PR #15, opened 2015-08-04
OverlayFS bug workaround

gablg1:
This PR enables the --overlayfs flag in CRIU which compensates for the OverlayFS bug in the Linux Kernel before version 4.2 (more about the bug at http://criu.org/Docker). More detailed information can be found in the comments below.


xemul:
We don't work via pull-requests (at least yet), all discussion happens on the mailing list :)
Thanks anyway.

PR #30, opened 2015-09-18
Remove multiple command validation as cpuinfo requires it

marcosnils:
`criu cpuinfo [dump | check]` can't be used through the command line as this validation kicks in. 
Other commands will fail due to required arguments so I believe it's not necessary anymore.

Signed-off-by: Marcos Lilljedahl <marcosnils@gmail.com>


xemul:
@marcosnils , thanks a lot for you contribution, it's really appreciated.

From the criu project start the contribution process is by sending (and sometimes discussing) the patches on the mailing list (criu@openvz.org). This is where all the criu developers are. That said, I have to I ask you to re-submit the patch to the mailing list.

Thanks in advance and sorry for inconvenience :)

PR #91, opened 2015-11-29
tmp-dir option

nwmcsween:
Compile tested.


xemul:
Hi.

Thanks for the contribution, it's really appreciated. However, CRIU development process is mailing-list based from the day 0, so would you please re-send the patches to criu@openvz.org.

Thanks!

PR #291, opened 2017-03-11
Speed up Travis jobs

kolyshkin:
Here are a few speedups I was able to come up with.

First, do kernel preparation (mostly git clone) in parallel with the main thread.
Second, use Google CDN to save a few seconds :)
Finally, use -j4 for make where appropriate.


kolyshkin:
Sorry, wrong repo.

PR #310, opened 2017-04-28
Update Makefile

mainboy:


avagin:
Could you add Signed-off-by line to the commit message?

In CRIU we have mailing-list based development, so all patches are to be sent to the criu@openvz.org list. For further contributions, please, send your patches there so that other people could also review the code. Thank you

xemul:
@avagin , so what's your decision on that?

PR #403, opened 2017-10-30
Fixed `else` statement

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. Warnings, found using PVS-Studio:
criu/compel/arch/x86/src/lib/infect.c       216     warn    V547 Expression '!sigframe->is_native' is always true.


avagin:
@lunixoid It is good that you found this error and fixed it. Thank you for this. Now you need to  submit this patch properly:

As minimum you need to write a more detailed commit message and add a signed-off line.

PR #413, opened 2017-11-20
glibc fixes

sorki:
Fixes needed for building with glibc 2.12.1. Not sure if this should go stable branch though.


adrianreber:
Thanks for the patch, but this has already been fixed on the master branch since July: d77d8cc402e597bb38edfb4e81db89fd7d96d0b6

PR #517, opened 2018-06-22
README.md correct travis links

grooverdan:
Codacy is unfixed currently as it looks like it isn't getting updates.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-[![master](https://travis-ci.org/xemul/criu.svg?branch=master)](https://travis-ci.org/xemul/criu)
-[![development](https://travis-ci.org/xemul/criu.svg?branch=criu-dev)](https://travis-ci.org/xemul/criu)
+[![master](https://travis-ci.org/checkpoint-restore/criu.svg?branch=master)](https://travis-ci.org/checkpoint-restore/criu)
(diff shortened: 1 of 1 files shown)


rppt:
@grooverdan, thanks for the fix.

We are old fashioned, though, and we send patches to criu@openvz.org mailing list rather than do GitHub pull requests.
Can you please send you patch to the list?

grooverdan:
Its a bit to go though for a most likely once-off uncontentious URL fix. Especially for a not-so-old fashioned dev who remembers it was something to do with ```git format-patch``` and even then was only partially successful. I might get to it, but I'd really prefer not to have to.

rst0git:
Hi @grooverdan

I send your patch to the mailing list
https://lists.openvz.org/pipermail/criu/2018-June/041767.html

FYI: `git send-email` makes it very easy to send a set patches to the mailing list. For example I send your patch with:
```
git send-email -1 --to criu@openvz.org --annotate --subject-prefix=PATCH
```

avagin:
Thanks for all who were involved.
https://github.com/checkpoint-restore/criu/commit/b24a54b63ef8c8ba497f7e1b20bb8096add0abe3

grooverdan:
thanks @rst0git

PR #523, opened 2018-07-08
Here are patches for CRIU v3.10

avagin:


avagin:
https://criu.org/Download/criu/3.10

Issue #557, opened 2018-09-19
Restore failed, can somebody explain the error message more convenient 


lovemyliwu:
Version: 2.12.1
```
 19688: Debug:  Setting 1 queue seq to 3907549365
 19688: Debug:  Setting 2 queue seq to 3670003764
 19688: Debug:  Restoring TCP options
 19688: Debug:          Will turn SAK on
 19688: Debug:          Will set snd_wscale to 9
 19688: Debug:          Will set rcv_wscale to 7
 19688: Debug: Will set mss clamp to 1444
 19688: Debug:  Setting 1 queue seq to 310756607
 19688: Debug:  Setting 2 queue seq to 212263591
 19688: Debug:  Restoring TCP options
 19688: Debug:          Will turn SAK on
 19688: Debug:          Will set snd_wscale to 9
 19688: Debug:          Will set rcv_wscale to 7
 19688: Debug: Will set mss clamp to 1444
 19688: Debug:  Setting 1 queue seq to 1873424399
 19688: Debug:  Setting 2 queue seq to 2635627206
 19688: Debug:  Restoring TCP options
 19688: Debug:          Will turn SAK on
 19688: Debug:          Will set snd_wscale to 9
 19688: Debug:          Will set rcv_wscale to 7
 19688: Debug: Will set mss clamp to 1444
 19688: Debug:  Setting 1 queue seq to 245079095
 19688: Debug:  Setting 2 queue seq to 3997547249
 19688: Debug:  Restoring TCP options
 19688: Debug:          Will turn SAK on
 19688: Debug:          Will set snd_wscale to 9
 19688: Debug:          Will set rcv_wscale to 7
 19688: Debug: Will set mss clamp to 1444
 19688: Debug:  Setting 1 queue seq to 2823590389
 19688: Debug:  Setting 2 queue seq to 4095220991
 19688: Debug:  Restoring TCP options
 19688: Debug:          Will turn SAK on
 19688: Debug:          Will set snd_wscale to 9
 19688: Debug:          Will set rcv_wscale to 7
 19688: Debug: Will set mss clamp to 1444
 19688: Debug:  Setting 1 queue seq to 1631574800
 19688: Debug:  Setting 2 queue seq to 454306393
 19688: Error: Can't connect inet socket back
Error (criu/cr-restore.c:1926): Restoring FAILED.
```

I find after some times retry, 1% rate chance to restore success.

adrianreber:
[...]

PR #564, opened 2018-10-19
add macro to ignore cant connect inet socket back error

lovemyliwu:
On my case, cant connect inet socket back error not the fatal error when do restore.

These changes related to #557


adrianreber:
Thanks for your patches. CRIU, however, does not use pull requests for patch submission. Could you send the patches as descried in https://criu.org/How_to_submit_patches

lovemyliwu:
@adrianreber ok

PR #573, opened 2018-11-07
Minor install cleanups

kolyshkin:
1. Do not install `.gitignore` to includes :)
2. Do not use `-rf` for `cp`

Also, hereby I nominate the first commit to be my entry to "the most trivial commit of the year" contest!


kolyshkin:
> Also, hereby I nominate the first commit to be my entry to "the most trivial commit of the year" contest!

With regrets, I retract my nomination, as the patch is not so trivial now :(

PR #602, opened 2019-02-02
c-lib: strdup for service_address and service_binary

martinwuehrer:
The functions criu_(local_)set_service_address and
criu_(local_)set_service_binary are the only functions that
do not create a copy of the given string arguments.
This may lead to problems, if the original string gets freed
but criu relies on them.

Additionally, the function criu_local_init_opts() assigns the
default service_binary now to opt->service_binary instead of
opt->service_address which is in my opinion the preferred way
since both are types of an anonymous union.

Signed-off-by: Martin Wührer <martin.wuehrer@artech.at>

### DIFF
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -35,6 +35,20 @@ struct criu_opts {
 static criu_opts *global_opts;
 static int saved_errno;
 
+void criu_free_service(criu_opts *opts)
+{
+	switch(opts->service_comm) {
+		case CRIU_COMM_SK:
+			free((void*)(opts->service_address));
+			break;
+        case CRIU_COMM_BIN:
+			free((void*)(opts->service_binary));
+			break;
+        default:
+			break;
+	}
+}
+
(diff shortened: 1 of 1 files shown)


lib/c/criu.c
```diff
@@ -47,10 +61,11 @@ void criu_set_service_comm(enum criu_service_comm comm)
 
 void criu_local_set_service_address(criu_opts *opts, const char *path)
 {
+	criu_free_service(opts);
 	if (path)
-		opts->service_address = path;
+		opts->service_address = strdup(path);
```
avagin: strdup() can return NULL, do we need to handle this?

avagin: There are more places where we don't check returns codes of strdup. Could you send another patch which will fix all of them?

martinwuehrer: This fix is added in #606. (0ed4db3)
But note, that it requires the change of the return type of the affected functions.

PR #605, opened 2019-02-05
Remove unused/unnecessary macro constants

rst0git:


criu/uffd.c
```diff
@@ -770,7 +770,7 @@ static int ud_open(int client, struct lazy_pages_info **_lpi)
 		pr_flags |= PR_REMOTE;
 	ret = open_page_read(lpi->pid, &lpi->pr, pr_flags);
 	if (ret <= 0) {
-		ret = -1;
+		pr_err("open_page_read error\n");
```
rppt: Nice catch for the unused `ret`.
For the error message
```
lp_err(lpi, "Failed to open pagemap\n");
```
 please

rst0git: Thanks, just wondering, when should we use `pr_perror` vs `pr_err` vs `lp_err`?

rppt: `lp_err` (and other `lp_<log>`) are uffd.c specific wrappers for `pr_err` etc that print userfault file  descriptor and the PID of the process it is attached to. I'd say they should be used in contexts related to a certain `lazy_pages_info` instance.

`pr_perror` is CRIU version of `perror` and it should be used when `errno` value has meaning, e.g. after failing a `syscall`.

PR #606, opened 2019-02-06
c-lib: added strdup-null checks, simplify the interface

martinwuehrer:
This pull request consists of two commits. The first one contains missing `strudp()` checks (as requested in #602)

The other commit is a proposal that simplifies the c-interface, as it does not require to call `*set_service_comm()` anymore.
I think it would be a good idea to omit these calls in favour of `*set_service_address()`, `*set_service_fd()` or `*set_service_binary()`. As these calls already set the intention of the library user.
And it can not happen that the user sets the wrong service_comm value to the `set_service_*()` calls.
E.g. If `*set_service_comm()` was set to socket, but a string to the binary was given
via `*set_service_binary()` it would probably cause to undefined behavior that can be prevented by a better library design.

### DIFF
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -59,18 +59,23 @@ void criu_set_service_comm(enum criu_service_comm comm)
 	criu_local_set_service_comm(global_opts, comm);
 }
 
-void criu_local_set_service_address(criu_opts *opts, const char *path)
+int criu_local_set_service_address(criu_opts *opts, const char *path)
 {
 	criu_free_service(opts);
-	if (path)
+	if (path) {
 		opts->service_address = strdup(path);
-	else
+	} else {
 		opts->service_address = strdup(CR_DEFAULT_SERVICE_ADDRESS);
+	}
+	if(opts->service_address == NULL) {
+		return -ENOMEM;
+	}
+	return 0;
 }
 
-void criu_set_service_address(const char *path)
+int criu_set_service_address(const char *path)
 {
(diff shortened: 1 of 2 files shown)


adrianreber:
There is already a commit since 3.11 which breaks ABI (not sure if the switch from char to const char counts as API change) as far as I can tell, so we need to bump the SO version for the next release anyway (https://github.com/checkpoint-restore/criu/commit/a5370ca536944962793784a36528452ca841563c). Although, until now the break was minimal.

This now definitely breaks API. The last time that I used libcriu was probably in 2013. So I cannot tell if this newly proposed interface is better than the old interface.

Another question is, are there actually many libcriu users for whom this change will be a problem.

All CRIU users I am aware of are using the CLI or the RPC interface. So I cannot tell if this newly proposed interface is better and worth breaking API. But I also do not know if there are enough libcriu users for us to be worried that breaking API and bumping SO name is really a problem for someone.

avagin:
> Another question is, are there actually many libcriu users for whom this change will be a problem.

I don't think that we have users who use libcriu in production.

> All CRIU users I am aware of are using the CLI or the RPC interface. So I cannot tell if this newly proposed interface is better and worth breaking API. But I also do not know if there are enough libcriu users for us to be worried that breaking API and bumping SO name is really a problem for someone.

For me, all these changes look reasonable, so I vote to merge them.

adrianreber:
I am also in favour of merging. We just need to remember to increase the SO version before 3.12.

PR #607, opened 2019-02-07
cr-service: Don't log return value of parse_options()

rst0git:
The function parse_options() should return an integer 0, 1 or 2, where:

* 0 := Success
* 1 := Failure
* 2 := CRIU should display help text

The current behaviour always creates a warning message which shows the return value of that function. With this change only the values that are different than 0 (Success) will be logged.

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -277,7 +277,6 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 
 		rpc_cfg_file = req->config_file;
 		i = parse_options(0, NULL, &dummy, &dummy, PARSING_RPC_CONF);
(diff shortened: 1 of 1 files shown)


adrianreber:
Yes, this is anyway just a debug printf from when I wrote it. It just appeared recently when we started to print out messages before logging was initialized. We can either remove it completely or handle it like in this PR.

rst0git:
>  It just appeared recently when we started to print out messages before logging was initialized. We can either remove it completely or handle it like in this PR.

It make sense to just remove it (instead if handling it that way) because if the return value is not 0 CRIU will fail, and since the logging was not initialized, this message will be lost anyway.

PR #608, opened 2019-02-08
mount: realloc info->source when resolve ext mount

rst0git:
Use `realloc()` to change the size of the memory block pointed to by `info->source` instead of allocating new memory area and releasing the old one. If `info->source` is NULL then the call is equivalent to `malloc(size)`.

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -491,14 +491,13 @@ static int try_resolve_ext_mount(struct mount_info *info)
 			int len;
 
 			len = strlen(val) + sizeof("dev[]");
-			source = xmalloc(len);
+			source = xrealloc(info->source, len);
 			if (source == NULL)
 				return -1;
 
 			snprintf(source, len, "dev[%s]", val);
 			info->fstype = fstype_auto();
 			BUG_ON(info->fstype->code != FSTYPE__AUTO);
-			xfree(info->source);
 			info->source = source;
 			return 1;
 		}


criu/mount.c
```diff
@@ -487,19 +487,16 @@ static int try_resolve_ext_mount(struct mount_info *info)
...
 		val = external_lookup_by_key(devstr);
 		if (!IS_ERR_OR_NULL(val)) {
-			char *source;
 			int len;
 
 			len = strlen(val) + sizeof("dev[]");
-			source = xmalloc(len);
-			if (source == NULL)
+			info->source = xrealloc(info->source, len);
+			if (info->source == NULL)
```
0x7f454c46: realloc() doesn't touch original memory if re-allocation failed
That means that info->source original pointer will leak now.
It's not that we're precise about alloc/free in case of error in criu..
But don't see the reason to add additional leak for a fallback.
Otherwise, the patch makes sense to me.

rst0git: > realloc() doesn't touch original memory if re-allocation failed
That means that info->source original pointer will leak now.

Thanks for catching that, I will update the commit to fix it.

> It's not that we're precise about alloc/free in case of error in criu..
But don't see the reason to add additional leak for a fallback.

If we are to be precise about alloc/free in case of error then
we could use tools like Valgrind to check for memory leaks.

criu/mount.c
```diff
@@ -491,15 +491,15 @@ static int try_resolve_ext_mount(struct mount_info *info)
 			int len;
 
 			len = strlen(val) + sizeof("dev[]");
-			source = xmalloc(len);
-			if (source == NULL)
+			source = xrealloc(info->source, len);
+			if (source == NULL) {
+				xfree(info->source);
```
avagin: It is dangerous. mnt_entry_free() will call xfree(info->source) again and it will be a double free fault. You don't need to call xfree here or you need to set info->source to NULL after that. I would prefer to not call xfree.

PR #611, opened 2019-02-11
remote: Ignore interupt signals

rst0git:
When an interrupt signal ([even `SIGWINCH` when `strace` is used](https://unix.stackexchange.com/a/465194)) is received while epoll_wait() sleeps, it will return a value of -1 and set errno to `EINTR`, which is not an error and should be ignored.

### DIFF
--- a/criu/img-remote.c
+++ b/criu/img-remote.c
@@ -832,7 +832,9 @@ void accept_image_connections() {
 		int n_events, i;
 
 		n_events = epoll_wait(epoll_fd, events, EPOLL_MAX_EVENTS, 250);
-		if (n_events < 0) {
+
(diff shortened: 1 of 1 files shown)


criu/img-remote.c
```diff
@@ -832,7 +832,7 @@ void accept_image_connections() {
 		int n_events, i;
 
 		n_events = epoll_wait(epoll_fd, events, EPOLL_MAX_EVENTS, 250);
-		if (n_events < 0) {
+		if (n_events < 0 && errno != EINTR) {
```
avagin: can you add a comment that epoll_wait isn't restarted after interrupted by a signal?

avagin: http://man7.org/linux/man-pages/man7/signal.7.html

PR #615, opened 2019-02-13
 A few tweaks and cleanups in criu/img-*.c

rst0git:


criu/img-cache.c
```diff
@@ -1,13 +1,10 @@
...
-#include "img-remote.h"
-#include "criu-log.h"
-#include <sys/socket.h>
-#include <netinet/in.h>
-#include <fcntl.h>
 #include "cr_options.h"
+#include "img-remote.h"
 #include "util.h"
 
+#include <sys/socket.h>
```
avagin: Can we move linux headers before criu headers?

rst0git:
I am closing this because I am working on a patch series that aims to integrate image-proxy and image-cache with dump/restore.

The idea is that a user should be able to perform the following operations to achieve live migration:
```sh
# Example 1

dst# criu restore --remote --port <port> ...
src# criu dump --remote --address <src> --port <port> ...

# Example 2

dst# criu restore --remote --ps-socket <fd> ...
src# criu pre-dump --remote --ps-socket <fd> ...
src# criu pre-dump --remote --ps-socket <fd> ...
src# criu dump --remote --ps-socket <fd> ...

# Example 3

dst# criu restore --remote --lazy-pages --port <port> ...
src# criu dump --remote --lazy-pages --address <src> --port <port> ...
```

PR #616, opened 2019-02-14
Fix pathname passed to access()

ashu-mehra:
The pathname passed to access() in check_path_remap() should
not include the "." prefix present in link names.

Signed-off-by: Ashutosh Mehra <asmehra1@in.ibm.com>

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -1178,6 +1178,9 @@ static int check_path_remap(struct fd_link *link, const struct fd_parms *parms,
 		 */
 		if (pid != 0) {
 			bool is_dead = strip_deleted(link);
+			mntns_root = mntns_get_root_fd(nsid);
+			if (mntns_root < 0)
+				return -1;
 
 			/* /proc/<pid> will be "/proc/1 (deleted)" when it is
 			 * dead, but a path like /proc/1/mountinfo won't have
@@ -1189,7 +1192,7 @@ static int check_path_remap(struct fd_link *link, const struct fd_parms *parms,
 			 */
 			if (!is_dead) {
 				*end = 0;
-				is_dead = access(rpath, F_OK);
+				is_dead = faccessat(mntns_root, rpath, F_OK, 0);
 				*end = '/';
 			}
 
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -209,6 +209,7 @@ TST_NOFILE	:=				\
 		unlink_multiple_largefiles	\
 		config_inotify_irmap		\
 		thp_disable			\
+		pid_file			\
 #		jobctl00			\
 
 ifneq ($(SRCARCH),arm)
--- a/test/zdtm/static/pid_file.c
+++ b/test/zdtm/static/pid_file.c
@@ -1,0 +1,52 @@
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc 	= "Check that environment didn't change";
+const char *test_author	= "Andrei Vagin <avagin@gmail.com>";
+
+int main(int argc, char **argv)
+{
+	int fd, fd2;
+	struct stat st, st2;
+
+	test_init(argc, argv);
+
+	fd = open("/proc/1/status", O_RDONLY);
+	if (fd < 0) {
+		pr_perror("Unable to open /proc/1/status");
+		return 1;
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	fd2 = open("/proc/1/status", O_RDONLY);
+	if (fd2 < 0) {
+		pr_perror("Unable to open /proc/1/status");
+		return 1;
+	}
+	if (fstat(fd, &st)) {
+		pr_perror("fstat");
+		return 1;
+	}
+	if (fstat(fd2, &st2)) {
+		pr_perror("fstat");
+		return 1;
+	}
+	close(fd);
+	close(fd2);
+
+	if (st.st_ino != st2.st_ino) {
+		fail("inode numbers mismatch");
+		return 1;
+	}
+
+	pass();
+	return 0;
+}


ashu-mehra:
Not sure why the travis-ci build for centos failed. The build for my fork passed fine: https://travis-ci.org/ashu-mehra/criu/builds/493116600

adrianreber:
I have restarted the CentOS test case.

criu/files-reg.c
```diff
@@ -1189,7 +1189,7 @@ static int check_path_remap(struct fd_link *link, const struct fd_parms *parms,
 			 */
 			if (!is_dead) {
 				*end = 0;
-				is_dead = access(rpath, F_OK);
+				is_dead = access(&rpath[1], F_OK); /* Ignore '.' at the start of the filename */
```
rppt: I'm not really familiar with that code, but does `rpath` here always starts with `.`?

ashu-mehra: Before the call to `check_path_remap`, the caller checks the path is regular by checking character at index 1 with '/':
https://github.com/checkpoint-restore/criu/blob/1a9bb4c911dd00e66d19e7a219e1caf252ce9c96/criu/files-reg.c#L1352
So the caller is also expecting the path name to start with `.`

avagin: Can you explain why should not the pathname  include the "." prefix? How did you find this problem? How can we reproduce it? Can we improve one of our test or create a new one to test this code?

ashu-mehra: I got a simple testcase to recreate this:
```
#include <stdio.h>
#include <unistd.h>

int main(void) {
	FILE *cgroupFile = fopen("/proc/1/cgroup", "r");
	int count = 0;
	while (1) {
		printf("%d\n", count);
		count += 1;
		sleep(5);
	}
	if (NULL != cgroupFile) {
		fclose(cgroupFile);
	}
}
```

I am trying to checkpoint and restore this process in a docker container based on ubuntu:18.04. Dumping the process worked fine but restore gives following error:
```
(00.014639)     13: Restoring 13 to 13 pgid 
(00.015036)     13: Error (criu/cr-restore.c:1542): Pid 2 do not match expected 1
(00.015684)     13: Error (criu/cr-restore.c:1298): 2 exited, status=1
(00.015766) Error (criu/cr-restore.c:2171): Restoring FAILED
```

In dump.log I noticed following messages which made me suspicious:
```
(00.015204) Dumping opened files (pid: 13)
(00.015223) ----------------------------------------
(00.015353) 13 fdinfo 0: pos:                0 flags:           100000/0
(00.015409) Dumping path for 0 fd via self 8 [/dev/null]
(00.015554) 13 fdinfo 1: pos:                0 flags:           100001/0
(00.015605) Dumping path for 1 fd via self 10 [/root/log]
(00.015665) 13 fdinfo 2: pos:                0 flags:           100001/0
(00.015724) 13 fdinfo 3: pos:                0 flags:           100000/0
(00.015755) Dumping path for 3 fd via self 12 [/proc/1/cgroup]
(00.015774) Dumping dead process remap of 1
(00.015847) ----------------------------------------
```

Looks like criu determined pid 1 as dead. This statement is generated by `check_path_remap` for `/proc/1/cgroup` file. When I debugged the code, I found the call to `access()` is failing even though `/proc/1/cgroup` still exists. That's when I realized we are actually passing `./proc/1/cgroup` to `access()`. After adding this fix, I am able to restore the process properly.
Hope this helps.

avagin: @ashu-mehra Now I understand the problem. Thank you for the explanation. I think you fix doesn't work correctly for processes in another mount namespaces. because criu will try to resolve this patch in a criu mount namespace. Could you try this:
```diff --git a/criu/files-reg.c b/criu/files-reg.c
index e9dcf5efa..b30da8c99 100644
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -1178,6 +1178,9 @@ static int check_path_remap(struct fd_link *link, const struct fd_parms *parms,
                 */
                if (pid != 0) {
                        bool is_dead = strip_deleted(link);
+                       mntns_root = mntns_get_root_fd(nsid);
+                       if (mntns_root < 0)
+                               return -1;
 
                        /* /proc/<pid> will be "/proc/1 (deleted)" when it is
                         * dead, but a path like /proc/1/mountinfo won't have
@@ -1189,7 +1192,7 @@ static int check_path_remap(struct fd_link *link, const struct fd_parms *parms,
                         */
                        if (!is_dead) {
                                *end = 0;
-                               is_dead = access(rpath, F_OK);
+                               is_dead = faccessat(mntns_root, rpath, F_OK, 0);
                                *end = '/';
                        }
```

avagin: And here is a test:
```
diff --git a/test/zdtm/static/Makefile b/test/zdtm/static/Makefile
index a31c202d5..912168cfc 100644
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -209,6 +209,7 @@ TST_NOFILE  :=                              \
                unlink_multiple_largefiles      \
                config_inotify_irmap            \
                thp_disable                     \
+               pid_file                        \
 #              jobctl00                        \
 
 ifneq ($(SRCARCH),arm)
diff --git a/test/zdtm/static/pid_file.c b/test/zdtm/static/pid_file.c
new file mode 100644
index 000000000..b89dc1790
--- /dev/null
+++ b/test/zdtm/static/pid_file.c
@@ -0,0 +1,53 @@
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+
+#include "zdtmtst.h"
+
+const char *test_doc   = "Check that environment didn't change";
+const char *test_author        = "Andrei Vagin <avagin@gmail.com>";
+
+{
+       int fd, fd2;
+       struct stat st, st2;
+
+       test_init(argc, argv);
+
+       fd = open("/proc/1/status", O_RDONLY);
+       if (fd < 0) {
+               pr_perror("Unable to open /proc/1/cgroup");
+               return 1;
+       }
+
+       test_daemon();
+       test_waitsig();
+
+       fd2 = open("/proc/1/status", O_RDONLY);
+       if (fd2 < 0) {
+               pr_perror("Unable to open /proc/1/cgroup");
+               return 1;
+       }
+       if (fstat(fd, &st)) {
+               pr_perror("fstat");
+               return 1;
+       }
+       if (fstat(fd2, &st2)) {
+               pr_perror("fstat");
+               return 1;
+       }
+       close(fd);
+       close(fd2);
+
+       if (st.st_ino != st2.st_ino) {
+               fail("inode numbers mismatch");
+               return 1;
+       }
+
+       pass();
+       return 0;
+}
```

You can use the next command to validate your changes:
```
python test/zdtm.py run -t zdtm/static/pid_file
```

ashu-mehra: @avagin Sorry for responding late. I verified your fix and it works fine for my use-case. I think its better I close this PR, and you can create a new PR with your fix and test changes.

ashu-mehra:
Closing this as @avagin has a better fix for the same issue.

avagin:
@ashu-mehra I think you misunderstand me. You found this issue and created a fix. Your fix is right, but doesn't handle all case. I just showed how to do it properly. So feel free to update your patch and submit it here.

ashu-mehra:
I was busy with other stuff at work. I will try to address it this week.

ashu-mehra:
Note to myself:
When I ran `python test/zdtm.py run -t zdtm/static/pid_file`, I get following error message:
```
# ./test/zdtm.py run -t zdtm/static/pid_file
The kernel is tainted: '4097\n'
  File "./test/zdtm.py", line 2303, in <module>
    opts['action'](opts)
  File "./test/zdtm.py", line 1972, in run_tests
    launcher = Launcher(opts, len(torun))
  File "./test/zdtm.py", line 1668, in __init__
    raise Exception("The kernel is tainted: %r" % self.__taint)
Exception: The kernel is tainted: '4097\n'
```

This can be worked around by adding `--ignore-taint` flag as:
`python test/zdtm.py run --ignore-taint -t zdtm/static/pid_file`

ashu-mehra:
@avagin I have updated the change set as per your suggestions. Please review.

avagin:
Merged. Thanks a lot!

ashu-mehra:
@avagin Thanks for your help in fixing the issue!

PR #617, opened 2019-02-15
socket-tcp-skip-in-flight: Don't fail on EAGAIN

rst0git:
The server socket is marked as non-blocking, and if the client doesn't connect, `accept()` will fail with `EAGAIN` (or `EWOULDBLOCK`), which is not an error, and we should try again.

### DIFF
--- a/test/zdtm/static/socket-tcp-skip-in-flight.c
+++ b/test/zdtm/static/socket-tcp-skip-in-flight.c
@@ -1,3 +1,4 @@
+#include <poll.h>
 #include "zdtmtst.h"
 
 #ifdef ZDTM_IPV4V6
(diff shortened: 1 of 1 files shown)


test/zdtm/static/socket-tcp-skip-in-flight.c
```diff
@@ -64,11 +64,13 @@ int main(int argc, char **argv)
 		return -1;
 	}
 
-	fd = tcp_accept_server(fd_s);
-	if (fd < 0) {
-		fail("Unable to accept a new connection");
-		return 1;
-	}
+	do {
+		fd = tcp_accept_server(fd_s);
```
avagin: I think it would be better to use poll or epoll here.

test/zdtm/static/socket-tcp-skip-in-flight.c
```diff
@@ -64,12 +66,15 @@ int main(int argc, char **argv)
 		return -1;
 	}
 
+	memset(poll_set, '\0', sizeof(poll_set));
+	poll_set[0].fd = fd_s;
+	poll_set[0].events = POLLIN;
+	poll(poll_set, 1, -1);
```
avagin: I think it would be better to handle errors of poll...

PR #621, opened 2019-02-18
crit: display help message when using python3

adrianreber:
Running crit with python2 gives following minimal help message:

 $ crit/crit
 usage: crit [-h] {decode,encode,info,x,show} ...
 crit: error: too few arguments

Using a python3 only system crit shows the following error:

 $ crit/crit
   File "crit/crit", line 6, in <module>
     cli.main()
   File "/home/criu/crit/pycriu/cli.py", line 334, in main
     opts["func"](opts)
 KeyError: 'func'

Using this patch the python3 output changes to:

 $ crit/crit
 usage: crit [-h] {decode,encode,info,x,show} ...
 crit: error: the following arguments are required: parser

The help message is not 100% the same but almost. Better than a
traceback in any way.

### DIFF
--- a/lib/py/cli.py
+++ b/lib/py/cli.py
@@ -331,6 +331,11 @@ def main():
 
 	opts = vars(parser.parse_args())
 
+	if not opts:
+		sys.stderr.write(parser.format_usage())
+		sys.stderr.write("crit: error: too few arguments\n")
+		sys.exit(1)
+
 	opts["func"](opts)
 
 if __name__ == '__main__':


avagin:
> crit: error: the following arguments are required: parser

This error is unclear. We have nothing about the parser argument in the help message. What do you think about something like this:

```
diff --git a/lib/py/cli.py b/lib/py/cli.py
index 12597dbfb..abaf0720c 100755
--- a/lib/py/cli.py
+++ b/lib/py/cli.py
@@ -331,6 +331,11 @@ def main():
 
        opts = vars(parser.parse_args())
 
+       if not opts:
+               sys.stderr.write(parser.format_usage())
+               sys.stderr.write("crit: error: too few arguments\n")
+               sys.exit(1)
+
        opts["func"](opts)
 
 if __name__ == '__main__':
```

adrianreber:
Your suggestion was also what my first version looked like, but then the behaviour of the python2 and then python3 version are different. I can change it to what you suggested.

adrianreber:
@avagin Updated the PR.

adrianreber:
> > Your suggestion was also what my first version looked like, but then the behavior of the python2 and then python3 version are different. I can change it to what you suggested.
> 
> I am not sure that I understand what you mean. Can you elaborate with more details?

Sorry, the comment was no totally correct. My initial patch was printing out the help and not the usage information using python3. That way the output was different. Your suggestion produces exactly the same output. You can disregard my comment.

PR #633, opened 2019-02-24
util: Introduce fd_set_nonblocking()

rst0git:
Combine the functionality of `socket_set_non_blocking()` and `socket_set_blocking()` into a new function, called `fd_set_nonblocking()`, and move it in `criu/util.c` to enable reusability throughout the code base.

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -1139,6 +1139,24 @@ int fd_has_data(int lfd)
 	return ret;
 }
 
+void fd_set_nonblocking(int fd, bool on)
+{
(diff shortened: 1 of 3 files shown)


criu/util.c
```diff
@@ -1139,6 +1139,24 @@ int fd_has_data(int lfd)
 	return ret;
 }
 
+void fd_nonblocking(int fd, bool on)
```
avagin: maybe fd_set_nonblocking?

rst0git: Thank you, fd_set_nonblocking is better.

PR #634, opened 2019-02-24
Sort includes in criu/img-*.c

rst0git:
Sort and remove unused/unnecessary include stements in `criu/img-*.c`

### DIFF
--- a/criu/img-proxy.c
+++ b/criu/img-proxy.c
@@ -1,10 +1,8 @@
 #include <unistd.h>
 
-#include "img-remote.h"
-#include "criu-log.h"
-#include <fcntl.h>
-#include <sys/socket.h>
 #include "cr_options.h"
+#include "criu-log.h"
+#include "img-remote.h"
 #include "util.h"
 
 int image_proxy(bool background, char *local_proxy_path)
--- a/criu/img-remote.c
+++ b/criu/img-remote.c
@@ -1,26 +1,16 @@
-#include <unistd.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <sys/types.h>
+#include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/epoll.h>
-#include <netinet/in.h>
-#include <netdb.h>
-#include "xmalloc.h"
-#include "criu-log.h"
-#include "img-remote.h"
-#include "images/remote-image.pb-c.h"
-#include "protobuf-desc.h"
-#include <fcntl.h>
-#include "servicefd.h"
-#include "common/compiler.h"
+#include <sys/un.h>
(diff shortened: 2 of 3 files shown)


criu/img-remote.c
```diff
@@ -1,26 +1,12 @@
-#include <unistd.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <sys/types.h>
-#include <sys/socket.h>
```
avagin: In this file, sys/socket.h is included here because socket() is called in this file.

avagin: We should not rely on criu headers in such cases. Maybe we should not include linux headers in criu headers...

rst0git: I agree that headers such as `sys/socket.h` and `sys/epoll.h` are used in this file and it is good to have them listed at the beginning as a hint about what functionalities are being used in the implementation that follows. However, if we were to do that consistently we will end up with a lot of duplicated code between *.c files and it is likely (in the future) for a developer to forget to remove an include statement when they remove a specific function call.
@avagin Considering both pros and cons, which headers do you think should be kept?
(I am happy to update the patch.)

avagin: > However, if we were to do that consistently we will end up with a lot of duplicated code between *.c files and it is likely (in the future) for a developer to forget to remove an include statement when they remove a specific function call.

Do you know where sys/socket.h is included now? Is it a side effect? Does the same side effect exists in other libc-s?

criu/img-remote.c
```diff
@@ -1,26 +1,14 @@
...
-#include "xmalloc.h"
-#include "criu-log.h"
+#include <sys/un.h>
+#include <unistd.h>
+
 #include "img-remote.h"
-#include "images/remote-image.pb-c.h"
-#include "protobuf-desc.h"
-#include <fcntl.h>
-#include "servicefd.h"
```
avagin: get_service_fd is used in this file

criu/img-remote.c
```diff
@@ -1,26 +1,14 @@
-#include <unistd.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <sys/types.h>
+#include <netinet/in.h>
 #include <sys/socket.h>
 #include <sys/epoll.h>
-#include <netinet/in.h>
-#include <netdb.h>
-#include "xmalloc.h"
```
avagin: xmalloc is used in this file

criu/img-proxy.c
```diff
@@ -1,10 +1,7 @@
 #include <unistd.h>
 
-#include "img-remote.h"
-#include "criu-log.h"
```
avagin: pr_info, pr_perror are used in this file

PR #638, opened 2019-03-01
c-lib: Support to build a static archive

udzura:
This will be useful to bind libcriu features into one binary. e.g. container runtime or some ops tools.

Would anyone have time to check this out?

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -6,7 +6,7 @@ UAPI_HEADERS		:= lib/c/criu.h images/rpc.proto
 # File to keep track of files installed by setup.py
 CRIT_SETUP_FILES	:= lib/.crit-setup.files
 
-all-y	+= lib-c lib-py
+all-y	+= lib-c lib-a lib-py
 
 #
 # C language bindings.
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -42,7 +42,7 @@ clean-lib:
 	$(Q) $(MAKE) $(build)=lib/py clean
 .PHONY: clean-lib
 clean: clean-lib
-cleanup-y	+= lib/c/$(CRIU_SO) lib/c/criu.pc
+cleanup-y	+= lib/c/$(CRIU_SO) lib/c/$(CRIU_A) lib/c/criu.pc
 mrproper: clean
 
 install: lib-c lib-py crit/crit lib/c/criu.pc.in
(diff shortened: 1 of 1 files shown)


lib/Makefile
```diff
@@ -19,8 +20,12 @@ ldflags-so		+= -lprotobuf-c
 lib/c/$(CRIU_SO): lib/c/built-in.o
 	$(call msg-link, $@)
 	$(Q) $(CC) -shared $(cflags-so) -o $@ $^ $(ldflags-so) $(LDFLAGS)
+lib/c/$(CRIU_A): lib/c/built-in.o
+	$(call msg-link, $@)
+	$(Q) $(AR) rcs $@ $^
 lib-c: lib/c/$(CRIU_SO)
-.PHONY: lib-c
+lib-a: lib/c/$(CRIU_A)
```
0x7f454c46: don't you need to add it to `all-y` to build .a?

cyrillos: yes, it should be added into `all-y` target in the same file

0x7f454c46: Probably, it's also worth to add `install` rule if we plan providing .a library.

udzura: Thabjs for feedback! I've added a commit.

avagin:
ping @0x7f454c46 @cyrillos

cyrillos:
ack

avagin:
@udzura pls take a look what fails in travis:
https://travis-ci.org/checkpoint-restore/criu/jobs/503457651

rst0git:
> @udzura pls take a look what fails in travis:

The failure is because `lib/c/libcriu.a` was not removed after `make clean`. This should fix it.

```diff
diff --git a/lib/Makefile b/lib/Makefile
index 7b95956a..4a131e88 100644
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -42,7 +42,7 @@ clean-lib:
        $(Q) $(MAKE) $(build)=lib/py clean
 .PHONY: clean-lib
 clean: clean-lib
-cleanup-y      += lib/c/$(CRIU_SO) lib/c/criu.pc
+cleanup-y      += lib/c/$(CRIU_SO) lib/c/$(CRIU_A) lib/c/criu.pc
 mrproper: clean

 install: lib-c lib-py crit/crit lib/c/criu.pc.in
```

udzura:
Thanks for advice!! Sorry for late to reply. I will try to rebase.

rst0git:
@udzura thank you for updating the PR!

Would it be possible also to add `CRIU_A` to the `install` and `uninstall` phony targets?

udzura:
Pushed dbe4527 . I will revert this commit if it is judged to be unnecessary.

cyrillos:
Looks ok to me

avagin:
Applied. Thanks!

PR #640, opened 2019-03-02
[WIP] util: Enable socket zero-copy transmission

rst0git:
The `MSG_ZEROCOPY` flag enables copy avoidance for socket send calls.

The kernel is permissive when applications pass undefined flags to the send system call. By default it simply ignores these.

https://www.kernel.org/doc/html/latest/networking/msg_zerocopy.html

### DIFF
--- a/criu/img-remote.c
+++ b/criu/img-remote.c
@@ -940,11 +940,20 @@ static int64_t send_image_async(struct roperation *op)
 	struct rimage *rimg = op->rimg;
 	bool close_fd = op->close_fd;
 	int n;
+	int flags = 0;
 
-	n = write(
+#ifdef MSG_ZEROCOPY
+	flags |= MSG_ZEROCOPY;
+#endif
+
+	n = send(
 		fd,
 		op->curr_sent_buf->buffer + op->curr_sent_bytes,
-		min(BUF_SIZE, op->curr_sent_buf->nbytes) - op->curr_sent_bytes);
+		min(BUF_SIZE, op->curr_sent_buf->nbytes) - op->curr_sent_bytes,
+		flags);
+
+	if (do_read_notification(fd))
+		return -1;
 
 	if (n > -1) {
 		op->curr_sent_bytes += n;
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -136,12 +136,21 @@ static inline int send_psi_flags(int sk, struct page_server_iov *pi, int flags)
 		return -1;
 	}
 
+#ifdef MSG_ZEROCOPY
+	if (flags & MSG_ZEROCOPY && do_read_notification(sk))
+		return -1;
+#endif
+
 	return 0;
 }
 
 static inline int send_psi(int sk, struct page_server_iov *pi)
 {
-	return send_psi_flags(sk, pi, 0);
+	int flags = 0;
+#ifdef MSG_ZEROCOPY
+	flags |= MSG_ZEROCOPY;
+#endif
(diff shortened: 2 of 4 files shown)


rst0git:
cc: @rppt

criu/img-remote.c
```diff
@@ -940,11 +940,20 @@ static int64_t send_image_async(struct roperation *op)
...
+#endif
 
-	n = write(
+	n = send(
 		fd,
 		op->curr_sent_buf->buffer + op->curr_sent_bytes,
-		min(BUF_SIZE, op->curr_sent_buf->nbytes) - op->curr_sent_bytes);
+		min(BUF_SIZE, op->curr_sent_buf->nbytes) - op->curr_sent_bytes,
+		flags);
+
```
rppt: @rst0git , are you sure it's worth the effort?

The kernel docs say: "MSG_ZEROCOPY is generally only effective at writes over around 10 KB" and much of the transfers here would be way less than 10 KB...

rst0git: That is a good point, here we only send the image header which has very small size.

criu/page-xfer.c
```diff
@@ -136,12 +136,21 @@ static inline int send_psi_flags(int sk, struct page_server_iov *pi, int flags)
...
 	return 0;
 }
 
 static inline int send_psi(int sk, struct page_server_iov *pi)
 {
-	return send_psi_flags(sk, pi, 0);
+	int flags = 0;
+#ifdef MSG_ZEROCOPY
+	flags |= MSG_ZEROCOPY;
+#endif
```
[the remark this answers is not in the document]
rst0git: Thanks, I will update the patch.

criu/util.c
```diff
@@ -1312,6 +1346,13 @@ int setup_tcp_server(char *type)
 		goto out;
 	}
 
+#ifdef SO_ZEROCOPY
```
avagin: An existence of SO_ZEROCOPY doesn't guarantee that zerocopy is supported by kernel and vice-versa.

criu/util.c
```diff
@@ -1286,6 +1287,39 @@ static int get_sockaddr_in(struct sockaddr_storage *addr, char *host)
 	return 0;
 }
 
+int do_read_notification(int fd)
```
avagin: we need to add a comment here which explains what this function does

criu/page-xfer.c
```diff
@@ -1242,6 +1251,10 @@ int connect_to_page_server_to_recv(int epfd)
 
 int request_remote_pages(unsigned long img_id, unsigned long addr, int nr_pages)
 {
+	int flags = MSG_DONTWAIT;
+#ifdef MSG_ZEROCOPY
```
avagin: this should be a runtime check. CRIU can't be compiled in a docker container which is running on an old kernel.

criu/img-remote.c
```diff
@@ -940,11 +940,20 @@ static int64_t send_image_async(struct roperation *op)
...
+#ifdef MSG_ZEROCOPY
+	flags |= MSG_ZEROCOPY;
+#endif
 
-	n = write(
+	n = send(
 		fd,
 		op->curr_sent_buf->buffer + op->curr_sent_bytes,
-		min(BUF_SIZE, op->curr_sent_buf->nbytes) - op->curr_sent_bytes);
```
avagin: Now op->curr_sent_buf->buffer can't be modifies immediately after this call. We need a comment here which explains life-time of this buffer.

PR #646, opened 2019-03-09
Set RLIMIT_NOFILE to max during pre-dump

rst0git:


rst0git:
> I'm just curious, why this was not done from the beginning.

Actually it was done from the beginning, and then it was changed (https://github.com/checkpoint-restore/criu/commit/a148e87469a36dc75e263259e39ed30336343fc8), I assume because during `criu dump` we use `CHUNK_MODE` and we do not need to set `RLIMIT_NOFILE` to max.

criu/page-pipe.c
```diff
@@ -120,7 +92,12 @@ static struct page_pipe_buf *ppb_alloc(struct page_pipe *pp,
 		}
 		cnt_add(CNT_PAGE_PIPES, 1);
 
-		ppb->pipe_size = fcntl(ppb->p[0], F_GETPIPE_SZ, 0) / PAGE_SIZE;
+		if (fcntl(ppb->p[0], F_SETPIPE_SZ, PIPE_MAX_SIZE * PAGE_SIZE) < 0) {
```
avagin: This operation will require to allocate 8 sequential pages in the kernel space, so it has a good chance to fail. I think it was a reason why we have this logic when a pipe size is changed step by step.

rst0git: Looking at [`pipe_set_size()`](https://github.com/torvalds/linux/blob/ea295481b6e313b4ea3ca2720ffcafd6005b5643/fs/pipe.c#L1027L1109), I noticed that when we grow a pipe's buffer the kernel would just allocate a new buffer and free the old one.
However, in the current implementation we never actually set the pipe's buffer size to `PIPE_MAX_SIZE`:
```C
	unsigned long new_size = ppb->pipe_size << 1;
...
	if (new_size > PIPE_MAX_SIZE)
                return 1;
```
> This operation will require to allocate 8 sequential pages in the kernel space, so it has a good chance to fail. I think it was a reason why we have this logic when a pipe size is changed step by step.

Do you mean that we should set the pipe's buffer to a value less than `PIPE_MAX_SIZE` or we should try to handle `ENOMEM` by doing so?

avagin: > Looking at pipe_set_size(), I noticed that when we grow a pipe's buffer the kernel would just allocate a new buffer and free the old one.

You are right. And the kernel uses kmalloc to allocate this buffer. kcalloc doesn't handle well big allocations. You can try to rework this code to use kvmalloc instead of kcalloc.

> However, in the current implementation we never actually set the pipe's buffer size to PIPE_MAX_SIZE:

You can fix this ;)

> Do you mean that we should set the pipe's buffer to a value less than PIPE_MAX_SIZE or we should try to handle ENOMEM by doing so?

Here are two points:
1. We should not allocate more than we need. A process can be small and we don't need "big" pipe for iot. Each time when we set a pipe size to PIPE_MAX_SIZE, we reduce our chance to set this size to the next pipe.
2. Right now we handle "ENOMEM". If ppb_resize_pipe fails, we create another pipe.

rst0git:
cc: @cyrillos 
Closing in favour of https://lists.openvz.org/pipermail/criu/2019-March/043572.html

Issue #612, opened 2019-02-12
fc/rawhide: error: taking address of packed member of 'struct user_i387_ia32_struct' may result in an unaligned pointer value


avagin:
```
  CC       compel/src/lib/log.o
  CC       compel/arch/x86/src/lib/cpu.o
  CC       compel/arch/x86/src/lib/infect.o
[91mcompel/arch/x86/src/lib/infect.c: In function 'compel_convert_from_fxsr':
compel/arch/x86/src/lib/infect.c:112:37: error: taking address of packed member of 'struct user_i387_ia32_struct' may result in an unaligned pointer value [-Werror=address-of-packed-member]
  112 |  struct fpreg *to = (struct fpreg *)&env->st_space[0];
      |                                     ^~~~~~~~~~~~~~~~~
```

https://travis-ci.org/checkpoint-restore/criu/jobs/491756052

avagin:
Cc: @adrianreber

adrianreber:
Will look into it. There is currently another error on criu-dev travis with rawhide, but still an error:
```
In file included from compel/include/uapi/compel/asm/sigframe.h:7,
                 from compel/plugins/std/infect.c:14:
compel/include/uapi/compel/asm/fpu.h:15: error: "FP_XSTATE_MAGIC2_SIZE" redefined [-Werror]
   15 | #define FP_XSTATE_MAGIC2_SIZE  sizeof(FP_XSTATE_MAGIC2)
      | 
In file included from /usr/include/signal.h:291,
                 from compel/include/uapi/compel/plugins/std/syscall-types.h:12,
                 from compel/include/uapi/compel/plugins/std/syscall-64.h:5,
                 from compel/include/uapi/compel/plugins/std/syscall.h:8,
                 from compel/include/uapi/compel/plugins/std.h:5,
                 from compel/plugins/std/infect.c:1:
/usr/include/bits/sigcontext.h:29: note: this is the location of the previous definition
   29 | #define FP_XSTATE_MAGIC2_SIZE sizeof (FP_XSTATE_MAGIC2)
      | 
```

PR #649, opened 2019-03-13
criu: fix gcc9 rawhide compilation errors

adrianreber:
This fixes errors on current Fedora rawhide with gcc 9.

Most of the changes is to remove `__packed` to remove errors like:
```
 error: taking address of packed member of ‘union <anonymous>’ may result in an unaligned pointer value [-Werror=address-of-packed-member]
```
CC: @cyrillos @0x7f454c46 

There have similar changes last year for gcc 8 but that was more adding alignments. Not sure if removing `__packed` is correct.

This also fixes:
```
  CC       criu/cr-restore.o
In file included from criu/include/restore.h:6,
                 from criu/cr-restore.c:91:
criu/cr-restore.c: In function ‘sigreturn_restore’:
criu/arch/x86/include/asm/restore.h:10:2: error: listing the stack pointer register ‘rsp’ in a clobber list is deprecated [-Werror=deprecated]
   10 |  asm volatile(       \
      |  ^~~
```
and:
```
  CC       criu/sk-unix.o
In file included from criu/include/imgset.h:5,
                 from criu/sk-unix.c:17:
criu/sk-unix.c: In function ‘dump_one_unix_fd’:
criu/include/log.h:51:2: error: ‘%s’ directive argument is null [-Werror=format-overflow=]
   51 |  print_on_level(LOG_ERROR,     \
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   52 |          "Error (%s:%d): " LOG_PREFIX fmt,  \
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   53 |          __FILE__, __LINE__, ##__VA_ARGS__)
      |          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
criu/sk-unix.c:440:5: note: in expansion of macro ‘pr_err’
  440 |     pr_err("Unix socket %d with unreachable peer %d (%d/%s)\n",
      |     ^~~~~~
```

### DIFF
--- a/compel/arch/x86/src/lib/include/uapi/asm/fpu.h
+++ b/compel/arch/x86/src/lib/include/uapi/asm/fpu.h
@@ -12,7 +12,9 @@
 
 #define FP_XSTATE_MAGIC1		0x46505853U
 #define FP_XSTATE_MAGIC2		0x46505845U
+#ifndef FP_XSTATE_MAGIC2_SIZE
 #define FP_XSTATE_MAGIC2_SIZE		sizeof(FP_XSTATE_MAGIC2)
+#endif
 
 #define XSTATE_FP			0x1
 #define XSTATE_SSE			0x2
@@ -261,7 +263,7 @@ struct xsave_struct_ia32 {
 		struct ymmh_struct	ymmh;
 		uint8_t			extended_state_area[EXTENDED_STATE_AREA_SIZE];
 	};
-} __aligned(FXSAVE_ALIGN_BYTES) __packed;
+} __aligned(FXSAVE_ALIGN_BYTES);
 
 typedef struct {
 	/*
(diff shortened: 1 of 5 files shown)


cyrillos:
Ack

0x7f454c46:
@adrianreber I'm up for removing __packed attributes — I guess, I've profusely scattered them all around just to be sure that the result is what I expect from a compiler, quite pessimistically.
IIRC, with much @avagin work we're running ia32 tests on linux-next (as TravisCI doesn't have 4.10 kernel) - so we should notice any regression before it may burst.

FWIW,
Reviewed-by: Dmitry Safonov <0x7f454c@gmail.com>

Thanks!

adrianreber:
Thanks everyone. For the record: I see that the rawhide travis test is now green.

Fixes: #612

PR #648, opened 2019-03-12
Fix SELinux process labeling and label sockets correctly

adrianreber:
This changes is motivated by https://github.com/containers/libpod/issues/2334

The main problem when trying to checkpoint a container on a SELinux enabled system is, that the connection from the parasite daemon to the main CRIU process is blocked by SELinux. The error looks something like:
```
type=AVC msg=audit(1550142323.799:1167): avc:  denied  { connectto } for  pid=23569 comm="top" path=002F6372746F6F6C732D70722D3233363139 scontext=system_u:system_r:container_t:s0:c245,c463 tcontext=unconfined_u:system_r:container_runtime_t:s0-s0:c0.c1023 tclass=unix_stream_socket permissive=0
```
That SELinux blocks a connect() from the container process to the outside of the container is understandable.

The solution discussed and which has been implemented in this PR is, that CRIU now labels the socket on a SELinux enabled system with the same SELinux context as the parasite daemon running in the container. With the socket correctly labeled Podman can now checkpoint a container on a SELinux enabled system without denials, because the parasite daemon can access the main CRIU process.

This PR also removes CRIU limitation to only work with the SELinux context 'unconfined_t'. According to @tych0 this limitation was introduced as, at that time, no one was sure if the SELinux implementation actually works. With the correct SELinux policies it works correctly. The biggest problem is, that usually it is not expected that a process changes its own process context. CRIU, however, has to do it. With this change a restored container has the same process context as during checkpointing.

For the container use case the Fedora package container-selinux has introduced a policy to allow CRIU to transition from one process context to another (https://github.com/containers/container-selinux/commit/891a85fee0f55ea32100c3f47a273ea522f4f2fc https://github.com/containers/container-selinux/commit/a2fc0309642a6430525cdaebcb3f1c8efde45fe2).

Processes started from the shell are usually running as 'unconfined_t' if no special policy exists and as CRIU will then also be running as 'unconfined_t' it should just work as before. For any process running in another process context additional SELinux policies have to be defined in the future.

PR #657, opened 2019-03-26
restore: support restoring threads with SELinux

adrianreber:
Restoring a multi-threaded process with CRIU's SELinux support fails
because SELinux does not always support changing the process context of
a multi-threaded process.

Reading the man-page for setcon(), to change the context of a running
process, it states that changing the SELinux context of a multi-threaded
process can only work 'if the new security context is bounded by the old
security context'.

To be able to restore a process without the need to have 'the new
security context [] bounded by the old security context', this sets the
SELinux process context before creating the threads. Thus all threads
are created with the process context of the main process.

### DIFF
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -43,6 +43,7 @@
 
 #include "images/creds.pb-c.h"
 #include "images/mm.pb-c.h"
+#include "images/inventory.pb-c.h"
 
 #include "shmem.h"
 #include "restorer.h"
@@ -179,7 +180,8 @@ static int lsm_set_label(char *label, int procfd)
 	return 0;
 }
 
-static int restore_creds(struct thread_creds_args *args, int procfd)
+static int restore_creds(struct thread_creds_args *args, int procfd,
+			 int lsm_type)
 {
 	CredsEntry *ce = &args->creds;
 	int b, i, ret;
@@ -296,8 +298,16 @@ static int restore_creds(struct thread_creds_args *args, int procfd)
 		return -1;
 	}
 
-	if (lsm_set_label(args->lsm_profile, procfd) < 0)
-		return -1;
+	if (lsm_type != LSMTYPE__SELINUX) {
+		/*
+		 * SELinux does not support setting the process context for
+		 * threaded processes. So this is skipped if running with
+		 * SELinux and instead the process context is set before the
+		 * threads are created.
+		 */
+		if (lsm_set_label(args->lsm_profile, procfd) < 0)
+			return -1;
+	}
 	return 0;
 }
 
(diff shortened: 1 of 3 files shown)


adrianreber:
This is a follow-up to #648 

The main problem with #648 is that SELinux only partially supports changing the process context of a threaded process. According to the manpage:

>  A multi-threaded application can perform a setcon() prior to creating any child threads, in which case all of the child threads will inherit the new context.  However, prior to Linux 2.6.28, setcon() would fail if there are any other threads running in the same process since this would yield an inconsistency among the security contexts of threads sharing the same memory space.  Since Linux 2.6.28, setcon() is permitted for threads  within  a multi-threaded process if the new security context is bounded by the old security context, where the bounded relation is defined through typebounds statements in the policy and guarantees that the new security context has a subset of the permissions of the old security context.

For my current use case (getting Podman checkpoint/restore with SELinux working) this pull request works with the following two policies:
```
allow container_t container_var_lib_t:file append;
allow container_t sysctl_kernel_t:file write;
```
The first line is to enable the restorer to write to the log-file and should not be a problem. The second line could be a bit more problematic, but is is necessary so that each thread can write to `ns_last_pid`. As far as I know `ns_last_pid` requires CAP_SYS_ADMIN and seccomp will be restored later, so that once the container is restored it is not possible to write to `ns_last_pid` anyway.

The approach in this PR only works if all threads are using the same process context, but if it is not possible to change the context of a single thread during restore, it is also not possible to have a container with threads running with different process context. So the restore is limited, but only in the same way the original process startup is limited.

In parallel I am having the discussion if the approach in this PR with the necessary policy changes can be implemented in the corresponding selinux policy.

CC: @tych0  (just FYI, it's complicated :wink:)

criu/pie/restorer.c
```diff
@@ -296,8 +297,17 @@ static int restore_creds(struct thread_creds_args *args, int procfd)
 		return -1;
 	}
 
-	if (lsm_set_label(args->lsm_profile, procfd) < 0)
-		return -1;
+	if (lsm_type != 1) {
```
avagin: Can we use LSMTYPE__SELINUX here? If we can't, pls write a few lines in the comment about this.

adrianreber: I thought we cannot include the protobuf headers in the restorer, but did not actually verify it. I just tried it and I can use LSMTYPE__SELINUX without a problem. Will push an update.

adrianreber:
The feedback I got so far from SELinux experts is that setting the process context of threads is only possible if the security context is bounded and this cannot change. On Fedora bounded security context are rarely used at all.

So this approach to set the process context before creating the threads seems to be the right thing to do and the additional policies, especially the one to write to ns_last_pid, are acceptable.

adrianreber:
The necessary policy to label ns_last_pid correctly has been merged into Fedora's SELinux policy package: https://github.com/fedora-selinux/selinux-policy/commit/f1ba302a9f2b7b249b5ac8df2705be7869575bb6

There should also soon be a new container-selinux package which actually allows CRIU to write to ns_last_pid from threads.

@avagin: From my side everything is ready for 3.12.

PR #660, opened 2019-03-29
mount: check mounts after restoring

Snorch:
Patches 1-4 do a minor cleanup of mounts code.

Patch 5 brings in the option --check-mounts on restore to check that at
the end of restore we have properly restored tree of mounts in each
mount namespace of the container.

We've seen cases of container migration passing fine without error, but
actually some mounts changed their attributes or even some new mounts
appeared. Potentially these "inaccuracies" during restore can accumulate
over several migrations of the same container, and we will finally see
unexpected crap in mounts and won't be able to find out the initial
problem. With --check-mounts we will catch initial problem and fix it.

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -330,7 +330,7 @@ static bool mounts_equal(struct mount_info *a, struct mount_info *b)
  */
 static char *mnt_roots;
 
-static struct mount_info *mnt_build_ids_tree(struct mount_info *list, struct mount_info *yard_mount)
+static struct mount_info *mnt_build_ids_tree(struct mount_info *list)
 {
 	struct mount_info *m, *root = NULL;
 
@@ -351,41 +351,14 @@ static struct mount_info *mnt_build_ids_tree(struct mount_info *list, struct mou
 
 		if (!parent) {
 			/* Only a root mount can be without parent */
-			if (root == NULL && m->is_ns_root) {
+			if (!root && m->is_ns_root) {
 				root = m;
-				if (!yard_mount)
-					continue;
+				continue;
 			}
 
-			if (!root) {
-				pr_err("No parent found for mountpoint %d (@%s)\n",
-					m->mnt_id, m->mountpoint);
-				return NULL;
-			}
-
-			pr_debug("Mountpoint %d (@%s) w/o parent %d\n",
-				 m->mnt_id, m->mountpoint, m->parent_mnt_id);
-
-			if (!mounts_sb_equal(root, m) ||
-			    strcmp(root->root, m->root)) {
-				pr_err("Nested mount namespaces with different "
-				       "roots %d (@%s %s) %d (@%s %s) are not supported yet\n",
-				       root->mnt_id, root->mountpoint, root->root,
-				       m->mnt_id, m->mountpoint, m->root);
-				return NULL;
-			}
-
-			/* Mount all namespace roots into the roots yard. */
-			parent = yard_mount;
-			if (unlikely(!yard_mount)) {
-				pr_err("Nested mount %d (@%s %s) w/o root insertion detected\n",
-				       m->mnt_id, m->mountpoint, m->root);
-				return NULL;
-			}
-
-			pr_debug("Mountpoint %d (@%s) get parent %d (@%s)\n",
-				 m->mnt_id, m->mountpoint,
-				 parent->mnt_id, parent->mountpoint);
+			pr_err("No parent found for mountpoint %d (@%s)\n",
+				m->mnt_id, m->mountpoint);
+			return NULL;
 		}
 
 		m->parent = parent;
@@ -396,9 +369,6 @@ static struct mount_info *mnt_build_ids_tree(struct mount_info *list, struct mou
 		pr_err("No root found for tree\n");
 		return NULL;
 	}
-
-	if (yard_mount)
-		return yard_mount;
 
 	return root;
 }
@@ -997,8 +967,7 @@ static int resolve_shared_mounts(struct mount_info *info, int root_master_id)
 	return 0;
 }
 
-static struct mount_info *mnt_build_tree(struct mount_info *list,
-					 struct mount_info *root_mp)
+static struct mount_info *mnt_build_tree(struct mount_info *list)
 {
 	struct mount_info *tree;
 
@@ -1007,7 +976,7 @@ static struct mount_info *mnt_build_tree(struct mount_info *list,
 	 */
 
 	pr_info("Building mountpoints tree\n");
-	tree = mnt_build_ids_tree(list, root_mp);
+	tree = mnt_build_ids_tree(list);
 	if (!tree)
 		return NULL;
 
@@ -1690,7 +1659,7 @@ struct mount_info *collect_mntinfo(struct ns_id *ns, bool for_dump)
 		return NULL;
 	}
 
-	ns->mnt.mntinfo_tree = mnt_build_tree(pm, NULL);
+	ns->mnt.mntinfo_tree = mnt_build_tree(pm);
 	if (ns->mnt.mntinfo_tree == NULL)
 		goto err;
 
@@ -2881,7 +2850,7 @@ static int get_mp_mountpoint(char *mountpoint, struct mount_info *mi, char *root
 	return 0;
 }
 
-static int collect_mnt_from_image(struct mount_info **pms, struct ns_id *nsid)
+static int collect_mnt_from_image(struct mount_info **head, struct mount_info **tail, struct ns_id *nsid)
 {
 	MntEntry *me = NULL;
 	int ret, root_len = 1;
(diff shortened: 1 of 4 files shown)


cyrillos:
@Snorch didn't you fixed mount order nit yesterday? Is it applicable to this series? Or it was another bug?

Snorch:
@cyrillos Yes I've already updated these.

mount: rework mount tree build step on restore
...
v2: prepend children to the root_yard in merge_mount_trees so that the
order in merged tree persists

criu/crtools.c
```diff
@@ -422,6 +422,9 @@ int main(int argc, char *argv[], char *envp[])
 "			Namespace can be specified as either pid or file path.\n"
 "			OPTIONS can be used to specify parameters for userns:\n"
 "			    user:PID,UID,GID\n"
+"* Sanity checks:\n"
+"  --check-mounts        Check that mount trees restored fine in each mntns,\n"
+"                        check they are identical to what we have in images.\n"
```
rst0git: Perhaps this should be added to the man page (`Documentation/criu.txt`) too?

Documentation/criu.txt
```diff
@@ -514,6 +514,12 @@ are not adequate, but this can be suppressed by using *--cpu-cap=none*.
    restored process.
    This option requires running *lazy-pages* daemon.
 
+*--check-mounts*::
+   Additional sanity check for mount tree restore. These checks that restored
+   mount tree is identical to what we have in images. These can be used to
+   debug problems with mounts and as an additional protection from corrupting
+   the mount tree.
```
Snorch: Added comment about the option to documentation.

criu/crtools.c
Snorch: done.

criu/cr-restore.c
```diff
@@ -3227,6 +3227,10 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 		if (root_ns_mask & CLONE_NEWNS &&
 		    remount_readonly_mounts())
 			goto err_nv;
+		if (opts.check_mounts &&
+		    root_ns_mask & CLONE_NEWNS &&
```
avagin: (root_ns_mask & CLONE_NEWNS)

criu/mount.c
```diff
@@ -3821,3 +3821,133 @@ int remount_readonly_mounts(void)
 	 */
 	return call_helper_process(ns_remount_readonly_mounts, NULL);
 }
+
+enum tree_step {
```
avagin: need a comment which explains these constants

criu/mount.c
```diff
@@ -3821,3 +3821,133 @@ int remount_readonly_mounts(void)
...
+	*step = STEP_STOP;
+	return NULL;
+}
+
+static void mnt_resort_siblings_full(struct mount_info *tree)
+{
+	struct mount_info *m, *p;
+	LIST_HEAD(list);
+
+	pr_info("\tResorting siblings full on %d\n", tree->mnt_id);
```
avagin: what does this mean?

criu/mount.c
```diff
@@ -3821,3 +3821,133 @@ int remount_readonly_mounts(void)
...
+	while (!list_empty(&tree->children)) {
+		m = list_first_entry(&tree->children, struct mount_info, siblings);
+		list_del(&m->siblings);
+
+		list_for_each_entry(p, &list, siblings)
+			if (strcmp(p->ns_mountpoint, m->ns_mountpoint) < 0)
+				break;
+
+		list_add_tail(&m->siblings, &p->siblings);
+		mnt_resort_siblings_full(m);
```
avagin: recursion? what is a max depth of it? Do we really need it? Could you write a comment about what is going on here?

criu/cr-restore.c
Snorch: Done.

criu/mount.c
Snorch: Done.

criu/mount.c
Snorch: I copied these function from mnt_resort_siblings with slight change. The prototype function has the same message to tell the root of the tree we are proceeding now. May be I need to rewrite the message to `pr_info("\tResorting siblings on %d (full)\n", tree->mnt_id);` for clearness. Is it ok?

criu/mount.c
Snorch: Actually max depth is max depth of mount tree for a mntns. For the tree from images the depth is ok because we've already done mnt_resort_siblings on it with the same depth (Though I don't see any prove that recursion in mnt_resort_siblings will not fail). But for the restored tree (which is expected in case of some error in mount restore be a different tree) we can't say the depth is ok. So I think either 1) we need a depth check (likely non-recursive dfs) before calling mnt_resort_siblings* or 2) we need to do mnt_resort_siblings non-recursive way which will be harder because we at the same time need to alter the tree in these function (may be we can do it with our regular non-recursive dfs next, need to think more about it).

I prefer (1) but I'm not sure about safe limit on the depth of the recursion, not sure if just testing it will be enough. What do you think?

criu/mount.c
Snorch: Fixed print.

criu/mount.c
Snorch: Done (2) - removed recursion from both current patch and mnt_resort_siblings.

Snorch: Done (2) - removed recursion from both current patch and mnt_resort_siblings. But yet I have some problem with tests -> will try fixing it.

Snorch:
seccomp_filter_* problems looks fixed by https://github.com/checkpoint-restore/criu/issues/725
```
Test zdtm/static/seccomp_filter_threads FAIL at zdtm/static/seccomp_filter_threads die 
Test zdtm/static/seccomp_filter_tsync FAIL at zdtm/static/seccomp_filter_tsync die
```
So rebase to latest criu and rerun tests.

Snorch:
Add a fix for binfmt_misc vs --check-mounts interaction: "mount/check-mounts: remove auxiliary cr-time mount from tree after umount". These is a first problem found by --check-mounts. It is more a cosmetic fix, but I'm very happy to know about every such unobvious  thing about mounts. I think that adding auxiliary mounts to the mount tree is not a good practice.

Snorch:
One more improvement: don't stop on first error when checking mounts, we need to see errors in all mntnses to find the first error.

Snorch:
> I think that adding auxiliary mounts to the mount tree is not a good practice.

I have an example when it can be bad:

```
# Preparation
[root@silo tmp]# mount -t tmpfs tmpfs test-mounts/
# Should be shared
[root@silo tmp]# cat /proc/self/mountinfo | grep test-mounts
838 72 0:223 / /tmp/test-mounts rw,relatime shared:990 - tmpfs tmpfs rw
[root@silo tmp]# mkdir test-mounts/dir1/dir2 -p
[root@silo tmp]# mount --bind test-mounts/ test-mounts/dir1/dir2/
[root@silo tmp]# cat /proc/self/mountinfo | grep test-mounts
838 72 0:223 / /tmp/test-mounts rw,relatime shared:990 - tmpfs tmpfs rw
1001 838 0:223 / /tmp/test-mounts/dir1/dir2 rw,relatime shared:990 - tmpfs tmpfs rw
```
```
# Mount the mount which can't be unmounted =)
[root@silo tmp]# mount -t tmpfs test-tmpfs test-mounts/dir1/dir2/dir1/
[root@silo tmp]# umount test-mounts/dir1/dir2/dir1/
umount: test-mounts/dir1/dir2/dir1/: mountpoint not found
[root@silo tmp]# cat /proc/self/mountinfo | grep test-mounts
838 72 0:223 / /tmp/test-mounts rw,relatime shared:990 - tmpfs tmpfs rw
1001 838 0:223 / /tmp/test-mounts/dir1/dir2 rw,relatime shared:990 - tmpfs tmpfs rw
1007 1001 0:224 / /tmp/test-mounts/dir1/dir2/dir1 rw,relatime shared:1002 - tmpfs test-tmpfs rw
1008 838 0:224 / /tmp/test-mounts/dir1 rw,relatime shared:1002 - tmpfs test-tmpfs rw
```

Snorch:
It looks like nobody wants --check-mounts feature (be able to check that our mounts restore with many heuristics really done it's job right), at least implemented in criu, not in "post-dump"(or may be even some new "post-freeze")"/"pre-resume" scripts, may be I need to rework it these way... But in scripts I think it will be more performance consuming, than in criu.

So I offer a compromise =)

Maybe we can at least take the preparational  patches from these series? They look quiet self consistent and good in general.

76e361672 mount: remove useless check in populate_mnt_ns

We have a useless check, why not remove it?

e52caca43 ns: make rst_new_ns_id static

We have exported function which is never used outside, why not make it static?

153c3e606 ns/restore/image: do not read namespace images for non-namespaced case

We try to open images of mnt/net namespaces when we were explicitly told not to do it, why not fix it? 

f5c4bb319 mount: rework mount tree build step on restore

These one is not a simple one. Cons: it changes the code quiet a lot. Pros: it separates extra complexity of merging trees from mnt_build_ids_tree to separate function merge_mount_trees; it makes nsid->mnt.mntinfo_tree mean something, now they (AFAICR) all equal to the root_yard and can be removed =); and it simplifies populate_mnt_ns. 

51d957474 mount: make mnt_resort_siblings nonrecursive and reuse friendly

Obviously, non-recursive is much better than recursive, we surely need to fix it, else once we will hit the stack limit.

Other "Pro" for these patches is that we use them in VZ7 criu for almost a year and have no problem with them, but it would create additional pain to port our patches to mount engine back to criums.

Snorch:
Ripped out STEP and left only 5 preparational patches.

avagin:
Applied, thanks!

Snorch:
Thanks a lot!

PR #663, opened 2019-03-31
Add TLS support

rst0git:


avagin:
Where is a test?

rst0git:
The changes in this PR currently follow the convention that *"new features are disabled by default"*, and therefore a `--tls-enable` option is introduced. However, in terms of security, the practice that applications should be *"secure by default and have the best security it can without having the user to turn it on"* seems to be more important.

An alternative solution could be to set `--tls-enable` by default, and/or use similar behaviour to [skopeo](https://github.com/containers/skopeo)'s `--tls-verify` option, which is described as:

> Require HTTPS and verify certificates when contacting registries (default: true). If explicitly set to true, then TLS verification will be used. If set to false, then TLS verification will not be used. If not specified, TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf

This might require to auto-generate X509 certificates for CRIU in a manner similar to [LXD](https://lxd.readthedocs.io/en/latest/security/), and ask the user to add
@rppt @avagin what do you think?

avagin:
> @rppt @avagin what do you think?

I would suggest to disable it by default initially. When we will be sure that this works, we will switch to skopeo's behavior.

And I would like to be able to choose a default behavior in a criu config file.

avagin:
And we need a wiki page which will describe how all these work, how to use this and how to generate certificates.

criu/page-xfer.c
```diff
@@ -561,9 +583,16 @@ static int page_server_check_parent(int sk, struct page_server_iov *pi)
 	if (ret < 0)
 		return -1;
 
-	if (write(sk, &ret, sizeof(ret)) != sizeof(ret)) {
-		pr_perror("Unable to send response");
-		return -1;
+	if (opts.tls_psk_key) {
```
0x7f454c46: Can we introduce helpers for all those checks if (opts.tls_psk_key)?
I struggle to find a proper name, so it's up to you, but something like:
```
int __send(int sockfd, const void *buf, size_t len)
{
        size_t transferred = opts.tls_psk_key ? tls_send(buf, len) : write(sk, buf, len);

        return (transferred - len);
}
```
Same for recv

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+ssize_t tls_send_data_from_fd(int fd, unsigned long len)
+{
+	ssize_t copied;
+	ssize_t total_sent = 0;
+	char *buffer = xmalloc(len);
+
+	if (!buffer)
+		return -1;
+
+	while (len > total_sent) {
```
0x7f454c46: drop `total_sent`, please - just decrease `len` whenever sent successful
`while (len) {`

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+		copied = read(fd, (buffer + total_sent), (len - total_sent));
```
0x7f454c46: you can read() here into the beginning of the buffer here - overriding the content already sent

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+/*
+ * Read data from a file descriptor, then encrypt and send it with GnuTLS.
+ * This function is used for cases when we would otherwise use splice()
+ * to transfer data from PIPE to TCP socket.
+ */
+ssize_t tls_send_data_from_fd(int fd, unsigned long len)
+{
+	ssize_t copied;
+	ssize_t total_sent = 0;
+	char *buffer = xmalloc(len);
```
0x7f454c46: I wonder if we need to limit the size to be xmalloc()ed here.
Maybe something like (more question rather than suggestion):
```
#define SPLICE_BUF_SZ_MAX (PIPE_BUF * 100)  /* some reasonable limit */
[..]
char *buffer = alloca(min(len, SPLICE_BUF_SZ_MAX));
```

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+/*
+ * Read and decrypt data with GnuTLS and write it to a file descriptor.
+ * This function is used for cases when we would otherwise use splice()
+ * to transfer data from a TCP socket to a PIPE.
+ */
+ssize_t tls_recv_data_to_fd(int fd, unsigned long len)
+{
+	ssize_t received;
+	ssize_t total_written = 0;
+	char *buffer = xmalloc(len);
```
0x7f454c46: ditto: get rid of `total_written` and probably (?) add a limit

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+{
+	int ret;
+
+	ret = tls_set_priority(&session);
+	if (ret != GNUTLS_E_SUCCESS) {
+		tls_perror("Failed to set priority", ret);
+		return -1;
+	}
+
+	connfdPtr = xmalloc(sizeof(int));
```
0x7f454c46: xmalloc() can fail and leak on the error-path.
But, do you need to allocate `connfdPtr`?
Won't `gnutls_transport_set_ptr(session, (gnutls_transport_ptr_t)&tls_sk)` work?

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+	ret = gnutls_init(&session, GNUTLS_CLIENT);
+	if (ret != GNUTLS_E_SUCCESS) {
+		tls_perror("Failed to initialize session", ret);
+		return -1;
+	}
+
+	ret = gnutls_psk_allocate_client_credentials(&c_psk_cred);
+	if (ret != 0) {
+		tls_perror("Failed to allocate credentials structure", ret);
+		return -1;
```
0x7f454c46: gnutls_deinit()?

criu/page-xfer.c
rst0git: >Can we introduce helpers for all those checks if (opts.tls_psk_key)?

That is a good idea, thanks! Another thing is that `tls_send()` does not currently pass the `flags` argument for `send()` (e.g. `MSG_DONTWAIT`). I will try to add this too.

criu/tls.c
```diff
@@ -0,0 +1,336 @@
...
+	}
+
+	key.size = strlen(opts.tls_psk_key);
+	key.data = xmalloc(key.size);
+	memcpy(key.data, opts.tls_psk_key, key.size);
```
0x7f454c46: Do we need malloc() & memcpy() here, or can just pass the key?
BTW, malloc() can fail.

criu/uffd.c
```diff
@@ -1469,5 +1470,10 @@ int cr_lazy_pages(bool daemon)
 
 	ret = handle_requests(epollfd, events, nr_fds);
 
+	if (opts.tls_psk_key) {
```
0x7f454c46: ditto: detail of realization can be hidden inside to simplify uffd code :)

rst0git:
**Update:** There is some performance overhead of using TLS and it could be reduced with the [in-kernel implementation](https://www.kernel.org/doc/Documentation/networking/tls.txt).

criu/tls.c
```diff
@@ -0,0 +1,343 @@
...
+{
+	ssize_t copied;
+	unsigned long buf_size = min(len, (unsigned long)SPLICE_BUF_SZ_MAX);
+	void *buf = xmalloc(buf_size);
+
+	if (!buf)
+		return -1;
+
+	while (len > 0) {
+		copied = read(fd, buf, buf_size);
```
avagin: min(len, buf_size)?

criu/tls.c
```diff
@@ -0,0 +1,343 @@
...
+		if (copied < 0) {
+			pr_perror("Can't read from pipe");
+			goto err;
+		}
+
+		if (copied == 0)
+			break;
+
+		int ret, sent;
```
avagin: all variables have to be declared at the beginning of the block

criu/tls.c
```diff
@@ -0,0 +1,343 @@
...
+
+	while (len > 0) {
+		copied = read(fd, buf, buf_size);
+		if (copied < 0) {
+			pr_perror("Can't read from pipe");
+			goto err;
+		}
+
+		if (copied == 0)
+			break;
```
avagin: need to report an error in this case

criu/page-xfer.c
```diff
@@ -206,7 +227,7 @@ static int open_page_server_xfer(struct page_xfer *xfer, int fd_type, unsigned l
 	/* Push the command NOW */
 	tcp_nodelay(xfer->sk, true);
 
-	if (read(xfer->sk, &has_parent, 1) != 1) {
+	if (__recv(xfer->sk, &has_parent, 1, 0) != 1) {
 		pr_perror("The page server doesn't answer");
```
avagin: will tls_recv set a proper errno wich is reported by pr_perror?

criu/tls.c
```diff
@@ -0,0 +1,343 @@
...
+	}
+	while (ret == GNUTLS_E_INTERRUPTED || ret == GNUTLS_E_AGAIN);
+	tls_sk_flags = 0;
+
+	if (ret == GNUTLS_E_REHANDSHAKE) {
+		tls_perror("Re-handshake is not supported", ret);
+		return -1;
+	}
+
+	if (gnutls_error_is_fatal(ret)) {
```
avagin: do we need to report non-fatal errors?

criu/page-xfer.c
rst0git: the value for the `errno` variable will be set by `recv()` which is called by `gnutls_record_recv()` (in the `tls_recv()` function).

For example, by terminating the `criu dump` process in the middle of a post-copy migration, on the destination side, I'm getting the following error message:
```
Error (criu/tls.c:128): tls: Failed receiving data: Error in the pull function.
Error (criu/page-xfer.c:1236): Error reading async data from page server: Connection reset by peer 
```

However, `gnutls_record_recv()` may also fail with a different error, such as `GNUTLS_E_INVALID_SESSION`, `GNUTLS_E_INTERNAL_ERROR`, etc. In this case the value of `errno` will not be set.

criu/tls.c
rst0git: [non-fatal errors](https://github.com/gnutls/gnutls/blob/0a47c08db07d0a01018aec34d10a294aed79fdd1/lib/errors.c#L442L461) are:
```C
static const gnutls_error_entry non_fatal_error_entries[] = {
	ERROR_ENTRY(N_("Success."), GNUTLS_E_SUCCESS),
	ERROR_ENTRY(N_("A TLS warning alert has been received."),
		    GNUTLS_E_WARNING_ALERT_RECEIVED),
	ERROR_ENTRY(N_("A heartbeat pong message was received."),
		    GNUTLS_E_HEARTBEAT_PONG_RECEIVED),
	ERROR_ENTRY(N_("A heartbeat ping message was received."),
		    GNUTLS_E_HEARTBEAT_PING_RECEIVED),
	ERROR_ENTRY(N_("Resource temporarily unavailable, try again."),
		    GNUTLS_E_AGAIN),
	ERROR_ENTRY(N_("The transmitted packet is too large (EMSGSIZE)."),
		    GNUTLS_E_LARGE_PACKET),
	ERROR_ENTRY(N_("Function was interrupted."), GNUTLS_E_INTERRUPTED),
	ERROR_ENTRY(N_("Rehandshake was requested by the peer."),
		    GNUTLS_E_REHANDSHAKE),
	ERROR_ENTRY(N_("Re-authentication was requested by the peer."),
		    GNUTLS_E_REAUTH_REQUEST),
	/* Only non fatal (for handshake) errors here */
	{NULL, NULL, 0}
};
```
Currently re-authentication, re-handshake and alerts are not used/expected, so I don't think that logging non-fatal errors would be very useful.

criu/page-xfer.c
avagin: I mean that pr_perror always prints errno, but __recv can call tls_recv which doesn't set errno and the error message will contain errno which has nothing with a described error...

rst0git: I changed tls_send() and tls_recv() to set `errno = EIO` in case of error.

criu/tls.c
rst0git: Since non-fatal errors are not expected I changed this condition for tls_send() and tls_recv() to return -1 in case of any error.

rst0git:
I updated the PR by modifying `tls_send()` and `tls_recv()` to propagate the `errno` value of the underlying push/pull callback function. This is needed to properly handle [asynchronous operations](https://www.gnutls.org/manual/html_node/Asynchronous-operation.html) (i.e. `page_server_async_read()`).

PR #664, opened 2019-04-04
zdtm: Replace imp module with importlib

rst0git:
[imp.load_source()](https://docs.python.org/2.7/library/imp.html#imp.load_module) has been deprecated. The recommended alternative API for loading a module is [exec_module()](https://docs.python.org/3.4/library/importlib.html#importlib.abc.Loader.exec_module).

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -18,7 +18,6 @@ import sys
 import linecache
 import random
 import string
-import imp
 import fcntl
 import errno
 import datetime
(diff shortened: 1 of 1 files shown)


adrianreber:
The linter is not happy with your python code changes. A few empty lines are missing I think. But thanks for working on this and trying to removing this warning.

rst0git:
> The linter is not happy with your python code changes. A few empty lines are missing I think.

Thank you for pointing this out. I updated the commit.

PR #665, opened 2019-04-05
crtools: Print error messages from check_options()

rst0git:
When `check_options()` exits with an error (return value != 0) the logging is not yet initialised, and therefore the error messages are not printed out.

Since this affects only command-line usage, and only when `check_options()` reports an error, flush the early log messages to standard error.


adrianreber:
Makes sense :+1:

Issue #341, opened 2017-06-29
Make it possible to show relative timestamps in criu logs


xemul:
There are two places to fix.

1. `print_ts()` from `criu/log.c`
2. `sbuf_log_init()` from `compel/plugins/std/log.c`

Also callers of `std_log_set_fd()` and `parasite_init_daemon()` might need to correctly pass the starting timeval into parasite and restorer contexts.

mitulkarnik:
I am interested to implement this enhancement. As per the suggestions given by Radostin on mailing list I am looking for possible solutions. 
Candidates:
clock_gettime() with CLOCK_MONOTONIC  parameter 
exploring more on similar lines

mitulkarnik:
I am exploring the possibility of using clock_gettime with Clock_Id as CLOCK_MONOTONIC_RAW which eliminates the skews caused by ntp or adjtime. I want to know if this approach works on all platforms 
something like 

`clock_gettime(CLOCK_MONOTONIC_RAW, &timespec);`
I have created the patch for the same and am able to generate the relative timestamps as well. 
Have doubt if this approach works.  Will send out the patch soon

xemul:
Actually, the idea behind this issue was quite simple -- when printing a new line in logs get the time of the previous line, subtract it from the current time and print the result. We all wanted to see in logs the amt of time each line takes to roughly estimate the "weight" of each operation.

nishant-08:
Hello.
Is this enhancement still desirable? I found this interesting and would like to work on this. My idea is pretty similar to what xemul mentioned, i.e. subtracting previous time from current time.

xemul:
Yes. Patches are welcome

PR #668, opened 2019-04-08
Relative timestamps

sagartewari01:
Adds relative timestamps as requested in issue #341. The decision to print relative timestamps is handed over to flags as of now. For `criu` command line tool, the flag could be tied to an argument. How should `compel` set that flag?
I'll make the changes once these questions are resolved.


sagartewari01:
I made only one commit over the latest master. Don't know why three commits are showing up.

compel/plugins/std/log.c
```diff
@@ -51,11 +54,19 @@ static void sbuf_log_init(struct simple_buf *b)
 	 */
 	b->bp = b->buf;
 
-	if (start.tv_sec != 0) {
+	if (start.tv_sec != 0 || print_ts_diffs) {
 		struct timeval now;
 
 		sys_gettimeofday(&now, NULL);
-		timediff(&start, &now);
+        struct timeval *pivot = print_ts_diffs ? &last_t : &start;
```
avagin: You have to follow the coding style:
https://www.kernel.org/doc/html/v4.10/process/coding-style.html

avagin:
> I made only one commit over the latest master. Don't know why three commits are showing up.
git fetch
git rebase origin/master
And then you need to push chnages to github again.

avagin:
> added relative timestamps, flag TODO

What does this "flag TODO" mean?

You need to write more details in the commit message.

sagartewari01:
@avagin I've updated the pull request.

compel/plugins/std/log.c
```diff
@@ -13,6 +13,8 @@ struct simple_buf {
 	void (*flush)(struct simple_buf *b);
 };
 
+int print_ts_diffs = 1;
```
avagin: Pls write a comment before this variable. It is used only in this file, so you need to mark it as static.

compel/plugins/std/log.c
```diff
@@ -51,11 +54,19 @@ static void sbuf_log_init(struct simple_buf *b)
 	 */
 	b->bp = b->buf;
 
-	if (start.tv_sec != 0) {
+	if (start.tv_sec != 0 || print_ts_diffs) {
 		struct timeval now;
 
 		sys_gettimeofday(&now, NULL);
-		timediff(&start, &now);
+		struct timeval *pivot = print_ts_diffs ? &last_t : &start;
```
avagin: all variables have to be declared at the beginning of a block where they are used.

compel/plugins/std/log.c
```diff
@@ -51,11 +54,19 @@ static void sbuf_log_init(struct simple_buf *b)
...
+		struct timeval curr = now;
+		timediff(pivot, &now);
+		if (now.tv_sec == curr.tv_sec && now.tv_usec == curr.tv_usec) {
+			// first entry will be zero
+			now.tv_sec = 0;
+			now.tv_usec = 0;
+		}
+		if (print_ts_diffs) last_t = curr;
```
avagin:
```
if (print_ts_diff)
	last_t  = curr;
```

compel/plugins/std/log.c
```diff
@@ -41,6 +43,7 @@ static inline void pad_num(char **s, int *n, int nr)
 
 static void sbuf_log_init(struct simple_buf *b)
 {
+	static struct timeval last_t;
```
avagin: why do we need another variable? Cat we use start instead of last_t?

criu/include/cr_options.h
```diff
@@ -139,6 +139,7 @@ struct cr_options {
 	pid_t			tree_id;
 	int			log_level;
 	char			*imgs_dir;
+	int         relative_timestamps;
```
avagin: you need to use tabs instead of spaces for alignments

PR #675, opened 2019-04-16
libcriu: bump SONAME to 2

adrianreber:
libcriu.so used to have the SONAME

$ readelf -d lib/c/libcriu.so   | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libcriu.so.1]

The recent changes to libcriu (removed and added functions, changes from
`char *`-args to `const char *`) are breaking ABI and API. This requires
a new SONAME of 2. This patch changes the SONAME to 2:

$ readelf -d lib/c/libcriu.so   | grep SONAME
 0x000000000000000e (SONAME)             Library soname: [libcriu.so.2]


adrianreber:
CC: @martinwuehrer (because most of the changes are from you)

CC: @avagin (to include for the new release)

PR #676, opened 2019-04-17
sk-inet: restore SO_BROADCAST option

### LINKED ISSUE #673: after udp restore: system call to 'sendto' report errno=EACCES Permission denied
I am experimenting with udp checkpoint feature. I have a udp client program and a udp server program. It seems to be OK by looking at the log. But when I tried to restore back to the checkpoint. The udp program failed to call `sendto`, with errno=EACCES (Permission denied). It seems the socket has lost 'SO_BROADCAST' attribute after restored.

My system configuration is as follows:
CentOS Linux release 7.1.1503 (Core)
CRIU 3.9

The following are the commands that I issued to create the problem.
criu dump -t 14455 --tcp-established --shell-job -vvvv -o dump.log && echo OK
date;criu restore -d -vvv --tcp-established --shell-job -o restore.log && echo OK

Code:
```c++
#include <sys/socket.h>
#include <sys/ioctl.h>
#include <linux/types.h>
#include <sys/types.h>
#include <arpa/inet.h>
#include <unistd.h>
#include <string.h>
#include <stdlib.h>
#include <stdio.h>
#include <signal.h>
#include <string.h>
#include <errno.h>

static int main_srv(int argc, char **argv)
{
    printf("main_srv\n");
    struct sockaddr_in cli;
    char request[400], reply[400];
    int clilen, n;
    int sk;
    sk = socket(AF_INET, SOCK_DGRAM, 0); 
    /* set reuse and non block for this socket */
    int on=1;
    setsockopt(sk, SOL_SOCKET, SO_REUSEADDR, (char *)&on, sizeof(on));
    setsockopt(sk, SOL_SOCKET, SO_BROADCAST, (char *)&on, sizeof(on));

    struct sockaddr_in servaddr; 
    memset(&servaddr, 0, sizeof(servaddr)); 
    servaddr.sin_family = AF_INET; 
    servaddr.sin_addr.s_addr = htonl(INADDR_ANY);
    servaddr.sin_port = htons(10000); 
    
    if (bind(sk, (struct sockaddr *) &servaddr, sizeof(servaddr)) != 0)
    {
          perror("Can't bind socket");
          return -1;
    }
    printf("bind to 10000\n");

    for (;;) {
        clilen = sizeof(cli);
        if ( (n=recvfrom(sk, request, 400, 0, (struct sockaddr *)&cli, &clilen)) < 0 ) {
            perror("recv from server error");
        }
        
        if ( sendto(sk, request, 400, 0, (struct sockaddr *)&cli, sizeof(cli)) != 400 ) {
            perror("sendto from server error");
        }
    }
}

static int main_cl(int argc, char **argv)
{
    printf("main_cl\n");
    char request[400], reply[400];
    int n;
    struct sockaddr_in servaddr; 
    int sk = socket(AF_INET, SOCK_DGRAM, 0); 
    if (sk < 0) {
        perror("Can't create socket");
        return -1;
    }
    /* set reuse and non block for this socket */
    int on=1;
    setsockopt(sk, SOL_SOCKET, SO_REUSEADDR, (char *)&on, sizeof(on));
    setsockopt(sk, SOL_SOCKET, SO_BROADCAST, (char *)&on, sizeof(on));

    struct sockaddr_in targetaddr; 
    memset(&targetaddr, 0, sizeof(targetaddr)); 
    targetaddr.sin_family = AF_INET; 
    //targetaddr.sin_addr.s_addr = htonl(INADDR_ANY);
    targetaddr.sin_addr.s_addr = inet_addr("127.255.255.255");
    targetaddr.sin_port = htons(10000); 

    int val = 0;
    while (1) {
        val++;
        snprintf(request, 400, "%d", val);
        if ( sendto(sk, request, 400, 0, (struct sockaddr *)&targetaddr, sizeof(targetaddr)) != 400 ) {
            perror("sendto from client error");
        }
        
        if ( (n=recvfrom(sk, reply, 400, 0, (struct sockaddr *)NULL, (int*) NULL)) < 0 ) {
            perror("recv from client error");
        }

        printf("UDP %s -> %s\n", request, reply);
        sleep(2);
    }
    
}

{
    if (strcmp(argv[1], "server")==0 )
        return main_srv(argc, argv);
    else if (strcmp(argv[1],"client")==0 )
        return main_cl(argc, argv);

    printf("Bad usage\n");
    return 1;
}
```

rst0git:
Inet sockets may have broadcasting capability enabled. The `SO_BROADCAST` option is used to enable this feature. It is a Boolean flag option, which is defined, fetched, and set with the int data type. During
checkpoint, CRIU should detect the state of this flag, and during restore, it should be set appropriately.

Fixes #673

### DIFF
--- a/test/zdtm/static/socket_udp-broadcast.c
+++ b/test/zdtm/static/socket_udp-broadcast.c
@@ -1,0 +1,47 @@
+#include <sys/socket.h>
+#include <sys/types.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "test checkpoint/restore of SO_BROADCAST\n";
+const char *test_author = "Radostin Stoyanov <rstoyanov1@gmail.com>\n";
+
+/* Description:
+ * Create UDP socket, set SO_BROADCAST and verify its value after restore.
+ */
+
+int main(int argc, char **argv)
+{
+	int sockfd;
+	int val;
+	socklen_t len = sizeof(val);
+
+	test_init(argc, argv);
+
+	sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sockfd < 0) {
+		pr_perror("Can't create socket");
+		return 1;
+	}
+
+	if (setsockopt(sockfd, SOL_SOCKET, SO_BROADCAST, &(int){ 1 }, len)) {
+		pr_perror("setsockopt");
+		return 1;
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	if (getsockopt(sockfd, SOL_SOCKET, SO_BROADCAST, &val, &len)) {
+		pr_perror("getsockopt");
+		return 1;
+	}
+
+	if (len != sizeof(val) || val != 1) {
+		fail("SO_BROADCAST not set");
+		return 1;
+	}
+
+	pass();
+	return 0;
+}
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -742,6 +742,10 @@ static int post_open_inet_sk(struct file_desc *d, int sk)
 	if (!val && restore_opt(sk, SOL_SOCKET, SO_REUSEPORT, &val))
 		return -1;
 
+	val = ii->ie->opts->so_broadcast;
+	if (!val && restore_opt(sk, SOL_SOCKET, SO_BROADCAST, &val))
+		return -1;
+
 	return 0;
 }
 
--- a/criu/sockets.c
+++ b/criu/sockets.c
@@ -566,6 +566,11 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 		pr_debug("\tset no_check for socket\n");
 		ret |= restore_opt(sk, SOL_SOCKET, SO_NO_CHECK, &val);
 	}
+	if (soe->has_so_broadcast && soe->so_broadcast) {
+		val = 1;
+		pr_debug("\tset broadcast for socket\n");
+		ret |= restore_opt(sk, SOL_SOCKET, SO_BROADCAST, &val);
+	}
 
 	tv.tv_sec = soe->so_snd_tmo_sec;
 	tv.tv_usec = soe->so_snd_tmo_usec;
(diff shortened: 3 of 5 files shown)


0x7f454c46:
Could you add a zdtm test for this?
I believe, the policy is to add a test for every new feature to make sure it won't be broken with some changes in future. Thanks.

test/zdtm/static/socket_udp-broadcast.c
```diff
@@ -0,0 +1,40 @@
...
+
+	test_init(argc, argv);
+
+	sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sockfd < 0) {
+		fail("Can't create socket");
+		return 1;
+	}
+
+	setsockopt(sockfd, SOL_SOCKET, SO_BROADCAST, &(int){ 1 }, sizeof(int));
```
avagin: Need to handle a return code here

test/zdtm/static/socket_udp-broadcast.c
```diff
@@ -0,0 +1,40 @@
...
+{
+	int sockfd;
+	int val;
+	socklen_t len = sizeof(val);
+
+	test_init(argc, argv);
+
+	sockfd = socket(AF_INET, SOCK_DGRAM, 0);
+	if (sockfd < 0) {
+		fail("Can't create socket");
```
avagin: pr_perror()

test/zdtm/static/socket_udp-broadcast.c
```diff
@@ -0,0 +1,40 @@
...
+		return 1;
+	}
+
+	setsockopt(sockfd, SOL_SOCKET, SO_BROADCAST, &(int){ 1 }, sizeof(int));
+
+	test_daemon();
+	test_waitsig();
+
+	getsockopt(sockfd, SOL_SOCKET, SO_BROADCAST, &val, &len);
+	if (len != sizeof(val) || val != 1) {
```
avagin: val can be used uninitialized here

rst0git: Thank you for the code review.

PR #678, opened 2019-04-17
Prepare v3.12

avagin:


adrianreber:
It has all the changes I am looking forward to. The bump of the SONAME needs to be included, but besides that it seems to include all I need.

rst0git:
@avagin Would it be possible to apply https://lists.openvz.org/pipermail/criu/2019-April/043788.html?

avagin:
> @avagin Would it be possible to apply https://lists.openvz.org/pipermail/criu/2019-April/043788.html? Otherwise make install fails when CC=gcc is not explicitly set.

Where do you see this problem?

rst0git:
> Where do you see this problem?

https://github.com/checkpoint-restore/criu/pull/678/files#diff-fc33c6390093a82b53b4b985d745e34fR9

```sh
$ cat Makefile 
all:
	$(CC ?= gcc)
	echo $(CC) $(origin CC)
$ make
echo cc default
cc default

```

avagin:
@rst0git I still don't understand what you mean. Does make fail in your enviroment?

rst0git:
Yes, it does.

```sh
$ git clone https://github.com/avagin/criu/
$ cd criu/
$ git checkout v3.12
$ make -j20 && sudo make install
...
make[2]: Nothing to be done for 'all'.
  INSTALL  criu/criu
  LINK     compel/libcompel.so
make: shared: Command not found
make: [Makefile.compel:73: compel/libcompel.so] Error 127 (ignored)
  CC       compel/arch/x86/src/lib/handle-elf.o
  CC       compel/src/lib/handle-elf.o
  CC       compel/src/main.o
  LINK     compel/built-in.o
  LINK     compel/compel
make: O2: Command not found
make: [Makefile.compel:66: compel/compel] Error 127 (ignored)
make[1]: *** No rule to make target 'compel/compel', needed by 'install'.  Stop.
make: *** [Makefile.install:44: install-compel] Error 2
```

0x7f454c46:
Hmm, late to the party (was on PTO).
I don't think this $(origin CC) hack is a proper fix.
We prune all default variable defines by using `-R` flag in nmk (I wish we used `--no-builtin-variables` in the source rather than shorthand, but whatever).
So, the actual problem @rst0git observes seems to come from 3b0c8dbd4837:
```
-compel/%: .FORCE
-       $(Q) $(MAKE) $(build)=compel $@
-
-test/compel/%: .FORCE
-       $(Q) $(MAKE) $(build)=compel $@
+include Makefile.compel
```
While I'm very up to moving into including sub-Makefiles rather than recursive make - we probably want to fix the actual problem, rather than paper-over `CC` definition.

PR #681, opened 2019-04-21
util: cr_daemon: Drop keep_fd argument

rst0git:
When running lazy-pages in daemon mode, file descriptor 3 is reused after fork to keep the opened UNIX socket. However, fd 3 happens to correspond to the opened image directory, and therefore this causes the following error:

```sh
$ criu lazy-pages -D <PATH> --page-server --address <ADDR> --port <PORT> -d
...
(06.835596) Error (criu/image.c:470): Unable to open pagemap-1.img: Not a directory
(06.835855) Error (criu/uffd.c:773): uffd: 1-7: Failed to open pagemap
```

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -629,7 +629,7 @@ int close_status_fd(void)
 	return close_safe(&opts.status_fd);
 }
 
-int cr_daemon(int nochdir, int noclose, int *keep_fd, int close_fd)
+int cr_daemon(int nochdir, int noclose, int close_fd)
 {
 	int pid;
 
@@ -651,16 +651,6 @@ int cr_daemon(int nochdir, int noclose, int *keep_fd, int close_fd)
 
 		if (close_fd != -1)
 			close(close_fd);
-
-		if ((*keep_fd != -1) && (*keep_fd != 3)) {
-			fd = dup2(*keep_fd, 3);
-			if (fd < 0) {
-				pr_perror("Dup2 failed");
-				return -1;
-			}
-			close(*keep_fd);
-			*keep_fd = fd;
-		}
 
 		fd = open("/dev/null", O_RDWR);
 		if (fd < 0) {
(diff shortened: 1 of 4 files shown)


criu/util.c
```diff
@@ -652,8 +652,8 @@ int cr_daemon(int nochdir, int noclose, int *keep_fd, int close_fd)
 		if (close_fd != -1)
 			close(close_fd);
 
-		if ((*keep_fd != -1) && (*keep_fd != 3)) {
-			fd = dup2(*keep_fd, 3);
+		if (*keep_fd != -1) {
+			fd = dup(*keep_fd);
```
avagin: If one of 0,1, or 2 isn't used in this process, dup() will return it, and then we will overwrite it by /dev/null, so we can't use dup here, don't we?

criu/util.c
```diff
@@ -652,8 +652,8 @@ int cr_daemon(int nochdir, int noclose, int *keep_fd, int close_fd)
 		if (close_fd != -1)
 			close(close_fd);
 
-		if ((*keep_fd != -1) && (*keep_fd != 3)) {
-			fd = dup2(*keep_fd, 3);
```
avagin: The first question what we need to ask our self is why we added this dup2.  In this code, we set 0, 1, 2 to /dev/null, but keep_fd can be one of this file descriptors.

criu/util.c
rst0git: Thank you for pointing this out. Perhaps we could use `fcntl(F_DUPFD)` instead?

avagin: Yes, we can. But I see another problem here. keep_fd is used to protect a file descriptor which is going to be used after cr_daemon. Here we have two such descriptros. Why do we want to handle only one of them?

rst0git: > keep_fd is used to protect a file descriptor which is going to be used after cr_daemon.

I don't understand why, though, after fork() the child will inherit a copy of all file descriptors, right? Why do we need to 'protect' `keep_fd`?

avagin: Why do we need keep_fd?

criu/util.c
rppt: that seems to be ancient history
@xemul, any comments?

criu/util.c
rst0git: @avagin https://github.com/checkpoint-restore/criu/pull/681/commits/bfb73aecf032c1779190d88c5224b546772ff259 Would that work? Do we want a test for it?

PR #685, opened 2019-04-29
This also adds the correct labels to sockets during restore

adrianreber:
If restoring a process in a SELinux environment, sockets created during restore must also have the correct labels. This sets the label for all sockets to the label the process had during restore.

As with all SELinux related labels, this assumes that all processes and sockets restored are using the same label (that is what is currently done for SELinux containers with Podman).

### DIFF
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -23,12 +23,17 @@
 #include "files.h"
 #include "image.h"
 #include "log.h"
+#include "lsm.h"
+#include "kerndat.h"
+#include "pstree.h"
 #include "rst-malloc.h"
 #include "sockets.h"
 #include "sk-inet.h"
 #include "protobuf.h"
 #include "util.h"
 #include "namespaces.h"
+
+#include "images/inventory.pb-c.h"
 
 #undef  LOG_PREFIX
 #define LOG_PREFIX "inet: "
@@ -804,11 +809,17 @@ static int open_inet_sk(struct file_desc *d, int *new_fd)
 	if (set_netns(ie->ns_id))
 		return -1;
 
+	if (run_setsockcreatecon(fle->fe))
+		return -1;
+
 	sk = socket(ie->family, ie->type, ie->proto);
 	if (sk < 0) {
 		pr_perror("Can't create inet socket");
 		return -1;
 	}
+
+	if (reset_setsockcreatecon())
+		return -1;
 
 	if (ie->v6only) {
 		if (restore_opt(sk, SOL_IPV6, IPV6_V6ONLY, &yes) == -1)
@@ -895,6 +906,7 @@ done:
 	}
 
 	*new_fd = sk;
+
 	return 1;
 err:
 	close(sk);
--- a/test/zdtm/static/selinux01.c
+++ b/test/zdtm/static/selinux01.c
@@ -1,0 +1,200 @@
+#include <unistd.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/mount.h>
+#include <sys/socket.h>
+#include <sys/xattr.h>
+#include <linux/limits.h>
+#include <signal.h>
+#include "zdtmtst.h"
+
+/* Enabling the right policy happens in selinux00.hook and selinx00.checkskip */
+
+const char *test_doc	= "Check that a SELinux socket context is restored";
+const char *test_author	= "Adrian Reber <areber@redhat.com>";
+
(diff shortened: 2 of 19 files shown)


adrianreber:
This was detected during the restore of a container. After the restore the container was no longer able to `accept()`:
```
avc:  denied  { accept } for  pid=8503 comm="java" lport=8009 scontext=system_u:system_r:container_t:s0:c99,c952 tcontext=unconfined_u:system_r:container_runtime_t:s0 tclass=tcp_socket permissive=0
```
With this patch the restored container can `accept()` connections again.

avagin:
Where is a test?

criu/sk-inet.c
```diff
@@ -804,6 +812,44 @@ static int open_inet_sk(struct file_desc *d, int *new_fd)
...
+		if (current->core[0]->thread_core &&
+		    current->core[0]->thread_core->creds &&
+		    current->core[0]->thread_core->creds->lsm_profile) {
+			/*
+			 * thread_core, thread_core->creds and
+			 * thread_core->creds->lsm_profile are all optional
+			 * protobuf fields. This makes sure these exist
+			 * before accessing them.
+			 */
+			int ret = setsockcreatecon_raw(current->core[0]->
```
avagin: A socket label can be different than a process label, can't it? Should we use fgetxattr(..., "security.selinux") to dump a socket label?

adrianreber:
Added test.

criu/sk-inet.c
adrianreber: Yes, a socket can have a different label. This is the minimal fix for the container (Podman) use case where everything in the container has the same label. It is similar to the current handling of threads. CRIU currently restores all threads with the same SELinux label. In theory threads could have different process labels (if those labels are bounded).

This fix is still needed as we need to make sure that sockets created in the future (after restore) have at least the label of the process.

In a follow up patch I can read and store all socket labels and re-create the correctly on restore.

The correct fix would also require checking if the user has used setsockcreatecon*() and restore that value as well.

If possible I would like to get this into CRIU and then the follow-up would read out /proc/PID/attr/sockcreate to see if something is set and store that for newly created sockets and read out the socket label to restore it (fgetxattr).

avagin: > Yes, a socket can have a different label. This is the minimal fix for the container (Podman) use case where everything in the container has the same label. It is similar to the current handling of threads. CRIU currently restores all threads with the same SELinux label. In theory threads could have different process labels (if those labels are bounded).

I'm ok with this, but we have to be sure that we will not restore a wrong label for a socket. For this we need to dump a socket label and then on restore we need to check that a socket label and a process label are the same.

test/zdtm/static/selinux01.c
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if (setprofile())
+		return -1;
+
+	/* Open our test socket */
+	int sk = socket(AF_INET, SOCK_STREAM, 0);
+	memset(ctx, 0, 1024);
+	/* Read out the socket label */
+	fgetxattr(sk, "security.selinux", ctx, 1024);
```
avagin: Need to handle errors of fgetxattr

adrianreber:
I completely changed it and I just saw that it no longer fixes the problem I tried to fix :frowning: 

Should have tested my actual problem earlier. Anyway, now I am trying to restore sockcreate only if it is set and I am using fgetxattr() to restore the labels of the sockets. This works pretty good for the included test cases, but not for my actual problem.

adrianreber:
fsetxattr() succeeds, but the sockets are not labelled correctly, need figure out why

adrianreber:
This now works for my use case and it correctly restores the socket labels from xattr and also correctly configures sockcreate.

avagin:
Thanks!

adrianreber:
Thanks for merging. I see a lot of CI errors with:
```
criu/lsm.c: In function ‘selinux_get_sockcreate_label’:
criu/lsm.c:119:2: error: ignoring return value of ‘fscanf’, declared with attribute warn_unused_result [-Werror=unused-result]
  fscanf(f, "%ms", output);
  ^~~~~~~~~~~~~~~~~~~~~~~~
make[2]: *** [/mnt/scripts/nmk/scripts/build.mk:119: criu/lsm.o] Error 1
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [criu/Makefile:78: criu/built-in.o] Error 2
make: *** [Makefile:228: criu] Error 2
```
I deliberately ignored the return value of `fscanf()` here but it seems I cannot. Any ideas?

0x7f454c46:
Would `!!fscanf(f, "%ms", output);` work?
(or maybe we want a special macro for such cases)

PR #691, opened 2019-05-03
lsm: fix compiler error 'unused-result'

adrianreber:
Reading out the xattr 'security.selinux' of checkpointed sockets with
fscanf() works (at least in theory) without checking the result of
fscanf(). There are, however, multiple CI failures when ignoring the
return value of fscanf().

This adds some extra (unnecessary) return code checking to make CI
happy.

See CI errors in #685

### DIFF
--- a/criu/lsm.c
+++ b/criu/lsm.c
@@ -111,20 +111,22 @@ err:
 static int selinux_get_sockcreate_label(pid_t pid, char **output)
 {
 	FILE *f;
+	int ret;
 
 	f = fopen_proc(pid, "attr/sockcreate");
 	if (!f)
 		return -1;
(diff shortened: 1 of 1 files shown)


avagin:
> This adds some extra (unnecessary) return code checking to make CI
happy.

It is necessary. In your case, fscanf has to allocate memory for a string, so it can return ENOMEM and we need to handle this error properly.

adrianreber:
@avagin, good point. I am using now ferror() to see if it actual is an error or EOF. Or should I just check for ENOMEM?

avagin:
I would check all errors of fscanf(), it should not be a problem.

PR #695, opened 2019-05-06
Preserve xattr when mounting cgroups

Snorch:
1-st patch - fix error handling while on it.
2-nd patch - get xattr cgroup mount option from existing mount to suppress noisy kernel warnings.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1698,11 +1698,11 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
 			pr_debug("\tMaking controller dir %s (%s)\n", paux, opt);
 			if (mkdir(paux, 0700)) {
 				pr_perror("\tCan't make controller dir %s", paux);
-				return -1;
+				goto err;
 			}
 			if (mount("none", paux, "cgroup", 0, opt) < 0) {
 				pr_perror("\tCan't mount controller dir %s", paux);
-				return -1;
+				goto err;
 			}
 		}
 
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1102,6 +1102,31 @@ static int userns_move(void *arg, int fd, pid_t pid)
 		close(fd);
 	}
 
+	if (fd < 0 && !pr_quelled(LOG_DEBUG)) {
+		int tmp = fd;
+
+		pr_err("Failed to openat %s at cgyard fd %d, will check upper dirs\n", (char *)arg, cg);
+		while (1) {
+			char *c;
+
+			c = strrchr(arg, '/');
+			if (!c)
+				break;
+			c[0] = '\0';
+
+			fd = openat(cg, arg, O_PATH);
+			if (fd < 0) {
+				pr_perror("Failed to openat %s too\n", (char *)arg);
+			} else {
+				pr_err("First successful openat for %s\n", (char *)arg);
+				close(fd);
+				break;
+			}
+		}
+
+		fd = tmp;
+	}
+
 	if (err < 0) {
 		pr_perror("Can't move %s into %s (%d/%d)", pidbuf, (char *)arg, err, fd);
 		return -1;
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -24,6 +24,7 @@
 #include "protobuf.h"
 #include "images/core.pb-c.h"
 #include "images/cgroup.pb-c.h"
+#include "mount.h"
 
 /*
  * This structure describes set of controller groups
@@ -51,6 +52,13 @@ static u32 cg_set_ids = 1;
 
 static LIST_HEAD(cgroups);
 static unsigned int n_cgroups;
+
+/*
+ * Should be preinitialized for cgroup_mountinfo_contains, these is to be able
+ * to test if cgroup mounts on host contain the specified options. Currently
+ * only to check 'xattr' option.
+ */
+struct mount_info *host_mountinfo;
 
 static CgSetEntry *find_rst_set_by_id(u32 id)
 {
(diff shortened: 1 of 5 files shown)


criu/cgroup.c
```diff
@@ -542,6 +545,84 @@ static int add_freezer_state(struct cg_controller *controller)
...
+			continue;
+
+		return 1;
+	}
+
+	return 0;
+}
+
+/*
+ * Test if there is a cgroup mount witch contains the options provided.
```
rst0git: s/witch/which/

Snorch: Thanks, updated.

0x7f454c46:
It sounds like it's very easy to write a test to check preserve for cgroup's mount option..
Could you add one to zdtm to prevent this being broken in future?

0x7f454c46:
@Snorch ah, I see. Sounds like a minor and no one will suffer if someone (>_>) will accidentally break it in future. dmesg check in `zdtm.py` sounds like overkill to me, I'm fine without a test.

Snorch:
Issue with cgroupns test does not reproduce for me and I'm not sure how my code could cause it... So I add another patch to find out if upper cgroup directories exist at the time of failure.

avagin:
@Snorch do we need to dump/restore xattr-s for cgroups?

rst0git:
according to https://github.com/torvalds/linux/commit/03b1cde6b22f625ae832b939bc7379ec1466aec5

>What would the use case be for this?
Attaching meta information to services, in an easily discoverable way. For example, in systemd we create one cgroup for each service, and could then store data like the main pid of the specific service as an xattr on the cgroup itself. That way we'd have almost all service state in the cgroupfs, which would make it possible to terminate systemd and later restart it without losing any state information.

Snorch:
@avagin I think we need to dump/restore these xatts, I managed to set them with "setfattr -n trusted.example -v example /sys/fs/cgroup/systemd/test" and after c/r the xattr is lost.

Snorch:
Rebased on new criu-dev with zero changes.

criu/cgroup.c
```diff
@@ -586,10 +668,17 @@ static int collect_cgroups(struct list_head *ctls)
 		if (!opts.manage_cgroups)
 			continue;
 
+		snprintf(mopts, sizeof(mopts), "xattr,%s", cc->name);
```
avagin: Pls write a comment why we need to care about xattr

criu/cgroup.c
```diff
@@ -52,6 +53,8 @@ static u32 cg_set_ids = 1;
 static LIST_HEAD(cgroups);
 static unsigned int n_cgroups;
 
+struct mount_info *host_mountinfo;
```
avagin: pls write a comment why we need this global variable

criu/cgroup.c
```diff
@@ -1700,6 +1793,17 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
...
+			if (ret < 0) {
+				pr_err("Failed to check xattr in host mountinfo\n");
+				goto err;
+			}
+			if (ret) {
+				int ooff = strlen(opt);
+				snprintf(opt + ooff, sizeof(opt) - ooff, ",xattr");
+			}
+
 			if (mount("none", paux, "cgroup", 0, opt) < 0) {
```
avagin: If we parse a host mount info, we can bind-mount a host cgroup mount instead of mounting a new one. In this case, we will not need to care about options. What do you think about this?

Snorch: If cgroup is mounted readonly on host we will get readonly bind-mount. These new mount was introduced initially to get writable cgroup mount.

criu/cgroup.c
Snorch: done.

criu/cgroup.c
Snorch: done.

criu/cgroup.c
avagin: > If cgroup is mounted readonly on host we will get readonly bind-mount. These new mount was introduced initially to get writable cgroup mount.

These new mounts was introduced, because we ditn't want to search it. It was easier to mount it, instead of parsing mountinfo and etc.

> upd: More other we don't always have a mount on host, so we still would need two cases where we do both mount and bind-mount

Is this a problem? A cgroup mount can be in another namespace and we will not find a proper options for it here. Is this a problem?

Can we always mount cgroups with xattr?

Snorch: >>If cgroup is mounted readonly on host we will get readonly bind-mount. These new mount was introduced initially to get writable cgroup mount.
>
>These new mounts was introduced, because we ditn't want to search it. It was easier to mount it, instead of parsing mountinfo and etc.

But there is still a posibility of readonly mount on host, thus making separate cgroup mount looks more general solution for me then binding (which in general requires additional writable remount).

Actually, I was wrong about the initials. I found these patch:

```
commit dda742ec570a3f6fea6d88c9b82e22306bcb1422
Author: Andrew Vagin <avagin@openvz.org>
Date:   Wed Jul 16 17:46:00 2014 +0400

    cgroup: don't look up existing mount point

    A mount point, which is mounted by someone else, may be umounted in
    any moment.
```

So 1) If someone unmounts the host's-mount, the bind-mount may fail... 2) And even more my code will not see xattr in mountinfo and likely some warnings will be still printed. But (2) is not so important as warnings are a fair cost for inproper mounts setup on host.

>
>>upd: More other we don't always have a mount on host, so we still would need two cases where we do both mount and bind-mount
>
>>Is this a problem?

Either we chouse between mount/bind-mount or we chouse between xattr/noxattr I think it will be almost the same amount of code. We still need to parse mountinfo in any case.

>>A cgroup mount can be in another namespace and we will not find a proper options for it here. Is this a problem?

If we have no mount on host but it actually existed previousely and kernel still holds 'xattr' option somthere deep inside for us, we have no way to suppress the 'missmatch' warning when mounting.

```
[root@snorch criu]# echo "cgroup DEBUG" >  /dev/kmsg
[root@snorch criu]# cat /proc/self/mountinfo | grep systemd
29 27 0:26 / /sys/fs/cgroup/systemd rw,nosuid,nodev,noexec,relatime shared:6 - cgroup cgroup rw,xattr,name=system
[root@snorch criu]# umount /sys/fs/cgroup/systemd
[root@snorch criu]# mount -t cgroup -onone,name=systemd cgroup /sys/fs/cgroup/systemd
[root@snorch criu]# cat /proc/self/mountinfo | grep systemd
29 27 0:26 / /sys/fs/cgroup/systemd rw,relatime shared:6 - cgroup cgroup rw,xattr,name=systemd
[root@snorch criu]# dmesg | grep "DEBUG\|mount" | tail -n2
[616485.628903] cgroup DEBUG
[616541.852775] new mount options do not match the existing superblock, will be ignored
```

>
>Can we always mount cgroups with xattr?

No, these way we will have missmatches for all other cgroups which initially had no option.

To conclude: to avoid 1)fall back to normal mount when src is unmounted, 2)fall back to normal mount on bindmount fail on src racy umount, 3)additional remount of the mount to rw if src is ro - I believe it is better to leave the current approach.

Snorch:
> do we need to dump/restore xattr-s for cgroups?

Yes we do in general.

Lets look on who can use it actually, initially it was enabled by systemd for it's nspawn:

a383724e42 systemd,nspawn: use extended attributes to store metadata

But now these logic is replaced and they don't put xattrs on cgroups:

(systemd-debuginfo-219-63.vl7.9.x86_64)

[root@silo systemd-219]# git grep setxattr | grep -v "\<user\."
src/journal/coredump.c:                k = fsetxattr(fd, xattrs[i], info[i], strlen(info[i]), XATTR_CREATE);
src/shared/copy.c:                        if (fsetxattr(fdt, p, bufb, m, 0) < 0)

^
always user.* - these is not a valid xattr for cgroup files

src/shared/smack-util.c:                r = lsetxattr(path, "security.SMACK64", label, strlen(label), 0);
src/shared/smack-util.c:                r = fsetxattr(fd, "security.SMACK64", label, strlen(label), 0);
src/shared/smack-util.c:                r = fsetxattr(fd, "security.SMACK64IPOUT", label, strlen(label), 0);
src/shared/smack-util.c:                r = fsetxattr(fd, "security.SMACK64IPIN", label, strlen(label), 0);
src/shared/smack-util.c:                r = lsetxattr(path, "security.SMACK64", label, strlen(label), 0);

^
The CONFIG_SECURITY_SMACK is not set in VZ7, so likely they are not used.

src/tmpfiles/tmpfiles.c:                if (lsetxattr(path, *name, *value, n, 0) < 0) {

^
Here one can do anything including xattr on cgroups if configured in /etc/tmpfiles.d/*.conf

Checking xattrs on a working system does not show any on my nodes, with:

find /sys/fs/cgroup/systemd/ -exec getfattr -d -m ".*" {} \;

So to conclude systemd itself does not use xattr on cgroups. Someone else is free to use them it's hard to estimate if someone really does it.

Disabling xattr on systemd cgroup would be a good option, though not yet sure if it is easy, as it is compiled in systemd.

Snorch:
With new --cgroup-yard option on dump and on restore we don't need these workaround anymore, we can show criu the path to host's cgroups and criu won't need to mount them. No mount -> no xattr option missmatch.

PR #697, opened 2019-05-15
Fixes for coverity and clang warnings

adrianreber:
I have been running some code analysis against criu 3.12 and this fixes a few of the reported errors.

All tests are still successful, but as this touches many different parts of criu I am not I did not break something.

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -776,6 +776,7 @@ static struct collect_image_info remap_cinfo = {
 static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_dev)
 {
 	struct cr_img *img;
+	int exit_code = -1;
 	GhostFileEntry gfe = GHOST_FILE_ENTRY__INIT;
 	Timeval atim = TIMEVAL__INIT, mtim = TIMEVAL__INIT;
 
@@ -812,7 +813,7 @@ static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_de
 	}
 
 	if (pb_write_one(img, &gfe, PB_GHOST_FILE))
-		return -1;
+		goto err_out;
 
 	if (S_ISREG(st->st_mode)) {
 		int fd, ret;
@@ -826,7 +827,7 @@ static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_de
 		fd = open(lpath, O_RDONLY);
 		if (fd < 0) {
 			pr_perror("Can't open ghost original file");
-			return -1;
+			goto err_out;
 		}
 
 		if (gfe.chunks)
@@ -835,11 +836,13 @@ static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_de
 			ret = copy_file(fd, img_raw_fd(img), st->st_size);
 		close(fd);
 		if (ret)
-			return -1;
+			goto err_out;
 	}
 
+	exit_code = 0;
+err_out:
 	close_image(img);
-	return 0;
+	return exit_code;
 }
 
 struct file_remap *lookup_ghost_remap(u32 dev, u32 ino)
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -627,6 +627,7 @@ static int collect_threads(struct pstree_item *item)
 {
 	struct seccomp_entry *task_seccomp_entry;
 	struct pid *threads = NULL;
+	struct pid *tmp = NULL;
 	int nr_threads = 0, i = 0, ret, nr_inprogress, nr_stopped = 0;
 
 	task_seccomp_entry = seccomp_find_entry(item->pid->real);
@@ -643,9 +644,11 @@ static int collect_threads(struct pstree_item *item)
 	}
 
 	/* The number of threads can't be less than already frozen */
-	item->threads = xrealloc(item->threads, nr_threads * sizeof(struct pid));
-	if (item->threads == NULL)
-		return -1;
+	tmp = xrealloc(item->threads, nr_threads * sizeof(struct pid));
+	if (tmp == NULL)
+		goto err;
+
+	item->threads = tmp;
 
 	if (item->nr_threads == 0) {
 		item->threads[0].real = item->pid->real;
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -573,7 +573,7 @@ static int do_dump_one_inet_fd(int lfd, u32 id, const struct fd_parms *p, int fa
 		ie.ip_opts->raw = NULL;
 
 	if (pb_write_one(img_from_set(glob_imgset, CR_FD_FILES), &fe, PB_FILE))
-		goto err;
+		err = -1;
 err:
 	ip_raw_opts_free(&ipopts_raw);
 	release_skopts(&skopts);
--- a/criu/util.c
+++ b/criu/util.c
@@ -878,6 +878,12 @@ void split(char *str, char token, char ***out, int *n)
 		cur++;
 	}
 
+	if (*n == 0) {
+		/* This can only happen if str == NULL */
+		*out = NULL;
+		*n = -1;
+		return;
+	}
 
 	*out = xmalloc((*n) * sizeof(char *));
 	if (!*out) {
@@ -1088,7 +1094,7 @@ int run_tcp_server(bool daemon_mode, int *ask, int cfd, int sk)
 		ret = cr_daemon(1, 0, cfd);
 		if (ret == -1) {
 			pr_err("Can't run in the background\n");
-			goto out;
+			goto err;
 		}
 		if (ret > 0) { /* parent task, daemon started */
 			close_safe(&sk);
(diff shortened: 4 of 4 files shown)


adrianreber:
Related: #698

criu/mount.c
```diff
@@ -1196,7 +1196,8 @@ bool mnt_is_overmounted(struct mount_info *mi)
 
 static int set_is_overmounted(struct mount_info *mi)
 {
-	mnt_is_overmounted(mi);
+	if (mnt_is_overmounted(mi))
```
avagin: mnt_is_overmounted returns false if a mount is overmounted, why should we return -1 here? Cc: @Snorch

criu/pagemap.c
```diff
@@ -486,6 +486,9 @@ static void free_pagemaps(struct page_read *pr)
 {
 	int i;
 
+	if (pr->pmes == NULL)
+		return;
+
```
avagin: I think here is the right fix for this warning:
```
diff --git a/criu/pagemap.c b/criu/pagemap.c
index 8d2e16320..8141f6793 100644
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -737,11 +737,13 @@ static int init_pagemaps(struct page_read *pr)
 
                pr->nr_pmes++;
                if (pr->nr_pmes >= nr_pmes) {
+                       PagemapEntry **new;
                        nr_pmes += nr_realloc;
-                       pr->pmes = xrealloc(pr->pmes,
+                       new = xrealloc(pr->pmes,
                                            nr_pmes * sizeof(*pr->pmes));
-                       if (!pr->pmes)
+                       if (!new)
                                goto free_pagemaps;
+                       pr->pmes = new;
                }
        }
 
```

criu/mount.c
Snorch: True is returned then mount is overmounted  and false if it is not-overmounted. both states are valid, also the state is cached on on mount_info. We call mnt_is_overmounted here to prepare the cache (in next steps of mount restore we slightly corrupt mount tree so to have the right result of mnt_is_overmounted we need to compute it here), there is no sense in checking the return code.

adrianreber: Thanks. I will drop this.

adrianreber:
Dropped the commit with mnt_is_overmounted and applied the changes from @avagin 

Let me know if I should drop or change anything else.

avagin:
```￼
autofs: fix coverity RESOURCE_LEAK ￼…
autofs: fix coverity RESOURCE_LEAK ￼…
autofs: fix coverity RESOURCE_LEAK ￼…
autofs: fix coverity RESOURCE_LEAK ￼…
```

Could you merge patches with the same type of issues per-subsystem?

nit: "fix clang warning " is a bad commit message subject...

adrianreber:
> Could you merge patches with the same type of issues per-subsystem?

Sure, I will merge the corresponding patches together.

> nit: "fix clang warning " is a bad commit message subject...

Will also change that.

criu/net.c
```diff
@@ -2059,6 +2059,7 @@ static inline int restore_iptables(int pid)
 		return -1;
 	if (empty_image(img)) {
 		ret = 0;
+		close_image(img);
 		goto ipt6;
```
rst0git: There is one more resource leak in `criu/net.c`, `tmp_file` is not closed in case of error in the `restore_ip_dump()` function. @adrianreber Could you fix this one as well, please?

adrianreber: done, please have a look

adrianreber:
squashed, reworded, rebased, force pushed

criu/net.c
rst0git: thank you!

criu/config.c
```diff
@@ -92,9 +92,10 @@ int parse_statement(int i, char *line, char **configuration)
 			}
 		}
 
-		offset += strlen(configuration[i]);
+		if (configuration[i])
```
avagin: can you explain when configuration[i] can be null here?

criu/files-reg.c
```diff
@@ -776,6 +776,7 @@ static struct collect_image_info remap_cinfo = {
 static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_dev)
 {
 	struct cr_img *img;
+	int ret = 0;
```
avagin: Usually, we call this variable as exit_code and initialize it to -1.

criu/proc_parse.c
```diff
@@ -853,6 +853,7 @@ int parse_pid_stat(pid_t pid, struct proc_pid_stat *s)
 		return -1;
 	}
 
+	buf[n] = 0;
```
avagin: buf is always null-terminated

criu/seize.c
```diff
@@ -645,7 +645,7 @@ static int collect_threads(struct pstree_item *item)
 	/* The number of threads can't be less than already frozen */
 	item->threads = xrealloc(item->threads, nr_threads * sizeof(struct pid));
 	if (item->threads == NULL)
-		return -1;
```
avagin: we need to restore a previous value of item->threads

criu/sk-inet.c
```diff
@@ -572,8 +572,7 @@ static int do_dump_one_inet_fd(int lfd, u32 id, const struct fd_parms *p, int fa
 	if (type != SOCK_RAW)
 		ie.ip_opts->raw = NULL;
 
-	if (pb_write_one(img_from_set(glob_imgset, CR_FD_FILES), &fe, PB_FILE))
-		goto err;
+	pb_write_one(img_from_set(glob_imgset, CR_FD_FILES), &fe, PB_FILE);
```
avagin: do we need to return -1 if pb_write_one failed?

criu/util.c
```diff
@@ -870,12 +870,17 @@ void split(char *str, char token, char ***out, int *n)
 	int i;
 	char *cur;
 
+	if (str == NULL)
```
avagin: This should be out of this function. The function has to be called of non-null strings.

criu/config.c
adrianreber: `} else if ((configuration[i] = strtok_r(line + offset, "\"", &quotedptr))) {` could assign NULL to configuration[i]

criu/proc_parse.c
adrianreber: Ah, because of the struct member end which comes after buf.

criu/seize.c
adrianreber: Sorry, I do not understand what you mean? Should I just drop this?

criu/sk-inet.c
adrianreber: We did not until now, but setting err to -1 would be probably a good thing.

criu/seize.c
avagin: if item->thread was not null, we will have a memory leak. Should we do something like this:
tmp = xrealloc(....)
if (tmp == NULL)
  return -1;
item->thread = tmp;

criu/config.c
avagin: We need to write a comment what it means here. If we expect to see null here, we need to describe a case when this can happen. If we don't expect to see NULL here, we need to report an error and log it.

criu/util.c
adrianreber: Not sure why I did this. Seems I misinterpreted one report. Will remove it.

criu/files-reg.c
adrianreber: I changed the variable name, but it needs an additional patch to fix `copy_file()` to return 0 on success.

criu/config.c
adrianreber: I had a closer look at this and this is a false positive. In theory the report is correct, but strtok is searching for `"` which is checked before to be there, so strtok will always find something and never return NULL. These checks are not necessary. I dropped the patch.

criu/seize.c
adrianreber: fixed

criu/util.c
```diff
@@ -1109,7 +1111,7 @@ int run_tcp_server(bool daemon_mode, int *ask, int cfd, int sk)
 		return -1;
 
 	if (sk >= 0) {
-		ret = *ask = accept(sk, (struct sockaddr *)&caddr, &clen);
+		*ask = accept(sk, (struct sockaddr *)&caddr, &clen);
 		if (*ask < 0)
 			pr_perror("Can't accept connection to server");
```
avagin: goto out?

avagin: And could you rename out to err?

adrianreber: done

criu/util.c
```diff
@@ -878,6 +878,8 @@ void split(char *str, char token, char ***out, int *n)
 		cur++;
 	}
 
+	if (*n == 0)
```
avagin: *out = NULL?

avagin: it is possible only if str is null, I am not sure that we want to handle this case. But if we want, we need to add a comment here.

adrianreber: I added a comment. Let me know if I should just drop it.

criu/files-reg.c
```diff
@@ -826,20 +827,22 @@ static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_de
...
-			return -1;
+			goto err_out;
 		}
 
 		if (gfe.chunks)
-			ret = copy_file_to_chunks(fd, img, st->st_size);
+			exit_code = copy_file_to_chunks(fd, img, st->st_size);
 		else
-			ret = copy_file(fd, img_raw_fd(img), st->st_size);
+			exit_code = copy_file(fd, img_raw_fd(img), st->st_size);
```
avagin: The whole idea of exit_code is not to use it as a return value for other functions

avagin: ping...

adrianreber: I hope I got it finally right.

avagin:
Applied. Thanks a lot!

PR #701, opened 2019-05-18
zdtm: Fix memory and resource leaks

rst0git:
These errors were found by Cppcheck 1.84

### DIFF
--- a/test/zdtm/static/s390x_runtime_instr.c
+++ b/test/zdtm/static/s390x_runtime_instr.c
@@ -147,9 +147,11 @@ int main(int argc, char **argv)
 			test_waitsig();
 			skip("RI not supported");
 			pass();
+			free(buf);
 			return 0;
 		}
 		fail("Fail with error %d", errno);
+		free(buf);
 		return -1;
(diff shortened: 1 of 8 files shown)


test/zdtm/static/sock_peercred.c
```diff
@@ -67,6 +67,7 @@ int main(int argc, char **argv)
 	socklen_t len;
 	char *stack;
 	pid_t pid;
+	int ret = 1;
```
avagin: Usually, we use exit_code name for this variable.

test/zdtm/static/s390x_runtime_instr.c
```diff
@@ -142,6 +142,7 @@ int main(int argc, char **argv)
 	memset(&ricb, 0, sizeof(ricb));
 	/* Initialize the default RI control block in the kernel */
 	if (syscall(__NR_s390_runtime_instr, 1, NULL) < 0) {
+		free(buf);
```
avagin: free can overwrite errno

PR #703, opened 2019-05-22
make: Use asciidoctor by default

rst0git:
The final release of [asciidoc](https://github.com/asciidoc/asciidoc) was on Sep 29, 2017 and the development is continued under [asciidoctor](https://github.com/asciidoctor/asciidoctor).

In commit 0493724 was added support for this tool by introducing `USE_ASCIIDOCTOR`. However, using asciidoctor by default might be a better option.

With this change CRIU will use asciidoctor if installed. Otherwise, it will fallback to asciidoc.

### DIFF
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -36,7 +36,7 @@ RUN apk add \
 	bash \
 	go \
 	e2fsprogs \
-	asciidoc xmlto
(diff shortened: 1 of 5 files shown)


0x7f454c46:
Mistakenly cancelled the build on the Travis-CI (I thought it's on my branch).
Already restarted, sorry for any possible noise.

PR #704, opened 2019-05-23
python: Drop six dependency

rst0git:
From the python-six module is used only six.string_types in the is_string() function. An alternative solution is to use basestring with additional if statement for Python 3 compatibility.

This change avoids the dependency on the six module.

However, this module is required by junit_xml and it is not listed as a dependency in the CentOS 7 package python2-junit_xml.


avagin:
[root@fc24 criu]# git grep six scripts/
scripts/build/Dockerfile.centos:        python-six \

PR #712, opened 2019-05-29
files-reg: Fix return value of copy_file()

rst0git:
https://lists.openvz.org/pipermail/criu/2019-May/044010.html

The function `copy_file()` is used only in the `files-reg.c` file. In this PR it is moved from the `util.c` to `files-reg.c` and its return value is set to 0 on success.

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -296,6 +296,58 @@ static int copy_file_from_chunks(struct cr_img *img, int fd, size_t file_size)
 	}
 }
 
+static int copy_file(int fd_in, int fd_out, size_t bytes)
+{
+	ssize_t written = 0;
+	size_t chunk = bytes ? bytes : 4096;
+	char *buffer;
+	ssize_t ret;
+
+	buffer = xmalloc(chunk);
+	if (buffer == NULL) {
+		pr_perror("failed to allocate buffer to copy file");
+		return -1;
+	}
+
+	while (1) {
+		if (opts.remote) {
+			ret = read(fd_in, buffer, chunk);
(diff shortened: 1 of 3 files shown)


avagin:
Why have our tests not caught up this bug?

rst0git:
> Why have our tests not caught up this bug?

This is a good question. Looking closer at the code, we can see that:

```C
int copy_file(int fd_in, int fd_out, size_t bytes)
{
	ssize_t written = 0;
	ssize_t ret;
...
	while (1) {   /* <--- This is an infinite loop  */
		if (opts.remote) {
			ret = read(fd_in, buffer, chunk);
			if (ret < 0) {
				pr_perror("Can't read from fd_in\n");
				ret = -1;
				goto err;
			}
			if (write(fd_out, buffer, ret) != ret) {
				pr_perror("Couldn't write all read bytes\n");
				ret = -1;
				goto err;
			}
		} else
			ret = sendfile(fd_out, fd_in, NULL, chunk);

		if (ret < 0) {
			pr_perror("Can't send data to ghost file");
			ret = -1;
			goto err;
		}

		if (ret == 0) {   /*  <--- This is the exit condition */
			if (bytes && (written != bytes)) {
				pr_err("Ghost file size mismatch %zu/%zu\n",
						written, bytes);
				ret = -1;
				goto err;
			}
			break;
		}
		written += ret;
	}
err:
	xfree(buffer);
	return ret;  /*  <-- Actually, we return 0 here */
}
```
I'm sorry for the confusion. It seems like the return value is correct, it is just not obvious.

PR #715, opened 2019-05-29
criu: Add --keep-on-exec option

criupatchwork:
Purge CONFIG_VDSO from CRIU and add arm32 helpers.                                                                                                                                            
Tested on RPI2 B+ board with vdso00/vdso02/vdso-proxy                                                                                                                                         
and fault-injected vdso proxification.                                                                                                                                                        
                                                                                                                                                                                              
Changes to v2:                                                                                                                                                                                
- added a big comment to criu-check test                                                                                                                                                      
- more precise messages when the check fails                                                                                                                                                  
- patch to omit inserting trampolines into vsyscall on ia32


criu/util.c
```diff
@@ -515,6 +515,27 @@ int cr_system(int in, int out, int err, char *cmd, char *const argv[], unsigned
 	return cr_system_userns(in, out, err, cmd, argv, flags, -1);
 }
 
+static int cmp_keep_on_exec(const void *__a, const void *__b)
+{
```
rst0git: @cyrillos this function is also defined in `criu/config.c`. Could we reuse it to avoid code duplication?

cyrillos: I happen to miss it. Will take a look, thanks!

cyrillos: [PATCH v2] criu: Add --keep-on-exec option

xemul:
Who needs which fds opened by criu in its action scripts?

cyrillos:
> Who needs which fds opened by criu in its action scripts?

libvzctl passes own files which are not supposed to be closed inside criu, in particular `statusfd` and `waitfd`.

xemul:
I see. Then, this should be done the other way.

1. All descriptors opened by criu are opened with O_CLOEXEC flag set (AFAIR either the open or openat allows this in one systemcall)
2. Any caller of CRIU sets O_CLOEXEC on descriptors it wants
3. We drop the close_all_fds() alltogether
4. We do not merge this PR

:)

xemul:
Oh and one more question. If libvzctl _hopes_ that some files will pass through criu to the action scripts, how does it make sure that criu FDs restore engine won't dup2 anything into it?

cyrillos:
> Oh and one more question. If libvzctl _hopes_ that some files will pass through criu to the action scripts, how does it make sure that criu FDs restore engine won't dup2 anything into it?

As far as I remember scripts are executed from root mount namespace while the restore of target files with desired descriptor number happens _inside_ designated namespace, thus we won't interfere with fds opened by a caller.

But I might be wrong.

cyrillos:
> I see. Then, this should be done the other way.

I'll think about it. You know there is no rush since I made a fast fix for vz7 criu instance which simply works by now. Surely we need a general solution in a long term.

xemul:
> with desired descriptor number happens _inside_ designated namespace, thus we won't interfere with fds opened by a caller.

No, this doesn't work this way. File _descriptors_ do not care the namespace you created them in :)

cyrillos:
> > with desired descriptor number happens _inside_ designated namespace, thus we won't interfere with fds opened by a caller.
> 
> No, this doesn't work this way. File _descriptors_ do not care the namespace you created them in :)

They *do* care because they are bound to `task_struct::fds` and the problem may appear if only you've cloned with fds shared, which we don't.

When we create fd we create them with the number os provide us, and if there are already inherited fds from a caller we won't touch them. Then once we created new task inside desired pid-namespace we transfer previously created fd via unix scm and reopen them with desired number, but they are opened inside own `task_struct::fds` and can't interfere with fds inerited from libvzctl.

cyrillos:
To be honest there is a one corner case -- service descriptors, need to revisit this moment, I fear that theoretically we can close inherited fds by accident in case if libvzctl provide us fd with number somewhere close to fds limits.

xemul:
Wait, wait. Consider libvzctl has a descriptor nr 5 that he wants to pass into the aciton script. If fork()+exec()-s criu, criu gets fd 5 which he doesn't know about. Then criu does fds restore magic and kills the 5th fd with anything (with temporary fd holder or with final fd), then it calls action script which expects to see something from vzctl in fd 5. BOOM!

cyrillos:
> Wait, wait. Consider libvzctl has a descriptor nr 5 that he wants to pass into the aciton script. If fork()+exec()-s criu, criu gets fd 5 which he doesn't know about. Then criu does fds restore magic and kills the 5th fd with anything (with temporary fd holder or with final fd), then it calls action script which expects to see something from vzctl in fd 5. BOOM!

The fds criu creates to restore children's fds are generated via plain open() call so that number won't be 5 as it is already busy. And criu won't close fd=5 in the *root* criu process. Actually this is the fact libvzctl been relaying on (until we start calling close_fds() and killed this logic :)

PR #728, opened 2019-06-26
Remove unexpected inotify events

Snorch:
"I've mentioned the problem that after c/r each inotify watch which
monitors close (IN_CLOSE_NOWRITE) events, gets an unexpected close
event. That happens because our algorithm first opens the file to be
watched, second adds watch on it through /proc/pid/fd/X path, and
third closes the file - adds a close event..."


avagin:
```
[root@fc24 criu]# git diff
diff --git a/test/zdtm/static/inotify04.c b/test/zdtm/static/inotify04.c
index 2650dce0c..0ecd9987e 100644
--- a/test/zdtm/static/inotify04.c
+++ b/test/zdtm/static/inotify04.c
@@ -82,6 +82,12 @@ int main (int argc, char *argv[])
                pr_perror("Failed inotify_add_watch");
                return 1;
        }
+       fd = open(test_file_path, O_WRONLY, 0644);
+       if (fd < 0) {
+               pr_perror("Failed to create %s", test_file_path);
+               return 1;
+       }
+       close(fd);
 
        test_daemon();
        test_waitsig();
@@ -90,7 +96,7 @@ int main (int argc, char *argv[])
        if (ret < 0) {
                fail("Failed to read inotify events");
                return 1;
-       } else if (n != 0) {
+       } else if (n != 2) {
                fail("Found %d unexpected inotify events", n);
                return 1;
        }
[root@fc24 criu]# python test/zdtm.py run -t zdtm/static/inotify04 --iter 0 -f h
=== Run 1/1 ================ zdtm/static/inotify04
======================== Run zdtm/static/inotify04 in h ========================
Start test
./inotify04 --pidfile=inotify04.pid --outfile=inotify04.out --dirname=inotify04.test
Send the 15 signal to  36
Wait for zdtm/static/inotify04(36) to die for 0.100000
Removing dump/zdtm/static/inotify04/36
======================= Test zdtm/static/inotify04 PASS ========================
[root@fc24 criu]# python test/zdtm.py run -t zdtm/static/inotify04  -f h
=== Run 1/1 ================ zdtm/static/inotify04
======================== Run zdtm/static/inotify04 in h ========================
Start test
./inotify04 --pidfile=inotify04.pid --outfile=inotify04.out --dirname=inotify04.test
Run criu dump
Run criu restore
Send the 15 signal to  36
Wait for zdtm/static/inotify04(36) to die for 0.100000
############### Test zdtm/static/inotify04 FAIL at result check ################
08:23:12.301:    36: Read 0 events
08:23:12.301:    36: FAIL: inotify04.c:100: Found 0 unexpected inotify events (errno = 11 (Resource temporarily unavailable))

```

Snorch:
You create two ievents open and close before c/r and they are lost after c/r.

Why do you think that I've also fixed that pre c/r events are lost, it was always like these and these patch does not try to fix these... I don't think that it even always possible with current interface.

Snorch:
Bad close.

avagin:
@Snorch why do you think that I remember that we don't restore events? Could you print an error or warning if an inotify fd has any events on dump?

Snorch:
@avagin I found that we already print a warning when there are inotify events in queue, with the patch below and your patch we see it. Will send separately:

```
[snorch@snorch criu]$ git diff test/zdtm.py 
index 57eb68a49..718147a57 100755
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1937,7 +1937,7 @@ default_test = {}
                        before.append(l)
                        if len(before) > 5:
                                before.pop(0)
-                       if "Error" in l:
+                       if "Error" in l or "Warn" in l:
                                if first:
                                        print_fname(fname, 'log')
                                        print_sep("grep Error", "-", 60)
```
Result:
```
(00.002390) netlink: Collect netlink sock 0x3ab47d
(00.002392) netlink: Collect netlink sock 0x3ab491
(00.002393) netlink: Collect netlink sock 0x3ab47b
(00.002395) netlink: Collect netlink sock 0x3ab47c
(00.002423) Warn  (criu/image.c:137): Failed to open parent directory
(00.003532) 54 fdinfo 2: pos:                0 flags:           102001/0
(00.003554) 54 fdinfo 3: pos:                0 flags:           100000/0
(00.003561) Dumping path for 3 fd via self 17 [/home/snorch/devel/criu/test/zdtm/static]
(00.003578) 54 fdinfo 4: pos:                0 flags:             4000/0
(00.003598) Warn  (criu/fsnotify.c:336): fsnotify: The 0x000007 inotify events will be dropped
```

criu/fsnotify.c
```diff
@@ -602,6 +606,31 @@ static int restore_one_inotify(int inotify_fd, struct fsnotify_mark_info *info)
 
 err:
 	close_safe(&target);
+
+	/*
+	 * As we close target after setting up an inotify on it, we create
+	 * an unwanted inotify event, we need to remove these event.
```
avagin: s/these/this/

criu/fsnotify.c
```diff
@@ -602,6 +606,31 @@ static int restore_one_inotify(int inotify_fd, struct fsnotify_mark_info *info)
...
+	 */
+	if (!ret && (mask & IN_CLOSE_NOWRITE)) {
+		char buff[EVENT_BUF_SIZE];
+		struct inotify_event *ie = (struct inotify_event *)buff;
+		int ielen;
+		int len;
+
+		len = read(inotify_fd, buff, EVENT_BUF_SIZE);
```
avagin: could you add a comment that we don't restore queued events, so we can read all events here.

criu/fsnotify.c
```diff
@@ -602,6 +606,31 @@ static int restore_one_inotify(int inotify_fd, struct fsnotify_mark_info *info)
...
+		if (len < 0) {
+			pr_perror("Failed to read event from inotify %#x", inotify_fd);
+			return -1;
+		}
+
+		ielen = EVENT_SIZE + ie->len;
+		if ((len != ielen) || (ie->mask != IN_CLOSE_NOWRITE)) {
+			pr_perror("Only one close ievent expected, len %d != %d on %#x",
```
avagin: we need to have two separate errors:
- received more than one event
- the event has a wrong type

Snorch: done.

criu/fsnotify.c
Snorch: done.

criu/fsnotify.c
Snorch: done.

Snorch:
Found the problem that if we recreate N inotify fds on the same file, these can end up in multiple close/open events in each file, will try to rework.

```
diff --git a/test/zdtm/static/inotify04.c b/test/zdtm/static/inotify04.c
index 2650dce0c..fb9293024 100644
--- a/test/zdtm/static/inotify04.c
+++ b/test/zdtm/static/inotify04.c
@@ -53,7 +53,7 @@ int main (int argc, char *argv[])
 {
        unsigned int mask = IN_ALL_EVENTS;
        char test_file_path[PATH_MAX];
-       int fd, ifd, ret;
+       int fd, ifd, ifd2, ret;
        unsigned int n;
 
        test_init(argc, argv);
@@ -78,11 +78,22 @@ int main (int argc, char *argv[])
                return 1;
        }
 
+       ifd2 = inotify_init1(IN_NONBLOCK);
+       if (ifd2 < 0) {
+               pr_perror("Failed inotify_init");
+               return 1;
+       }
+
        if (inotify_add_watch(ifd, test_file_path, mask) < 0) {
                pr_perror("Failed inotify_add_watch");
                return 1;
        }
 
+       if (inotify_add_watch(ifd2, test_file_path, mask) < 0) {
+               pr_perror("Failed inotify_add_watch");
+               return 1;
+       }
+
        test_daemon();
        test_waitsig();
 
@@ -95,7 +106,17 @@ int main (int argc, char *argv[])
                return 1;
        }
 
+       ret = inotify_read_events(ifd, &n);
+       if (ret < 0) {
+               fail("Failed to read inotify events");
+               return 1;
+       } else if (n != 0) {
+               fail("Found %d unexpected inotify events", n);
+               return 1;
+       }
+
        close(ifd);
+       close(ifd2);
        unlink(test_file_path);
        pass();

############### Test zdtm/static/inotify04 FAIL at result check ################
14:30:36.166:    36: Event       0x20
14:30:36.166:    36: Event       0x10
14:30:36.166:    36: Read 2 events
14:30:36.166:    36: FAIL: inotify04.c:105: Found 2 unexpected inotify events (errno = 11 (Resource temporarily unavailable))

##################################### FAIL #####################################```

**Comment by Snorch** [comment]
Updated with multiple inotify events.

**Comment by Snorch** [comment]
Fixed unexpected BUGON trigger (coredump) in cleanup_current_inotify_events.

File: criu/include/restorer.h [hunk 4]
```diff
@@ -264,14 +264,21 @@ enum {
...
 	 * purely to make sure all tasks be in sync.
 	 */
 	CR_STATE_FORKING,
+	/*
+	 * All tasks oppened all their fds. Stage is needed for inotify events
+	 * cleanup. These is also a special stage like CR_STATE_FORKING and is
+	 * barriered by the root task, this task also switches the stage (into
+	 * restoring).
+	 */
+	CR_STATE_FDS,
```
avagin: I am not sure that we want to add a new stage to fix this minor problem. A new stage will affect performance even if processes don't have inotify descriptors.

Snorch: From second time look, It might be easy to do it without new stage. => will do.

Snorch: upd: Not so easy. For now the closest sync point where all fds for all processes are prepared is after restore_finish_stage(task_entries_local, CR_STATE_RESTORE); in __export_restore_task which is in PIE code. I'm not common with PIE peculiarities, if I use cleanup_current_inotify_events from there I get compile problems:

Error (compel/src/lib/handle-elf-host.c:335): Unexpected undefined symbol: `cleanup_current_inotify_events'. External symbol in PIE?

Some advice will be appreciated.

Snorch:
Please reopen, I'm still working on getting rid of new stage

Snorch:
Updated without additional fds restore stage.

criu/include/restorer.h
```diff
@@ -177,6 +177,9 @@ struct task_restore_args {
 	pid_t				*zombies;
 	unsigned int			zombies_n;
 
+	int				*inotify_fds;
```
avagin: we need a comment here which explains why we need inotify_fds in restore...

criu/pie/restorer.c
```diff
@@ -1243,6 +1244,73 @@ static inline int restore_child_subreaper(int child_subreaper)
...
+ */
+static int cleanup_inotify_events(int inotify_fd)
+{
+	char buf[EVENT_BUFF_SIZE * 8];
+	int ret;
+
+	while (1) {
+		ret = fd_poll(inotify_fd);
+		if (ret < 0) {
+			pr_err("Failed to poll from inotify fd\n");
```
avagin: : %d", ret - we want to know errno

criu/pie/restorer.c
```diff
@@ -1243,6 +1244,73 @@ static inline int restore_child_subreaper(int child_subreaper)
...
+			return -1;
+		}
+	}
+
+	return 0;
+}
+
+/*
+ * When we restore inotifies we can open and close files we create a watch
+ * for. So wee need to cleanup these auxiliary events which we've generated.
```
avagin: Pls add in a comment that we don't restore inotify events, so we can just clean all events from a queue.

criu/pie/restorer.c
```diff
@@ -1243,6 +1244,73 @@ static inline int restore_child_subreaper(int child_subreaper)
...
 
+static int fd_poll(int inotify_fd)
+{
+	struct pollfd pfd = {inotify_fd, POLLIN, 0};
+#ifdef SYS_poll
+	return sys_poll(&pfd, 1, 0);
+#else
+	struct timespec tmo = {0, 0};
+
+	return sys_ppoll(&pfd, 1, &tmo, NULL, sizeof(sigset_t));
```
avagin: Why we can't use ppoll in all cases?

criu/include/restorer.h
Snorch: done.

criu/pie/restorer.c
Snorch: done.

criu/pie/restorer.c
Snorch: done.

criu/pie/restorer.c
Snorch: Replaced sys_poll with sys_ppoll completely. Thanks, didn't even thought about it, now it looks better.

Snorch:
@avagin Thanks a lot for you comments! Reworked the patch accordingly.

PR #729, opened 2019-06-27
py: Reformat everything into pep8 style

xemul:
As discussed on the mailing list, current .py files formatting does not
conform to the world standard, so we should better reformat it. For this
the yapf tool is used. The command I used was

  yapf -i $(find -name *.py)

Signed-off-by: Pavel Emelyanov <xemul@virtuozzo.com>


adrianreber:
Thanks, this will make things much easier in the future. It was always difficult to get the editor to use the right formatting for CRIU's python files. Overall I am :+1: . CI is also happy besides a single unrelated failure.

rst0git:
Unfortunately `yapf` has missed the `\t` in doc-strings. Here is a [patch.txt](https://github.com/checkpoint-restore/criu/files/3339303/patch.txt) that should fix that.

rst0git:
Perhaps we should also update `scripts/flake8.cfg`?

xemul:
> Unfortunately `yapf` has missed the `\t` in doc-strings. Here is a [patch.txt](https://github.com/checkpoint-restore/criu/files/3339303/patch.txt) that should fix that.

Thanks! I'll merge it and push the updated branch shortly.

xemul:
> Perhaps we should also update `scripts/flake8.cfg`?

Good point. I've played a bit with it, it seems like yapf has fixed us everything but

 E501 line too long
 W504 line break after binary operator

and the latter one seems to be conflicting with W503? %)

coredump/criu_coredump/coredump.py
```diff
@@ -36,795 +36,802 @@
...
+
+            if not found:
+                continue
+
+            if "in_parent" in m and m["in_parent"] == True:
+                ppid = self.pstree[pid]["ppid"]
+                return self._get_page(ppid, page_no)
+            else:
+                with open(self._imgs_dir + "/" + "pages-" + str(pages_id) +
+                          ".img") as f:
```
rst0git: This line could be reformatted to:
```
with open(self._imgs_dir + "/pages-%s.img" % pages_id) as f:
```

lib/py/images/pb2dict.py
```diff
@@ -29,350 +29,400 @@
...
-	if _marked_as_odev(field):
-		return "%d:%d" % (os.major(value), os.minor(value))
-	else:
-		return "%d:%d" % (value >> kern_minorbits, value & ((1 << kern_minorbits) - 1))
+    if _marked_as_odev(field):
+        return "%d:%d" % (os.major(value), os.minor(value))
+    else:
+        return "%d:%d" % (value >> kern_minorbits, value &
+                          ((1 << kern_minorbits) - 1))
+
```
rst0git:
```
        return "%d:%d" % (value >> kern_minorbits,
                          value & ((1 << kern_minorbits) - 1))
```

PR #730, opened 2019-06-28
zdtm: Refactor seccomp_filter_{threads,tsync}

rst0git:
As discusses on the [musl mailing list](https://www.openwall.com/lists/musl/2019/06/26/7) when libc api is used to create a POSIX thread, and this thread is killed by seccomp, this breaks a fundamental assumption the C runtime relies on, causing any libc call (i.e. pthread_join) after the kill to have undefined behaviour.

In order to work around the issue we could use `SECCOMP_RET_ERRNO` instead of `SECCOMP_RET_KILL`. This filter will set a magic value to user space as `errno` without executing the system call.


cyrillos:
Acked-by: Cyrill Gorcunov <gorcunov@gmail.com>

PR #732, opened 2019-07-01
zdtm: make grep_errors also grep warnings

Snorch:
It is inspired by the discussion about inotify fix:
https://github.com/checkpoint-restore/criu/pull/728#issuecomment-506929427

From one point of view, warnings might be important to understand why we
detect some visible change in the environment after c/r-ing the process,
and if this change is expected or not. So we should add "Warn" messages
to the output.

From over point, these warnings if they are expected, can spoil our
final logs with a lot of unnecessary details.

I see warnings from dedup when it tries to find region in parent hit the log limits:
```
From: https://api.travis-ci.org/v3/job/552665729/log.txt
The job exceeded the maximum log length, and has been terminated.
```

### DIFF
--- a/criu/image.c
+++ b/criu/image.c
@@ -126,6 +126,18 @@ int inventory_save_uptime(InventoryEntry *he)
 	return 0;
 }
 
+/*
+ * This function is intended to get an inventory image from previous (parent)
+ * dump iteration. We use dump_uptime from the image in detect_pid_reuse().
+ *
+ * You see that these function never fails by itself, it only prints warnings
+ * to better understand reasons why we don't found a proper image, failing here
+ * is too early. We get to detect_pid_reuse() only if we have a parent pagemap
+ * and that's the proper place to fail: we know that there is a parent pagemap
+ * but we don't have (can't access, etc) parent inventory => can't detect
+ * pid-reuse => fail.
+ */
+
 InventoryEntry *get_parent_inventory(void)
 {
 	struct cr_img *img;
@@ -134,7 +146,15 @@ InventoryEntry *get_parent_inventory(void)
 
 	dir = openat(get_service_fd(IMG_FD_OFF), CR_PARENT_LINK, O_RDONLY);
 	if (dir == -1) {
-		pr_warn("Failed to open parent directory\n");
+		/*
+		 * We print the warning below to be notified that we had some
+		 * unexpected problem on open. For instance we have a parent
+		 * directory but have no access. Having no parent inventory
+		 * when also having no parent directory is an expected case of
+		 * first dump iteration.
+		 */
+		if (errno != ENOENT)
+			pr_warn("Failed to open parent directory\n");
 		return NULL;
 	}
 
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -87,11 +87,15 @@ int dedup_one_iovec(struct page_read *pr, unsigned long off, unsigned long len)
 
 		ret = pr->seek_pagemap(pr, off);
 		if (ret == 0) {
-			pr_warn("Missing %lx in parent pagemap\n", off);
-			if (off < pr->cvaddr && pr->cvaddr < iov_end)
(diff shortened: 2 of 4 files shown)


avagin:
```
(00.250989) Warn  (criu/pagemap.c:90): Missing 7f84103e3000 in parent pagemap
(00.250999) 	p 0x7f84103f5000 [1]
(00.251002) Warn  (criu/pagemap.c:90): Missing 7f84103f5000 in parent pagemap
(00.251010) 	p 0x7f841041f000 [1]
(00.251012) Warn  (criu/pagemap.c:90): Missing 7f841041f000 in parent pagemap
(00.251020) 	p 0x7f8410457000 [1]
(00.251031) Warn  (criu/pagemap.c:90): Missing 7f8410457000 in parent pagemap
(00.251040) 	p 0x7f8410463000 [1]


The job exceeded the maximum log length, and has been terminated.
```

Travis jobs fail because we print too many warning messages. I suggest to convert them to debug messages or introduce warn_once.

Cc: @rppt

Snorch:
Add two more patches with small log improvements, so that logs don't overflow with:

Warn  (criu/image.c:137): Failed to open parent directory
Warn  (criu/pagemap.c:90): Missing 7f84103e3000 in parent pagemap

criu/image.c
```diff
@@ -134,7 +134,8 @@ InventoryEntry *get_parent_inventory(void)
 
 	dir = openat(get_service_fd(IMG_FD_OFF), CR_PARENT_LINK, O_RDONLY);
 	if (dir == -1) {
-		pr_warn("Failed to open parent directory\n");
+		if (errno != ENOENT)
```
avagin: pls add a comment why this can happen

criu/pagemap.c
```diff
@@ -87,11 +87,13 @@ int dedup_one_iovec(struct page_read *pr, unsigned long off, unsigned long len)
 
 		ret = pr->seek_pagemap(pr, off);
 		if (ret == 0) {
-			pr_warn("Missing %lx in parent pagemap\n", off);
-			if (off < pr->cvaddr && pr->cvaddr < iov_end)
+			if (off < pr->cvaddr && pr->cvaddr < iov_end) {
+				pr_debug("No range %lx-%lx in pagemap\n", off, pr->cvaddr);
```
avagin: PARENT pagemap?

Snorch: Actually we seek pagemap in pr which is current page_read and given in arguments, I think that old message is wrong about "parent" so I removed this.

criu/image.c
Snorch: done.

criu/image.c
```diff
@@ -134,7 +134,12 @@ InventoryEntry *get_parent_inventory(void)
 
 	dir = openat(get_service_fd(IMG_FD_OFF), CR_PARENT_LINK, O_RDONLY);
 	if (dir == -1) {
-		pr_warn("Failed to open parent directory\n");
+		/*
+		 * If we have no parent directory (no previous dump) then we
+		 * don't have parent inventory image too, it's expected.
+		 */
+		if (errno != ENOENT)
+			pr_warn("Failed to open parent directory\n");
```
avagin: you have to report an error if errno isn't ENOENT, don't you?

Snorch: We print the warning "Failed to open parent directory" from get_parerent_inventory() to be notified that we had some unexpected problem on open, for instance we have a parent directory but have no access. But failing here is too early, we get to detect_pid_reuse() only if we have a parent pagemap, that's the proper place to fail: we know that there is a parent pagemap but we can't access the parent inventory => can't detect pid-reuse => fail.

Snorch:
Updated outdated message in detect_pid_reuse while on it.

criu/image.c
avagin: This sounds reasonable. Could you add a comment with this explanation?

Snorch: Sure, done. Moved the general part before function.

PR #734, opened 2019-07-05
Add support for migrating CHILD_SUBREAPER prctl

mclapinski:
1. Checkpoint it via parasite.
2. Restore it before forking.
3. Add ZDTM test.

### DIFF
--- a/criu/include/restorer.h
+++ b/criu/include/restorer.h
@@ -217,6 +217,7 @@ struct task_restore_args {
 	unsigned			page_size;
 #endif
 	int				lsm_type;
+	int				child_subreaper;
 } __aligned(64);
 
 /*
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -52,6 +52,10 @@
 #define PR_SET_PDEATHSIG 1
 #endif
 
+#ifndef PR_SET_CHILD_SUBREAPER
+#define PR_SET_CHILD_SUBREAPER 36
+#endif
+
 #ifndef FALLOC_FL_KEEP_SIZE
 #define FALLOC_FL_KEEP_SIZE     0x01
 #endif
@@ -1231,6 +1235,14 @@ static bool vdso_needs_parking(struct task_restore_args *args)
 	return !vdso_unmapped(args);
 }
 
+static inline int restore_child_subreaper(int child_subreaper)
+{
+	if (child_subreaper)
+		return sys_prctl(PR_SET_CHILD_SUBREAPER, child_subreaper, 0, 0, 0);
+	else
+		return 0;
+}
+
 /*
  * The main routine to restore task via sigreturn.
  * This one is very special, we never return there
@@ -1731,6 +1743,7 @@ long __export_restore_task(struct task_restore_args *args)
 			    args->lsm_type);
 	ret = ret || restore_dumpable_flag(&args->mm);
 	ret = ret || restore_pdeath_sig(args->t);
+	ret = ret || restore_child_subreaper(args->child_subreaper);
 
 	futex_set_and_wake(&thread_inprogress, args->nr_threads);
 
--- a/test/zdtm/static/child_subreaper.c
+++ b/test/zdtm/static/child_subreaper.c
@@ -1,0 +1,36 @@
+#include <stdlib.h>
+#include <sys/prctl.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc	= "Check that child subreaper attribute is restored";
+const char *test_author	= "Michał Cłapiński <mclapinski@google.com>";
+
+int main(int argc, char **argv)
+{
+	test_init(argc, argv);
+
+	int cs_before = 1;
+	int ret = prctl(PR_SET_CHILD_SUBREAPER, cs_before, 0, 0, 0);
+	if (ret) {
+		pr_perror("Can't set child subreaper attribute, err = %d", ret);
+		exit(1);
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	int cs_after;
+	ret = prctl(PR_GET_CHILD_SUBREAPER, (unsigned long)&cs_after, 0, 0, 0);
+	if (ret) {
+		pr_perror("Can't get child subreaper attribute, err = %d", ret);
+		exit(1);
+	}
+
+	if (cs_before != cs_after)
+		fail("%d != %d\n", cs_before, cs_after);
+	else
+		pass();
+
+	return 0;
+}
--- a/test/zdtm/static/child_subreaper_and_reparent.c
+++ b/test/zdtm/static/child_subreaper_and_reparent.c
@@ -1,0 +1,142 @@
+#include <unistd.h>
+#include <stdlib.h>
+#include <signal.h>
+#include <sys/mman.h>
+#include <sys/wait.h>
+#include <sys/prctl.h>
+
+#include "zdtmtst.h"
+#include "lock.h"
+
+const char *test_doc	= "Check that child subreaper does not affect reparenting";
+const char *test_author	= "Pavel Tikhomirov <ptikhomirov@virtuozzo.com>";
+
+enum {
+	TEST_FORK,
+	TEST_SAVE,
+	TEST_CRIU,
+	TEST_CHECK,
+	TEST_EXIT,
+};
+
+struct shared {
+	futex_t fstate;
+	int parent_before_cr;
+	int parent_after_cr;
+} *sh;
+
+int orphan()
+{
+	/*
+	 * Wait until reparented to the pidns init. (By waiting
+	 * for the SUBREAPER to reap our parent.)
+	 */
+	futex_wait_until(&sh->fstate, TEST_SAVE);
+
+	sh->parent_before_cr = getppid();
+
+	/* Return the control back to MAIN worker to do C/R */
+	futex_set_and_wake(&sh->fstate, TEST_CRIU);
+	futex_wait_until(&sh->fstate, TEST_CHECK);
+
+	sh->parent_after_cr = getppid();
+
+	futex_set_and_wake(&sh->fstate, TEST_EXIT);
+	return 0;
+}
+
+int helper()
+{
+	int pid;
+
+	pid = fork();
+	if (pid < 0) {
+		pr_perror("Failed to fork");
+		return 1;
+	} else if (pid == 0) {
+		exit(orphan());
+	}
+	return 0;
+}
+
+int subreaper()
+{
+	int pid, ret, status;
+
+	setsid();
+
+	pid = fork();
+	if (pid < 0) {
+		pr_perror("Failed to fork");
+		return 1;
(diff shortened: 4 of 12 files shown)


rst0git:
Hi @mclapinski, would it be possible to add a ZDTM test for this feature to make sure it will not break with some changes in the future?

mclapinski:
Hi @rst0git, done.

0x7f454c46:
> 2. Restore it before forking.

Not sure about that.. I believe we set SIGCHILD handler in criu to catch bugs or unexpected deaths of children. Look at `sigchld_handler()` in `criu/cr-restore.c`. The main part (except reporting) is `futex_abort_and_wake()`, which will stop restore and prevent criu from hanging in such cases.

So, we either need to add SIGCHILD handling to the sub-reaper being restored (which is problematic, as it might be a process, that wasn't C/R'ed), or restore the subreaper much later - even so it will add a prctl-per-thread.

Cc'ing @Snorch as he has eaten a bunch of children, subreapers and all those fantastic animals at some point.

mclapinski:
@0x7f454c46 
Restoring child subreaper attribute after forking will work only on kernels containing this [fix](https://github.com/torvalds/linux/commit/749860ce242798fb090557a5a7868dee40af9268) so starting from v4.11. For older kernels restoring this attribute after forking will have no effect on children this process already has.

mclapinski:
@0x7f454c46 
Sorry but I'm new here. I don't understand what you mean by `sub-reaper being restored (which is problematic, as it might be a process, that wasn't C/R'ed)`. To me this PR is about C/Ring the subreaper process and its children, so I don't really understand how it could not have been C/R'ed.

0x7f454c46:
@mclapinski one can `criu dump -t <pid>` which has some foreign subreaper already

mclapinski:
> @mclapinski one can `criu dump -t <pid>` which has some foreign subreaper already

Sure, but you said `sub-reaper being restored`, so I got lost. Is the subreaper process being C/R'ed or is it foreign and not being C/R'ed?

mclapinski:
I see it like this:
1. We are C/Ring a process that is the subreaper, so running `PR_GET_CHILD_SUBREAPER` will return `1`. In that case we have to restore that property before forking due to the bug in kernels older than 4.11
2. We are C/Ring a process that isn't the subreaper but one of it ancestors is. In that case we don't care, because child subreaper property is assigned to that ancestor not to us, so there's nothing we should do.

0x7f454c46:
Ok, let me try again: I like (1) restoring at the forking stage, it looks neat and simple-designed.
But for (2): how will criu acknowledge that a child has unexpectedly died (killed by user/OOM/segfaulted)? As per my poor understanding, SIGCHLD will be delivered to subreaper and criu will just hang on futex waiting for nr_in_progress to cease.

mclapinski:
But what does that have to do with that PR? I'm changing the behaviour of C/R of process that is a child subreaper, not of a process that has a child subreaper among its ancestors. Behaviour in (2) doesn't change with this PR. In (2) parasite runs `PR_GET_CHILD_SUBREAPER`, it returns `0` (meaning we are not a subreaper) and we set it to `0` during restore.

0x7f454c46:
@mclapinski Oh, I see. Sorry, by some reason I got confused by what `prctl()` interface does.
Thanks for describing. Now I understand implementation a bit better and it looks quite good to me.

Though, I got another question: what will happen if during restore we get unexpected SIGCHLD from side-activity? (Like a task dies after being reparented to criu). Sounds like it will break restore, which might be just fine - but we probably should document it somewhere..

Snorch:
@mclapinski, I wrote a test to show you the problem, see here https://github.com/Snorch/criu/tree/child_subreaper_and_reparent

Before your "Add support for migrating CHILD_SUBREAPER prctl" patch there is no fail in these test, after the patch though we have:
```
[root@snorch criu]# ./test/zdtm.py run -f uns -t zdtm/static/child_subreaper_and_reparent
=== Run 1/1 ================ zdtm/static/child_subreaper_and_reparent
============= Run zdtm/static/child_subreaper_and_reparent in uns ==============
Start test
./child_subreaper_and_reparent --pidfile=child_subreaper_and_reparent.pid --outfile=child_subreaper_and_reparent.out
Run criu dump
Run criu restore
Send the 15 signal to  73
Wait for zdtm/static/child_subreaper_and_reparent(73) to die for 0.100000
###### Test zdtm/static/child_subreaper_and_reparent FAIL at result check ######
11:33:15.590:     4: FAIL: child_subreaper_and_reparent.c:67: Parent missmatch before 1 after 5 (errno = 10 (No child processes))

```
That is because you restore CHILD_SUBREAPER _before_ tasks are reparented and they get wrong parent. (see prepare_pstree_ids for reference)

 >Restoring child subreaper attribute after forking will work only on kernels containing this fix so starting from v4.11. For older kernels restoring this attribute after forking will have no effect on children this process already has.

I think we should not restore CHILD_SUBREAPER on kernels before fix at all, because we can't diferentiate children before the flag set and children after it. These way we are free to set the flag as late as possible to not affect process tree topology restore.

mclapinski:
@Snorch You are right about the problem, thanks for the test and explanation.

> I think we should not restore CHILD_SUBREAPER on kernels before fix at all, because we can't diferentiate children before the flag set and children after it. These way we are free to set the flag as late as possible to not affect process tree topology restore.

Ok, I can do that but in Travis we use Xenial, which I think means Ubuntu 16.04.0 therefore v4.4 kernel. If that's the case my test will fail since v4.4 doesn't have your patch to kernel. What do you think?

mclapinski:
> Ok, I can do that but in Travis we use Xenial, which I think means Ubuntu 16.04.0 therefore v4.4 kernel. If that's the case my test will fail since v4.4 doesn't have your patch to kernel. What do you think?

Oh, that's actually not true, since in my test I'm just checking if child subreaper attribute is set, not if it was properly set up for children. But if someone was to write such a test it would fail. Should I care?

Snorch:
@mclapinski Looks like travis kernel should have the fix:
[snorch@snorch ~]$ curl https://api.travis-ci.org/v3/job/555886995/log.txt 2>/dev/null | grep "Runtime kernel version"
Runtime kernel version: 4.15.0-1032-gcp

Not sure if our docker template can/do change kernel version shown inside, but I doubt it does.

You can add new kdat check to enable/disable CHILD_SUBREAPER logic. Either kernel version check (which is not 100% accurate) or create three processes SUBREAPER, HELPER and CHILD like I do in test, set the subreaper flag on the SUBREAPER, kill HELPER and check CHILD is or is not a child of SUBREAPER after these.

mclapinski:
> You can add new kdat check to enable/disable CHILD_SUBREAPER logic. Either kernel version check (which is not 100% accurate) or create three processes SUBREAPER, HELPER and CHILD like I do in test, set the subreaper flag on the SUBREAPER, kill HELPER and check CHILD is or is not a child of SUBREAPER after these.

That sounds pretty bad actually. Can't I just restore it after fork also for kernels without the fix? Before my change the behaviour was broken (one would lose child_subreaper property for both existing and new children) and after my change it will also be broken (one would lose child_subreaper property for existing children but keep it for new children). I don't think the situation after my change is worse than before my change.

Btw. I pushed new code that restores this property after forking. Please review it if you can.

test/zdtm/static/child_subreaper_existing_child.c
```diff
@@ -0,0 +1,135 @@
...
+
+	/* Cleanup: reap the SUBREAPER */
+	waitpid(pid, &status, 0);
+	if (!WIFEXITED(status) || WEXITSTATUS(status)) {
+		fail("Wrong exit status: %d", status);
+		return 1;
+	}
+
+	if (pid != sh->parent_after_cr)
+		fail("Parent missmatch before %d after %d", pid, sh->parent_after_cr);
```
Snorch: Confusing message here, you check if parent after orphan re-parent is expected (is a SUBREAPER). And maybe change naming for parent_after_cr too.

Snorch:
Except a small nit in test these looks good to me, after fixing you can add

Also you can add child_subreaper_and_reparent test to your series if you like (with similar rework you did in child_subreaper_existing_child to make it look better).

note: CR_STATE_RESTORE_SIGCHLD looks a proper place for a flag restore when all helpers and zombies already dead and all reparenting finished.

mclapinski:
@Snorch done

criu/pie/restorer.c
```diff
@@ -1711,6 +1723,10 @@ long __export_restore_task(struct task_restore_args *args)
 	if (ret)
 		goto core_restore_end;
 
+	ret = restore_child_subreaper(&args->child_subreaper);
```
Snorch: Sorry, my bad, I confused you with my previous message... what I meant in my note that the place you called restore_child_subreaper was absolutely right. Should've wrote ``` _after_ CR_STATE_RESTORE_SIGCHLD looks a proper place```

test/zdtm/static/child_subreaper.c
```diff
@@ -0,0 +1,36 @@
+#include <stdlib.h>
```
Snorch: And it looks like you accidentally merged zdtm test patch with criu patch. Please make them separate patches.

test/zdtm/static/child_subreaper_and_reparent.c
```diff
@@ -19,132 +19,120 @@ enum {
...
+	futex_wait_until(&sh->fstate, TEST_EXIT);
 	return 0;
 }
 
-int main(int argc, char **argv)
-{
-	void *ptr;
-
-	BUG_ON(sizeof(*fstate) + sizeof(*parent_before_cr) + sizeof(*parent_after_cr) > 4096);
+int main(int argc, char **argv) {
```
Snorch: We follow linux coding style. Please place beginning brace at the new line for functions.

Snorch:
And for child_subreaper_and_reparent, there is no need in two patches, merge your changes with the initial patch.

mclapinski:
@Snorch done

Snorch:
Looks good to me.

criu/include/restorer.h
```diff
@@ -50,6 +50,11 @@ struct restore_posix_timer {
 	int overrun;
 };
 
+struct restore_child_subreaper {
+	bool has_child_subreaper;
```
avagin: why do you need this has_child_subreaper? child_subreaper is PID, so it is always greater than 0. If child_subreaper is zero, child_subreaper isn't set. Hm?

mclapinski: `child_subreaper` is not PID.

avagin: Ok. It is even better:). child_subreaper can be 0 or 1. If it is 0 we need to do nothing. If it is one we need to call prctl(PR_SET_CHILD_SUBREAPER). I still don't understand why we need has_child_subreaper.

avagin:
Applied. Thanks!

mclapinski:
Thanks!

PR #739, opened 2019-07-10
make: Insert version macros in criu.h

### LINKED ISSUE #738: want CRIU_VERSION_NUMBER in header
I want like `CRIU_VERSION_NUMBER` in criu.h.

For example, in the case of https://github.com/checkpoint-restore/criu/commit/9ede9253f5d04e7ade626ece8b96bb66bf6dc938, binding like mruby-criu can be changed with backward compatibility.

rst0git:
Including the macros that provide information about the version of CRIU is required by projects that use libcriu to preserve backward compatibility.

Closes #738

### DIFF
--- a/lib/Makefile
+++ b/lib/Makefile
@@ -1,6 +1,6 @@
 CRIU_SO			:= libcriu.so
 CRIU_A			:= libcriu.a
-UAPI_HEADERS		:= lib/c/criu.h images/rpc.proto
+UAPI_HEADERS		:= lib/c/criu.h images/rpc.proto criu/include/version.h
 
 #
 # File to keep track of files installed by setup.py
--- a/lib/c/criu.h
+++ b/lib/c/criu.h
@@ -21,8 +21,10 @@
 
 #include <stdbool.h>
 
+#include "version.h"
+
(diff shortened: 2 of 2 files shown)


lib/Makefile
```diff
@@ -54,6 +54,7 @@ install: lib-c lib-a lib-py crit/crit lib/c/criu.pc.in
 	$(Q) install -m 755 lib/c/$(CRIU_A) $(DESTDIR)$(LIBDIR)/$(CRIU_A)
 	$(Q) mkdir -p $(DESTDIR)$(INCLUDEDIR)/criu/
 	$(Q) install -m 644 $(UAPI_HEADERS) $(DESTDIR)$(INCLUDEDIR)/criu/
+	$(Q) cat criu/include/version.h >> $(DESTDIR)$(INCLUDEDIR)/criu/criu.h
```
xemul: Why don't we explicitly add the

    #include "version.h"

line into criu.h and install the former one into criu/ subdir?

rst0git: Thank you for suggesting that, I updated the PR.

PR #743, opened 2019-07-17
Replace libprotobuf-c0-dev with libprotobuf-c-dev

thaJeztah:
The `libprotobuf-c0-dev` virtual package is no longer available
in Debian Buster, but is provided by `libprotobuf-c-dev`, which
is available.

https://packages.debian.org/stretch/libprotobuf-c0-dev

> Virtual Package: libprotobuf-c0-dev
>
> This is a virtual package. See the Debian policy for a definition of virtual packages.
>
> Packages providing libprotobuf-c0-dev
> libprotobuf-c-dev
> Protocol Buffers C static library and headers (protobuf-c)

### DIFF
--- a/contrib/debian/dev-packages.lst
+++ b/contrib/debian/dev-packages.lst
@@ -1,7 +1,7 @@
 # Required packages for development in Debian
 build-essential
 libprotobuf-dev
-libprotobuf-c0-dev
+libprotobuf-c-dev
 protobuf-c-compiler
 protobuf-compiler
 python-protobuf
--- a/criu/Makefile.packages
+++ b/criu/Makefile.packages
@@ -11,7 +11,7 @@ REQ-RPM-PKG-NAMES	+= $(PYTHON)-future
 REQ-RPM-PKG-TEST-NAMES  += libaio-devel
 
 REQ-DEB-PKG-NAMES	+= libprotobuf-dev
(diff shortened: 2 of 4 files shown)


thaJeztah:
Looks like CI completed, but didn't report back

thaJeztah:
ping @avagin @kolyshkin ptal 🤗

adrianreber:
So Debian 9 is Stretch and Debian 10 is Buster, right? Those are the two currently active releases, right? And both releases have `libprotobuf-c-dev`, right? But the Dockerfile and travis scripts are running on Ubuntu 16.04 (I think). Is the package on Ubuntu also named without the `0`? Looking at the travis result everything still seems to work and I also see it without `0` on 18.04. So :+1: from me.

thaJeztah:
oh, ahum, yes I came to this because I ran into the problem in https://github.com/moby/moby/pull/39549 and assumed you're using Debian here (😊😅) ; I'll double-check the Ubuntu packages, but they're very likely the same (😅)

avagin:
I don't seen any problem with Debian. Our s390 job uses the s390x/debian:latest base image:
https://travis-ci.org/checkpoint-restore/criu/jobs/559945752

avagin:
Applied, thanks!

thaJeztah:
Thanks!

PR #746, opened 2019-07-17
restorer: print an error if the prctl syscall failed

avagin:


avagin:
Cc: @mclapinski

criu/pie/restorer.c
```diff
@@ -328,10 +328,18 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
...
-		return sys_prctl(PR_SET_PDEATHSIG, ta->pdeath_sig, 0, 0, 0);
-	else
+	int ret;
+
+	if (!ta->pdeath_sig)
 		return 0;
+
+	ret = sys_prctl(PR_SET_PDEATHSIG, ta->pdeath_sig, 0, 0, 0);
+	if (ret) {
+		pr_err("Unable to set PR_SET_PDEATHSIG(%d): %d", ta->pdeath_sig, ret);
```
rst0git: please add `\n` at the end of the error message.

avagin: Fixed. Thanks!

images/core.proto
```diff
@@ -52,7 +52,7 @@ message task_core_entry {
 	//optional int32		tty_nr		= 16;
 	//optional int32		tty_pgrp	= 17;
 
-	optional int32			child_subreaper	= 18;
+	optional bool			child_subreaper	= 18;
```
mclapinski: I think it will work fine, but technically the documentation says:
```
PR_GET_CHILD_SUBREAPER (since Linux 3.4)
    Return the "child subreaper" setting of the caller, in the location pointed to by (int *) arg2.
```

criu/pie/restorer.c
```diff
@@ -328,10 +328,18 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
...
-	else
+	int ret;
+
+	if (!ta->pdeath_sig)
 		return 0;
+
+	ret = sys_prctl(PR_SET_PDEATHSIG, ta->pdeath_sig, 0, 0, 0);
+	if (ret) {
+		pr_err("Unable to set PR_SET_PDEATHSIG(%d): %d\n", ta->pdeath_sig, ret);
+		return -1;
```
mclapinski: You could remove this line and just `return ret` below. Same in the other function.

images/core.proto
rst0git: in [protobuf-c.h](https://github.com/racker/protobuf-c/blob/0dd05d11684fcd476088119ed3331ee7de0d321d/src/google/protobuf-c/protobuf-c.h#L88) bool is
```C
/** Boolean type. */
typedef int protobuf_c_boolean;
```
but using a boolean is better because it indicates that the value should be 0 or 1.

avagin: > I'm okay with bool, but I used int since the documentation says:

If you get a value which isn't 0 or 1, you have to fail, because you don't know what it means...

criu/pie/restorer.c
avagin: It will be correct too, but I like my way a bit more.

PR #749, opened 2019-07-29
mount: delay setting unbindable flag where possible to fix mount restore

Snorch:


criu/mount.c
```diff
@@ -1851,10 +1851,15 @@ static int restore_shared_options(struct mount_info *mi, bool private, bool shar
 			mi->mnt_id, mi->mountpoint, private, shared, slave);
 
 	if (mi->flags & MS_UNBINDABLE) {
-		if (shared || slave)
+		if (shared || slave) {
 			pr_warn("%s has both unbindable and sharing, ignoring unbindable\n", mi->mountpoint);
-		else
+		} else {
+			if (!mnt_is_overmounted(mi)) {
+				pr_debug("Temporary leave unbindable mount %s as private\n", mi->mountpoint);
```
avagin: need a comment which will explain why we need to do this. And are you sure that this will not break the restore of shared mounts?

criu/mount.c
```diff
@@ -2472,6 +2477,14 @@ static int do_close_one(struct mount_info *mi)
 	return 0;
 }
 
+static int set_unbindable(struct mount_info *mi)
+{
+	if (mi->flags & MS_UNBINDABLE && !mnt_is_overmounted(mi)) {
```
avagin: if ((mi->flags & MS_UNBINDABLE) && !mnt_is_overmounted(mi))

criu/mount.c
```diff
@@ -2472,6 +2477,14 @@ static int do_close_one(struct mount_info *mi)
...
+		return mount(NULL, mi->mountpoint, NULL, MS_UNBINDABLE, NULL);
```
avagin: we need to print an error message if mount fails. I see that we already have this problem in our code, so it would be good to fix them too.

avagin:
Pls describe in commit message why we need to delay setting unbindable flag

criu/mount.c
Snorch: I mentioned that you ask me to change such conditions for several times already. Do we have some code-style rule for it? Why operator precedence is not enough?

Snorch:
Comment, messages and parentheses done.

criu/mount.c
Snorch: > And are you sure that this will not break the restore of shared mounts?

Unbindable mounts don't have shared_id or master_id set, so when binding from it we will treat exactly the same as we do for a private mount, and I set it to be private, everything should be fine. We already have similar fix in vz7-criu for 2 months.

criu/mount.c
avagin: No, we don't. For me, this form looks more readable. I found that your form is widely used in the Linux kernel sources, so I think you can leave it as is.

Snorch: Lets be lazy and leave it as in latest push - with parentheses.

criu/mount.c
```diff
@@ -3193,6 +3218,7 @@ static int populate_mnt_ns(void)
 
 	ret = mnt_tree_for_each(pms, do_mount_one);
 	mnt_tree_for_each(pms, do_close_one);
+	mnt_tree_for_each(root_yard_mp, set_unbindable);
```
avagin: bzzzzz. I don't like this. First, we don't need to do this if ret isn't zero. Second, can we do this without enumerating all mounts?

We can try to resurrect this https://lkml.org/lkml/2017/1/23/712

Snorch: > bzzzzz. I don't like this. First, we don't need to do this if ret isn't zero. Second, can we do this without enumerating all mounts?

Done.

>We can try to resurrect this https://lkml.org/lkml/2017/1/23/712

I don't think it has any connection with unbindable flag, we can set these flag at any moment to any visible mount.

Offtop: I think, generally, that the kernel interface to add an existing mount to an existing shared group (or slave group) is a good thing. But I'm still not sure how to convince MS kernel guys to allow us this.

Snorch:
@avagin ping

criu/mount.c
Snorch: Ret check, special delayed_unbindable list to improve enumeration done.

Snorch:
Can it be merged? It is even needed for mounts-v2.

Snorch:
Thanks!

Issue #748, opened 2019-07-25
[Travis-CI, Fedora Rawhide] Error: Unable to find a match: python2-future


0x7f454c46:
```
Fedora - Modular Rawhide - Developmental packag  64 kB/s | 3.9 MB     01:01    
Fedora - Rawhide - Developmental packages for t  21 MB/s |  74 MB     00:03    
Last metadata expiration check: 0:00:01 ago on Thu Jul 25 19:32:11 2019.
Package python3-pip-19.1.1-1.fc31.noarch is already installed.
No match for argument: python2-future
Package sudo-1.8.27-1.fc31.x86_64 is already installed.
Package tar-2:1.32-1.fc31.x86_64 is already installed.
Error: Unable to find a match: python2-future
The command '/bin/sh -c dnf install -y 	ccache 	findutils 	gcc 	git 	gnutls-devel 	iproute 	iptables 	libaio-devel 	libasan 	libcap-devel 	libnet-devel 	libnl3-devel 	make 	procps-ng 	protobuf-c-devel 	protobuf-devel 	python2-protobuf 	python2 	python-ipaddress 	python-yaml 	python3-pip 	python2-future 	python3-PyYAML 	python3-future 	python3-protobuf 	python3-junit_xml 	sudo 	tar 	which 	e2fsprogs 	rubygem-asciidoctor 	kmod' returned a non-zero code: 1
Makefile:28: recipe for target 'fedora-rawhide' failed
make[1]: *** [fedora-rawhide] Error 1
make[1]: Leaving directory '/home/travis/build/0x7f454c46/criu/scripts/build'
Makefile:27: recipe for target 'fedora-rawhide' failed
make: *** [fedora-rawhide] Error 2
```

adrianreber:
Fedora is removing a lot of python2 packages, this is probably related. I can look into providing a fix.

0x7f454c46:
Thanks, @adrianreber

adrianreber:
Should be fixed in criu-dev.

PR #751, opened 2019-07-29
Remove python2-future from Fedora Dockerfiles

adrianreber:
This removes python2 from Fedora based Dockerfiles as it is not needed and will be removed from future Fedora releases. This PR also adds an environment variable to use other container runtimes for testing.


adrianreber:
Fix for #748

adrianreber:
Hmm, exactly fedora-rawhide is still failing. Need to check locally once more. I am pretty sure there were no errors locally.

Issue #346, opened 2017-06-29
Compel: use (rt)-vdso/vvar for timings in PIE blobs


0x7f454c46:
In PIE blobs (parasite/restorer) we use timings for logs - check `sbuf_log_init()`:
```
	/*
	 * Format:
	 *
	 * (time)pie: pid: string-itself
	 */
```
This is all done in that function with raw syscalls:
```
		sys_gettimeofday(&now, NULL);
		timediff(&start, &now);
```
If we omit calling raw syscalls and use vdso for the purpose of logging, that will eliminate as much as one-syscall-per-PIE-message. Getting time without switching to kernel will speed up C/R, keeping logs as informative as they were.

During restore (in the restorer blob) - we know the position of vvar/vdso pair and vdso-symtable.
So, it should be simple to use gettimeofday from vvar page, calling vdso-provided function.

On the dumping - it's a bit more complex. We have vdso-symtable parsed in criu, but yet don't know if seized task has the same vdso-symtable. So, as we'll parse the symtable afterward anyway for dumping needs, this could be done during parasite's initialization and then using gettimeofday from seized task.

PR #753, opened 2019-07-30
vdso: Use gettimeofday() from vdso for timings

avagin:
Historically, PIE uses syscalls for printing timings.
The idea was written a while ago here:
https://github.com/checkpoint-restore/criu/issues/346

While at it, I've found that we print timings only in restorer blob, but
not in parasite. It may be intentional or just missed
std_log_set_start() call in parasite_init_daemon().
Anyway, I've decided to keep it as-is for now.

Also, adding support for timings from vdso in ia32 turned to be more
complex (see a comment in the last patch), so I decided to start from
native applications.

I've measured the gain on zdtm tests:
https://gist.github.com/0x7f454c46/7147b519e49f9ad01706c27ab1334fa7
It seems that criu is really voluble on restoring mappings :)

Anyway, the gain is not as huge as I've expected (I thought we print
much more), but still removes 50-100 syscalls per most of zdtm tests.
And it doesn't bring much complexity as we already have all information
about vdso mappings and symbol tables in restorer.
So, remove syscalls for free and keep being the same informative.

Cc: Cyrill Gorcunov <gorcunov@gmail.com>

Dmitry Safonov (8):
  compel/log: Use enum as parameter for std_log_set_loglevel()
  compel/std/uapi: Provide setter for gettimeofday()
  vdso/restorer: Try best to preserve vdso during restore
  vdso/restorer: Always track vdso/vvar positions in vdso_maps_rt
  restorer/parasite-vdso: Don't move vvar if failed to move vdso
  seccomp/restorer: Disable gtod from vdso in strict mode
  vdso: Add compatible property to vdso_maps
  restorer: Use gettimeofday() from rt-vdso for log timings

 compel/plugins/include/uapi/std/log.h |  16 ++-
 compel/plugins/std/log.c              |  18 ++-
 criu/arch/aarch64/include/asm/vdso.h  |   1 +
 criu/arch/arm/include/asm/vdso.h      |   1 +
 criu/arch/ppc64/include/asm/vdso.h    |   1 +
 criu/arch/s390/include/asm/vdso.h     |   3 +-
 criu/arch/x86/include/asm/vdso.h      |   3 +-
 criu/include/parasite-vdso.h          |   4 +-
 criu/include/util-vdso.h              |   1 +
 criu/pie/parasite-vdso.c              | 164 +++++++++++++++-----------
 criu/pie/restorer.c                   | 115 ++++++++++++++----
 criu/vdso.c                           |   6 +-
 12 files changed, 227 insertions(+), 106 deletions(-)

### DIFF
--- a/compel/plugins/include/uapi/std/log.h
+++ b/compel/plugins/include/uapi/std/log.h
@@ -1,10 +1,12 @@
 #ifndef COMPEL_PLUGIN_STD_LOG_H__
 #define COMPEL_PLUGIN_STD_LOG_H__
 
+#include "compel/loglevels.h"
+
 #define STD_LOG_SIMPLE_CHUNK	256
 
 extern void std_log_set_fd(int fd);
-extern void std_log_set_loglevel(unsigned int level);
+extern void std_log_set_loglevel(enum __compel_log_levels level);
 extern void std_log_set_start(struct timeval *tv);
 extern int std_vprint_num(char *buf, int blen, int num, char **ps);
 extern void std_sprintf(char output[STD_LOG_SIMPLE_CHUNK], const char *format, ...)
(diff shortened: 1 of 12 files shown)


cyrillos:
I already reviewed it via email. Anyway
Reviewed-by: Cyrill Gorcunov <gorcunov@gmail.com>

avagin:
Applied, thanks

PR #755, opened 2019-07-31
scripts: Remove Python 2 dependency for Fedora tests

rst0git:
In https://github.com/checkpoint-restore/criu/pull/751 have been removed some of the Python 2 packages that are installed in Fedora Dockerfiles. There are a few necessary changes to allow the tests to work in environment with Python 3 only.

cc @adrianreber

### DIFF
--- a/lib/py/images/pb2dict.py
+++ b/lib/py/images/pb2dict.py
@@ -263,7 +263,7 @@ def get_bytes_dec(field):
 def is_string(value):
     # Python 3 compatibility
     if "basestring" in __builtins__:
-        string_types = basestring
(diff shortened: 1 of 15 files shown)


adrianreber:
Thanks!

avagin:
Why do tests in Travis pass without these changes?

rst0git:
Because by installing `python-ipaddress` and `python-yaml` we install Python 2 and executable files with `#!/usr/bin/env python2` shebang are able to run. With these changes Python 2 is no longer installed and we use only Python 3 to run the tests.

avagin:
I was thinking about something like this. Thanks you for clarifying.

Issue #747, opened 2019-07-22
Error: ipv4: FIB table does not exist.


avagin:
```
======================= Run zdtm/static/netns_sub in ns ========================
Start test
Test is SUID
./netns_sub --pidfile=netns_sub.pid --outfile=netns_sub.out
Run criu dump
=[log]=> dump/zdtm/static/netns_sub/42/1/dump.log
------------------------ grep Error ------------------------
(00.074841) 	Running ip route save
(00.076377) 	Running ip route save
(00.078832) 	Running ip -6 route save
(00.079549) 	Running ip route save
Error: ipv4: FIB table does not exist.
Dump terminated
(00.082025) Error (criu/util.c:632): exited, status=2
(00.082057) Error (criu/net.c:1735): IP tool failed on route save
(00.082399) Error (criu/namespaces.c:1090): Namespaces dumping finished with error 65280
(00.082501) Unlock network
(00.082506) Running network-unlock scripts
(00.084140) 	Running ip rule save
Error: ipv4: FIB table does not exist.
Dump terminated
(00.089039) Error (criu/util.c:632): exited, status=2
(00.089071) Error (criu/net.c:1735): IP tool failed on route save
(00.101841) 	Unseizing 45 into 1
(00.101854) 	Unseizing 46 into 1
(00.101865) 	Unseizing 48 into 1
(00.101876) 	Unseizing 47 into 1
(00.101923) Error (criu/cr-dump.c:1754): Dumping FAILED.
------------------------ ERROR OVER ------------------------
################# Test zdtm/static/netns_sub FAIL at CRIU dump #################
```

avagin:
https://ci.openvz.org/job/CRIU/job/CRIU-ppc64le/job/criu-dev/4546/consoleFull

adrianreber:
Already discussed in https://lists.openvz.org/pipermail/criu/2019-May/043923.html, unfortunately with no result.

https://www.spinics.net/lists/netdev/msg559904.html

adrianreber:
On 3.13 I can still see this error on current rawhide: `Error: ipv4: FIB table does not exist`

Snorch:
https://www.spinics.net/lists/netdev/msg559739.html

They say that iproute package is guilty, I have the same on iproute-5.2.0-1.fc30.x86_64.

Snorch:
Still not fixed:
```
[root@snorch snorch]# unshare -n
[...]

PR #758, opened 2019-08-08
test: bring the lo interface up in each network namespace

avagin:
This is needed to workaround the problem with "ip route save":
(00.113153) 	Running ip route save
Error: ipv4: FIB table does not exist.

Signed-off-by: Andrei Vagin <avagin@gmail.com>


avagin:
https://github.com/checkpoint-restore/criu/issues/747

PR #759, opened 2019-08-08
Add new command line option: --cgroup-yard

mclapinski:
Instead of creating cgroup yard in CRIU, now we can create it externally
and pass it to CRIU. Useful if somebody doesn't want to grant
CAP_SYS_ADMIN to CRIU.


avagin:
We need a test for this option.

Documentation/criu.txt
```diff
@@ -266,10 +266,28 @@ For example, the command line for the above example should look like this:
...
-    is not used, the predefined properties are merged with the provided ones.
+*--cgroup-yard* 'path'::
+    Instead of trying to mount cgroups in CRIU, provide a path to a directory
+    with already created cgroup yard. Useful if you don't want to grant
+    CAP_SYS_ADMIN to CRIU. For every cgroup mount there should be exactly one
+    directory. If there is only one controller in this mount, the dir's name
+    should be just the name of the controller. If there are multiple controllers
+    comounted, the directory name should have them be separated by a comma.
++
+For example, if */proc/cgroups* looks like this:
```
rst0git:
```diff
diff --git a/Documentation/criu.txt b/Documentation/criu.txt
index 4f9a4af5b..cd9d72930 100644
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -275,19 +275,24 @@ For example, the command line for the above example should look like this:
     comounted, the directory name should have them be separated by a comma.
 +
 For example, if */proc/cgroups* looks like this:
++
+----------
 #subsys_name hierarchy num_cgroups enabled
 cpu          1         1           1
 devices      2         2           1
 freezer      2         2           1
+----------
 +
 then you can create the cgroup yard by the following commands:
 +
+----------
 mkdir private_yard
 cd private_yard
 mkdir cpu
 mount -t cgroup -o cpu none cpu
 mkdir devices,freezer
 mount -t cgroup -o devices,freezer none devices,freezer
+----------
 
 *--tcp-established*::
     Checkpoint established TCP connections.
```

mclapinski:
I've added the test. Since it's my first time writing those hooks, any suggestions are appreciated.

test/zdtm/static/cgroup_yard.hook
```diff
@@ -0,0 +1,62 @@
...
+
+			if cgr.startswith("name="):
+				ctrl = cgr[len("name="):]
+				opts = "none," + cgr
+			else:
+				ctrl = cgr
+				opts = cgr
+
+			os.mkdir(yard + "/" + ctrl)
+			subprocess.call(["mount", "-t", "cgroup", "none", yard + "/" + ctrl, "-o", opts])
```
rst0git: What about using `subprocess.check_call()` to catch non-zero return code?

test/zdtm/static/cgroup_yard.hook
```diff
@@ -0,0 +1,62 @@
...
+	except:
+		pass
+
+	try:
+		os.rmdir(os.path.join(tname, "subcg00"))
+	except:
+		pass
+
+	subprocess.call(["umount", tname])
+	os.rmdir(tname)
```
rst0git: Can we replace `os.rmdir` with something like `shutil.rmtree(tname, ignore_errors=True)`?

test/zdtm/static/cgroup_yard.desc
```diff
@@ -0,0 +1 @@
+{'flavor': 'h', 'flags': 'suid', 'opts': '--manage-cgroups --cgroup-yard ../../../../../../external_yard'}
```
avagin: we need a comment which explains these ../../.../../..

avagin: can we just specify --cgroup-yard=/sys/fs/cgroup?

test/zdtm/static/cgroup_yard.hook
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
+
+import sys
+import os
+import subprocess
+import tempfile
+
+yard = "external_yard"
+
+if sys.argv[1] == "--pre-dump":
```
avagin: could you write comments for each actions which describe what they are doing

rst0git:
We should also update https://criu.org/CGroups

rst0git:
@mclapinski  Is the `--cgroup-yard` option intended to be used only with `criu dump`?

test/zdtm/static/cgroup_yard.desc
mclapinski: No, the special test cgroup `zdtmtst` is not mounted there.

test/zdtm/static/cgroup_yard.hook
mclapinski: I think we can but `shutil.rmtree` will try to remove the files in that directory first, fail on all of them and ignore that errors and then remove the directory. I'm not sure if it's the cleanest solution.

mclapinski:
> @mclapinski Is the `--cgroup-yard` option intended to be used only with `criu dump`?

No, it's intended to be used with both `dump` and `restore`.

rst0git:
>

> 
> 
> > @mclapinski Is the `--cgroup-yard` option intended to be used only with `criu dump`?
> 
> No, it's intended to be used with both `dump` and `restore`.

Thank you for clarifying this.

`Documentation/criu.txt` has separate section for both `dump` and `restore` options and currently `--cgroup-yard` is described only for `dump`. We should also add description for `restore`.

criu/cgroup.c
```diff
@@ -1652,20 +1675,28 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
...
-		goto err;
+		cg_yard = xstrdup(paux);
+		if (!cg_yard) {
+			rmdir(paux);
+			return -1;
+		}
+
+		if (make_yard(cg_yard))
+			goto err;
+	}
```
rst0git: a few lines below we do:
```C
/* Create controller if not yet present */
if (access(paux, F_OK)) {
	pr_debug("\tMaking controller dir %s (%s)\n", paux, opt);
	if (mkdir(paux, 0700)) {
		pr_perror("\tCan't make controller dir %s", paux);
		return -1;
	}
	if (mount("none", paux, "cgroup", 0, opt) < 0) {
		pr_perror("\tCan't mount controller dir %s", paux);
		return -1;
	}
}
```
should we skip this step when `opts.cgroup_yard` is used?

mclapinski: If external yard is passed then `access` will return 0 and we don't run the code inside this `if`. I could add another `if` here checking if external yard was passed but I don't really see the reason since nothing would change.

test/zdtm/static/cgroup_yard.hook
avagin: if os.accesss(path, os.F_OK):
  os.rmdir(path)

PR #761, opened 2019-08-10
kerndat: mark functions as static which are used in kerndat.c only

avagin:


criu/kerndat.c
```diff
@@ -157,70 +157,28 @@ static void kerndat_mmap_min_addr(void)
 		 (unsigned long)kdat.mmap_min_addr);
 }
 
-static int kerndat_files_stat(bool early)
+static int kerndat_files_stat()
```
rppt: nit: static int kerndat_files_stat(**void**)

PR #771, opened 2019-08-26
Introduce new command line option: --cgroup-dump-only-below-criu

mclapinski:
Currently CRIU will not migrate empty cgroups below the task's cgroup if CRIU runs in the same cgroup as the task ([proof](https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/cgroup.c#L905)). This flag overrides that behaviour and causes CRIU to dump everything below the CRIU's cgroup.

This flag fixes another issue. Let's imagine we created cgroup A to contain some task. But the task created 2 sub-cgroups: A1 and A2. Then moved itself to A1. We start migration and put CRIU in cgroup A. In that situation only sub-cgroup A1 would be migrated since A2 doesn't contain any task and no task runs directly in A. But the expected behaviour is to migrate everything below cgroup A since everything there belongs to the task.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -439,10 +439,13 @@ static int add_cgroup_properties(const char *fpath, struct cgroup_dir *ncd,
 	return 0;
 }
 
-static int add_cgroup(const char *fpath, const struct stat *sb, int typeflag)
+static int add_cgroup(const char *fpath, const struct stat *sb, int typeflag, struct FTW *ftwbuf)
 {
 	struct cgroup_dir *ncd = NULL, *match;
 	int exit_code = -1;
+
+	if (opts.cgroup_dump_only_below_criu && ftwbuf->level == 0)
+		return 0;
 
 	if (typeflag == FTW_D) {
 		int mtype;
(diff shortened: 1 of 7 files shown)


avagin:
Thank you for contribution. I'm on vacation right now, will review this and other PR-s after the Linux Plumbers.

criu/cgroup.c
```diff
@@ -439,11 +439,14 @@ static int add_cgroup_properties(const char *fpath, struct cgroup_dir *ncd,
 	return 0;
 }
 
-static int add_cgroup(const char *fpath, const struct stat *sb, int typeflag)
+static int add_cgroup(const char *fpath, const struct stat *sb, int typeflag, struct FTW *ftwbuf)
 {
 	struct cgroup_dir *ncd = NULL, *match;
 	int exit_code = -1;
 
+	if (opts.cgroup_dump_only_below_criu && ftwbuf->level == 0)
```
avagin: could you add a comment about what this means?

criu/config.c
```diff
@@ -517,6 +517,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{ "tls-key",			required_argument,	0, 1095},
 		BOOL_OPT("tls", &opts.tls),
 		{"tls-no-cn-verify",		no_argument,		&opts.tls_no_cn_verify, true},
+		BOOL_OPT("cgroup-dump-only-below-criu", &opts.cgroup_dump_only_below_criu),
```
avagin: you need to add a description for this option into Documentation/criu.txt and criu/crtools.c

avagin:
Maybe it is better to specify a cgroup path bellow which cgroups have to be dumped?

And we need to think about how to test this new option. You can look at other cgroups tests in test/zdtm/static/.

Cc: @cyrillos

cyrillos:
> Maybe it is better to specify a cgroup path bellow which cgroups have to be dumped?

Looks a way better from architecture pov I think.

PR #783, opened 2019-09-13
Couple of fixes to build and run libcriu tests

ashu-mehra:
libcriu tests are currently broken. This patch fixes couple of
issues to allow the building and running libcriu tests.

1. lib/c/criu.h got updated to include version.h which is present
at "criu/include", but the command to compile libcriu tests is not
specifying "criu/include" in the path to be searched for header
files. This resulted in compilation error.
This can be fixed by adding "-I ../../../../../criu/criu/include"
however it causes more problems as "criu/include/fcntl.h" would now
hide system defined fcntl.h
Solution is to use "-iquote ../../../../../criu/criu/include"
which applies only to the quote form of include directive.

2. Secondly, libcriu.so major version got updated to 2 but
libcriu/run.sh still assumes verion 1. Instead of just updating the
version in libcriu/run.sh to 2, this patch updates the libcriu/Makefile
to use "CRIU_SO_VERSION_MAJOR" so that future changes to major version
of libcriu won't cause same problem again.

Signed-off-by: Ashutosh Mehra <asmehra1@in.ibm.com>

### DIFF
--- a/scripts/travis/travis-tests
+++ b/scripts/travis/travis-tests
@@ -161,6 +161,9 @@ ip net add test
 ./test/zdtm.py run -t zdtm/static/env00 -k always
 ./test/crit-recode.py
 
+# libcriu testing
+make -C test/others/libcriu run
+
 make -C test/others/shell-job
 
 if ! [ -x "$(command -v flake8)" ]; then
--- a/test/others/libcriu/Makefile
+++ b/test/others/libcriu/Makefile
@@ -1,3 +1,5 @@
+include ../../../../criu/Makefile.versions
+
 TESTS += test_sub 
 TESTS += test_self
 TESTS += test_notify
@@ -19,8 +21,16 @@ endef
 $(foreach t, $(TESTS), $(eval $(call genb, $(t))))
 
 %.o: %.c
-	gcc -c $^ -I../../../../criu/lib/c/ -I../../../../criu/images/ -o $@ -Werror
+	gcc -c $^ -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o $@ -Werror
 
-clean:
+clean: libcriu_clean
 	rm -rf $(TESTS) $(TESTS:%=%.o) lib.o
 .PHONY: clean
+
+libcriu_clean:
+	rm -f libcriu.so.${CRIU_SO_VERSION_MAJOR}
+.PHONY: libcriu_clean
+
+libcriu:
+	ln -s ../../../../criu/lib/c/libcriu.so libcriu.so.${CRIU_SO_VERSION_MAJOR}
+.PHONY: libcriu
--- a/test/others/libcriu/run.sh
+++ b/test/others/libcriu/run.sh
@@ -5,14 +5,13 @@ source ../env.sh || exit 1
 
 echo "== Clean"
 make clean
+make libcriu
 rm -rf wdir
-rm -f ./libcriu.so.1
 
 echo "== Prepare"
 mkdir -p wdir/i/
 
 echo "== Run tests"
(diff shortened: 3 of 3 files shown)


adrianreber:
Would be good to include this test in the travis runs via `scripts/travis/travis-tests` now that it is fixed. Besides that the changes look reasonable.

adrianreber:
@ashu-mehra Thanks for the travis integration. If this does not break travis testing I :+1:

ashu-mehra:
I see some failures in my personal build. Checking them.

scripts/travis/travis-tests
```diff
@@ -161,6 +161,9 @@ ip net add test
 ./test/zdtm.py run -t zdtm/static/env00 -k always
 ./test/crit-recode.py
 
+# libcriu testing
+./test/others/libcriu/run.sh
+
```
rst0git: There are a few tests listed from the `other` folder above. You can add something like:
```
make -C test/others/libcriu/ run
```

ashu-mehra: Right, that should prevent the errors I am seeing. Thanks for the suggestion.

ashu-mehra:
`test_sub` and `test_iters` have failed on some archs. Are they known to fail?

rst0git:
> Are they known to fail?

I can't reproduce the error outside Travis.
I think it will be good to add the following change to make it easier to understand why the test failed:
```diff
--- a/test/others/libcriu/run.sh
+++ b/test/others/libcriu/run.sh
@@ -27,6 +27,7 @@ function run_test {
                mkdir wdir/i/$1/
                if ! setsid ./$1 ${CRIU} wdir/i/$1/ < /dev/null &>> wdir/i/$1/test.log; then
                        echo "$1: FAIL"
+                       cat wdir/i/$1/*.log
                        RESULT=1
                fi
        fi

```

rst0git:
The tests failed with `Can't exec criu swrk: No such file or directory` because the `criu` binary was not found in `PATH`. The following should fix it:
```diff
--- a/test/others/libcriu/run.sh
+++ b/test/others/libcriu/run.sh
@@ -13,7 +13,7 @@ mkdir -p wdir/i/
 
 echo "== Run tests"
 export LD_LIBRARY_PATH=.
-export PATH="`dirname ${BASH_SOURCE[0]}`/../../:$PATH"
+export PATH="`dirname ${BASH_SOURCE[0]}`/../../../criu/:$PATH"
 
 RESULT=0
```

ashu-mehra:
@rst0git I tried your fix and it worked. But I was not certain why only two tests `test_sub` and `test_iters`  are failing. On closer look I realized these two tests are missing a call to `criu_set_service_binary()` before the restore operation.
That's the reason dump always succeeds but restore fails.
I have also updated `run.sh` to display contents of log files on failure. 
With these changes my personal build is passing now.

rst0git:
@ashu-mehra I also noticed that `criu_set_service_binary()` was missing and I think that this is on purpose. It is testing the [default (initialisation) values](https://github.com/checkpoint-restore/criu/blob/3a96d3713de37c96e8beb604add383ea671d57a7/lib/c/criu.c#L262L263).

```C
	opts->service_comm	= CRIU_COMM_BIN;
	opts->service_binary	= strdup(CR_DEFAULT_SERVICE_BIN);
```
However, the relative path to `criu` was incorrect since 88aaae3 and db6cb69.

ashu-mehra:
@rst0git If thats the case then your suggestion of fixing the PATH would be the way to go. I will update my last commit based on your suggestion.

ashu-mehra:
I think this looks ready now.

avagin:
@adrianreber @rst0git could you review this pr?

rst0git:
LGTM

avagin:
The third commit should be before the second one. Travis tests should pass after each commit.

PR #784, opened 2019-09-14
test/other: Remove 'criu exec' tests

rst0git:
The 'criu exec' functionality has been deprecated and it is no longer supported.


avagin:
It was deprecated, because @xemul and @cyrillos promised to implement this feature in compel. Where is "compel exec"?  ;)

cyrillos:
I think it won't be implemented for a while :) Maybe someday in vague future.

PR #785, opened 2019-09-14
image-desc: Remove CR_FD_FILE_LOCKS_PID

rst0git:
The support for per-pid images with locks has been dropped with commit d040219 and `CR_FD_FILE_LOCKS_PID` is not used.


avagin:
Should we drop FILE_LOCKS_MAGIC too?

rst0git:
No, FILE_LOCKS_MAGIC is used by `filelocks.img`

PR #787, opened 2019-09-15
GSoC 19: Optimizing the Pre-dump Algorithm

criupatchwork:
This patch series implements optimization of pre-dumping algorithm
as part of GSoC 2019 project.

In current pre-dumping, the target process needs to be frozen
till all the memory pages are drained into pipes. Then the target
process gets unfrozen and pages collected into pipes are written
into image files at the end of pre-dump. This approach has two
problems. First, target process remains frozen for longer duration.
Second, pipes induce memory pressure in the system. If memory
utilization during pre-dump is nearly equal to system's memory,
then this risks running into out-of-memory failures as the pipe
pages are not reclaimable.

The new implementation of pre-dump solves above mentioned two
issues. In this, the target process is frozen untill memory
mappings are collected. Then the process will unfreeze and
continue. The pages of target process are copied while process
is running. We use process_vm_readv syscall to copy pages from
process memory to user-space buffer, by using collected mappings.
Since page copying and process execution are happening
simultaneously, there is possibility that process might have
modified some mappings after they have been collected. This
results in race over memory mappings.

This patch series handles the race over mappings, utilizes
user-space buffer through process_vm_readv syscall to copy pages
and reduces the frozen time for target process by allowing it to
run as soon as memory mappings are collected. We call this new
approach of pre-dumping as "read mode pre-dump". New CLI option
--pre-dump-mode is added for it, which takes "splice" or "read"
as input. "splice" mode is traditional parasite way of pre-dumping
and is set by default.

Evaluation of new approach:
---------------------------

Performance:
------------
$ ./test/zdtm.py run --pre 5 -t zdtm/static/maps04 -f h

Splice mode -
pre-dump: 0.59
pre-dump: 0.06
pre-dump: 0.07
pre-dump: 0.06
pre-dump: 0.06
dump: 0.12

$ ./test/zdtm.py run --pre 5 -t zdtm/static/maps04 --pre-dump-mode=read -f h

Read mode -
pre-dump: 0.66
pre-dump: 0.06
pre-dump: 0.06
pre-dump: 0.06
pre-dump: 0.06
dump: 0.12

Average drop : ~ 7%
Maximum drop : ~ 13%

Freeze time:
------------
$ ./test/zdtm.py run --pre 5 -t zdtm/static/maps04 --show-stats -f h

Splice mode -
pre-dump: 122235
pre-dump: 77514
pre-dump: 77912
pre-dump: 74940
pre-dump: 71977
dump: 131208

$ ./test/zdtm.py run --pre 5 -t zdtm/static/maps04 --pre-dump-mode=read --show-stats -f h

Read mode -
pre-dump: 82365
pre-dump: 67521
pre-dump: 68385
pre-dump: 68706
pre-dump: 67144
dump: 160998

Average reduction : ~18%
Maximum reduction : ~35%

Abhishek Dubey (7):
  Adding --pre-dump-mode option
  Skip generating iov for non-PROT_READ memory
  Skip adding PROT_READ to non-PROT_READ mappings
  Adding cnt_sub for stats manipulation
  Handle vmsplice failure for read mode pre-dump
  read mode pre-dump implementation
  Refactor time accounting macros

 Documentation/criu.txt    |   6 +
 criu/config.c             |  10 ++
 criu/cr-dump.c            |   9 +-
 criu/crtools.c            |   2 +
 criu/include/cr_options.h |   7 +
 criu/include/page-xfer.h  |   4 +
 criu/include/stats.h      |   1 +
 criu/mem.c                |  88 +++++++++--
 criu/page-pipe.c          |  10 ++
 criu/page-xfer.c          | 382 ++++++++++++++++++++++++++++++++++++++++++++++
 criu/stats.c              |  12 ++
 test/zdtm.py              |   9 +-
 12 files changed, 524 insertions(+), 16 deletions(-)

### DIFF
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -156,6 +156,12 @@ In addition, *page-server* options may be specified.
     Turn on memory changes tracker in the kernel. If the option is
     not passed the memory tracker get turned on implicitly.
 
+*--pre-dump-mode*='mode'::
+    There are two 'mode' to operate pre-dump algorithm. The 'splice' mode
+    is parasite based, whereas 'read' mode is based on process_vm_readv
+    syscall. The 'read' mode incurs reduced frozen time and reduced
+    memory pressure as compared to 'splice' mode. Default is 'splice' mode.
+
 *dump*
 ~~~~~~
 Performs a checkpoint procedure.
--- a/criu/config.c
+++ b/criu/config.c
@@ -276,6 +276,7 @@ void init_opts(void)
 	opts.empty_ns = 0;
 	opts.status_fd = -1;
 	opts.log_level = DEFAULT_LOGLEVEL;
+	opts.pre_dump_mode = PRE_DUMP_SPLICE;
 }
 
 bool deprecated_ok(char *what)
@@ -517,6 +518,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{ "tls-key",			required_argument,	0, 1095},
 		BOOL_OPT("tls", &opts.tls),
 		{"tls-no-cn-verify",		no_argument,		&opts.tls_no_cn_verify, true},
+		{ "pre-dump-mode",		required_argument,	0, 1097},
 		{ },
 	};
 
@@ -814,6 +816,14 @@ int parse_options(int argc, char **argv, bool *usage_error,
 			break;
 		case 1095:
 			SET_CHAR_OPTS(tls_key, optarg);
+			break;
+		case 1097:
+			if (!strcmp("read", optarg)) {
+				opts.pre_dump_mode = PRE_DUMP_READ;
+			} else if (strcmp("splice", optarg)) {
+				pr_err("Unable to parse value of --pre-dump-mode\n");
+				return 1;
+			}
 			break;
 		case 'V':
 			pr_msg("Version: %s\n", CRIU_VERSION);
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -1515,7 +1515,8 @@ static int cr_pre_dump_finish(int status)
 		mem_pp = dmpi(item)->mem_pp;
 
 		if (opts.pre_dump_mode == PRE_DUMP_READ)
-			ret = 0;  /* Call optimized pre-dump here */
+			ret = page_xfer_predump_pages(item->pid->real,
+                                                      &xfer, mem_pp);
 		else
 			ret = page_xfer_dump_pages(&xfer, mem_pp);
 
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -1514,11 +1514,13 @@ static int cr_pre_dump_finish(int status)
 
 		mem_pp = dmpi(item)->mem_pp;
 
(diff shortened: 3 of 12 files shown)


avagin:
@Snorch we read pagemap, create a list of all pages that have to be dumped then clear soft-dirty bits, unfreeze processes and start dumping pages. Where is a race here?

Snorch:
@avagin Thanks for explanation! Now I see that it is enough to lock only "soft-dirty clear + pagemap-read" to have consistent pre-dumps.

Another problem comes in mind, again - may be I'm missing something. What if: we dumped pagemap of the process and released it, now we are dumping it's pages. Then process dies and another process reuses it's pid? Syscall process_vm_readv works on pid number, so it won't fill any difference and we will continue reading pagemap. Assume we are "lucky" and the new process has exactly the same page offsets in it's pagemap. We will produce a predump filled with inconsistent combination of two processes. Now more, the second process had all pages with cleaned soft dirty, because it was forked from other process which pagemap was just read and bits were cleared. These means that on final dump we will use pages from the predump.

The "lucky" process may be an attacker which tries to steel memory form the first process.

0x7f454c46:
@Snorch Yeah, that's a good concern.

Probably, we could check at least the creation time for a process to verify that's the same one (as we do between pre-dumps now..)
I wish we could use pidfd instead though

avagin:
> These means that on final dump we will use pages from the inconsistent predump.

We will not. On the final dump, we will detect that this process has died. I am agree that we have an issue here. A user can request to dump their processes tree and then they will receive images with some memory of other processes.

We can use pidfd and check that a target process is still alive after dumping all pages. If it is not, we can remove all its images.

Snorch:
Yes, we can fix it either with:
1) holding open pidfd on the process to check if it is still alive and memory is valid;
or
2) checking that process creation time is strictly less than the time of pagemap reed (under freeze "lock") for a (pre-)dump, which makes these (pre-)dump valid.

avagin:
@Snorch could you file a new issue to this and assigned it to @dubeyabhishek? Thanks.

Snorch:
@avagin 
>We will not.

>checking that process creation time is strictly less than the time of pagemap reed (under freeze "lock") for a (pre-)dump, which makes these (pre-)dump valid.

After looking into the code I see that inventory_save_uptime() and __parasite_dump_pages_seized()->generate_vma_iovs() and __parasite_dump_pages_seized()->task_reset_dirty_track() are still all done then all tasks are frozen. So our current detect_pid_reuse() should make it's job fine for an optimized pre-dump too.

Sorry that I didn't get it from the beginning...

Snorch:
> I am agree that we have an issue here. A user can request to dump their processes tree and then they will receive images with some memory of other processes.

>could you file a new issue to this

Now when I've rethought everything... These optimization makes pre-dump possibly containing memory of other processes, yes. But may be it's not an issue? - Same thing will happen if we do the pre-dump a bit later, it will contain different data and possibly of different processes. We can't freeze everything immediately anyway, isn't it?

So no issue and we are fine, OK?

avagin:
Applied.

PR #791, opened 2019-09-15
crtools: flush early log using destructor

rst0git:
This patch series aims to address some of the issues described in:

https://lists.openvz.org/pipermail/criu/2019-September/044469.html
https://lists.openvz.org/pipermail/criu/2019-September/044470.html


0x7f454c46:
LGTM, sounds like a good improvement, thanks!

PR #793, opened 2019-09-15
Cleanup do_full_int80()

criupatchwork:


0x7f454c46:
I don't mind. Though, I remember that we've previously removed `"rsp"` from clobbers on some asm statements. See commit 901f5d48ec6e.
Probably, worth to check if gcc9 will still build criu without errors..

PR #794, opened 2019-09-15
test/cgroup_yard: always clean up a test cgroup yard

criupatchwork:


test/zdtm/static/cgroup_yard.hook
```diff
@@ -29,42 +30,22 @@ if sys.argv[1] == "--pre-dump":
...
+			os.path.join(tname, "subcg00")]:
+		if os.access(cg, os.F_OK):
+			os.rmdir(cg)
 
 	subprocess.call(["umount", tname])
 	os.rmdir(tname)
+
+if sys.argv[1] == "--clean":
+	if os.access(yard, os.F_OK):
+		subprocess.call(["umount", "-l", yard])
```
mclapinski: If there are still some fds open, lazy umount will success and we will not notice the hanging fds. Wouldn't `umount -R` be a better idea? It should fail if there are some fds open.

avagin: clean has to be always successful. All errors should be detected in other hooks (post-restore, post-dump).

avagin: Actually, we can leave your code which umounts cgroups in post-restore. Is it enough to address a case what you are thinking about?

And we check file descriptors in zdtm.py, you can look at get_visible_state() for more details.

avagin: I don't want to use umount -R, because it reads /proc/mounts, but there is a known race in the linux kernel when /proc/mounts doesn't show some mounts or show some mounts twice.

mclapinski: If we check fds in zdtm.py, then I think all options are good. `umount -l` is the cleanest one I think.

PR #801, opened 2019-09-24
Cgroup detect unsupported xattrs

Snorch:
Inspired by: https://github.com/checkpoint-restore/criu/pull/695
We don't c/r xattrs on cgroups, but they can be set. But we don't know any user of these xattrs so for now we can just check that nobody uses them.


Snorch:
@avagin ping

Snorch:
Thanks a lot for a review, and sorry for stupid mistakes. Reworked all according to your comments.

Snorch:
Looks like trusted.invocation_id xattr on cgroupfs mounts appeared, I both see it for cgroup-v1 and cgroup-v2. So we should implement full support instead now.

PR #802, opened 2019-09-24
cgroup: fix cg_yard leak on error path in prepare_cgroup_sfd

Snorch:


avagin:
Applied, thanks!

PR #803, opened 2019-09-24
restorer/inotify: allow several retries on cleanup due to interrupted…

Snorch:
… poll

We've got in our tests on VZ7:

(05.718449) pie: 104: Cleaning inotify events from 5
(05.718835) pie: 330: Cleaning inotify events from 3
(05.719046) pie: 1: Cleaning inotify events from 23
(05.719164) pie: 80: Cleaning inotify events from 7
(05.719185) pie: 1: Error (criu/pie/restorer.c:1287): Failed to poll from inotify fd: -4
(05.719202) pie: 95: Cleaning inotify events from 6
(05.719269) pie: 1: Error (criu/pie/restorer.c:1890): Restorer fail 1

Looks like we just need to retry in case we were interrupted.

### DIFF
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -1327,17 +1327,24 @@ static int fd_poll(int inotify_fd)
  */
 #define EVENT_BUFF_SIZE ((sizeof(struct inotify_event) + PATH_MAX))
 
+/* We can get interrupted in poll, so try some more */
+#define MAX_POLL_RETRY 5
+
 /*
  * Read all available events from inotify queue
  */
 static int cleanup_inotify_events(int inotify_fd)
 {
 	char buf[EVENT_BUFF_SIZE * 8];
-	int ret;
+	int ret, retry = 0;
 
-	while (1) {
+	while (retry < MAX_POLL_RETRY) {
 		ret = fd_poll(inotify_fd);
 		if (ret < 0) {
+			if (ret == -EINTR) {
+				retry++;
+				continue;
+			}
 			pr_err("Failed to poll from inotify fd: %d\n", ret);
 			return -1;
 		} else if (ret == 0) {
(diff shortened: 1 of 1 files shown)


avagin:
> Looks like we just need to retry in case we were interrupted.

Do you understand how it can be interrupted?

criu/pie/restorer.c
```diff
@@ -1327,17 +1327,24 @@ static int fd_poll(int inotify_fd)
...
  */
 static int cleanup_inotify_events(int inotify_fd)
 {
 	char buf[EVENT_BUFF_SIZE * 8];
-	int ret;
+	int ret, retry = 0;
 
-	while (1) {
+	while (retry < MAX_POLL_RETRY) {
 		ret = fd_poll(inotify_fd);
```
avagin: we call poll with zero timeout, it shouldn't block. why doesn't this mean that it should not be interrupted?

criu/pie/restorer.c
```diff
@@ -1327,17 +1327,24 @@ static int fd_poll(int inotify_fd)
  */
 #define EVENT_BUFF_SIZE ((sizeof(struct inotify_event) + PATH_MAX))
 
+/* We can get interrupted in poll, so try some more */
+#define MAX_POLL_RETRY 5
```
avagin: Nack. We don't expect any external signals, so we have to know when ppoll can be interrupted and how many times.

criu/pie/restorer.c
avagin: I read the kernel code quickly and if I understand it right, EINTR can be returned only if there is no event to report by.

Snorch:
Sorry, I don't have any idea which signal interrupted us and why (maybe it was connected with stack overflow problem fixed by @dsafonov), and logs are lost now.

Looking on kernel code of ppoll, we should be interrupted by signal and queue should be empty. So we can instead of retrying just handle retcode -EINTR same as 0, and just stop polling as queue is empty.

I understand that these is hiding the problem, will try to rerun our test without retries, to check if it can be reproduced after a fix for overflow.

Snorch:
Closing as it does not reproduce now.

PR #816, opened 2019-10-08
service: Use space on stack for msg buffer

rst0git:
RPC messages are have fairly small size and using space on the stack might be a better option. This change follows the pattern used with do_pb_read_one() and pb_write_one().

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -27,6 +27,7 @@
 #include "cr-service.h"
 #include "cr-service-const.h"
 #include "page-xfer.h"
+#include "protobuf.h"
 #include "net.h"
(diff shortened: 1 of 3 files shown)


criu/cr-service.c
```diff
@@ -80,43 +84,47 @@ static int recv_criu_msg(int socket_fd, CriuReq **req)
...
 
-	xfree(buf);
-	return 0;
+	ret = 0;
 err:
-	xfree(buf);
-	return -1;
+	if (buf != (void *)&local)
+		xfree(buf);
+	return ret;
```
avagin: pls declare another variable for a function exit code. Usually, we call it exit_code.

PR #817, opened 2019-10-09
page-xfer: Don't set SPLICE_F_NONBLOCK to vmsplice

rst0git:
cc: @dubeyabhishek

### DIFF
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -846,10 +846,10 @@ int page_xfer_predump_pages(int pid, struct page_xfer *xfer,
 
 		bufvec.iov_base = userbuf;
 		bufvec.iov_len = bytes_read;
-		ret = vmsplice(ppb->p[1], &bufvec, 1, SPLICE_F_NONBLOCK);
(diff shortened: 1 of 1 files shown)


criu/page-xfer.c
```diff
@@ -846,10 +846,10 @@ int page_xfer_predump_pages(int pid, struct page_xfer *xfer,
 
 		bufvec.iov_base = userbuf;
 		bufvec.iov_len = bytes_read;
-		ret = vmsplice(ppb->p[1], &bufvec, 1, SPLICE_F_NONBLOCK);
+		ret = vmsplice(ppb->p[1], &bufvec, 1, 0);
```
avagin: SPLICE_F_NONBLOCK is set intentionally. We know the size of this pipe and we call vmsplice exactly for this size. And nobody reads from this pipe when vmsplice is being called, so if this call will be blocked, it will be blocked forever.

rst0git: Thank you for explaining this!

PR #819, opened 2019-10-13
security: change CR_FD_PERM from rw-r--r-- to rw-------

rst0git:
When creating a container snapshot (e.g. using runc) all image files are by default as world-readable.

Therefore, unprivileged users on the system may be able to read data stored in container or process checkpoint.

For example, most of the files stored in the following paths have `-rw-r--r--` permissions:

(Podman) `/var/lib/containers/storage/overlay-containers/*/userdata/checkpoint/`
(LXD) `/var/snap/lxd/common/lxd/snapshots/*/*/state`

However, it should be noted that, an unprivileged user would be able to read the image files only if the parent directory has `o+x` permissions.

This patch also modifies `do_open_image()` to always use `userns_call()`. When user namespace is not being restored, this is equivalent to directly calling `openat()`.

`userns_call` has been updated (fixed) to send back the value of `errno`. This is necessary to properly handle (skip) missing image files.

### DIFF
--- a/criu/image.c
+++ b/criu/image.c
@@ -460,28 +460,24 @@ static int do_open_image(struct cr_img *img, int dfd, int type, unsigned long of
 
 	flags = oflags & ~(O_NOBUF | O_SERVICE | O_FORCE_LOCAL);
 
-	if (opts.remote && !(oflags & O_FORCE_LOCAL))
+	if (opts.remote && !(oflags & O_FORCE_LOCAL)) {
 		ret = do_open_remote_image(dfd, path, flags);
-	else {
+	} else {
 		/*
-		 * For pages images dedup we need to open images read-write on
-		 * restore, that may require proper capabilities, so we ask
-		 * usernsd to do it for us
+		 * When userns is enabled proper capabilities may be required
+		 * to open images, so we ask usernsd to do it for us
 		 */
-		if (root_ns_mask & CLONE_NEWUSER &&
-		    type == CR_FD_PAGES && oflags & O_RDWR) {
-			struct openat_args pa = {
-				.flags = flags,
-				.err = 0,
-				.mode = CR_FD_PERM,
-			};
-			snprintf(pa.path, PATH_MAX, "%s", path);
-			ret = userns_call(userns_openat, UNS_FDOUT, &pa, sizeof(struct openat_args), dfd);
-			if (ret < 0)
-				errno = pa.err;
-		} else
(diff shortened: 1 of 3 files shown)


adrianreber:
This sounds correct. Cannot really comment on the actual code changes as I have never looked at that before, but having image files as 600 make sense.

criu/image.c
```diff
@@ -460,28 +460,23 @@ static int do_open_image(struct cr_img *img, int dfd, int type, unsigned long of
...
-				errno = pa.err;
-		} else
-			ret = openat(dfd, path, flags, CR_FD_PERM);
+		struct openat_args pa = {
+			.flags = flags,
+			.err = 0,
+			.mode = CR_FD_PERM,
+		};
+		snprintf(pa.path, PATH_MAX, "%s", path);
+		ret = userns_call(userns_openat, UNS_FDOUT, &pa, sizeof(struct openat_args), dfd);
```
avagin: How does this affect performance? userns_call should be much slower than the openat system call.

criu/image.c
```diff
@@ -460,28 +460,23 @@ static int do_open_image(struct cr_img *img, int dfd, int type, unsigned long of
...
-	if (opts.remote && !(oflags & O_FORCE_LOCAL))
+	if (opts.remote && !(oflags & O_FORCE_LOCAL)) {
 		ret = do_open_remote_image(dfd, path, flags);
-	else {
-		/*
-		 * For pages images dedup we need to open images read-write on
-		 * restore, that may require proper capabilities, so we ask
-		 * usernsd to do it for us
+	} else {
+		/* When userns is enabled proper capabilities may be required
```
avagin: /*
 * When ..

criu/image.c
rst0git:
```C

int __userns_call(const char *func_name, uns_call_t call, int flags,
		  void *arg, size_t arg_size, int fd)
{
...
	if (!usernsd_pid)
		return call(arg, fd, getpid());
```
Any overhead of `userns_call()` should be only between `start_usernsd()` and `stop_usernsd()`.

avagin:
> However, it should be noted that, an unprivileged user would be able to read the image files only if the parent directory has o+x permissions.

Maybe it is better to set a right permission for the image directory?

Cc: @0x7f454c46 @xemul @rppt @cyrillos

cyrillos:
> Maybe it is better to set a right permission for the image directory?
sounds like a plan

Issue #820, opened 2019-10-14
Include tests for validating Java APIs work fine with CRIU


nidhi1605:
The goal of this issue is to include tests to verify if Java Applications work with CRIU. We plan to introduce Java based CRIU APIs so that a Java App can do self checkpoint and restore.

We have included Java functional tests that uses the Java File based APIs, Memory mapping APIs and Socket based APIs and places the process in various states before check pointing and validates if these resources are still accessible after restore. It also validates if the file contents or socket connections are in the expected state.

The file based tests are here in our local repo - https://github.com/nidhi1605/criu/tree/javaTests/test/javaTests
Let us know if the format of the tests are fine, we will create a PR here.

adrianreber:
> We plan to introduce Java based CRIU APIs so that a Java App can do self checkpoint and restore.

Where do you plan to introduce those APIs?

If you could share some additional information it would make it a lot easier to understand what this is exactly about. I had a look at the branch and there are a lot of commits and it is hard to tell what this is really about.

Right now it seems you want to include a Java wrapper, maybe it could be called library, into the test directory. One of my main questions is, where should the Java <-> CRIU interface live which you want to test. If it is a library it might leave under lib/, but I know there is an external Ruby lib and recently will also moved the go-bindings out of the main tree.

I also see that the test uses maven, from my last encounter with maven I remember it takes forever. So I am a bit worried about our test runtime if this is included.

My main questions right now are, where should the Java binding live? If we know that location it makes sense to talk about tests. At least for me. Some background about your use case would also be helpful.

dinogun:
Hi @adrianreber, Thank you for your prompt response, Really appreciate it ! 

The question that we have is about adding Java application tests for the existing CRIU Interface to begin with. The goal is to augment the existing tests by adding Java application tests that test CRIU. I think this can be optional tests so that they have to be specifically invoked for them to run. That way they do not add additional overhead for your everyday testing.

The main aim is to help the Java ecosystem to get more confidence in CRIU and to demonstrate that regular Java applications work well with CRIU. (and hopefully iron out any issues that we find along the way). This would be the first step.

The next step would be to introduce Java API's for doing checkpoint-restart, so that Java applications can self checkpoint. We are still a long way from this and there have been some design discussions at IBM and others and I think needs a broader participation from the Java community to help arrive at the right design.

[...]

PR #825, opened 2019-10-18
Add Java-based CRIU Tests.

nidhi1605:
As noted in issue #820, we are adding sample Java tests to check CRIU APIs. Currently this only checks File-based CRIU APIs. Based on feedback we can add the remaining Java tests that we have been working on.

These tests take 1.5 minutes when run the first time as maven needs to build its repo. Subsequent runs on the same machine will take about 6 seconds as the required files will be cached locally.

### DIFF
--- a/test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
+++ b/test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
@@ -1,0 +1,450 @@
+package org.criu.java.tests;
+
+import org.testng.Assert;
+import org.testng.annotations.AfterTest;
+import org.testng.annotations.BeforeSuite;
+import org.testng.annotations.Parameters;
+import org.testng.annotations.Test;
+
+import java.io.*;
+import java.nio.MappedByteBuffer;
+import java.nio.channels.FileChannel;
+import java.nio.channels.FileChannel.MapMode;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.nio.file.StandardOpenOption;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+
+public class CheckpointRestore {
+	private MappedByteBuffer mappedByteBuffer = null;
+	private String testName = "";
+	private String logFolder = Helper.LOG_FOLDER + "/";
+	private String outputFolder = Helper.OUTPUT_FOLDER_NAME + "/";
+
+	/**
+	 * Create CRlog and output directory if they don't exist.
+	 * Delete directories containing .img files from failed Checkpoint-Restore if 'neverCleanFailures' property is not set to true.
+	 *
+	 * @throws IOException
+	 */
+	@BeforeSuite
+	void suiteSetup() throws IOException {
+		System.out.println("Tests are to be run as a privileged user having capabilities mentioned in ReadMe");
+		boolean neverCleanFailures = Boolean.getBoolean("neverCleanFailures");
+		Path logDir = Paths.get(logFolder);
+		Path outputDir = Paths.get(outputFolder);
+		if (!Files.exists(logDir)) {
+			System.out.println("Logs directory does not exist, creating it");
+			Files.createDirectory(logDir);
+		}
+		if (!Files.exists(outputDir)) {
+			System.out.println("Output directory does not exist, creating it");
+			Files.createDirectory(outputDir);
+		}
+		/*
+		 * Delete the directories containing the img files from failed Checkpoint-Restore.
+		 */
+		if (!neverCleanFailures) {
+			File output = new File(outputFolder);
+			String[] name = output.list();
+			for (int i = 0; null != name && i < name.length; i++) {
+				File testFolder = new File(outputFolder + name[i]);
+				if (testFolder.isDirectory()) {
+					String[] list = testFolder.list();
+					File file;
+					if (null != list) {
+						for (int j = 0; j < list.length; j++) {
+							file = new File(outputFolder + name[i] + "/" + list[j]);
+							if (!file.isDirectory()) {
+								Files.delete(file.toPath());
+							}
+						}
+					}
+				}
+				Files.delete(testFolder.toPath());
+			}
+		}
+	}
+
+	/**
+	 * Create the output folder for the test in case it does not exist
+	 *
+	 * @param testName Name of the java test
+	 * @throws IOException
+	 */
+	private void testSetup(String testName) throws IOException {
+		Path testFolderPath = Paths.get(outputFolder + testName + "/");
+		if (!Files.exists(testFolderPath)) {
+			System.out.println("Test Folder does not exist creating it");
+			Files.createDirectory(testFolderPath);
+		}
+	}
+
+	/**
+	 * Read the pid of process from the pid file of test
+	 *
+	 * @param name Name of the java test
+	 * @return pid Process id of the java test process
+	 * @throws IOException
+	 */
+	private String getPid(String name) throws IOException {
+		name = outputFolder + testName + "/" + name + Helper.PID_APPEND;
+		File pidfile = new File(name);
+		BufferedReader pidReader = new BufferedReader(new FileReader(pidfile));
+		String pid = pidReader.readLine();
+		pidReader.close();
+		return pid;
+	}
+
+	/**
+	 * @param testName      Name of the java test
+	 * @param checkpointOpt Additional options for checkpoint
+	 * @param restoreOpt    Additional options for restore
+	 * @throws Exception
+	 */
+	@Test
+	@Parameters({"testname", "checkpointOpt", "restoreOpt"})
+	public void runtest(String testName, String checkpointOpt, String restoreOpt) throws Exception {
+		this.testName = testName;
+		String name = Helper.PACKAGE_NAME + "." + testName;
+		String pid;
+		int exitCode;
+
+		System.out.println("======= Testing " + testName + " ========");
+
+		testSetup(testName);
+
+		File f = new File(Helper.MEMORY_MAPPED_FILE_NAME);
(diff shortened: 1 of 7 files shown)


adrianreber:
Can you remove the merge commit and squash the update readme commit in the previous commit?

nidhi1605:
Removed the merge commit and squashed the other 2 commits.

rst0git:
Would it be possible to make the source files non-executable (i.e. `chmod 644`), use tab indentation consistently and remove trailing white-space?

nidhi1605:
> Would it be possible to make the source files non-executable (i.e. `chmod 644`), use tab indentation consistently and remove trailing white-space?

Done

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
```diff
@@ -0,0 +1,345 @@
...
+
+	/**
+	 * Create CRlog and output directory if they don't exist.
+	 * Delete failed dump and restore logs if they exist.
+	 *
+	 * @throws IOException
+	 */
+	@BeforeSuite
+	void suiteSetup() throws IOException {
+	System.out.println("Tests are to be run as root user!!");
```
rst0git: This is slightly misleading. If you run the tests as *root* user inside an unprivileged container are they going to fail?

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
```diff
@@ -0,0 +1,345 @@
...
+	}
+
+	/**
+	 * Create the output folder for the test in case it does not exist.
+	 *
+	 * @param testName Name of the java test
+	 * @throws IOException
+	 */
+	private void testSetup(String testName) throws IOException {
+	Path testFolderPath = Paths.get(outputFolder + testName + "/");
```
rst0git: the body of the methods in this file should be indented?

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
```diff
@@ -0,0 +1,345 @@
...
+	 */
+	@Test
+	@Parameters({"filename"})
+	public void runtest(String testName) throws Exception {
+	this.testName = testName;
+	String name = Helper.PACKAGE_NAME + "." + testName;
+	String pid;
+	int exitCode;
+
+	System.out.println("=======Testing " + testName + " ========");
```
rst0git: s/=Testing/= Testing/

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
```diff
@@ -0,0 +1,345 @@
...
+
+	Assert.assertNotEquals(currentState, Helper.STATE_TERMINATE, testName + ": ERROR: Checkpoint-Restore failed");
+	Assert.assertNotEquals(currentState, Helper.STATE_FAIL, testName + ": ERROR: Test Failed, Check Log for details");
+	Assert.assertEquals(currentState, Helper.STATE_PASS, testName + " ERROR: Unexpected State of Mapped Buffer");
+	System.out.println("-----" + "PASS" + "-----");
+
+	}
+
+	/**
+	 * Remove .img files, stats-dump and stats-restore files from Log Directory
```
rst0git: Actually keeping these files is useful for debugging, for example, the stats-* files contain information about how much time it took for checkpoint/restore and the .img files can be used to inspect the state of a process.

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
```diff
@@ -0,0 +1,345 @@
...
+	 * test as failed
+	 *
+	 * @param pid Pid of process to be checkpointed
+	 * @throws IOException
+	 * @throws InterruptedException
+	 */
+	private void checkpoint(String pid) throws IOException, InterruptedException {
+	System.out.println("Checkpointing process " + pid);
+
+	ProcessBuilder builder = new ProcessBuilder("criu", "dump", "--tcp-established", "--shell-job", "-t", pid, "-vvv", "-D", logFolder, "-o", "dump.log");
```
rst0git: It would be better to specify CRIU options per test case. Also we should be testing with the compiled `criu` executable not the one installed on the system.

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
```diff
@@ -0,0 +1,345 @@
...
+	 * state and mark test as failed
+	 *
+	 * @throws IOException
+	 * @throws InterruptedException
+	 */
+	private void restore() throws IOException, InterruptedException {
+	System.out.println("Restoring process");
+
+	String currentPath = System.getProperty("user.dir");
+	ProcessBuilder builder = new ProcessBuilder(currentPath + "/../../scripts/criu-ns", "restore", "-d", "-vvv", "--tcp-established", "--shell-job", "-D", logFolder, "-o", "restore.log");
```
rst0git: Why do we need to use `criu-ns`? It is just a wrapper around criu that creates a new PID namespace.

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
nidhi1605: Corrected Indentation.

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
nidhi1605: Done.

avagin:
Could you add a travis job to run this test?
https://github.com/checkpoint-restore/criu/blob/criu-dev/.travis.yml

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
nidhi1605: If I do not use criu-ns to restore the process in a new namespace, the restore fails because thread-id is not available( thread-id mismatch).
```
(00.027183) pie: 13541: Error (criu/pie/restorer.c:555): Thread pid mismatch 13541/13540
(00.027193) pie: 13541: Error (criu/pie/restorer.c:609): Restorer abnormal termination for 13532
(00.027316) Error (criu/cr-restore.c:2294): Restoring FAILED.
```

adrianreber: Using `criu-ns` you create a new PID namespace and thus avoid PID collisions. Is the test runner part (a thread) of the process you are checkpointing/restoring?

If you are trying to checkpoint and restore yourself while still running this will not work. You either need to checkpoint/restore a separate process or include the restore in your startup code, but if there are still processes or threads running of the process you are trying to restore it will result in PID collisions. Restoring in a new PID namespace is a workaround, but I am not sure that is the actual thing you want to test.

Maybe some description in your test case what exactly you are trying to do here would be helpful.

nidhi1605: CheckpointRestore.java is our test runner, and it does not checkpoint itself. In CheckpointRestore.java, we run the test (FileRead.java) as a new process and it is this process we checkpoint and restore. The CheckpointRestore.java upon receiving a message from the test process checkpoints the test process, there is no "leave-running" option used when checkpointing the process. After the process has been checkpointed (and killed), CheckpointRestore.java restores the test process(currently in a new namespace using criu-ns).  So there is no thread that remains running.
What happens if after checkpointing, when the process has been killed, a thread-id that had been used by the test process gets used by some other process running on the machine. Won't the restore fail because the thread-id is not available.

nidhi1605: No, it does not always fail. It fails sometimes if we are not in a new namespace.

adrianreber: > No, it does not always fail. It fails sometimes if we are not in a new namespace.

Do you know which process is blocking the restore? Some random process from the system or a process involved in the restore?

chandrams: We believe it is a random process, but need to confirm on that. 

Nidhi, can you retain the .img files and restore once again after confirming the thread ID is not in use.

nidhi1605: I restored the process again after ensuring the thread-id was not in use, and it restored successfully. It should be a random process, however I am unable to confirm that it was not a process involved in restore. Is there a way to do so? I am looking at the processes running on my system along with the thread-id of the threads used by them ( ps -elfT ) but I cannot find the thread-id which caused failure over there.

chandrams: @adrianreber - Are there any logging options that we can specify to log the process that was holding the thread-id when restore fails?

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
nidhi1605: I have cleared this up, and mentioned the capabilities required by the user running these tests in README.

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
nidhi1605: Now the .img files, log files stats-dump and stats-restore are saved in a different folder on failure. This folder is deleted the next time these tests are run. However if we do not want the folder to be deleted on the following runs, the -DneverCleanFailures option can be used, as mentioned in README

test/javaTests/src/org/criu/java/tests/CheckpointRestore.java
nidhi1605: CRIU options are now specified per test case, and the compiled CRIU is being used.

avagin:
Applied, thanks!

PR #831, opened 2019-11-01
travis: start to use aarch64 hardware

adrianreber:
With the newly introduced aarch64 at Travis it is possible for the CRIU test-cases to switch to aarch64.

Travis uses unprivileged LXD containers on aarch64 which blocks many of the kernel interfaces CRIU needs. So for now this only tests building CRIU natively on aarch64 instead of using the Docker+QEMU combination.

All tests based on Docker are not working on aarch64 is there currently seems to be a problem with Docker on aarch64. Maybe because of the nesting of Docker in LXD.

The problem with Docker on LXD was reported here: https://travis-ci.community/t/docker-container-download-fails/5674/2

Once this is fixed I will try to move all other arm32/arm64 related tests to native hardware.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,9 @@ env:
   - TR_ARCH=x86_64
   - TR_ARCH=x86_64      CLANG=1
   - TR_ARCH=armv7hf
-  - TR_ARCH=aarch64
   - TR_ARCH=ppc64le
   - TR_ARCH=s390x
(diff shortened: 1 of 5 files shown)


0x7f454c46:
Thanks for doing this! Looks good to me.
15 mins build for arm64 => 1.5 min, yay!

avagin:
Applied. Thanks a lot!

PR #834, opened 2019-11-04
Run all tests on real hardware (without any qemu emulation anymore)

adrianreber:
This changes the rawhide-aarch64 build and the armv7hf to use Travis aarch64 hardware.

This PR required a few more changes. Docker container can no longer run with --privileged as it is running in an unprivileged LXD container.

The Travis aarch64 system has uname of armv8l when running under `setarch linux32`. To  enable arm32 builds I added armv8l detection to the Makefile. If it is an armv8l system CRIU will be built just as for an armv7hf system. Not verified that the output binary actually works. Only used for compilation tests.


adrianreber:
Something went wrong with my branches. Need to rebase.

adrianreber:
The ppc64le build timed out. I thought I was able to restart builds in Travis, but the button to restart the build is gone.

adrianreber:
The ppc64le job without CLANG takes 50 minutes and the ppc64le job with CLANG takes only 20 minutes. As 50 minutes is the maximum Travis runtime for a test we seem to hit this more often.

It is interesting that the same test requires less than half the time with CLANG=1.

Makefile
```diff
@@ -45,6 +44,15 @@ ifeq ($(ARCH),arm)
...
 
+        ifeq ($(ARMV),8)
+                # Untested hack to run the arm32 bit test on Travis aarch64 with 'setarch linux32'.
+                # Running 'setarch linux32 uname -m' returns armv8l on travis aarch64.
+                USERCFLAGS += -march=armv7-a
+                ARMV := 7
+        endif
+
+        DEFINES		:= -DCONFIG_ARMV$(ARMV) -DCONFIG_VDSO_32
+
```
0x7f454c46: I think it makes sense - it's in reality not armv7, but aarch32 which is instruction-compatible with first.
My only concern would be that vdso on arm32 differ from compatible vdso on aarch64, but so far we check only build - it shouldn't make any difference.

Please, change the commentary from `Untested hack to run the arm32` to something more appropriate after Travis-CI passes.

adrianreber: > Please, change the commentary from `Untested hack to run the arm32` to something more appropriate after Travis-CI passes.

Travis-CI has already passed. Should I change it before or after merge?

0x7f454c46: Whatever is more suitable - probably a patch on the top that just changes the comment won't need much testing..

adrianreber:
And now there is also support for ppc64le and s390 from Travis: https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z

Now we can drop emulation for all arches.

Makefile
adrianreber: Will do it in a follow up.

adrianreber:
I tried the s390x and ppc64le support here: https://travis-ci.org/adrianreber/criu/builds/611253563

s390x test times are done from 18 minutes to 41 seconds.

ppc64le from 20 minutes (sometime 50 minutes) to less than 2 minutes.

adrianreber:
Restarting the ppc64le test. Time out after 50 minutes.

0x7f454c46:
> I tried the s390x and ppc64le support here: https://travis-ci.org/adrianreber/criu/builds/611253563
> s390x test times are done from 18 minutes to 41 seconds.
> ppc64le from 20 minutes (sometime 50 minutes) to less than 2 minutes.

That's very promising, would you send patches/PR?
I'm eager to Ack them, haha

0x7f454c46:
> Restarting the ppc64le test. Time out after 50 minutes.

I don't think it's a problem with this PR - I have added include `compiler.h` in some other patches set and it times out on ppc64 by the same reason (on criu-dev it passes with 30 sec margin to 50 mins, which is very narrow).

adrianreber:
> That's very promising, would you send patches/PR?

Patches are ready and I will open a new PR as soon as this one is merged.

scripts/travis/Makefile
```diff
@@ -23,17 +26,29 @@ define DOCKER_JSON
 endef
 
 export DOCKER_JSON
-$(TARGETS):
-	echo "$$DOCKER_JSON" > /etc/docker/daemon.json
-	systemctl restart docker
+
+ifeq ($(UNAME),x86_64)
```
avagin: Need a comment here to explain why we need different options

adrianreber: done

adrianreber:
I included the ppc64le and s390x changes also in this PR.

adrianreber:
@alicefr : just as an FYI, the s390x test case is now running on real hardware, without qemu emulation. (as discussed previously)

alicefr:
@adrianreber thanks! Great news :)

avagin:
Applied. Thanks a lot! 
great-job-graphics-2416-greatjob

PR #835, opened 2019-11-06
Checkpoint only specified controllers

mclapinski:
Before this change CRIU would checkpoint all controllers, even the ones not specified in --cgroup-dump-controller. That becomes a problem if there's a cgroup controller on the checkpointing machine that doesn't exist on the restoring machine even if CRIU is instructed not to dump that controller. After that change everything works as expected.


avagin:
@Snorch @cyrillos could you review this pr?

avagin:
Applied, thanks!

PR #837, opened 2019-11-07
Run java functional tests on travis

nidhi1605:
Add travis job to run Java functional tests on travis, as per #825 

Signed-off-by: Nidhi Gupta <itsnidhi16@gmail.com>

### DIFF
--- a/scripts/build/Dockerfile.openj9-alpine
+++ b/scripts/build/Dockerfile.openj9-alpine
@@ -1,0 +1,33 @@
+FROM adoptopenjdk/openjdk8-openj9:alpine
+
+RUN apk update && apk add \
+	bash \
+	build-base \
+	ccache \
+	coreutils \
+	git \
+	gnutls-dev \
+	libaio-dev \
+	libcap-dev \
+	libnet-dev \
+	libnl3-dev \
+	pkgconfig \
+	protobuf-c-dev \
+	protobuf-dev \
+	python \
+	sudo \
+	maven \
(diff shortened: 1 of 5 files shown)


scripts/build/Dockerfile.openj9-alpine
```diff
@@ -0,0 +1,35 @@
...
+	protobuf-c-dev \
+	protobuf-dev \
+	python \
+	sudo \
+	maven
+
+COPY . /criu
+WORKDIR /criu
+
+RUN apk add \
```
adrianreber: Why are those packages not installed in the previous run of `apk add`?

nidhi1605: I have changed it. All the packages are now installed together.

adrianreber:
Looks good. All tests are passing. Thanks!

scripts/travis/travis-tests
```diff
@@ -170,6 +170,8 @@ make -C test/others/make/ run CC="$CC"
 make -C test/others/shell-job/ run
 make -C test/others/rpc/ run
 
+mvn -f test/javaTests/pom.xml test
```
avagin: I don't think that we need to add this test here. We have two separate travis jobs to run this test and I think these should be enough.

nidhi1605: I have removed the test from travis-tests.sh.

avagin:
Applied, thanks!

PR #842, opened 2019-11-10
restorer: allocate a stack guard page

avagin:
An accidental overflow could corrupt the area located below a stack,
so let's map an extra page there without read-write permissions.


criu/cr-restore.c
```diff
@@ -3532,8 +3535,15 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 		seccomp_rst_reloc(&thread_args[i]);
 		thread_args[i].seccomp_force_tsync = rsti(current)->has_old_seccomp_filter;
 
-		thread_args[i].mz = mz + i;
-		sigframe = (struct rt_sigframe *)&mz[i].rt_sigframe;
+		/* Allocate a guard page. */
+		if (mprotect(mem_zones + i * restore_mem_zone_size, page_size(), PROT_NONE)) {
```
0x7f454c46: Not `page_size()`, please, but `RESTORE_STACK_REDZONE`

0x7f454c46:
I guess, there are still two issues I see:
1. `bootstrap_len` is a continuous block so we might fail to find this large amount for restorer - may be worth to split it on multiple vmas (though, it may be not a problem on 64-bit systems).
2. Now the redzone is one page, but in the original failure, it was an array that has a couple of kilobytes size, which means 4Kb maybe not enough for structures used on restorer's stack.

Anyway, the patch seems good to me and is a nice start that may catch new issues.
With a minor nit, you can add my `Reviewed-by: Dmitry Safonov <0x7f454c46@gmail.com>` or which way is now preferred :)
(I don't have a button that would mark PR as reviewed)

0x7f454c46:
@avagin ping? :)

PR #848, opened 2019-11-12
net: add nftables c/r

criupatchwork:
This PR is autogenerated from: https://patchwork.criu.org/series/3871


cyrillos:
Looks good to me as well.

criu/util.c
```diff
@@ -795,6 +795,45 @@ struct vma_area *alloc_vma_area(void)
...
 
+static int __check_cmd_exists(const char *path)
+{
+	return (access(path, F_OK) == 0);
+}
+
+int check_cmd_exists(const char *cmd)
+{
+	int ret = 0;
+	char buf[255];
```
Snorch: It is generally a bad practice to use magic numbers without defining them. We need to use PATH_MAX here.

mihalicyn:
https://github.com/checkpoint-restore/criu/pull/861

PR #849, opened 2019-11-14
nftables support (uses nft utility)

mihalicyn:


mihalicyn:
https://github.com/checkpoint-restore/criu/pull/861

PR #850, opened 2019-11-14
unix sysctl max-dgram-qlen preserve

mihalicyn:
This patch set resolves problem of changing sysctl net.unix.max_dgram_qlen value after C/R

### DIFF
--- a/criu/net.c
+++ b/criu/net.c
@@ -210,6 +210,19 @@ char *devconfs6[] = {
 #define MAX_CONF_OPT_PATH IFNAMSIZ+60
 #define MAX_STR_CONF_LEN 200
 
+static const char *unix_conf_entries[] = {
+	"max_dgram_qlen",
+};
+
+/*
+ * MAX_CONF_UNIX_PATH = (sizeof(CONF_UNIX_FMT) - strlen("%s"))
+ * 					  + MAX_CONF_UNIX_OPT_PATH
+ */
+#define CONF_UNIX_BASE		"net/unix"
+#define CONF_UNIX_FMT		CONF_UNIX_BASE"/%s"
+#define MAX_CONF_UNIX_OPT_PATH	32
+#define MAX_CONF_UNIX_PATH	(sizeof(CONF_UNIX_FMT) + MAX_CONF_UNIX_OPT_PATH - 2)
+
 static int net_conf_op(char *tgt, SysctlEntry **conf, int n, int op, char *proto,
 		struct sysctl_req *req, char (*path)[MAX_CONF_OPT_PATH], int size,
 		char **devconfs, SysctlEntry **def_conf)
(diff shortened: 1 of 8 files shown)


mihalicyn:
@avagin in mailing list asked several questions about patch set. Let me quote and answer here.

```
> test_init creates a new network namespace. Why do we need one more? Do
> you want to add the "h" flavor?
```
I've removed unnecessary `unshare()`. In `h` flavor test is meaningless because we don't change net namespace.

```
> +++ b/test/zdtm/static/netns_sub_sysctl.desc
> @@ -0,0 +1,4 @@
> +{
> +    'flavor': 'ns',

uns?
```
In `uns` flavor test will fail because of lack /proc/sys/net/unix/max_dgram_qlen

mihalicyn:
```
> > --- a/criu/net.c
> > +++ b/criu/net.c
> > @@ -210,6 +210,15 @@ char *devconfs6[] = {
> >  #define MAX_CONF_OPT_PATH IFNAMSIZ+60
> >  #define MAX_STR_CONF_LEN 200
> >  
> > +static char *unix_conf_entries[] = {
> > +	"max_dgram_qlen",
> > +};
> > +
> > +#define CONF_UNIX_BASE		"net/unix"
> > +#define CONF_UNIX_FMT		CONF_UNIX_BASE"/%s"
> > +#define MAX_CONF_UNIX_OPT_PATH	32

Need to add a comment which explains this "-2"
```
done

all previous comments/suggestions resolved

cc @avagin

mihalicyn:
@rst0git @0x7f454c46 @adrianreber 
Guys, could you look, please?

all tests except `podman` passed

adrianreber:
I am waiting for upstream podman to release 1.6.4. Once that is in the PPA we are using for testing the test should work again. So right now please ignore the Podman error.

criu/net.c
```diff
@@ -339,6 +352,72 @@ static int ipv6_conf_op(char *tgt, SysctlEntry **conf, int n, int op, SysctlEntr
 			devconfs6, def_conf);
 }
 
+static int unix_conf_op(SysctlEntry ***rconf, size_t *n, int op)
+{
+	int i, ret = -1, flags = op == CTL_READ ? CTL_FLAGS_OPTIONAL : 0;
```
rst0git: could we simplify this line, for example, with something like:
```C
int flags = 0;
....
if (opts.weak_sysctls || op == CTL_READ)
	flags = CTL_FLAGS_OPTIONAL;
```

mihalicyn: sure. Thanks ;)

test/zdtm/lib/sysctl.c
```diff
@@ -0,0 +1,59 @@
...
+err:
+	close(fd);
+	return ret;
+}
+
+int sysctl_write_int(const char *name, int val)
+{
+	int fd;
+	int ret;
+	char buf[32];
```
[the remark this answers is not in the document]
mihalicyn: > `INT_MAX=32767` and `strlen("32767\n")=6`, why 32?

Hm, `INT_MAX=2147483647` (4 bytes), not `32767` is `SHORT_MAX` (2 byte integers).
Then, we could decrease stack usage from `32` to `16` without any problems.
P.S. `strlen("-2147483648")=11` and `strlen("2147483647")=10` and plus null-byte. So safe minimum is `12` bytes. Let's take `16` for beauty :D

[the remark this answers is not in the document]
mihalicyn: I've changed to `16`.

mihalicyn:
@rst0git thank you very much for the review!

test/zdtm/static/netns_sub_sysctl.c
```diff
@@ -0,0 +1,54 @@
...
+			pr_perror("Can't change %s", p->path);
+			return -1;
+		}
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	for (p = net_unix_params; p->path != NULL; p++) {
+		sysctl_read_int(p->path, &p->new);
```
0x7f454c46: nit: check the result of `sysctl_read_int()`

criu/net.c
```diff
@@ -339,6 +352,72 @@ static int ipv6_conf_op(char *tgt, SysctlEntry **conf, int n, int op, SysctlEntr
...
+static int unix_conf_op(SysctlEntry ***rconf, size_t *n, int op)
+{
+	int i, ret = -1, flags = 0;
+	char path[ARRAY_SIZE(unix_conf_entries)][MAX_CONF_UNIX_PATH] = { };
+	struct sysctl_req req[ARRAY_SIZE(unix_conf_entries)] = { };
+	SysctlEntry **conf = *rconf;
+
+	if (*n != ARRAY_SIZE(unix_conf_entries)) {
+		pr_err("unix: Unexpected entries in config (%u %u)\n",
+			(unsigned)*n, (unsigned)ARRAY_SIZE(unix_conf_entries));
```
0x7f454c46: Nit: `%zu` and down with the `(unsigned)` casts.

criu/net.c
```diff
@@ -339,6 +352,72 @@ static int ipv6_conf_op(char *tgt, SysctlEntry **conf, int n, int op, SysctlEntr
...
+		req[i].flags = flags;
+
+		switch (conf[i]->type) {
+		case SYSCTL_TYPE__CTL_32:
+			req[i].type = CTL_32;
+			req[i].arg = &conf[i]->iarg;
+			break;
+		default:
+			pr_err("unix: Unknown config type %d\n",
+				(unsigned)conf[i]->type);
```
0x7f454c46: Nit: casting to `(unsigned)` and printing as signed? Hmm :)

mihalicyn:
I've updated code according all recomendations/suggestions. Thank you, guys.

PR #843, opened 2019-11-10
mount: Bind-mount root via userns_call

criupatchwork:
This PR is autogenerated from: https://patchwork.criu.org/series/3860

PR #856, opened 2019-11-17
mount: Bind-mount root via userns_call

rst0git:
When restoring a runc container with enabled user namespace CRIU fails to mount the specified root directory because the path is under `/run/runc` which is inaccessible to unprivileged users.

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -1589,7 +1589,7 @@ static void restore_pgid(void)
 static int mount_proc(void)
 {
 	int fd, ret;
-	char proc_mountpoint[] = "crtools-proc.XXXXXX";
+	char proc_mountpoint[] = "/tmp/crtools-proc.XXXXXX";
 
 	if (root_ns_mask == 0)
 		fd = ret = open("/proc", O_DIRECTORY);
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -62,6 +62,7 @@ tests_root = None
 def clean_tests_root():
     global tests_root
     if tests_root and tests_root[0] == os.getpid():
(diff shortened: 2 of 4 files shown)


Snorch:
Except for test https://github.com/checkpoint-restore/criu/pull/843#issuecomment-554805622 these looks good for me:

test/zdtm.py
```diff
@@ -481,6 +484,13 @@ def start(self):
             # move into some semi-random state
             time.sleep(random.random())
 
+        if self.__flavor.ns:
```
Snorch: Does it work for uns?

avagin: yes

criu/cr-restore.c
```diff
@@ -1589,7 +1589,7 @@ static void restore_pgid(void)
 static int mount_proc(void)
 {
 	int fd, ret;
-	char proc_mountpoint[] = "crtools-proc.XXXXXX";
+	char proc_mountpoint[] = "/tmp/crtools-proc.XXXXXX";
```
avagin: do we remove it somewhere?

rst0git: in `open_detach_mount()` I think

test/zdtm.py
rst0git: fedora-asan tests fail for some reason
https://api.travis-ci.org/v3/job/614299079/log.txt

criu/cr-restore.c
avagin: BTW: you can try to replace this mkdir,mount,open,umount,rmdir by fsopen,fsmount.
https://github.com/torvalds/linux/blob/master/samples/vfs/test-fsmount.c

PR #861, opened 2019-11-21
nftables support (without fork/exec)

mihalicyn:
This patchset adds nftables dump/restore functionality in CRIU using libnftables API.
Also it ads zdtm test-case.

Warning: Currently, CentOS not ships nftables-devel package as Fedora do. I've removed installing nft into our test docker containers to make nft tests skipped in travis.
Ubuntu 18.04 also lacks libnftables package.


mihalicyn:
I've marked PR as a draft because currently, we meet a problem in containers migration - some users have both iptables and nftables in their containers in this situation are possible to catch a bug with network lock in criu (because network lock uses iptables).

@cyrillos

criu/net.c
```diff
@@ -1818,6 +1819,52 @@ static inline int dump_iptables(struct cr_imgset *fds)
...
+	nft = nft_ctx_new(NFT_CTX_DEFAULT);
+	if (!nft)
+		return -1;
+
+	img = img_from_set(fds, CR_FD_NFTABLES);
+	img_fd = dup(img_raw_fd(img));
+	if (img_fd < 0) {
+		pr_perror("dup() failed");
+		goto nft_ctx_free_out;
+	}
```
Snorch: Why you need a dup here? Why you can't use the fd returned by img_raw_fd? You also forget to close img_fd.

criu/net.c
```diff
@@ -1818,6 +1819,52 @@ static inline int dump_iptables(struct cr_imgset *fds)
...
+
+	nft_ctx_set_output(nft, fp);
+
+	if (nft_run_cmd_from_buffer(nft, "list ruleset")) {
+		goto fp_close_out;
+	}
+
+	fflush(fp);
```
Snorch: Why we need fflush here, isn't it enough to fclose?

criu/net.c
Snorch: Ah ok I see in fdopen man, that fclose will close the fd. My bad.

criu/net.c
```diff
@@ -1818,6 +1819,52 @@ static inline int dump_iptables(struct cr_imgset *fds)
...
+static inline int dump_nftables(struct cr_imgset *fds)
+{
+	int ret = -1;
+	struct cr_img *img;
+	int img_fd;
+	FILE *fp;
+	struct nft_ctx *nft;
+
+	/* we not dump nftables if nft utility isn't present */
+	if (!check_cmd_exists("nft"))
```
Snorch: Looks like we don't need check_cmd_exists in case we use a static library.

criu/net.c
```diff
@@ -1818,6 +1819,52 @@ static inline int dump_iptables(struct cr_imgset *fds)
...
+	img = img_from_set(fds, CR_FD_NFTABLES);
+	img_fd = dup(img_raw_fd(img));
+	if (img_fd < 0) {
+		pr_perror("dup() failed");
+		goto nft_ctx_free_out;
+	}
+
+	fp = fdopen(img_fd, "w");
+	if (!fp) {
+		pr_perror("fdopen() failed");
```
Snorch: close(img_fd);

criu/net.c
mihalicyn: >  img_fd.

I've checked other places in CRIU, and realized that in all places we don't close descriptor that returned by img_raw_fd. Therefore, I need to dup descriptor -> then close dupped. I've used fclose() and fclose() always closes underlying fd.

criu/net.c
mihalicyn: my bad :(

criu/net.c
mihalicyn: yeah. You are right. We need to think about it. Which checks we need to do in that place to decide - dump nft or not. Current check with `check_cmd_exists("nft")` not best solution.

criu/Makefile
```diff
@@ -30,6 +30,8 @@ CFLAGS			+= -iquote .
 CFLAGS			+= $(shell pkg-config --cflags libnl-3.0)
 CFLAGS			+= $(CONFIG-DEFINES)
 
+LDFLAGS			+= -lnftables
```
0x7f454c46: nit:
```
LDFLAGS += $(shell pkg-config --libs libnftables)
```

mihalicyn: thanks! ;)

mihalicyn: > nit:
> 
> ```
> LDFLAGS += $(shell pkg-config --libs libnftables)
> ```

thanks ;)

mihalicyn:
Guys, time for hard questions :)

This patchset requires libnftables at least 0.9.1. Becase between 0.9.0 -> 0.9.1 API has been changed (`nft_run_cmd_from_buffer()` function now have 2 args, was 3 args).
I hope that in future our friends from netfilter team will not change the API of libnftables. :)

The question is: libnftables 0.9.1 is too fresh. In CentOS 8, stable Ubuntu, etc we have no prebuilt package with this version. What to do? I have several ideas:
1. Add nftables as a git submodule to CRIU and build with CRIU.
2. Simply add checks to CRIU about libnftables -> that libnftables present, have compatible API and so on. And require a user to build the corresponding version of libnftables. Also, I could write some bash script to simplify that process (but it's simple as `./configure && make && make install`).

Regards, Alex

mihalicyn:
I've pushed a new version with additional checks in Makefile - in this implementation, we check that libnftables is present and also try to compile a program that uses specific API. I think that it's better than hard-coding checks based on the concrete library version.

0x7f454c46:
Hi @mihalicyn, if you're going with makefile checks, make sure that if `libnftables.so.1` is not present in build-time, CRIU gets built without nftables support instead of failing the build.
On the other hand, I've tried `dlopen()` approach and it doesn't seems to be very ugly:
https://lists.openvz.org/pipermail/criu/2019-November/044844.html
I have rebased your branch on those patches in case you won't to look:
https://github.com/0x7f454c46/criu/commits/wip/nftables-v1
It compiles, not sure if it works ;-)
I'm not very proud of those boilerplate functions with `shared_libs_lookup_once()`, but it may be cleaned up on top with a single-line macro I think..

mihalicyn:
Hi @0x7f454c46! Looks great! To my shame, I didn't know that by the suffix in name of the library we can determine that API has been changed. (If I understand correctly if the developers of netfilter change the API again, the library will be called libnftables.so.2, and so on)
Thank you!
Let's try that approach!

avagin:
@mihalicyn I don't understand what the status of the idea with dlopen. Could you implement a version when on compile time we will check whether a right version of this library installed of not. And if it is installed, criu will be compiled with the support of dumping/restoring nftables.

mihalicyn:
@avagin yes, of course! The current version on code already has compile-time checks about libnftables, but it supports only the last version of libnftables. As we discussed before, we could implement the support of previous libnftables API without problems. I will do that in a few days.
Thank you.

mihalicyn:
Warning: Currently, CentOS not ships nftables-devel package as Fedora do. I've removed installing nft into our test docker containers to make nft tests skipped in travis.

mihalicyn:
Something wrong with out PowerPC64 LE tests, and Java tests also...

adrianreber:
> Something wrong with our PowerPC64 LE tests, and Java tests also...

Yes, I have also seen the ppc64le errors. As the tests are not even starting I would blame Travis. Seems like their setup is not working. Java tests are known to not work due to a Ubuntu kernel bug. Workaround is available at: #891

mihalicyn:
@adrianreber couldn't you review this patchset? I would be very glad to receive your comments.

Makefile.config
```diff
@@ -23,6 +23,23 @@ else
         $(info Note: Building without GnuTLS support)
 endif
 
+ifeq (x$(call pkg-config-check,libnftables),xy)
```
Snorch: Sorry in advance as I have small experience in makefiles, but why not just ?
```ifeq ($(call pkg-config-check,libnftables),y)```

File: Makefile.config [hunk 7]
```diff
@@ -23,6 +23,23 @@ else
...
+	LIB_NFTABLES	:= $(shell pkg-config --libs libnftables)
+        ifeq ($(call try-cc,$(FEATURE_TEST_NFTABLES_LIB_API_0),$(LIB_NFTABLES)),true)
```
Snorch: Tabs vs spaces in the same hunk.

criu/net.c
```diff
@@ -1897,6 +1901,55 @@ static inline int dump_iptables(struct cr_imgset *fds)
...
+		close(img_fd);
+		goto nft_ctx_free_out;
+	}
+
+	nft_ctx_set_output(nft, fp);
+
+#if defined(CONFIG_HAS_NFTABLES_LIB_API_0)
+	if (nft_run_cmd_from_buffer(nft, "list ruleset", strlen("list ruleset")))
+#elif defined(CONFIG_HAS_NFTABLES_LIB_API_1)
+	if (nft_run_cmd_from_buffer(nft, "list ruleset"))
```
Snorch: "list ruleset" x3 times - we may add a define for it

Makefile.config
mihalicyn: As far as I know, very old shells did not handle empty strings correctly in comparisons. Nowadays it's just a tradition. :)

Snorch: On line 13 we have
```ifeq ($(call pkg-config-check,libselinux),y)```
with no problem somehow, why?

mihalicyn: I've fixed that to avoid meaningless holywar. :)

mihalicyn: I agree, that if we use `ifeq` without `x` in upper code, we should do that uniformly in all codebase.

0x7f454c46:
Applied, thanks!
I've added you @mihalicyn to `test_author` in zdtm test on merge.
P.S.
Sorry for diverting the discussion to dlopen() idea - I thought it could help you with the ABI problem.

PR #878, opened 2019-12-12
sk-netlink:  Handle case of in-use netlink peer ids

Hello1024:
Netlink sockets normally kernel-assign a peer id based on the requesting pid at bind time. CRIU saves this peer ID and tries to restore it.   At restore time, it's possible the id is already in use (assigned to another process, perhaps in another pid namespace).

This patch tries to use the original id, but if that fails then allows the kernel to auto-allocate a new peer id.   We make a warning message, because it is possible (although rather unlikely) the application cares what its peer id is.


avagin:
@xemul  what do you think about this?

xemul:
Just in case :) I think it makes sense, but as the pr is heavily outdated, let's close one.

PR #879, opened 2019-12-12
net:  Add ability to inherit file descriptors of TCP sockets

Hello1024:
Many other types of sockets work with `--inherit-fd`.  This extends this ability to inet sockets (UDP, TCP, listening sockets, and everything else INET).


avagin:
@Hello1024 could you add a test for this? All other tests for --inherit-fd are in test/inhfd/.

xemul:
Outdated. Please, resubmit of still relevant

PR #882, opened 2019-12-16
Restore processes using clone3() with set_tid

adrianreber:
This is my first attempt to use the newly introduced clone3() with set_tid to replace CRIU's use of ns_last_pid.

My current test case is a simple threaded process which can be restored using the changes in this PR.

This is still work in progress and a request for comments. I am not sure if I did the whole child handling correct so please have close look at these changes. I am especially unsure about the child stack setup.

This is currently still x86_64 only and once this is working I would start looking at the other architectures.

One very nice result of clone3() with set_tid is that by no longer using ns_last_pid, newly created processes after running CRIU are using different PIDs and PID collisions during testing by restoring the same process again and again does not happen any more.

### DIFF
--- a/criu/clone-noasan.c
+++ b/criu/clone-noasan.c
@@ -1,4 +1,10 @@
+#include <stdlib.h>
 #include <sched.h>
+#include <unistd.h>
+
+#include <compel/plugins/std/syscall-codes.h>
+
+#include "sched.h"
 #include "common/compiler.h"
 #include "log.h"
 #include "common/bug.h"
@@ -31,10 +37,36 @@
 int clone_noasan(int (*fn)(void *), int flags, void *arg)
 {
 	void *stack_ptr = (void *)round_down((unsigned long)&stack_ptr - 1024, 16);
+
 	BUG_ON((flags & CLONE_VM) && !(flags & CLONE_VFORK));
 	/*
 	 * Reserve some bytes for clone() internal needs
 	 * and use as stack the address above this area.
 	 */
 	return clone(fn, stack_ptr, flags, arg);
+}
+
+int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags,
+			   int exit_signal, pid_t pid)
+{
+	struct _clone_args c_args = {};
+
+	BUG_ON(flags & CLONE_VM);
+
+	/*
+	 * Make sure no child signals are requested. clone3() uses
+	 * exit_signal for that.
+	 */
+	BUG_ON(flags & 0xff);
+
+	pr_debug("Creating process using clone3()\n");
+
+	c_args.exit_signal = exit_signal;
+	c_args.flags = flags;
+	c_args.set_tid = ptr_to_u64(&pid);
+	c_args.set_tid_size = 1;
+	pid = syscall(__NR_clone3, &c_args, sizeof(c_args));
+	if (pid == 0)
+		exit(fn(arg));
+	return pid;
 }
--- a/criu/include/rst_info.h
+++ b/criu/include/rst_info.h
@@ -4,6 +4,7 @@
 #include "common/lock.h"
 #include "common/list.h"
 #include "vma.h"
+#include "kerndat.h"
 
 struct task_entries {
 	int nr_threads, nr_tasks, nr_helpers;
--- a/criu/include/sched.h
+++ b/criu/include/sched.h
@@ -1,0 +1,33 @@
+#ifndef __CR_SCHED_H__
+#define __CR_SCHED_H__
+
+#include <linux/types.h>
+
+#ifndef ptr_to_u64
+#define ptr_to_u64(ptr)	((__u64)((uintptr_t)(ptr)))
+#endif
+#ifndef u64_to_ptr
+#define u64_to_ptr(x)	((void *)(uintptr_t)x)
+#endif
+
+/*
+ * This structure is needed by clone3(). The kernel
+ * calls it 'struct clone_args'. As CRIU will always
+ * need at least this part of the structure (VER1)
+ * to be able to test if clone3() with set_tid works,
+ * the structure is defined here as 'struct _clone_args'.
+ */
+
+struct _clone_args {
+	__aligned_u64 flags;
+	__aligned_u64 pidfd;
+	__aligned_u64 child_tid;
+	__aligned_u64 parent_tid;
+	__aligned_u64 exit_signal;
+	__aligned_u64 stack;
+	__aligned_u64 stack_size;
+	__aligned_u64 tls;
+	__aligned_u64 set_tid;
+	__aligned_u64 set_tid_size;
+};
+#endif /* __CR_SCHED_H__ */
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -41,6 +41,7 @@
 #include "uffd.h"
 #include "vdso.h"
 #include "kcmp.h"
+#include "sched.h"
 
 struct kerndat_s kdat = {
 };
@@ -986,6 +987,44 @@ static int kerndat_tun_netns(void)
 	return check_tun_netns_cr(&kdat.tun_ns);
 }
 
+static bool kerndat_has_clone3_set_tid(void)
+{
+	pid_t pid;
+	struct _clone_args args = {};
+
+#ifndef CONFIG_X86_64
+	/*
+	 * Currently the CRIU PIE assembler clone3() wrapper is
+	 * only implemented for X86_64.
+	 */
+	kdat.has_clone3_set_tid = false;
+	return 0;
+#endif
+
+	args.set_tid = -1;
+	/*
+	 * On a system without clone3() this will return ENOSYS.
+	 * On a system with clone3() but without set_tid this
+	 * will return E2BIG.
+	 * On a system with clone3() and set_tid it will return
+	 * EINVAL.
+	 */
+	pid = syscall(__NR_clone3, &args, sizeof(args));
+
+	if (pid == -1 && (errno == ENOSYS || errno == E2BIG)) {
+		kdat.has_clone3_set_tid = false;
+		return 0;
+	}
+	if (pid == -1 && errno == EINVAL) {
+		kdat.has_clone3_set_tid = true;
+	} else {
+		pr_perror("Unexpected error from clone3\n");
+		return -1;
+	}
+
+	return 0;
+}
+
 int kerndat_init(void)
 {
 	int ret;
@@ -1059,6 +1098,8 @@ int kerndat_init(void)
 		ret = has_kcmp_epoll_tfd();
 	if (!ret)
 		ret = kerndat_has_fsopen();
+	if (!ret)
+		ret = kerndat_has_clone3_set_tid();
 
 	kerndat_lsm();
 	kerndat_mmap_min_addr();
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -35,6 +35,7 @@
 #include "sk-inet.h"
 #include "vma.h"
 #include "uffd.h"
+#include "sched.h"
 
 #include "common/lock.h"
 #include "common/page.h"
@@ -1771,16 +1772,19 @@ long __export_restore_task(struct task_restore_args *args)
 		long clone_flags = CLONE_VM | CLONE_FILES | CLONE_SIGHAND	|
 				   CLONE_THREAD | CLONE_SYSVSEM | CLONE_FS;
 		long last_pid_len;
+		pid_t thread_pid;
 		long parent_tid;
 		int i, fd = -1;
 
-		/* One level pid ns hierarhy */
-		fd = sys_openat(args->proc_fd, LAST_PID_PATH, O_RDWR, 0);
-		if (fd < 0) {
-			pr_err("can't open last pid fd %d\n", fd);
-			goto core_restore_end;
-		}
+		if (!args->has_clone3_set_tid) {
+			/* One level pid ns hierarhy */
+			fd = sys_openat(args->proc_fd, LAST_PID_PATH, O_RDWR, 0);
+			if (fd < 0) {
+				pr_err("can't open last pid fd %d\n", fd);
+				goto core_restore_end;
+			}
 
+		}
 		mutex_lock(&task_entries_local->last_pid_mutex);
 
 		for (i = 0; i < args->nr_threads; i++) {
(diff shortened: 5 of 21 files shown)


criu/include/sched.h
```diff
@@ -0,0 +1,34 @@
+#ifndef __CR_SCHED_H__
+#define __CR_SCHED_H__
+
+#include <linux/types.h>
```
brauner: Don't you need to include
`linux/sched.h`?

criu/include/sched.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef u64_to_ptr
+#define u64_to_ptr(x)	((void *)(uintptr_t)x)
+#endif
+
+#ifndef CLONE_ARGS_SIZE_VER0
+#define CLONE_ARGS_SIZE_VER0
+
+/* This structure is needed by clone3(). */
+
+struct _clone_args {
```
brauner: And why is this called `_clone_args` and not `_clone_args`? :)

criu/include/sched.h
```diff
@@ -0,0 +1,34 @@
...
+#include <linux/types.h>
+
+#ifndef ptr_to_u64
+#define ptr_to_u64(ptr)	((__u64)((uintptr_t)(ptr)))
+#endif
+#ifndef u64_to_ptr
+#define u64_to_ptr(x)	((void *)(uintptr_t)x)
+#endif
+
+#ifndef CLONE_ARGS_SIZE_VER0
```
brauner: There also is no size attached to this. Is that on purpose?

adrianreber: Looking at it now (one week later) it seems the whole file was started with other ideas than how I am using it right now.

The reason to call it `_clone_args` is to not collide with the kernel struct from the header (which is not included (which is wrong)). I always want to have a struct with the set_tid member to be able to test if it works. I am using the structure size defines (CLONE_ARGS_SIZE_VER1) mainly to check if the structure is defined at all.

criu/include/sched.h
brauner:
```c
#ifndef CLONE_ARGS_SIZE_VER0
#define CLONE_ARGS_SIZE_VER0 64
struct clone_args {
	__aligned_u64 flags;
	__aligned_u64 pidfd;
	__aligned_u64 child_tid;
	__aligned_u64 parent_tid;
	__aligned_u64 exit_signal;
	__aligned_u64 stack;
	__aligned_u64 stack_size;
	__aligned_u64 tls;
#define CLONE_ARGS_SIZE_VER1 80
	__aligned_u64 set_tid;
	__aligned_u64 set_tid_size;
};
#endif /* CLONE_ARGS_SIZE_VER0 */
```
I think the ifdef for CLONE_ARGS_SIZE_VER1 is not needed. A system where `VER0` is defined but not `VER1` is broken.

adrianreber:
This breaks CI much more than expected. Need to fix this and also the include file problems which @brauner mentioned.

criu/include/sched.h
brauner: The way I define it in kernel tests is:
```c
#ifndef CLONE_ARGS_SIZE_VER0
#define CLONE_ARGS_SIZE_VER0 64
#endif

#ifndef __NR_clone3
#define __NR_clone3 -1
struct clone_args {
	__aligned_u64 flags;
	__aligned_u64 pidfd;
	__aligned_u64 child_tid;
	__aligned_u64 parent_tid;
	__aligned_u64 exit_signal;
	__aligned_u64 stack;
	__aligned_u64 stack_size;
	__aligned_u64 tls;
#define CLONE_ARGS_SIZE_VER1 80
	__aligned_u64 set_tid;
	__aligned_u64 set_tid_size;
#define CLONE_ARGS_SIZE_VER2 88
	__aligned_u64 cgroup;
};
#endif /* __NR_clone3 */
```
`CLONE_ARGS_SIZE_VER0` is only exported starting with kernel `v5.4`. So the reliable test for whether or not `clone3()` is available at all should be to look at `__NR_clone3`.

adrianreber:
Travis results are starting to look better, but still failures with clang on x86_64 and non-clang builds on the other arches. Interesting. Will update to get it fixed.

criu/kerndat.c
```diff
@@ -986,6 +987,40 @@ static int kerndat_tun_netns(void)
 	return check_tun_netns_cr(&kdat.tun_ns);
 }
 
+static bool kerndat_has_clone3_set_tid(void)
+{
+	pid_t pid;
+	struct _clone_args args = {};
+
+#ifndef CONFIG_X86_64
+	kdat.has_clone3_set_tid = false;
```
avagin: is clone3_set_tid supported by other architectures?

criu/kerndat.c
```diff
@@ -986,6 +987,40 @@ static int kerndat_tun_netns(void)
...
+
+	args.set_tid = -1;
+	/*
+	 * On a system without clone3() this will return ENOSYS.
+	 * On a system with clone3() but without set_tid this
+	 * will return E2BIG.
+	 * On a system with clone3() and set_tid it will return
+	 * EINVAL.
+	 */
+	pid = syscall(SYS_clone3, &args, sizeof(args));
```
avagin: You define clone3 in plugins/std/syscalls, in this case you need to use __NR_clone3 instead of SYS_clone3.

And you will need to include  "compel/plugins/std/syscall-codes.h"

criu/kerndat.c
```diff
@@ -986,6 +987,40 @@ static int kerndat_tun_netns(void)
...
+
+	if (pid == -1 && (errno == ENOSYS || errno == E2BIG)) {
+		kdat.has_clone3_set_tid = false;
+		return 0;
+	}
+	if (pid == -1 && errno == EINVAL) {
+		kdat.has_clone3_set_tid = true;
+	} else {
+		pr_err("Unexpected error from clone3\n");
```
avagin: pr_perror

criu/clone-noasan.c
```diff
@@ -27,14 +29,80 @@
...
+	 * The stack_size seems to be pretty big and in the mmap() it is
+	 * even twice the size. By telling the kernel the stack is
+	 * STACK_SIZE and mmap()ing it twice the size the restore actually
+	 * works again. Any different combination did not work. This
+	 * is not understood.
+	 */
+
+#define STACK_SIZE 1024 * 1024 * 32
+
+	stack_ptr = mmap(NULL, STACK_SIZE * 2, PROT_READ | PROT_WRITE,
```
avagin: we don't need to allocate a stack here, we can use the current one by the same way how it works for libc clone.

criu/include/rst_info.h
```diff
@@ -76,11 +77,16 @@ extern struct task_entries *task_entries;
 
 static inline void lock_last_pid(void)
 {
+	/* No locking required if clone3() is available */
+	if (kdat.has_clone3_set_tid)
```
avagin: We can't do this. last_pid_mutex guarantees that we will not restore threads when there are helper processes (look at call_helper_process), because a helper process can use one of thread pid-s.

criu/pie/restorer.c
```diff
@@ -1771,17 +1772,20 @@ long __export_restore_task(struct task_restore_args *args)
...
+		if (!args->has_clone3_set_tid) {
+			/* One level pid ns hierarhy */
+			fd = sys_openat(args->proc_fd, LAST_PID_PATH, O_RDWR, 0);
+			if (fd < 0) {
+				pr_err("can't open last pid fd %d\n", fd);
+				goto core_restore_end;
+			}
 
-		mutex_lock(&task_entries_local->last_pid_mutex);
+			mutex_lock(&task_entries_local->last_pid_mutex);
```
avagin: we always need to hold the last_pid_mutex when we are forking threads, otherwise there may be a helper process which is using the thread pid.

criu/kerndat.c
adrianreber: It is supported on all CRIU architectures, but it needs an assembler wrapper which I currently only have implemented for x86_64. Once I have the assembler wrappers for all architectures, this will be removed.

adrianreber:
@avagin about the last_pid_mutex. I assumed we do not need to lock it if we are using clone3(). Without clone3() we had the possibility of writing to `ns_last_pid` and having a helper process spawned before creating the actual process or thread we want to restore. With clone3() we do not rely on `ns_last_pid` and therefore we avoid the race condition of accidentally creating a process with the PID we need for restore. With clone3() we do not change the PID allocation of other processes/threads as we used to when using `ns_last_pid`. Based on this I thought we do not need the locking, but maybe I misunderstood something.

Thinking more about it, if we are restoring in a PID namespace then the chances are high that all PIDs in that PID namespace are most likely close together and also close to 1, if we are talking about a container with only a few processes. If we spawn a helper process inside of that PID namespace it it much more likely that such a helper process will use one of the PIDs we need soon. So maybe it is important to have the locking if the helper processes are running in the destination PID namespace.

I assumed that if we are not writing to `ns_last_pid` we do not need to lock it to avoid the race condition of some other process taking away the PID we just selected via `ns_last_pid` as clone3() does not have to deal with this race condition.

avagin:
> I assumed that if we are not writing to ns_last_pid we do not need to lock it to avoid the race condition of some other process taking away the PID we just selected via ns_last_pid as clone3() does not have to deal with this race condition.

I don't understand why it does not have to deal with this race condition. Look at this
1. A: fork a helper process and this process gets the PID 23.
2. B: restores a process with the PID 23 and fails because the target PID 23 is already used
3. A helper process has been exited and released the PID 23.

Right now, last_pid_mutex guarantees that stem 2 will be executed before the first step or after the third step.

criu/clone-noasan.c
avagin: Actually, when we are forking a new process with a specified pid, clone_flags can't contain CLONE_VM, this means that you can add clone_with_pid_noasan:

```
int clone_with_pid_noasan(int (*fn)(void *), int flags, void *arg, pid_t pid) {
  struct _clone_args c_args = {};
  pid_t pid;

  BUG_ON(flags & CLONE_VM);

  c_args.flags = flags;
  c_args.set_tid = ptr_to_u64(&pid);
  c_args.set_tid_size = 1;
  pid = syscall(__NR_clone3, &c_args, sizeof(c_args);
  if (pid == 0) {
    exit(fn(args));
  }
  return pid;
}
```

adrianreber:
> > I assumed that if we are not writing to ns_last_pid we do not need to lock it to avoid the race condition of some other process taking away the PID we just selected via ns_last_pid as clone3() does not have to deal with this race condition.
> 
> I don't understand why it does not have to deal with this race condition. Look at this
> 
>     1. A: fork a helper process and this process gets the PID 23.
> 
>     2. B: restores a process with the PID 23 and fails because the target PID 23 is already used
> 
>     3. A helper process has been exited and released the PID 23.
> 
> 
> Right now, last_pid_mutex guarantees that stem 2 will be executed before the first step or after the third step.

Okay, I will add the locking back also for clone3().

criu/clone-noasan.c
adrianreber: Unfortunately the restored process crashes if using the stack the same way as with the libc clone().

criu/clone-noasan.c
```diff
@@ -27,14 +29,80 @@
...
+	 * as the stack pointer for the newly created process.
+	 */
+
+	/*
+	 * The stack_size seems to be pretty big and in the mmap() it is
+	 * even twice the size. By telling the kernel the stack is
+	 * STACK_SIZE and mmap()ing it twice the size the restore actually
+	 * works again. Any different combination did not work. This
+	 * is not understood.
+	 */
```
brauner: That's a worrying observation. No clue by what this is caused?

criu/clone-noasan.c
adrianreber: The restored process also crashes if using your clone_with_pid_noasan() example instead of my own assembler wrapper. Your code looks so much simpler than my assembler wrapper, but it does not work.

The interesting thing about the crash is, that not the process that is created with your code crashes, but the second thread of my test process segfaults when the first is restored with your code.

If I change `c_args.stack_size` in the thread restorer (pie/restorer.c) to `RESTORE_STACK_SIZE / 2` the restore works most of the time, but sometimes the restored thread segfaults.

criu/clone-noasan.c
adrianreber: Right now most of the stack setup, which was working with the old clone implementation (glibc wrapper and own assembler wrapper), is not understood. With the added parameter `stack_size` the whole thing has become more complicated and I do not yet understand why it works and why it crashes.

brauner: Why did you call the glibc clone but used your own assembler wrapper? That's why glibc clone exists in the first place to not have to do this yourself. 
The stack size was always required even with old clone just implicitly.

adrianreber: > Why did you call the glibc clone but used your own assembler wrapper?

I do not understand this. Something is missing in this sentence.

> The stack size was always required even with old clone just implicitly.

It was never a parameter passed to the kernel and it seems the value is not really used in the kernel besides adding it to the stack address.

brauner: > > Why did you call the glibc clone but used your own assembler wrapper?
> 
> I do not understand this. Something is missing in this sentence.

You said

> Right now most of the stack setup, which was working with the old clone implementation (glibc wrapper and own assembler wrapper), 

That sounds like you used glibc's clone() but still used an assembly wrapper. That sounds strange.

> 
> > The stack size was always required even with old clone just implicitly.
> 
> It was never a parameter passed to the kernel and it seems the value is not really used in the kernel besides adding it to the stack address.

Actually it was for some architectures but that's not the main point. The point is that in both cases the stack_size did matter because of stack direction.

criu/clone-noasan.c
brauner: Can you show that code you're testing this with, please?

criu/clone-noasan.c
```diff
@@ -27,14 +29,80 @@
...
  *	... wait for process to finish ...
  *	unlock_last_pid
+ *
+ * This locking is not required if clone3() is available. lock_last_pid()
+ * and unlock_last_pid() will be noops if clone3() is available.
  */
-int clone_noasan(int (*fn)(void *), int flags, void *arg)
+int clone_noasan(int (*fn)(void *), int flags, void *arg, pid_t pid)
 {
 	void *stack_ptr = (void *)round_down((unsigned long)&stack_ptr - 1024, 16);
```
brauner: Also, what is this?

rst0git: from https://github.com/checkpoint-restore/criu/commit/bcdd276003 :
```
 criu/clone: stack size is too small for Android

stack for clone is too small, child process will get
wild pointer, and segfault.
```
and from https://github.com/checkpoint-restore/criu/commit/ef65d98 :
```
clone_noasan: Allow to create CLONE_VM|CLONE_VFORK processes 
...

Currently parent touches child's stack, as in moment of clone() call
its stack pointer is above the child's (we allocate char stack[128]
on parent's stack). This prevents to create CLONE_VM|CLONE_VFORK
processes, because the child uses stack addresses occupied by parent.

The patch changes clone_noasan() behaviour and allows to do that
with the same memory consumption. We give a child memory, which
is not used by parent clone(), so parent's and child's stacks
have no intersection.
```

criu/clone-noasan.c
adrianreber: Right now I guess I am doing something wrong when setting up the stack in the thread (see criu/pie/restorer.c). I need to look closer at what I might be doing wrong there.

Thanks for all the feedback!

adrianreber: If I subtract the stack_size of the new stack pointer for restored threads before calling clone3() the restored process never crashes. Maybe that was already everything that needed to be changed. I will test it some more and then update the PR with my new code. Using Andrei's assembler free clone3() example.

criu/clone-noasan.c
avagin: @adrianreber I think something wrong in RUN_CLONE3_RESTORE_FN. I don't try to find what wrong in your code, I took RUN_CLONE_RESTORE_FN and modified it to use clone3. Here is my version and it works on my laptop:
https://github.com/avagin/criu/commit/2c9dc9d93ba3dc5a500138f1a3e3783269779fcb

Could you take a look at it?

adrianreber: @avagin You are not setting the stack_size anywhere in your code. The kernel does:
```
        if (kargs->stack == 0) {
                if (kargs->stack_size > 0)
                        return false;
        } else {
                if (kargs->stack_size == 0)
                        return false;

                if (!access_ok((void __user *)kargs->stack, kargs->stack_size))
                        return false;

#if !defined(CONFIG_STACK_GROWSUP) && !defined(CONFIG_IA64)
                kargs->stack += kargs->stack_size;
#endif
        }
```
You should not be able to create a process using clone3() as you are setting stack to newsp but stack_size is 0.

brauner: Depends on what kernel @avagin is on, but yes.

criu/kerndat.c
adrianreber: adding a comment to make it easier to understand

adrianreber:
I tried to address all open points. Mainly using Andrei's clone_noasan example without assembler and adding the locking back. Thanks so far for all comments and feedback.

adrianreber:
CI failure is related to the known broken test running in a docker container.

criu/kerndat.c
```diff
@@ -986,6 +987,44 @@ static int kerndat_tun_netns(void)
...
 }
 
+static bool kerndat_has_clone3_set_tid(void)
+{
+	pid_t pid;
+	struct _clone_args args = {};
+
+#ifndef CONFIG_X86_64
+	/*
+	 * Currently the CRIU PIE assembler clone3() wrapper is
```
avagin: pls file an issue to implement clone3() for other architectures

adrianreber:
Tracker bug to implement clone3 based assembler wrapper for other architectures : #890

avagin:
Is it still RFC or is it ready to be merged?

adrianreber:
Running zdtm on this branch gives me:
```
################### 2 TEST(S) FAILED (TOTAL 371/SKIPPED 29) ####################
 * zdtm/transition/fifo_loop(ns)
 * zdtm/static/netns-nft(uns)
```
The NFT test always fails with out without my patches and the fifo_loop test also fails sometimes independent of my patches. Removing the RFC flag. Ready to be merged from my point of view.

avagin:
@mihalicyn could you investigate what wrong with zdtm/static/netns-nft

adrianreber:
I see the following error on Fedora 31 with 5.5.0-rc1 kernel:
```
[root@fedora01 test]# ./zdtm.py run  -t  zdtm/static/netns-nft
=== Run 1/1 ================ zdtm/static/netns-nft
======================= Run zdtm/static/netns-nft in ns ========================
Start test
Test is SUID
./netns-nft --pidfile=netns-nft.pid --outfile=netns-nft.out --filename=netns-nft.test
Run criu dump
Run criu restore
Send the 15 signal to  89
Wait for zdtm/static/netns-nft(89) to die for 0.100000
############### Test zdtm/static/netns-nft FAIL at result check ################
Error: No such file or directory
list table inet netns-nft-zdtm-test
                ^^^^^^^^^^^^^^^^^^^
09:01:58.816:     4: FAIL: netns-nft.c:51: Can't get nft table (errno = 11 (Resource temporarily unavailable))

```

rst0git:
@adrianreber is `nftables-devel` installed? (https://github.com/checkpoint-restore/criu/issues/898)

adrianreber:
> @adrianreber is `nftables-devel` installed? (#898)

No. I was about to write 'then it should not fail', but my selinux tests are in the same camp. If the selinux runtime is installed and CRIU is built without libselinux-devel I get the same error. Installing the devel package fixes this indeed. Thanks.

0x7f454c46:
The patches LGTM.
I wish there were no assembly to clone in restorer, but maybe it's really necessary.
Anyway, assembly in the PR has good comments and I can experiment separately if something could be done to remove it.

PR #883, opened 2019-12-17
x86/cpu: cleanup and improve xfeatures_mask check

Snorch:
Make xfeatures_mask check explicit.

Split up comments about xsave_size and xfeatures_mask because we have
different approaches for them:

1) for xfeatures we need to have on destination all the features which we
had on source

2) for xsave_size we need to have smaller size on destination than on
source, because xsave operation on small allocated buffer may corrupt the
nearby data

Having single comment for both cases is less understandable.

### DIFF
--- a/criu/arch/x86/cpu.c
+++ b/criu/arch/x86/cpu.c
@@ -236,6 +236,7 @@ static int cpu_validate_features(compel_cpuinfo_t *cpu_info)
 		return -1;
 
 	if (opts.cpu_cap & CPU_CAP_FPU) {
+		uint64_t m;
 		/*
 		 * If we're requested to check FPU only ignore
 		 * any other bit. It's up to a user if the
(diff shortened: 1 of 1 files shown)


Snorch:
@cyrillos please take a look

criu/arch/x86/cpu.c
```diff
@@ -261,25 +262,37 @@ static int cpu_validate_features(compel_cpuinfo_t *cpu_info)
...
-		 * Make sure the xsave features are compatible. We already hit the
-		 * issue with libc where we've checkpointed the container on old
-		 * machine but restored on more modern one and libc fetched new
-		 * xsave frame size directly by xsave instruction with greedy
-		 * feature mask causing programs to misbehave.
+		 * Make sure the xsave features are compatible. We were relying
+		 * on our guess about hardware "backward compatibility" and
+		 * used ">" check here for a long time. But it looks better to
+		 * explicitly check that all xfeature bits from source are
+		 * present on the destination.
```
0x7f454c46: I'm not sure anybody would understand ">" here in 2021..
Maybe instead:
```
		 * Make sure the xsave features are compatible. Check that on destination
		 * there are all the features which were on the source.
```
Hmm?

criu/arch/x86/cpu.c
```diff
@@ -261,25 +262,37 @@ static int cpu_validate_features(compel_cpuinfo_t *cpu_info)
...
 		 */
-		if (cpu_info->xfeatures_mask > rt_cpu_info.xfeatures_mask) {
-			uint64_t m = cpu_info->xfeatures_mask & ~rt_cpu_info.xfeatures_mask;
+		if ((m = cpu_info->xfeatures_mask & ~rt_cpu_info.xfeatures_mask)) {
 			pr_err("CPU xfeatures has unsupported bits (%#llx)\n",
 			       (unsigned long long)m);
```
0x7f454c46: While at here, is it possible to print it with `"...%#"PRIx64")\n"`, removing needless cast?

criu/arch/x86/cpu.c
0x7f454c46: actually, "on a destination" and "on a source" (per my limited knowledge of ingles)

criu/arch/x86/cpu.c
```diff
@@ -261,25 +262,37 @@ static int cpu_validate_features(compel_cpuinfo_t *cpu_info)
...
 			return -1;
-		} else if (cpu_info->xsave_size != rt_cpu_info.xsave_size) {
-			pr_err("CPU xsave size mismatch (%u/%u)\n",
-			       cpu_info->xsave_size, rt_cpu_info.xsave_size);
-			return -1;
-		} else if (cpu_info->xsave_size_max != rt_cpu_info.xsave_size_max) {
-			pr_err("CPU xsave max size mismatch (%u/%u)\n",
-			       cpu_info->xsave_size_max, rt_cpu_info.xsave_size_max);
-			return -1;
+		} else {
```
rst0git: What is the purpose of `else` here (and `else if` below)? The `return -1` statement would prevent execution of subsequent statements, right?

criu/arch/x86/cpu.c
Snorch: Makes sense, but for me it sounds better to use definite article here, because we only have one destination and one source and they seam defined for me.

criu/arch/x86/cpu.c
Snorch: done.

criu/arch/x86/cpu.c
Snorch: Thanks for a good hint, removed extra else statements.

PR #884, opened 2019-12-19
files: fix ghost file error path

nviennot:
The error path is incorrectly leaving the ghost entry in the list


0x7f454c46:
Merged, thanks!

PR #886, opened 2019-12-19
Add memfd support

nviennot:
The following commits add support for memfd.

It reuses the ghost file logic to dump the content of the memfd into the image.


criu/util.c
```diff
@@ -470,6 +472,193 @@ int copy_file(int fd_in, int fd_out, size_t bytes)
...
+		buf = xmalloc(BUFSIZE);
+		if (!buf)
+			return -1;
+	}
+
+	while (len > 0) {
+		if (opts.remote) {
+			ret = pread(fd, buf, min_t(size_t, BUFSIZE, len), off);
+			if (ret <= 0) {
+				pr_perror("Can't read from ghost file");
```
mihalicyn: nits: I think that if we are generalizing this function by moving to `util.c` it would be great if we change error messages to more common without mentioning `ghost` files. What do you think about it? We could also add additional `pr_err` to calling code to make logs more informative if needed.

criu/util.c
```diff
@@ -470,6 +472,193 @@ int copy_file(int fd_in, int fd_out, size_t bytes)
...
+			if (write(img, buf, ret) != ret) {
+				pr_perror("Can't write to image");
+				xfree(buf);
+				return -1;
+			}
+			off += ret;
+		} else {
+			ret = sendfile(img, fd, &off, len);
+			if (ret <= 0) {
+				pr_perror("Can't send ghost to image");
```
mihalicyn: comment above

mihalicyn:
@nviennot 
great work! Thank you!

Please, take a look on CI test results. There are several compile-time issues.

Regards, Alex

criu/util.c
nviennot: I agree. I amended the commit.

I don't think we need an additional `pr_err` as the current error messages will immediately follow `pr_info("Dumping ghost file contents")` in the logs

criu/proc_parse.c
```diff
@@ -303,6 +304,23 @@ static int vma_get_mapfile_user(const char *fname, struct vma_area *vma,
 	}
 
 	vfi_dev = makedev(vfi->dev_maj, vfi->dev_min);
+
+	if (is_memfd(vfi_dev, fname)) {
+		struct fd_link link;
+		link.len = strlen(fname);
+		strcpy(link.name, fname);
```
mihalicyn: I think that it is safer to use `strlcpy` here. Sorry, I'm have a little paranoia. ;)
`strcpy(link.name, fname, sizeof(link.name));`

nviennot: Yes, I agree, fixed

nviennot:
Thank you for the appreciation, and thank you for reviewing :)

I've fixed the CI issues:
- `memfd_create()` is sometimes defined by libc, sometimes not. To avoid this issue, we use a config flag `CONFIG_HAS_MEMFD_CREATE` and implement our own syscall if needed.
For tests, we prefix the `memfd_create` syscall with `_`.
- In compat mode (32bits apps), `sys_open()` does not add a `O_LARGEFILE` flag to files, but in 64bits the flag is added forcefully. I don't think it's worth trying to restore the flag correctly. But I fixed one of the memfd test to avoid testing for it.

See here https://github.com/checkpoint-restore/criu/compare/e28943c..nviennot:memfd for the comparison between the original PR and now.

test/zdtm/static/memfd01.c
```diff
@@ -0,0 +1,112 @@
+#include <err.h>
```
avagin: pls, move the test to a separate patch.

test/zdtm/static/memfd01.c
```diff
@@ -0,0 +1,112 @@
...
+		/*
+		 * We pass O_LARGEFILE because in compat mode, our file
+		 * descriptor does not get O_LARGEFILE automatically, but the
+		 * restorer using non-compat open() is forced O_LARGEFILE.
+		 * This creates a flag difference, which we don't want to deal
+		 * with this at the moment.
+		 */
+		fd = open(buf, O_RDONLY | O_LARGEFILE);
+		if (fd < 0)
+			err(1, "Can't open memfd via proc");
```
avagin: can we use pr_perror() here and in other places

test/zdtm/static/memfd01.c
```diff
@@ -0,0 +1,112 @@
...
+	task_waiter_t t;
+	test_init(argc, argv);
+
+	task_waiter_init(&t);
+
+	int fd = _memfd_create("somename", MFD_CLOEXEC);
+	if (fd < 0)
+		err(1, "Can't call memfd_create");
+
+	pid_t pid = getpid();
```
avagin: According to the Linux kernel coding styles (we use it for criu), all variables have to be declared at the start of scope.

avagin:
@nviennot Great job! I will be on vacation for the next two weeks and then I will review this PR.

avagin:
@mclapinski if I am not mistaken, your team is interested in this functionality too.

nviennot:
@avagin, I've addressed your comments.

Enjoy your vacation :)

criu/memfd.c
```diff
@@ -0,0 +1,450 @@
...
+	 * before applying this seal.
+	 */
+
+	struct memfd_inode *inode;
+
+	list_for_each_entry(inode, &memfd_inodes, list) {
+		if (!inode->pending_seals)
+			continue;
+
+		int fd = memfd_open_inode(inode);
```
mihalicyn: Hm, I've missed that part -> as @avagin said before we always put variables declaration at the start of scope.

criu/memfd.c
```diff
@@ -0,0 +1,450 @@
...
+		if (fd < 0)
+			return -1;
+
+		int ret = fcntl(fd, F_ADD_SEALS, inode->pending_seals);
```
mihalicyn: the same here

criu/memfd.c
```diff
@@ -0,0 +1,450 @@
...
+#define F_SEAL_GROW	0x0004	/* prevent file from growing */
+#define F_SEAL_WRITE	0x0008	/* prevent writes */
+/* Linux 5.1+ */
+#define F_SEAL_FUTURE_WRITE	0x0010  /* prevent future writes while mapped */
+
+struct memfd_inode {
+	struct list_head	list;
+	u32			id;
+	union {
+		// Only for dump
```
mihalicyn: we could switch here to C-style (`/* .. */`) comments

nviennot:
@mihalicyn, I've addressed your comments.

In addition, I've opened PR #893 so that the code style is enforced by the compiler. This way we won't have to spend time on fixing variable declaration :) :)

avagin:
> It reuses the ghost file logic to dump the content of the memfd into the image.

We have the support of shared memory regions in CRIU. memfd is a shared memory file with some extra features. In this case, it should have the same format that we use for shared memory files. Does this sound reasonable?

CRIU supports iterative migrations of private memory regions and we have a task to add the support of shared memory regions. If memfd and shmem will have the same format, it will be easier to support them both together.

nviennot:
memfd behaves like a ghost file, or like a shmem region, depending on the context:
1) When a memfd file is not memory mapped, its semantics are similar to a ghost file.
2) When a memfd file is memory mapped and its file descriptor closed, its semantics are similar to a shmem region.

@avagin Overall, I think using shmem for memfd is a good idea, but it make sense to defer the implementation for another time, as it's not going to be an easy thing to do. A fair amount of work is needed from what I see when we are in scenario 1). At least, I don't see an easy way to do it.

We can also consider refactoring the ghost file code to use the shmem facilities as well for the following reason. Consider a regular ghost file that has been mmaped with `MAP_SHARED` for which its file descriptor is closed. Semantically, we are in the same scenario as 2). Memory can potentially be tracked for regular ghost files too.

It's worth noting that the current memfd implementation borrows ghost file code only from the following two functions calls:
- [`copy_fd_to_img(fd, img, mie.size, flags)`](https://github.com/checkpoint-restore/criu/pull/886/files#diff-72c517a9485da0dd6d0f6050ecc21c09R100)
- [`copy_img_to_fd(img, fd, mie->size, flags)`](https://github.com/checkpoint-restore/criu/pull/886/files#diff-72c517a9485da0dd6d0f6050ecc21c09R247)

It won't be difficult to refactor this code to use the shmem facilities in the future, should we decide to do so.

avagin:
> When a memfd file is not memory mapped, its semantics are similar to a ghost file.

You can get a file descriptor to a shared memory region by opening /proc/self/map_files/[start_addr]-[end-addr].

nviennot:
> > When a memfd file is not memory mapped, its semantics are similar to a ghost file.

> You can get a file descriptor to a shared memory region by opening /proc/self/map_files/[start_addr]-[end-addr].

Yes, but in this case, the memfd file is *not* memory mapped. When it is, that what's is happening in proc_parse.c, yes. 

> I was talking about the image format. It would be nice to have the same format for shmem and memfd. Actually, we can dump a content of a memfd file descriptor as shmem and put shmemid on memfd_inode_entry.

The image format of shmem, from my understanding, is a pair of "pages-%u" and "pagemap-shmem-%u" files. These are managed by the page_xfer facilities. So we have to reuse code. Using the page_xfer code looks pretty difficult judging by `do_dump_one_shmem`.

From what I see, the easiest for the dump a memfd is to call `do_dump_one_shmem(int fd, void *addr, struct shmem_info *si)`. The `si` struct needs only `shmid` to be set to the inode number of the file, `size` to be correct, and `pstate_map` to NULL.

In the restore path, we would use `int do_restore_shmem_content(void *addr, unsigned long size, unsigned long shmid)`.

Is this what you have in mind? If not, what would the implementation look like?

avagin:
> If not, what would the implementation look like?
Here is my quick POC:
https://github.com/avagin/criu/commit/39907eb8b6ff9404ae60b29ed6db50b15ac8fb32

nviennot:
Wonderful
I'm happy to revisit the memfd patches to reflect the shmem backing.

I think the ghost file logic extraction in util.c (util: extract ghost file copy logic) is still valuable as mentioned https://github.com/checkpoint-restore/criu/pull/886#issuecomment-567485690
I can make a separate PR for that, and rework the memfd implementation to use shmem in that way. Yes?

avagin:
> I can make a separate PR for that, and rework the memfd implementation to use shmem in that way. Yes?

Sounds good to me.

nviennot:
memfd now reuses shmem code.

If you are not liking the `err()` in the tests, let me know what you prefer.

avagin:
LGTM. Thanks!
* Could you rebase this PR on top of the new criu-dev (it has been rebase on the current master).
* Add a inhfd memfd testcase (look at test/inhfd/* and test/jenkins/criu-inhfd.sh)
* And add a bit more details into commit messages.

nviennot:
Super!

I've updated the pull request, I reworded some comments.
Let me know what you think about the inhfd test.

test/inhfd/memfd.py
```diff
@@ -0,0 +1,49 @@
...
+    cmd = "echo '{}' | gcc -x c - -o memfd_out && ./memfd_out && rm -f ./memfd_out".format(prog)
+    out = subprocess.check_output(cmd, shell=True)
+    return int(out)
+
+
+SYS_memfd_create = __get_SYS_memfd_create()
+
+
+def memfd_create(name, flags):
+    return libc.syscall(SYS_memfd_create, name.encode('utf8'), flags)
```
avagin: I think we can use the libc wrapper:
import ctypes
libc = ctypes.CDLL(None)
return libc.memfd_create(name.encode('utf8'), flags)

In this case, we will not need the C program. What do you think about this?

nviennot: Yes, very well. Thank you.

I was assuming an old libc that doesn't have `memfd_create()`.
But for the tests, it's okay to assume a recent libc.

I pushed the changes.

test/inhfd/memfd.py
```diff
@@ -0,0 +1,28 @@
+import os
+import ctypes
+libc = ctypes.CDLL(None)
+
+
+def memfd_create(name, flags):
+    return libc.memfd_create(name.encode('utf8'), flags)
+
+
+def create_fds():
```
avagin: why this test fails with this patch:
```
diff --git a/test/inhfd/memfd.py b/test/inhfd/memfd.py
index cbf6a0fb2..ddb215b7d 100755
--- a/test/inhfd/memfd.py
+++ b/test/inhfd/memfd.py
@@ -1,4 +1,5 @@
 import os
+import random
 import ctypes
 import subprocess
 libc = ctypes.CDLL(None)
@@ -25,10 +26,11 @@ SYS_memfd_create = __get_SYS_memfd_create()
 
 
 def memfd_create(name, flags):
-    return libc.syscall(SYS_memfd_create, name.encode('utf8'), flags)
+    return libc.memfd_create(name.encode('utf8'), flags)
 
 
 def create_fds():
+    idx = random.randint(0, 1024)
     def create_memfd_pair(name):
         fd = memfd_create(name, 0)
         fw = open('/proc/self/fd/{}'.format(fd), 'wb')
@@ -36,7 +38,7 @@ def create_fds():
         os.close(fd)
         return (fw, fr)
 
-    return [create_memfd_pair("name{}".format(i)) for i in range(10)]
+    return [create_memfd_pair("name{}".format(i + idx)) for i in range(10)]
 
 
 def filename(f):
```

nviennot: It works on my laptop, I'll try on travis-ci

nviennot: I added the added random int commit to another branch: https://github.com/nviennot/criu/commit/72543e4e76b24fb8adbe823b1adb7e230c9dbcc1

So far, the tests are passing: https://travis-ci.org/nviennot/criu/builds/646740596

avagin: It was my fault. I specified the --iter 10, but with this option, other tests fail too.

avagin: In travis, the centos job fails. I think you need to add memfd.py.checkskip script:
```
$ cat test/inhfd/memfd.py.checkskip 
#!/usr/bin/env python


import ctypes
libc = ctypes.CDLL(None)

libc.memfd_create("test", 0)
```

nviennot: I updated the branch. There's a failed podman travis test, but that's unrelated.
Otherwise, all the tests are green. Only centos skips the inhfd memfd test, other run it.

test/zdtm/static/Makefile
```diff
@@ -222,6 +222,7 @@ TST_NOFILE	:=				\
 		child_subreaper_and_reparent	\
 		memfd00				\
 		memfd01				\
+		memfd02				\
```
avagin: but the test is in another commit, isn't it?

nviennot: Good catch!
Fixed

Thank you for being thorough

avagin:
Applied, thanks a lot!

nviennot:
Woohoo!! ✨💃🎉
Thank you for reviewing :))

PR #888, opened 2019-12-20
files: file flags improvement

nviennot:


0x7f454c46:
@nviennot per my understanding it doesn't have any functional changes and is a clean-up, right?

nviennot:
Correct

nviennot:
Actually, it might be a bug fix. We want to avoid having files opened with `O_CLOEXEC` when restoring. Only file descriptors should be concerned with their `FD_CLOEXEC` flag

0x7f454c46:
@nviennot not sure I follow how it supposed to work after, i.e.:
```
int dump_one_reg_file(int lfd, u32 id, const struct fd_parms *p)
{
[..]
	RegFileEntry rfe = REG_FILE_ENTRY__INIT;
[..]
	rfe.flags	= p->flags;
```
and
```
static int do_open_reg_noseek(int ns_root_fd, struct reg_file_info *rfi, void *arg)
{
	return do_open_reg_noseek_flags(ns_root_fd, rfi, &rfi->rfe->flags);
}
```
I'm pretty sure I'm missing something, but it's not as trivial as I wish it to be..
I'll add reviewers and delay this for a while.
Also, there doesn't seem to be a zdtm test for O_CLOEXEC - having one would make me less concerned about breaking it. @nviennot, could you add one?

nviennot:
@0x7f454c46 good idea, I'll do this on Monday

criu/files.c
```diff
@@ -382,7 +382,13 @@ static int fill_fd_params(struct pid *owner_pid, int fd, int lfd,
...
 	p->fd		= fd;
 	p->pos		= fdinfo.pos;
-	p->flags	= fdinfo.flags;
+	/*
+	 * The kernel artificially adds the O_CLOEXEC flag on the file pointer
+	 * flags by looking at the flags on the file descriptor (see kernel
+	 * code fs/proc/fd.c). FD_CLOEXEC is a file descriptor property, which
+	 * is saved in fd_flags.
+	 */
+	p->flags	= fdinfo.flags & ~O_CLOEXEC;
```
rst0git: when creating a checkpoint of the following process `crit` shows a different result.
```C
#include <stdio.h>
#include <sys/types.h>
#include <sys/stat.h>
#include <fcntl.h>
#include <unistd.h>

int main()
{
	int fd1 = open("foo", O_CREAT | O_RDWR | O_CLOEXEC, 0644);
	int fd2 = open("bar", O_CREAT | O_RDWR, 0644);
	getchar();
	close(fd1);
	close(fd2);
	return 0;
}
```
with this change:
```
# crit show files.img | grep flags
               ...
                "flags": "O_RDWR | O_LARGEFILE", 
                "flags": "O_RDWR | O_LARGEFILE", 
```
without this change:
```
# crit show files.img | grep flags
               ...
                "flags": "O_RDWR | O_LARGEFILE | O_CLOEXEC",
                "flags": "O_RDWR | O_LARGEFILE", 
```

nviennot:
@0x7f454c46 I've added a test.

@Snorch I've documented the commit, and splitted the commit into two for clarity.

Thanks,
Nico

test/zdtm/static/file_cloexec.c
```diff
@@ -0,0 +1,63 @@
...
+
+int main(int argc, char *argv[])
+{
+	int fd1, fd2, fd3, fd4;
+
+	test_init(argc, argv);
+
+	fd1 = open("/", O_RDONLY | O_CLOEXEC);
+	if (fd1 < 0)
+		err(1, "Can't open()");
```
avagin: pls use pr_perror() to log errors

nviennot: There's a define on top that invokes pr_perror() and then exit().
Is it acceptable?

avagin: @nviennot I don't like this, but I think this is acceptable. I understand the problem what you are trying to solve here. What do you think about the following more generic solution:

```
diff --git a/test/zdtm/lib/zdtmtst.h b/test/zdtm/lib/zdtmtst.h
index 1fbf795bf..ffe18469c 100644
--- a/test/zdtm/lib/zdtmtst.h
+++ b/test/zdtm/lib/zdtmtst.h
@@ -122,11 +122,17 @@ extern int write_pidfile(int pid);
 /* message helpers */
 extern int test_log_init(const char *outfile, const char *suffix);
 extern int zdtm_seccomp;
-#define pr_err(format, arg...) \
-       test_msg("ERR: %s:%d: " format, __FILE__, __LINE__, ## arg)
-#define pr_perror(format, arg...)      \
+#define pr_err(format, arg...) ({ \
+       test_msg("ERR: %s:%d: " format, __FILE__, __LINE__, ## arg); \
+       1;  \
+})
+
+#define pr_perror(format, arg...)      ({ \
        test_msg("ERR: %s:%d: " format " (errno = %d (%s))\n", \
-               __FILE__, __LINE__, ## arg, errno, strerror(errno))
+               __FILE__, __LINE__, ## arg, errno, strerror(errno)); \
+       1; \
+})
+
 #define fail(format, arg...)   \
        test_msg("FAIL: %s:%d: " format " (errno = %d (%s))\n", \
                 __FILE__, __LINE__, ## arg, errno, strerror(errno))
diff --git a/test/zdtm/static/env00.c b/test/zdtm/static/env00.c
index 1feabfa9f..cb3b3417f 100644
--- a/test/zdtm/static/env00.c
+++ b/test/zdtm/static/env00.c
@@ -16,10 +16,8 @@ int main(int argc, char **argv)
 
        test_init(argc, argv);
 
-       if (setenv(envname, test_author, 1)) {
-               pr_perror("Can't set env var \"%s\" to \"%s\"", envname, test_author);
-               exit(1);
-       }
+       if (setenv(envname, test_author, 1))
+               return pr_perror("Can't set env var \"%s\" to \"%s\"", envname, test_author);
 
        test_daemon();
        test_waitsig();
```

nviennot: I think we should keep the `err()` semantics: http://man7.org/linux/man-pages/man3/err.3.html for the following reasons:
1) `return 1` works only when called from `main()`. When called from another function, `return -1` makes more sense. `exit(1)` always works.
2) We should try to keep functions that have well known semantics to ease readability.

PR #889, opened 2019-12-20
Rework mnt_is_external

Snorch:
These basicly does three steps:

1) rework mounts_sb_equal so that mnt_bind list for mnt_is_external will be right;
2) fix order of mnt_bind list setup and mnt_is_external and add BUG_ON to prevent these in future;
3) simplify mnt_is_external and split separate can_receive_master_from_external to use in resolve_shared_mounts.

### DIFF
--- a/criu/include/mount.h
+++ b/criu/include/mount.h
@@ -73,6 +73,7 @@ struct mount_info {
 	struct list_head	siblings;
 
 	struct list_head	mnt_bind;	/* circular list of derivatives of one real mount */
+	bool			mnt_no_bind;	/* no bind-mounts has been found for us */
 	struct list_head	mnt_share;	/* circular list of shared mounts */
 	struct list_head	mnt_slave_list;	/* list of slave mounts */
 	struct list_head	mnt_slave;	/* slave list entry */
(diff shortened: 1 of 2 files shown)


0x7f454c46:
Regarding the second commit: isn't `mnt_bind` always initialized in:
```
struct mount_info *mnt_entry_alloc()
{
[..]
		INIT_LIST_HEAD(&new->mnt_bind);
```
hmm?
_Adding later, looking at code:_
I'm sure you meant "not populated list". Is there a test for this?
Thanks!

Snorch:
Hi @0x7f454c46! Sad thing with test is: that it will require private external mount (so that old mnt_is_external does not find it from second mount) and it's bind-mount with external master (to get to mnt_is_external check in resolve_shared_mounts). Currently we does not support these in many other places and after the second patch test will still fail.

I've just mentioned that mnt_bind list is populated just after we use this list, it is surely wrong as we always see an empty list in resolve_shared_mounts->mnt_is_external.

0x7f454c46:
@Snorch right. As long as it's not urgent, I'll leave it to @avagin's review.
Thanks!

Snorch:
note: I'm preparing to add some more patches to these pull request with more improvements to external mounts, and hope some more tests in near future.

Snorch:
Reworked version is in https://github.com/checkpoint-restore/criu/pull/906. Added more tests and some more external mount related things.

Issue #860, opened 2019-11-19
error when checkpointing container: Can't lookup mount


giovannivenancio:
I have a freshly installed Ubuntu 18.04 and I get the following error when checkpointing a container:

`docker checkpoint create cr checkpoint1
Error response from daemon: Cannot checkpoint container cr: runc did not terminate sucessfully: criu failed: type NOTIFY errno 0 path= /run/containerd/io.containerd.runtime.v1.linux/moby/f2e3e31d2e1b75c6ad339f8931d38e296344f87917d7ef54b6de16a1268d2faf/criu-dump.log: unknown`

When inspecting the log, there is this error: (00.008480) Error (criu/files-reg.c:1338): Can't lookup mount=795 for fd=-3 path=/bin/sh

[criu-dump.log](https://github.com/checkpoint-restore/criu/files/3863610/criu-dump.log)

The container was created as follows:

`docker run --security-opt=seccomp:unconfined --name cr -d busybox /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'`

I'm using Docker 19.03.5 and CRIU 3.12.

Any ideas? Thanks in advance!

avagin:
What kernel do you use?

avagin:
(00.000079) Running on wraeclast Linux 5.0.0-36-generic #39~18.04.1-Ubuntu SMP Tue Nov 12 11:09:50 UTC 2019 x86_64

avagin:
Cc: @Snorch

Snorch:
Do you have mount 795 on host? If you still have dumpable ct running or you can reproduce it do:

```grep "^795\>" /proc/*/mountinfo```

If the above gives anything, please, also show the optput of:

 ```lsns | grep mnt```

1) You can have open fd in the dumpable process from the mount outside of dumpable mntns'es.
2) You can have open mount in detached mntns.
3) You can have open file on overlayfs and it sometimes shows mnt_id of pseudo mount which never existed.

If (1), you might need some external mount given to criu. (2) is not supported. For (3) I'm not sure if we can handle these.

Likely you case is (3):

[...]

PR #891, opened 2019-12-20
Fix tests on Ubuntu

adrianreber:
Podman Ubuntu packages are now available via OBS and no longer via PPA.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
   - TR_ARCH=local       CLANG=1 COMPAT_TEST=y
   - TR_ARCH=x86_64
   - TR_ARCH=x86_64      CLANG=1
-  - TR_ARCH=docker-test
   - TR_ARCH=openj9-test
 matrix:
   include:
@@ -59,6 +58,16 @@ matrix:
       dist: bionic
     - os: linux
       arch: amd64
+      env: TR_ARCH=docker-test
+      dist: bionic
+    - os: linux
+      arch: amd64
+      env: TR_ARCH=docker-test DIST=xenial
+      # On xenial it should be possible to test overlayfs;
+      # broken on the latest bionic kernel
+      dist: xenial
+    - os: linux
+      arch: amd64
       env: TR_ARCH=alpine      CLANG=1
       dist: xenial # test hangs on bionic
     - os: linux
@@ -79,9 +88,9 @@ matrix:
       dist: bionic
   allow_failures:
     - env: TR_ARCH=docker-test
+    - env: TR_ARCH=docker-test DIST=xenial
     - env: TR_ARCH=fedora-rawhide
     - env: TR_ARCH=local       GCOV=1
-    - env: TR_ARCH=podman-test
 script:
   - sudo make CCACHE=1 -C scripts/travis $TR_ARCH
 after_success:
--- a/scripts/travis/podman-test.sh
+++ b/scripts/travis/podman-test.sh
@@ -1,7 +1,13 @@
 #!/bin/bash
 set -x -e -o pipefail
 
-add-apt-repository -y ppa:projectatomic/ppa
+echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+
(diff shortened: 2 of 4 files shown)


scripts/travis/podman-test.sh
```diff
@@ -1,7 +1,10 @@
 #!/bin/bash
 set -x -e -o pipefail
 
-add-apt-repository -y ppa:projectatomic/ppa
+echo 'deb http://download.opensuse.org/repositories/devel:/kubic:/libcontainers:/stable/xUbuntu_18.04/ /' > /etc/apt/sources.list.d/devel:kubic:libcontainers:stable.list
+
+wget -nv https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_18.04/Release.key -O /tmp/Release.key
+apt-key add - < /tmp/Release.key
```
rst0git: `-O-` can be used to send the output to standard output and avoid the use of temporary file:
```
wget -nv -O- https://download.opensuse.org/repositories/devel:kubic:libcontainers:stable/xUbuntu_18.04/Release.key | apt-key add -
```

adrianreber: Thanks. Changed,

.travis.yml
```diff
@@ -56,7 +55,11 @@ matrix:
     - os: linux
       arch: amd64
       env: TR_ARCH=podman-test
-      dist: bionic
+      dist: xenial # latest bionic kernel breaks this test case
+    - os: linux
+      arch: amd64
+      env: TR_ARCH=docker-test
+      dist: xenial # latest bionic kernel breaks this test case
```
rst0git: it seems like `openj9-test` fails because of the issue described in https://github.com/checkpoint-restore/criu/issues/860. The following would fix this test as well:
```diff
diff --git a/.travis.yml b/.travis.yml
index 71ede34ea..c35e4650e 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,6 @@ env:
   - TR_ARCH=local       CLANG=1 COMPAT_TEST=y
   - TR_ARCH=x86_64
   - TR_ARCH=x86_64      CLANG=1
-  - TR_ARCH=openj9-test
 matrix:
   include:
     - os: linux
@@ -60,6 +59,10 @@ matrix:
       arch: amd64
       env: TR_ARCH=docker-test
       dist: xenial # latest bionic kernel breaks this test case
+    - os: linux
+      arch: amd64
+      env: TR_ARCH=openj9-test
+      dist: xenial # latest bionic kernel breaks this test case
     - os: linux
       arch: amd64
       env: TR_ARCH=alpine      CLANG=1
```

adrianreber:
@rst0git Yes, I also tried running the openj9 test with xenial and it worked. It would be good to exactly understand why the bionic kernel breaks container checkpointing.

adrianreber:
I am trying to fix the failing tests by switching to devicemapper instead of overlayfs.

Related to: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257.

We can either switch to an older distribution for Travis or disable `overlayfs` for our testing. Maybe doing both would also be a good idea. Let's see if my attempt to switch to `devicemapper` on the _bionic_ kernel gives us working tests again. If that works I will introduce the same test on _xenial_ based on `overlayfs`.

scripts/travis/podman-test.sh
```diff
@@ -13,6 +18,11 @@ apt-get update -qq
 
 apt-get install -qqy podman containernetworking-plugins
 
+# overlaysfs behaves differently on Ubuntu and breaks CRIU
+# https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
+#sed -e "s,driver.*$,driver = \"vfs\",g" -i /etc/containers/storage.conf
+echo 'driver = "vfs"' > /etc/containers/storage.conf
```
rst0git:
```sh
sed -i 's/^\(driver =\).*/\1 "vfs"/' /etc/containers/storage.conf
```

adrianreber: Thanks. I am still trying to figure out how to configure it all correctly. Lot's of changes all the time.

adrianreber:
From my side this is ready. docker-test is fixed by switching to devicemapper on bionic and introducing a xenial test with overlayfs. podman-test is fixed by using vfs as graphdriver and openj9-test is fixed by using devicemapper.

0x7f454c46:
@adrianreber, good work!
@mihalicyn, @rst0git thanks for reviews!

Merged, thanks!

PR #892, opened 2019-12-20
scripts: alpine: Install py2 packages with pip

rst0git:
The py-future package [has been renamed to py3-future](https://git.alpinelinux.org/aports/commit/main?id=316d44abaed13964e97eb43c095cd1b64e3943ad) and py2 package for yaml [has been dropped](https://git.alpinelinux.org/aports/commit/main?id=e369c1fd7707a73f2c3e2b11b613198d9a4106de).

### DIFF
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -27,9 +27,7 @@ RUN mv .ccache /tmp && make mrproper && ccache -sz && \
 	date && make -j $(nproc) CC="$CC" && date && ccache -s
 
 RUN apk add \
-	py-yaml \
 	py-pip	\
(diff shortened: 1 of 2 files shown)


adrianreber:
@rst0git Shouldn't we switch to Python 3 for alpine? Especially now that Python 2 is EOL?

rst0git:
I think we should, Alpine Linux are dropping Python 2 support.
https://wiki.alpinelinux.org/wiki/Python_package_policies#Python_2_deprecation
https://lists.sr.ht/~sircmpwn/alpine-devel/%3CBVLYYY50OMSF.91O7O5WI1KE7%40homura%3E

PR #893, opened 2019-12-21
Enforce kernel style

nviennot:
This PR includes two warnings that the kernel uses during compilation:
1) `-Wdeclaration-after-statement`: this enforces having variables declared at the top of scopes
2) `-Wstrict-prototypes`: this enforces full declaration of functions.
  Previously, when declaring `extern void func()`, one can call `func(123)` and have no compilation error. This is dangerous. The correct declaration is `extern void func(void)`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,7 +100,7 @@ export PROTOUFIX DEFINES
 DEFINES			+= -D_FILE_OFFSET_BITS=64
 DEFINES			+= -D_GNU_SOURCE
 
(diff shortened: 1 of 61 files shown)


mihalicyn:
cool suggestion!

0x7f454c46:
Merged, thanks!
@nviennot, please, be more verbose in the commit messages for history, thanks!

nviennot:
@0x7f454c46 I will try to be more verbose in commit messages in the future, thank you

PR #896, opened 2019-12-27
Add support of symlinks

mihalicyn:
This patchset adds support c/r of opened O_PATH fds and symlinks (ghost symlinks also)

Extract from commit messages:
1. fown: Don't fail on dumping files opened wit O_PATH
O_PATH opened files are special: they have empty
file operations in kernel space, so there not that
much we can do with them, even setting position is
not allowed. Same applies to a signal number for
owner settings.

2. files: allow dumping opened symlinks
To really open symlink file and not the regular file below it, one needs
to do open with O_PATH|O_NOFOLLOW flags. Looks like systemd started to
open /etc/localtime symlink this way sometimes, and before that nobody
actually used this and thus we never supported this in CRIU.

3. zdtm: add a test on open symlink migration

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -1817,11 +1817,17 @@ static int do_open_reg(int ns_root_fd, struct reg_file_info *rfi, void *arg)
 	if (fd < 0)
 		return fd;
 
-	if ((rfi->rfe->pos != -1ULL) &&
-			lseek(fd, rfi->rfe->pos, SEEK_SET) < 0) {
-		pr_perror("Can't restore file pos");
-		close(fd);
-		return -1;
+	/*
+	 * O_PATH opened files carry empty fops in kernel,
+	 * just ignore positioning at all.
+	 */
+	if (!(rfi->rfe->flags & O_PATH)) {
+		if (rfi->rfe->pos != -1ULL &&
+		    lseek(fd, rfi->rfe->pos, SEEK_SET) < 0) {
+			pr_perror("Can't restore file pos");
+			close(fd);
+			return -1;
+		}
 	}
 
 	return fd;
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -321,19 +321,53 @@ static int mkreg_ghost(char *path, GhostFileEntry *gfe, struct cr_img *img)
 	return ret;
 }
 
+static int mklnk_ghost(char *path, GhostFileEntry *gfe)
+{
+	if (!gfe->symlnk_target) {
+		pr_err("Ghost symlink target is NULL for %s. Image from old CRIU?\n", path);
+		return -1;
+	}
+
+	if (symlink(gfe->symlnk_target, path) < 0) {
+		/*
+		 * ENOENT case is OK
+		 * Take a look closer on create_ghost() function
+		 */
+		if (errno != ENOENT)
+			pr_perror("symlink(%s, %s) failed", gfe->symlnk_target, path);
+		return -1;
+	}
+
+	return 0;
+}
+
 static int ghost_apply_metadata(const char *path, GhostFileEntry *gfe)
 {
 	struct timeval tv[2];
 	int ret = -1;
 
-	if (chown(path, gfe->uid, gfe->gid) < 0) {
-		pr_perror("Can't reset user/group on ghost %s", path);
-		goto err;
-	}
+	if (S_ISLNK(gfe->mode)) {
+		if (lchown(path, gfe->uid, gfe->gid) < 0) {
+			pr_perror("Can't reset user/group on ghost %s", path);
+			goto err;
+		}
 
-	if (chmod(path, gfe->mode)) {
-		pr_perror("Can't set perms %o on ghost %s", gfe->mode, path);
-		goto err;
+		/*
+		 * We have no lchmod() function, and fchmod() will fail on
+		 * O_PATH | O_NOFOLLOW fd. Yes, we have fchmodat()
+		 * function and flag AT_SYMLINK_NOFOLLOW described in
+		 * man 2 fchmodat, but it is not currently implemented. %)
+		 */
+	} else {
+		if (chown(path, gfe->uid, gfe->gid) < 0) {
+			pr_perror("Can't reset user/group on ghost %s", path);
+			goto err;
+		}
+
+		if (chmod(path, gfe->mode)) {
+			pr_perror("Can't set perms %o on ghost %s", gfe->mode, path);
+			goto err;
+		}
 	}
 
 	if (gfe->atim) {
@@ -392,6 +426,9 @@ again:
 	} else if (S_ISDIR(gfe->mode)) {
 		if ((ret = mkdirpat(AT_FDCWD, path, gfe->mode)) < 0)
 			msg = "Can't make ghost dir";
+	} else if (S_ISLNK(gfe->mode)) {
+		if ((ret = mklnk_ghost(path, gfe)) < 0)
+			msg = "Can't create ghost symlink";
 	} else {
 		if ((ret = mkreg_ghost(path, gfe, img)) < 0)
 			msg = "Can't create ghost regfile";
@@ -779,6 +816,7 @@ static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_de
 	int exit_code = -1;
 	GhostFileEntry gfe = GHOST_FILE_ENTRY__INIT;
 	Timeval atim = TIMEVAL__INIT, mtim = TIMEVAL__INIT;
+	char pathbuf[PATH_MAX];
 
 	pr_info("Dumping ghost file contents (id %#x)\n", id);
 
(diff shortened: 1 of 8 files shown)


test/zdtm/static/symlink.c
```diff
@@ -0,0 +1,82 @@
...
+		pr_perror("cat't symlink to %s", test_symlink);
+		return 1;
+	}
+
+	fd = open(test_symlink, O_PATH | O_NOFOLLOW);
+	if (fd == -1) {
+		pr_perror("cat't open symlink %s", test_symlink);
+		return 1;
+	}
+
```
rst0git: `unlink(test_symlink);` ?

criu/files-reg.c
```diff
@@ -321,19 +321,48 @@ static int mkreg_ghost(char *path, GhostFileEntry *gfe, struct cr_img *img)
...
 }
 
+static int mklnk_ghost(char *path, GhostFileEntry *gfe, struct cr_img *img)
+{
+	if (!gfe->symlnk_target) {
+		pr_err("Ghost symlink target is NULL for %s. Image from old CRIU?\n", path);
+		return -1;
+	}
+
+	if (symlink(gfe->symlnk_target, path) < 0) {
```
rst0git: What if the folder where the `symlnk_target` has been created no longer exists?

test/zdtm/static/symlink.c
mihalicyn: yes, we need separate test, I think.

criu/files-reg.c
mihalicyn: > What if the folder of `path` no longer exists?

@rst0git Thank you for your point. If I understanding correctly, this case automatically handled because we call `mklnk_ghost` function only from `create_ghost` function. In which we already have a check for `ENOENT` error. It seems sufficient for me.

rst0git: In the case of `ENOENT`, CRIU will generate the following log message

```
Error (criu/files-reg.c:332): Could not create ghost symlink ... : No such file or directory
```
and execute:
```C
	/* Use grand parent, if parent directory does not exist */
	if (errno == ENOENT) {
		if (trim_last_parent(path) < 0) {
			pr_err("trim failed: @%s@\n", path);
			goto err;
		}
		goto again;
	}
```
As a result, it will try to create the symlink in the parent directory, potentially followed by successful restore. IMHO either the log message should not be an "error", or the restore should fail.

If the path in the ghost symlink is relative, would restoring it in a different directory make any difference?

mihalicyn: @rst0git thank you for the detailed explanation. I understood your point. I'm a little dumb %)
I've removed the error message and the behavior of `mklnk_ghost` should be the same as `mkreg_ghost` of course.

>If the path in the ghost symlink is relative, would restoring it in a different directory make any difference?

Great notice! But I think that in our case, it's really not important because we deal with ghost symlink and file descriptor opened with `O_PATH | O_NOFOLLOW` flags. In current kernels, all that we can do with that file descriptor is to `readlink()` / `fcntl()` / `fstat()` / `close()`.

mihalicyn: We discussed with Pavel (@Snorch) and after that discussion, I've decided to take a closer look at the potential problems with symlinks, created in the not original directory.

Idea of the test
1. Create symlink that points to file
2. open with `O_PATH | O_NOFOLLOW`
3. make symlink ghost (unlink it)
4. Open it using hacky procfs way with dereferencing: 
```
char lpath[1000];
sprintf(lpath, "/proc/self/fd/%d", fd);
ret = open(lpath, O_RDONLY);
if (ret == -1) {
	pr_perror("cat't open file");
	return 1;
}
```
5. And... we got a error:
`cat't open file (errno = 40 (Too many levels of symbolic links))`

All of that looks strange as for me. And this is a very weird and exotic way of using `O_PATH | O_NOFOLLOW` file descriptors, and also I've not found a way to crack the system. Perhaps,  this way exists but I don't know...

Snorch:
Please consider making a separate test for O_PATH, possibly you can just cherry-pick the one I prepared to check your changes. https://github.com/Snorch/criu/commit/6a6c42ff516a39c04ed01e019fd59877413d4a30

mihalicyn:
> Please consider making a separate test for O_PATH, possibly you can just cherry-pick the one I prepared to check your changes. (upd fix compilation) [Snorch@5d87107](https://github.com/Snorch/criu/commit/5d87107c46f67395d7d30a20f89d66f9495f8195)

I've added a test. Thank you!

test/zdtm/static/opath_file.c
```diff
@@ -0,0 +1,95 @@
...
+	file = fopen(path, "r");
+	if (!file) {
+		pr_perror("fopen");
+		return -1;
+	}
+
+	while (fgets(line, sizeof(line), file)) {
+		if (fdinfo_field(line, "flags")) {
+			if (sscanf(line, "%*s %lli", &val) != 1) {
+				pr_err("failed to read flags\n");
```
avagin: : %s", line

criu/pie/parasite.c
```diff
@@ -335,6 +335,19 @@ static int fill_fds_opts(struct parasite_drain_fd *fds, struct fd_opts *opts)
...
+		 * For O_PATH opened files there is no owner at all.
+		 */
+		flags = sys_fcntl(fd, F_GETFL, 0);
+		if (flags < 0) {
+			pr_err("fcntl(%d, F_GETFL) -> %d\n", fd, flags);
+			return -1;
+		}
+		if (flags & O_PATH) {
+			p->fown.pid = 0;
+			continue;
```
avagin: I don't like this "continue". It might be dangerous if someone will add something useful after filling up p->fown. I think we need to add a new helper which will contain the code to fill up p->fown.

criu/files-reg.c
```diff
@@ -321,19 +321,46 @@ static int mkreg_ghost(char *path, GhostFileEntry *gfe, struct cr_img *img)
 	return ret;
 }
 
+static int mklnk_ghost(char *path, GhostFileEntry *gfe)
+{
+	if (!gfe->symlnk_target) {
+		pr_err("Ghost symlink target is NULL for %s. Image from old CRIU?\n", path);
+		return -1;
+	}
```
0x7f454c46: so, you're going to break images back-compatibility?
also, shouldn't it be `gfe->has_symlnk_target`?

criu/files-reg.c
```diff
@@ -321,19 +321,46 @@ static int mkreg_ghost(char *path, GhostFileEntry *gfe, struct cr_img *img)
...
+
+	if (symlink(gfe->symlnk_target, path) < 0)
+		return -1;
```
0x7f454c46: `pr_perror()`

criu/files-reg.c
```diff
@@ -392,6 +419,9 @@ static int create_ghost(struct ghost_file *gf, GhostFileEntry *gfe, struct cr_im
 	} else if (S_ISDIR(gfe->mode)) {
 		if ((ret = mkdirpat(AT_FDCWD, path, gfe->mode)) < 0)
 			msg = "Can't make ghost dir";
+	} else if (S_ISLNK(gfe->mode)) {
+		if ((ret = mklnk_ghost(path, gfe)) < 0)
+			msg = "Can't create ghost symlink";
```
0x7f454c46: we probably want to be strict in images and report an error if
`!S_ISLNK(gfe->mode) && gfe->has_symlnk_target`

criu/files-reg.c
```diff
@@ -812,6 +843,36 @@ static int dump_ghost_file(int _fd, u32 id, const struct stat *st, dev_t phys_de
...
+		if (ret < 0) {
+			pr_perror("Can't readlinkat");
+			goto err_out;
+		}
+
+		pathbuf[ret] = 0;
+
+		if (ret != st->st_size) {
+			pr_err("Buffer for readlinkat is too small: ret %u, st_size %u, buf %u %s\n",
+					(unsigned)ret, (unsigned)st->st_size, (unsigned)PATH_MAX, pathbuf);
```
0x7f454c46: `ret %d` (or better make it ssize_t to be the same type as syscall)
`size %"PRIu64" `
and no need for any cast..

criu/files-reg.c
mihalicyn: No, I'm not going to break back-compatibility, and this check does not break compatibility because for old images we have next situation:

1. Let suppose that we have CRIU without these changes and use old images.
2. Also, let suppose, that we have opened fd to symlink with `O_PATH`
3. Then we do `criu dump`... all fine
4. Then we do `criu restore` and we will fail on restore. Because we have this check:
https://github.com/checkpoint-restore/criu/blob/43a7b34094ddda2a5f60dc4902a55a1d73f6f6ec/criu/files-reg.c#L1760
we will fail with the error, that file has bad mode, because with old criu + old image we will restore our ghost symlink as regular ghost file.

Now, let's take a look on case with CRIU + this patch + old image:
after `criu restore` we will also fail. :) But error message says us, that our image is bad and not contains symlink target.
Speaking strictly, this is incompatible change because we have changed behavior (error messages), but from another point of view in both cases, our `criu restore` will fail.

>also, shouldn't it be gfe->has_symlnk_target

As far as I know, for `optional string` fields and sub-messages there are no `has_` field. And we should simply check pointer.

mihalicyn:
@0x7f454c46 @avagin 
thanks for review! I will prepare fixes soon.

criu/files-reg.c
0x7f454c46: Fair enough :)

criu/files-reg.c
mihalicyn: > `ret %d` (or better make it ssize_t to be the same type as syscall)
> `size %"PRIu64" `
> and no need for any cast..

I'm sorry, PRId64 I think, as ssize_t is signed type. Right?

criu/files-reg.c
mihalicyn: > we probably want to be strict in images and report an error if
> `!S_ISLNK(gfe->mode) && gfe->has_symlnk_target`

I'm not sure, how to do that better way. If we want to be more strict with images, it also makes sense to refactor `create_ghost` function at all. There are also `gfe->has_rdev` parameter for instance.

criu/files-reg.c
0x7f454c46: Yep :)

criu/files-reg.c
0x7f454c46: Right, up to you - I don't insist

criu/files-reg.c
```diff
@@ -1817,11 +1817,17 @@ static int do_open_reg(int ns_root_fd, struct reg_file_info *rfi, void *arg)
...
 		return fd;
 
-	if ((rfi->rfe->pos != -1ULL) &&
+	/*
+	 * O_PATH opened files carry empty fops in kernel,
+	 * just ignore positioning at all.
+	 */
+	if (!(rfi->rfe->flags & O_PATH)) {
+		if ((rfi->rfe->pos != -1ULL) &&
 			lseek(fd, rfi->rfe->pos, SEEK_SET) < 0) {
```
avagin:
```
		if (rfi->rfe->pos != -1ULL &&
￼		    lseek(fd, rfi->rfe->pos, SEEK_SET) < 0) {
```

avagin:
ALL patches have to have only ONE your sob.

criu/files-reg.c
mihalicyn: Did you mean to just fix the indentation of `lseek`?

avagin: Yes and you can remove extra brackets.

mihalicyn:
> ALL patches have to have only ONE your sob.

ok

avagin:
Applied. Thanks!

PR #899, opened 2019-12-28
socket: Add c/r support for SO_LINGER and SO_OOBINLINE

rst0git:
This PR adds checkpoint/restore support for `SO_LINGER` and `SO_OOBINLINE` socket options.

### DIFF
--- a/criu/sockets.c
+++ b/criu/sockets.c
@@ -526,6 +526,7 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 {
 	int ret = 0, val = 1;
 	struct timeval tv;
+	struct linger so_linger;
 	/* In kernel a bufsize value is doubled. */
 	u32 bufs[2] = { soe->so_sndbuf / 2, soe->so_rcvbuf / 2};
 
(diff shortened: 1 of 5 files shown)


0x7f454c46:
Ok, LGTM

criu/sockets.c
```diff
@@ -566,6 +566,10 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 		pr_debug("\tset broadcast for socket\n");
 		ret |= restore_opt(sk, SOL_SOCKET, SO_BROADCAST, &val);
 	}
+	if (soe->has_so_oobinline && soe->so_oobinline) {
+		pr_debug("\tset oobinline for socket\n");
+		ret |= restore_opt(sk, SOL_SOCKET, SO_OOBINLINE, &val);
```
avagin: do we dump OOB data? I think we don't. Have you looked how hard to c/r them?

rst0git: OOB data is limited to 1 byte. When OOB message has size > 1 byte, only the last 1 byte of the message would be received as OOB, and the rest of the message can be received without the `MSG_OOB` flag. I'm not sure how we could restore data as OOB, however, with the following change CRIU will append the OOB data to the recv queue:
```diff
diff --git a/soccr/soccr.c b/soccr/soccr.c
index 20eabfbd..783f5234 100644
--- a/soccr/soccr.c
+++ b/soccr/soccr.c
@@ -378,6 +378,7 @@ err_recv:
 int libsoccr_save(struct libsoccr_sk *sk, struct libsoccr_sk_data *data, unsigned data_size)
 {
        struct soccr_tcp_info ti;
+       char oob_data;
 
        if (!data || data_size < SOCR_DATA_MIN_SIZE) {
                loge("Invalid input parameters\n");
@@ -400,6 +401,12 @@ int libsoccr_save(struct libsoccr_sk *sk, struct libsoccr_sk_data *data, unsigne
        if (get_queue(sk->fd, TCP_RECV_QUEUE, &data->inq_seq, data->inq_len, &sk->recv_queue))
                return -5;
 
+       if (recv(sk->fd, &oob_data, 1, MSG_OOB|MSG_PEEK) > 0) {
+               sk->recv_queue = realloc(sk->recv_queue, data->inq_len + 1);
+               memcpy(sk->recv_queue + data->inq_len, &oob_data, 1);
+               data->inq_len += 1;
+       }
+
        if (get_queue(sk->fd, TCP_SEND_QUEUE, &data->outq_seq, data->outq_len, &sk->send_queue))
                return -6;
```
@avagin should I add this change to the PR?

PR #902, opened 2020-01-09
travis: switch alpine to python3

adrianreber:
Now that Python 2 has officially reached its end of life also switch the Alpine based test to Python 3.


0x7f454c46:
LGTM

PR #903, opened 2020-01-09
Switch openj9 alpine tests to python3

nidhi1605:
Signed-off-by: Nidhi Gupta <itsnidhi16@gmail.com>


adrianreber:
Thanks, but I just realized, that test case is never run. Would it make sense to add it to `.travis.yaml` or is there not much value to test the java checkpoint/restore also on alpine? If not maybe we should just remove that Dockerfile? What do you think, should we remove the file or enable the test?

Issue #412, opened 2017-11-20
Restore failed on Oracle Database Server 12c R2


dineshputchala:
Trying to verify the checkpoint/restore feature on Oracle Database Server 12c R2

There was similar issue (https://github.com/checkpoint-restore/criu/issues/255) last year when I tried on some non-production docker-1.10.0-dev version.

This time I tried on latest version as this checkpoint/restore is enabled in experimental version of regular release of docker.

Steps followed are:

Enabled experimental flag on  "Docker version 17.06.2-ol, build d02b7ab"

bash-4.2$ docker run -d --env-file  db_env.dat -p :1521 -p  :5500  --name tc --security-opt seccomp:unconfined   store/oracle/database-enterprise:12.2.0.1
b1ed6b3ff854241230e357432e779238e4b0a14a32ea9b0661f87697161ac51c

Created checkpoint once the db came up,

bash-4.2$ docker checkpoint create tc tc_ck1
tc_ck1

bash-4.2$ docker checkpoint ls tc
CHECKPOINT NAME
tc_ck1

Trying to start the container again using checkpoint,

bash-4.2$ docker start --checkpoint tc_ck1 tc
Error response from daemon: oci runtime error: criu failed: type NOTIFY errno 0
log file: /var/lib/docker/containers/b1ed6b3ff854241230e357432e779238e4b0a14a32ea9b0661f87697161ac51c/checkpoints/tc_ck1/criu.work/restore-2017-11-17T02:06:14.324615919-08:00/restore.log

dineshputchala:
Attached the restore.log 

[restore.log](https://github.com/checkpoint-restore/criu/files/1486450/restore.log)

xemul:
Ugh, there's some problem with AIO ring. Cc @tkhai and @avagin

tkhai:
@dineshputchala, which kernel version do you use?

adrianreber:
From the shell version and the used glibc and other information in the restore.log this could be CentOS or RHEL. Which CRIU version are you using?

Strange that the CRIU version is not visible in the restore.log. We should also put the kernel version in the dump and restore log.

CRIU on CentOS/RHEL needs an extra patch if build from sources: https://git.centos.org/blob/rpms!criu.git/c7/SOURCES!aio-fix.patch

adrianreber:
My tests with migrating the oracle database have always failed probably due to problems with monotonic time.

adrianreber:
[...]

Issue #296, opened 2017-03-17
Absence of time namespace causes some apps (Oracle) to crash after restore


steffengy:
# Summary
Running an Oracle 11g XE container works as expected.
Creating and restoring a snapshot (using docker, which uses CRIU) itself also works.

The problem is that after the restore, the oracle processes crash becauseof a segfault.
(likely SysVIPC related, according to a stacktrace walkthrough [here](https://community.oracle.com/message/14023602#14023602).

The comments in the reproduction below explain some of what seems to go on.

# Reproduction
The following was run on a fresh Ubuntu 16.10.
```shell
# Install Docker on Host ( 17.03.0-ce / 3a232c8 )
sudo apt-get install apt-transport-https ca-certificates curl software-properties-common
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
sudo add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
sudo apt-get update && sudo apt-get install -y docker-ce

# Compile CRIU [ 36ebce8d9cf1b05cec691c642fc63fc8bfe44918 ] & put Docker into experimental mode
sudo apt-get install -y libaio-dev build-essential libnet1-dev pkg-config libnet1-dev
sudo apt-get install -y libprotobuf-dev libprotobuf-c0-dev protobuf-c-compiler protobuf-compiler python-protobuf libnl-3-dev libcap-dev
sudo apt-get install -y --no-install-recommends asciidoc xmlto
git clone --depth=1 https://github.com/xemul/criu.git && cd criu && make install
echo "{\"experimental\": true}" > /etc/docker/daemon.json
sudo service docker restart

# Run a oracle DB
docker run -d --shm-size=2g -p 1521:1521 -p 8080:8080 --name inspiring_swirles alexeiled/docker-oracle-xe-11g
docker checkpoint create inspiring_swirles test
docker start inspiring_swirles --checkpoint test
docker exec -it inspiring_swirles /bin/bash
cat /u01/app/oracle/diag/rdbms/xe/XE/trace/alert_XE.log
# This shows a time drift first, but that's only a warning
#  VKTM detected a time drift. Please check trace file for more details.
#  GEN0 (ospid: 43): terminating the instance due to error 472
#  Instance terminated by GEN0, pid = 43
# So the actual error happened in GENO, so let's take a look at that (the number after gen0 might be different case)
[...]

PR #905, opened 2020-01-10
image: core -- Reserve start_time field

valeriyvdovin:
To ensure consistency of runtime environment processes within a container need to see same start time values over suspend/resume cycles. We introduce new field to the core image structure to store start time of a dumped process. Later same value would be restored to a newly created task. In future the feature is likely to be pulled here, so we reserve field id in the core's protobuf descriptor.

### DIFF
--- a/images/core.proto
+++ b/images/core.proto
@@ -53,6 +53,8 @@ message task_core_entry {
 	//optional int32		tty_pgrp	= 17;
 
 	optional bool			child_subreaper	= 18;
+	// Reserved for container relative start time
+	//optional uint64		start_time	= 19;
 }
 
 message task_kobj_ids_entry {


adrianreber:
You need to sign off your patches like described at https://criu.org/How_to_submit_patches#Sign_your_work

And a bit more details why this change, even as simple as it is, is needed.

0x7f454c46:
@valeriyvdovin why do you want to reserve it?
The reason why tty_* fields were reserved is to make Openvz's CRIU version images compatible with the mainstream version. Is that the case in this PR?

0x7f454c46:
Ah, I've noticed your `@virtuozzo.com` address.
Not sure if we've seen any progress with tty features or even a roadmap when they would go in mainstream.
So, up to @avagin if he thinks it's worth to reserve that.

mihalicyn:
#412 #296

valeriyvdovin:
> Ah, I've noticed your `@virtuozzo.com` address.
> Not sure if we've seen any progress with tty features or even a roadmap when they would go in mainstream.
> So, up to @avagin if he thinks it's worth to reserve that.
This 'start_time' field is rather straight-forward and is the most basic stat of a process and should be covered by criu's basic functionality, so in our vz criu we definitely want to support it to fix issues with processes that depend on correct value of this field and in the light of that I don't see any reason to not reserve it in mainline, so we could pull it here later

Snorch:
We've already had these problem in VZ7 criu, if we use some id in core image we have a risk that these same field id will be used in ms criu, and we won't to have compatible images between mscriu and vzcriu. Last time we've introduced  task_core_entry_VZ730 =) https://src.openvz.org/projects/OVZ/repos/criu/browse/images/core.proto#27 And we don't want this pain again.

I'm not sure how much time it may take to implement start_time in mscriu, and how much time it will take to add proper interface for it in mskernel. But we can at least try implementing it in vzcriu and vzkernel first =).

PR #906, opened 2020-01-13
mount: Improvements for external mounts and sharing

Snorch:
Patch 1 is minor unrelated improvement of error printing in mount.c
Patch 2 is a fix for mnt_ext_master test to make it correspond to it's name
Patches 3-5 are to make external mounts resolve only in root mntns of CT
Patches 6-7 small fix of mnt_is_external and rework to make it easier to understand
(sadly no test as fail can happen only in too complex case for current criu)
Patches 8-9 are to properly resolve internal/external sharing
Patch 10 is to properly resolve internal/external slavely
(sadly no test as we don't support internal slavery fully)


Snorch:
Rebased on latest criu.

criu/mount.c
```diff
@@ -765,6 +780,9 @@ static int resolve_external_mounts(struct mount_info *info)
 		if (m->parent == NULL || m->is_ns_root)
 			continue;
 
+		/* Check mount has external sharing */
+		m->internal_sharing = !has_external_sharing(ext_ns->mnt.mntinfo_list, m);
```
avagin: what will has_external_sharing return on restore? This idea to look up external sharing in a criu namespace looks wrong for me. On dump, we can look up a mount. But how it should work on restore is unclear for me. What if we will restore on another host where shared, master ids for mounts are completely different?

Snorch: >what will has_external_sharing return on restore?
These function is called on dump only.
>This idea to look up external sharing in a criu namespace looks wrong for me. On dump, we can look up a mount. But how it should work on restore is unclear for me. What if we will restore on another host where shared, master ids for mounts are completely different?
On restore one should use m->internal_sharing same as it was before.

Sorry if I miss something.

avagin: > These function is called on dump only.
Oops. I misread the for_dump argument. I still don't understand the idea of resolve_external_mounts. Let's think that we don't use any auto ext mount options. We explicitly specify external mounts. In this case, I don't understand why we need to enumerate all mounts in the criu namespace.

Snorch: You miss that I'm trying to detect external _sharing_ not external mounts.

1) Even if you know that the mount is manually marked as external mount if it is shared it has two options: a)external share (need to inherit share) and b) internal share (need to create own share).

2) If "user" does not mark the mount as external mount it still can (or not) have external sharing, we need to detect this external dependecy and e.g. fail as we can't resolve external deps without helpof the "user".

Snorch:
Updated due to lost signedof in the patch "zdtm: add external sharing to mnt_ext_sharing test"

More other it looks like github somehow reorders patches... Original order: 

[snorch@snorch criu]$ git log --oneline --reverse | tail -n11
4c6c3370d mount: show more info about why we can't mount
99f8befc6 zdtm: fix mnt_ext_master test to correspond to it's name
9a59e7a38 mount: remove source check from mounts_sb_equal
78916a792 mount: prohibit external mounts in non-root mntns
2f9b07a70 zdtm: add new mnt_ext_sharing test
cc71c03f5 mount: setup mnt_bind list before using it in mnt_is_external
6b8791b15 mount: split mnt_is_external and can_receive_master_from_external
5b948f108 mount: make honest external/internal sharing detection
f9c06f7bd zdtm: add external sharing to mnt_ext_sharing test
687661cf2 mount: add external_slavery detection
b923b20f2 mount: add a message about found dev-mapping

avagin:
> 687661c mount: add external_slavery detection

Where is a test for this?

Snorch:
> > [687661c](https://github.com/checkpoint-restore/criu/commit/687661cf24711730e609f14eb63c0418d2078ca9) mount: add external_slavery detection
> 
> Where is a test for this?

The only test I can think of will still  fail, I only fix internal/external slave detection, not it's handling yet.

Patch 10 is to properly resolve internal/external slavely
(sadly no test as we don't support internal slavery fully)

Note 2: Sadly criu is not yet ready for a test in which we have external
mount being a slave of other mount in testct. But at least internal
slavery would not silently switch to external slavery after c/r.

avagin:
@Snorch pleas describe this behavior on wiki.

avagin:
> Patch 10 is to properly resolve internal/external slavely
(sadly no test as we don't support internal slavery fully)

Could you more detailed here? What is the problem? 

>  But at least internal slavery would not silently switch to external slavery after c/r.

Will we get an error?

Snorch:
> > Patch 10 is to properly resolve internal/external slavely
> > (sadly no test as we don't support internal slavery fully)
> 
> Could you more detailed here? What is the problem?

To fully support internal slavery we need to have some new mechanism, which inherits proper slave id from mounts which are mounted later. I don't think we need it, as if we have your kernel patch to add mounts to shared groups this part would be much easier.

> 
> > But at least internal slavery would not silently switch to external slavery after c/r.
> 
> Will we get an error?

No without check-mounts we don't detect creation of wrong mounts. But I create a hint we can use later to fix it, and replace internal slave with just private mount, these should break tree topology much less than it was before...

Snorch:
>Will we get an error?

Sorry for been slow, it looks like a good Idea to report error here. Can add it if you agree.

Snorch:
No functional update: s/1/true/ and some debugging added so that we can see how it works in logs.

Snorch:
@avagin sorry for delay with wiki, at first there was a problem with my account. Later it was hard to decide where is a proper place for this info and how to describe it better. So I put it to https://criu.org/External_bind_mounts#Sharing_for_external_bindmounts as a new paragraph. I chose self-descriptive example way.

Snorch:
Rebase again.

v3 changes:
Fixup for mnt_no_bind checks, for root_yard and cr_time which are not in mountinfo list.
Fail on internal slavery of external mount as it is not fully supported yet. (we were talking about this above, but now I think that failing is better than restoring changed state)

@avagin While there is no much movement here, I already have some more related patches on top of it: a) binds of external mounts need to wait them mounted, b) slavery without sharing need to be also propagated, c) handle ineternal/external sharing/slavery for root mount. Do I need wait while it is merged or add patches to this series?

https://src.openvz.org/projects/OVZ/repos/criu/commits/df10952fdf0f8c8abe65dd4a0b7aa87af56e202c
https://src.openvz.org/projects/OVZ/repos/criu/commits/50b909e5e8c1812bff79f19b227b7850dbdf589e
https://src.openvz.org/projects/OVZ/repos/criu/commits/08f4e7f2d4e8c11d7065500c5eb721b765756449
https://src.openvz.org/projects/OVZ/repos/criu/commits/51a7999d0e0c317eacd05d36e00770561bcc07ee
https://src.openvz.org/projects/OVZ/repos/criu/commits/ca5cd6b5d46abbc2535a1a821d49f3501cc73266

cyyzero:
I saw that [external bind mounts page](https://criu.org/External_bind_mounts) has mentioned this pull request. Has these commits been merged? As this pr is still open now.

Snorch:
@cyyzero Thanks for reminder, this pr should be closed. Some patches are already merged some are reworked in [Mount-v](https://criu.org/Mount-v2) and should appear in next release. Note that we don't plan to support external sharing only external slavery, it is a design decision at least for now.

PR #907, opened 2020-01-13
pagemap-cache: pread may return early with length less than requested

geminijun:
When we have a very large VMA, the size_map here is very long, tne pread may return early with length less than requested, we have to re-try to get all the data


criu/pagemap-cache.c
```diff
@@ -154,10 +154,16 @@ static int pmc_fill_cache(pmc_t *pmc, const struct vma_area *vma)
 	BUG_ON(pmc->map_len < size_map);
 	BUG_ON(pmc->fd < 0);
 
-	if (pread(pmc->fd, pmc->map, size_map, PAGEMAP_PFN_OFF(pmc->start)) != size_map) {
-		pmc_zap(pmc);
-		pr_perror("Can't read %d's pagemap file", pmc->pid);
-		return -1;
+	size_t have_read = 0;
```
rppt: Please move this declaration to the function declaration block, this will make travis happy :)

criu/pagemap-cache.c
```diff
@@ -154,10 +154,16 @@ static int pmc_fill_cache(pmc_t *pmc, const struct vma_area *vma)
...
+	while (have_read < size_map) {
+		ssize_t ret = pread(pmc->fd, (char *)pmc->map + have_read, size_map - have_read,
+		                    PAGEMAP_PFN_OFF(pmc->start) + have_read);
+		if (ret == -1) {
```
rppt: CRIU uses `< 0` to check syscall return value, please keep this consistent

0x7f454c46:
Please, sign your work by adding `Signed-off-by: ...` to your commit message.

criu/pagemap-cache.c
avagin: do we need to report an error if ret is 0 here?

PR #908, opened 2020-01-14
[RFC] compel: add -ffreestanding to force gcc not to use builtin memcpy...

mihalicyn:
compel: add -ffreestanding to force gcc not to use builtin memcpy, memset

This patch fixes the problem with SSE (xmm) registers corruption on amd64
architecture. The problem was that gcc generates parasite blob that uses
xmm registers, but we don't preserve this registers in CRIU when injecting
parasite. Also, gcc, even with -nostdlib option uses builtin memcpy,
memset functions that optimized for amd64 and involves SSE registers.

It seems, that optimal solution is to use -ffreestanding gcc option
to compile parasite. This option implies -fno-builtin and also it designed
for OS kernels compilation/another code that suited to work on non-hosted
environments and could prevent future sumilar bugs.

To check that you amd64 CRIU build affected by this problem you could simply
objdump -dS criu/pie/parasite.o | grep xmm
Output should be empty.

Reported-by: Diyu Zhou <zhoudiyupku at gmail.com>
Signed-off-by: Alexander Mikhalitsyn <alexander@mihalicyn.com>

### DIFF
--- a/compel/src/main.c
+++ b/compel/src/main.c
@@ -21,7 +21,7 @@
 
 #define CFLAGS_DEFAULT_SET					\
 	"-Wstrict-prototypes "					\
-	"-fno-stack-protector -nostdlib -fomit-frame-pointer "
+	"-fno-stack-protector -nostdlib -fomit-frame-pointer -ffreestanding "
 
 #define COMPEL_CFLAGS_PIE	CFLAGS_DEFAULT_SET "-fpie"
 #define COMPEL_CFLAGS_NOPIC	CFLAGS_DEFAULT_SET "-fno-pic"


mihalicyn:
I think that I need to provide additional details.

The problem was reported by Diyu Zhou (@vmexit) with a very stable reproduce program example.
Very helpful discussion:
https://lists.openvz.org/pipermail/criu/2019-September/044522.html

I've tried to make the next step in problem investigation and found, that problem is that corruption occurs not with FPU registers, but with SSE registers. Target program uses SSE registers for doubles (compiler optimizations).

This PR is attempt to develop a proper and universal solution to this and all similar potential problems with the parasite. I decided to choose `-ffreestanding` option because it suited for code that will work on non-hosted environment (OS kernels for instance). But tests were shown, that for this particular problem `-fno-builtin` seems enough.

Thanks to Dmitry (@0x7f454c46) for help and advices. ;)

cyrillos:
Before parasite takes place we do save the complete xsave frame and restore it back on parasite cleanup, don't we? IOW the idea was to be able to use any instructions inside parasite code.

If the compiler option you set fixups the problem I don't mind to merge as a fast fix (since I simply don't have time to dive into this issue right now). But maybe we should put some note somewhere in code to investigate the issue deeper once time permit.

mihalicyn:
> Before parasite takes place we do save the complete xsave frame and restore it back on parasite

for example in compel/arch/s390/src/lib/infect.c there are following comment:
```
 * We don't store floating point and vector registers here because we
 * assue that compel/pie code does not change them.
```

But I agree, maybe, it will be better to save a full CPU context before switching to a parasite.

I want to receive comments from team because of this change (adding full CPU context save/restore before parasite) not so simple, and should be implemented on every arch separately (as I know).

cyrillos:
I don't remember from my head if we save complete xsave frame on s86 or not. But if we not then definitely we should and this is not that trivial. Thus your patch should be merged I think.

0x7f454c46:
Yes, I tend to agree here with @cyrillos. I believe, the plan we discussed was:
1. Introduce a fault-injection test for corrupting fpu/sse/mpx registers in parasite on purpose and run it against related zdtm tests.
2. Make sure that for all threads parasite restores extended register set upon curing.
The (2) is important not only due to any optimisations those may use the registers but also in case of UB or anything unexpected in parasite - it should be able to cure the affected task from itself.

The registers are already saved on x86: `compel_get_thread_regs()` => `get_task_regs()`.

And again I agree with @cyrillos that we can apply the patch from pull request as it fixes a real issue.
Though, I believe when we get (1) & (2) working, the patch may be reverted as it won't affect anything and only make gcc generate un-optimized code.

0x7f454c46:
The downside of having this is that gcc won't inline `__builtin_memcpy()` and friends..
But as it fixes real issues, I better revert it later when the parasite code is ready :)

mihalicyn:
@0x7f454c46 

>I'll cook some patches to preserve the extended register set, according to the plan above.
As it will be convenient for your ;) I can try to develop this patch with registers preservation. It seems interesting :)

PR #910, opened 2020-01-15
s390x: remove stack pointer from clobber list

adrianreber:
Just like on all other supported architectures gcc complains about the stack pointer register being part of the clobber list: `error: listing the stack pointer register ‘15’ in a clobber list is deprecated [-Werror=deprecated]`

This removes the stack pointer from the clobber list.

`zdtm.py run -a` still runs without any errors after this change.


rppt:
cc @alicefr

adrianreber:
Same change as #909 

CC: @carnil 

@alicefr this change is pretty simple and the same as on all other architectures, maybe you can confirm that it is correct. So far everything still seems to work.

alicefr:
@adrianreber it seems correct to me

0x7f454c46:
LGTM

Issue #890, opened 2019-12-20
Implement clone3() assembler wrapper for non x86_64 architectures


adrianreber:
Pull request #882 adds clone3() based process and thread restore for x86_64 only.

For the PIE restorer we need assembler wrappers for clone3() for all other architectures.

This is a tracker bug to get the clone3() wrapper implemented for arm, aarch64, ppc64le and s390x.

I am happy to work on those implementations but if someone wants to work on one of those wrappers please let me know.

adrianreber:
clone3() based assembler wrapper implemented for all architectures.

PR #911, opened 2020-01-16
s390x: use clone3() if possible

adrianreber:
This adds the parasite clone3() with set_tid wrapper for s390x.

In contrast to the x86_64 implementation the thread start address and arguments are not put on the thread stack but passed via r4 and r5. As those registers are caller-saved they still contain the correct value (thread start address and arguments) after returning from the syscall.

Tested on 5.5.0-rc6.

Thanks to @fweimer and @huth for your help in understanding this.

CC: @alicefr 

Also see #882 and #890


fweimer:
Minor nit: r4 and r5 are caller-saved according to the userspace calling convention. They are callee-saved for the syscall. This is what allows the code to use them without saving them on the stack.

PR #914, opened 2020-01-17
ppc64le: use clone3() if possible

adrianreber:
This adds the parasite clone3() with set_tid wrapper for ppc64le.

Almost the same as #911 and depends on #911

Also see #882 and #890


avagin:
@adrianreber how have you tested these changes?

adrianreber:
Fedora 31 with a 5.5.0-rc6 kernel. Complete zdtm run. Only failures are cgroupv2  related and are happening with and without this change.

avagin:
@adrianreber  Could you rebase this pr?

PR #918, opened 2020-01-20
aarch64: use clone3() if possible

adrianreber:
This adds the parasite clone3() with set_tid wrapper for aarch64.
    
Tested on Fedora 31 with 5.5.0-rc6.

Depends on #911 and #914 

Also see #882 and #890


adrianreber:
Closing in favour of #925

Issue #895, opened 2019-12-24
Update Makefile.version in criu-dev to match master


adrianreber:
Running `criu--version` in `criu-dev` still returns 3.12. Would be nice to have criu-dev report the correct version.

0x7f454c46:
I believe, it should be updated by rebase of `criu-dev`.
I'll leave it up to Andrei as long as it's not urgent for you.

avagin:
Yes, we need to rebase criu-dev on master. @adrianreber if you have time, you can try to do this.

adrianreber:
Fixed with #919

PR #919, opened 2020-01-20
Rebase criu-dev on master 3.13

adrianreber:
@avagin as discussed in #895 I rebased criu-dev on top of master.

Most work was getting all the python re-formating right.


0x7f454c46:
LGTM.
You could probably while at it, drop:

31b557e88fdf Revert "travis-ci: Allow failures on ppc64le"
3e742447c436 travis-ci: Allow failures on ppc64le
f1e9373235df Revert "zdtm: Set mode with mkdir/mknod"
4775b8e50708 zdtm: Set mode with mkdir/mknod

but totally no hard feelings - maybe they're worth to keep for history.

avagin:
Done. Thanks a lot!

Issue #913, opened 2020-01-17
arm: Remove stack pointer from clobber list?


carnil:
Hi

Building `criu` (in this case based on 3.13) in Debian on armhf fails to build, similarly as already reported for ppc64el and s390x by @adrianreber.

I think the solution would look similar here, but explictly filling as issue.

The build log can be fetched here:
https://buildd.debian.org/status/fetch.php?pkg=criu&arch=armhf&ver=3.13-4&stamp=1579242990&raw=0

adrianreber:
Just tried to fix this, but CRIU does not seem to work on Fedora 31 armhfp with gcc 9.

I can get it to compile but it fails with: `Error (criu/vdso.c:615): vdso: Failed to fill self vdso symtable`

Any ideas what is necessary to get CRIU working on armhfp?

PR #921, opened 2020-01-21
arm: remove stack pointer from clobber list

adrianreber:
Just like on all other supported architectures gcc complains about the stack pointer register being part of the clobber list. This removes the stack pointer from the clobber list.

Fixes: #913 

Only compile tested so far.

### DIFF
--- a/criu/arch/arm/include/asm/restore.h
+++ b/criu/arch/arm/include/asm/restore.h
@@ -16,7 +16,7 @@
 		     : "r"(new_sp),					\
 		       "r"(restore_task_exec_start),			\
 		       "r"(task_args)					\
-		     : "sp", "r0", "r1", "memory")
+		     : "r0", "r1", "memory")
 
 static inline void core_get_tls(CoreEntry *pcore, tls_t *ptls)
 {


adrianreber:
After applying a patch from @0x7f454c46 I was able to run all tests on a system without VDSO support again. I am only seeing the usual cgroupv2 related errors.

avagin:
> After applying a patch from @0x7f454c46 

What patch are you talking about?

rst0git:
@adrianreber AFAIU the clobbering the stack pointer register is deprecated. GCC expects the value of the stack pointer after an asm statement to be the same as it was before, and no asm can validly clobber the stack pointer in the usual sense.

https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=99063eeecbfb15fece3c72747a9dd14077c41afd
https://github.com/ARM-software/CMSIS_5/commit/a07b56505471fd789111bd7dba726461449cc261

Perhaps we should remove `sp` from the clobber list in `criu/arch/aarch64/include/asm/restorer.h` as well?

adrianreber:
@rst0git Do you mean https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/arch/aarch64/include/asm/restorer.h#L52 ? That seems not to be used anywhere.

rst0git:
@adrianreber you are right, `ARCH_FAIL_CORE_RESTORE` is not used anywhere.

0x7f454c46:
LGTM
Will submit arm !CONFIG_VDSO fix separately in another PR.

PR #922, opened 2020-01-22
vdso: repair C/R on nodes with CONFIG_VDSO unset

0x7f454c46:
Apparently, after removing `CONFIG_VDSO` ifdefs from CRIU, C/R was broken on !CONFIG_VDSO duts.
Much help provided by @adrianreber to test this.
Unfortunately, there doesn't seem to be a way to disable vdso in kernel, except `vdso=` boot cmdline or .config option. So, in the result zdtm is not covering this platform-related code.


0x7f454c46:
Rebased over new criu-dev

PR #923, opened 2020-01-22
compel/criu: A bunch of different cleanups

0x7f454c46:
Remove some dead code that was found while reviewing mips  PR.


adrianreber:
Thanks. We just also talked about one of those things in #921 .

0x7f454c46:
I've force-pushed the branch: instead of removing `prologue.S` in  19f416d65c8ae09e394ad2ce49fa9acc6c239b4a, I think we better to remove the whole std plugin. I thought we use it, but it's still completely unused, untested and there is no futher plan of developing it AFAIK.
+Cc: @cyrillos in case he cares/minds.

cyrillos:
I don't mind.

compel/plugins/shmem/shmem.c
```diff
@@ -4,7 +4,6 @@
 #include <compel/plugins/shmem.h>
 #include <compel/plugins/std/syscall.h>
 #include "shmem.h"
-#include "std-priv.h"
```
rst0git: It seems like we use `std_ctl_sock()` below in `shmem_create()`:
```
ret = sys_write(std_ctl_sock(), &spi, sizeof(spi));
```

0x7f454c46: Right you are.
It seems like we don't build the shmem plugin either:
```
compel-plugins		+= compel/plugins/std.lib.a compel/plugins/fds.lib.a
```
I need to think about what we need from all this and probably re-use it for it's dead now too..

avagin:
@0x7f454c46 could you rebase this CL?

0x7f454c46:
@avagin I will delay this PR a bit - I'm thinking if I could reuse compel's plugins (std/shmem) instead of removing them as dead-code.. For a while, closing it - will reopen & rebase if won't come with ideas.

PR #925, opened 2020-01-22
arm/aarch64: use clone3() if possible

adrianreber:
This is the last architecture specific change to make CRIU use clone3() with set_tid if available. Just as on all other architectures this adds a clone3() based assembler wrapper to be used in the restorer code.

Tested on Fedora 31 with the same 5.5.0-rc6 kernel as on the other architectures.

Depends on the other architecture specific assembler wrapper pull request.


avagin:
@adrianreber could you rebase this pr? Thanks!

adrianreber:
Rebased. This includes the changes for aarch64 and arm. This are the two last assembler wrappers. All the other assembler changes for clone3() have been merged.

Issue #926, opened 2020-01-24
clone3() returns EINVALID if exit_signal set


adrianreber:
Trying to run CRIU to restore a runc based container does not work as clone3() fails.

Looking at the kernel there is a check if CLONE_PARENT is set and exit_signal is set and if both are set the kernel returns EINVALID.

Talking with @brauner about this he mentioned that the kernel always ignored the exit_signal if CLONE_PARENT is set even with clone(). The exit_signal of the thread group leader will always be used if CLONE_PARENT is specified.

CLONE_PARENT is mentioned a few times and I am not sure how to correctly handle it. My quick fix is to just unset `exit_signal` if CLONE_PARENT is set. But only for clone3():
```diff
-       c_args.exit_signal = exit_signal;
+       if (!(flags && CLONE_PARENT))
+               c_args.exit_signal = exit_signal;
```

As CLONE_PARENT seems to be necessary, according to the comments in `cr-restore.c` to work around some bugs in older kernels, this seems to be the best fix right now.

@avagin in other ideas how solve this best?

brauner:
Do you pass anything other than SIGCHLD somewhere?

adrianreber:
A quick grep through the code seems to indicate it is only SIGCHLD.

adrianreber:
Possible fix that works for me: #930

adrianreber:
Fixed with #930

PR #930, opened 2020-01-25
clone3: handle clone3() with CLONE_PARENT

adrianreber:
clone3() explicitly blocks setting an exit_signal if CLONE_PARENT is specified. With clone() it also did not work, but there was no error message. The exit signal from the thread group leader is taken.

runc restores are currently failing if clone3() is present as clone3() returns EINVALID if exit_signal is set and CLONE_PARENT is specified.

This fix works for me, but I am not sure if it is the right approach.

### DIFF
--- a/criu/clone-noasan.c
+++ b/criu/clone-noasan.c
@@ -61,7 +61,19 @@ int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags,
 
 	pr_debug("Creating process using clone3()\n");
 
-	c_args.exit_signal = exit_signal;
+	/*
+	 * clone3() explicitly blocks setting an exit_signal
+	 * if CLONE_PARENT is specified. With clone() it also
+	 * did not work, but there was no error message. The
+	 * exit signal from the thread group leader is taken.
+	 */
+	if (!(flags & CLONE_PARENT)) {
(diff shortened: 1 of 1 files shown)


adrianreber:
Also see #926

criu/clone-noasan.c
```diff
@@ -61,7 +61,14 @@ int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags,
...
 
-	c_args.exit_signal = exit_signal;
+	/*
+	 * clone3() explicitly blocks setting an exit_signal
+	 * if CLONE_PARENT is specified. With clone() it also
+	 * did not work, but there was no error message. The
+	 * exit signal from the thread group leader is taken.
+	 */
+	if (!(flags & CLONE_PARENT))
+		c_args.exit_signal = exit_signal;
```
avagin: should we failt if exit_signal isn't equal to SIGCHLD?

adrianreber: No idea, but if you think it would be correct, I can add that. Let me add it and push a new version of this PR.

avagin: Yes, I think it would be correct.

adrianreber: Now failing if `exit_signal != SIGCHLD`

PR #935, opened 2020-01-30
Man fixes

kolyshkin:
This is mostly fixes for better rendering with asciidoctor (which is used by default since #703).

Please see detailed descriptions in individual commits.

**UPDATE**: man page looks fine in Fedora 31 (criu-3.13-5.fc31). I think these changes will be much needed by Fedora 32 since they will get rid of Python 2 (and therefore asciidoc).

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -54,7 +54,7 @@ ifneq ($(USE_ASCIIDOCTOR),)
 	$(Q) $(ASCIIDOC) -b manpage -d manpage -o $@ $<
 else
 	$(Q) $(ASCIIDOC) -b docbook -d manpage -o $(patsubst %.1,%.xml,$@) $<
-	$(Q) $(XMLTO) man -m custom.xsl $(patsubst %.1,%.xml,$@) 2>/dev/null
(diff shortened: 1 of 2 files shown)


kolyshkin:
> For the TODO in the commit message, which monospaced text are you referring to?

Code and command line examples, listings etc. For example,
```console
$ cd Documentation
$ make criu.pdf
$ evince criu.pdf # or whatever your PDF viewer is
```

Look for the word `freezer` in the resulting PDF (in my case it's page 5) and you'll see the text is not monospaced as it should be. If you do the same when using asciidoc, you'll see monospaced text in there.

So far I tried using `----` or `....` as delimiters, as described in https://powerman.name/doc/asciidoc#_blocks, as well as `[literal]` or `[listing]` or `[source]` attributes, as described in https://asciidoctor.org/docs/user-manual/#literal-text-and-blocks (NOTE: not all of it is compatible with asciidoc!) Alas, nothing works.

I'm afraid the fix belongs to asciidoctor's manpage backend. Looks like that asciidoc+xmlto inserts
```groff
.fam C
......
.fam
```
pair around the liternal text, while asciidoctor does not.

kolyshkin:
Trying to fix this in asciidoctor: https://github.com/asciidoctor/asciidoctor/pull/3561

PR #929, opened 2020-01-24
Rawhide build warnings

0x7f454c46:
Fixes: #928

Issue #928, opened 2020-01-24
criu/img-remote.c:1099:2: error: 'strncpy' specified bound 4096 equals destination size [-Werror=stringop-truncation]


adrianreber:
Fedora Rawhide switched to gcc 10 and Travis sees following error on aarch64:
```
criu/img-remote.c: In function 'push_snapshot_id':
criu/img-remote.c:1099:2: error: 'strncpy' specified bound 4096 equals destination size [-Werror=stringop-truncation]
 1099 |  strncpy(rn.snapshot_id, snapshot_id, PATH_MAX);
      |  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  CC       criu/sk-tcp.o
cc1: all warnings being treated as errors
make[2]: *** [/criu/scripts/nmk/scripts/build.mk:118: criu/img-remote.o] Error 1
```

https://travis-ci.org/checkpoint-restore/criu/jobs/641275533

0x7f454c46:
I was looking into in yesterday: https://github.com/0x7f454c46/criu/commits/rawhide-warnings

Issue #927, opened 2020-01-24
Fedora Rawhide Travis Failure - x86_64


adrianreber:
Fedora Rawhide is now shipping a pre-release of GCC 10 and it fails on x86_64:
```
  LINK     criu/arch/x86/crtools.built-in.o
ld: criu/arch/x86/crtools.o:/criu/criu/include/cr_options.h:159: multiple definition of `rpc_cfg_file'; criu/arch/x86/cpu.o:/criu/criu/include/cr_options.h:159: first defined here
```

https://travis-ci.org/checkpoint-restore/criu/jobs/641275534

adrianreber:
Also fixed.

PR #936, opened 2020-02-02
criu: fix build failure against gcc-10

trofi:
On gcc-10 (and gcc-9 -fno-common) build fails as:

```
ld: criu/arch/x86/crtools.o:criu/include/cr_options.h:159:
  multiple definition of `rpc_cfg_file'; criu/arch/x86/cpu.o:criu/include/cr_options.h:159: first defined here
make[2]: *** [scripts/nmk/scripts/build.mk:164: criu/arch/x86/crtools.built-in.o] Error 1
```

gcc-10 will change the default from -fcommon to fno-common:
https://gcc.gnu.org/PR85678.

The error also happens if CFLAGS=-fno-common passed explicitly.

Reported-by: Toralf Förster
Bug: https://bugs.gentoo.org/707942

### DIFF
--- a/criu/config.c
+++ b/criu/config.c
@@ -30,6 +30,7 @@
 #include "common/xmalloc.h"
 
 struct cr_options opts;
+char *rpc_cfg_file;
 
 static int count_elements(char **to_count)
 {
--- a/criu/include/cr_options.h
+++ b/criu/include/cr_options.h
@@ -156,7 +156,7 @@ struct cr_options {
 };
 
 extern struct cr_options opts;
-char *rpc_cfg_file;
+extern char *rpc_cfg_file;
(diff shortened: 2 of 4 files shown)


adrianreber:
Thanks for the patches. We have seen this also on Fedora a couple of days ago and there is already a pull request for this: #929 (but needs rebasing)

Also see #928 and #927 

@0x7f454c46 is this PR the same fix you provided? Looks like #929 has the same changes as this PR and then also the `strncpy()` fixes. We could merge this PR and if you rebase #929 just drop the `extern` related patches.

0x7f454c46:
@adrianreber Yeah, that is, let's merge this one :)

PR #937, opened 2020-02-03
In tests check current working directory permissions in filesystem operations with absolute paths.

valeriyvdovin:
Any filesystem syscall, that needs to navigate to inode by it's
absolute path performs successive lookup operations for each part of the
path. Lookup operation includes access rights check.
Usually but not always zdtm tests processes fall under 'other' access
category. Also, usually directories don't have 'x' bit set for other.
In case when bit 'x' is not set and user-ID and group-ID of a process
relate it to 'other', test's will not succeed in performing these
syscalls which are most of filesystem api, that has const char *path
as part of it arguments (open, openat, mkdir, bind, etc).
The observable behavior of that is that zdtm tests fail at file
creation ops on one system and pass on the other. The above is not
immediately clear to the developer by just looking at failed test's logs.
Investigation of that is also not quick for a developer due to the
complex structure of zdtm runtime where nested clones with
NAMESPACE flags take place alongside with bind-mounts.

As an additional note: 'get_current_dir_name' is documented as returning
EACCESS in case when some part of the path lacks read/list permissions.
But in fact it's not always so. Practice shows, that test processes can
get false success on this operation only to fail on later call to
something like mkdir/mknod/bind with a given path in arguments.

'get_cwd_check_perm' is a wrapper around 'get_current_dir_name'. It also
checks for permissions on the given filepath and logs the error. This
directs the developer towards the right investigation path or even
eliminates the need for investigation completely.

In the same PR added helper function that uses the above wrapper
to create absolute paths for a unix socket's with filling of sockaddr_un structre.
initialize unix socket address to an absolute file path, derived from
current working directory of the test + relative filename of a resulting
socket. Because the former code used cwd = get_current_dir_name() as
part of absolute filename generation, the resulting filepath could later
cause failure of bind systcall due to unchecked permissions and
introduce confusing permission errors.

### DIFF
--- a/test/zdtm/lib/unix.c
+++ b/test/zdtm/lib/unix.c
@@ -1,0 +1,19 @@
+#include <sys/socket.h>
+#include <sys/un.h>
+#include "zdtmtst.h"
+#include "fs.h"
+
+int unix_fill_sock_name(struct sockaddr_un *name, char *relFilename)
+{
+	char *cwd;
+
+	if (get_cwd_check_perm(&cwd)) {
+		pr_err("failed to get current working directory with valid permissions.\n");
+		return -1;
+	}
+
+	name->sun_family = AF_LOCAL;
+	ssprintf(name->sun_path, "%s/%s", cwd, relFilename);
+	return 0;
+}
+
--- a/test/zdtm/lib/Makefile
+++ b/test/zdtm/lib/Makefile
@@ -4,7 +4,7 @@ CFLAGS	+= $(USERCFLAGS)
 
 LIB	:= libzdtmtst.a
 
-LIBSRC	:= datagen.c msg.c parseargs.c test.c streamutil.c lock.c ns.c tcp.c fs.c sysctl.c
+LIBSRC	:= datagen.c msg.c parseargs.c test.c streamutil.c lock.c ns.c tcp.c unix.c fs.c sysctl.c
 LIBOBJ	:= $(LIBSRC:%.c=%.o)
 
 BIN	:= groups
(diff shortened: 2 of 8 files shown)


adrianreber:
I am not familiar with the details of that code, but that sounds like something which I was also having problems before. Sounds like a good idea to include that check.

You definitely need to add the Signed-off-by to your commits according to CRIU's guidelines.

valeriyvdovin:
> I am not familiar with the details of that code, but that sounds like something which I was also having problems before. Sounds like a good idea to include that check.
> 
> You definitely need to add the Signed-off-by to your commits according to CRIU's guidelines.
Sure.

test/zdtm/lib/unix.c
```diff
@@ -0,0 +1,25 @@
...
+	if (get_cwd_check_perm(&cwd)) {
+		pr_err("failed to get current working directory with valid permissions.\n");
+		return -1;
+	}
+
+	if (strlen(relFilename) + strlen(cwd) + 1 >= sizeof(name->sun_path)) {
+		pr_err("filename \"%s/%s\" is too long for sockaddr_un initialiazation.\n", 
+			cwd, relFilename);
+		return -1;
+	}
```
Snorch: ssprintf handles path length check itself, and calls abort() to create a coredump so it would be easier to investigate the problem.

Snorch:
I agree with these change as I also spent some time on investigating this once before.

test/zdtm/lib/unix.c
valeriyvdovin: Agreed

PR #939, opened 2020-02-04
Use cgroup v2 freezer if detected

adrianreber:
With these changes I am able to use runc again on a cgroup v2 system. I am also preparing a small runc patch to fall back to non-freezer based checkpointing if necessary.

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -20,6 +20,7 @@
 #include "seccomp.h"
 #include "seize.h"
 #include "stats.h"
+#include "string.h"
 #include "xmalloc.h"
 #include "util.h"
 #include <compel/compel.h>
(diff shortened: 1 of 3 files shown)


adrianreber:
Thanks @rst0git. I addressed all the things you pointed out.

rst0git:
@adrianreber maybe we can add a test in Travis for cgroup v2 (e.g. https://github.com/opencontainers/runc/commit/ccd4436fc46dec671c45dcf69fa8c51a283b3fe5)?

adrianreber:
Yes, I am currently experimenting with exactly this setup to figure out how to use it for CRIU. I was planing to do this independently of this PR as this requires some time to get it working with Travis as it always takes a lot of waiting to change something in Travis and make sure everything works. As long as it does not break existing functionality I think we could introduce additional tests in another PR.

criu/seize.c
```diff
@@ -77,13 +78,34 @@ const char *get_real_freezer_state(void)
...
+		if (strlcpy(state, thawed, sizeof(state)) >= sizeof(state))
+			return -1;
+	} else if (new_state == FROZEN) {
+		if (strlcpy(state, frozen, sizeof(state)) >= sizeof(state))
+			return -1;
+	} else {
+		return -1;
+	}
+
+	lseek(fd, 0, SEEK_SET);
```
avagin: need to check a return code

criu/seize.c
```diff
@@ -92,13 +114,24 @@ static int freezer_restore_state(void)
...
-	if (write(fd, frozen, sizeof(frozen)) != sizeof(frozen)) {
-		pr_perror("Unable to freeze tasks");
-		close(fd);
+	return fd;
+}
+
+static int freezer_restore_state(void)
+{
+	int fd;
+	int exit_code;
```
avagin: exit_code is used only in this patter:
````
{
  int exit_code = -1;
  ...
  ...
  exit_code = 0;
out:
  return exit_code;
}
```

In this case, I would rename exit_code to ret;

criu/seize.c
```diff
@@ -361,12 +393,10 @@ static int freeze_processes(void)
 	pr_debug("freezing processes: %lu attempts with %lu ms steps\n",
 		 nr_attempts, step_ms);
 
-	snprintf(path, sizeof(path), "%s/freezer.state", opts.freeze_cgroup);
-	fd = open(path, O_RDWR);
-	if (fd < 0) {
-		pr_perror("Unable to open %s", path);
+	fd = freezer_open();
+	if (fd == -1)
```
avagin: can it be fd < 0

criu/seize.c
```diff
@@ -71,6 +74,66 @@ static enum freezer_state get_freezer_state(int fd)
...
+
+	/*
+	 * cgroupv2 freezer uses cgroup.freeze to control the state. The file
+	 * can return 0 or 1. 1 means the cgroup is frozen; 0 means it is not
+	 * frozen. Writing 1 to an unfrozen cgroup can freeze it. Freezing can
+	 * take some time and if the cgroup has finished freezing can be
+	 * seen in cgroup.events: frozen 0|1.
+	 */
+
+	lseek(fd, 0, SEEK_SET);
```
avagin: Need to check the return code

criu/seize.c
```diff
@@ -815,6 +901,28 @@ static int collect_task(struct pstree_item *item)
 	return -1;
 }
 
+static int cgroup_version(void)
+{
+	char path[PATH_MAX];
+	struct stat st;
+
+	snprintf(path, sizeof(path), "%s/freezer.state", opts.freeze_cgroup);
+	if (stat(path, &st) == 0) {
```
avagin: can we use access instead of stat?

adrianreber:
@avagin I addressed all you comments and force pushed.

PR #950, opened 2020-02-12
[RFC] Travis: Add names for jobs

0x7f454c46:
As I use to forget what's the difference between `TR_ARCH=local` and `TR_ARCH=x86_64`, I suggest adding some descriptions for jobs.
The idea is to clearly state: architecture + build/zdtm + compiler or other peculiarities.
At first, they look unfamiliar, but I guess some time needed to get used to them.


adrianreber:
Good idea. The difference between `local` and `x86_64` is indeed confusing.

One of the test listed in `allow_failures` is not being ran at all: `TR_ARCH=local       GCOV=1`.

They do look indeed unfamiliar, but I think this is a good change.

adrianreber:
@0x7f454c46 Moving away from Travis this probably makes not much sense any more. Do you want to close it?

0x7f454c46:
@adrianreber Yes. Thanks for reminding - I've managed to forget about this PR.

PR #957, opened 2020-02-18
 criu(8): Add documentation for --enable-fs 

rst0git:
This option was introduced with commit https://github.com/checkpoint-restore/criu/commit/e2c38245c613df5e36dcf0253c7652f928e46abf.

### DIFF
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -90,6 +90,19 @@ The following levels are available:
 *-L*, *--libdir* 'path'::
     Path to plugins directory.
 
+*--enable-fs* ['fs'[,'fs'...]]::
+    Specify a comma-separated list of filesystem names that should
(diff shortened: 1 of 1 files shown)


avagin:
Could you describe this option on the wiki as well?

Snorch:
Maybe we can also mention that --enable-fs does not fit with --external dev[]: , see try_resolve_ext_mount, external dev mounts only determined for FSTYPE__UNSUPPORTED.

Also I'm not sure about other place where we rely on FSTYPE__UNSUPPORTED in do_mount_one: fstype will not be filled for btrfs on restore, not sure if it's bad or not though.

rst0git:
> Could you describe this option on the wiki as well?

https://criu.org/CLI/opt/--enable-fs

PR #916, opened 2020-01-17
util: extract ghost file copy logic

nviennot:
This is a refactor of the ghost file copy logic.
It was originally intended for the memfd code, but still a good addition to have. See https://github.com/checkpoint-restore/criu/pull/886#issuecomment-575792993

PR #968, opened 2020-02-29
memfd: don't corrupt a state of the dumped fd

avagin:
Right now, criu uses a dumped fd to dump content of a memfd "file".

Here are two reasons why we should not do this:
* a state of a dumped fd doesn't have to be changed, but now criu calls
  lseek on it. This can be workarounded by using pread.
* a dumped descriptor can be write-only.

Reported-by: Mr Jenkins
Cc: @nviennot

### DIFF
--- a/criu/memfd.c
+++ b/criu/memfd.c
@@ -119,6 +119,7 @@ out:
 static struct memfd_inode *dump_unique_memfd_inode(int lfd, const char *name, const struct stat *st)
 {
 	struct memfd_inode *inode;
+	int fd;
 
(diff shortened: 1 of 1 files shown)


nviennot:
Ah very well.

I though having this covered, but it was in the ghost file logic https://github.com/checkpoint-restore/criu/pull/916/files#diff-dfaf4b21e68230a85cd2f08277fe7cc9R567-R579

The amount of commonality between ghost files and shmem makes me wonder if the two subsystems should be combined in some ways

LGTM

PR #969, opened 2020-02-29
zdtm/inhfd: force python to read new data from a file

avagin:
python 2.7 doesn't call the read system call if it's read file to the
end once. The next seek allows to workaround this problem.

inhfd/memfd.py hangs due to this issue.

Reported-by: Mr Jenkins
Signed-off-by: Andrei Vagin <avagin@gmail.com>


avagin:
Cc: @nviennot

nviennot:
great, LGTM

PR #977, opened 2020-03-05
zdtm/fifo_loop: don't try to write more than pipe size

avagin:
... otherwise write() can block.

Reported-by: Mr Jenkins

### DIFF
--- a/test/zdtm/transition/fifo_loop.c
+++ b/test/zdtm/transition/fifo_loop.c
@@ -39,6 +39,7 @@ int main(int argc, char **argv)
 	int i;
 	uint8_t buf[0x100000];
 	char *file_path;
+	int pipe_size;
 
 	test_init(argc, argv);
 
@@ -104,6 +105,13 @@ int main(int argc, char **argv)
 		exit(1);
 	}
 
+	pipe_size = fcntl(writefd, F_SETPIPE_SZ, sizeof(buf));
+	if (pipe_size != sizeof(buf)) {
+		pr_perror("fcntl(writefd, F_GETPIPE_SZ) -> %d", pipe_size);
+		kill(0, SIGKILL);
+		exit(1);
+	}
+
 	file_path = path[i - 1];
 	readfd = open(file_path, O_RDONLY);
 	if (readfd < 0) {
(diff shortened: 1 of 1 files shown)


test/zdtm/transition/fifo_loop.c
```diff
@@ -125,7 +132,7 @@ int main(int argc, char **argv)
 		uint8_t rbuf[sizeof(buf)], *p;
 
 		datagen(buf, sizeof(buf), NULL);
-		wlen = write(writefd, buf, sizeof(buf));
+		wlen = write(writefd, buf, pipe_size);
```
mihalicyn: Maybe I'm wrong, but it looks like `min(sizeof(buf), pipe_size)` is safer.

Also, here
https://github.com/checkpoint-restore/criu/blob/0fd9f79ad03251bedc9ff653fcb94a2cbdaae15d/test/zdtm/transition/fifo_loop.c#L88

we have write with `sizeof(buf)` size. Shouldn't we need to use `min(sizeof(buf), pipe_size)` here too?

test/zdtm/transition/fifo_loop.c
```diff
@@ -39,6 +39,7 @@ int main(int argc, char **argv)
 	int i;
 	uint8_t buf[0x100000];
```
rst0git:
```
$ ag 0x100000
...
test/zdtm/transition/fifo_dyn.c
31:	uint8_t buf[0x100000];

test/zdtm/transition/fifo_loop.c
40:	uint8_t buf[0x100000];

test/zdtm/transition/pipe_loop00.c
34:	uint8_t buf[0x100000];

test/zdtm/transition/socket_loop00.c
35:	uint8_t buf[0x100000];
```
The default pipe buffer size is 0x100000 == 1048576
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/fs/pipe.c?id=HEAD#n37

Why not set the pipe's buffer size instead? At least it may save a few clock cycles every time we run the tests.

avagin: This is a good idea. Thanks!

test/zdtm/transition/fifo_loop.c
```diff
@@ -104,6 +105,13 @@ int main(int argc, char **argv)
 		exit(1);
 	}
 
+	pipe_size = fcntl(writefd, F_SETPIPE_SZ, sizeof(buf));
+	if (pipe_size != sizeof(buf)) {
+		pr_perror("fcntl(writefd, F_GETPIPE_SZ) -> %d", pipe_size);
```
rst0git: s/F_GETPIPE_SZ/F_SETPIPE_SZ/g

Issue #924, opened 2020-01-22
CRIU 3.12 cross-compile issue: compel-host-bin (stack smashing detected)


shrkamat:
I am facing CRIU cross-compile failure for aarch64 target
Looks like some problem with running compel-host-bin for intermediate code generation

**Compile options**
```
export PKG_CONFIG_PATH=/output/aarch64/lib/pkgconfig
make ARCH=aarch64 CROSS_COMPILE=/usr/bin/aarch64-linux-gnu- USERCFLAGS="-I/output/aarch64/include -I/output/aarch64/include/libnl3 -I/usr/aarch64-linux-gnu/include -L/usr/aarch64-linux-gnu/lib -L//output/
aarch64/lib64 -L//output/aarch64/lib -lprotobuf-c -lpthread -lrt"

```
**Error**
```
./compel/compel-host hgen -f criu/pie/parasite.built-in.o -o criu/pie/parasite-blob.h
*** stack smashing detected ***: <unknown> terminated
criu/pie/Makefile:49: recipe for target 'criu/pie/parasite-blob.h' failed
make[2]: *** [criu/pie/parasite-blob.h] Aborted (core dumped)
make[2]: *** Deleting file 'criu/pie/parasite-blob.h'
criu/Makefile:61: recipe for target 'pie' failed
make[1]: *** [pie] Error 2
Makefile:228: recipe for target 'criu' failed
make: *** [criu] Error 2
```

**Steps to reproduce**
```
git clone https://github.com/shrkamat/criu-aarch64
cd criu-aarch64
./run.sh
# Inside docker shell
source ./build.sh
```

P.S. I had to make one small change to Makefiles.packages
https://github.com/shrkamat/criu-aarch64/blob/master/criu/patches/Fix-CRIU-find-packages.patch
I am not clear why the try-cc step fails, does it not use cross-compiler ?

avagin:
Cc: @0x7f454c46

0x7f454c46:
Thanks for the Cc.
I'll check it next week - a little sick right now.

PR #978, opened 2020-03-05
Travis: Add arm64 cross build test

0x7f454c46:
To do that, I had to clean up compel headers a bit.
Fixes: https://github.com/checkpoint-restore/criu/issues/924


scripts/build/Dockerfile.aarch64-cross
```diff
@@ -0,0 +1,45 @@
+FROM dockcross/base:latest
+
+# Add the cross compiler sources
+RUN echo "deb http://ftp.us.debian.org/debian/ jessie main" >> /etc/apt/sources.list && \
```
adrianreber: jessie is called 'oldoldstable' on the Debian homepage, does this also exist for the latest Debian release, 'buster'?

0x7f454c46: I've added 357d36ef5722 on the top to use buster for both cross-build tests.

PR #979, opened 2020-03-05
travis: add ppc64-cross test

rppt:
Now when @0x7f454c46 did the heavy lifting of resolving compel headers mess, it's easy to add another cross-build target :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -90,6 +90,10 @@ jobs:
       arch: amd64
       env: TR_ARCH=aarch64-cross
       dist: bionic
+    - os: linux
+      arch: amd64
+      env: TR_ARCH=ppc64-cross
+      dist: bionic
   allow_failures:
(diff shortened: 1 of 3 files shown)


0x7f454c46:
I like the idea - cross-compilation takes ~2-3 mins, so it shouldn't be critical from one side and from the other there were times when Travis's ppc platform job had bad times.

0x7f454c46:
Also, there _actually_ seems to be people using cross-compilation for CRIU :-)

avagin:
It is interesting whether there is a project which has more travis jobs than CRIU or not.

0x7f454c46:
@avagin we're far from Apache:
https://arxiv.org/pdf/1904.09416.pdf

avagin:
@rppt could you rebase this pr?

rppt:
> @rppt could you rebase this pr?

done

PR #980, opened 2020-03-10
crit-recode: skip (not try to parse) nftables raw image

mihalicyn:
Reported-by: Mr Jenkins

fix tests https://ci.openvz.org/job/CRIU/job/CRIU-crit/job/criu-dev/6027/consoleText

### DIFF
--- a/test/crit-recode.py
+++ b/test/crit-recode.py
@@ -47,6 +47,8 @@ for imgf in find.stdout.readlines():
         continue
     if imgf_b.startswith(b'ip6tables-'):
         continue
+    if imgf_b.startswith(b'nftables-'):
+        continue
     if imgf_b.startswith(b'route-'):
         continue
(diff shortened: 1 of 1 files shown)


adrianreber:
Looks correct. Some explanation would have been helpful why these type of images are skipped. If I understand it correctly it is because the image file for nftables (as defined in criu/include/magic.h) are of the type `RAW_IMAGE_MAGIC`.

mihalicyn:
> Looks correct. Some explanation would have been helpful why these type of images are skipped. If I understand it correctly it is because the image file for nftables (as defined in criu/include/magic.h) are of the type `RAW_IMAGE_MAGIC`.

Yes that's it. I've written about "raw image" in the first line of commit message, but I will add some information to commit description section too.

Thanks.
Alex.

PR #981, opened 2020-03-11
autofs: fix integer overflow in mount options parsing

mihalicyn:
In real life cases pipe_ino param could be larger that INT_MAX,
but in autofs_parse() function we using atoi function, that uses
4 byte integers. It's a bug.

Example of mount info from real case:
(00.508286) 	type autofs source /etc/auto.misc mnt_id 2824 s_dev 0x4b9 / @
./misc flags 0x300000 options fd=5,pipe_ino=3480845226,pgrp=95929,timeout=300,
minproto=5,maxproto=5,indirect

3480845226 > 2147483647 (32-bit wide signed int max value) => we have a problem

It causes a error:
(03.195915) Error (criu/pipes.c:529): The packetized mode for pipes is not supported yet

Signed-off-by: Alexander Mikhalitsyn (Virtuozzo) <alexander@mihalicyn.com>

### DIFF
--- a/criu/autofs.c
+++ b/criu/autofs.c
@@ -62,25 +62,53 @@ int autofs_parse(struct mount_info *pm)
 {
 	long pipe_ino = AUTOFS_OPT_UNKNOWN;
 	char **opts;
-	int nr_opts, i;
+	int nr_opts, i, ret;
 
 	split(pm->options, ',', &opts, &nr_opts);
 	if (!opts)
 		return -1;
+
 	for (i = 0; i < nr_opts; i++) {
 		if (!strncmp(opts[i], "pipe_ino=", strlen("pipe_ino=")))
-			pipe_ino = atoi(opts[i] + strlen("pipe_ino="));
+			if (xatol(opts[i] + strlen("pipe_ino="), &pipe_ino)) {
+				pr_err("pipe_ino (%s) mount option parse failed\n", opts[i] + strlen("pipe_ino="));
+				ret = -1;
+				goto free;
+			}
 	}
-	for (i = 0; i < nr_opts; i++)
-		xfree(opts[i]);
-	free(opts);
+
+	/*
+	 * We must inform user about bug if pipe_ino is greater than UINT32_MAX,
+	 * because it means that something changed in Linux Kernel virtual fs
+	 * inode numbers generation mechanism. What we have at the moment:
+	 * 1. struct inode i_ino field (include/linux/fs.h in Linux kernel)
+	 * has unsigned long type.
+	 * 2. get_next_ino() function (fs/inode.c), that used for generating inode
+	 * numbers on virtual filesystems (pipefs, debugfs for instance)
+	 * has unsigned int as return type.
+	 * So, it means that ATM it is safe to keep uint32 type for pipe_id field
+	 * in pipe-data.proto.
+	 */
+	if (pipe_ino > UINT32_MAX) {
+		pr_err("overflow: pipe_ino > UINT32_MAX\n");
+		ret = -1;
+		goto free;
+	}
 
 	if (pipe_ino == AUTOFS_OPT_UNKNOWN) {
 		pr_warn("Failed to find pipe_ino option (old kernel?)\n");
-		return 0;
+		ret = 0;
+		goto free;
 	}
 
-	return autofs_gather_pipe(pipe_ino);
+	ret = autofs_gather_pipe(pipe_ino);
+
+free:
+	for (i = 0; i < nr_opts; i++)
+		xfree(opts[i]);
+	xfree(opts);
+
+	return ret;
 }
 
 static int autofs_check_fd_stat(struct stat *stat, int prgp, int fd,


mihalicyn:
Unfortunately, I couldn't provide real applicable test-case for that. But I've managed to reproduce the bug.

Crazy program that makes internal kernel ino counter sufficiently big:
```
#include <stdio.h>
#include <unistd.h>
#include <sys/stat.h>
#include <stdlib.h>

unsigned int get_last_pipe_ino()
{
    int pipefd[2];
    struct stat sb;

    if (pipe(pipefd) == -1)
    {
        perror("pipe");
        exit(1);
    }

    if (fstat(pipefd[0], &sb) == -1)
    {
        perror("fstat");
        exit(1);
    }

    close(pipefd[0]);
    close(pipefd[1]);

    return sb.st_ino;
}

int main(int argc, char *argv[])
{
    int pipefd[2];
    struct stat sb;
    unsigned int start_ino, end_ino;
    int need_ino_to_overflow;

    start_ino = get_last_pipe_ino();
    need_ino_to_overflow = __INT_MAX__ - (int) start_ino;

    if (need_ino_to_overflow <= 0)
    {
        printf("Inode numbers already larger that INT_MAX\n");
        exit(0);
    }

    printf("start ino: %u to overflow: %u\n", start_ino, need_ino_to_overflow);

    #pragma omp parallel for private(pipefd) num_threads(6)
    for (unsigned int i = 0; i < need_ino_to_overflow; i++)
    {
        if (pipe(pipefd) == -1)
        {
            perror("pipe");
            exit(1);
        }

        close(pipefd[0]);
        close(pipefd[1]);
    }

    end_ino = get_last_pipe_ino();

    printf("end_ino: %u ino diff: %u INT MAX: %d overflowed(?): %d\n", end_ino, (end_ino - start_ino), __INT_MAX__, (end_ino > __INT_MAX__));

    return 0;
}
```

```
gcc -fopenmp fill_ino.c -o fill_ino
./fill_ino
```

Test:
```
sudo ./test/zdtm.py run -t zdtm/static/autofs
```

Result:
```
========================= Run zdtm/static/autofs in ns =========================
Start test
Test is SUID
./autofs --pidfile=autofs.pid --outfile=autofs.out --dirname=autofs.test
Run criu dump
=[log]=> dump/zdtm/static/autofs/58/1/dump.log
(00.012542) Dumping pipe 22 with id 0x11 pipe_id 0x80b374b4
(00.012544) Dumping data from pipe 0x80b374b4 fd 22
(00.012561) 64 fdinfo 9: pos:                0 flags:            40001/0
(00.012565) Dumping pipe 23 with id 0x12 pipe_id 0x80b374b4
(00.012567) Error (criu/pipes.c:504): The packetized mode for pipes is not supported yet
(00.012572) ----------------------------------------
(00.012579) Error (criu/cr-dump.c:1348): Dump files (pid: 64) failed with -1
(00.015469) 	Unseizing 58 into 1
(00.015474) 	Unseizing 62 into 1
(00.015483) 	Unseizing 63 into 1
(00.015489) 	Unseizing 64 into 1
(00.015508) Error (criu/cr-dump.c:1775): Dumping FAILED.
################## Test zdtm/static/autofs FAIL at CRIU dump ###################
```

criu/autofs.c
```diff
@@ -62,25 +62,53 @@ int autofs_parse(struct mount_info *pm)
...
+	 * I know, you want to ask "Why we simply not change type in image to uint64"? ;)
+	 * Let me explain this moment. We need to look at fs/inode.c in Linux kernel.
+	 * There is a get_next_ino() function, take a look on it's return type =>
+	 * currently, it has unsigned int type. Looks weird. struct inode i_ino has
+	 * unsigned long type, but get_next_ino() returns only unsigned int.
+	 * (Read big comment above get_next_ino in Linux kernel for explanation)
+	 * 
+	 * So, we don't need to do anything with our images at that moment.
+	 */
+	BUG_ON(pipe_ino > UINT32_MAX);
```
Snorch: What if pipe_ino < 0 (big enough unsigned long can overflow long)? And may be we need the same BUG_ON in parse_options() too.

0x7f454c46: Why should it be `BUG_ON()` in the first place?
Instead of crashing CRIU, could we return error-code and handle it?

mihalicyn: > What if pipe_ino < 0 (big enough unsigned long can overflow long)? And may be we need the same BUG_ON in parse_options() too.

Good point, but it doesn't possible at the moment, because:
https://github.com/torvalds/linux/blob/ac309e7744bee222df6de0122facaf2d9706fa70/fs/autofs/inode.c#L93

I'm not sure, should we or not to make checks for that case. It's up to you.

mihalicyn: @0x7f454c46 thanks, I've replaced `BUG_ON(..)` to appropriate check

criu/autofs.c
```diff
@@ -62,25 +62,57 @@ int autofs_parse(struct mount_info *pm)
...
+	}
+
+	/*
+	 * We must inform user about bug if pipe_ino
+	 * is greater than UINT32_MAX, because, from kernel side of view
+	 * it is possible situation, because i_ino field in struct inode
+	 * (look at include/linux/fs.h in Linux kernel) has unsigned long type.
+	 * But, at the moment, pipe_id field in pipe-data.proto has uint32 type.
+	 *
+	 * I know, you want to ask "Why we simply not change type in image to uint64"? ;)
```
avagin: Pls rephrase this comment in more formal way without smiles ;).

criu/autofs.c
```diff
@@ -62,25 +62,57 @@ int autofs_parse(struct mount_info *pm)
...
+				pr_err("pipe_ino (%s) mount option parse failed\n", opts[i] + strlen("pipe_ino="));
+				ret = -1;
+				goto free;
+			}
+	}
+
+	/*
+	 * We must inform user about bug if pipe_ino
+	 * is greater than UINT32_MAX, because, from kernel side of view
+	 * it is possible situation, because i_ino field in struct inode
```
avagin: From kernel side of view, it is impossible situation. Despite the unsigned long  type of i_no, it can't be greater than UINT_MAX, because pipe inode numbers are generated by get_next_ino() which return type is uint.

PR #990, opened 2020-03-19
Moved sigact dump/restore code into separate file

prakritigoyal19:
In this patch, I have moved the sigact dump/restore code into a new file. I have created the file sigaction.c for the same. I removed the function parasite_dump_sigacts_seized() from parasite-syscall.c and prepare_sigactions() and everything it calls from cr-restore.c.

In parasite-syscall.h, I have removed the call for the function deleted in parasite-syscall.c, parasite_dump_sigacts_seized().

I have included the sigaction.h in cr-dump.c, as they call these functions.

I have added a line in Makefile.crtools to create an object file sigaction.o


criu/cr-dump.c
```diff
@@ -1,52 +1,18 @@
...
-#include <fcntl.h>
-
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <sys/vfs.h>
-#include <sys/time.h>
-#include <sys/wait.h>
-
-
-#include <sched.h>
```
Snorch: Oh wow, I **didn't** mean to ask you to remove all the unused headers in all the existing files =). (Though it might be a matter of another patch-set.)

I mean to remove all unused headers **you add** in you patch-set **only**.

Also you should merge fixes to your original commits into your original commits. In the patch-set each commit should be "good" there is no such a situation where first commit is breaking something and second commit is fixing (cleanup-ing as in your case) the first one.

prakritigoyal19: Oh, sorry. I was not aware of these guidelines, though I have learned a lot in the past few days. Thanks a lot for your inputs. 
Just to clarify, should I restore the headers which I have removed or let the file be as it is now?
Also, should I revert to the previous three commits and send a fresh 'good' commit?

criu/cr-dump.c
```diff
@@ -1,52 +1,18 @@
...
-#include <signal.h>
-#include <limits.h>
-#include <unistd.h>
-#include <errno.h>
-#include <string.h>
-
-#include <fcntl.h>
-
-#include <sys/types.h>
-#include <sys/stat.h>
```
avagin: This is wrong. fstat is used in this file.
https://linux.die.net/man/2/fstat
```
$ cat criu/cr-dump.c  | grep fstat
	if (fstat(fd, &p->stat) < 0) {
		pr_perror("Can't fstat exe link");
	if (fstatfs(lfd, &fst)) {
```

criu/cr-dump.c
Snorch: 1) Let's restore. So If your patchset as a whole **adds** some new headers then just check if they (new ones) are really required.
2) You can merge your original commit and all fixup changes together in a single fresh commit in the same branch an push it with force "-f" to replace commits in this pull-request.

Glad to help!

criu/sigaction.c
```diff
@@ -0,0 +1,308 @@
...
+static int restore_native_sigaction(int sig, SaEntry *e)
+{
+	rt_sigaction_t act;
+	int ret;
+
+	ASSIGN_TYPED(act.rt_sa_handler, decode_pointer(e->sigaction));
+	ASSIGN_TYPED(act.rt_sa_flags, e->flags);
+	ASSIGN_TYPED(act.rt_sa_restorer, decode_pointer(e->restorer));
+	BUILD_BUG_ON(sizeof(e->mask) != sizeof(act.rt_sa_mask.sig));
+	memcpy(act.rt_sa_mask.sig, &e->mask, sizeof(act.rt_sa_mask.sig));
```
avagin, line 94: memcpy requires to include string.h:
http://man7.org/linux/man-pages/man3/memcpy.3.html

criu/sigaction.c
```diff
@@ -0,0 +1,308 @@
...
+	}
+
+	if (sa_inherited(sig - 1, &act))
+		return 1;
+
+	/*
+	 * A pure syscall is used, because glibc
+	 * sigaction overwrites se_restorer.
+	 */
+	ret = syscall(SYS_rt_sigaction, sig, &act, NULL, sizeof(k_rtsigset_t));
```
avagin, line 108: syscall requires to include unistd.h and sys/syscall.h:
http://man7.org/linux/man-pages/man2/syscall.2.html

pls review other external calls and add headers that are required for them.

criu/sigaction.c
Snorch: They are implicitly included via parasite.h, but to be explicit we need to add them.

avagin: We are trying to not look at implicitly included headers. When we use a function from glibc or from other third-party libraries, we include all headers explicitly that are required for it.

criu/sigaction.c
```diff
@@ -1,3 +1,11 @@
+#include <stdio.h>
```
avagin, line 1: you need this merge these changes into the first commit.

criu/sigaction.c
```diff
@@ -1,3 +1,11 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+#include <errno.h>
+#include <err.h>
```
avagin, line 7: why do we need this err.h?

xemul:
The pr is heavily outdated. Please, rebase and open a new one of needed.

PR #992, opened 2020-03-20
Fix deadlock for ghost DGRAM unix socket

mihalicyn:
This patch fixes deadlock that appears on ghost DGRAM unix sockets.
Problem is that wake_connected_sockets() function *should* be called
strictly after fle->stage >= FLE_OPEN.

Explanation:
Consider situation, we have ghost unix DGRAM socket (peer socket),
and also have several sockets that connected to this peer socket.

How restore of that picture works?
In files subsystem we have open_fdinfos(pstree_item*) function that calls open_fd()
function for *every* fd of task. open_fd() function calls appropriate
file descriptor "open" handler that may return "1" which means "try again later".
This retcode means, that some additional resources is needed to fully restore file
descriptor. For *ghost* UNIX sockets, for instance, we need to have peer socket
file descriptor *before* we can open and restore client sockets. Here is the main problem.
open_fdinfos() called from separate tasks simultaneously, so, when we get "1" retcode
we stay on futex (wait_fds_event() function) and waiting for someone another task
restore some resource and notify us that we can retry opening of file descriptor.

With *ghost* UNIX socket I've managed to caught next behaviour.
1. From one task (that holds client socket) open_fdinfos() called open_fd() that called
open_unixsk_standalone(). In open_unixsk_standalone we have check that means
"if socket have peer and that peer is GHOST and that peer fle->stage < FLE_OPEN"
return "try again". Ok. So, this task will stay on wait_fds_event().

2. Second task, that holds *peer* tried to open peer socket fd. So,
it also calls open_fd() -> open_unixsk_standalone() -> opening socket
-> bind_unix_sk() -> in bind_unix_sk we have call to wake_connected_sockets().
So, after that call we will "wake up" task from first point and it may proceed
fd restoring. Yes? No. In first point we need to "peer_fle->stage >= FLE_OPEN"
but fle->stage of our peer socket will become FLE_OPEN in open_fd(). After we
return from open_unixsk_standalone we proceed to setup_and_serve_out() where we have
appropriate stage change.
Between call of wake_connected_sockets and moment when we set stage to FLE_OPEN
should pass very small amount of time. But it may happen, so we "wake up"
tasks that holds client sockets but did not have enough time to change fle->stage
to FLE_OPEN. Exactly that case I've managed to reproduce.
(Really, ossec-hids application managed to reproduce this problem at first %) )

This patch changes place, where we call wake_connected_sockets(). After
introducing callback in file_desc_ops on_stage_change it's very easy.

### DIFF
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -1845,7 +1845,7 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
 {
 	struct unix_sk_info *queuer = ui->queuer;
 	struct unix_sk_info *peer = ui->peer;
-	struct fdinfo_list_entry *fle, *fle_peer;
+	struct fdinfo_list_entry *fle;
 	int sk;
 
 	fle = file_master(&ui->d);
@@ -1859,10 +1859,12 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
 	 * into it in a special way.
 	 */
 	if (peer && (peer->flags & USK_GHOST_FDSTORE)) {
-		fle_peer = file_master(&peer->d);
-		if (fle_peer->stage < FLE_OPEN) {
+		/*
+		 * Here we return 1 which says "try again" to
+		 * the caller if peer is not yet bounded.
+		 */
+		if (peer_is_not_prepared(peer))
 			return 1;
-		}
 	}
 
 	if (fle->stage == FLE_OPEN)
--- a/test/zdtm/static/sk-unix-dgram-ghost.c
+++ b/test/zdtm/static/sk-unix-dgram-ghost.c
@@ -1,0 +1,230 @@
+#ifndef _GNU_SOURCE
+# define _GNU_SOURCE
+#endif
+
+#include <stdlib.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <string.h>
+#include <stdbool.h>
+#include <limits.h>
+#include <errno.h>
+#include <dirent.h>
+#include <sys/wait.h>
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+#include <sys/stat.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc	= "Check data of bound ghost DGRAM unix socket and possibility to connect";
+const char *test_author	= "Alexander Mikhalitsyn <alexander@mihalicyn.com>";
+
+/*
+ * PROCESS_NUM | FREEZE_FREQ
+ * 3           | 1 / 5
+ * 4           | 1 / 5
+ * 5           | 2 / 5
+ * 10          | 10 / 10
+ */
+#define PROCESSES_NUM 10
+
+#define MSG "hello"
+char filename[PATH_MAX];
+char *dirname;
+TEST_OPTION(dirname, string, "directory name", 1);
+
+static int fill_sock_name(struct sockaddr_un *name, const char *filename)
+{
+	char *cwd;
+
+	cwd = get_current_dir_name();
+	if (strlen(filename) + strlen(cwd) + 1 >= sizeof(name->sun_path)) {
+		pr_err("Name %s/%s is too long for socket\n",
+		       cwd, filename);
+		return -1;
+	}
+
+	name->sun_family = AF_LOCAL;
+	ssprintf(name->sun_path, "%s/%s", cwd, filename);
+	return 0;
+}
+
+static int client(int i, task_waiter_t t)
+{
+	struct sockaddr_un addr;
+	int sk;
+
+	sk = socket(AF_UNIX, SOCK_DGRAM, 0);
+	if (sk < 0) {
+		pr_perror("open client %d", i);
+		return 1;
+	}
+
+	if (fill_sock_name(&addr, filename) < 0) {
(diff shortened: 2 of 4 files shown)


test/zdtm/static/sk-unix-dgram-ghost.c
```diff
@@ -0,0 +1,220 @@
...
+	task_waiter_complete(&t, 3);
+
+	return 0;
+}
+
+void child_exited(int signo)
+{
+	int status;
+
+	if (wait(&status) == -1) {
```
avagin, line 100: you need to call wait in a loop because sigchld from a few children can be merged.

test/zdtm/static/sk-unix-dgram-ghost.c
```diff
@@ -0,0 +1,220 @@
...
+
+	test_msg("child %d: MSG was sent\n", i);
+
+	/* notify server that we sent data */
+	task_waiter_complete(&t, 3);
+
+	return 0;
+}
+
+void child_exited(int signo)
```
avagin, line 96: static void ...

avagin:
Will this patch fix the issue?
```
index 048ff44ae..d8466948f 100644
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -1758,7 +1758,7 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
 {
        struct unix_sk_info *queuer = ui->queuer;
        struct unix_sk_info *peer = ui->peer;
-       struct fdinfo_list_entry *fle, *fle_peer;
+       struct fdinfo_list_entry *fle;
        int sk;
 
        fle = file_master(&ui->d);
@@ -1772,8 +1772,7 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
         * into it in a special way.
         */
        if (peer && (peer->flags & USK_GHOST_FDSTORE)) {
-               fle_peer = file_master(&peer->d);
-               if (fle_peer->stage < FLE_OPEN) {
+               if (peer->bound == 0) {
                        return 1;
                }
        }
```

test/zdtm/static/sk-unix-dgram-ghost.c
mihalicyn: Thanks. Fixed!

test/zdtm/static/sk-unix-dgram-ghost.c
mihalicyn: Sure. Fixed!

mihalicyn:
@avagin I've taken your patch as fix instead of my first version. Because your patch is simpler and shorter.
Thanks!

Just to keep history, I've attached patch files of the first fix version.

[0002-sk-unix-wake-connected-sockets-strictly-after-fle-st.patch.txt](https://github.com/checkpoint-restore/criu/files/5147399/0002-sk-unix-wake-connected-sockets-strictly-after-fle-st.patch.txt)
[0001-files-add-on_stage_change-callback-to-file_desc_ops.patch.txt](https://github.com/checkpoint-restore/criu/files/5147400/0001-files-add-on_stage_change-callback-to-file_desc_ops.patch.txt)

criu/sk-unix.c
```diff
@@ -1787,10 +1787,12 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
 	 * into it in a special way.
 	 */
 	if (peer && (peer->flags & USK_GHOST_FDSTORE)) {
-		fle_peer = file_master(&peer->d);
-		if (fle_peer->stage < FLE_OPEN) {
+		/*
+		 * Here we return 1 which says "try again" to
+		 * the caller if peer is not yet bounded.
+		 */
+		if (peer->bound == 0)
```
Snorch, line 1794: Probably it should be "if (peer_is_not_prepared(peer)) return 1;".

Snorch (changes requested):

criu/sk-unix.c
mihalicyn: Yes, technically it will be more correct. I've also tried to reproduce deadlock with STREAM unix socket, but the design on interaction with stream sockets implies that we open "connection" sockets *after* server socket. And this circumstance is very important and breaks possible deadlock reproduce. I even put a comment in zdtm test which says that to reproduce a deadlock we need to have a situation where peer socket is the *last* file descriptor of the holding process. I've also tried to "move" file descriptor using `dup2` but it doesn't help me to reproduce the problem.

mihalicyn: Fixed

test/zdtm/static/sk-unix-dgram-ghost.c
```diff
@@ -0,0 +1,230 @@
...
+	 * So, if we for example, move task_waiter_init() after server
+	 * socket creation we loose reproduce.
+	 */
+	srv = socket(AF_UNIX, SOCK_DGRAM | SOCK_NONBLOCK, 0);
+	if (srv < 0) {
+		pr_perror("open srv");
+		ret = 1;
+		goto clean;
+	}
+
```
mihalicyn, line 165: Comment: if we add open of *any* file descriptor we will lose reproduce.
```
int fds[2]; 
//...
pipe(fds);
```
will break, and so on.

Snorch (approved):

avagin:
What is going on with this PR? Could you rebase it?

mihalicyn:
gentle ping @avagin

PR #998, opened 2020-03-24
memfd: save all memfd inodes in one image

avagin:
Per-object image is acceptable if we expect to have 1-3 objects
per-container. If we expect to have more objects, it is better to save
them all into one image. There are a number of reasons for this:
* We need fewer system calls to read all objects from one image.
* it is faster to save or move one image.


avagin:
@nviennot, could you review this? Thanks!

criu/memfd.c
```diff
@@ -235,20 +227,58 @@ static struct memfd_inode *memfd_alloc_inode(int id)
...
+
+	return 0;
+}
+
+static struct collect_image_info memfd_inode_cinfo = {
+	.fd_type = CR_FD_MEMFD_INODE,
+	.pb_type = PB_MEMFD_INODE,
+	.priv_size = sizeof(struct memfd_inode_node),
+	.collect = collect_one_memfd_inode,
+	.flags = COLLECT_NOFREE,
```
mihalicyn: nit: it's not so important (CRIU works for a small amount of time), but as far as I understand from implementation https://github.com/checkpoint-restore/criu/blob/8cf5f298ebefe49e2977769039849348dc199e9f/criu/protobuf.c#L244
`COLLECT_NOFREE` implies not free memory after collecting object from pb image. As I see from code we using pointers to `MemfdInodeEntry`, so, we need to have a valid pointer for a long time. But when we finished with memfd restoration shouldn't we call `memfd_inode_entry__free_unpacked` somewhere? It looks like we currently not have the general infrastructure of "destructors" for that.

avagin: We don't want to spend time to release these objects because from the moment when we know that we don't need them to the moment when the criu process will die is just one step of resuming processes.

criu/memfd.c
```diff
@@ -235,20 +227,58 @@ static struct memfd_inode *memfd_alloc_inode(int id)
 	return inode;
 }
 
-extern int restore_memfd_shm(int fd, u64 id, u64 size);
+static LIST_HEAD(inode_list);
```
nviennot: We already a list of inodes that we maintain during restore. Let's reuse it instead of creating a second one. Here's a commit that implement this idea: https://github.com/nviennot/criu/commit/c6ddea4eb119bde1e1e99eb81002ab2603eecfea

Note that the `struct memfd_inode` has a union for dump and restore parts. It doesn't really help us. The only common parts of the restore and dump inode are the list_head node, and the inode id. Might as well have two different structs `struct memfd_dump_inode` and `struct memfd_restore_inode`. Code would be clearer. What do you think?

avagin: Thank you for the feedback. I addressed both your suggestions.

PR #1002, opened 2020-03-26
Add support for time namespaces

avagin:
The time namespace allows for per-namespace offsets to the system
monotonic and boot-time clocks.

C/R of time namespaces are very straightforward. On dump, criu enters a
target time namespace and dumps currents clocks values, then on restore,
criu creates a new namespace and restores clocks values.


criu/cr-check.c
```diff
@@ -1266,6 +1266,14 @@ static int check_kcmp_epoll(void)
...
 }
 
+static int check_time_namespace(void)
+{
+	if (!kdat.has_timens)
+		return -1;
+
+	return 0;
+}
+
```
0x7f454c46: Nit: as kerndat is cached, you probably want to print here something if timens is unsupported by kernel.
Also, I think it's worth to call this in `cr_check()` for checking it in `criu check --all`

test/zdtm/lib/ns.c
```diff
@@ -207,6 +209,36 @@ static void ns_sig_hand(int signo)
...
+
+        if (write(fd, buf, len) != len) {
+                fprintf(stderr, "/proc/self/timens_offsets: %m");
+		return -1;
+	}
+
+        close(fd);
+
+        return 0;
+}
```
0x7f454c46: nit: something with the indention here: are those spaces?
also, fprintf() stands out

test/zdtm/lib/ns.c
```diff
@@ -207,6 +209,36 @@ static void ns_sig_hand(int signo)
...
+        char buf[4096];
+
+        if (clk_id == CLOCK_MONOTONIC_COARSE || clk_id == CLOCK_MONOTONIC_RAW)
+                clk_id = CLOCK_MONOTONIC;
+
+        len = snprintf(buf, sizeof(buf), "%d %ld 0", clk_id, offset);
+
+        fd = open("/proc/self/timens_offsets", O_WRONLY);
+        if (fd < 0) {
+                fprintf(stderr, "/proc/self/timens_offsets: %m");
```
0x7f454c46: nit: I'd hint what has failed
"open(/proc/self/timens_offsets): %m"

test/zdtm/lib/ns.c
```diff
@@ -207,6 +209,36 @@ static void ns_sig_hand(int signo)
...
+		return -1;
+	}
+
+        if (write(fd, buf, len) != len) {
+                fprintf(stderr, "/proc/self/timens_offsets: %m");
```
0x7f454c46: nit: "write(/proc/self/timens_offsets): %m"
`close(fd)`

test/zdtm_ct.c
```diff
@@ -5,6 +5,69 @@
...
+
+        if (write(fd, buf, len) != len) {
+                fprintf(stderr, "/proc/self/timens_offsets: %m");
+		return -1;
+	}
+
+        close(fd);
+
+        return 0;
+}
```
0x7f454c46: ditto

test/zdtm/static/time.c
```diff
@@ -0,0 +1,47 @@
...
+		a = ts.tv_sec * NSEC_PER_SEC + ts.tv_nsec;
+		b = tss[i].tv_sec * NSEC_PER_SEC + tss[i].tv_nsec;
+		if (a < b) {
+			fail("%d: %lld %lld", clocks[i], a, b);
+			return 1;
+		}
+		if (a > b + 60 * 60 * NSEC_PER_SEC) {
+			fail("%d: %lld %lld", clocks[i], a, b);
+			return 1;
+		}
```
0x7f454c46: I'd say hour is good to prevent false-positives, but this won't fail without timens (as tests finish earlier than in an hour).
What do you think if the test would rather check that the date is far in the future, like gettimeofday() + 200days.
And using crit we could add those 200 days into the image.
Just an idea :)

test/zdtm/static/timens_nested.c
```diff
@@ -0,0 +1,53 @@
...
+		return -1;
+	}
+
+	if (pid == 0) {
+		while (1)
+			sleep(1);
+		return 0;
+	}
+
+
```
Snorch: Double '\n" again

test/zdtm/static/timens_nested.c
```diff
@@ -0,0 +1,53 @@
...
+int main(int argc, char **argv)
+{
+	pid_t pid;
+
+	test_init(argc, argv);
+
+	if (unshare(CLONE_NEWTIME)) {
+		pr_perror("unshare");
+		return 1;
+	}
```
Snorch: If we only have /proc/pid/ns/time_for_children set without fork below after migration it would be lost AFAICS.

adrianreber:
If clone3() ever gets the possibility to clone a process with an offset directly would it make sense from your point of view to use it? Or just do it always like in this PR?

avagin:
@adrianreber yes, it will.

test/zdtm/static/time.c
avagin: For ns and uns flavors, we run the test in a separate time namespace with 10 and 20 hours offsets for monotonic and boottime clocks, so if we don't dump/restore time namespace, this test will fail.

PR #1008, opened 2020-04-01
pstree: improve error handling in read_pstree_image

Snorch:
First don't free pstree_item as they are allocated with shmalloc on
restore. Second always pstree_entry__free_unpacked PstreeEntry. Third
remove all breaks replacing them with implict goto err, so that it would
be easier to understand that we are on error path.

Sadly there is no much use in xfree-ing pi->threads because in case of
an error we still have ->threads unfreed from previous entries anyway.

But at least some cleanup can be done here.

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -498,11 +498,109 @@ static int read_pstree_ids(struct pstree_item *pi)
 	return 0;
 }
 
+/*
+ * Returns <0 on error, 0 on eof and >0 on successful read
+ */
+static int read_one_pstree_item(struct cr_img *img, pid_t *pid_max)
+{
+	struct pstree_item *pi;
+	PstreeEntry *e;
+	int ret, i;
+
+	ret = pb_read_one_eof(img, &e, PB_PSTREE);
+	if (ret <= 0)
+		return ret;
+
+	ret = -1;
+	pi = lookup_create_item(e->pid);
+	if (pi == NULL)
+		goto err;
+	BUG_ON(pi->pid->state != TASK_UNDEF);
+
+	/*
+	 * All pids should be added in the tree to be able to find
+	 * free pid-s for helpers. pstree_item for these pid-s will
+	 * be initialized when we meet PstreeEntry with this pid or
+	 * we will create helpers for them.
+	 */
+	if (lookup_create_item(e->pgid) == NULL)
+		goto err;
+	if (lookup_create_item(e->sid) == NULL)
+		goto err;
+
+	pi->pid->ns[0].virt = e->pid;
+	if (e->pid > *pid_max)
+		*pid_max = e->pid;
+	pi->pgid = e->pgid;
+	if (e->pgid > *pid_max)
+		*pid_max = e->pgid;
+	pi->sid = e->sid;
+	if (e->sid > *pid_max)
+		*pid_max = e->sid;
+	pi->pid->state = TASK_ALIVE;
+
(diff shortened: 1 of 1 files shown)


nviennot:
1) I would rename the `close` label to `out`. It seems more established as a practice when I grep around in the CRIU codebase.

2) I would prefer the refactor to use `while ((ret = pb_read_one_eof(img, &e, PB_PSTREE)) <= 0)` (a.k.a while there's a pstree entry). I think it would increase readability as it uses the `while` condition semantics.

So how about:

```c
{
    PstreeEntry *e = NULL;
    ...
    while ((ret = pb_read_one_eof(img, &e, PB_PSTREE)) <= 0) {
        ...
    }

out:
    if (e)
        pstree_entry__free_unpacked(e, NULL);
    close_image(img);
    return ret;
err:
    ret = -1;
    goto out;
}
```

We could also migrate pieces of CRIU to Rust, and avoid all this deallocation dance =)

avagin:
> I would rename the close label to out. It seems more established as a practice when I grep around in the CRIU codebase.

+1

> I would prefer the refactor to use while ((ret = pb_read_one_eof(img, &e, PB_PSTREE)) <= 0) (a.k.a while there's a pstree entry). I think it would increase readability as it uses the while condition semantics.

We don't use assignments inside conditional expressions. For me, this doesn't look more readable. I have seen a few times bugs in such expressions which were really hard to find.

nviennot:
> We don't use assignments inside conditional expressions. 

What about something similar to what can be found in `man strtok`. It may look a bit contrived, I'm not sure, but worth considering.

```c
for (ret = pb_read_one_eof(img, &e, PB_PSTREE)); ret > 0;
     ret = pb_read_one_eof(img, &e, PB_PSTREE)) {
    ...
}
```

avagin:
This looks ok for me, but it is up to Pavel what to choose.

Snorch:
Thanks a lot for your reviews!

>>>I would prefer the refactor to use while ((ret = pb_read_one_eof(img, &e, PB_PSTREE)) <= 0) (a.k.a while there's a pstree entry). I think it would increase readability as it uses the while condition semantics.
>>We don't use assignments inside conditional expressions. For me, this doesn't look more readable. I have seen a few times bugs in such expressions which were really hard to find.
>What about something similar to what can be found in man strtok.

I personally like ```while ((ret = func()) > 0)``` way, but I see how errors can be made in this construction.

"For" thing for me does not look better than just:

```
ret = func();
while (ret > 0)
        ret = func();
```

So I updated the pr with the latter.

Moreover it looks nice to separate read_pstree_item helper to handle free and close on error paths much simplier.

nviennot:
That makes a lot of sense to extract the body in a function. Much cleaner.

nitpicks:
1) I would rename `read_pstree_item` into `read_next_pstree_item` to suggest we have an iterator
2) The while loop can be replaced with a `do { ret = func() } while (ret > 0);`. I don't feel strongly about this one.

Snorch:
> That makes a lot of sense to extract the body in a function. Much cleaner.
> 
> nitpicks:
> 
>     1. I would rename `read_pstree_item` into `read_next_pstree_item` to suggest we have an iterator

read_one_pstree_item sounds better for me, is it ok?

> 
>     2. The while loop can be replaced with a `do { ret = func() } while (ret > 0);`. I don't feel strongly about this one.

Year do saves us from pre-loop call.

PR #1010, opened 2020-04-05
Adding --images-dir and --prev-images dir to manual.

### LINKED ISSUE #1009: Add flag `--prev-images-dir` to `page-server`
Hey,

I was running into a problem when doing an iterative migration with `page-server`. Memory tracking seemed to be disabled. After investigating a bit, I came across #553 just to realize that there is a `--prev-images-dir` flag for `page-server` that seems not to be documented.

I think it should be added, I could do that myself.

csegarragonz:
Closes #1009

### DIFF
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -637,6 +637,13 @@ Launches *criu* in page server mode.
     Useful for intercepting page-server traffic e.g. to add encryption
     or authentication.
 
+*--images-dir* 'path'::
+    Use 'path' as a base directory where to store the sets of image files.
+
+*--prev-images-dir* 'path'::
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks for you PR. Not sure it is actually needed. Looking at the man-page I see:
```
   Common options
       Common options are applicable to any command.
```

This includes `-D, --images-dir` and `--prev-images-dir`. So technically it is documented, but maybe the wording could be improved, or mention that the page-server also takes all common options.

csegarragonz:
You are right, I missed that line. So yes it's technically not needed.

Thanks for the quick answer!

PR #1013, opened 2020-04-08
img: fix image_name() when image is empty

nviennot:
When an image is opened but errored with a ENOENT error, the image is still valid. Later on, do_pb_read_one() can fail and will invoke image_name(). The image fd is EMPTY_IMG_FD (-404). read_fd_link fails.

### DIFF
--- a/criu/protobuf.c
+++ b/criu/protobuf.c
@@ -25,8 +25,13 @@ static char *image_name(struct cr_img *img)
 	int fd = img->_x.fd;
 	static char image_path[PATH_MAX];
 
-	if (read_fd_link(fd, image_path, sizeof(image_path)) > 0)
+	if (lazy_image(img))
+		return img->path;
+	else if (empty_image(img))
+		return "(empty-image)";
+	else if (fd >= 0 && read_fd_link(fd, image_path, sizeof(image_path)) > 0)
 		return image_path;
+
 	return NULL;
 }
 
--- a/criu/protobuf.c
+++ b/criu/protobuf.c
@@ -20,16 +20,16 @@
 #include "protobuf.h"
 #include "util.h"
 
-static char *image_name(struct cr_img *img)
+#define  image_name(img, buf) __image_name(img, buf, sizeof(buf))
+static char *__image_name(struct cr_img *img, char *image_path, size_t image_path_size)
 {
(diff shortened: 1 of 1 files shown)


criu/protobuf.c
```diff
@@ -25,7 +25,7 @@ static char *image_name(struct cr_img *img)
 	int fd = img->_x.fd;
 	static char image_path[PATH_MAX];
 
-	if (read_fd_link(fd, image_path, sizeof(image_path)) > 0)
+	if (fd >= 0 && read_fd_link(fd, image_path, sizeof(image_path)) > 0)
```
Snorch, line 28: Maybe we can do something like:

```
if (lazy_image(img)) {
        return img->path;
} else if (!empty_image(img)) {
        if (read_fd_link(fd, image_path, sizeof(image_path)) > 0)
                return image_path;
}

return NULL;
```

nviennot: Excellent suggestion. I made some modifications. I still want to do a `fd >= 0` test just in case we add new image failures.

nviennot:
Also, it occurs to me that the there's a `static char image_path[PATH_MAX]` in this function. It seems that we are wasting 4K of resident memory for no good reason. Should this be addressed?

nviennot:
image_path is now on the stack. I wish the stack space was only reserved when about to print an error message though.

criu/protobuf.c
```diff
@@ -48,14 +47,15 @@ static char *image_name(struct cr_img *img)
 
 int do_pb_read_one(struct cr_img *img, void **pobj, int type, bool eof)
 {
+	char imgp[PATH_MAX];
```
avagin, line 50: could you

nviennot: This was my first choice, but then it makes some of the pr_err to wrap line. Should I still rename?

nviennot: This was my first choice, but then it makes some of the pr_err to wrap line. Should I still rename?

avagin: what do you think about this https://github.com/avagin/criu/commit/797f27605bbd7f9ebf0bd15a5c4e23305fac80dc ?

nviennot: sold

PR #1015, opened 2020-04-08
mem: avoid re-opening CR_FD_PAGES when not needed

nviennot:
This commit introduces an optimization when rsti(t)->vma_io is empty. This optimization provides support for streaming a non-seekable image as CR_FD_PAGES is not re-oepened.

### DIFF
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -1404,6 +1404,20 @@ static int prepare_vma_ios(struct pstree_item *t, struct task_restore_args *ta)
 	struct cr_img *pages;
 
 	/*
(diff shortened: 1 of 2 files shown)


criu/mem.c
```diff
@@ -1403,6 +1403,20 @@ static int prepare_vma_ios(struct pstree_item *t, struct task_restore_args *ta)
 {
 	struct cr_img *pages;
 
+	/*
+	 * We optimize the case when rsti(t)->vma_io is empty.
+	 *
+	 * This is useful for for remote images, where all VMAs are premapped
```
avagin, line 1409: what are remove images?

PR #1016, opened 2020-04-09
img-read: reaching EOF unexpectedly in read_page_img_cache() should fail

nviennot:
Before, reaching EOF too soon would trigger an infinite loop.

### DIFF
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -416,7 +416,11 @@ static int maybe_read_page_img_cache(struct page_read *pr, unsigned long vaddr,
 	pr_debug("\tpr%lu-%u Read page from self %lx/%"PRIx64"\n", pr->img_id, pr->id, pr->cvaddr, pr->pi_off);
 	while (1) {
 		ret = read(fd, buf + curr, len - curr);
-		if (ret < 0) {
+		if (ret == 0) {
+			pr_err("Reached EOF unexpectedly while reading page from image\n");
+			return -1;
+		}
+		else if (ret < 0) {
 			pr_perror("Can't read mapping page %d", ret);
 			return -1;
 		}


criu/pagemap.c
```diff
@@ -416,7 +416,11 @@ static int maybe_read_page_img_cache(struct page_read *pr, unsigned long vaddr,
 	pr_debug("\tpr%lu-%u Read page from self %lx/%"PRIx64"\n", pr->img_id, pr->id, pr->cvaddr, pr->pi_off);
 	while (1) {
 		ret = read(fd, buf + curr, len - curr);
-		if (ret < 0) {
+		if (ret == 0) {
+			pr_err("Reached EOF unexpectedly while reading page from image\n");
+			return -1;
+		}
+		if (ret <= 0) {
```
Snorch, line 423: There is never equality at this point already so "<" is enough. And also please change second "if" to "else if" for code readability. Except that looks good -> adding my Acc.

nviennot: Good catch, I amended the commit.

Thank you for the review.

criu/pagemap.c
```diff
@@ -416,7 +416,11 @@ static int maybe_read_page_img_cache(struct page_read *pr, unsigned long vaddr,
 	pr_debug("\tpr%lu-%u Read page from self %lx/%"PRIx64"\n", pr->img_id, pr->id, pr->cvaddr, pr->pi_off);
 	while (1) {
 		ret = read(fd, buf + curr, len - curr);
-		if (ret < 0) {
+		if (ret == 0) {
+			pr_err("Reached EOF unexpectedly while reading page from image\n");
+			return -1;
+		}
```
Snorch, line 422: Continuation (else if) of the statement should be on the same line as the closing bracket https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces

Snorch: https://github.com/checkpoint-restore/criu/pull/1022

Snorch: https://github.com/checkpoint-restore/criu/pull/1022

PR #1019, opened 2020-04-11
Optimize find_unix_sk_by_ino()

### LINKED ISSUE #339: Optimize find_unix_sk_by_ino()
The routine in question does linear list search, but we may have many unix sockets in the dump. Turn it into a hash or tree.

ZeyadYasser:
1. Generalize sockets hashtable so it could be used in restoration. Now the hashtable stores a void pointer (obj) to entry, which could point to socket_desc or unix_sk_info.
2. Refactor lookup functions names to be more descriptive.
3. Optimize find_unix_sk_by_ino(). Replaced the loop with a hashtable lookup.

Fixes #339

### DIFF
--- a/criu/include/sockets.h
+++ b/criu/include/sockets.h
@@ -16,10 +16,15 @@ struct cr_imgset;
 struct nlmsghdr;
 struct cr_img;
 
+struct socket_table_entry {
+	unsigned int			ino;
+	struct socket_table_entry	*next;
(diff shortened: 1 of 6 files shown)


ZeyadYasser:
I think my implementation leaks memory as the table entries are never freed, I will close the PR and hopefully come up with a solution.

PR #1020, opened 2020-04-12
Optimize find_unix_sk_by_ino()

### LINKED ISSUE #339: Optimize find_unix_sk_by_ino()

ZeyadYasser:
Optimize find_unix_sk_by_ino().
Replaced the linear search with a hashtable lookup.
Fixes #339

### DIFF
--- a/criu/include/sockets.h
+++ b/criu/include/sockets.h
@@ -62,6 +62,8 @@ extern int unix_sk_id_add(unsigned int ino);
 extern int unix_sk_ids_parse(char *optarg);
 extern int unix_prepare_root_shared(void);
 
+extern void init_sk_info_hash(void);
+
 extern int do_dump_opt(int sk, int level, int name, void *val, int len);
 #define dump_opt(s, l, n, f)	do_dump_opt(s, l, n, f, sizeof(*f))
 extern int do_restore_opt(int sk, int level, int name, void *val, int len);
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -903,6 +903,7 @@ struct unix_sk_info {
 	struct unix_sk_info	*peer;
 	struct pprep_head	peer_resolve; /* XXX : union with the above? */
 	struct file_desc	d;
+	struct hlist_node	hash; /* To lookup socket by ino */
 	struct list_head	connected; /* List of sockets, connected to me */
 	struct list_head	node; /* To link in peer's connected list  */
 	struct list_head	scm_fles;
@@ -934,11 +935,25 @@ struct scm_fle {
 #define USK_PAIR_SLAVE		0x2
 #define USK_GHOST_FDSTORE	0x4	/* bound but removed address */
 
+#define SK_INFO_HASH_SIZE	32
+
+static struct hlist_head sk_info_hash[SK_INFO_HASH_SIZE];
+
+void init_sk_info_hash(void)
+{
+	int i;
+
+	for (i = 0; i < SK_INFO_HASH_SIZE; i++)
+		INIT_HLIST_HEAD(&sk_info_hash[i]);
+}
+
 static struct unix_sk_info *find_unix_sk_by_ino(int ino)
 {
 	struct unix_sk_info *ui;
+	struct hlist_head *chain;
 
-	list_for_each_entry(ui, &unix_sockets, list) {
+	chain = &sk_info_hash[ino % SK_INFO_HASH_SIZE];
+	hlist_for_each_entry(ui, chain, hash) {
 		if (ui->ue->ino == ino)
 			return ui;
 	}
@@ -2044,6 +2059,7 @@ static int init_unix_sk_info(struct unix_sk_info *ui, UnixSkEntry *ue)
 	INIT_LIST_HEAD(&ui->node);
 	INIT_LIST_HEAD(&ui->scm_fles);
 	INIT_LIST_HEAD(&ui->ghost_node);
+	INIT_HLIST_NODE(&ui->hash);
 
 	return 0;
 }
@@ -2135,6 +2151,7 @@ static int collect_one_unixsk(void *o, ProtobufCMessage *base, struct cr_img *i)
 		list_add_tail(&ui->ghost_node, &unix_ghost_addr);
 	}
 
+	hlist_add_head(&ui->hash, &sk_info_hash[ui->ue->ino % SK_INFO_HASH_SIZE]);
 	list_add_tail(&ui->list, &unix_sockets);
 	return file_desc_add(&ui->d, ui->ue->id, &unix_desc_ops);
 }
--- a/criu/files.c
+++ b/criu/files.c
@@ -1749,5 +1749,6 @@ struct collect_image_info files_cinfo = {
 int prepare_files(void)
 {
 	init_fdesc_hash();
+	init_sk_info_hash();
 	return collect_image(&files_cinfo);
 }


criu/sk-unix.c
```diff
@@ -2195,6 +2185,7 @@ static int add_fake_queuer(struct unix_sk_info *ui)
 	peer->ue->ino = FAKE_INO;
 	file_desc_add(&peer->d, peer_ue->id, &unix_desc_ops);
 	list_del_init(&peer->d.fake_master_list);
+	add_socket_table_entry(FAKE_INO, (struct socket_table_entry *)peer);
```
mihalicyn, line 2188: I think it doesn't make sense to add fake sockets to sockets hashtable, because all fake sockets have the same inode number (0) and we use fake sockets only as queuers (and we search for it by using `find_queuer_for` function).

ZeyadYasser:
@mihalicyn Thanks for the advice.

criu/sockets.c
```diff
@@ -425,48 +425,63 @@ static int restore_socket_filter(int sk, SkOptsEntry *soe)
...
 
-struct socket_desc *lookup_socket_ino(unsigned int ino, int family)
+struct socket_table_entry *lookup_socket_ino(unsigned int ino)
+{
+	struct socket_table_entry *e;
+
+	for (e = sockets[ino % SK_HASH_SIZE]; e; e = e->next) {
+		if (e->ino == ino)
+			return e;
+	}
```
Snorch, line 437: hlist_for_each_entry

criu/sockets.c
```diff
@@ -425,48 +425,63 @@ static int restore_socket_filter(int sk, SkOptsEntry *soe)
...
-int sk_collect_one(unsigned ino, int family, struct socket_desc *d, struct ns_id *ns)
+void add_socket_table_entry(unsigned ino, struct socket_table_entry *ste)
 {
-	struct socket_desc **chain;
+	struct socket_table_entry **chain;
+
+	ste->ino = ino;
+	chain = &sockets[ino % SK_HASH_SIZE];
+	ste->next = *chain;
+	*chain = ste;
```
Snorch, line 474: hlist_add_head

criu/sk-unix.c
```diff
@@ -894,6 +894,7 @@ int fix_external_unix_sockets(void)
 }
 
 struct unix_sk_info {
+	struct	socket_table_entry ste;
```
Snorch, line 897: hlist_node

criu/include/sockets.h
```diff
@@ -16,12 +16,17 @@ struct cr_imgset;
...
+	struct socket_table_entry	*next;
+};
+
 struct socket_desc {
-	unsigned int		family;
-	unsigned int		ino;
-	struct socket_desc	*next;
-	struct ns_id		*sk_ns;
-	int			already_dumped;
+	struct socket_table_entry	ste;
```
Snorch, line 25: hlist_node

criu/sockets.c
```diff
@@ -425,48 +425,63 @@ static int restore_socket_filter(int sk, SkOptsEntry *soe)
 	return ret;
 }
 
-static struct socket_desc *sockets[SK_HASH_SIZE];
+static struct socket_table_entry *sockets[SK_HASH_SIZE];
```
Snorch, line 428: hlist_head

criu/sockets.c
ZeyadYasser: Thanks for the advice, I will try doing this, just one question.
Should the lookup table be initialized to NULL like [here](https://github.com/checkpoint-restore/criu/blob/328c1d540df827fa5abfbbc2056b7cdcb0648818/criu/files.c#L66)? I am asking because the previous implementation for socket_desc was not initializing the table and the tests passed successfully.
If it needs initialization, I guessing that could be done [here](https://github.com/checkpoint-restore/criu/blob/328c1d540df827fa5abfbbc2056b7cdcb0648818/criu/files.c#L1749), please correct me if I am wrong.

ZeyadYasser: @Snorch 
I faced a problem, I can't use `struct unix_sk_info` from `sockets.c` because it is only defined in `sk_unix.c`.
I am thinking of two solutions:
1. This is a workaround and I think it is ugly:
- Add a function to get the `hlist_head` by ino.
- Then `hlist_for_each_entry` to get the `struct unix_sk_info` entry in `sk_unix.c`.
2. Another solution is leaving the hashtable implementation for `struct socket_desc` as it was originally and creating a table in `sk_unix.c` only for `struct unix_sk_info`, this is not generalized but will work well.

btw I think `container_of` could be used `struct socket_table_entry` if we don't want it to be always be the first member.
maybe something like this:
```c
#define table_entry(ptr, type, member) container_of(ptr,type,member)
```
What is your opinion? Any suggestions?

Snorch: >Should the lookup table be initialized to NULL ...

Technically for file_desc_hash we can skip it because global variables are zero-initialized. But for me personally explicitly using INIT_HLIST_HEAD protects us from any internal changes in hlist (which will likely never happen I  know, but to be on the safe side).

>I am guessing that could be done here ...

Initializing hlists in prepare_files looks like a right place on restore, but you miss dump.

>I faced a problem, I can't use struct unix_sk_info from sockets.c because it is only defined in sk_unix.c. 1. ... 2. ...

I like (2) we don't want both unix_sk_info-s and socket_desc-s in one list anyway, why not separating lists for readability.

You should likely resurect find_unix_sk_by_ino, it will lookup through unix_sk_info-s and lookup_socket_ino will lookup through socket_desc-s. Something like:

```
static struct hlist_head sockets[SK_HASH_SIZE];

struct unix_sk_info {
	...
	struct hlist_node h;
};


static struct unix_sk_info *find_unix_sk_by_ino(int ino)
{
	struct unix_sk_info *ui;

	hlist_for_each_entry(ui, &sockets[ino % SK_HASH_SIZE], h) {
		if (ui->ue->ino == ino)
			...
	}
	...
}

struct socket_desc{
	...
	struct hlist_node h;
};

static struct socket_desc *lookup_socket_ino(int ino)
{
	struct socket_desc *d;

	hlist_for_each_entry(d, &sockets[ino % SK_HASH_SIZE], h) {
		if (d->ino == ino)
			...
	}
	...
}
```

ZeyadYasser: Done, But I think we should keep the PR focused on one thing at a time.
I only optimized ```find_unix_sk_by_ino()```, I think a separate PR should be made to refactor other hashtable implementations, like [this](https://github.com/checkpoint-restore/criu/blob/996a11d708fd5a5fc6166fe8cfaee80095e920a3/criu/sockets.c#L428) and [this](https://github.com/checkpoint-restore/criu/blob/996a11d708fd5a5fc6166fe8cfaee80095e920a3/criu/sk-unix.c#L118).

criu/sk-unix.c
```diff
@@ -2044,13 +2051,18 @@ static int init_unix_sk_info(struct unix_sk_info *ui, UnixSkEntry *ue)
...
 	return 0;
 }
 
 int unix_prepare_root_shared(void)
 {
 	struct unix_sk_info *ui;
+	int i;
+
+	for (i = 0; i < SK_INFO_HASH_SIZE; i++) /* initialize hashtable */
+		INIT_HLIST_HEAD(&sk_info_hash[i]);
```
Snorch, line 2065: Looks like you add to the hlist before hlist initialization:

```
cr_restore_tasks
| crtools_prepare_shared
|   prepare_files
|     collect_image(&files_cinfo)
|       collect_one_file
|         collect_one_file_entry(&unix_sk_cinfo)
|           collect_one_unixsk
|             hlist_add_head(&sk_info_hash)
|
+-restore_root_task
    fork_with_pid
      restore_task_with_children
        if (current->parent == NULL) # root task
          root_prepare_shared
            unix_prepare_root_shared
              INIT_HLIST_HEAD(&sk_info_hash)
```

>Initializing hlists in prepare_files looks like a right place on restore, but you miss dump.

You can add new init_sk_info_hash() helper and call it from there.

criu/sk-unix.c
```diff
@@ -937,8 +942,10 @@ struct scm_fle {
 static struct unix_sk_info *find_unix_sk_by_ino(int ino)
 {
 	struct unix_sk_info *ui;
+	struct hlist_head *chain;
```
Snorch, line 945: Extra space in the end of the line.

ZeyadYasser:
@mihalicyn @Snorch , Is there something missing that I should do to get this merged? It has been 3 weeks.

avagin:
Could you rebase this PR?

ZeyadYasser:
> Could you rebase this PR?

Done

ZeyadYasser:
@avagin any updates?

ZeyadYasser:
@avagin All CI tests are now passing, thanks to @adrianreber .
Are there any changes I need to make?

PR #1026, opened 2020-04-15
kerndat remove duplicate call to kerndat_nsid()

Snorch:
Func kerndat_nsid() is called twice.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1102,8 +1102,6 @@ int kerndat_init(void)
 	if (!ret)
 		ret = kerndat_socket_netns();
 	if (!ret)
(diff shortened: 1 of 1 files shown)


criu/kerndat.c
```diff
@@ -1081,8 +1081,6 @@ int kerndat_init(void)
 		ret = kerndat_tun_netns();
 	if (!ret)
 		ret = kerndat_socket_unix_file();
-	if (!ret)
-		ret = kerndat_nsid();
```
avagin, line 1085: could you remove the next call. It is better to have kerndat_nsid and kerndat_link_nsid close to each other.

Snorch: Done.

PR #1031, opened 2020-04-18
Prepare 3.14 (part3)

avagin:
* Add time namespace support
* A few fixes here and there


avagin:
This is the final part.

avagin:
https://criu.org/Download/criu/3.14

avagin:
I suggest @nviennot to choose the name for this release. The format is "material"/"bird breed". Here is the list of all previous releases https://criu.org/Download/criu

nviennot:
I'm happy and honoured to pick a name!
How about **Platinum Peacock**

avagin:
> How about Platinum Peacock

Sounds good to me. I like that it is something real:

white-peacock

kolyshkin:
@avagin can you maybe add https://github.com/checkpoint-restore/criu/pull/1033? It's cosmetic but soft of important for runc integration tests.

Issue #575, opened 2018-11-19
runc lazy-pages test case needs "readonly": false (Error (criu/mount.c:1119): mnt: Can't create a temporary directory: Read-only file system)


adrianreber:
The test case for lazy-pages in runc still needs:

```
sed -i 's;"readonly": true;"readonly": false;' config.json
```

Without this change I am seeing following error during dumping:
```
# (00.058834) mnt: 	206: 30:/ @ ./sys/fs/cgroup
# (00.058837) mnt: Mount is not fully visible ./sys/fs/cgroup
# (00.058873) mnt: 	mount has children ./sys/fs/cgroup
# (00.058899) Error (criu/mount.c:1119): mnt: Can't create a temporary directory: Read-only file system
# (00.058906) mnt: 	mount is overmounted or has children ./sys/fs/cgroup
# (00.066086) mnt: 	205: 2f:/ @ ./sys
# (00.066099) mnt: 	204: 2a:/ @ ./dev/mqueue
# (00.066132) mnt: 	203: 2e:/ @ ./dev/shm
# (00.069285) mnt: 	202: 2d:/ @ ./dev/pts
# (00.069297) mnt: 	201: 2c:/ @ ./dev
# (00.069300) mnt: Mount is not fully visible ./dev
# (00.069325) mnt: 	mount has children ./dev
# (00.069337) Error (criu/mount.c:1119): mnt: Can't create a temporary directory: Read-only file system
# (00.069341) mnt: 	mount is overmounted or has children ./dev
# (00.076074) mnt: 	200: 2b:/ @ ./proc
# (00.076086) mnt: 	199: 800001:/tmp/busyboxtest/rootfs @ ./
# (00.076106) Dumping file-locks
```

Is this something we can fix in CRIU?

avagin:
Could you try the last version of criu? I think @Snorch fixed this problem.

adrianreber:
Latest criu-dev fails with for the lazy-pages runc test case:
```
# (00.003799) Collecting netns 9/8227
# (00.003801) Switching to 8227's net for collecting sockets
# (00.003857) unix: 	Collected: ino 510676 peer_ino 0 family    1 type    5 state  7 name (null)
# (00.003967) Error (criu/libnetlink.c:27): ERROR -2 reported by netlink (No such file or directory)
# (00.003972) Warn  (criu/sockets.c:791): sockets: The current kernel doesn't support ipv4 raw_diag module
[...]

Issue #520, opened 2018-07-03
Regression since release 3.8


adrianreber:
Starting with release 3.8 I see errors in runc using CRIU with read-only root.

(00.098979) mnt: 	217: 3a:/ @ ./sys/fs/cgroup
(00.098981) mnt: Mount is not fully visible ./sys/fs/cgroup
(00.099005) mnt: 	mount has children ./sys/fs/cgroup
(00.099022) Error (criu/mount.c:1095): mnt: Can't create a temporary directory: Read-only file system
(00.099042) Error (criu/mount.c:1545): mnt: Unable to dump a file system for 217:./sys/fs/cgroup

If runc would upgrade their criu version in CI it would also fail there. Using CRIU 3.7 this error does not happen.

One workaround is to tell runc to mount /tmp as read-write tmpfs and then it works.

avagin:
@adrianreber Could you bisect a commit where it was broken?

adrianreber:
git bisect tells me

b6cfb1ce29485186de32ff5a30a56b471d5cb884 is the first bad commit

```
commit b6cfb1ce29485186de32ff5a30a56b471d5cb884
Author: Pavel Tikhomirov <ptikhomirov@virtuozzo.com>
Date:   Fri Dec 22 11:22:26 2017 +0300

    mount: make open_mountpoint handle overmouts properly```

adrianreber:
I am using runc's test suite to trigger this error:

bats -t tests/integration/checkpoint.bats

This is, as always with my tests, on the CentOS/RHEL kernel. So there might also be something related to that kernel.

avagin:
I think we need something like this https://github.com/avagin/criu/commit/b264318361d11cd90b0ee95c987c69128a852fa3

Snorch:
>  I think we need something like this avagin/criu@b264318
I don't think so.

First, change in open_mountpoint actually does not change anything AFAICS, as if "!mnt_is_overmounted(pm)" is true we just get clean fd or go to err and "if (fd < 0) {" is never entered, if "!mnt_is_overmounted(pm)" is false then "if (fd < 0) {" check is always entered, we had the same with just "else".

Second, ns_open_mountpoint should return clean mount, nothing protects us from having mount pm with children in open_mountpoint's "else" path (mount can both have children an be overmounted).

Not yet sure how to fix it...

Snorch:
[...]

PR #1033, opened 2020-04-20
get_clean_mount: demote an error to a warning

kolyshkin:
The error here is not fatal since it is handled in ns_open_mountpoint(),
(at least since #520 is fixed), so demote this error to a warning
to avoid confusing a user.

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -1104,7 +1104,7 @@ static char *get_clean_mnt(struct mount_info *mi, char *mnt_path_tmp, char *mnt_
 	if (mnt_path == NULL && errno == ENOENT)
 		mnt_path = mkdtemp(mnt_path_root);
 	if (mnt_path == NULL) {
-		pr_perror("Can't create a temporary directory");
+		pr_warn("Can't create a temporary directory: %s\n", strerror(errno));
 		return NULL;
 	}
 


kolyshkin:
Would be nice to backport it to 3.14 as well

kolyshkin:
Inspired by runc intergration tests that grep for Error in criu log files.

Snorch:
@kolyshkin Can you please also list the exact error you got in you tests?

kolyshkin:
> Can you please also list the exact error you got in you tests?

Here you are:
```
# runc --criu /usr/local/sbin/criu checkpoint --work-path ./work-dir test_busybox (status=0):
# 
# (00.013986) mnt: 	497: 2f:/ @ ./dev/shm
# (00.016411) mnt: 	496: 2e:/ @ ./dev/pts
# (00.016420) mnt: 	495: 2b:/ @ ./dev
# (00.016422) mnt: Mount is not fully visible ./dev
# (00.016442) mnt: 	mount has children ./dev
# (00.016455) Error (criu/mount.c:1107): mnt: Can't create a temporary directory: Read-only file system
```

The cause is container's root is read-only (`readonly: true` in container's config.json in runc terms).

The integration tests code looks like this (a small example, slightly simplified):
```bash
    set -e 
...
    # checkpoint the running container
    runc --criu "$CRIU" checkpoint --work-path ./work-dir test_busybox
    cat ./work-dir/dump.log | grep -B 5 Error || true
    [ "$status" -eq 0 ]
...
```

The idea is to print any errors, if found in the log, before we check the exit code (and exit if it failed because of `set -e`).

Now,
1 we have to add `|| true` because there is an "Error" in the log.
2 we still see this error and it is confusing.

I want to demote it to a warning to fix these things.

criu/mount.c
```diff
@@ -1109,7 +1109,7 @@ static char *get_clean_mnt(struct mount_info *mi, char *mnt_path_tmp, char *mnt_
 	}
 
 	if (mount(mi->mountpoint, mnt_path, NULL, MS_BIND, NULL)) {
-		pr_perror("Can't bind-mount %d:%s to %s",
+		pr_warn("Can't bind-mount %d:%s to %s: %m",
```
avagin, line 1112: pls replace %m on ": %s", ..., strerror(errno))". It is safer. %m will print errno of a last syscall, but pr_warn can call something else before Xprintf.

kolyshkin: So, what you see is v 3.

* initially, I did it by implementing `pr_pwarn`. When I decided it's not worth it for one user only, and no one else will use it.
* v2 was doing `strerror(errno)` as you are suggesting.
* v3 is what you see, I did git grep '%m[^s]' and decided it is OK to use `%m`.

Anyway, let me rollback to v2.

kolyshkin: Also, `\n` is missing here!

kolyshkin: Also, I am not fixing the error I am seeing :facepalm:

kolyshkin: updated

kolyshkin:
@avagin updated, PTAL

Snorch:
>Also, I am not fixing the error I am seeing facepalm

Now it looks good, sorry for late response.

kolyshkin:
This actually fixes https://github.com/checkpoint-restore/criu/issues/575

PR #1039, opened 2020-04-23
Also report clone3() errors correctly

adrianreber:
Without clone3() CRIU was able to detect a process with a wrong PID only in the already created child process. With clone3() this error can happen before the process is created.

The clone3() function will now return -errno if something does not work correctly and in the case of EEXIST this error will also be correctly forwarded to an RPC client.

This was detected by running test/others/libcriu on a clone3() system.

### DIFF
--- a/criu/clone-noasan.c
+++ b/criu/clone-noasan.c
@@ -70,6 +70,7 @@ int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags,
 	if (!(flags & CLONE_PARENT)) {
 		if (exit_signal != SIGCHLD) {
 			pr_err("Exit signal not SIGCHLD\n");
+			errno = EINVAL;
 			return -1;
 		}
 		c_args.exit_signal = exit_signal;
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -1439,6 +1439,8 @@ static inline int fork_with_pid(struct pstree_item *item)
 
 	if (ret < 0) {
 		pr_perror("Can't fork for %d", pid);
+		if (errno == EEXIST)
+			set_cr_errno(EEXIST);
 		goto err_unlock;
 	}
 


criu/cr-restore.c
```diff
@@ -1438,7 +1438,9 @@ static inline int fork_with_pid(struct pstree_item *item)
 	}
 
 	if (ret < 0) {
-		pr_perror("Can't fork for %d", pid);
+		pr_perror("Can't fork for %d (errno %d)", pid, ret);
```
Snorch, line 1441: Maybe "(errno %d)" is excess if you call pr_perror anyway? Except this looks good.

adrianreber: I can remove that line. I just added it as pr_perror reports it as a string and the test case printed out that it expected 17 and comparing that to the errno string ('file exist') required an extra lookup for me.

Snorch: I'm not strictly against it but it attracts my eye as something strange. Let's leave it up to you and @avagin.

We already have 3 more places in code:

```
[snorch@snorch criu]$ git blame criu/cr-dedup.c | grep "pr_perror.*errno"
d866bd196b cr-dedup.c      (Tikhomirov Pavel 2013-12-17 19:27:08 +0400  22) 		pr_perror("Can't enter previous snapshot folder, error=%d", errno);
d866bd196b cr-dedup.c      (Tikhomirov Pavel 2013-12-17 19:27:08 +0400  32) 				pr_perror("Failed readdir, error=%d", errno);
[snorch@snorch criu]$ git blame criu/mem.c | grep "pr_perror.*errno"
ba78d15557 criu/mem.c (Kir Kolyshkin         2017-03-23 15:02:11 -0700   67) 			pr_perror("Can't reset %d's dirty memory tracker (%d)", pid, errno);
```

I'm guilty too =)

kolyshkin: We can add `errno %d` to `pr_error`, and drop these five cases.

kolyshkin: Or fix the tests. Maybe it's better to fix the tests.

adrianreber:
rebased;

Not sure what the status of the review is? Does it need rework or not?

criu/clone-noasan.c
```diff
@@ -80,5 +80,7 @@ int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags,
 	pid = syscall(__NR_clone3, &c_args, sizeof(c_args));
 	if (pid == 0)
 		exit(fn(arg));
+	if (pid < 0)
+		return -errno;
```
avagin, line 84: This function can return -1, but this isn't -EPERM. Do we really need to return -errno here? Maybe it is better to write a comment that this functions returns valid errno in cases of errors and initialize it properly where it returns -1.

adrianreber: How do I initialize it properly?

avagin: if (exit_signal != SIGCHLD) {
		pr_err("Exit signal not SIGCHLD\n");
                errno = EINVAL;
		return -1;
}

criu/cr-restore.c
adrianreber: I removed the line, so we do not have to solve this discussion here.

criu/clone-noasan.c
adrianreber: I changed the code like you suggested.

PR #1041, opened 2020-04-23
lib/c: add criu_get_version()

adrianreber:
Although the CRIU version is exported in macros in version.h it only contains the CRIU version of libcriu during build time.

As it is possible that CRIU is upgraded since the last time somethingwas built against libcriu, this adds functions to query the actual CRIU binary about its version.

### DIFF
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -1657,3 +1657,68 @@ int criu_restore_child(void)
 {
 	return criu_local_restore_child(global_opts);
 }
+
+int criu_local_get_version(criu_opts *opts)
+{
+	int ret = -1;
+	CriuReq req	= CRIU_REQ__INIT;
+	CriuResp *resp	= NULL;
+
+	saved_errno = 0;
+
+	req.type	= CRIU_REQ_TYPE__VERSION;
+	req.opts	= opts->rpc;
+
+	ret = send_req_and_recv_resp(opts, &req, &resp);
+	if (ret)
+		goto exit;
+
+	if (resp->success) {
+		ret = resp->version->major_number * 10000;
+		ret += resp->version->minor_number * 100;
+		if (resp->version->has_sublevel)
+			ret += resp->version->sublevel;
+		if (resp->version->gitid) {
+			/* Taken from runc: a git release -> minor + 1 */
+			ret -= (ret % 100);
+			ret += 100;
+		}
+	} else {
+		ret = -EBADE;
+	}
+
+exit:
+	if (resp)
+		criu_resp__free_unpacked(resp, NULL);
+
+	swrk_wait(opts);
+
+	errno = saved_errno;
+
+	return ret;
+}
+
+int criu_get_version(void)
+{
+	return criu_local_get_version(global_opts);
+}
+
+int criu_local_check_version(criu_opts *opts, int minimum)
+{
+	int version;
+
+	version = criu_local_get_version(opts);
+
+	if (version < 0)
+		return version;
+
+	if (minimum <= version)
+		return 1;
+
+	return 0;
+}
+
+int criu_check_version(int minimum)
+{
+	return criu_local_check_version(global_opts, minimum);
+}
--- a/lib/c/criu.h
+++ b/lib/c/criu.h
@@ -158,6 +158,35 @@ typedef void *criu_predump_info;
 int criu_dump_iters(int (*more)(criu_predump_info pi));
 
 /*
+ * Get the version of the actual binary used for RPC.
+ *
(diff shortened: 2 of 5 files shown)


lib/c/criu.h
```diff
@@ -157,6 +157,35 @@ int criu_restore_child(void);
...
+ *
+ * As this library is just forwarding all tasks to an
+ * independent (of this library) CRIU binary, the actual
+ * version of the CRIU binary can be different then the
+ * hardcoded values in the libary (version.h).
+ * To be able to easily check the version of the CRIU binary
+ * the function criu_get_version() returns the version
+ * in the following format:
+ *
+ * (major * 10000) + (minor * 100) + sublevel
```
Snorch, line 171: Maybe it is better to use bit shift based version format? If for instance 128 is enough for version number we can have:
```
#define V_MAJOR_SHIFT 14
#define V_MINOR_SHIFT 7
(major<<V_MAJOR_SHIFT) + (minor<<V_MINOR_SHIFT) + sublevel
```

adrianreber: If I understand it correctly that would be pretty difficult to read (at least for me).

Instead of getting 31400 for CRIU 3.14 I would get 50944 ((3 << 14)+(14 << 7)). Especially for the check_version() function it would not be very obvious in the code for which version I want to check. Not convinced that this would make sense.

Snorch: May be I chouse bad shift, 8 looks nice:

>>> hex(((3 << 16)+(14 << 8)))
'0x30e00'

avagin:
Could you rebase this PR?

lib/c/criu.h
adrianreber: I would prefer to leave it the way it is to easily read number when in decimal.

runc uses the decimal approach already for CRIU version detection and the go-criu code (which is a copy of the runc code) also

lib/c/criu.c
```diff
@@ -1657,3 +1657,71 @@ int criu_restore_child(void)
...
+	return criu_local_get_version(global_opts);
+}
+
+int criu_local_check_version(criu_opts *opts, int minimum)
+{
+	int version;
+
+	version = criu_local_get_version(opts);
+
+	printf("version %d\n", version);
```
avagin, line 1712: I don't think that this is a good idea to print anything from the library.

lib/c/criu.c
```diff
@@ -1657,3 +1657,71 @@ int criu_restore_child(void)
...
 	return criu_local_restore_child(global_opts);
 }
+
+int criu_local_get_version(criu_opts *opts)
+{
+	int ret = -1;
+	CriuReq req	= CRIU_REQ__INIT;
+	CriuResp *resp	= NULL;
+
+	saved_errno = 0;
```
avagin, line 1667: saved_errno = errno ? Why do we need this?

lib/c/criu.c
adrianreber: No, this is wrong. Some debug printfs. Will remove it.

lib/c/criu.c
adrianreber: recv_resp() sets saved_errno to the errno in the case of a failure during recv(). This way we can forward the errno of the RPC communication to the library user.

lib/c/criu.c
```diff
@@ -1657,3 +1657,68 @@ int criu_restore_child(void)
...
+}
+
+int criu_local_check_version(criu_opts *opts, int minimum)
+{
+	int version;
+
+	version = criu_local_get_version(opts);
+
+	if (version < 0)
+		return version;
```
rst0git, line 1713: > Returns < 0 if there was an error.

`return -version;`?

lib/c/criu.c
```diff
@@ -1657,3 +1657,68 @@ int criu_restore_child(void)
...
+		ret += resp->version->minor_number * 100;
+		if (resp->version->has_sublevel)
+			ret += resp->version->sublevel;
+		if (resp->version->gitid) {
+			/* Taken from runc: a git release -> minor + 1 */
+			ret -= (ret % 100);
+			ret += 100;
+		}
+	} else {
+		ret = -EBADE;
```
avagin, line 1687: I don't like the idea to use both return value and errno for error codes. We have to choose only one of them.

adrianreber: This code structure is just copied from all the other functions in the library. I have no problem changing it, just wanted to point out, that almost all functions in the library are structured like this. This behaviour is documented here https://github.com/checkpoint-restore/criu/blob/criu-dev/lib/c/criu.h#L120

Issue #777, opened 2019-09-07
checkpoint/restore support for bpf-map fd


rst0git:
We need to add support for checkpoint/restore of bpf-map file descriptors.

This is a simple example of how to open such file descriptor:

```C
#include <stdio.h>
#include <unistd.h>
#include <linux/bpf.h>
#include <sys/syscall.h>

static inline int sys_bpf(enum bpf_cmd cmd, union bpf_attr *attr,
			  unsigned int size)
{
	return syscall(__NR_bpf, cmd, attr, size);
}

int bpf_create_map(enum bpf_map_type map_type,
		   unsigned int key_size,
		   unsigned int value_size,
		   unsigned int max_entries)
{
	union bpf_attr attr = {
	  .map_type    = map_type,
	  .key_size    = key_size,
	  .value_size  = value_size,
	  .max_entries = max_entries
	};

	return sys_bpf(BPF_MAP_CREATE, &attr, sizeof(attr));
}

int main()
{
	int fd = bpf_create_map(BPF_MAP_TYPE_PROG_ARRAY, sizeof(int), sizeof(int), 4);
	printf("bpf-map fd: %d\n", fd);
	getchar();
	return 0;
}
```

When trying to checkpoint this process CRIU fails with:
```
Error (criu/files-ext.c:95): Can't dump file 3 of that type [600] (anon anon_inode:bpf-map)
...
Error (criu/cr-dump.c:1754): Dumping FAILED.
```

PR #1044, opened 2020-04-25
Dumping BPF map files (Issue #777)

abhishekvijeev:
Hi,

I have implemented functionality to dump information about BPF map files and also created a template ZDTM static test file for BPF maps. The test currently fails because restore functionality has not yet been implemented. I am currently working on this.

My first goal is to get checkpoint and restore fully working for simple BPF map files, after which I will enhance CRIU to deal with other kinds of BPF files. Kindly let me know whether you would prefer me to prioritize something else.

I await your kind feedback.

Thank you.


Snorch:
Hi @abhishekvijeev!

Let's start with small nits on how you do a pr to criu:

1) Please add a Siged-off-by line in each commit.
2) Each commit should have commit message formated as:
```
component1[/component2/...]: short commit summary

longer commit description which describes what, why and how exectly you do

Signed-off-by: Name Surname <e-mail>
```
3) We want to have pr's containing commits with clean code only, you need to squash all fixup commit to their base and force push the pr. (I mean "Code formatting" commit is a bad practice)

4) Another thing all related travis tests should pass, it looks like you have some of them failing:

```
===================== Run zdtm/static/bpf_create_map in h ======================
Start test
./bpf_create_map --pidfile=bpf_create_map.pid --outfile=bpf_create_map.out
Run criu dump
=[log]=> dump/zdtm/static/bpf_create_map/47/1/dump.log
(00.003791) 47 fdinfo 2: pos:              0xe flags:           102001/0
(00.003812) 47 fdinfo 3: pos:                0 flags:           100000/0
(00.003822) Dumping path for 3 fd via self 18 [/home/travis/build/checkpoint-restore/criu/test/zdtm/static]
(00.003841) 47 fdinfo 4: pos:                0 flags:                2/0x1
(00.003868) Error (criu/proc_parse.c:1749): No data left in proc file while parsing bpfmap
(00.003879) Error (criu/proc_parse.c:2114): parse_fdinfo_pid_s: error parsing [map_type:	3] for 19: Success
(00.003887) ----------------------------------------
(00.003898) Error (criu/cr-dump.c:1349): Dump files (pid: 47) failed with -1
(00.004127) 47 was stopped
(00.004215) Unlock network
(00.004220) Unfreezing tasks into 1
(00.004223) 	Unseizing 47 into 1
(00.004235) Error (criu/cr-dump.c:1775): Dumping FAILED.
############## Test zdtm/static/bpf_create_map FAIL at CRIU dump ###############
Send the 9 signal to  47
Wait for zdtm/static/bpf_create_map(47) to die for 0.100000
```

abhishekvijeev:
Hi @Snorch ,

Thank you very much for the feedback.

I have made the necessary changes to the commits (will keep this in mind henceforth). Kindly let me know if I can explain any of the commit messages more clearly.

Also, the reason that majority of tests were failing was because I had implemented incomplete boiler-plate code for testing c/r of BPF maps. I have removed the test from the PR for now, and will add it once it is complete (I have described below my thoughts on creating tests for BPF maps). 
Therefore, these tests pass successfully now. However, 3 of the tests (#4568.13, #4568.14 and 4568.20) for the ARM platform fail with the following error:

criu/proc_parse.c: In function 'parse_bpfmap':
criu/proc_parse.c:1724:32: error: format '%lu' expects argument of type 'long unsigned int *', but argument 3 has type 'uint64_t *' {aka 'long long unsigned int *'} [-Werror=format=]
  if (sscanf(str, "map_flags: %lu", &bpf->map_flags) != 1)
                              ~~^   ~~~~~~~~~~~~~~~
                              %llu
criu/proc_parse.c:1730:30: error: format '%lu' expects argument of type 'long unsigned int *', but argument 3 has type 'uint64_t *' {aka 'long long unsigned int *'} [-Werror=format=]
  if (sscanf(str, "memlock: %lu", &bpf->memlock) != 1)
                            ~~^   ~~~~~~~~~~~~~
                            %llu


I tried changing the format specifier to %llu, but the code didn't compile on my PC (x86_64). Does this test fail because compilers view the data type 'uint64_t' differently for ARM64 and x86_64? If this is the case, how should I resolve this issue?

Currently, this PR accomplishes 2 tasks:
a) Checkpoint - read information about BPF maps from the proc filesystem and dump this into image files
b) Restore - read information back from the image files and create new BPF map files with the same parameters preserved during checkpointing

However, a significant part of the problem remains unsolved. BPF map files typically contain data in the form of key value pairs. CRIU must therefore checkpoint and restore this data as well. According to my current understanding, the data contained within BPF maps is not made available via the proc
filesystem. Therefore, the way I'm thinking of accomplishing this is by iterating through all the key-value pairs stored in a map when it is about to be checkpointed, and saving them in a new image (similar to what is currently done for pipes' data). Kindly let me know your thoughts on this.

Also, BPF maps can be persisted on /sys/fs/bpf/ using the BPF_PIN_FD command argument. In this case, during restoration, instead of creating a new map with the same parameters (the way this PR currently works), CRIU must be able to open the map on the sys filesystem. This is another problem I will solve.

Once I have accomplished this, I will create 2 ZDTM tests that accomplish the following:

a)  - Checkpoint a process that has a BPF map with data
     - Restore the BPF map and its data
     - Check whether the restored map's parameters and data are the same as before checkpointing

b)  - Checkpoint a process that has a BPF map persisted on sysfs
     - Restore the BPF map and its data
     - Check whether the map is the same as the one persisted on sysfs

Thank you,
Abhishek

Snorch:
>how should I resolve this issue?
PRIu64 should help to print unit64_t https://en.cppreference.com/w/c/types/integer

>iterating through all the key-value pairs stored in a map when it is about to be checkpointed, and saving them in a new image
You may also consider protobuf repeated key-value pair entry.

Thanks,
Pavel.

abhishekvijeev:
Thanks!

avagin:
Could you rebase this PR?

abhishekvijeev:
Sure @avagin , it's done.

abhishekvijeev:
I see that the following RPC test is failing:
https://travis-ci.org/github/checkpoint-restore/criu/jobs/684654477#L17709

I tried reproducing the error on my PC by executing the following command:
"make -C test/others/rpc/ run"

'test-c' fails with the following output (obtained from test/others/rpc/build/output_c):

"build/imgs_c
Fail
build/imgs_c
Fail
Cant connect to socket: Connection refused
build/imgs_c
build/imgs_c
Fail
build/imgs_c
Fail
build/imgs_c
Fail
build/imgs_c
Fail
build/imgs_c
Fail
build/imgs_c
Success
build/imgs_c
Success
Restored"

The socket connection was failing. I therefore re-ran the test as the root user, and it passed. Am I correct in assuming that instructing Travis to execute the test as root user is the correct solution to this problem? If so, how can I ensure that Travis passes it?

Thank you.

avagin:
Could you push the test too? It will help to review these changes.

abhishekvijeev:
The template test I had created earlier for BPF maps?

abhishekvijeev:
Hi @Snorch @avagin ,

I don't think it's possible to checkpoint a BPF map file's data, as Linux's BPF system call API doesn't provide a means to duplicate BPF map data (like 'tee' for pipes). I have explained the problem in this email:
https://lists.openvz.org/pipermail/criu/2020-May/044995.html

It's meaningless to continue work on BPF map files (https://github.com/checkpoint-restore/criu/issues/777) unless Linux's BPF API evolves to support dumping map data. I mean, I don't see the point in only reading info from the proc filesystem and restoring an identical map without its data. Do let me know if you think otherwise (i.e if you think there is value in only dumping and restoring a map without its data). Else, I shall close this PR.

Thank you.

abhishekvijeev:
Hi @Snorch  @avagin ,

Could you kindly let me know whether I am correct in assuming that I've hit a dead end and if so, can I close this PR? 

Thank you.

avagin:
@abhishekvijeev This is not a dead end, you hit a new challenge of extending linux API-s.  While we were developing CRIU, we met similar problems a few time. Now, you need to discuss a new API for dumping BPF files with the kernel community and implement it.

abhishekvijeev:
Thank you for the response @avagin 
Sure, I'll get to work on this.

abhishekvijeev:
Hi @avagin @Snorch ,

For a ZDTM test that checks the consistency of BPF map data after a restore, I require access to a hash table (where keys and values are integers). Is there a library you would recommend I use, or should I implement a simple hash table with chaining myself? 

Thank you.

Snorch:
You may consider using hlist hash table, see include/common/list.h .

abhishekvijeev:
Thank you.

abhishekvijeev:
Including list.h unfortunately gives me the following error due multiple definitions of __stringify_1:

In file included from ../../../include/common/list.h:9,
                 from bpf_hash_data.c:6:
../../../include/common/compiler.h:17: error: "__stringify_1" redefined [-Werror]
   17 | #define __stringify_1(x...) #x
      | 
In file included from bpf_hash_data.c:5:
../lib/zdtmtst.h:113: note: this is the location of the previous definition
  113 | #define __stringify_1(x)        #x
      | 
In file included from ../../../include/common/list.h:9,
                 from bpf_hash_data.c:6:
../../../include/common/compiler.h:18: error: "__stringify" redefined [-Werror]
   18 | #define __stringify(x...) __stringify_1(x)
      | 
In file included from bpf_hash_data.c:5:
../lib/zdtmtst.h:114: note: this is the location of the previous definition
  114 | #define __stringify(x)          __stringify_1(x)

Any suggestions?

abhishekvijeev:
Hi @Snorch @avagin ,

This PR is ready for review. It currently enables CRIU to c/r the data and meta-data of BPF_MAP_TYPE_HASH and BPF_MAP_TYPE_ARRAY maps. I have implemented ZDTM tests for these as well.

I'd like to bring a few things to your attention:

1) Travis CI fails due to 2 reasons:
        a) The code depends on a new library - libbpf, which must be installed
        b) The code uses features that are only available in Linux kernel versions 5.6 or newer
Therefore, please help me solve this issue.

2) In criu/include/magic.h, I have currently defined two magic numbers based on 2 randomly selected Russian cities - Alapayevsk and Arkhangelsk. Kindly let me know your preference of cities - I will change them to reflect your preferences.

Thank you.

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+	unsigned int count;
+	DECLARE_LIBBPF_OPTS(bpf_map_batch_opts, opts,
+		.elem_flags = 0,
+		.flags = 0,
+	);
+
+	for (map_data = bpf_hash_table[map_id & BPFMAP_DATA_HASH_MASK]; map_data != NULL; 
+		map_data = map_data->next) {
+
+		if(map_data->bde->map_id == map_id)
```
mihalicyn, line 78: formatting `if (`

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+			break;
+	}
+
+	if(!map_data || map_data->bde->count == 0) {
```
mihalicyn, line 82: formatting `if (`

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+		return 0;
+	}
+
+	bde = map_data->bde;
+	count = bde->count;
+
+	keys = mmap(NULL, bde->keys_bytes, PROT_READ | PROT_WRITE,
+				MAP_SHARED | MAP_ANONYMOUS, 0, 0); 
+	if(keys == MAP_FAILED) {
+		pr_perror("Can't map memory for BPF map keys\n");
```
mihalicyn, line 93: when `pr_perror` is used no need to add `\n` at the end

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+		goto err;
+	}
+	memcpy(keys, map_data->data, bde->keys_bytes);
+
+	values = mmap(NULL, bde->values_bytes, PROT_READ | PROT_WRITE,
+				MAP_SHARED | MAP_ANONYMOUS, 0, 0); 
+	if(values == MAP_FAILED) {
```
mihalicyn, line 100: formatting

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+		pr_perror("Can't map memory for BPF map values\n");
```
mihalicyn, line 101: no need to add `\n`

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+	munmap(values, value_size * max_entries);
+	munmap(out_batch, key_size);
+	return -1;
+}
+
+static int dump_one_bpfmap(int lfd, u32 id, const struct fd_parms *p)
+{
+	BpfmapFileEntry bpf = BPFMAP_FILE_ENTRY__INIT;
+	FileEntry fe = FILE_ENTRY__INIT;
+    int ret;
```
mihalicyn, line 233: spaces instead of tabs

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+
+	if (parse_fdinfo(lfd, FD_TYPES__BPFMAP, &bpf))
+		return -1;
+
+    switch(bpf.map_type) {
```
mihalicyn, line 238: `switch (`

criu/files.c
```diff
@@ -539,6 +540,8 @@ static int dump_one_file(struct pid *pid, int fd, int lfd, struct fd_opts *opts,
 			ops = &signalfd_dump_ops;
 		else if (is_timerfd_link(link))
 			ops = &timerfd_dump_ops;
+		else if(is_bpfmap_link(link))
```
mihalicyn, line 543: `else if (is_bpf...`

test/zdtm/static/bpf_array.c
```diff
@@ -0,0 +1,155 @@
...
+		.value_size = sizeof(int),
+		.max_entries = max_entries,
+		.map_flags = BPF_F_NUMA_NODE,
+	};
+	DECLARE_LIBBPF_OPTS(bpf_map_batch_opts, opts,
+		.elem_flags = 0,
+		.flags = 0,
+	);
+
+    keys = mmap(NULL, max_entries * sizeof(int),
```
mihalicyn, line 75: spaces instead of tabs

test/zdtm/static/bpf_array.c
```diff
@@ -0,0 +1,155 @@
...
+
+	test_init(argc, argv);
+
+	map_fd = bpf_create_map_xattr(&xattr);
+	if(map_fd == -1) {
+		pr_perror("Can't create BPF map\n");
+		goto err;
+	}
+
+    if(map_batch_update(map_fd, max_entries, keys, values))
```
mihalicyn, line 95: spaces

mihalicyn:
Great job!

criu/bpfmap.c
```diff
@@ -0,0 +1,323 @@
...
+		pr_perror("Can't map memory for BPF map out_batch\n");
+		goto err;
+	}
+
+	ret = bpf_map_lookup_batch(lfd, in_batch, out_batch, keys, values, &count, &opts);
+	if(ret && errno != ENOENT) {
+		pr_perror("Can't perform a batch lookup on BPF map\n");
+		goto err;
+	}
+
```
mihalicyn, line 197: shouldn't we have a check about `count == 0`?

criu/Makefile.packages
```diff
@@ -7,6 +7,7 @@ REQ-RPM-PKG-NAMES	+= protobuf-python
 REQ-RPM-PKG-NAMES	+= libnl3-devel
 REQ-RPM-PKG-NAMES	+= libcap-devel
 REQ-RPM-PKG-NAMES	+= $(PYTHON)-future
+REQ-RPM-PKG-NAMES	+= libbpf-devel
```
rst0git, line 10: Could you add this dependency in the Dockerfile templates (in `scripts/build/`) to enable Travis to run the tests?

rst0git:
> The code uses features that are only available in Linux kernel versions 5.6 or newer

The Fedora 31 kernel (`scripts/travis/vagrant.sh`) should be able to run the BPF test cases.

abhishekvijeev:
Thank you very much @mihalicyn and @rst0git 

I will make the necessary changes.

criu/Makefile.packages
abhishekvijeev: Currently, the helper functions and structures for BPF batch operations (used in this PR) are not available in 'stable' releases of libbpf-dev. For instance, Debian has the latest version of libbpf-dev (v1:0.0.9-1) only in its 'testing' and 'unstable' repositories. I see from the Dockerfile templates that only 'stable' repositories are added to the package sources list file. Therefore, the code will not compile. 

Would you like me to update the Dockerfile templates anyway?

criu/bpfmap.c
abhishekvijeev: If count == 0 after the lookup, I store it as 0 in the protobuf image (without any data).
During restore, I check to see whether the count is 0 and if this is the case, return without trying to restore data:
https://github.com/checkpoint-restore/criu/pull/1044/files#diff-72d5d16cff7b2f3309c3fd148760d60aR82-R85

Would you like me to change the logic such that it checks for count == 0 during the dump phase instead of restore? If so, I would have to return from the function only after freeing the memory allocated above. I think the cleanest way to do this would be by adding a goto label 'pass' here:
https://github.com/checkpoint-restore/criu/pull/1044/files#diff-72d5d16cff7b2f3309c3fd148760d60aR216
The code would then become:

`ret = bpf_map_lookup_batch(lfd, in_batch, out_batch, keys, values, &count, &opts);

	if (ret && errno != ENOENT) {
		pr_perror("Can't perform a batch lookup on BPF map");
		goto err;
	}
        if (count == 0)
                goto pass;`

Please let me know which approach you think would be better.

Thank you.

abhishekvijeev:
All formatting inconsistencies have been fixed.

avagin:
The travis jobs fail, you need to fix them:
https://travis-ci.org/github/checkpoint-restore/criu/builds/713274776?utm_source=github_status&utm_medium=notification

abhishekvijeev:
Hi,

I've tried fixing Travis tests by installing ```libbpf-dev``` across the different linux distros and ISAs being tested. The commit that adds these changes is [here](https://github.com/checkpoint-restore/criu/pull/1044/commits/18828e369200a2e53e08d0797ece6110aa500874).

Currently, most of the tests that fail do so because the kernel is not v5.6+ and therefore does not support BPF batch operations. Is there any way by which we can get the test machines to use kernel v5.6+?

The tests that fail for reasons other than an old kernel are as follows:

* TR_ARCH=x86_64

    ```
    In file included from criu/bpfmap.c:3:0:
        /usr/include/bpf/bpf.h:76:23: error: field 'expected_attach_type' has incomplete type
        enum bpf_attach_type expected_attach_type;
    ```

    [This thread](https://github.com/libbpf/libbpf/issues/53#issuecomment-580962821) says that the cause of this problem is old linux headers. Will this be fixed if the kernel version is upgraded to 5.6+?

* TR_ARCH=centos
        libbpf requires libc >= 2.26, which I tried installing from CentOS repos. However, as you can see [here](https://travis-ci.org/github/checkpoint-restore/criu/jobs/717312166#L1241), manually installing a newer version of libc breaks things. Since libc >= 2.26 is available in CentOS 8, would it be possible to run Travis tests using CentOS 8?

* TR_ARCH=fedora-asan
        The version of libbpf installed doesn't support BPF batch operations. This should be resolved once the package is upgraded to its latest version.

Any suggestions that could help resolve these problems?

Thank you.

adrianreber:
> Hi,
> 
> I've tried fixing Travis tests by installing ```libbpf-dev``` across the different linux distros and ISAs being tested. The commit that adds these changes is [here](https://github.com/checkpoint-restore/criu/pull/1044/commits/18828e369200a2e53e08d0797ece6110aa500874).
> 
> Currently, most of the tests that fail do so because the kernel is not v5.6+ and therefore does not support BPF batch operations. Is there any way by which we can get the test machines to use kernel v5.6+?

Have a look at the vagrant based test case. That uses an up2date kernel based on Fedora 32. Also see https://github.com/checkpoint-restore/criu/pull/1155 for another vagrant based test with a very new kernel.

> The tests that fail for reasons other than an old kernel are as follows:
> 
> * TR_ARCH=x86_64
> 
>     ```
>     In file included from criu/bpfmap.c:3:0:
>         /usr/include/bpf/bpf.h:76:23: error: field 'expected_attach_type' has incomplete type
>         enum bpf_attach_type expected_attach_type;
>     ```
> 
>     [This thread](https://github.com/libbpf/libbpf/issues/53#issuecomment-580962821) says that the cause of this problem is old linux headers. Will this be fixed if the kernel version is upgraded to 5.6+?

Travis kernels will probably not be updated for something like two years.

> * TR_ARCH=centos
>         libbpf requires libc >= 2.26, which I tried installing from CentOS repos. However, as you can see [here](https://travis-ci.org/github/checkpoint-restore/criu/jobs/717312166#L1241), manually installing a newer version of libc breaks things. Since libc >= 2.26 is available in CentOS 8, would it be possible to run Travis tests using CentOS 8?

Introduce a CentOS 8 based test. We should test that. It will probably also run in a container like CentOS 7. Do not update any packages outside of the official package set. That will break too many things.

> * TR_ARCH=fedora-asan
>         The version of libbpf installed doesn't support BPF batch operations. This should be resolved once the package is upgraded to its latest version.
 
Which version of Fedora is that based on? Maybe just update to the latest Fedora version for that test case.

> Any suggestions that could help resolve these problems?
> 
> Thank you.
> 
>

avagin:
@abhishekvijeev could you look at https://github.com/avagin/criu/commits/0a6644caeb08fd39aeee828cedce4c52b1014b48

* I have updated base images for build containers.
* CRIU compiled without the support of bpf files if libbpf isn't installed.
* The bpf tests are executed in the TR_ARCH=vagrant-fedora-no-vdso job.

abhishekvijeev:
Thanks so much Adrian, Andrei!

So basically, I just need to worry about executing the ZDTM tests (bpf_array and bpf_hash) on the fedora vagrant machine. Since the other machines use an older kernel, we don't install libbpf or run any of the BPF code (by setting a config flag depending on whether libbpf is installed or not). Is this correct? 

I have a few more questions:

* CONFIG_HAS_LIBBPF is [set](https://github.com/avagin/criu/commit/0a6644caeb08fd39aeee828cedce4c52b1014b48#diff-6e8f6ae888b05f88a8c1c4ca525515b1R21) only if libbpf is installed. And if libbpf is installed, we [compile bpfmap.c](https://github.com/avagin/criu/commit/0a6644caeb08fd39aeee828cedce4c52b1014b48#diff-45e0b3cae0c8e1125f265fd694dd1c44R92). But, why is bpfmap.c  [not compiled](https://travis-ci.org/github/avagin/criu/jobs/717841154) for TR_ARCH=x86_64, even though we have [installed libbpf](https://github.com/avagin/criu/commit/0a6644caeb08fd39aeee828cedce4c52b1014b48#diff-69e92cca26cce77c47ba1936fcf5d303R27)? Am I missing something?

* Is it necessary to update the [fedora-asan](https://github.com/avagin/criu/commit/f788b5eb84bb0a7ce2aae17caf600c35ccce38a3#diff-dc705e73de3c3991927a87a98dd7d872R1) and [x86-64](https://github.com/avagin/criu/commit/0a6644caeb08fd39aeee828cedce4c52b1014b48#diff-009fdba7365eaa1d5c4feb49cb1cd164R1) base images (since we won't be using libbpf)?

* Shall I go ahead with creating a new CentOS 8 based test?

adrianreber:
> * Is it necessary to update the [fedora-asan](https://github.com/avagin/criu/commit/f788b5eb84bb0a7ce2aae17caf600c35ccce38a3#diff-dc705e73de3c3991927a87a98dd7d872R1) and [x86-64](https://github.com/avagin/criu/commit/0a6644caeb08fd39aeee828cedce4c52b1014b48#diff-009fdba7365eaa1d5c4feb49cb1cd164R1) base images (since we won't be using libbpf)?

From my point of view it is not necessary if we cannot run the tests there.

> * Shall I go ahead with creating a new CentOS 8 based test?

You are of course welcome to do that, but if you do not need it, probably not. There should be some support for bpf on rhel/centos 8, as far as I know.

avagin:
> CONFIG_HAS_LIBBPF is set only if libbpf is installed. And if libbpf is installed, we compile bpfmap.c. But, why is bpfmap.c not compiled for TR_ARCH=x86_64, even though we have installed libbpf? Am I missing something?

Step 7/12 : RUN pkg-config --print-errors libbpf
1001 ---> Running in 4d41b8fe2cf2
1002Package libelf was not found in the pkg-config search path.
1003Perhaps you should add the directory containing `libelf.pc'
1004to the PKG_CONFIG_PATH environment variable
1005Package 'libelf', required by 'libbpf', not found
1006The command '/bin/sh -c pkg-config --print-errors libbpf' returned a non-zero code: 1

avagin:
Here is a two-line fix https://github.com/avagin/criu/commit/863a9cbf743b900a782375c08dada820a7ee8dc8

abhishekvijeev:
Thank you very much.

I will fix this PR.

abhishekvijeev:
Hi,

Travis tests pass now. The Fedora rawhide tests fail, but I see that this has been occurring across [CRIU builds](https://travis-ci.org/github/checkpoint-restore/criu/builds) over that past couple of days - I therefore assume that this is a temporary problem independent of this PR.

I have not added a CentOS 8 test or upgraded the docker base images because there was no need to. I will be glad to do these if necessary.

Please let me know if I can improve this PR in any way before it's merged.

Thank you.

avagin:
Applied. Thanks a lot!

abhishekvijeev:
That's great!

Thanks so much to the community for helping me work towards getting this PR merged!

PR #1046, opened 2020-04-27
Fix error logging string in libsoccr tcp-test

jabedude:
I found two minor string errors while studying tcp-conn test to better understand the libsoccr API. This patch fixes them :)


avagin:
Applied, thanks

PR #1047, opened 2020-04-29
criu swrk: fix usage, allow common options

kolyshkin:
TL;DR: this makes possible to use `-v` with `criu swrk`, and removes showing
default usage which is useless in swrk mode.

1. Since criu swrk command is not described in usage, there is no sense
   in showing it. Instead, show a one-line hint about how to use it.

2. In case some global options (like `-v`) are used, `argv[1]` might not
   point to "swrk". Use `optind` to point to a correct non-option
   argument. Same for `fd` argument.

3. While at it, also error out in case we have extra arguments.

Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

### DIFF
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -87,12 +87,18 @@ int main(int argc, char *argv[], char *envp[])
 		return 1;
 	if (ret == 2)
 		goto usage;
+	if (optind >= argc) {
+		pr_err("command is required\n");
+		goto usage;
+	}
 
 	log_set_loglevel(opts.log_level);
 
-	if (!strcmp(argv[1], "swrk")) {
-		if (argc < 3)
-			goto usage;
+	if (optind < argc && !strcmp(argv[optind], "swrk")) {
+		if (argc != optind+2) {
+			fprintf(stderr, "Usage: criu swrk <fd>\n");
+			return 1;
+		}
 		/*
 		 * This is to start criu service worker from libcriu calls.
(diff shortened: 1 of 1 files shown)


rst0git:
> this makes possible to use -v with criu swrk

> Since criu swrk command is not described in usage, there is no sense in showing it. 

The usage also contains information about options like `-v`.
Would it make sense to show the description of such options for `criu swrk`?

kolyshkin:
> Would it make sense to show the description of such options for criu swrk?

I get it that `swrk` was deliberately excluded from `usage`. I just wanted to show something that makes sense if the command failed (`usage` did not make sense since swrk is not even there).

So, it definitely does not make sense to show usage as it is.

avagin:
Could you rebase this PR?

kolyshkin:
rebased

criu/crtools.c
```diff
@@ -90,17 +90,19 @@ int main(int argc, char *argv[], char *envp[])
 
 	log_set_loglevel(opts.log_level);
 
-	if (!strcmp(argv[1], "swrk")) {
-		if (argc < 3)
-			goto usage;
+	if (!strcmp(argv[optind], "swrk")) {
```
Snorch, line 93: What would happen if optind >= argc ?
```
[snorch@snorch criu]$ criu/criu -v4
Segmentation fault
```

kolyshkin: nice catch! fixed

PR #1051, opened 2020-05-02
Remote image: code simplification, fixes, and performance improvements

nviennot:
See the commit messages for details.

Aside from code simplification and bug fixes, the important contribution is the use of pipes instead of UNIX sockets for local communication, enabling highly efficient data transfers for proxies that use splice().

### DIFF
--- a/criu/img-remote.c
+++ b/criu/img-remote.c
@@ -1041,28 +1041,22 @@ static int pull_snapshot_ids(void)
 {
 	int n, sockfd;
 	SnapshotIdEntry *ls;
-	struct snapshot *s = NULL;
+	struct snapshot *s;
 
 	sockfd = read_remote_image_connection(NULL_SNAPSHOT_ID, PARENT_IMG);
 
-	/* The connection was successful but there is not file. */
-	if (sockfd < 0) {
-		if (errno != ENOENT) {
-			pr_err("Unable to open snapshot id read connection\n");
-			return -1;
-		}
-		return 0;
-	}
+	if (sockfd < 0)
+		return -1;
 
 	while (1) {
 		n = pb_read_obj(sockfd, (void **)&ls, PB_SNAPSHOT_ID);
 		if (!n) {
 			close(sockfd);
-			return n;
+			return 0;
 		} else if (n < 0) {
 			pr_err("Unable to read remote snapshot ids\n");
 			close(sockfd);
-			return n;
+			return -1;
 		}
 
 		s = new_snapshot(ls->snapshot_id);
@@ -1073,9 +1067,6 @@ static int pull_snapshot_ids(void)
 		add_snapshot(s);
 		pr_info("[read_snapshot ids] parent = %s\n", ls->snapshot_id);
 	}
-	free(ls);
-	close(sockfd);
-	return n;
 }
 
 int push_snapshot_id(void)
@@ -1121,8 +1112,10 @@ int get_curr_snapshot_id_idx(void)
 	struct snapshot *si;
 	int idx = 0;
 
-	if (list_empty(&snapshot_head))
-		pull_snapshot_ids();
+	if (list_empty(&snapshot_head)) {
+		if (pull_snapshot_ids() < 0)
+			return -1;
+	}
 
 	list_for_each_entry(si, &snapshot_head, l) {
 		if (!strncmp(si->snapshot_id, snapshot_id, PATH_MAX))
@@ -1139,8 +1132,10 @@ char *get_snapshot_id_from_idx(int idx)
 {
 	struct snapshot *si;
 
-	if (list_empty(&snapshot_head))
-		pull_snapshot_ids();
+	if (list_empty(&snapshot_head)) {
+		if (pull_snapshot_ids() < 0)
+			return NULL;
+	}
 
 	/* Note: if idx is the service fd then we need the current
 	 * snapshot_id idx. Else we need a parent snapshot_id idx.
--- a/criu/img-remote.c
+++ b/criu/img-remote.c
@@ -474,8 +474,9 @@ static struct roperation *handle_accept_cache_read(int cli_fd,
 			pr_perror("Error writing reply header for unexisting image");
 		close(cli_fd);
 		xfree(rop);
+		return NULL;
 	}
-	return NULL;
+	return rop;
 }
 
 static void forward_remote_image(struct roperation *rop)
@@ -505,6 +506,19 @@ static void forward_remote_image(struct roperation *rop)
 	event_set(epoll_fd, EPOLL_CTL_ADD, rop->fd, EPOLLOUT, rop);
 }
 
+static void dismiss_pending_image_files(void)
+{
+	struct roperation *rop, *tmp;
+
+	list_for_each_entry_safe(rop, tmp, &rop_pending, l) {
+		if (write_reply_header(rop->fd, ENOENT) < 0)
+			pr_perror("Error writing reply header for unexisting image");
+		close(rop->fd);
+		list_del(&rop->l);
+		xfree(rop);
+	}
+}
+
 static void handle_remote_accept(int fd)
 {
 	char path[PATH_MAX];
@@ -525,6 +539,13 @@ static void handle_remote_accept(int fd)
 	/* This means that the no more images are coming. */
 	else if (!ret) {
 		finished_remote = true;
+
+		/*
+		 * For all pending files to be sent that we have collected, we
+		 * now know they will never come. If they did come, they would
+		 * have been forwarded in finish_cache_write().
+		 */
+		dismiss_pending_image_files();
 		pr_info("Image Proxy connection closed.\n");
 		return;
 	}
@@ -566,7 +587,8 @@ static void handle_local_accept(int fd)
 
 	if (read_header(cli_fd, snapshot_id, path, &flags) < 0) {
 		pr_err("Error reading local image header\n");
-		goto err;
+		close(cli_fd);
+		return;
 	}
 
 	if (snapshot_id[0] == NULL_SNAPSHOT_ID && path[0] == FINISH) {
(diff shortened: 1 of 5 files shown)


adrianreber:
Not really familiar with this code but this looks good. I restarted the one failed Travis job because some repository troubles.

@nviennot Are you actively using this code? This is one of the features not present on master for some time already. Would be great to get it into shape to move it to the master branch.

nviennot:
Cool

I implemented an external proxy to do high performance checkpoint/restore. It's called criu-image-streamer. While I don't need CRIU's proxy/cache servers, I need CRIU to support remote images.

From what I can tell, the current proxy/cache server most likely harbor more bugs.

Once I open-source criu-image-streamer (within a week or two), my plan is to propose that it replaces the existing proxy/cache servers.

nviennot:
The tool I was referring to is now open-source: https://github.com/twosigma/criu-image-streamer

adrianreber:
Your image caching looks very interesting. Cool.

You need to rebase and currently the TLS based tests are broken because of expired certificates. Just that you know. PR to fix already exist.

nviennot:
Cool!

I've rebased the branch, and tested it.

criu/img-remote.c
```diff
@@ -178,6 +179,56 @@ static int setup_UNIX_client_socket(char *path)
...
+ * greatly improve performance. Data can be further spliced() by the proxy
+ * to transfer data with minimal overhead.
+ * Transfer rates of up to 15GB/s can be seen with this technique.
+ */
+#define SEND_FD_FOR_READ	0
+#define SEND_FD_FOR_WRITE	1
+#define RECV_FD			2
+static int upgrade_to_pipe(int sockfd, int mode)
+{
+	return sockfd;
```
rst0git, line 193: Is this `return sockfd;` intentional?

criu/img-remote.c
```diff
@@ -178,6 +179,56 @@ static int setup_UNIX_client_socket(char *path)
...
+		ret = fds[direction];
+		goto out_pipe;
+
+err_pipe:
+		close(fds[direction]);
+out_pipe:
+		close(fds[other_direction]);
+err:
+		close(sockfd);
+		return ret;
```
rst0git, line 228: maybe this part can be simplified to:
```C
		if (send_fd(sockfd, NULL, 0, fds[other_direction]) < 0)
			close(fds[direction]);
		else
			ret = fds[direction];

		close(fds[other_direction]);
err:
		close(sockfd);
		return ret;
```

criu/img-remote.c
```diff
@@ -950,7 +950,7 @@ int read_remote_image_connection(char *snapshot_id, char *path)
 		return -1;
 	}
 
-	if (!error || (snapshot_id[0] == NULL_SNAPSHOT_ID && path[0] != FINISH))
+	if (!error)
 		return sockfd;
```
rst0git, line 1016: With this change, the following `if (error == ENOENT) ` becomes redundant?

criu/img-remote.c
```diff
@@ -181,6 +179,56 @@ static int setup_UNIX_client_socket(char *path)
...
+
+	/*
+	 * If the other end of the pipe closes, the kernel will want to kill
+	 * us with a SIGPIPE. These signal must be ignored.
+	 */
+
+	if (mode == RECV_FD) {
+		int ret = recv_fd(sockfd);
+		close(sockfd);
+		return ret;
```
rst0git, line 203: `int ret = -1;` could be declared before the if-statement, then the following common part can be moved at the end of the function.
```C
		close(sockfd);
		return ret;
```

criu/img-remote.c
nviennot: Good catch, wow.

Not sure how this slipped through the cracks. Sorry.

criu/img-remote.c
nviennot: Yes, very good

criu/img-remote.c
nviennot: I don't think so. If error is ENOENT, we return -ENOENT, otherwise, we return -1 for all other errors.
We could return -error, but I don't recommend. It's better to control what kind of errors we allow to be returned.

criu/img-remote.c
nviennot: Very well, I pushed your suggestions.

criu/img-remote.c
rst0git: You are right. Thank you for clarifying.

nviennot:
There's a bug in the last commit where pending requests are not getting upgraded to pipes correctly. I'll be back with a fix.

nviennot:
I think it's good.

I use this little script to test the proxy/cache servers:
```bash
#!/bin/bash
sleep 3 & # The app to be checkpointed
APP_PID=$!

sudo criu/criu image-cache --port 5000 -v &
sleep 0.3
sudo criu/criu image-proxy --address 127.0.0.1 --port 5000 -v &
sleep 0.3

sudo criu/criu -D . --remote dump -t $APP_PID --shell-job -v &
sleep 0.3
sudo criu/criu -D . --remote restore --shell-job -v

wait
# Should exit successfully
```

There's definitely more bugs lurking in the proxy/cache code, but I don't think it's worth chasing them, as [criu-image-streamer](https://github.com/twosigma/criu-image-streamer) is a good replacement.

criu/img-remote.c
```diff
@@ -950,7 +950,7 @@ int read_remote_image_connection(char *snapshot_id, char *path)
 		return -1;
 	}
 
-	if (!error || (snapshot_id[0] == NULL_SNAPSHOT_ID && path[0] != FINISH))
```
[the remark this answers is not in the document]
nviennot: Good thinking.

printf() is actually safe with %s and NULL strings. When doing a `printf("%s", NULL)`, the printed string would be`(null)`, so we are good.

avagin:
> I use this little script to test the proxy/cache servers:

You can use zdtm.py to test the remote mode:

```
[root@fc24 criu]# python test/zdtm.py run --remote -t zdtm/static/env00 -f h -k always
=== Run 1/1 ================ zdtm/static/env00
========================== Run zdtm/static/env00 in h ==========================
Start test
./env00 --pidfile=env00.pid --outfile=env00.out --envname=ENV_00_TEST
Adding image cache
Adding image proxy
Run criu dump
ERROR: bad page counts, stats = 28 real = 0(0)
Run criu restore
Send the 15 signal to  44
Wait for zdtm/static/env00(44) to die for 0.100000
========================= Test zdtm/static/env00 PASS ==========================
```

but you will need to do a small hack

```
index b4a2b28b9..a832ed1c2 100755
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1219,7 +1219,7 @@ class criu:
         if (stats_written != r_pages) or (r_off != 0):
             print("ERROR: bad page counts, stats = %d real = %d(%d)" %
                   (stats_written, r_pages, r_off))
-            raise test_fail_exc("page counts mismatch")
+            #raise test_fail_exc("page counts mismatch")
 
     def dump(self, action, opts=[]):
         self.__iter += 1
```

need to investigate what wrong is there.

avagin:
> The external high performance proxy/cache I wrote is now open-source: https://github.com/twosigma/criu-image-streamer

This is awesome! I don't have objections to deprecate the criu remote proxy. It is not in the master branch, because it has not been completed and developers who worked on it moved to other projects.

Your implementation of the criu proxy can live in your repo or we can merge it in the criu repo. This is up to you. But if we want to merge all other required bits in the master branch, we need to start testing the proxy mode in travis-ci.

nviennot:
Wonderful!

Merging criu-image-streamer into the CRIU repo makes the most sense. As it will be the source of truth, I'll happily kill the twosigma repo.

avagin:
> What do you advise to do in terms of pull request? Should I take out the existing cache/proxy, and add criu-image-streamer?

Lets merge this pull request and then you will create a new pr which will add criu-image-streamer and remove the existing cache/proxy.

> Also, is it fine that it's Apache 2 licensed?

I think it should be okay. This is the separate tool and it can be licensed differently. @xemul what do you think about this?

criu/img-remote.c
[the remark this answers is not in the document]
Snorch: It depends on I think

```[snorch@snorch tmp]$ cat printf.c
#include <stdio.h>

int main ()
{
	printf("%s\n", NULL);
}
[snorch@snorch tmp]$ gcc -o printf printf.c 
[snorch@snorch tmp]$ ./printf 
Segmentation fault (core dumped)
```

adrianreber:
As this is Rust based I think it would make more sense to have it in another repository under checkpoint-restore. This is especially me talking from a packaging standpoint. Rust, with its own build-system, is packaged differently. If I would package it, it would be a separate package but it would be based on the same tarball if it is in the main criu repository.

Also, your proxy has the same license problem as go-criu. I see https://github.com/twosigma/criu-image-streamer/blob/master/proto/remote-image.proto and it says it is taken from the CRIU repository. The file is however released as GPLv2 and you are releasing it as Apache 2.0.
In go-criu we are only including the generated result and not the original proto file.

I would really love for CRIU to have all `images/*proto` files licensed under Apache 2.0 exactly for cases like this. According to git 26 persons have touched `images/`.

avagin:
> If I would package it, it would be a separate package but it would be based on the same tarball if it is in the main criu repository.

Another option is to create a new repo in https://github.com/checkpoint-restore/.

adrianreber:
> > If I would package it, it would be a separate package but it would be based on the same tarball if it is in the main criu repository.
> 
> Another option is to create a new repo in https://github.com/checkpoint-restore/.

That was what I was trying to propose :wink:

avagin:
I have merged this PR, so let's continue the discussion in https://github.com/checkpoint-restore/criu/issues/1061.

PR #1055, opened 2020-05-06
Update certificates for failing tls based tests

adrianreber:
When using zdtm.py with --tls it started to fail as the certificates seem to have expired. Following commands have been used to re-generate the certificate:

```
            # Generate CA key and certificate
            echo -ne "ca\ncert_signing_key" > temp
            certtool --generate-privkey > cakey.pem
            certtool --generate-self-signed \
                --template temp \
                --load-privkey cakey.pem \
                --outfile cacert.pem

            # Generate server key and certificate
            echo -ne "cn=$HOSTNAME\nencryption_key\nsigning_key" > temp
            certtool --generate-privkey > key.pem
            certtool --generate-certificate \
                --template temp \
                --load-privkey key.pem \
                --load-ca-certificate cacert.pem \
                --load-ca-privkey cakey.pem \
                --outfile cert.pem
            rm temp cakey.pem
```

Without this tests will fail in Travis.


adrianreber:
CC: @rst0git

PR #1056, opened 2020-05-07
Allow CRIU to support restoring into an existing PID namespace

adrianreber:
Trying to checkpoint one container of a pod in CRI-O fails with
```
Error (criu/namespaces.c:1081): Can't dump a pid namespace without the process init
```
because, as far as I understand it, pods can share namespaces. It seems the PID namespace if controlled by the initial container in that pod which runs a `pause` process.

To enable checkpointing containers out of a pod this PR enables CRIU to restore out of PID namespace and into another namespace. I can restore process from the host PID namespace in another PID namespace. Out of PID namespace into the host PID namespace. All directions are working.

Restoring into an existing PID namespace comes with possibility of PID collisions.

Using `clone3()` the whole thing is pretty simple: `setns(); clone3();`

Not using `clone3()` makes it much more complicated as it is necessary to write to `ns_last_pid` of the destination PID namespace and thus a helper is required: `setns()`, `fork()`, `open('ns_last_pid')`, `write()`, `close()`, `waitpid()`.

This seems to work so far, all tests are happy. After implementing it I am not sure it makes sense to checkpoint single containers out of a pod or if it would make more sense to checkpoint the complete pod. Not sure.


test/pki/cacert.pem
```diff
@@ -1,23 +1,23 @@
 -----BEGIN CERTIFICATE-----
```
rst0git, line 1: This commit was merged in the `criu-dev` branch.

criu/include/magic.h
```diff
@@ -96,6 +96,7 @@
 #define FILES_MAGIC		0x56303138 /* Toropets */
 #define MEMFD_INODE_MAGIC	0x48453499 /* Dnipro */
 #define TIMENS_MAGIC		0x43114433 /* Beslan */
+#define PIDNS_MAGIC		0x12345678
```
avagin, line 99: TODO: create correct magic

test/others/ns_ext/run.sh
```diff
@@ -0,0 +1,74 @@
...
+done
+
+if [[ "$NS" == "net" ]]; then
+	pid=$(cat pidfile)
+	pid2=$(cat pidfile2)
+elif [[ "$NS" == "pid" ]]; then
+	pid2=$(pgrep -f ". _run.sh pidfile $RND1" -n)
+	pid=$(pgrep -f "sh _run.sh pidfile2 $RND2" -n)
+fi
+
```
avagin, line 46: pls add some comments in this file, otherwise it is hard to figure out what is going on here.

test/others/ns_ext/run.sh
```diff
@@ -0,0 +1,74 @@
...
+touch $MNT1
+mount --bind /proc/$pid/ns/$NS $MNT1
+touch $MNT2
+mount --bind /proc/$pid2/ns/$NS $MNT2
+mkdir -p images
+ino=$(ls -iL $MNT1 | awk '{ print $1 }')
+ino2=$(ls -iL $MNT2 | awk '{ print $1 }')
+exec 33< $MNT1
+exec 34< $MNT2
+$CRIU dump -v4 -t $pid -o dump.log -D images --external $NS[$ino]:test_ns --external $NS[$ino2]:test_ns2
```
avagin, line 56: we need to check criu exit code. you can consider using set -e -o pipefail for the whole script.

criu/cr-restore.c
```diff
@@ -1385,36 +1408,91 @@ static inline int fork_with_pid(struct pstree_item *item)
...
+				 * so much easier and simpler. As long as CRIU supports
+				 * clone() this is needed.
+				 */
+				helper_pid = fork();
+				if (helper_pid < 0) {
+					pr_perror("Cannot fork ns_last_pid writer");
+					goto err_unlock;
+				}
+			}
+			if (helper_pid <= 0) {
```
avagin, line 1465: helper_pid == 0?

criu/cr-restore.c
```diff
@@ -1385,36 +1408,91 @@ static inline int fork_with_pid(struct pstree_item *item)
...
+				if (ret == -1 && helper_pid == -1)
+					goto err_unlock;
+				if (helper_pid == 0)
+					exit(ret);
+			}
+			if (helper_pid > 0) {
+				/* We forked and this is the parent. */
+				int status;
+
+				ret = waitpid(-1, &status, 0);
```
avagin, line 1476: waitpid(helper_pid, ...

criu/cr-restore.c
```diff
@@ -1385,36 +1408,91 @@ static inline int fork_with_pid(struct pstree_item *item)
...
+				if (ret < 0) {
+					pr_perror("Cannot wait for ns_last_pid writer");
+					goto err_unlock;
+				}
+				if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
```
avagin, line 1481: a process could be killed. IMHO: It is better to check status != 0.

criu/cr-restore.c
```diff
@@ -1385,36 +1408,91 @@ static inline int fork_with_pid(struct pstree_item *item)
...
+				}
+			}
+			if (helper_pid <= 0) {
+				ret = write_ns_last_pid(pid);
+				if (ret == -1 && helper_pid == -1)
+					goto err_unlock;
+				if (helper_pid == 0)
+					exit(ret);
+			}
+			if (helper_pid > 0) {
```
0x7f454c46, line 1472: I think, it would be a little easier to understand if you move `if (helper_pid <= 0)` into a helper funtion and `if (helper_pid > 0)` case into the branch with `if (pid_ns && pid_ns->ext_key)`

criu/cr-restore.c
```diff
@@ -1385,36 +1408,91 @@ static inline int fork_with_pid(struct pstree_item *item)
...
+					ret = -1;
+					goto err_unlock;
+				}
 			}
-			close(fd);
 		}
 	} else {
-		BUG_ON(pid != INIT_PID);
+		if (!(pid_ns && pid_ns->ext_key))
+			BUG_ON(pid != INIT_PID);
```
0x7f454c46, line 1490: Here too, if possible let's print error to the user and `return -1`.

criu/cr-restore.c
```diff
@@ -2163,9 +2244,6 @@ static int restore_root_task(struct pstree_item *init)
 	if (prepare_userns_hook())
 		return -1;
```
0x7f454c46, line 2245: Probably, `prepare_userns_hook()` should be moved too

Documentation/criu.txt
```diff
@@ -237,6 +237,12 @@ In other words, do not use it unless really needed.
     Tell *criu* that one end of a pair of UNIX sockets (created by
     *socketpair*(2)) with the given _id_ is OK to be disconnected.
 
+*--external* **pid[**__inode__**]:**__name__::
```
0x7f454c46, line 240: maybe `pidns[inode]:name`?
I don't insist though.

test/others/ns_ext/run.sh
```diff
@@ -0,0 +1,74 @@
...
+	pid=$(pgrep -f "^sh _run.sh pidfile2 $RND2")
+fi
+new_ino=$(ls -iL /proc/$pid/ns/$NS | awk '{ print $1 }')
+new_ino2=$(ls -iL /proc/$pid2/ns/$NS | awk '{ print $1 }')
+[ "$ino" != "$new_ino" ] && {
+	echo FAIL
+	exit 1
+}
+[ "$ino2" != "$new_ino2" ] && {
+	echo FAIL
```
0x7f454c46, line 70: Probably, it could print what has failed before exiting.

adrianreber:
Thanks for the reviews everyone. I will update this PR.

@avagin Can you create a new magic for me?

images/pidns.proto
```diff
@@ -0,0 +1,5 @@
+syntax = "proto2";
+
+message pidns_entry {
+	optional string ext_key		= 1;
```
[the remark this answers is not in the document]
adrianreber: I was also thinking about this. Not sure. I was looking at the external network namespace support implementation and that uses optional. With PID namespace support it is however the only entry in the protobuf definition. Having the only field as optional is kind of strange.

So, yes, `required` would be probably a good idea right now as the file will only exist if `ext_key` is set. If we ever get another entry here then it would be better to be `optional`.

Interestingly I just read that the `proto3` format has dropped `required` and `optional` completely.

I will change it to `required` if we think that this is better. I am unsure myself what do to with it.

Documentation/criu.txt
adrianreber: The reason for `pid` and not `pidns` is again the external network namespace support which uses `net`. I do not really care what name it has. Just following existing naming.

criu/cr-restore.c
adrianreber: No, because I tried to have only one call to write_ns_last_pid(), but the code has changed because I used `call_in_child_process()` as suggested by @Snorch

criu/cr-restore.c
```diff
@@ -1335,9 +1335,32 @@ static bool needs_prep_creds(struct pstree_item *item)
 	return (!item->parent && ((root_ns_mask & CLONE_NEWUSER) || getuid()));
 }
 
+static int write_ns_last_pid(pid_t pid)
```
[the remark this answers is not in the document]
adrianreber: done

criu/cr-restore.c
```diff
@@ -1335,9 +1335,32 @@ static bool needs_prep_creds(struct pstree_item *item)
...
+	int len;
+	int fd;
+
+	fd = open_proc_rw(PROC_GEN, LAST_PID_PATH);
+	if (fd < 0)
+		return -1;
+
+	len = snprintf(buf, sizeof(buf), "%d", pid - 1);
+	if (write(fd, buf, len) != len) {
+		pr_perror("%d: Write %s to %s", pid, buf,
```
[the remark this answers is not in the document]
adrianreber: done

criu/cr-restore.c
```diff
@@ -1385,36 +1408,91 @@ static inline int fork_with_pid(struct pstree_item *item)
...
-				close(fd);
-				goto err_unlock;
+			if (pid_ns && pid_ns->ext_key) {
+				/*
+				 * Restoring into another namespace requires a helper
+				 * to write to LAST_PID_PATH. Using clone3() this is
+				 * so much easier and simpler. As long as CRIU supports
+				 * clone() this is needed.
+				 */
+				helper_pid = fork();
```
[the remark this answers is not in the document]
adrianreber: done

criu/cr-restore.c
adrianreber: done

adrianreber:
I was able to address most things which came up during the review:

  * more comments in the test script
  * switched to `call_in_child_process` instead of doing `fork` manually
  * renamed function to `set_next_pid`
  * replaced `BUG` macros with error messages

If it should be called `pid` or `pidns` when using this via `--external` is still open, but because the network namespace support also only uses `net` and not `netns` I am :+1: for `pid` even if `pidns` is clearer.

I am also unsure if the protobuf field should be `required` or `optional`.

I also need someone to create a magic value which fits the existing scheme.

criu/include/magic.h
avagin: You can use 0x61157326 /* Surgut */ or you can choose any other city that you like.
I think you can even choose a German city. For example, it can be 0x52311324 /* Berlin */, but in this case, you need to fix a comments which is right before the list of this constants.

adrianreber:
Tests are happy right now, but sometimes I see errors like this (centos):
```
+ ../../../criu/criu dump -v4 -t 31156 -o dump.log -D images --external 'pid[4026532585]:test_ns' --external 'pid[4026532585]:test_ns2'
+ RESULT=1
+ cat images/dump.log
+ grep -B 5 Error
(00.004863) Dumping task cwd id 0x6 root id 0x7
(00.004909) Dumping file-locks
(00.004915) 
(00.004916) Dumping pstree (pid: 31156)
(00.004918) ----------------------------------------
(00.004919) Error (criu/pstree.c:297): The root process 8 is not a session leader. Consider using --shell-job option
(00.007373) Unlock network
(00.007383) Unfreezing tasks into 1
(00.007386) 	Unseizing 31156 into 1
(00.007401) Error (criu/cr-dump.c:1775): Dumping FAILED.
+ '[' 1 '!=' 0 ']'
+ echo 'CRIU dump failed'
CRIU dump failed
```
or on alpine
```
+ ../../../criu/criu dump -v4 -t 1216 -o dump.log -D images --external 'pid[4026532743]:test_ns' --external 'pid[4026532743]:test_ns2'
+ RESULT=1
+ + cat images/dump.log
grep -B 5 Error
(00.003015) Wait for ack 70 on daemon socket
pie: 7: __sent ack msg: 70 70 0
pie: 7: Daemon waits for command
(00.003029) Fetched ack: 70 70 0
(00.003033) sid=0 pgid=0 pid=7
(00.003037) Error (criu/cr-dump.c:1332): A session leader of 1216(7) is outside of its pid namespace
--
(00.003425) 1216 (native) is going to execute the syscall 15, required is 15
(00.003442) 1216 was stopped
(00.003702) Unlock network
(00.003711) Unfreezing tasks into 1
(00.003714) 	Unseizing 1216 into 1
(00.003730) Error (criu/cr-dump.c:1775): Dumping FAILED.
+ '[' 1 '!=' 0 ']'
+ echo 'CRIU dump failed'
CRIU dump failed
+ echo FAIL
FAIL
```

Not sure why this happens. Is this somehow racy? But it is during `dump` which did not really change in this PR. Any ideas recommendations if my test setup is wrong?

Snorch:
Maybe the difference is in how you run your root process, setsid wrapper might help.

adrianreber:
> Maybe the difference is in how you run your root process, setsid wrapper might help.

Could you give some more details what you mean. I am already using setsid and I also tried it with a setsid before the `unshare`. The problem is that I only see it sometimes on alpine or centos. Non of the other distribution have failed so far.

Snorch:
>Error (criu/cr-dump.c:1332): A session leader of 1216(7) is outside of its pid namespace

It means that your root task with pid [1216, 7] (1216 in criu pidns and 7 in it's own pidns) has sid [smth, 0], which means that session leader is from lower pid namespace, likely from criu one or even lower.

Your task [1216, 7] should do setsid before exec and have sid [1216, 7], so that where will be no problem anymore. You are obviousely using setsid somehow wrong as your root task is not a session leader.

adrianreber:
Are there any further comments regarding `required` or `optional` in the protobuf file?

adrianreber:
Another ping. The only open point so far is `optional` or `required`.

test/others/ns_ext/run.sh
```diff
@@ -0,0 +1,74 @@
...
+touch $MNT1
+mount --bind /proc/$pid/ns/$NS $MNT1
+touch $MNT2
+mount --bind /proc/$pid2/ns/$NS $MNT2
+mkdir -p images
+ino=$(ls -iL $MNT1 | awk '{ print $1 }')
+ino2=$(ls -iL $MNT2 | awk '{ print $1 }')
+exec 33< $MNT1
+exec 34< $MNT2
+$CRIU dump -v4 -t $pid -o dump.log -D images --external $NS[$ino]:test_ns --external $NS[$ino2]:test_ns2
```
avagin, line 56: in the case of pidns, ino and ino2 will be equal to each other?

test/others/ns_ext/run.sh
```diff
@@ -0,0 +1,74 @@
...
+	rm -f pidfile pidfile2 pidfile3
+}
+
+CRIU=../../../criu/criu
+if [[ "$NS" == "net" ]]; then
+	setsid unshare -n bash -c 'unshare -n sh _run.sh pidfile2 & unshare -n sh _run.sh pidfile3 & ip link add xxx type veth && ip link add mymacvlan1 link xxx type macvlan mode bridge && . _run.sh pidfile' < /dev/zero &> output &
+elif [[ "$NS" == "pid" ]]; then
+	RND1=$RANDOM
+	RND2=$RANDOM
+	setsid unshare -p -f setsid bash -c "setsid sh _run.sh pidfile2 $RND2 & . _run.sh pidfile $RND1" < /dev/zero &> output &
```
avagin, line 31: It would be nice to have a comment which explains these three setsid-s

test/others/ns_ext/run.sh
avagin: And I think you don't dump pid2, do you? It would be nice to have some comments in this script.

adrianreber: I hope you are aware that you wrote that script initially before I extended it to also test restoring into an existing PID namespace :wink:

test/others/ns_ext/run.sh
adrianreber: I don't know. Only with a`setsid` at all three places it succeeds with Ubuntu, CentOS and Alpine. I do not understand this part. It seems, especially the alpine busybox behaves differently.

test/others/ns_ext/run.sh
avagin: Very often, I write fewer comments than I have to;).

PR #959, opened 2020-02-20
libcriu: Add orphan pts master

rst0git:
The orphan pts master option was introduced with https://github.com/checkpoint-restore/criu/commit/6afe523d97d59e6bf29621b8aa0e6a4332f710fc to enable checkpoint/restore of containers with a pty pair used as console.

PR #1063, opened 2020-05-13
lib/c: extend receive to handle incoming FDs

adrianreber:
When using libcriu with the notify callback functionality CRIU transmits an FD during 'orphan-pts-master' back to libcriu user. This is message is sent via sendmsg() to transmit the FD and not via write() as all other protobuf messages.

libcriu was using recv() and to be able to receive the FD this needs to be changed to recvmsg() and if an FD is attached to it (currently only for 'orphan-pts-master' this FD is stored in a variable which can be retrieved with the function criu_get_orphan_pts_master_fd().

This (and #959) are the last changes to make it possible to use crun instead of runc to checkpoint and restore a container with Podman.

### DIFF
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -34,6 +34,7 @@ struct criu_opts {
 
 static criu_opts *global_opts;
 static int saved_errno;
+static int orphan_pts_master_fd = -1;
 
 void criu_free_service(criu_opts *opts)
 {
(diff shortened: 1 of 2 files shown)


lib/c/criu.c
```diff
@@ -1134,24 +1135,60 @@ int criu_set_page_server_address_port(const char *address, int port)
...
 	}
 
-	len = recv(socket_fd, buf, len, MSG_TRUNC);
+	io.iov_base = buf;
+	io.iov_len = len;
+	msg_hdr.msg_iov = &io;
+	msg_hdr.msg_iovlen = 1;
+	msg_hdr.msg_control = buf + len;
+	msg_hdr.msg_controllen = 32;
+	len = recvmsg(socket_fd, &msg_hdr, 0);
```
rst0git, line 1170: Should the `MSG_TRUNC` flag be set?

lib/c/criu.c
```diff
@@ -1134,24 +1135,60 @@ int criu_set_page_server_address_port(const char *address, int port)
...
+
+	/*
+	 * This will be NULL if no FD is in the message. Currently
+	 * only a response with script set to 'orphan-pts-master'
+	 * has an FD in the ancillary data.
+	 */
+	cmsg = CMSG_FIRSTHDR(&msg_hdr);
```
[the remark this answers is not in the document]
adrianreber: yes, I moved it to the right place. I think we do not need explicit error handling for `len == 0`.

lib/c/criu.c
adrianreber: done

lib/c/criu.c
```diff
@@ -1134,24 +1135,60 @@ int criu_set_page_server_address_port(const char *address, int port)
 
 static CriuResp *recv_resp(int socket_fd)
 {
+	struct msghdr msg_hdr = {0};
 	unsigned char *buf = NULL;
-	int len;
+	struct cmsghdr * cmsg;
```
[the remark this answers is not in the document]
adrianreber: done

adrianreber:
I will include a test once #959 has been merged.

lib/c/criu.c
```diff
@@ -1134,29 +1135,65 @@ int criu_set_page_server_address_port(const char *address, int port)
...
+		if (cmsg->cmsg_type != SCM_RIGHTS) {
+			errno = EINVAL;
+			goto err;
+		}
+		/* CTRUNC will be set if msg_hdr.msg_controllen is too small. */
+		if (msg_hdr.msg_flags & MSG_CTRUNC) {
+			errno = ENFILE;
+			goto err;
+		}
+		memcpy(&orphan_pts_master_fd, CMSG_DATA(cmsg), sizeof(int));
```
avagin, line 1194: orphan_pts_master_fd = *(int *)CMSG_DATA(cmsg)

lib/c/criu.c
```diff
@@ -1134,29 +1135,65 @@ int criu_set_page_server_address_port(const char *address, int port)
...
-	int len;
+	struct cmsghdr *cmsg;
 	CriuResp *msg = 0;
+	struct iovec io;
+	int len;
 
+	/* Check the size of the waiting data. */
 	len = recv(socket_fd, NULL, 0, MSG_TRUNC | MSG_PEEK);
 	if (len == -1) {
 		perror("Can't read request");
```
avagin, line 1148: Is it ok to print error messages from the library?

lib/c/criu.c
```diff
@@ -1134,29 +1135,65 @@ int criu_set_page_server_address_port(const char *address, int port)
...
 		goto err;
 	}
 
-	buf = malloc(len);
+	/*
+	 * If there is an FD attached to the protobuf message from CRIU
+	 * the FD will be in the ancillary data. Let's reserve additional
+	 * 32 bytes for that.
+	 */
+	buf = malloc(len+32);
```
avagin, line 1157: cmsg_len =  CMSG_LEN(sizeof(int));
buf = malloc(len + cmsg_len)

lib/c/criu.c
adrianreber: Not sure. Another interface might have been better where the error messages are handled by the library user, but there are multiple perror()s in the library right now.

lib/c/criu.c
adrianreber: I initially had it like this and have changed it back to be like this now.

lib/c/criu.c
adrianreber: done

adrianreber:
Rebased and applied changes from review.

lib/c/criu.c
adrianreber: Now, thanks to Travis I remember why I did a memcpy. On ARCH=x86_64 I get
```
error: dereferencing type-punned pointer will break strict-aliasing rules
```

adrianreber:
Is there anything missing here?

lib/c/criu.c
avagin: In this case, we need to have a comment which explains why we use memcpy here.

adrianreber: done

adrianreber:
ping

adrianreber:
Rebased and force pushed.

PR #1069, opened 2020-05-16
[RFC] swrk: send notification instead of using status fd

kolyshkin:
When we use swrk, we have a mechanism to send notifications over RPC.
It is cleaner and more straightforward than sending `\0` to status fd.

For now, both mechanisms are supported, although status fd request
option is now deprecated, so a warning is logged in case it's used.

Guess we can remove it in a few years.

The corresponding runc PR is here: https://github.com/opencontainers/runc/pull/2414

### DIFF
--- a/criu/action-scripts.c
+++ b/criu/action-scripts.c
@@ -29,6 +29,7 @@ static const char *action_names[ACT_MAX] = {
 	[ ACT_PRE_RESUME ]	= "pre-resume",
 	[ ACT_POST_RESUME ]	= "post-resume",
 	[ ACT_ORPHAN_PTS_MASTER ] = "orphan-pts-master",
+	[ ACT_STATUS_READY ]    = "status-ready",
 };
 
 struct script {
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -1385,7 +1385,7 @@ int cr_service(bool daemon_mode)
 	if (setup_sigchld_handler())
 		goto err;
 
-	if (close_status_fd())
+	if (status_ready())
 		goto err;
 
 	while (1) {
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -671,6 +671,8 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	}
 
 	if (req->has_status_fd) {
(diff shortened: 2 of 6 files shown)


kolyshkin:
@avagin @adrianreber PTAL

kolyshkin:
CI failure  on alpine:

```
================== Run zdtm/static/socket-tcp4v6-local in uns ==================
10067Start test
10068./socket-tcp4v6-local --pidfile=socket-tcp4v6-local.pid --outfile=socket-tcp4v6-local.out
10069Run criu dump
10070Run criu restore
10071Send the 15 signal to  115
10072Wait for zdtm/static/socket-tcp4v6-local(115) to die for 0.100000
10073########## Test zdtm/static/socket-tcp4v6-local FAIL at result check ###########
10074Test output: ================================
1007501:54:50.097:     5: ERR: tcp.c:128: can't connect to server (errno = 106 (Socket is connected))
1007601:54:50.097:     4: ERR: socket-tcp4v6-local.c:201: read less then have to (errno = 11 (Resource temporarily unavailable))
10077
10078 <<< ================================
```

Is this a known flaky test?

adrianreber:
I restarted the test and now it passed. Have not seen this fail before.

Right now I am still undecided about this. It seems to make sense to use the existing notification channel. From that point I like it. On the other hand, the simplicity of passing down an FD to CRIU and let CRIU write to that FD also seems like the right thing to do. If the FD interface would not exist I would say :+1: , but with the existing interface it feels redundant, but also easier to understand and use.

Undecided but why not. So LGTM.

kolyshkin:
> Right now I am still undecided about this. It seems to make sense to use the existing notification channel. From that point I like it. On the other hand, the simplicity of passing down an FD to CRIU and let CRIU write to that FD also seems like the right thing to do. If the FD interface would not exist I would say ￼ , but with the existing interface it feels redundant, but also easier to understand and use.

Try to look at it not from a CRIU developer but from a user's perspective. Currently, runc has to use two different mechanisms -- fd passing (for "I'm ready" communication) and rpc notifications (for everything else). This seems odd.

It's not that one mechanism is better than the other. It's more about unification. With this, all the notifications are going through the same channel.

adrianreber:
@kolyshkin Can you update https://criu.org/Action_scripts ?

PR #1070, opened 2020-05-17
actions: create file for daily rebuild

adrianreber:
This adds a definition to do a daily rebuild of all cross-compile tests on the master and criu-dev branch.

Follow up on #1067 to also have daily cron-like builds like defined in Travis for the builds moved to github actions.

### DIFF
--- a/.github/workflows/cross-compile-daily.yml
+++ b/.github/workflows/cross-compile-daily.yml
@@ -1,10 +1,8 @@
-name: Cross Compile Tests
+name: Daily Cross Compile Tests
 
 on:
-  push:
-    branches: [criu-dev, master]
-  pull_request:
-    branches: [criu-dev, master]
+  schedule:
+    - cron:  '30 * * * *'
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


.github/workflows/cross-compile.yml
```diff
@@ -5,8 +5,6 @@ on:
     branches: [ criu-dev ]
   pull_request:
     branches: [ criu-dev ]
```
avagin, line 7: offtopic: do we need the master branch here?

.github/workflows/cross-compile.yml
```diff
@@ -5,8 +5,6 @@ on:
...
-  schedule:
-    - cron:  '55 5 * * *'
```
avagin, line 9: why do we need to move this in a new file?

.github/workflows/cross-compile.yml
adrianreber: yes, that sounds useful, let met add that.

.github/workflows/cross-compile.yml
adrianreber: Maybe because I was not able to correctly understand the documentation :wink: 

What I wanted to do is set up a daily build for `master` and `criu-dev`. The problem is that it is not possible, as far as I understood it, to specify a branch in the `schedule` section. To have a daily build on two branches I have to use the matrix thing like I did and change the checkout section to use that. This means, that I am no longer checking out the PR this is supposed to run, but always `master` and `criu-dev`. Something like this.

From my understanding it is basically a short-coming that I can tell which actions to run the schedule on easily. I am also not happy to have the same content, what to build, in two files now, but I have not found a better solution.

adrianreber:
@avagin can this be merged?

avagin:
Pls you rebase this pr and I will merge it. Sorry for the inconvenience.

adrianreber:
Rebased and force pushed.

Issue #58, opened 2015-09-24
Library/API to access potentially stacked pagemap+pages


xemul:
Pulled from criu.org/Todo

The [memory dumps](http://criu.org/Memory_dumps) are quite sophisticated. It would be nice to have a library or API to access the data in them using some simple API. IOW -- API-ize the page_read.c

PR #1071, opened 2020-05-18
Replace CONFIG_ARMV* with CONFIG_CPU_V*

thaJeztah:
based on this discussion: https://lists.denx.de/pipermail/u-boot/2015-December/239646.html,
CONFIG_ARMV* was replaced by CONFIG_CPU_V*, so changing it here as well:

> The arch/arm/lib/cache-cp15.c checks for CONFIG_ARMV7 and if this macro is
> set, it configures TTBR0 register. This register must be configured for the
> cache on ARMv7 to operate correctly.
>
> The problem is that noone actually sets the CONFIG_ARMV7 macro and thus the
> TTBR0 is not configured at all. On SoCFPGA, this produces all sorts of minor
> issues which are hard to replicate, for example certain USB sticks are not
> detected or QSPI NOR sometimes fails to write pages completely.
>
> The solution is to replace CONFIG_ARMV7 test with CONFIG_CPU_V7 one. This is
> correct because the code which added the test(s) for CONFIG_ARMV7 was added
> shortly after CONFIG_ARMV7 was replaced by CONFIG_CPU_V7 and this code was
> not adjusted correctly to reflect that change.


thaJeztah:
Some background for this patch (and a **big** disclaimer!).

I was running into a problem in our build pipeline, where CRIU failed to compile on our ARM machines;

> Hmm.. static builds on ARM are failing (both ARM32 and ARM64);
>
> ```
> #58 52.08 In file included from include/common/lock.h:9:0,
> #58 52.08                  from compel/plugins/std/infect.c:5:
> #58 52.08 include/common/asm/atomic.h:61:2: error: #error ARM architecture version (CONFIG_ARMV*) not set or unsupported.
> #58 52.08  #error ARM architecture version (CONFIG_ARMV*) not set or unsupported.
> #58 52.08   ^~~~~
> ```

This turned out to be because we were testing some Jenkins nodes to do arm32 builds on arm64 machines (and possibly some mis-configuration), but before that I started some Googling to find about those variables, and discovered the discussion linked in the top description. Doing a search in the kernel code, I confirmed that `CONFIG_CPU_VXX` look to be what's used in the kernel now; https://github.com/torvalds/linux/blob/192ffb7515839b1cc8457e0a8c1e09783de019d3/arch/arm/boot/compressed/misc.c#L33

I am _not_ a C developer, nor do I know much about the CRIU internals, so perhaps these were not renamed _for a reason_. I had this branch around on my machine (it's mostly just a "find/replace"), so thought to at least open it as a pull request for the maintainers here to have a look. Feel free to take this patch, modify it if needed, or close it if it doesn't make sense :joy:

Also happy to make changes and update the PR if there are suggestions.

thaJeztah:
@kolyshkin @avagin ptal 🤗

adrianreber:
@thaJeztah We are using the following steps to build CRIU on arm32 on Travis's aarch64 systems:

https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/build/Dockerfile.linux32.tmpl

Would that help you?

thaJeztah:
@adrianreber thanks! I'll check with @stefanscherer to see if that helps (IIRC, dynamically linked builds worked, but static builds failed, but as mentioned, it could've been an incomplete configuration of those machines)

adrianreber:
@thaJeztah I had a closer look at your code changes. Does that actually solve your problem? Because it seems you are only renaming a define CRIU is setting itself. As far as I understand it, it could be named whatever we want as I would not expect the internal CONFIG_CPU_V7 to be exported to user-space headers in the kernel.

thaJeztah:
No, it didn't solve the problem :joy: but had the branch still around, and thought it might be of interest to you (given that these were renamed in the kernel). Admitted, perhaps they were not renamed on purpose (to prevent custom values  with conflicting with the _actual_ values?)

StefanScherer:
Thanks @adrianreber for the link. We also use `setarch linux32`, but already for running dockerd and containerd with 32bit binaries. I‘ll take a closer look what we can do in our CI environment.

0x7f454c46:
Please, try running this:
https://github.com/0x7f454c46/criu/commit/1ac22e91d9be54be224c78c3affeaa7db05fef66
to see what happens with `$ARMV`

thaJeztah:
@0x7f454c46 cherry-picked your commit in this branch

0x7f454c46:
@thaJeztah, please run it in your build environment: it should fail if the scripts fail to properly detect arm architecture version (armv7/armv8/armv6) which I suppose is your original problem.
So, it'll be possible to fix the script knowing what differs in your env.
Thanks,
Dmitry

adrianreber:
If this is still relevant for someone please rebase and reopen. Closing it for now.

PR #1074, opened 2020-05-20
travis: switch travis to Python 3

adrianreber:
Python2 has reached its EOL. Let's switch to Python3 in our Travis runs.

We are still using Python 2 in the CentOS 7 test run.

### DIFF
--- a/scripts/travis/travis-tests
+++ b/scripts/travis/travis-tests
@@ -1,12 +1,26 @@
 #!/bin/bash
 set -x -e
 
-TRAVIS_PKGS="protobuf-c-compiler libprotobuf-c-dev libaio-dev python-future
-		libgnutls28-dev libgnutls30 libprotobuf-dev protobuf-compiler
-		libcap-dev libnl-3-dev gdb bash python-protobuf python-yaml
-		libnet-dev util-linux asciidoctor libnl-route-3-dev
-		python-junit.xml python-ipaddress time ccache flake8
-		libbsd-dev"
+TRAVIS_PKGS="protobuf-c-compiler libprotobuf-c-dev libaio-dev libgnutls28-dev
+		libgnutls30 libprotobuf-dev protobuf-compiler libcap-dev
+		libnl-3-dev gdb bash libnet-dev util-linux asciidoctor
+		libnl-route-3-dev time ccache flake8 libbsd-dev"
+
+
+if [ -e /etc/lsb-release ]; then
+	# This file does not exist on non Ubuntu
+	. /etc/lsb-release
+	if [ "$DISTRIB_RELEASE" = "16.04" ]; then
+		# There is one last test running on 16.04 because of the broken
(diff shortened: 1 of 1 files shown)


adrianreber:
This also fixes the latest criu-dev alpine failures.

scripts/travis/travis-tests
```diff
@@ -59,6 +73,10 @@ travis_prep () {
 	apt-get update -qq
 	apt-get install -qq --no-install-recommends $TRAVIS_PKGS
 	chmod a+x $HOME
+
+	# zdtm uses an unversioned python binary to run the tests.
+	# let's point python to python3
+	ln -sf /usr/bin/python3 /usr/bin/python
```
avagin, line 79: I don't like this hack. Why we can't use the default python version for each distribution?

avagin:
> This also fixes the latest criu-dev alpine failures.

This has to be in a separate commit. And I would create another PR for this

scripts/travis/travis-tests
adrianreber: Not every distribution sets a python link. When there was only python2, python was a link to python2. When python3 was released there was a PEP that there should be not python link to python3. But that was later changed. I think. So there are some distributions which are providing a python to python3 link and some do not provide such a link.

As we are using an unversioned python in zdtm we need an unversioned python link. From what I have seen neither ubuntu nor alpine do have a python to python3 link. At least the versions I checked.

adrianreber:
> > This also fixes the latest criu-dev alpine failures.
> 
> This has to be in a separate commit. And I would create another PR for this

Will do that.

scripts/travis/travis-tests
avagin: Why do we need to use python3 everwhere? 

For example, we know that Fedora uses python3 by default:
```
[avagin@laptop ~]$ env python -V
Python 3.7.6
[avagin@laptop ~]$ cat /etc/fedora-release 
Fedora release 31 (Thirty One)
```

I think it is enough if one of jobs will run on python3, is it?

adrianreber:
alpine fixes available in #1089

scripts/travis/travis-tests
adrianreber: > I think it is enough if one of jobs will run on python3, is it?

I think it is more important to run as much on Python 3 and only use Python2 for old distributions like CentOS 7 and Ubuntu 16.04.

rst0git: I think Adrian refers to [PEP 394 -- The "python" Command on Unix-Like Systems](https://www.python.org/dev/peps/pep-0394/).
An alternative solution could be:
```
python3 -m venv virtualenv
source virtualenv/bin/activate
```

avagin:
@adrianreber could you rebase this pr

adrianreber:
rebased and force pushed

PR #1075, opened 2020-05-20
Fix three small code checker reports

adrianreber:


adrianreber:
See #1076 #1077 #1078 for unfixed reports from the same run.

PR #1080, opened 2020-05-23
tests: run clang based scan-build

adrianreber:
This adds a clang based code analysis to our CI runs. The result is published as a zip file and github keeps these artifacts for 90 days.


adrianreber:
No longer sure if this is really useful. It would make it easier to look at the results of the static analyzer, but if nobody ever looks it does not make much sense. On the other hand it does not hurt running this.

Maybe if we would publish the results at a place where it is not put in a zip file it would make it easier to have a look. Do we have a place where we could push the resulting html pages for easier consumption?

avagin:
How useful are these results? Can we detect new issues for PR-s?

adrianreber:
> How useful are these results? Can we detect new issues for PR-s?

We could. Once we are able to have no warnings when running scan-build new warnings could be reported during PRs. First goal would be to come to a state were no warnings are left during scan-build. Either we fix them, or we mark them as false positive.

It would also be nice if we could automatically upload the results to criu.org via SSH. github offers the possibility to add secrets which could be used to store an SSH key which can connect to criu.org and upload the result. Having a place to easily see the results via browser would make this more useful than it being a github action artifact.

Possible steps to get this into PR testing:
 * Have the possibility to upload the results to criu.org (or somewhere else)
 * get CRIU warning free by fixing or marking as invalid
 * activate the results in PRs to not introduce new warnings

adrianreber:
scan-build does detect too many false positives. Closing.

PR #1084, opened 2020-05-29
Restrict Checks for Open/MMAPed Files - CLI & Build-ID

AnorexicAtticusFinch:
This adds CLI options and the build-ID file verification method in addition to the current file size check as part of GSoC 2020.

This will make CRIU use the build-ID validation method by default whenever possible (The file size check is now a preliminary check) and if that cannot be done, it will use only the file size check. To explicitly use only the file size check all the time, the following command line option can be used: `--file-validation="filesize"`
In other words, `--file-validation="buildid"` will not make a difference as this is the default method.

Mentors: @avagin @0x7f454c46

### DIFF
--- a/criu/config.c
+++ b/criu/config.c
@@ -280,6 +280,7 @@ void init_opts(void)
 	opts.status_fd = -1;
 	opts.log_level = DEFAULT_LOGLEVEL;
 	opts.pre_dump_mode = PRE_DUMP_SPLICE;
+	opts.file_validation_method = FILE_VALIDATION_DEFAULT;
 }
 
 bool deprecated_ok(char *what)
@@ -420,6 +421,22 @@ static int parse_join_ns(const char *ptr)
 	return 0;
 }
 
+static int parse_file_validation_method(struct cr_options *opts, const char *optarg)
+{
+	if (!strcmp(optarg, "filesize"))
+		opts->file_validation_method = FILE_VALIDATION_FILE_SIZE;
+	else if (!strcmp(optarg, "buildid"))
+		opts->file_validation_method = FILE_VALIDATION_BUILD_ID;
+	else
+		goto Esyntax;
+
+	return 0;
+
+Esyntax:
+	pr_err("Unknown file validation method `%s' selected\n", optarg);
+	return -1;
+}
+
 /*
  * parse_options() is the point where the getopt parsing happens. The CLI
  * parsing as well as the configuration file parsing happens here.
@@ -523,6 +540,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{"tls-no-cn-verify",		no_argument,		&opts.tls_no_cn_verify, true},
 		{ "cgroup-yard",		required_argument,	0, 1096 },
 		{ "pre-dump-mode",		required_argument,	0, 1097},
+		{ "file-validation",		required_argument,	0, 1098	},
 		{ },
 	};
 
@@ -831,6 +849,10 @@ int parse_options(int argc, char **argv, bool *usage_error,
 				pr_err("Unable to parse value of --pre-dump-mode\n");
 				return 1;
 			}
+			break;
+		case 1098:
+			if (parse_file_validation_method(&opts, optarg))
+				return 2;
 			break;
 		case 'V':
 			pr_msg("Version: %s\n", CRIU_VERSION);
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -12,6 +12,7 @@
 #include <sched.h>
 #include <sys/capability.h>
 #include <sys/mount.h>
+#include <elf.h>
 
 #ifndef SEEK_DATA
 #define SEEK_DATA	3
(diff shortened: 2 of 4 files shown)


criu/files-reg.c
```diff
@@ -1380,6 +1380,14 @@ static bool should_check_size(int flags)
 	return true;
 }
 
+/* Stores the size of a file so that it can be validated while restoring. */
+static inline void store_validation_data_file_size(RegFileEntry *rfe, const struct fd_parms *p)
+{
```
rppt, line 1385: To simplify addition of new validation methods, you might want to call this `store_validation_data` from the beginning and update the comment respectively.
If it grows too large we can split it later

criu/files-reg.c
```diff
@@ -1726,6 +1733,18 @@ static int rfi_remap(struct reg_file_info *rfi, int *level)
 	return 0;
 }
 
+/* Compares the file's size with the stored value. */
+static inline bool validate_with_file_size(const struct stat *fd_status,
+					const struct reg_file_info *rfi)
```
rppt, line 1738: and here `validate_file` should be enough, IMO

avagin:
Commit messages has to be wrapped at 75 columns.

images/regfile.proto
```diff
@@ -4,13 +4,17 @@ import "opts.proto";
...
+	required uint32		id			= 1;
+	required uint32		flags			= 2 [(criu).flags = "rfile.flags"];
+	required uint64		pos			= 3;
+	required fown_entry	fown			= 5;
+	required string		name			= 6;
+	optional sint32		mnt_id			= 7 [default = -1];
+	optional uint64		size			= 8;
+	optional bool		ext			= 9;
+	optional uint32         mode    		= 10;
+	repeated uint32 	build_id 		= 11;
```
avagin, line 16: I don't think that we need to add all of them in one commit. I think it is better to add a field together with the implementation of its check. And you can create a separate PR for each check. It is easier to review small PR-s.

criu/files-reg.c
AnorexicAtticusFinch: The code for the build-id method is getting kinda large so I broke everything down into smaller functions preemptively

AnorexicAtticusFinch:
> Commit messages has to be wrapped at 75 columns.

Oh, I'll keep that in mind from now on.

images/regfile.proto
AnorexicAtticusFinch: In that case, I'll use this for the CLI options and build-ID?

criu/files-reg.c
rppt: That's ok to have several functions, but there still need to be "the top level validation" function and `dump_one_reg_file` will call it.
And I think having simple checks like file size open coded in that function is ok.

images/regfile.proto
avagin: I don't understand what you mean. Could you elaborate?

AnorexicAtticusFinch: > I don't think that we need to add all of them in one commit. I think it is better to add a field together with the implementation of its check. And you can create a separate PR for each check. It is easier to review small PR-s.

I'll implement the build-ID check in this PR along with the command line arguments

avagin: yes, this is right.

criu/files-reg.c
AnorexicAtticusFinch: @rppt I've made a new commit along with the build-id check, is this fine now?

rppt: @AnorexicAtticusFinch my point was that from the very beginning you could call the function store_validation_data() and use it for size validation, and then add a function that performs  build-id validation and call it from `store_validation_data()` in the next commit.
Also, I don't think that `_for_restore()` suffix has any useful meaning and IMO can be simply dropped.

AnorexicAtticusFinch: > @AnorexicAtticusFinch my point was that from the very beginning you could call the function store_validation_data() and use it for size validation, and then add a function that performs build-id validation and call it from `store_validation_data()` in the next commit.
Yeah that's what I've done, I'm calling the build-id functions from store_validation_data and validate_data (store_validation_data_build_id and validate_with_build_id respectively).
> Also, I don't think that `_for_restore()` suffix has any useful meaning and IMO can be simply dropped.
Yeah that makes sense, I'll change the names in the next commit.

criu/files-reg.c
```diff
@@ -1380,11 +1387,148 @@ static bool should_check_size(int flags)
...
+	if (!rfe->build_id) {
+		pr_warn("Build-ID (For validation) could not be set for file %s\n",
+				rfe->name);
+		return false;
+	}
+
+	rfe->n_build_id = build_id_size;
+	for (i = 0; i < build_id_size; i++) {
+		rfe->build_id[i] = build_id[i];
+	}
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: Build-id is a repeated uint32 in regfile.proto and build_id is an array of uchar, that's why I didn't use memcpy here

criu/config.c
```diff
@@ -420,6 +421,22 @@ static int parse_join_ns(const char *ptr)
 	return 0;
 }
 
+static int parse_file_validation_method(struct cr_options *opts, const char *optarg)
+{
+	if (!strcmp(optarg, "filesize"))
+		opts->file_validation_method = FILE_VALIDATION_FILE_SIZE;
+	else if (!strcmp(optarg, "buildid"))
```
avagin, line 428: do we want to be able to specify filesize and buildid together?

avagin:
Could you add numbers that will show how fast (slow) the build-id check?

criu/config.c
AnorexicAtticusFinch: Regardless of what option the user picks (Build-id, checksum or just file size), the file size check is done first because if the file sizes aren't equal then there's no point in comparing the checksum or the build-id so it would be faster to immediately put out an error.

AnorexicAtticusFinch:
> Could you add numbers that will show how fast (slow) the build-id check?

I used the time command to measure the time taken for 2 tests that I chose at random.
zdtm/transition/shmem:
file size: 3.782s
build-id: 4.153s

zdtm/static/maps04:
file size: 35.317s
build-id: 35.720s

criu/files-reg.c
```diff
@@ -1475,12 +1475,12 @@ static bool store_validation_data_build_id(RegFileEntry *rfe, int lfd)
 	snprintf(buf, sizeof(buf), "/proc/self/fd/%d", lfd);
 	fd = open(buf, O_RDONLY);
 	if (fd < 0) {
-		pr_warn("Build-ID (For validation) could not be obtained for file %s because can't open the file\n",
+		pr_info("Build-ID (For validation) could not be obtained for file %s because can't open the file\n",
```
avagin: I think this must be an error and we need to stop dumping processes in this case.

criu/files-reg.c
```diff
@@ -1475,12 +1475,12 @@ static bool store_validation_data_build_id(RegFileEntry *rfe, int lfd)
...
 				rfe->name);
 		return false;
 	}
 	if (fstat(fd, &st) < 0) {
-		pr_warn("Build-ID (For validation) could not be obtained for file %s because can't fstat the file\n",
+		pr_info("Build-ID (For validation) could not be obtained for file %s because can't fstat the file\n",
```
avagin: the same here

criu/files-reg.c
```diff
@@ -1489,14 +1489,14 @@ static bool store_validation_data_build_id(RegFileEntry *rfe, int lfd)
...
-		pr_warn("Build-ID (For validation) could not be obtained for file %s\n",
+		pr_info("Build-ID (For validation) could not be obtained for file %s\n",
 				rfe->name);
 		return false;
 	}
 
 	rfe->build_id = xmalloc(sizeof(int) * build_id_size);
 	if (!rfe->build_id) {
-		pr_warn("Build-ID (For validation) could not be set for file %s\n",
+		pr_info("Build-ID (For validation) could not be set for file %s\n",
```
avagin: the same here

criu/files-reg.c
```diff
@@ -1403,7 +1403,6 @@ static int get_build_id(const int fd, const struct stat *fd_status,
 	file_header = (Elf_ptr(Ehdr) *) mmap(0, fd_status->st_size,
 						PROT_READ, MAP_PRIVATE, fd, 0);
 	if (file_header == MAP_FAILED) {
-		pr_warn("HERE!\n");
```
avagin: * we have to log an error in this case
* You added this code in this PR, so you need to merge any fixes of this code in a commit where this code has been introduced. git commit --interactive and git rebase --interactive can help you to do this. We want to have a clean history in the upstream repo.

criu/files-reg.c
```diff
@@ -1475,29 +1483,21 @@ static bool store_validation_data_build_id(RegFileEntry *rfe, int lfd)
 	snprintf(buf, sizeof(buf), "/proc/self/fd/%d", lfd);
 	fd = open(buf, O_RDONLY);
 	if (fd < 0) {
-		pr_info("Build-ID (For validation) could not be obtained for file %s because can't open the file\n",
```
avagin: pr_perror or you need to explain why open can fail here

criu/files-reg.c
```diff
@@ -1574,6 +1574,8 @@ static int get_build_id(const int fd, const struct stat *fd_status,
 		return -1;
 	start_addr = (char *) mmap(0, fd_status->st_size,
 					PROT_READ, MAP_PRIVATE, fd, 0);
+	if (start_addr == MAP_FAILED)
```
avagin, line 1577: This has to be merged in a commit where you introduce this code.

criu/files-reg.c
```diff
@@ -1574,6 +1574,8 @@ static int get_build_id(const int fd, const struct stat *fd_status,
 		return -1;
 	start_addr = (char *) mmap(0, fd_status->st_size,
 					PROT_READ, MAP_PRIVATE, fd, 0);
```
avagin, line 1576: If this is a file mapping, pls add MAP_FILE

Snorch:
@AnorexicAtticusFinch Please take a look at instruction on how to do PRs https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#pull-request-guidelines

Andrew mentioned that fixup commits are not allowed, you will find more information about it in the link.

criu/files-reg.c
```diff
@@ -1380,11 +1387,148 @@ static bool should_check_size(int flags)
...
+	}
+
+	*build_id = (unsigned char *) xmalloc(note_header->n_descsz);
+	if (!*build_id) {
+		munmap(file_header, fd_status->st_size);
+		return -1;
+	}
+
+	memcpy(*build_id,
+		(void *) ((size_t) note_header + sizeof(Elf_ptr(Nhdr)) + note_header->n_namesz),
```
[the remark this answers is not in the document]
avagin: And I think we need to verify that we will not fault here. Can we trust that note_header contain valid data? What if n_namesz is much bigger that the file mapping?

criu/files-reg.c
```diff
@@ -1380,12 +1387,276 @@ static bool should_check_size(int flags)
...
+	int build_id_size, i;
+	char buf[32];
+	int fd;
+	struct stat st;
+
+	snprintf(buf, sizeof(buf), "/proc/self/fd/%d", lfd);
+	fd = open(buf, O_RDONLY);
+	if (fd < 0)
+		return false;
+	if (fstat(fd, &st) < 0) {
```
avagin, line 1610: fd_param contains stat data for this fd so we probably can avoid this fstat here.

criu/files-reg.c
```diff
@@ -1726,6 +2000,77 @@ static int rfi_remap(struct reg_file_info *rfi, int *level)
 	return 0;
 }
 
+/*
+ * Compares the file's build-id with the stored value.
+ * https://fedoraproject.org/wiki/Releases/FeatureBuildId#Unique_build_ID
```
avagin, line 2005: * I think we need a more detailed comment here. The link can be dead in a month.
* I see that this function can return -1, 0 or 1. Pls describe what each exit code means.

criu/files-reg.c
```diff
@@ -1726,6 +2000,77 @@ static int rfi_remap(struct reg_file_info *rfi, int *level)
...
+	}
+
+	if (opts.file_validation_method == FILE_VALIDATION_BUILD_ID)
+		result = validate_with_build_id(fd, fd_status, rfi);
+
+
+	if (result == -1)
+		pr_info("File %s could only be validated with file size\n",
+				rfi->path);
+	if (result)
```
avagin, line 2069: return !!result

criu/files-reg.c
AnorexicAtticusFinch: Why though?
For example, in the ZDTM test "static/remap_dead_pid", the file "/proc/6/mountinfo" could not be opened so only file size was used for validation and the test passed.
If I had put out an error and stopped dumping then the test would've failed.

Log File (From an older commit): https://api.travis-ci.org/v3/job/696469323/log.txt

criu/files-reg.c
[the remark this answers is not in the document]
AnorexicAtticusFinch: This was taken care of in the subsequent commits.

AnorexicAtticusFinch:
@avagin @Snorch I've reordered and squashed some commits, is it fine now or should I squash them all into 1 commit?

images/regfile.proto
```diff
@@ -4,13 +4,17 @@ import "opts.proto";
...
+	optional uint32 	checksum 		= 12;
+	optional uint32 	checksum_config 	= 13;
+	optional uint32 	checksum_parameter 	= 14;
```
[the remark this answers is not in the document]
rppt: @AnorexicAtticusFinch please address this comment as well.

rppt:
> @avagin @Snorch I've reordered and squashed some commits, is it fine now or should I squash them all into 1 commit?

I think 3 commits are fine.
Could you please open a clean pull request for the new version of the patches as described in [CONTRIBUTING.md#pull-request-guidelines](CONTRIBUTING.md#pull-request-guidelines)

images/regfile.proto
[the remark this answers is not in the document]
AnorexicAtticusFinch: Simply creating a new message type and including that in reg_file_entry isn't working and causes a crash while dumping.
The crash is occurring after dump_one_reg_file is called on all the files so my guess is that it is crashing when the data is being serialized? Am I missing something?

criu/files-reg.c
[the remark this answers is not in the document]
avagin: These subsequent commits have to be merged into this one.

images/regfile.proto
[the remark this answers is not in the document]
avagin: Show me the code which crashes and I will try to figure out a reason.

criu/files-reg.c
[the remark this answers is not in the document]
AnorexicAtticusFinch: Yeah that's what I've done, right now there are only 3 commits - adding fields to regfile.proto, adding build-id functionality and adding the cli option.

criu/files-reg.c
avagin: Your example is a special case and I think we don't need to call store_validation_data... in this case.

images/regfile.proto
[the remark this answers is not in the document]
rppt: Did you explicitly allocate the new sub-message?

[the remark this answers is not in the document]
AnorexicAtticusFinch: > Show me the code which crashes and I will try to figure out a reason.

I pushed my test code here: https://github.com/AnorexicAtticusFinch/criu/tree/testing-vdata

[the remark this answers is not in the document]
AnorexicAtticusFinch: > Did you explicitly allocate the new sub-message?

Yep, I did that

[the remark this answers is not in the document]
avagin: > I pushed my test code here: https://github.com/AnorexicAtticusFinch/criu/tree/testing-vdata

I can't reproduce the problem. It works for me.

Issue #1061, opened 2020-05-12
Replace the existing proxy/cache servers on twosigma/criu-image-streamer


avagin:
@nviennot implemented a new image proxy which works faster and less buggy than the exiting one:

```
I implemented an external proxy to do high performance checkpoint/restore. It's called
criu-image-streamer.

From what I can tell, the current proxy/cache server most likely harbor more bugs. Once
I open-source criu-image-streamer (within a week or two), my plan is to propose that it replaces
the existing proxy/cache servers.@nviennot implemented a new proxy/cache server and here
is an idea to replace the existing one 

```

The start of this discussion is in
https://github.com/checkpoint-restore/criu/pull/1051#issuecomment-623511160

Cc: @adrianreber @xemul

avagin:
https://github.com/twosigma/criu-image-streamer

nviennot:
I am not clear on how I should add the streamer to the criu project:
1) I can add it to the main criu repo in a subfolder.
2) I can create a new repo at https://github.com/checkpoint-restore/criu-image-streamer

If we go with 1) I could add a Makefile feature flag for compiling the streamer. But then should it be on or off by default? @adrianreber had well founded reservations due to packaging concerns.

In both cases, I will take out the existing proxy/cache servers and update documentation

adrianreber:
From the packaging standpoint I would prefer a separate repository. As long as CRIU and criu-image-server do not require each other during build time I think it makes sense. With good documentation in the wiki and in the man-pages it should be possible for interested users to find criu-image-server even if it is in another repository.

avagin:
@nviennot I created https://github.com/checkpoint-restore/criu-image-streamer. You have the write access to this repo.

nviennot:
Thank you

I'm waiting for approval for transfering the ownership of the reposutiry from our org to yours. Stay tunned.

PR #1087, opened 2020-06-01
img-remote: introduction of criu-image-streamer

nviennot:
Following up on #1061
The proxy/cache servers are to be replaced with https://github.com/checkpoint-restore/criu-image-streamer

nits: took out some inline keywords that the compile will inline anyways.


adrianreber:
The travis alpine errors should be fixed with: #1089

images/remote-image.proto
```diff
@@ -7,10 +7,7 @@ message local_image_entry {
 }
 
 message remote_image_entry {
-	required string name		= 1;
-	required string snapshot_id	= 2;
-	required uint32 open_mode	= 3;
-	required uint64 size		= 4;
+	/* No longer used */
 }
```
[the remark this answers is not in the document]
nviennot: Yes.

I am also tempted to do the following:
1) Remove the snapshot_id logic. This was originally used for incremental checkpoints support. However, the ids are not meaningful. Also, there should be an efficient mechanism to inform the presence of pages that have not changed. This mechanism should also be thought out with filesystem dedup that would be done by the caller. 
2) Rename the "remote" image naming to "streaming" image. The remote name can be confusing with the lazy page server, which is also remote. It would also be consistent with the naming of criu-image-streamer tool.
3) Remove the code paths where `opts.remote` is tested to avoid using `sendfile()` in `criu/util.c` and `criu/files-reg.c`. There's no reason not to use read+write loops when `sendfile()` should be just fine.

what do you think?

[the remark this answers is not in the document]
avagin: SGTM. We need to run tests. I think zdtm.py has the --remote option for this. Could you fix it to use criu-image-streamer?

nviennot:
I finished the implementation.
I'm missing tests, but I'd be happy to receive some feedback.
Once this is merged, https://github.com/checkpoint-restore/criu-image-streamer/pull/3 can be merged too.
Thank you

scripts/travis/travis-tests
```diff
@@ -159,6 +159,9 @@ LAZY_OPTS="-p 2 -T $LAZY_TESTS $LAZY_EXCLUDE $LAZY_FLAVORS $ZDTM_OPTS"
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
 
+# criu-image-streamer test
+./test/zdtm.py run --stream -p 2 -T '.*(ghost|fifo|unlink).*' $ZDTM_OPTS
```
avagin, line 163: Can we move this in a separate travis job? travis-tests is too big already.

nviennot:
I think it's ready!
Refer to the commit messages to see what has been done.
Let me know what you guys think.

Thanks,
Nico

scripts/install-criu-image-streamer.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -eux
+
+# Install Rust toolchain
+curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+
+# Clone criu-image-streamer in a sibling directory of the criu project directory
+cd $(dirname "$0")/../../
+# TODO change dev branch to master once PR is merged
+git clone https://github.com/checkpoint-restore/criu-image-streamer.git -b dev
```
avagin, line 10: With --depth 1, this will be faster

nviennot: Added, thank you

criu/pagemap.c
```diff
@@ -636,24 +631,12 @@ static int try_open_parent(int dfd, unsigned long id, struct page_read *pr, int
...
-		pfd = dfd == get_service_fd(IMG_FD_OFF) ?
-			get_curr_snapshot_id_idx() - 1 : dfd - 1;
-		if (pfd < 0)
-			goto out;
-	} else {
-		pfd = openat(dfd, CR_PARENT_LINK, O_RDONLY);
-		if (pfd < 0 && errno == ENOENT)
-			goto out;
-	}
+	if (opts.stream)
```
rst0git, line 634: It may useful to add a comment here why parent images are not used when using criu in stream mode.

test/zdtm.py
```diff
@@ -28,7 +28,15 @@
 
 import yaml
 
-os.chdir(os.path.dirname(os.path.abspath(__file__)))
+SCRIPT_DIR = os.path.dirname(os.path.realpath(__file__))
+os.chdir(SCRIPT_DIR)
+
+# Instead of passing the location of the image streamer binary as an argument
+# (e.g., --criu-image-streamer-bin), the caller can modify PATH to include the
+# binary location. We add a default location as well.
```
rst0git, line 36: What do you think about adding `make install` to criu-image-streamer?

criu/pagemap.c
nviennot: Yes, thank you

test/zdtm.py
nviennot: I merged your PR on criu-image-streamer.

I don't necessarily see a reason to change how:
1) Travis compiles criu-image-streamer (and doesn't install it)
2) CRIU locates the streamer binary in ../criu-image-streamer in addition to regular PATH
Maybe you do though. What do you think?

rst0git: No, it just makes it easier to run the tests locally when `criu-image-streamer` is installed.

test/zdtm.py
```diff
@@ -28,7 +28,15 @@
...
+IMG_STREAMER_PATH = os.path.realpath(os.path.join(SCRIPT_DIR, "../../criu-image-streamer"))
```
avagin, line 37: can we specify a path to criu-image-streamer as an argument? For example, it can be --image-streamer /x/y/criu-image-streamer

criu/crtools.c
```diff
@@ -368,9 +366,7 @@ int main(int argc, char *argv[], char *envp[])
 "                            veth[IFNAME]:OUTNAME{@BRIDGE}\n"
 "                            macvlan[IFNAME]:OUTNAME\n"
 "                            mnt[COOKIE]:ROOT\n"
-"\n"
-"  --remote              dump/restore images directly to/from remote node using\n"
-"                        image-proxy/image-cache\n"
+"  --stream              dump/restore images using criu-image-streamer\n"
```
rst0git, line 369: Could you add a description for `--stream` in `Documentation/criu.txt`?

test/zdtm.py
nviennot: We can. Would it be acceptable to just prepend the specified directory in PATH?

criu/crtools.c
nviennot: Yes, I forgot that one

Done

test/zdtm.py
avagin: Sorry, I was out for a while.

> Would it be acceptable to just prepend the specified directory in PATH?

Yes, it would.

avagin:
It looks like this pr doesn't reuse the existing proxy/remote cache code. If I am not mistaken, it is better to rebase this PR on the master branch, and then we will drop existing proxy/cache patches from the criu-dev branch and merge this PR. @nviennot what do you think?

adrianreber:
I was thinking the same as @avagin. This would make handling `master` and `criu-dev` so much easier.

nviennot:
Yes, I made a new PR: #1126

PR #1089, opened 2020-06-03
travis: fix alpine builds

adrianreber:
With the latest version of the alpine container image it seems that alpine changed a few package names. This adapts the alpine container to solve the travis failures.

This also includes a small change to zdtm.py because the latest run of flake8 complains about having variables named 'l'.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -2077,11 +2077,11 @@ def grep_errors(fname):
     print_next = False
     before = []
     with open(fname, errors='replace') as fd:
-        for l in fd:
-            before.append(l)
+        for line in fd:
+            before.append(line)
             if len(before) > 5:
                 before.pop(0)
-            if "Error" in l or "Warn" in l:
+            if "Error" in line or "Warn" in line:
                 if first:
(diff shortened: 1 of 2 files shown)


avagin:
I don't see this commit in the criu-dev branch. Can we cherry-pick proper commits from the criu-dev branch?

test/zdtm.py
```diff
@@ -2077,11 +2077,11 @@ def grep_errors(fname):
     print_next = False
     before = []
     with open(fname, errors='replace') as fd:
-        for l in fd:
-            before.append(l)
+        for line in fd:
```
avagin, line 2080: zdtm.py and Dockerfile.alpint changes should be in separate patches because these changes are independent.

adrianreber:
> I don't see this commit in the criu-dev branch. Can we cherry-pick proper commits from the criu-dev branch?

I meant, that once this is merged on criu-dev it would be nice to have it on master to avoid broken travis runs.

adrianreber:
Spilt into two commits.

PR #1090, opened 2020-06-03
Fix Travis on master branch

adrianreber:
Because the cron triggered travis runs are failing every time because of the expired certificates this bring the fix from `criu-dev` to master.

No need for a new release I would say, just to make Travis happy so that we know if we are seeing a real error instead of assuming it always fails.


adrianreber:
We would probably also need #1089 on the master branch to have error free travis run on master.

PR #1094, opened 2020-06-05
uffd: uffd_open prints info, caller prints error

angieenee:
When uffd_open is called from kerndat_uffd, userfaultfd failure is not
considered an error, so the goal is to suppress the error message --
instead, we print this message as info.

If the function fails, it is the responsibility of the caller
to print the error message.

### DIFF
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -269,7 +269,7 @@ int uffd_open(int flags, unsigned long *features)
 
 	uffd = syscall(SYS_userfaultfd, flags);
 	if (uffd == -1) {
-		pr_perror("Lazy pages are not available");
+		pr_info("Lazy pages are not available: %s\n", strerror(errno));
 		return -errno;
 	}
 


adrianreber:
Hello @angieenee , welcome to CRIU!

Thanks for your patch, but I am not sure it is necessary. It uses `pr_perror` which is defined as:
```
#define pr_perror(fmt, ...)                                             \
        pr_err(fmt ": %s\n", ##__VA_ARGS__, strerror(errno))
```

Which seems already to do exactly what you want.

Also, your commit message needs to be much more detailed. All the information you added to the pull request should have been part of your commit message. That would have been a very good commit messages. We also required commits to carry your 'Signed-off-by'.

angieenee:
> Hello @angieenee , welcome to CRIU!
> 
> Thanks for your patch, but I am not sure it is necessary. It uses `pr_perror` which is defined as:
> 
> ```
> #define pr_perror(fmt, ...)                                             \
>         pr_err(fmt ": %s\n", ##__VA_ARGS__, strerror(errno))
> ```
> 
> Which seems already to do exactly what you want.
> 
> Also, your commit message needs to be much more detailed. All the information you added to the pull request should have been part of your commit message. That would have been a very good commit messages. We also required commits to carry your 'Signed-off-by'.

Ah I see! Thank you so much for your feedback and for bearing with my first patch -- I'll be sure to make necessary changes in the future.

adrianreber:
> Ah I see! Thank you so much for your feedback and for bearing with my first patch -- I'll be sure to make necessary changes in the future.

No problem. Don't worry. Everybody had to send their first patch at some point. Happy to have you contributing. Happy to review your patches.

If you still feel changes are needed here, please rework your patch, if the current code works for you, please close the PR.

criu/uffd.c
```diff
@@ -269,7 +269,7 @@ int uffd_open(int flags, unsigned long *features)
 
 	uffd = syscall(SYS_userfaultfd, flags);
 	if (uffd == -1) {
-		pr_perror("Lazy pages are not available");
+		pr_info("Lazy pages are not available: %s\n", strerror(errno));
```
avagin, line 272: I think the idea of this patch is to suppress the error message when uffd_open is called from kerndat_uffd, because in this case it isn't an error.

avagin:
> Also, your commit message needs to be much more detailed. All the information you added to the pull request should have been part of your commit message. That would have been a very good commit messages. We also required commits to carry your 'Signed-off-by'.

I completely agree with this. We are following the Linux Kernel coding style, so I suggest to read this article: https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#describe-your-changes

angieenee:
Hi @avagin -- yes, I believe this patch's resolution is to display as info instead of error. I can make necessary adjustments to follow the Linux Kernel coding style if the maintainers believe this patch is beneficial.

avagin:
@angieenee I think this patch requires some work. When uffd_open is called from kerndat_uffd, this message has to be printed as warning or info, but on other cases it has to be printed as error.

criu/uffd.c
```diff
@@ -269,7 +269,11 @@ int uffd_open(int flags, unsigned long *features)
 
 	uffd = syscall(SYS_userfaultfd, flags);
 	if (uffd == -1) {
-		pr_info("Lazy pages are not available: %s\n", strerror(errno));
```
rppt: Hello @angieenee, please rearrange your branch so that you have a single commit with a descriptive change log on top of criu-dev.
You can either create a new branch and add the change there or use `git rebase --interactive` to modify the history of your branch

criu/uffd.c
```diff
@@ -269,7 +269,11 @@ int uffd_open(int flags, unsigned long *features)
...
+		if (flags == 0) {
```
rppt, line 272: I wouldn't rely on `flags` to select the logging level. Consider adding an explicit parameter to `uffd_open` and then use this parameter.
Another possibility is to drop the `pr_error()` if `syscal(SYS_userfaultfd)` fails and update the error message in `setup_uffd()`:
```
diff --git a/criu/uffd.c b/criu/uffd.c
index 33b34ba25..b0baff08a 100644
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -315,7 +315,7 @@ int setup_uffd(int pid, struct task_restore_args *task_args)
         */
        task_args->uffd = uffd_open(O_CLOEXEC | O_NONBLOCK, &features);
        if (task_args->uffd < 0) {
-               pr_perror("Unable to open an userfaultfd descriptor");
+               pr_perror("Lazy pages are not available");
                return -1;
        }

File: criu/uffd.c [hunk 4]
```diff
@@ -269,7 +269,11 @@ int uffd_open(int flags, unsigned long *features)
...
+			pr_info("Lazy pages are not available: %s\n", strerror(errno));
+			return -errno
```
rppt, line 274: Please compile your changes before creating a pull request

mclapinski:
This is not ready for review yet. Let's close this PR, I'll work with @angieenee to clean this up and we'll reopen it when it's ready. Thanks everybody for help.

angieenee:
Hello! After further deliberation, I've changed the error to info in uffd_open and left the rest of the error messages as they were. This is because the callers of this function already print error messages where needed. 

Please feel free to take a look and add any questions or concerns!

Issue #1037, opened 2020-04-23
ZDTM VDSO test fails with vdso=0


adrianreber:
This is a tracking bug for the disabled ZDTM VDSO test in #945

If running with `vdso=0` zdtm should detect it and not run tests requiring VDSO support.

0x7f454c46:
I'll put it in my todo.. Though, I don't mind if you beat me at fixing it :-)

PR #1106, opened 2020-06-16
test: skip vdso test on non-vdso systems

adrianreber:
Fixes: #1037

### DIFF
--- a/test/zdtm/static/vdso01.checkskip
+++ b/test/zdtm/static/vdso01.checkskip
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+grep -q "\[vdso\]" /proc/self/maps
--- a/scripts/travis/vagrant.sh
+++ b/scripts/travis/vagrant.sh
@@ -45,9 +45,8 @@ fedora-no-vdso() {
 	vagrant reload
 	ssh default cat /proc/cmdline
 	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; make -j 4'
-	# Excluding the VDSO test as we are running without VDSO
 	# Excluding two cgroup tests which seem to fail because of cgroup2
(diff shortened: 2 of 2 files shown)


adrianreber:
CC: @0x7f454c46

0x7f454c46:
Alternatively, it could be this:
```
diff --git a/test/zdtm/static/vdso01.c b/test/zdtm/static/vdso01.c
index 269688c5e00b..5e5181d0503e 100644
--- a/test/zdtm/static/vdso01.c
+++ b/test/zdtm/static/vdso01.c
@@ -313,6 +313,9 @@ static int vdso_fill_self_symtable(struct vdso_symtable *s)
                break;
        }
 
+       if (s->vma_start == VDSO_BAD_ADDR)
+               skip("Kernel doesn't provide vDSO");
+
        test_msg("[vdso] %lx-%lx\n", s->vma_start, s->vma_end);
 err:
        fclose(maps);
```
But I don't really mind - either way seems good to me.

test/zdtm/static/vdso01.checkskip
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+grep "\[vdso\]" /proc/self/maps -q
```
avagin, line 3: nit: grep -q "\[vdso\]" /proc/self/maps 
￼

0x7f454c46: err, I think backslashes there are important :)

avagin: You are right and I suggest moving -q before non-option arguments. You have not seen backslashes in my comment because I didn't take into account that this is markdown.

adrianreber: I moved the `-q` and force pushed.

avagin:
Applied, thanks!

PR #1107, opened 2020-06-17
travis: Cache Vagrant VM images

rst0git:
This commit enables caching for Vagrant VM images in Travis-CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,10 @@
 language: c
 os: linux
 dist: bionic
-cache: ccache
+cache:
+  ccache: true
+  directories:
+  - /home/travis/.vagrant.d/boxes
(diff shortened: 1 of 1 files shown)


adrianreber:
Any specific reason for this? I guess the idea is to avoid re-downloading the vagrant image between builds, but it seems Travis says to not cache things like this: https://docs.travis-ci.com/user/caching/#things-not-to-cache

They are talking about docker images and debian packages, because it does not really help for large files. They still have to be downloaded.

My question is if you have seen an actual improvement by adding the vagrant images to the cache.

rst0git:
@adrianreber You are right, there isn't actual improvement because large files take as long to download from the cache as from the original source.

PR #1108, opened 2020-06-17
Add error logging to config + crtools init

angieenee:
CRIU sometimes returns 1 from main() with no explanation. Even
increasing the log level is insufficient because that was not set until
after some initialization. Changes made add more logging in the case of
initialization errors in config and crtools.

### DIFF
--- a/criu/config.c
+++ b/criu/config.c
@@ -587,8 +587,10 @@ int parse_options(int argc, char **argv, bool *usage_error,
 			opts.final_state = TASK_ALIVE;
 			break;
 		case 'x':
-			if (optarg && unix_sk_ids_parse(optarg) < 0)
+			if (optarg && unix_sk_ids_parse(optarg) < 0) {
+				pr_err("Failed to parse unix socket inode from optarg: %s\n", optarg);
 				return 1;
+			}
 			opts.ext_unix_sk = true;
 			break;
 		case 't':
@@ -648,13 +650,17 @@ int parse_options(int argc, char **argv, bool *usage_error,
 					goto bad_arg;
 
 				*aux = '\0';
-				if (veth_pair_add(optarg, aux + 1))
+				if (veth_pair_add(optarg, aux + 1)) {
+					pr_err("Failed to add veth pair: %s, %s.\n", optarg, aux + 1);
 					return 1;
+				}
 			}
 			break;
 		case 1049:
-			if (add_script(optarg))
+			if (add_script(optarg)) {
+				pr_err("Failed to add action-script: %s.\n", optarg);
 				return 1;
+			}
 			break;
 		case 1051:
 			SET_CHAR_OPTS(addr, optarg);
@@ -724,12 +730,16 @@ int parse_options(int argc, char **argv, bool *usage_error,
 				return 0;
 			break;
 		case 1064:
-			if (!add_skip_mount(optarg))
+			if (!add_skip_mount(optarg)) {
+				pr_err("Failed to add skip-mnt: %s\n", optarg);
 				return 1;
+			}
 			break;
 		case 1065:
-			if (!add_fsname_auto(optarg))
+			if (!add_fsname_auto(optarg)) {
+				pr_err("Failed while parsing --enable-fs option: %s", optarg);
 				return 1;
+			}
 			break;
 		case 1068:
 			SET_CHAR_OPTS(freeze_cgroup, optarg);
@@ -738,8 +748,10 @@ int parse_options(int argc, char **argv, bool *usage_error,
 			opts.ghost_limit = parse_size(optarg);
 			break;
 		case 1070:
-			if (irmap_scan_path_add(optarg))
+			if (irmap_scan_path_add(optarg)) {
+				pr_err("Failed while parsing --irmap-scan-path option: %s", optarg);
 				return -1;
+			}
 			break;
 		case 1071:
 			SET_CHAR_OPTS(lsm_profile, optarg);
@@ -765,13 +777,17 @@ int parse_options(int argc, char **argv, bool *usage_error,
 					goto bad_arg;
 
 				*aux = '\0';
-				if (ext_mount_add(optarg, aux + 1))
+				if (ext_mount_add(optarg, aux + 1)) {
+					pr_err("Could not add external mount when initializing config: %s, %s\n", optarg, aux + 1);
 					return 1;
+				}
 			}
 			break;
 		case 1073:
-			if (add_external(optarg))
+			if (add_external(optarg)) {
+				pr_err("Could not add external resource when initializing config: %s\n", optarg);
 				return 1;
+			}
 			break;
 		case 1074:
 			if (!strcmp("net", optarg))
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -69,8 +69,10 @@ int main(int argc, char *argv[], char *envp[])
 	BUG_ON(get_service_fd(SERVICE_FD_MIN+1) <
 	       get_service_fd(SERVICE_FD_MAX-1));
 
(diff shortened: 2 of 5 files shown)


criu/config.c
```diff
@@ -587,8 +587,10 @@ int parse_options(int argc, char **argv, bool *usage_error,
 			opts.final_state = TASK_ALIVE;
 			break;
 		case 'x':
-			if (optarg && unix_sk_ids_parse(optarg) < 0)
+			if (optarg && unix_sk_ids_parse(optarg) < 0) {
+				pr_err("unix_sk_ids_parse failed.\n");
 				return 1;
```
rppt, line 592: `unix_sk_ids_parse()` has informative error messages on the error paths. I don't think that adding "unix_sk_ids_parse failed" will improve user's understanding of what happened. Moreover, it may only confuse them ;-)

criu/config.c
```diff
@@ -648,13 +650,17 @@ int parse_options(int argc, char **argv, bool *usage_error,
 					goto bad_arg;
 
 				*aux = '\0';
-				if (veth_pair_add(optarg, aux + 1))
+				if (veth_pair_add(optarg, aux + 1)) {
+					pr_err("veth_pair_add failed.\n");
 					return 1;
```
rppt, line 655: `veth_pair_add()` will report an allocation error or the error occurred during further processing. I didn't check all the paths, but those I've checked have error reporting.

criu/config.c
```diff
@@ -648,13 +650,17 @@ int parse_options(int argc, char **argv, bool *usage_error,
...
+				}
 			}
 			break;
 		case 1049:
-			if (add_script(optarg))
+			if (add_script(optarg)) {
+				pr_err("add_script failed.\n");
```
rppt, line 661: `add_script()` may fail only because of memory allocation and all the `x*alloc()` invocations have error reporting.

mclapinski:
@rppt 
Are you sure we should only report errors on one level? I don't think I agree with that approach. What if some function is called in multiple places? I think it's better for the user to see what exactly failed and where it was called from in that case. If pr_err displayed stacktrace I might have agreed with you, but it doesn't. When I debug something I prefer to see errors on each level.

rppt:
@mclapinski 
> Are you sure we should only report errors on one level? I don't think I agree with that approach. What if some function is called in multiple places? 

But that's not the case here, at several places at least. Most of the added reports are for functions that are only called once.

> I think it's better for the user to see what exactly failed and where it was called from in that case. If pr_err displayed stacktrace I might have agreed with you, but it doesn't. When I debug something I prefer to see errors on each level.

I'm not saying we should not report errors at the top level by any means. But I still think that we do not want simply add
 
```
if (foo()) {
	pr_err("foo failed\n");
	return -1;
}
```

everywhere.

This is either debug aid and then it should be pr_debug message or it is a message for a user and then the text should be more informative.

And, in cases when a function is called only once and it already has error report, I think another `pr_err()` is excessive and only adds unnecessary churn.

mclapinski:
@rppt 
I think the
```
if (foo()) {
	pr_err("foo failed\n");
	return -1;
}
```
is better than nothing which we often have right now. @angieenee can go through all of the functions and check if they all log on error but 1) she can miss something, 2) that's harder to maintain. People can add new returns without logging and then the user has no idea why something failed. With @angieenee's approach, users will always have at least some idea what failed.

The pr_debug approach has another issue - until CRIU parses the command line setting that changes the log level we won't see debug messages since the default log level is `warn`.

rppt:
@mclapinski 

> I think the
> 
> ```
> if (foo()) {
> 	pr_err("foo failed\n");
> 	return -1;
> }
> ```
> 
> is better than nothing which we often have right now. @angieenee can go through all of the functions and check if they all log on error but 1) she can miss something, 2) that's harder to maintain. People can add new returns without logging and then the user has no idea why something failed. With @angieenee's approach, users will always have at least some idea what failed.

I can agree with your point to have an log message for each error. And if it was an automated replacement for **all** cases that silently return error, I would accept the "foo failed" error messages.
But I think that for a dozen calls it would be better to use meaningful error messages.

mclapinski:
@rppt 
I agree that meaningful messages would be better but okay messages are better than no messages. We had issues with CRIU failing without any useful logs so we added some logs. We don't actually understand all of those functions and so writing meaningful messages would require substantial research.

That being said - I understand if you really don't like those messages. In that case I think it would be easier for @angieenee to just go through all of the functions and add logging there on the low level errors (which I guess is closer to your original idea).

mclapinski:
@rppt
Could you state your final opinion?

rppt:
@mclapinski sorry for the delay.
I think many of these functions can fail only if memory allocation fails, these cases should be reflected in the error messages. 
As for the more complex cases it would be nice to see the high level effect of the error. For instance for  'veth_pair_add()`  it could be
```
pr_err("Failed to add  veth pair: %s %s\n", optarg, aux + 1);
```

criu/config.c
angieenee: Thanks for your patience and suggestions, @rppt 
I've updated the error messages to be more user-friendly. I'm open to feedback if the updated messaging is still not quite up-to-standard :)

angieenee:
Thanks for the patience + contributing feedback to these logging changes, everyone! I'm definitely learning a lot by reading through your suggestions and CRIU code. I've updated the error messaging to be more user-friendly and descriptive. Please give these changes another look and feel free to provide additional feedback -- even if it's the commit message or the way I committed. 

Thanks again!

criu/crtools.c
```diff
@@ -192,7 +194,7 @@ int main(int argc, char *argv[], char *envp[])
 		return 1;
 
 	if (kerndat_init()) {
-		pr_err("kerndat_init failed.\n");
+		pr_err("Could not initialize kerndat.\n");
```
rppt, line 197: Maybe "Could not initialize kernel features detection"

criu/config.c
```diff
@@ -778,14 +778,14 @@ int parse_options(int argc, char **argv, bool *usage_error,
...
-					pr_err("ext_mount_add failed.\n");
+					pr_err("Could not add external mount when initializing config: %s, %s\n", optarg, aux + 1);
 					return 1;
 				}
 			}
 			break;
 		case 1073:
 			if (add_external(optarg)) {
-				pr_err("add_external failed.\n");
+				pr_err("Could not add external when initializing config: %s\n", optarg);
```
rppt, line 788: I think 
"Could not add external **resource** ..." would be better

criu/crtools.c
```diff
@@ -137,8 +139,10 @@ int main(int argc, char *argv[], char *envp[])
 		}
 
 		opts.exec_cmd = xmalloc((argc - optind) * sizeof(char *));
-		if (!opts.exec_cmd)
+		if (!opts.exec_cmd) {
+			pr_err("Memory allocation failed.\n");
```
mclapinski, line 143: xmalloc already logs, please revert

angieenee:
Hey hi hello! 
Exactly one travis job is failing: 
```
Failed tests: 
  CheckpointRestore.runtest:226->restore:442 SocketsListen: ERROR: Error during restore: exitCode of restore process was not zero.
```
This job seems to [pass](https://travis-ci.org/github/angieenee/criu/jobs/703722202) on my personal repo, so I was wondering if anyone could provide some feedback. Thanks so much!

rppt:
> ```
> Failed tests: 
>   CheckpointRestore.runtest:226->restore:442 SocketsListen: ERROR: Error during restore: exitCode of restore process was not zero.
> ```

I don't think that this CI hiccup is related to your changes.

avagin:
Applied, thanks a lot!

PR #1109, opened 2020-06-18
GSoC 2020: Generate binary logs using flog

prakritigoyal19:
The changes in this pr generate all log files as binary logs, but when the logs are to be printed directly to the terminal, it uses the regular printer.

These changes are failing for three architectures:
1. Centos - Due to its incompatibility with C11
2. Alpine
3. Alpine CLANG=1

I request your feedback on how to improvise these changes and correct these errors.


cyrillos:
In centos it is expected to fail, on alpine the output is really strange
```
=[log]=> dump/zdtm/transition/thread-bomb/44/4/dump.log
b'\x00\x10\x00viou\xcd\xab\xcd\xab`\x00\x00\x00\x01\x00\x00\x00\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd \x00\x00\x00\x00\x00\x00\x00[\x00\x00\x00\x00\x00\x00\x00Dumping memfd:%s contents (id %#x, shmid: %#x, size: %lu)'
b'\x00\x10\x00ame\xcd\xab\xcd\xabP\x00\x00\x00\x01\x00\x00\x00\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd \x00\x00\x00\x00\x00\x00\x00I\x00\x00\x00\x00\x00\x00\x00page-pipe: Create page pipe for %u segs'
b'\x00\x10\x00#x, s\xcd\xab\xcd\xabX\x00\x00\x00\x01\x00\x00\x00\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd \x00\x00\x00\x00\x00\x00\x00P\x00\x00\x00\x00\x00\x00\x00page-pipe: Will grow page pipe (iov off is %u)'
b'\x00\x10\x00e: %lu\xcd\xab\xcd\xab0\x00\x00\x00\x01\x00\x00\x00\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd \x00\x00\x00\x00\x00\x00\x00-\x00\x00\x00\x00\x00\x00\x00No %s image'
b'\x00\x10\x00 \xcd\xab\xcd\xabx\x00\x00\x00\x01\x00\x00\x00\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd\xef\xbf\xbd \x00\x00\x00\x00\x00\x00\x00u\x00\x00\x00\x00\x00\x00\x00Error (%s:%d): page-xfer: No parent image found, though parent directory is set: %s'
```

Looks like the stack is broken somewhere.

prakritigoyal19:
I had assumed this occurs because the zdtm test suite is trying to read from a binary file, and this had happened in all other architectures as well.

If this is the cause of failure for alpine architecture, then I should be able to fix it after building the decoder and using it in places where it greps from the log file.

cyrillos:
> I had assumed this occurs because the zdtm test suite is trying to read from a binary file, and this had happened in all other architectures as well.
> 
> If this is the cause of failure for alpine architecture, then I should be able to fix it after building the decoder and using it in places where it greps from the log file.

The real error comes from python
```
Try -h|--help for more info
Connecting to CRIU in swrk mode.
  File "./config_file.py", line 185, in <module>
    test_rpc_without_configuration_file()
  File "./config_file.py", line 124, in test_rpc_without_configuration_file
    check_results(resp, log_file)
  File "./config_file.py", line 114, in check_results
    search_in_log_file(log, 'The criu itself is within dumped tree')
  File "./config_file.py", line 94, in search_in_log_file
    if message not in f.read():
  File "/usr/lib/python3.8/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 12: invalid start byte
```
So we've been searching something in log file and failed because now it is in binary format. I've no clue what this code is doing. Maybe @avagin knows.

adrianreber:
This is the config file code, which changes the log files via the config file and then it looks at the log file to see if it works. If the log file no longer contains the expected strings it fails. So this test needs to be adapted to first decode the binary log file.

prakritigoyal19:
Alright, I will start working on the decoder as of now and revisit all these issues after that is done.

prakritigoyal19:
The logs are now generated in binary, only when the user specifies it with a -b option. This could be used for testing of this feature until the decoder is built.

xemul:
Obsoleted by pr#1192

Issue #1034, opened 2020-04-21
Second checkpoint using cgroup2 freezer fails


adrianreber:
I am mainly looking for some additional input or ideas why the following is happening.

I am currently trying to switch crun in combination with Podman to use the freezer to stop processes in the container: `criu_set_freeze_cgroup()`.

I am able checkpoint and restore a Java (Wildfly) based container multiple times without a problem without using the freezer cgroup.

If I use the freezer I can checkpoint the container only once. The second checkpoint of a previously restored container fails with:
```
(00.000504) Detected cgroup V2 freezer
(00.000507) freezing processes: 100000 attempts with 100 ms steps
(00.000562) cgroup.freeze=0
(00.000577) cgroup.freeze=1
(00.100902) cgroup.freeze=1
(00.101099) freezing processes: 1 attempts done
(00.101265) SEIZE 15459: success (-> shell script PID 1 in the container)
(00.102551) Seized task 15459, state 1
(00.102590) seccomp: Collected tid_real 15459 mode 0x2
(00.102697) Collected (0 attempts, 0 in_progress)
(00.102827) Seized task 15481, state 0 (-> JVM started by the above shell script)
(00.103098) Error (compel/src/lib/infect.c:235): Unseizable non-zombie 15481 found, state S, err -1/10
(00.103210) Seized task 15481, state 0
(00.103334) Error (compel/src/lib/infect.c:235): Unseizable non-zombie 15481 found, state S, err -1/10
(00.103397) Collected (-1 attempts, 1 in_progress)
(00.103433) Unlock network
(00.103444) Unfreezing tasks into 1
(00.103451) 	Unseizing 15459 into 1
(00.103534) Error (criu/cr-dump.c:1776): Dumping FAILED.
```
Any ideas why using the cgroup2 freezer fails with a process which has been restored in contrast of the process has been started by the container? Any ideas what the compel error means?

adrianreber:
@avagin who could help me with this? Who knows compel the best?

avagin:
I am sorry for the late response. Could you explain how to reproduce this?

avagin:
I was able to reproduce this issue. First of all, I want to mention that zdtm.py --freezecg doesn't work with the cgroup v2. This means that we don't test this in travis, Jenkins, etc. If something isn't tested, it doesn't work or will be broken soon. This case is another proof of the previous sentence. Please, please, please, write tests for all features.

avagin:
[...]

PR #1110, opened 2020-06-19
Add the initial support for cgroupv2

avagin:


test/zdtm.py
```diff
@@ -1645,7 +1676,11 @@ def get_freezer(desc):
         return noop_freezer()
 
     fd = desc.split(':')
-    fr = cg_freezer(path=fd[0], state=fd[1])
+
+    if os.access("/sys/fs/cgroup/user.slice/cgroup.procs", os .F_OK):
```
Snorch, line 1680: Are unified cgroups always mounted just in /sys/fs/cgroup, not /sys/fs/cgroup/unified? (I mean one can do something like https://blog.christophersmart.com/2019/12/15/enabling-docker-in-fedora-31-by-reverting-to-cgroups-v1/)

adrianreber:
This PR fixes for me the problems reported in https://github.com/checkpoint-restore/criu/issues/1034

Using Podman and crun with CRIU and this PR I can checkpoint and restore a container multiple times using the cgroup2 freezer.

Thanks.

test/zdtm.py
adrianreber: Starting with Fedora 31 there is only cgroup2 available and it is mounted at /sys/fs/cgroup. I expect that every systemd based distribution will mount cgroup2 at that location if it switches to cgroup2 only.

avagin:
@adrianreber With this pr, criu creates cgroups and moves tasks in them, but it doesn't restore any cgroup properties (memory.max, cpu.weight, etc). If you want, you can start working on restoring these properties. I don't have a plan to work on this in the near future.

adrianreber:
Do you want to merge this PR or wait for complete cgroup2 support based on this PR? Having this PR merged would be an improvement to the current code.

test/zdtm.py
avagin: @Snorch In your example, cgroupv2 will be mounted in /sys/fs/cgroup/unified, but the freezer controller will be mounted separately in /sys/fs/cgroup/freezer/. I have checked and my changes works fine in this case.

PR #1121, opened 2020-06-25
compel: Replace sh_addr with sh_offset

MaskRay:
criu/pie/parasite.built-in.o is an relocatable output (ld -r)
which have zero section addresses when linked with LLD (LLVM linker).
(GNU ld also uses zero addresses in many cases.)

.text (index 1) has a sh_addr of 0 and sh_offset of sizeof(Elf_Ehdr).
For any subsequent section, sh_addr = sh_offset.

GNU ld (non-zero section addresses but this is not reliable):

```
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .text             PROGBITS        0000000000000000 000040 003f22 00  AX  0   0 16
  [ 2] .rela.text        RELA            0000000000000000 007348 002370 18   I  6   1  8
  [ 3] .data             PROGBITS        0000000000003f30 003f70 000058 00  WA  0   0 16
  [ 4] .rodata           PROGBITS        0000000000003fa0 003fe0 000f56 00   A  0   0 32
```

Use sh_offset to make the result work with LLD and more robust with GNU ld.

-----

This addresses one problem when linked with [LLD](http://lld.llvm.org/). It passes our internal tests when linked with GNU ld. There are other issues I haven't figure out yet.


MaskRay:
There may be some issues. I'm investigating it.

0x7f454c46:
Could you add another commit with a test with lld build to Travis-CI?
We already have clang build jobs, so let's have a job with clang's linker to make sure it won't be accidentally broken in future.

MaskRay:
> Could you add another commit with a test with lld build to Travis-CI?
> We already have clang build jobs, so let's have a job with clang's linker to make sure it won't be accidentally broken in future.

I'd be happy to do that. Unfortunately I noticed that this patch is insufficient. There may be another place where linker differences cause criu to fail.

cyrillos:
Wait. There is no problem of having virtual address equal to zero. The padding is needed to fill gap between sections which are not aligned in size but VMA aligned. I don't get which exactly problem we're solving here?

What happens when we're generating binary blob with llvm-ld? The generated blob is not loaded correctly or what?

MaskRay:
No idea. I am a linker person who knows nearly nothing about criu:/

This steps runs the executable which has compel/src/lib/handle-elf-host.c as one of its source files:
```
./compel/compel-host hgen -f criu/pie/parasite.built-in.o -o criu/pie/parasite-blob.h
```

This does not work for LLD linked `criu/pie/parasite.built-in.o` (sh_addr fields are all zeros):

```
% readelf -WS criu/pie/parasite.built-in.o                                                                                                            
There are 9 section headers, starting at offset 0x9840:                                                                                               
                                                                           
Section Headers:                     
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al                                                              
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0                                                              
  [ 1] .text             PROGBITS        0000000000000000 000040 003f22 00  AX  0   0 16                                                              
  [ 2] .data             PROGBITS        0000000000000000 003f70 000058 00  WA  0   0 16                                                              
  [ 3] .rodata           PROGBITS        0000000000000000 003fe0 000f45 00 WAMS  0   0 32                                                             
  [ 4] .rela.text        RELA            0000000000000000 004f28 002370 18   I  6   1  8                                                              
  [ 5] .rela.rodata      RELA            0000000000000000 007298 000138 18   I  6   3  8                                                              
  [ 6] .symtab           SYMTAB          0000000000000000 0073d0 001998 18      8 132  8                                                              
  [ 7] .shstrtab         STRTAB          0000000000000000 008d68 000047 00      0   0  1                                                              
  [ 8] .strtab           STRTAB          0000000000000000 008daf 000a8c 00      0   0  1
```

I think criu is making use of an unreliable feature in GNU ld (-r output has non-zero sh_addr in some cases). As a case where this setting is unreliable: if the sections `.text1 .text .data` and .text is empty => .text and .data will have zero sh_addr.

cyrillos:
> If you want to try lld by yourself. LDFLAGS += -fuse-ld=lld

OK, I'll try to do so on the week (no promises though, lack of spare time :( Actually the idea in first place was not to depend on VMA since it might be zero. Still you're right that we've been relying on gnu's `ld` heavily.

MaskRay:
Hi Cyrill, do you have luck with LLD? :)

cyrillos:
Oh. I'm really-really sorry, didn't get time :( I hope I'll have a piece of spare this weekend. Don't worry I remember about this issue and wont leave it. Maybe Dima manage to take a look earlier? @0x7f454c46

cyrillos:
To be honest the section's processing in compel code is not _that_ stable. When I've been developing it of course I only tested with gnu's ld.

cyrillos:
True, I see the same when compiling with `clang` and linking with `lld`

Just like
```
CC=clang LD=ld.lld make V=1 -j 8 2>&1 > err
```

And output as
```
Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 0]                   NULL            0000000000000000 000000 000000 00      0   0  0
  [ 1] .text             PROGBITS        0000000000000000 000040 003ed8 00  AX  0   0 16
  [ 2] .data             PROGBITS        0000000000000000 003f18 000050 00  WA  0   0  8
  [ 3] .rodata           PROGBITS        0000000000000000 003f70 000ef5 00 WAMS  0   0 16
  [ 4] .rela.text        RELA            0000000000000000 004e68 002df0 18   I  7   1  8
  [ 5] .rela.rodata      RELA            0000000000000000 007c58 000450 18   I  7   3  8
  [ 6] .llvm_addrsig     LOOS+0xfff4c03  0000000000000000 0080a8 000008 00   E  0   0  1
  [ 7] .symtab           SYMTAB          0000000000000000 0080b0 001938 18      9 129  8
  [ 8] .shstrtab         STRTAB          0000000000000000 0099e8 000055 00      0   0  1
  [ 9] .strtab           STRTAB          0000000000000000 009a3d 000b14 00      0   0  1
```

cyrillos:
Actually we rely heavily on `sh_addr` but we have to do a part of linker work instead and try to emulate "loading" procedure where later we will apply relocations manually.

You propose looks ok to me though I fear that we have to re-walk all elf handling code since there are a number of references to `sh_addr`. I think the GNU `ld` does a part of our job and precalculate virtual addresses in turn ld.lld does not leaving it to loader (I'm guessing, not sure, didn't read any code).

That said Ack to your changes but more places to be checked still and I'm not sure who gonna be doing it (if ever :-)

cyrillos:
IOW, as far as I understand the GNU's ld does align sections and precalculates virtual addresses by self but this is a feature of it, not a standard thus we should change our code and use section's offsets and calculate virtual addresses by self if they are not present in object files.

MaskRay:
Obsoleted by #1227

PR #1122, opened 2020-06-26

angieenee:
When groups already have desired values, we should make a check to skip calling
sys_setgroup. Furthermore, this check allows us to restore without errors when in an
unprivileged user namespace that cannot use setgroups().

### DIFF
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -202,10 +202,32 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
 	 * Setup supplementary group IDs early.
 	 */
 	if (args->groups) {
-		ret = sys_setgroups(ce->n_groups, args->groups);
-		if (ret) {
-			pr_err("Can't setup supplementary group IDs: %d\n", ret);
+		// We may be in an unprivileged user namespace where we're not
+		// allowed to call setgroups. If the current list of groups is already
+		// what we want, skip the call to setgroups.
+
+		int n = sys_getgroups(0, NULL);
+		int len = n * sizeof(unsigned int);
+		bool groups_are_ok;
+		unsigned int* gids = (unsigned int*)sys_mmap(NULL, len, PROT_WRITE | PROT_READ,
+			MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+		if (gids == MAP_FAILED) {
+			pr_err("Unable to mmap %d bytes\n", len);
 			return -1;
+		}
+		n = sys_getgroups(n, gids);
+		groups_are_ok = (n == ce->n_groups && !memcmp(gids, args->groups, len));
+		if (sys_munmap(gids, len)) {
+			pr_err("Unable to munmap\n");
+			return -1;
+		}
+		if (groups_are_ok) {
+			pr_debug("getgroups() looks good, not calling setgroups()\n");
+		} else {
+			if (sys_setgroups(ce->n_groups, args->groups)) {
+				pr_err("Can't setgroups([%zu gids])\n", ce->n_groups);
+				return -1;
+			}
 		}
 	}
 


criu/pie/restorer.c
```diff
@@ -202,11 +202,33 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
 	 * Setup supplementary group IDs early.
 	 */
 	if (args->groups) {
-		ret = sys_setgroups(ce->n_groups, args->groups);
-		if (ret) {
-			pr_err("Can't setup supplementary group IDs: %d\n", ret);
+		// We may be in an unprivileged user namespace where we're not
+		// allowed to call setgroups. If the current list of groups is already
+		// what we want, skip the call to setgroups.
```
rppt, line 207: Please use classic C comments to keep this consistent with the rest of the code

criu/pie/restorer.c
```diff
@@ -202,11 +202,33 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
...
+
+		int n = sys_getgroups(0, NULL);
```
rppt, line 209: `sys_getgroups()` may fail.
Please add a check for its return value. This might require some adjustments to the declaration block as we generally don't mix declarations and code.

criu/pie/restorer.c
```diff
@@ -202,11 +202,33 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
...
+		unsigned int* gids = (unsigned int*)sys_mmap(NULL, len, PROT_WRITE | PROT_READ,
+			MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+		if (gids == MAP_FAILED) {
+			pr_err("Unable to mmap %d bytes\n", len);
+			return -1;
+		}
+		n = sys_getgroups(n, gids);
+		groups_are_ok = (n == ce->n_groups && !memcmp(gids, args->groups, len));
+		if (sys_munmap(gids, len)) {
+			pr_err("Unable to munmap\n");
```
avagin, line 221: The travis jobs fail because sys_nunmap returns EINVAL if len is zero.

criu/pie/restorer.c
```diff
@@ -202,11 +202,33 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
...
 			return -1;
 		}
+		if (groups_are_ok) {
```
avagin, line 224: Can we move this check out of the pie code?

criu/pie/restorer.c
```diff
@@ -202,11 +202,33 @@ static int restore_creds(struct thread_creds_args *args, int procfd,
...
+		int n = sys_getgroups(0, NULL);
+		int len = n * sizeof(unsigned int);
+		bool groups_are_ok;
+		unsigned int* gids = (unsigned int*)sys_mmap(NULL, len, PROT_WRITE | PROT_READ,
+			MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+		if (gids == MAP_FAILED) {
+			pr_err("Unable to mmap %d bytes\n", len);
+			return -1;
+		}
+		n = sys_getgroups(n, gids);
```
avagin, line 218: you don't need to call sys_getgroups if n isn't equal to ce->n_groups

criu/pie/restorer.c
mclapinski: According to man, it cannot fail if `size` is 0. Unless we want to take into account future changes to getgroups.

criu/pie/restorer.c
mclapinski: Do you mean we should call the getgroups and do the comparison before restorer and then pass that information into restorer? I guess the groups can't change in between, right?

criu/pie/restorer.c
angieenee: Good catch! Thank you.

criu/pie/restorer.c
angieenee: I don't think it'd hurt to provide an additional check in the case that getgroups does get modified.

criu/pie/restorer.c
avagin: yep. This is exactly what I mean. And we will not need to call mmap/munamp, we will be able to use xmalloc/xfree.

mclapinski: great idea, thanks

criu/cr-restore.c
```diff
@@ -3128,6 +3128,24 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
...
+	bool ret;
+	gid_t* gids;
+
+	n = getgroups(0, NULL);
+	if (n != n_groups)
+		return false;
+	if (n == 0)
+		return true;
+	len = n * sizeof(gid_t);
+	gids = (gid_t*)xmalloc(len);
```
rppt, line 3142: There is no need to cast `xmalloc` return value.
But, it should be checked for `!NULL`.

criu/cr-restore.c
```diff
@@ -3128,6 +3128,24 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
...
+	n = getgroups(n, gids);
```
rppt, line 3143: `getgroups` may fail, it's return value should be verified.

criu/cr-restore.c
```diff
@@ -3128,6 +3128,24 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
 	thread_args->creds_args = args;
 }
 
+static bool groups_are_okay(gid_t* groups, int n_groups) {
```
rppt, line 3131: I think a better name would be `groups_match()` as it conveys that we compare the supplied groups with existing ones.
Another nit, the opening brace should be on a separate line.

criu/cr-restore.c
```diff
@@ -3128,6 +3128,24 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
...
+		return false;
+	if (n == 0)
+		return true;
+	len = n * sizeof(gid_t);
+	gids = (gid_t*)xmalloc(len);
+	n = getgroups(n, gids);
+	ret = !memcmp(gids, groups, len);
+	xfree(gids);
+	return ret;
+}
```
rppt, line 3147: And in general, don't save on empty lines between code chunks, they make the code more readable ;-)

criu/cr-restore.c
angieenee: We **can** check if `getgroups` fails here, but the Linux man page for `getgroups` states that it will fail either if gids has an invalid address or if n is "less than the number of supplementary group IDs, but is not zero." 
Given this, I'll be sure to follow your other suggestions to ensure that gids is valid -- however, `getgroups` should not fail here after those changes.

criu/cr-restore.c
angieenee: Good catch. Thank you!

criu/cr-restore.c
rppt: It may also fail with `-ENOMEM` or `-EPERM`:

```
ENOMEM Out of memory.
EPERM  The calling process has insufficient privilege (the caller  does not have the CAP_SETGID capability in the user namespace in which it resides).
EPERM (since Linux 3.19) The use of setgroups() is denied in this user namespace.  See the description of /proc/[pid]/setgroups in user_namespaces(7).
```

https://man7.org/linux/man-pages/man2/getgroups.2.html

angieenee: That man page contains the possible failures for both `getgroups` and `setgroups`. You are absolutely correct in that `setgroups` could fail for additional reasons, but I'd defer to the `getgroups` failures for this line of code.

rppt: You are right, I've skipped over the line "setgroups() can additionally fail" :)

Still, checking a system call return value is a good practice even if you know for sure that there should be no error. 
Some code around this call may change and this assumption will not hold anymore, or the kernel implementation will change and it may return error in some case that was not anticipated.

It is better to have the check for the system call return value from the beginning to be on the safe side.

angieenee:
I've restored this PR to the original commit, and I'm closing it as suggested. I'll open a new PR with restructured changes.

PR #1123, opened 2020-06-27
Restrict Checks for Open/MMAPed Files - CLI & Build-ID

AnorexicAtticusFinch:
This adds CLI options and the build-ID file validation method in addition to the current file size check as part of GSoC 2020.

This will make CRIU use the build-ID validation method by default whenever possible (The file size check is now a preliminary check) and if that cannot be done, it will use only the file size check. To explicitly use only the file size check all the time, the following command line option can be used: --file-validation="filesize"
In other words, --file-validation="buildid" will not make a difference as this is the default method.

Using the time command to measure the time taken for 2 tests that were chosen at random.
zdtm/transition/shmem:
file size: 3.782s
build-id: 4.153s

zdtm/static/maps04:
file size: 35.317s
build-id: 35.720s

Mentors: @avagin @0x7f454c46

Link to the previous PR: 
https://github.com/checkpoint-restore/criu/pull/1084

Changes from the previous PR:
rebase on the current criu-dev
cosmetic changes
get_build_id calls appropriate 32-bit or 64-bit functions as necessary
pointer safety checks
other minor changes


criu/files-reg.c
```diff
@@ -1380,12 +1381,290 @@ static bool should_check_size(int flags)
...
+		start_addr[2] != 0x4c || start_addr[3] != 0x46) {
+		munmap(start_addr, fd_status->st_size);
+		return -1;
+	}
+
+	if (start_addr[4] == ELFCLASS64)
+		return get_build_id_64((Elf64_Ehdr *) start_addr, fd_status, build_id);
+	else if (start_addr[4] == ELFCLASS32)
+		return get_build_id_32((Elf32_Ehdr *) start_addr, fd_status, build_id);
+	else
```
avagin, line 1594: nit: I think we can remove this and previous else.

criu/files-reg.c
```diff
@@ -1380,12 +1381,290 @@ static bool should_check_size(int flags)
...
+	if (fd_status->st_size < 5)
+		return -1;
+	start_addr = (char *) mmap(0, fd_status->st_size,
+					PROT_READ, MAP_PRIVATE | MAP_FILE, fd, 0);
+	if (start_addr == MAP_FAILED) {
+		pr_warn("Couldn't mmap file with fd %d", fd);
+		return -1;
+	}
+
+	if (start_addr[0] != 0x7f || start_addr[1] != 0x45 ||
```
avagin, line 1584: pls add a comment which explains these checks.

criu/files-reg.c
```diff
@@ -1380,12 +1381,290 @@ static bool should_check_size(int flags)
...
+	*build_id = (unsigned char *) xmalloc(size);
+	if (!*build_id) {
+		munmap(file_header, fd_status->st_size);
+		return -1;
+	}
+
+	memcpy(*build_id, (void *) note_header, size);
+
+	munmap(file_header, fd_status->st_size);
+	return size;
```
avagin, line 1561: err:
    munmap(file_header, fd_status->st_size);
    return -1;

criu/files-reg.c
```diff
@@ -1380,12 +1381,290 @@ static bool should_check_size(int flags)
...
+	}
+
+	note_header = (Elf64_Nhdr *) (program_header->p_offset + (char *) file_header);
+	if (note_header <= (Elf64_Nhdr *) file_header) {
+		munmap(file_header, fd_status->st_size);
+		return -1;
+	}
+
+	note_header_end = (Elf64_Nhdr *) (file_header_end - sizeof(Elf64_Nhdr));
+	num_iterations = 500;
```
avagin, line 1525: pls, add a comment which explains this magic number.

criu/files-reg.c
```diff
@@ -1380,12 +1381,290 @@ static bool should_check_size(int flags)
...
+ * obtained, else -1.
+ */
+static int get_build_id(const int fd, const struct stat *fd_status,
+				unsigned char **build_id)
+{
+	char *start_addr;
+
+	if (fd_status->st_size < 5)
+		return -1;
+	start_addr = (char *) mmap(0, fd_status->st_size,
```
avagin, line 1577: The file can be very big. Are you sure that we need to map the whole file? Maybe we can avoid this somehow? And If we can't, we need to check the elf magic before mapping a file.

AnorexicAtticusFinch: This shouldn't be a problem since demand paging is used right? mmap is only anticipatory and it will fault and demand the required pages from the file system only when needed.

AnorexicAtticusFinch:
Since I'll have to squash this fix-up commit, I'll close this PR and open a fresh one?

avagin:
> Since I'll have to squash this fix-up commit, I'll close this PR and open a fresh one?

I don't need to create a new PR for this.

criu/files-reg.c
avagin: On x32 systems, the address space size is only 4Gb. CRIU has to work on arm32.

AnorexicAtticusFinch: Instead of mapping the file to check for the magic number, I'll read the first 5 bytes and only if it's an ELF file I'll map it?

AnorexicAtticusFinch:
The cross compile tests failed due to a hash sum mismatch while upgrading packages, is there a way to restart the tests?

criu/files-reg.c
avagin: Even in this case, you need to think about how to handle a case when you can't map a file due to its size.

AnorexicAtticusFinch: One possible solution is that if a file is greater than 3 GB (4GB minus 1 GB for the system reserved address spaces), regardless of which validation method the user has opted for, I read the first 1024 bytes (Some large number of bytes) and use the checksum of these bytes for validation.

Another solution would be to always mmap at most 3 GB (If the file is greater, I'll just pretend it's only 3 GB and try finding the build-ID or checksum).

AnorexicAtticusFinch: I'm not sure which would be a better approach but I think it's the second one.
@0x7f454c46 Is there a better solution?

0x7f454c46: Well, yes, I'd use some "big-enough" chunk to mmap(). Let's say, 1Mb (define it as a constant in one place in case there's need to change it) and when you found where are the section headers, mmap() only them (with this 1 Mb window), then mmap() another 1Mb on top of previous while looking for build-id.

AnorexicAtticusFinch: @0x7f454c46 That seems unnecessary though, since the build-id exists within the first hundreds of bytes (I tested this by printing the location, in bytes, of the build-id relative to the start of each file and ran the ZDTM test suite) so I can just mmap() the first 1 MB and if the build-id doesn't exist within this, it's highly probable that it doesn't exist at all and I will call the checksum method instead. Wouldn't this be better?

0x7f454c46: Well, for computing the checksum you still will need this window (1 Mb or whatever sensable limit).

AnorexicAtticusFinch: @0x7f454c46 Yes but for finding the build-id this is unnecessary since only the first chunk is needed.

criu/files-reg.c
```diff
@@ -1380,12 +1381,299 @@ static bool should_check_size(int flags)
...
+	
+	munmap(start_addr, min_t(size_t, fd_status->st_size, 1048576));
+	return -1;
+}
+
+/*
+ * Finds and stores the build-id of a file, if it exists, so that it can be validated
+ * while restoring.
+ * Returns 1 if the build-id of the file could be stored, 0 if there was an error
+ * or -1 if the build-id could not be obtained.
```
avagin, line 1609: I think it would be more consistent if it will return -1 on an error and 0 if the build-id could not be obtained.

criu/files-reg.c
```diff
@@ -1380,12 +1381,299 @@ static bool should_check_size(int flags)
...
+ */
+static int store_validation_data_build_id(RegFileEntry *rfe, int lfd,
+						const struct fd_parms *p)
+{
+	unsigned char *build_id = NULL;
+	int build_id_size, i;
+	int fd;
+
+	if (p->stat.st_size < 5)
```
avagin, line 1618: pls add a comment what 5 means here

criu/files-reg.c
```diff
@@ -1380,12 +1381,299 @@ static bool should_check_size(int flags)
...
+
+	if (read(fd, buf, 5) != 5)
+		return -1;
+
+	/* 
+	 * The first 4 bytes contain a magic number identifying the file as an
+	 * ELF file. They contain the characters ‘\x7f’, ‘E’, ‘L’, and
+	 * ‘F’, respectively.
+	 */
+	if (buf[0] != 0x7f || buf[1] != 0x45 || buf[2] != 0x4c || buf[3] != 0x46)
```
avagin, line 1581: #include <elf.h>
...
memcpy(buf, ELFMAG, SELFMAG);

criu/files-reg.c
```diff
@@ -1380,12 +1381,299 @@ static bool should_check_size(int flags)
...
+ * and then calling either the 32-bit or the 64-bit function as necessary.
+ * Returns the number of bytes of the build-id if it could be
+ * obtained, else -1.
+ */
+static int get_build_id(const int fd, const struct stat *fd_status,
+				unsigned char **build_id)
+{
+	char *start_addr, buf[5];
+
+	if (read(fd, buf, 5) != 5)
```
avagin, line 1573: read(fd, buf, SELFMAG) != SELFMAG

criu/files-reg.c
avagin: p->stat.st_size < SELFMAG

criu/files-reg.c
AnorexicAtticusFinch: But I will also need the 5th bit (SELFMAG + 1) since I need to know whether the file is 32-bit or 64-bit right?

criu/files-reg.c
AnorexicAtticusFinch: If I do this, I won't be able to simply return !!result in store_validation_data().

AnorexicAtticusFinch: Then I won't be able to simply return !!result in store_validation_data().

avagin:
@AnorexicAtticusFinch I forgot about one major thing. You need to develop tests. And we need to have a test which will fail expectedly. I think it can look something like this https://github.com/avagin/criu/commit/1e8d2f61ca840a0a4fa9808c3d2e0378cc1dd418

AnorexicAtticusFinch:
> @AnorexicAtticusFinch I forgot about one major thing. You need to develop tests. And we need to have a test which will fail expectedly. I think it can look something like this [avagin@d69dfbb](https://github.com/avagin/criu/commit/d69dfbb574733e5f0d03b74bd22b4e7371596)

In the timeline, I had allotted the entire 3rd work period to writing ZDTM tests. I'll write the tests for build-ID and checksum together, once all the work on the build-ID and checksum method are over.

I've finished implementing the latest changes you requested.

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+		goto err;
+
+	program_header_end = (Elf64_Phdr *) (file_header_end - sizeof(Elf64_Phdr));
+	num_iterations = file_header->e_phnum + 1;
+
+	/* 
+	 * If the file has a build-id, it will be in the PT_NOTE program header 
+	 * entry AKA the note sections.
+	 */
+	while (num_iterations-- && program_header <= program_header_end &&
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: This is already accounted for when program_header_end is calculated.

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+	note_header = (Elf64_Nhdr *) (program_header->p_offset + (char *) file_header);
+	if (note_header <= (Elf64_Nhdr *) file_header)
+		goto err;
+
+	note_header_end = (Elf64_Nhdr *) (file_header_end - sizeof(Elf64_Nhdr));
+	/* Arbitrary large number to prevent unnecessarily searching the entire file */
+	num_iterations = 500;
+
+	/* The note type for the build-id is NT_GNU_BUILD_ID. */
+	while (num_iterations-- && note_header <= note_header_end &&
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: This is already accounted for when note_header_end is calculated.

criu/files-reg.c
avagin: You will need to write three lines of code and it will be more readable:
if (ret < 0)
  return -1;
return 0;

AnorexicAtticusFinch: I've pushed this change

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+
+	if (!num_iterations || note_header >= note_header_end)
+		goto err;
+
+	/* 
+	 * If the size of the notes description is too large or is invalid
+	 * then the build-id could not be obtained.
+	 */
+	if (note_header->n_descsz <= 0 || note_header->n_descsz > 512)
+		goto err;
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: Done

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+	num_iterations = 500;
+
+	/* The note type for the build-id is NT_GNU_BUILD_ID. */
+	while (num_iterations-- && note_header <= note_header_end &&
+			note_header->n_type != NT_GNU_BUILD_ID)
+		note_header = (Elf64_Nhdr *) ((char *) note_header + sizeof(Elf64_Nhdr) +
+						note_header->n_namesz + note_header->n_descsz);
+
+	if (!num_iterations || note_header >= note_header_end)
+		goto err;
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: Done

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+	 */
+	if (strncmp(buf, ELFMAG, SELFMAG))
+		return -1;
+
+	/* 
+	 * If the build-id exists, then it will most likely be present in the
+	 * beginning of the file. Therefore at most only the first 1 MB of the 
+	 * file is mapped.
+	 */
+	start_addr = (char *) mmap(0, min_t(size_t, fd_status->st_size, 1048576),
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: Done

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+	 */
+	if (!file_header->e_phnum)
+		goto err;
+
+	/* 
+	 * If the build-id exists, then it will most likely be present in the
+	 * beginning of the file. Therefore at most only the first 1 MB of the 
+	 * file was mapped.
+	 */
+	file_header_end = min_t(size_t, fd_status->st_size, 1048576) + (size_t) file_header;
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: I've added this as an argument.

criu/files-reg.c
```diff
@@ -1380,12 +1381,304 @@ static bool should_check_size(int flags)
...
+
+	program_header = (Elf64_Phdr *) (file_header->e_phoff + (char *) file_header);
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: I've added this check and pushed the change.

criu/files-reg.c
```diff
@@ -1380,12 +1388,318 @@ static bool should_check_size(int flags)
...
+	rfe->build_id = xmalloc(sizeof(int) * build_id_size);
+	if (!rfe->build_id) {
+		pr_warn("Build-ID (For validation) could not be set for file %s\n",
+				rfe->name);
+		return -1;
+	}
+
+	rfe->n_build_id = build_id_size;
+	for (i = 0; i < build_id_size; i++)
+		rfe->build_id[i] = build_id[i];
```
0x7f454c46, line 1665: Bringing back the comment from the previous review:
> Build-id is a repeated uint32 in regfile.proto and build_id is an array of uchar, that's why I didn't use memcpy here

Allocate as:
`rfe->build_id = xzalloc(round_up(build_id_size, sizeof(uint32_t));`
That way you'll be able to use `memcpy()`.

Currently, by copying bytes into u32, you don't use 3/4 of the space in the array.

criu/files-reg.c
```diff
@@ -1726,6 +2044,78 @@ static int rfi_remap(struct reg_file_info *rfi, int *level)
...
+		return -1;
+	}
+
+	for (i = 0; i < build_id_size; i++)
+		if (build_id[i] != rfi->rfe->build_id[i]) {
+			pr_err("File %s has bad build-ID value %x at index %d (expect %x)\n",
+					rfi->path, build_id[i], i, rfi->rfe->build_id[i]);
+			xfree(build_id);
+			return -1;
+		}
```
0x7f454c46, line 2084: With the comment above about allocation, this could be:
`if (memcmp(build_id, rfi->rfe->build_id, build_id_size))`

criu/files-reg.c
```diff
@@ -1380,12 +1388,318 @@ static bool should_check_size(int flags)
...
+	if (note_header <= (Elf32_Nhdr *) file_header)
+		goto err;
+
+	note_header_end = (Elf32_Nhdr *) (file_header_end - sizeof(Elf32_Nhdr));
+	/* Arbitrary large number to prevent unnecessarily searching the entire file */
+	num_iterations = 500;
+
+	/* The note type for the build-id is NT_GNU_BUILD_ID. */
+	while (num_iterations-- && note_header <= note_header_end &&
+			note_header->n_type != NT_GNU_BUILD_ID)
```
0x7f454c46, line 1446:
```
note_header_end = (Elf32_Nhdr *)(program_header->p_filesz + note_header);
while (note_header <= note_header_end && ...)
```
hmm?

criu/files-reg.c
```diff
@@ -1380,12 +1388,318 @@ static bool should_check_size(int flags)
...
+
+	program_header_end = (Elf32_Phdr *) (file_header_end - sizeof(Elf32_Phdr));
+	num_iterations = file_header->e_phnum + 1;
+
+	/* 
+	 * If the file has a build-id, it will be in the PT_NOTE program header 
+	 * entry AKA the note sections.
+	 */
+	while (num_iterations-- && program_header <= program_header_end &&
+			program_header->p_type != PT_NOTE)
```
0x7f454c46, line 1428: I don't insist here, but I'd just call it `i` insted:
```
for (i = 0; i < file_header->e_phnum && program_header < program_header_end; i++) {
     if (program_header->p_type == PT_NOTE)
          break;
}
```
Also, note that `program_header <= program_header_end` should be without `=`.

0x7f454c46:
I'm also fine if you just force-push into the same review instead of creating a new one for each iteration of reviews.

criu/crtools.c
```diff
@@ -440,6 +440,9 @@ int main(int argc, char *argv[], char *envp[])
 "			Namespace can be specified as either pid or file path.\n"
 "			OPTIONS can be used to specify parameters for userns:\n"
 "			    user:PID,UID,GID\n"
+"  --file-validation [m]\n"
```
0x7f454c46, line 443: Nit here:
`--file-validation METHOD`
(`[m]` in square brackets usually means it's optional)

criu/files-reg.c
```diff
@@ -1380,12 +1388,318 @@ static bool should_check_size(int flags)
...
+				const int fd, size_t mapped_size)
+{
+	int size, num_iterations;
+	size_t file_header_end;
+	Elf32_Phdr *program_header, *program_header_end;
+	Elf32_Nhdr *note_header, *note_header_end;
+
+	file_header_end = (size_t) file_header + mapped_size;
+	if (file_header >= (Elf32_Ehdr *) (file_header_end - sizeof(Elf32_Ehdr)))
+		goto err;
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: Yep, I've made this change.

criu/files-reg.c
AnorexicAtticusFinch: Wouldn't this be a bit harder to read? I find it easier to look at the entire break condition together like in the while loop.

criu/files-reg.c
```diff
@@ -1726,6 +2044,78 @@ static int rfi_remap(struct reg_file_info *rfi, int *level)
...
+	}
+
+	result = validate_with_build_id(fd, fd_status, rfi);
+
+	if (result == -1)
+		return false;
+
+	if (!result)
+		pr_info("File %s could only be validated with file size\n",
+				rfi->path);
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: I had kept the log message outside since in the next PR I'll bring in the checksum code and that would be called if build-ID couldn't be used.

criu/files-reg.c
AnorexicAtticusFinch: Doing this seems to cause job 7 to fail (https://travis-ci.org/github/AnorexicAtticusFinch/criu/jobs/708335971).

criu/files-reg.c
AnorexicAtticusFinch: @0x7f454c46 I'm not sure why, but this change seems to cause job 7 to fail (https://travis-ci.org/github/AnorexicAtticusFinch/criu/jobs/708363457) but I've pushed this code anyway.

avagin:

AnorexicAtticusFinch:
> The criu-dev branch has been force-pushed. Please rebase this PR.

@avagin I tried rebasing but there were a bunch of merge conflicts. I resolved them but that led to 137 commits. I'll just close this PR and push my code again through a different PR from a different branch?

avagin:
@AnorexicAtticusFinch you need to run git rebase -i origin/criu-dev and remove all commits which are not yours:
https://asciinema.org/a/2yDg1mDlhtoIqBUGjtpiRA3Dv

And here is you PR after rebasing it on top of criu-dev: https://github.com/avagin/criu/tree/check-file-buildid

AnorexicAtticusFinch:
> @AnorexicAtticusFinch you need to run git rebase -i origin/criu-dev and remove all commits which are not yours:
> https://asciinema.org/a/2yDg1mDlhtoIqBUGjtpiRA3Dv
> 
> And here is you PR after rebasing it on top of criu-dev: https://github.com/avagin/criu/tree/check-file-buildid

The problem is that when I try to rebase there are no commits, just noop.

criu/files-reg.c
```diff
@@ -1726,6 +2044,78 @@ static int rfi_remap(struct reg_file_info *rfi, int *level)
...
+ * by checking the file's size, build-id and/or checksum with the same metadata
+ * that was stored before dumping.
+ * Checksum is calculated with CRC32C.
+ * Returns true if the metadata of the file matches the metadata stored while
+ * dumping else returns false.
+ */
+static bool validate_file(const int fd, const struct stat *fd_status,
+					const struct reg_file_info *rfi)
+{
+	int result = 1;
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: It is required to be initialized in case the user chooses to only use the file size check.

AnorexicAtticusFinch:
Nevermind, I had to run ```git rebase -i origin/criu-dev~137``` to see the commits. 
@avagin Thanks for the help!

0x7f454c46:
Looks better, but `num_iterations = 500;`.
Probably, the reason why Travis-CI job fails is because in this calculation:
```
		note_header = (Elf64_Nhdr *) ((char *) note_header + sizeof(Elf64_Nhdr) +
						note_header->n_namesz + note_header->n_descsz);
```
you don't take into account the required align:

>        Note sections contain a series of notes (see the struct definitions
>        below).  Each note is followed by the name field (whose length is
>        defined in n_namesz) and then by the descriptor field (whose length
>        is defined in n_descsz) and whose starting address has a 4 byte
>        alignment.  Neither field is defined in the note struct due to their
>        arbitrary lengths.

See the example from [man elf](https://man7.org/linux/man-pages/man5/elf.5.html) how-to calculate it properly.

AnorexicAtticusFinch:
> Looks better, but `num_iterations = 500;`.

Even after aligning, using `program_header->p_filesz` to limit the number of iterations causes job 7 to [fail](https://travis-ci.org/github/AnorexicAtticusFinch/criu/jobs/709661921).

avagin:
> @0x7f454c46 Even after aligning, using program_header->p_filesz to limit the number of iterations causes job 7 to fail.

You need to check that note_header_end is inside a mapping.

```
@@ -1502,6 +1497,8 @@ static int get_build_id_64(Elf64_Ehdr *file_header, unsigned char **build_id,
        pr_err("===");
 
        note_header_end = (Elf64_Nhdr *)((char *) note_header + program_header->p_filesz);
+       if (note_header_end > (Elf64_Nhdr *) (file_header_end - sizeof(Elf64_Nhdr)))
+               note_header_end = (Elf64_Nhdr *) (file_header_end - sizeof(Elf64_Phdr));
 
        /* The note type for the build-id is NT_GNU_BUILD_ID. */
        while (note_header <= note_header_end &&v
```

AnorexicAtticusFinch:
> You need to check that note_header_end is inside a mapping.

Made the change.

Issue #10, opened 2015-07-29
RO or absent /tmp causes dump to fail


xemul:
As described in isse #6 r/o /tmp causes failure in open_mountpoint() due to inability to mkdtemp(). Same will be for the case when /tmp is absent. Same will be for RO / :)

xemul:
aff53534dfa38bcbbaaf4dac01eaae14d6ae189f

PR #1125, opened 2020-06-29
Add error logging to kerndat init

angieenee:
CRIU sometimes returns 1 from main() with no explanation.
These changes add additional logging for initialization errors in kerndat.

There may be discourse surrounding the messages themselves -- specific alternatives are written and recorded elsewhere for each "foo failed" message. Given the large number of functions I've written error messages for in this commit, it may be safer to start with generic "foo failed" messages. These can be replaced depending on the preference of the CRIU maintainers. 

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -390,9 +390,10 @@ static int init_zero_page_pfn(void)
 	ret = vaddr_to_pfn(-1, (unsigned long)addr, &kdat.zero_page_pfn);
 	munmap(addr, PAGE_SIZE);
 
-	if (kdat.zero_page_pfn == 0)
+	if (kdat.zero_page_pfn == 0) {
+		pr_err("vaddr_to_pfn succeeded but kdat.zero_page_pfn is invalid.\n");
 		ret = -1;
-
+	}
 	return ret;
 }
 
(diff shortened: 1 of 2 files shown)


angieenee:
I'm currently looking into the travis failures.

0x7f454c46:
Please, check Travis-CI, it's very red.
Seems like `criu check` fails after this.

avagin:
> Fixed bug + log levels from previous commit (#10) …

The commit history of your pull request has to be clean. If you find a bug in our of your commits, you don't need to create another commit to fix it, instead of this, you have to merge the fix into the target commit.

criu/kerndat.c
```diff
@@ -1047,72 +1059,132 @@ int kerndat_init(void)
...
-		ret = kerndat_has_memfd_create();
-	if (!ret)
-		ret = kerndat_detect_stack_guard_gap();
-	if (!ret)
-		ret = kerndat_uffd();
-	if (!ret)
-		ret = kerndat_has_thp_disable();
+	if (check_pagemap()) {
+		pr_err("check_pagemap failed when initializing kerndat.\n");
+		ret = 1;
```
avagin, line 1072: ret = -1

nit: usually, we use -1 or another negative value in an error case

angieenee: Ah, thanks for letting me know! That's a good _**one**_ to know :)

PR #1129, opened 2020-07-03
mount: adjust log & fsnotify: rework redundant code

azhadchenko:
Adjust some log levels:
1) get_clean_mnt() pr_perror  -> pr_warn since the caller may resolve this situation or print error in the worst case (similar to this [commit](https://github.com/checkpoint-restore/criu/commit/5c5e7695a51318b17e3d982df8231ac83971641c))
2) mnt_is_dir() pr_perror->pr_warn since it may fail several times before finding suitable mountpoint. Added error in lookup_mnt_sdev. Other callers of mnt_is_dir() have corresponding errors.

Remove redundant call to open_handle() in check_open_handle() since alloc_openable() do exactly the same work plus some code.

### DIFF
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -209,7 +209,7 @@ static int open_handle(unsigned int s_dev, unsigned long i_ino,
 
 		mntfd = __open_mountpoint(m, -1);
 		if (mntfd < 0) {
-			pr_err("Can't open mount for s_dev %x, continue\n", s_dev);
+			pr_warn("Can't open mount for s_dev %x, continue\n", s_dev);
 			continue;
 		}
 
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -104,12 +104,19 @@ static int open_by_handle(void *arg, int fd, int pid)
 	return syscall(__NR_open_by_handle_at, fd, arg, O_PATH);
 }
 
+enum {
+	ERR_NO_MOUNT = -1,
+	ERR_NO_PATH_IN_MOUNT = -2,
+	ERR_GENERIC = -3
+};
+
 static char *alloc_openable(unsigned int s_dev, unsigned long i_ino, FhEntry *f_handle)
 {
 	struct mount_info *m;
 	fh_t handle;
 	int fd = -1;
 	char *path;
+	char suitable_mount_found = 0;
 
 	decode_handle(&handle, f_handle);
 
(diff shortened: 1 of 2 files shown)


criu/fsnotify.c
```diff
@@ -228,63 +233,60 @@ int check_open_handle(unsigned int s_dev, unsigned long i_ino,
...
+	 *    so the only portable solution is to carry the whole path
+	 *    to the watchee inside image.
+	 */
+	path = alloc_openable(s_dev, i_ino, f_handle);
 
+	if (!IS_ERR_OR_NULL(path)) {
 		pr_debug("\tHandle 0x%x:0x%lx is openable\n", s_dev, i_ino);
+		goto out;
+	}
+	else if (IS_ERR(path) && PTR_ERR(path) == -ENOMEM)
```
Snorch, line 259: You should read https://www.kernel.org/doc/html/v4.10/process/coding-style.html#placing-braces-and-spaces and fix the codding style here. Else should be on the same line with closing brace and if one branch has braces all branches should have braces.

avagin:
pls rebase this pr

criu/fsnotify.c
```diff
@@ -228,63 +231,60 @@ int check_open_handle(unsigned int s_dev, unsigned long i_ino,
 		FhEntry *f_handle)
 {
 	char *path, *irmap_path;
-	int fd = -1;
+	struct mount_info *mi;
 
 	if (fault_injected(FI_CHECK_OPEN_HANDLE)) {
-		fd = -1;
 		goto fault;
 	}
```
Snorch, line 238: nit: single line branch should omit brackets.

Snorch:
@avagin Can it be merged?

criu/fsnotify.c
```diff
@@ -228,63 +231,59 @@ int check_open_handle(unsigned int s_dev, unsigned long i_ino,
...
+	 *  - in case of migration the inodes might be changed as well
+	 *    so the only portable solution is to carry the whole path
+	 *    to the watchee inside image.
+	 */
+	path = alloc_openable(s_dev, i_ino, f_handle);
 
+	if (!IS_ERR_OR_NULL(path)) {
 		pr_debug("\tHandle 0x%x:0x%lx is openable\n", s_dev, i_ino);
+		goto out;
+	} else if (IS_ERR(path) && PTR_ERR(path) == -3) {
```
avagin, line 255: You need to define constants for these error codes and describe them in the comment for alloc_openable.

azhadchenko: I have added constants for error codes. They look self-explaining to me, but I can add additional comment if you want.

PR #1131, opened 2020-07-05
Restrict Checks for Open/MMAPed Files - CLI & Checksum

AnorexicAtticusFinch:
This adds CLI options and the checksum file validation method in addition to the previous build-ID and file size checks as part of GSoC 2020.

CRIU uses the build-ID validation method by default whenever possible and if that cannot be done, it will try and use the checksum method with the first 1 MB of the file. If this fails as well, it will use only the file size check.

To use the checksum method first and make the build-ID method the fallback:
--file-validation="chksm" --chksm-parameter=[N]
(To calculate checksum with the first N bytes of the file, at most 10 MB)

--file-validation="chksm-full"
(To calculate checksum with the entire file, at most 10 MB)

--file-validation="chksm-period" --chksm-parameter=[N]
(To calculate checksum with every Nth byte of the file)

To explicitly use only the file size check all the time, the following command line option can be used: --file-validation="filesize"
In other words, --file-validation="buildid" will not make a difference as this is the default method

Using the time command to measure the time taken for 2 tests.
zdtm/transition/shmem:
file size: 3.782s
build-id: 4.153s
chksum (Parameter = 1024): 4.465s
chksum-full: 4.722s
chksum-period (Parameter = 1024): 4.498s

zdtm/static/maps04:
file size: 35.317s
build-id: 35.720s
chksum (Parameter = 1024): 35.919s
chksum-full: 36.679s
chksum-period (Parameter = 1024): 36.476s

Mentors: @avagin @0x7f454c46


criu/files-reg.c
```diff
@@ -1602,6 +1608,125 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+		 * If the new iterator position is outside the mapped region (Bytes in the range 
+		 * low_bound and up_bound) of the file, the current region needs to be unmapped
+		 * and the next portion (At most the next 10MB) of the file needs to be mapped.
+		 */
+		if (*iter >= *up_bound && *up_bound != file_size) {
+			munmap(*file_header, *up_bound - *low_bound);
+			*iter -= *up_bound;
+			*low_bound = *up_bound;
+			*up_bound = min_t(size_t, file_size, *up_bound + CHKSM_CHUNK_SIZE);
+			*file_header = (unsigned char *) mmap(0, *up_bound - *low_bound,
```
avagin, line 1643: need to check errors of mmap

criu/files-reg.c
```diff
@@ -1602,6 +1608,125 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+	/*
+	 * The number of iterations is limited to 10485760 times for performance reasons.
+	 * 10485760 was chosen because regardless of the checksum configuration chosen,
+	 * at most 10MB worth of data of the file is collectively considered in the
+	 * calculation of the checksum since the checksum parameter can also never exceed
+	 * 10485760.
+	 */
+	while (!calculate_checksum_iterator_stop_condition(i) &&
+			(i >= low_bound && i < up_bound) &&
+			num_iter < 10485760 && file_header != MAP_FAILED) {
```
avagin, line 1700: CHKSM_CHUNK_SIZE?

criu/files-reg.c
```diff
@@ -1380,12 +1387,472 @@ static bool should_check_size(int flags)
...
+
+	*checksum = 0xFFFFFFFF;
+	calculate_checksum_iterator_init(&i);
+
+	/*
+	 * The number of iterations is limited to 10485760 times for performance reasons.
+	 * 10485760 was chosen because regardless of the checksum configuration chosen,
+	 * at most 10MB worth of data of the file is collectively considered in the
+	 * calculation of the checksum since the checksum parameter can also never exceed
+	 * 10485760.
```
avagin, line 1696: 10485760 has to be configurable

criu/files-reg.c
avagin: and I would say that if you call mmap here, you need to create an initial mapping in calculate_checksum_iterator_init and to unmap in calculate_checksum_iterator_stop_condition. WDYT?

images/regfile.proto
```diff
@@ -4,13 +4,17 @@ import "opts.proto";
...
+	required uint64		pos			= 3;
+	required fown_entry	fown			= 5;
+	required string		name			= 6;
+	optional sint32		mnt_id			= 7 [default = -1];
+	optional uint64		size			= 8;
+	optional bool		ext			= 9;
+	optional uint32         mode    		= 10;
+	repeated uint32 	build_id 		= 11;
+	optional uint32 	checksum 		= 12;
+	optional uint32 	checksum_config 	= 13;
```
avagin, line 18: you need to add comments what each of these fields mean

criu/include/cr_options.h
```diff
@@ -63,6 +63,20 @@ struct cg_root_opt {
 
 #define DEFAULT_TIMEOUT		10
 
+/*
+ * File validation options
+ */
+#define FILE_VALIDATION_FILE_SIZE		1
+#define FILE_VALIDATION_BUILD_ID		2
+#define FILE_VALIDATION_CHKSM			3
+#define FILE_VALIDATION_CHKSM_EVERY		4
```
avagin, line 72: Need to add a comment which describes what each of these constant means.

And I think they have to be in images/regfile.proto as enum

criu/config.c
```diff
@@ -832,6 +862,17 @@ int parse_options(int argc, char **argv, bool *usage_error,
...
 			break;
+		case 1098:
+			if (parse_file_validation_method(&opts, optarg))
+				return 2;
+			break;
+		case 1099:
+			opts.file_validation_chksm_parameter = atoi(optarg);
+			/* The checksum parameter has to be less then 10MB for performance reasons. */
+			if (opts.file_validation_chksm_parameter <= 0 ||
+					opts.file_validation_chksm_parameter > 10485760)
```
avagin, line 873: I don't think that we need this limit. We need to have a reasonable default value.

criu/config.c
```diff
@@ -420,6 +423,31 @@ static int parse_join_ns(const char *ptr)
 	return 0;
 }
 
+static int parse_file_validation_method(struct cr_options *opts, const char *optarg)
+{
+	if (!strcmp(optarg, "filesize")) {
+		opts->file_validation_method = FILE_VALIDATION_FILE_SIZE;
```
avagin, line 429: you have to describe all these options and their values in the usage message (crtools.c) and in Documentation/criu.txt

criu/config.c
```diff
@@ -420,6 +423,31 @@ static int parse_join_ns(const char *ptr)
...
+	} else if (!strcmp(optarg, "buildid")) {
+		opts->file_validation_method = FILE_VALIDATION_BUILD_ID;
+	} else if (!strcmp(optarg, "chksm")) {
```
avagin, line 432: nit: s/chksm/chksum/

criu/config.c
AnorexicAtticusFinch: I've removed this check, the default value is 1024.

criu/include/cr_options.h
AnorexicAtticusFinch: Why does the enum need to be in images/regfile.proto? Currently, I've left it in criu/include/cr_options.h.

criu/files-reg.c
AnorexicAtticusFinch: I removed the limit for the number of the iterations. Now the problem is what if the user chooses to checksum the entire file and the files turn out to be huge? It would take a very long time to calculate the checksum using all the bytes.

criu/files-reg.c
AnorexicAtticusFinch: I moved this entire block of code into calculate_checksum().

avagin:

PR #1135, opened 2020-07-10

angieenee:
When groups already have desired values, we skip calling setgroups.
Furthermore, this check allows us to restore when in an
unprivileged user namespace that cannot use setgroups.

This is a modified version of a [previous PR.](https://github.com/checkpoint-restore/criu/pull/1122)
v2: restructure the setgroups check out of PIE code and into a helper function
v1: add a check within PIE restore

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -3201,6 +3201,43 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
 	thread_args->creds_args = args;
 }
 
+static bool groups_match(gid_t* groups, int n_groups)
+{
+	int n, len;
+	bool ret;
+	gid_t* gids;
+
+	n = getgroups(0, NULL);
+	if (n == -1) {
+		pr_perror("Failed to get number of supplementary groups");
+		ret = false;
+	}
+	if (n != n_groups)
+		return false;
+	if (n == 0)
+		return true;
+
+	len = n * sizeof(gid_t);
+	gids = xmalloc(len);
+	if (gids == NULL)
+		return false;
+
+	n = getgroups(n, gids);
+	if (n == -1) {
+		pr_perror("Failed to get supplementary groups");
+		ret = false;
+	} else {
+		/*
+		* Because getgroups sorts gids,
+		* memcmp compares two sorted arrays
+		*/
+		ret = !memcmp(gids, groups, len);
+	}
+
+	xfree(gids);
+	return ret;
+}
+
 static struct thread_creds_args *
 rst_prep_creds_args(CredsEntry *ce, unsigned long *prev_pos)
 {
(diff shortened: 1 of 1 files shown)


criu/cr-restore.c
```diff
@@ -3128,6 +3128,36 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
...
+	if (n != n_groups)
+		return false;
+	if (n == 0)
+		return true;
+
+	len = n * sizeof(gid_t);
+	gids = xmalloc(len);
+	if (gids == NULL) {
+		return false;
+	}
```
rppt, line 3147: Nit: the braces here are not necessary.

avagin:

criu/cr-restore.c
```diff
@@ -3201,6 +3201,35 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
 	thread_args->creds_args = args;
 }
 
+static bool groups_match(gid_t* groups, int n_groups)
+{
+	int n, len;
+	bool ret;
+	gid_t* gids;
+
+	n = getgroups(0, NULL);
```
avagin, line 3210: pls report an error if getgroups failed

criu/cr-restore.c
```diff
@@ -3201,6 +3201,35 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
...
+	gids = xmalloc(len);
+	if (gids == NULL)
+		return false;
+
+	n = getgroups(n, gids);
+	if (n == -1) {
+		pr_perror("Failed to get supplementary groups");
+		ret = false;
+	} else {
+		ret = !memcmp(gids, groups, len);
```
avagin, line 3226: memcp will work only if both arrays are sorted. Are they sorted? if so please add a comment here.

angieenee: Good catch. It appears that the `groups` variable in this function is set from [here](https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/pie/parasite.c#L277) with `sys_getgroups`, so the order *should* be the same. However, on the off-chance that these two arrays are not in the same order, the restoring will revert to its default process as if this `groups_match` function wasn't added. Let me know if you'd prefer some form of sorting before calling the `memcpy`!

criu/cr-restore.c
angieenee: Done! Feel free to let me know if you'd like the error message to be modified.

criu/cr-restore.c
avagin: The right answer is that the Linux getgroups returns groups in the sorted order: https://github.com/torvalds/linux/blob/master/kernel/groups.c#L209.

There is no guarantee that this will not be changed in the future, but I think it will not be done soon.

I was lazy to read the kernel code before writing the previous comment. Pls, add a comment into the code.

criu/cr-restore.c
```diff
@@ -3201,6 +3201,43 @@ static void rst_reloc_creds(struct thread_restore_args *thread_args,
...
+
+	n = getgroups(n, gids);
+	if (n == -1) {
+		pr_perror("Failed to get supplementary groups");
+		ret = false;
+	} else {
+		/*
+		* Because getgroups sorts gids,
+		* memcmp compares two sorted arrays
+		*/
```
rppt, line 3233: The comment is short enough to fit into a single line ;-)
I'll update it myself during the pull.

And, for future reference, in multi-line comments the `*` should be aligned:

```
/*
 * This comment is very long and 
 * spans multiple lines
 */
```

rppt:
I've merged it manually

criu/cr-restore.c
angieenee: Ah, thank you for letting me know + for manually merging! I'll be sure my future comments a**_line_** with these guidelines :^)

PR #1140, opened 2020-07-16
tests: criu-image-streamer change dev branch to master branch

nviennot:
Now that the new protocol is in, we can move to the master branch on criu-image-streamer


avagin:
pls add the SOB line and the message body to the commit message.

nviennot:
Done, sorry for the trouble

PR #1142, opened 2020-07-17
pidns: fixup

adrianreber:
I pushed the wrong branch to the pidns PR

https://github.com/checkpoint-restore/criu/pull/1056

which resulted in the wrong patches getting merged.

This is the actual result from the review.

The code in `criu-dev` already works. When using `criu-dev` to checkpoint a container out of pod in cri-o I saw that the wrong magic number was in the pidns image file.

Sorry for the mixup of the branches, but this should be now the result from the review in the original PR. Mainly using a correct magic number and using `call_in_child_process()` instead of `fork()`.

### DIFF
--- a/test/others/ns_ext/run.sh
+++ b/test/others/ns_ext/run.sh
@@ -26,6 +26,7 @@ CRIU=../../../criu/criu
 if [[ "$NS" == "net" ]]; then
 	setsid unshare -n bash -c 'unshare -n sh _run.sh pidfile2 & unshare -n sh _run.sh pidfile3 & ip link add xxx type veth && ip link add mymacvlan1 link xxx type macvlan mode bridge && . _run.sh pidfile' < /dev/zero &> output &
 elif [[ "$NS" == "pid" ]]; then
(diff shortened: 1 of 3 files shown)


test/others/ns_ext/run.sh
```diff
@@ -54,19 +58,35 @@ ino2=$(ls -iL $MNT2 | awk '{ print $1 }')
...
 [ "$ino" != "$new_ino" ] && {
+	echo "Inode of new NS is different"
 	echo FAIL
 	exit 1
 }
 [ "$ino2" != "$new_ino2" ] && {
+	echo "Inode of new NS is different"
 	echo FAIL
 	exit 1
 }
```
rst0git, line 92: @adrianreber Could you add `make -C test/others/ns_ext run` in `scripts/travis/travis-tests`?

adrianreber: That is already there https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/travis/travis-tests#L215

rst0git: you are right, thanks

PR #1143, opened 2020-07-17
compel: GOT table corruption fix, and other code improvements

nviennot:
I was recently burnt by a bug in compel when writing https://github.com/nviennot/run_as_exe
The GOT table, used by certain symbols, was sharing the same memory space as the args region used when invoking a parasite service command. This is not a problem in the current CRIU implementation as there are no GOT symbols in its PIE blobs. But it could happen in the future.

At first, I tried allocating the GOT table via a linker script modification, but failed. So I opted to allocate the GOT table at runtime, as the current codebase intended.

This was tricky because the GOT table needs to be squeezed between the parasite blob and the args area. Sadly, the linker scripts were setting the location of the args section statically, forbidding additional room for a GOT table. So I made the args area dynamic. It's only then that I realized that most of the parasite service assembly entry code could be implemented in C.
This lead to further simplification of the compel implementation.

Each commit holds on its own. I apologize for such a PR as this may not be an easy review to do but I think it's best to address these GOT memory corruption now to avoid getting bitten in the future.

Thanks,
Nico


avagin:
@cyrillos I think we need to restore you from the ten years old snapshot. At that time, you had much more time to work on CRIU;).

cyrillos:
> @cyrillos I think we need to restore you from the ten years old snapshot. At that time, you had much more time to work on CRIU;).

Oh, Andrew, so true :-)

compel/src/lib/infect.c
```diff
@@ -862,11 +862,6 @@ static int compel_map_exchange(struct parasite_ctl *ctl, unsigned long size)
 	return ret;
```
rppt, line 868: > comepl: stop rounding up the parasite
  ^coMPEL ;-)

nviennot: woops
I fixed two comepl -> compel typos
thank you

0x7f454c46:
Merged manually.

nviennot:
Hello @0x7f454c46 

> Nice work! That was a pleasure to review :)
Thank you, I'm glad :)

Regarding the `BUG_ON()` alternative:
1) Instead of `ctl->rstack = (void *)round_down((unsigned long)ctl->rstack, 16);`, I'd rather `goto err`.
If we add code and mess up the parasite layout, we should know quickly, instead of having a success.
2) Or, we always do the round_down(), and round_up(), without conditionals. But it's kind of silly, because we know it's the case already.

0x7f454c46:
Agree, (1) looks good to me.
Added #1157

PR #1145, opened 2020-07-20
[WIP: gsoc 2020] Add decoder to crit and change logging structure

prakritigoyal19:
Changes through this pr:

1. Encoder:
a. At the moment, I have transferred all calls for logging to the flog.c file. In the flog_encode_msg(), it is checked whether binary logging is turned on or not. If the binary logging is not enabled, the control is transferred to vprint_on_level(), where the regular logging takes place. Else it is encoded in the binary.
b. To be able to achieve this, I had to merge flog to criu and had to make small changes to multiple files.
c. There is no change in the earlier functionality of regular logging. The binary logging while dumping of processes also works well, but while restoring, I am getting a few errors that I am investigating.

2. Decoder:
a. Added decoder to the crit folder.
b. To run the decoder:
i. First build criu normally
ii. Run using command: ./crit/decoder <input_file_path> <otput_file_path>
c. The decoder worked well when the file passed to it is wholly encoded via flog, but while logging, there are a few statements that are logged via the regular printer (this mostly happens when the logging is done via compel) and the decoder is not able to read through these statements. So I have commented out all the compel log messages for the time being.
The decoder currently passes for all architectures except Arm64 architecture. I am looking for ways to resolve that.

 
During logging via flog no timestamps are being printed, and neither is the pid being recorded into the files.

Signed-off-by: prakritigoyal19 <prakritigoyal19@gmail.com>


compel/plugins/std/log.c
```diff
@@ -355,6 +355,26 @@ void print_on_level(unsigned int loglevel, const char *format, ...)
 	sbuf_log_flush(&b);
 }
 
+int flog_encode_msg(int loglevel, unsigned int nargs, unsigned int mask, const char *format, ...){
+	/*TO DO: Change this to add binary logging here*/
```
xemul, line 359: @prakritigoyal19 , you have commented-out all the compel's messages, why is flog needed in compel after it?

criu/arch/aarch64/vdso-pie.c
```diff
@@ -22,9 +22,9 @@ int vdso_redirect_calls(unsigned long base_to, unsigned long base_from,
 		if (vdso_symbol_empty(&from->symbols[i]))
 			continue;
 
-		pr_debug("br: %lx/%lx -> %lx/%lx (index %d)\n",
+		/*pr_debug("br: %lx/%lx -> %lx/%lx (index %d)\n",
```
xemul, line 25: Why is criu logging message commented-out?

criu/arch/s390/crtools.c
```diff
@@ -86,8 +86,9 @@ static void print_core_gs_cb(CoreEntry *core)
 		pr_debug("       No GS_CB\n");
 		return;
 	}
-	for (i = 0; i < 4; i++)
+	for (i = 0; i < 4; i++){
```
xemul, line 89: Please, don't add irrelevant changes into a patch. Also criu's coding style doesn't welcome braces around single line blocks.

criu/autofs.c
```diff
@@ -360,10 +360,11 @@ static int autofs_revisit_options(struct mount_info *pm)
 				free(pm->options);
 
 				pm->options = xstrdup(token);
-				if (!pm->options)
+				if (!pm->options){
```
xemul, line 363: Irrelevant change.

criu/arch/x86/include/asm/compat.h
```diff
@@ -10,6 +10,7 @@
 #endif
 
 #include <sys/mman.h>
+#include "flog.h"
```
xemul, line 13: Try to avoid header into header inclusion. Here flog.h is not needed in compat.h itself, but is apparently used to fanout itself into other .c files. Better patch those .c files.

criu/config.c
```diff
@@ -864,12 +868,13 @@ int parse_options(int argc, char **argv, bool *usage_error,
 	return 0;
 
 bad_arg:
-	if (idx < 0) /* short option */
+	if (idx < 0){ /* short option */
```
xemul, line 871: Irrelevant change

criu/config.c
```diff
@@ -886,9 +891,9 @@ int check_options(void)
 	if (opts.weak_sysctls)
 		pr_info("Will skip non-existant sysctls on restore\n");
 
-	if (opts.deprecated_ok)
+	if (opts.deprecated_ok){
```
xemul, line 894: Irrelevant change

criu/cr-check.c
```diff
@@ -203,10 +203,11 @@ static int check_prctl_cat1(void)
 		pr_msg("Info  prctl: PR_SET_MM_MAP_SIZE is not supported\n");
 		ret = prctl(PR_SET_MM, PR_SET_MM_BRK, (unsigned long)sbrk(0), 0, 0);
 		if (ret < 0) {
-			if (errno == EPERM)
+			if (errno == EPERM){
```
xemul, line 206: Irrelevant change

criu/cr-dump.c
```diff
@@ -1564,9 +1564,9 @@ static int cr_pre_dump_finish(int status)
 	if (write_img_inventory(&he))
 		ret = -1;
 
-	if (ret)
+	if (ret){
```
xemul, line 1567: Irrelevant change

criu/cr-restore.c
```diff
@@ -1561,11 +1561,12 @@ static void sigchld_handler(int signal, siginfo_t *siginfo, void *data)
 		break;
 	}
 
-	if (exit)
+	if (exit){
```
xemul, line 1564: Irrelevant change

criu/cr-service.c
```diff
@@ -1280,12 +1280,13 @@ static void reap_worker(int signo)
 			return;
 		}
 
-		if (WIFEXITED(status))
+		if (WIFEXITED(status)){
```
xemul, line 1283: Irrelevant change

criu/files-reg.c
```diff
@@ -1553,10 +1553,11 @@ static void rm_parent_dirs(int mntns_root, char *path, int count)
 		if (prev)
 			*prev = '/';
 
-		if (unlinkat(mntns_root, path, AT_REMOVEDIR))
+		if (unlinkat(mntns_root, path, AT_REMOVEDIR)){
```
xemul, line 1556: Irrelevant change (I'm tired or marking these braces, sorry)

criu/flog.c
```diff
@@ -75,18 +78,20 @@ static uint64_t mbuf_size = sizeof(_mbuf);
...
 		return -1;
 	}
 	return 0;
 }
 
 /*extern char *rodata_start;
-extern char *rodata_end;
-*/
+extern char *rodata_end;*/
+
```
xemul, line 94: This change shouldn't belong to this patch.

criu/flog.c
```diff
@@ -150,66 +155,87 @@ int flog_close(int fdout)
...
 	size_t i;
+	int fdout;
+	unsigned int current_loglevel;
+
+	if (!opts.log_in_binary){
+
+		va_start(argptr, format);
+		vprint_on_level(loglevel, format, argptr);
+		va_end(argptr);
+
```
xemul, line 172: Put "return" here (or forward goto) to avoid indentation shift for the rest of the function.

xemul:
@prakritigoyal19 , great work! Let's update this PR like this:

1. Remove the irrelevant changes from it. In particular, there are many hunks that add braces to if's and loop's. Criu coding style doesn't welcome braces around single-line blocks.

2. Split this one big patch into several:

  a. comment-out all compel's messages (or better -- define the compel's print_on_level so that it preprocesses to empty)
  b. move flog from flog/ to criu/ (you have these hunks) as is, without logical changes
  c. intruduce the --binary-log option doing nothing
  d. wire the flog into criu properly (patch the print_on_level so that it makes selection of how to print based on the option)
  e. add the decoder
  f. use the decoder where appropriate (in zdtm.py)

3. Check (and fix) the CI failures

xemul:
@cyrillos , @avagin , currently compel shares the logging with criu in terms of file descriptor only, this makes flog integration to double work -- we need to have it in two places -- compel and criu. We need to come up with some better way of doing it. Typically libraries do not use their own logging, but instead have "register_logger" call that allows users to register the logging function that will be called by the library. How about walking this route with compel too?

cyrillos:
> @cyrillos , @avagin , currently compel shares the logging with criu in terms of file descriptor only, this makes flog integration to double work -- we need to have it in two places -- compel and criu. We need to come up with some better way of doing it. Typically libraries do not use their own logging, but instead have "register_logger" call that allows users to register the logging function that will be called by the library. How about walking this route with compel too?

Sounds like a plan

criu/arch/s390/crtools.c
prakritigoyal19: I added these braces to the code because when I tried to build criu after adding binary logging, I had gotten errors like: else without a previous if. This error was fixed after adding braces to it. 
I understand that this goes against the coding practice and have found a better way of doing it.

criu/flog.c
prakritigoyal19: Okay Sir, I will add a goto statement to transfer flow back to the regular printer.

criu/arch/x86/include/asm/compat.h
prakritigoyal19: Sure sir, I will make this change and keep in mind for the future too.

compel/plugins/std/log.c
prakritigoyal19: While compiling the code I am getting the error: external symbol in pie. This causes the build to fail with an error 255. To overcome this error, I added this to compel. I will look for alternatives to be able to fix this. 

Also, the rpc test of criu tries to read one of the messages written by compel. The message is: “The criu itself is within dumped tree”, from criu/compel/src/lib/infect.c. This causes the Travis CI to fail for multiple architectures after I comment out all the compel messages. Please let me know if I should allow only that message to be printed or should I try changing the message in rpc or would there be some other alternative to overcome this error.

prakritigoyal19:
@xemul After splitting this patch into smaller patches, would you like me to send each patch as a separate pull request or should I send it as different commits within a single pull request?

xemul:
> @xemul After splitting this patch into smaller patches, should I send each patch as a separate pull request or should I send it as different commits within a single pull request?

@prakritigoyal19 , all commits in one pill request

prakritigoyal19:
The log message "The criu itself is within dumped tree." is expected to be present in the log file by test/others/rpc/config_file.py file. Since the compel logging is disabled, this message from file compel/src/lib/infect.c is not being printed. This causes the rpc test to fail. Hence the CI failing for multiple architectures.

xemul:
@prakritigoyal19 , I see only one commit in the updated PR, where are the rest?

prakritigoyal19:
@xemul I have the rest of the commits ready to be sent. Since this one failed I thought it would be better to fix this first and then send the others.
Should I send the rest also right now?

xemul:
Yes, please. Make sure that the pr contains all the parts that were described previously

prakritigoyal19:
@xemul, The last part of the pr, using decoder wherever needed, is pending. According to the timeline, I had initially planned on doing it in the next phase. I will try to do it as soon as possible and send it across before the following week. 

I also had a few queries regarding the integration of the decoder:

1. How would the decoder be called? Currently to use the decoder, the command, `crit/decoder <input_filename> <output_filename>`, is used. Should I continue with this, or should it be changed?

2. Before adding the decoder to zdtm.py, should I convert all its logs to binary, or should there be a separate option to enable binary logging? Or should I add the decoder in the dormant mode for the time being?

criu/config.c
```diff
@@ -528,6 +528,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{ "status-fd",			required_argument,	0, 1088 },
 		BOOL_OPT(SK_CLOSE_PARAM, &opts.tcp_close),
 		{ "verbosity",			optional_argument,	0, 'v'	},
+		{ "binary-file",		optional_argument,	0, 'b' },
```
xemul, line 531: It should be "no_argument", since it's a boolean option. With the respective short option change.

criu/cr-restore.c
```diff
@@ -3824,15 +3824,19 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 
 	__gcov_flush();
 
-	pr_info("task_args: %p\n"
+	/*This message is commented out as
+	it causes a segmentation fault when restore
+	is run while using binary logging*/
```
xemul, line 3829: Have you managed to find out why it crashes?

criu/ipc_ns.c
```diff
@@ -38,7 +38,7 @@
 
 static void pr_ipc_desc_entry(unsigned int loglevel, const IpcDescEntry *desc)
 {
-	print_on_level(loglevel, "id: %-10d key: %#08x uid: %-10d gid: %-10d "
+	flog_encode(loglevel, "id: %-10d key: %#08x uid: %-10d gid: %-10d "
```
xemul, line 41: Why do you change callers? Is change in log.h not enough?

criu/arch/x86/restorer.c
```diff
@@ -2,6 +2,7 @@
 #include <unistd.h>
 
 #include "types.h"
+#include "flog.h"
```
xemul, line 5: Why is this header needed here?

compel/plugins/include/uapi/std/log.h
```diff
@@ -26,5 +26,6 @@ extern void std_sprintf(char output[STD_LOG_SIMPLE_CHUNK], const char *format, .
 	__attribute__ ((__format__ (__printf__, 2, 3)));
 extern void print_on_level(unsigned int loglevel, const char *format, ...)
 	__attribute__ ((__format__ (__printf__, 2, 3)));
+extern int flog_encode_msg(int loglevel, unsigned int nargs, unsigned int mask, const char *format, ...);
```
xemul, line 29: You should have commented out all logging from compel, why does flog forward declaration go here?

criu/flog.c
```diff
@@ -178,10 +200,12 @@ int flog_encode_msg(int fdout, unsigned int nargs, unsigned int mask, const char
 		 * a copy (FIXME implement rodata refs).
 		 */
 		if (mask & (1u << i)) {
-			p = memccpy(str_start, (void *)m->args[i], 0, mbuf_size - (str_start - mbuf));
-			if (p == NULL) {
-				fprintf(stderr, "No memory for string argument\n");
-				return -1;
+			if(m->args[i]){
```
xemul, line 203: Please, explain this change.

criu/flog.c
```diff
@@ -200,11 +224,17 @@ int flog_encode_msg(int fdout, unsigned int nargs, unsigned int mask, const char
...
 			return -1;
 	} else {
 		mbuf += m->size;
 		mbuf_size -= m->size;
 	}
 	return 0;
-}
\ No newline at end of file
+
+	regular_logging:
```
xemul, line 235: Since this is a goto-only code, then it's worth just moving into the respective if () branch. I thought you re-used some parts of this function in both if and else branches in the previous version of this patch, that's why asked for goto that time.

test/others/rpc/config_file.py
```diff
@@ -111,6 +111,7 @@ def check_results(resp, log):
         print('FAIL: Missing the expected error message in RPC response')
         sys.exit(-1)
     # Look into the log file for the same message
+    # This message is being searched into the log file.
```
xemul, line 114: Please remove this comment, I guess it's here by mistake.

criu/flog.c
```diff
@@ -33,9 +33,7 @@ static uint64_t mbuf_size = sizeof(_mbuf);
 	ffi_arg rc;
 	size_t i, ret;
 	char *fmt;
-
```
xemul: Please, remove these lines-removals, they just complicate the review.

criu/include/cr_options.h
```diff
@@ -174,6 +174,7 @@ struct cr_options {
 
 	/* This stores which method to use for file validation. */
 	int 			file_validation_method;
+	int 			log_in_binary;
```
xemul, line 177: Should be bool

criu/files.c
```diff
@@ -1005,7 +1005,11 @@ static int recv_fd_from_peer(struct fdinfo_list_entry *fle)
 		else if (ret)
 			return -1;
 
-		pr_info("Further fle=%p, pid=%d\n", tmp, fle->pid);
+		/*This message is commented out as
+		it causes a segmentation fault when restore
+		is run while using binary logging*/
```
xemul, line 1010: Same question here -- do you know the reason of this crash?

criu/flog.c
```diff
@@ -33,7 +36,9 @@ static uint64_t mbuf_size = sizeof(_mbuf);
 	ffi_arg rc;
 	size_t i, ret;
 	char *fmt;
+
```
xemul: Please, remove line-addings, they complicate the review.

rppt:
> OK, mostly cosmetic fixes left to be done. Please, update the commits and push the full PR again.
> Also it would be very nice if each commin in the PR contains not only the one-line description, but also a good comment describing the change at the top level (good commit comment says why it's done like it is, but not how).

@prakritigoyal19  you may want to take a look at [CONTRIBUTING.md](CONTRIBUTING.md) for guidance about commit messages and pull requests.

criu/cr-restore.c
prakritigoyal19: I did some investigation on this and found that it is linked to the default value of -1 assigned to the flog_genbit in the _Generic() macro. Today, when I changed it to 0, and the restore ran successfully with these messages as well. Should I make the _Generic return 0 as the default value to avoid this crash?

criu/ipc_ns.c
prakritigoyal19: The change in log.h only causes change when the log message is called via the macros pr_info, pr_err, etc. To ensure that all messages are being printed into the file are in the same format, I had to change all the callers. (In binary when binary logging is enabled. In readable form when not.)

criu/arch/x86/restorer.c
prakritigoyal19: Due to the presence of logging macros in this file and all the header files it includes, the absence of this header throws me an error: implicit declaration of function ‘flog_encode’. To avoid this error I had added this header here.

compel/plugins/include/uapi/std/log.h
prakritigoyal19: There are log messages in criu too that had initially used compel’s print_on_level() to print the messages into the file. I converted all the pr_msg, etc to call flog_encode_msg() instead of print_on_level(). Now, these log messages search for the flog’s function in compel’s log.c file. Removing the function header gives me an error: External symbol in PIE?. This causes the build to fail. I am not able to figure out why it is searching in compel/plugin/std/log.c instead of the flog.c file.

criu/flog.c
prakritigoyal19: There are a few log messages while dumping that try to print a null string into the file. The m->args[i] became null during this point. When the memccpy() function tried to copy this argument, it caused a segmentation fault. Thus to avoid this I have placed a check to ensure that the value being copied is not NULL, i.e. if the m->args[i] is null we skip the memccpy function.

criu/flog.c
prakritigoyal19: Alright Sir, I will make this change.
The if statement here determines whether the binary logging option is enabled. If it is, then we use the flog printer to store the messages into the file. If it is not, then the criu/log.c file’s vprint_on_level is called to log via the regular printer.

compel/src/lib/infect.c
```diff
@@ -230,7 +230,7 @@ int compel_wait_task(int pid, int ppid,
 	if (ret < 0 || WIFEXITED(status) || WIFSIGNALED(status)) {
 		if (ss->state != 'Z') {
 			if (pid == getpid())
-				pr_err("The criu itself is within dumped tree.\n");
+				pr_err("The criu itself is within dumped tree.\n"); //This message is written here
```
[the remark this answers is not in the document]
prakritigoyal19: This message causes the CI to fail at the rpc check. Thus I had marked it here. I shall remove this comment in the next pr.

criu/arch/x86/restorer.c
xemul: OK, so you took my words about not to include headers in headers too literally :) Here's how it's usually done:

- If you have a lot of .c files, that need something from some header, then you include this header into all of them. Even if there's a shortcut -- by including this header into some other one, which is in turn already present in all those .c files -- you still go and include the header into .c files.

- If you have a .h file, that needs something from another header, then you include header into header. There's another solution to this -- you go and include this another header in all the .c files, that have the 1st header, but it's not the way to go either.

Regarding this place -- the criu/arch/x86/restorer.c does not need the flog.h by itself. The flog.h turns out to be needed by restorer.c because restorer.c includes log.h which needs flog. This is the 1st item in my example, so include flog.h into log.h and that's it.

criu/cr-restore.c
xemul: No, please, find out why -1 is reported. Maybe it's unsupported data type and/or format specified?

criu/ipc_ns.c
xemul: Ah, I see.

criu/cr-restore.c
prakritigoyal19: The argument that is trying to be printed is a pointer of type _struct task_restore_args *_ and, this doesn’t fit into any of the other specified data types in the _Generic macro, and hence the -1 is returned. 
 
If I include this struct type explicitly in the _Generic list, the error is resolved. But in this pr_info, _task_args->clone_restore_fn_, is a function pointer. I had typecasted it to (void *) and then it did not return an error.

criu/arch/x86/restorer.c
prakritigoyal19: I have already included flog.h in the criu/include/log.h. Even after the addition of flog.h in log.h, during the compile time I get this error, if I do not include this header here.

PR #1148, opened 2020-07-22
Restrict Checks for Open/MMAPed Files - CLI & Checksum

AnorexicAtticusFinch:
This adds CLI options and the checksum file validation method in addition to the previous build-ID and file size checks as part of GSoC 2020.

CRIU uses the build-ID validation method by default whenever possible and if that cannot be done, it will try and use the checksum method with the first 1 MB of the file. If this fails as well, it will use only the file size check.

To use the checksum method first and make the build-ID method the fallback:
--file-validation="chksm" --chksm-parameter=[N]
(To calculate checksum with the first N bytes of the file, at most 10 MB)

--file-validation="chksm-full"
(To calculate checksum with the entire file, at most 10 MB)

--file-validation="chksm-period" --chksm-parameter=[N]
(To calculate checksum with every Nth byte of the file)

To explicitly use only the file size check all the time, the following command line option can be used: --file-validation="filesize"
In other words, --file-validation="buildid" will not make a difference as this is the default method

Using the time command to measure the time taken for 2 tests.
zdtm/transition/shmem:
file size: 3.782s
build-id: 4.153s
chksum (Parameter = 1024): 4.465s
chksum-full: 4.722s
chksum-period (Parameter = 1024): 4.498s

zdtm/static/maps04:
file size: 35.317s
build-id: 35.720s
chksum (Parameter = 1024): 35.919s
chksum-full: 36.679s
chksum-period (Parameter = 1024): 36.476s

Mentors: @avagin @0x7f454c46

This also adds ZDTM tests for build-ID and checksum validation methods.

Link to the previous PR:
https://github.com/checkpoint-restore/criu/pull/1131

Changes from the previous PR:
rebase on the current criu-dev
cosmetic changes
other minor changes


AnorexicAtticusFinch:
@avagin @0x7f454c46 What if the user chooses to checksum the entire file and the files turn out to be huge? Without some kind of limit, it would take a very long time to calculate the checksum using all the bytes.

criu/files-reg.c
```diff
@@ -1571,6 +1577,111 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+			up_bound = min_t(size_t, fd_status->st_size, up_bound + CHKSM_CHUNK_SIZE);
+			file_header = (unsigned char *) mmap(0, up_bound - low_bound,
+						PROT_READ, MAP_PRIVATE | MAP_FILE, fd, low_bound);
+		}
+
+		if (file_header == MAP_FAILED) {
+			pr_warn("Couldn't mmap file with fd %d", fd);
+			return false;
+		}
+
```
Snorch, line 1665: nit: Tabs on and empty line.

criu/files-reg.c
```diff
@@ -1571,6 +1577,111 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+		calculate_checksum_iterator_next(&i);
+
+		/*
+		 * If the new iterator position is outside the mapped region (Bytes in the range
+		 * low_bound and up_bound) of the file, the current region needs to be unmapped
+		 * and the next portion (At most the next 10MB) of the file needs to be mapped.
+		 */
+		if (i >= up_bound && up_bound != fd_status->st_size) {
+			munmap(file_header, up_bound - low_bound);
+			i -= up_bound;
```
Snorch, line 1654: Lets assume:
opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_FIRST
opts.file_validation_chksm_parameter == CHKSM_CHUNK_SIZE+5
and st_size > CHKSM_CHUNK_SIZE*2

In calculate_checksum you would cycle in while() loop untill i == CHKSM_CHUNK_SIZE-1, then calculate_checksum_iterator_next will increase i to CHKSM_CHUNK_SIZE, then you get to "if (i >= up_bound && up_bound != fd_status->st_size) {" true branche, and do "i -= up_bound", so that i == 0 now. Also you set low_bount = up_bound == CHKSM_CHUNK_SIZE. So while loop will stop on the next iteration because i(0) < low_bound(CHKSM_CHUNK_SIZE). You've only checked CHKSM_CHUNK_SIZE bytes, not what is expected CHKSM_CHUNK_SIZE+5.

Looks like the logic is broken.

criu/files-reg.c
AnorexicAtticusFinch: Removed any errant whitespaces.

criu/files-reg.c
AnorexicAtticusFinch: Sorry about that, I had overlooked it. I've fixed it and pushed the new code.

criu/files-reg.c
```diff
@@ -2035,6 +2199,59 @@ static int validate_with_build_id(const int fd, const struct stat *fd_status,
...
+ * stored or could not be obtained.
+ */
+static int validate_with_checksum(const int fd, const struct stat *fd_status,
+					const struct reg_file_info *rfi)
+{
+	u32 checksum;
+
+	if (!rfi->rfe->has_size)
+		return 1;
+
```
Snorch, line 2215: Again empty line with tab.

criu/files-reg.c
```diff
@@ -2035,6 +2199,59 @@ static int validate_with_build_id(const int fd, const struct stat *fd_status,
...
+		opts.file_validation_chksm_config = rfi->rfe->checksum_config;
+	}
+
+	/*
+	 * If the checksum parameter used while dumping is not the same as the current
+	 * checksum parameter, then the user is informed of this and the checksum parameter
+	 * used while dumping is considered. If the checksum config is to checksum the entire
+	 * file, then no changes need to be made since the checksum parameter isn't used anyway.
+	 */
+	if (opts.file_validation_chksm_config != FILE_VALIDATION_CHKSM_EVERY &&
```
Snorch, line 2237: Space in the end of line.

If you use "git show" they would be highlighted red. (at least in my git setup)

criu/files-reg.c
Snorch: Maybe we can add a test for this case?

criu/files-reg.c
AnorexicAtticusFinch: I didn't know that, thanks!

criu/files-reg.c
AnorexicAtticusFinch: Yeah I was planning on adding a test along these lines along with 2 others:
- `opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_PERIOD` and`opts.file_validation_chksm_parameter == CHKSM_CHUNK_SIZE+10`.
- Alter the file after dump but before restore, this should cause restore to fail. If restore succeeds then the test fails.

These 3 tests should be enough for the checksum method?

Snorch: > These 3 tests should be enough for the checksum method?

I would say we need to add at least one test for each checksum method (3 tests). And for tests with chksm-parameter we should likely have two tests 1)if parameter < CHKSM_CHUNK_SIZE and 2) > CHKSM_CHUNK_SIZE. So 5 tests would be nice.

For the exact case you've fixed we should alter e.g. the last byte of "the first 'N'' to check that we detect broken file even if only the 'N'-th byte is changed.

To sum-up: more tests are always better, but don't spend too much time on it =) Add some tests to PR and we will see if it's enough.

criu/files-reg.c
```diff
@@ -1571,6 +1577,111 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+	if (file_header == MAP_FAILED) {
+		pr_warn("Couldn't mmap file with fd %d", fd);
+		return false;
+	}
+
+	*checksum = 0xFFFFFFFF;
+	calculate_checksum_iterator_init(&i);
+
+	while (!calculate_checksum_iterator_stop_condition(i) &&
+			i >= low_bound && i < up_bound && i < fd_status->st_size) {
```
0x7f454c46, line 1644: This would be easier to understand if it was for () cycle.
You don't need checking low_bound every loop as the iterator only goes up.
Something like:
```
if (opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_FIRST)
     up_bound = min_t(up_bound, opts.file_validation_chksm_parameter);
if (opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_PERIOD)
     step = opts.file_validation_chksm_parameter;
else
     step = 1;
for (i = 0; i < up_bound; i += step)
{
/* ... */
}
```
At this moment the logic is obscure and hard to read.

criu/files-reg.c
```diff
@@ -2054,8 +2271,21 @@ static bool validate_file(const int fd, const struct stat *fd_status,
...
+				rfi->path);
+			result = validate_with_checksum(fd, fd_status, rfi);
+		}
+	} else if (opts.file_validation_method == FILE_VALIDATION_CHKSM) {
+		result = validate_with_checksum(fd, fd_status, rfi);
+		if (!result) {
+			pr_info("Could not validate with checksum, trying build-ID validation instead for file %s\n",
+				rfi->path);
+			result = validate_with_build_id(fd, fd_status, rfi);
+		}
```
0x7f454c46, line 2287: I don't see how failed validation with checksum can success with build-id.

criu/files-reg.c
Snorch: I generally I agree with @0x7f454c46, this code can be simplified. But at this point the logic looks at least semanticly _right_ to me, please don't break it =)

This:
up_bound = min_t(up_bound, opts.file_validation_chksm_parameter);
does not go well with:
up_bound = min_t(size_t, fd_status->st_size, low_bound + CHKSM_CHUNK_SIZE);

So be careful.

Snorch (approved):
Looks good. But lacks zdtm tests.

criu/files-reg.c
0x7f454c46: What I meant was:
```
up_bound = min_t(size_t, fd_status->st_size, low_bound + CHKSM_CHUNK_SIZE);
if (opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_FIRST)
     up_bound = min_t(up_bound, opts.file_validation_chksm_parameter);
```
and some update inside the loop on remap.

criu/include/cr_options.h
```diff
@@ -75,12 +75,48 @@ enum FILE_VALIDATION_OPTIONS
...
+	 * This constant indicates that the file validation should be done with the
+	 * checksum method using the first N (The checksum parameter) bytes of the file.
+	 */
+	FILE_VALIDATION_CHKSM_FIRST,
+
+	/*
+	 * This constant indicates that the file validation should be done with the
+	 * checksum method using every Nth (The checksum parameter) bytes of the file.
+	 */
+	FILE_VALIDATION_CHKSM_PERIOD
```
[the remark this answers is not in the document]
AnorexicAtticusFinch: Excessive commenting is just force of habit from my course projects. I've changed the comments and changed FILE_VALIDATION_CHKSM_EVERY to FILE_VALIDATION_CHKSM_FULL (Since the argument for it is checksum-full).

criu/files-reg.c
AnorexicAtticusFinch: Yeah, you're right. I've removed build-id as a fallback for checksum.

AnorexicAtticusFinch:
> Looks good. But lacks zdtm tests.

I plan on pushing zdtm tests for build-id and checksum through a separate PR since I have some doubts regarding them.

criu/files-reg.c
AnorexicAtticusFinch: I've moved some sections of code around, hopefully it's easier to read now. I wrote it this way to make it really easy to add new configurations.
To do so, only the `init, next and stop` functions need to be altered without worrying about the rest of code (Like mapping the next region of the file).
Even if the code for a new configuration is faulty, it won't crash since the iterator is validated outside those functions before accessing memory.

Snorch:
>> Looks good. But lacks zdtm tests.
> I plan on pushing zdtm tests for build-id and checksum through a separate PR since I have some doubts regarding them.

ZDTM tests are not only intended to test that you functionality will not break in future. But also to help _review_ your functionality, so it _should go along_ with your functionality. This way everybody can see (by running test on their own and experimenting with them) that you functionality works. We skip tests in some rare cases, but in this case I think we should have them.

@avagin Please correct me if I'm wrong.

AnorexicAtticusFinch:
> ZDTM tests are not only intended to test that you functionality will not break in future. But also to help _review_ your functionality, so it _should go along_ with your functionality. This way everybody can see (by running test on their own and experimenting with them) that you functionality works. We skip tests in some rare cases, but in this case I think we should have them.

Oh, well in that case I've pushed the tests but they're not complete.
They work and everything is done except that in the tests that should fail during restore, once restore fails the test is automatically considered a failure.
The `crfail` flag is only for tests that should fail during dumping, I couldn't find any such flag for failure during restore. I'm not sure how to proceed from here.

Snorch:
> The `crfail` flag is only for tests that should fail during dumping,

Maybe it's time to add new flag for fail on dump expected then.

AnorexicAtticusFinch:
@avagin @0x7f454c46 The added tests ran and passed on Travis yet this is the result:

    ################### 14 TEST(S) FAILED (TOTAL 408/SKIPPED 35) ###################
     * zdtm/static/file-validation-chksm05.c(h)
     * zdtm/static/file-validation-buildid03.c(h)
     * zdtm/static/file-validation-chksm02.desc(h)
     * zdtm/static/file-validation-chksm03.c(h)
     * zdtm/static/file-validation-buildid02.c(h)
     * zdtm/static/file-validation-buildid00.c(h)
     * zdtm/static/file-validation-buildid01.c(h)
     * zdtm/static/file-validation-chksm02.c(h)
     * zdtm/static/file-validation-chksm01.c(h)
     * zdtm/static/file-validation-chksm01.desc(h)
     * zdtm/static/file-validation-chksm04.desc(h)
     * zdtm/static/file-validation-chksm04.c(h)
     * zdtm/static/file-validation-chksm03.desc(h)
     * zdtm/static/file-validation-chksm00.c(h)

Have I missed something? Why is this happening?

0x7f454c46:
Probably, some of the files you've added have executable bit set?

AnorexicAtticusFinch:
> Probably, some of the files you've added have executable bit set?

Yeah, that was the reason. I've re-uploaded the files. Thanks!

criu/files-reg.c
```diff
@@ -1581,6 +1587,110 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+				u32 *checksum)
+{
+	int tmp;
+	u32 byte, mask;
+	u64 up_bound, low_bound = 0, i = 0;
+	unsigned char *file_header;
+
+	/* At most, the first 10MB of the file is mapped. */
+	up_bound = min_t(size_t, fd_status->st_size, CHKSM_CHUNK_SIZE);
+	file_header = (unsigned char *) mmap(0, up_bound,
```
avagin, line 1643: The typecasting isn't needed here.

criu/files-reg.c
```diff
@@ -1581,6 +1587,110 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+		*iter = 0;
+	else if (opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_FIRST)
+		*iter = 0;
+	else if (opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_PERIOD)
+		*iter = 0;
+}
+
+static inline void calculate_checksum_iterator_next(u64 *iter)
+{
+	if (opts.file_validation_chksm_config == FILE_VALIDATION_CHKSM_FULL)
```
avagin, line 1602: use switch instead of multiple if-s

criu/files-reg.c
```diff
@@ -1581,6 +1587,110 @@ static int get_build_id(const int fd, const struct stat *fd_status,
 	return ret;
 }
 
+static inline void calculate_checksum_iterator_init(u64 *iter)
```
avagin, line 1590: The name of this function is too long. It is used just in this file, so you can use a shorter name. The code will be more readable.

criu/files-reg.c
```diff
@@ -1581,6 +1587,110 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+		/*
+		 * If the new iterator position is outside the mapped region (Bytes in the range
+		 * low_bound and up_bound) of the file, the current region needs to be unmapped
+		 * and the next portion (At most the next 10MB) of the file needs to be mapped.
+		 */
+		if (i >= up_bound && up_bound != fd_status->st_size && i < fd_status->st_size) {
+			munmap(file_header, up_bound - low_bound);
+			low_bound = round_down(i, sysconf(_SC_PAGE_SIZE));
+			up_bound = min_t(size_t, fd_status->st_size, low_bound + CHKSM_CHUNK_SIZE);
+			file_header = (unsigned char *) mmap(0, up_bound - low_bound,
```
avagin, line 1680: This is not a file header. I think you need to rename it to "buf" or something like this.

criu/files-reg.c
```diff
@@ -1629,6 +1739,47 @@ static int store_validation_data_build_id(RegFileEntry *rfe, int lfd,
...
+	if (fd < 0) {
+		pr_err("Checksum (For validation) could not be obtained for file %s because can't open the file\n",
+				rfe->name);
+		return -1;
+	}
+
+	if (!calculate_checksum(fd, &(p->stat), &checksum)) {
+		close(fd);
+		pr_info("Could not obtain checksum for file %s\n", rfe->name);
+		return 0;
```
avagin, line 1767: Why do we not report an error in this case?

criu/files-reg.c
```diff
@@ -1581,6 +1587,110 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+	*checksum = 0xFFFFFFFF;
+	calculate_checksum_iterator_init(&i);
+
+	while (!calculate_checksum_iterator_stop_condition(i) &&
+			i >= low_bound && i < up_bound) {
+
+		byte = file_header[i-low_bound];
+		*checksum = *checksum ^ byte;
+		for (tmp = 0; tmp < 8; tmp++) {
+			mask = -(*checksum & 1);
```
avagin, line 1659: How fast is this implementation of crc32? Have you looked at how it is implemented in the Linux kernel or other projects?

criu/files-reg.c
```diff
@@ -1581,6 +1587,110 @@ static int get_build_id(const int fd, const struct stat *fd_status,
...
+	up_bound = min_t(size_t, fd_status->st_size, CHKSM_CHUNK_SIZE);
+	file_header = (unsigned char *) mmap(0, up_bound,
+						PROT_READ, MAP_PRIVATE | MAP_FILE, fd, 0);
+	if (file_header == MAP_FAILED) {
+		pr_warn("Couldn't mmap file with fd %d", fd);
+		return false;
+	}
+
+	*checksum = 0xFFFFFFFF;
+	calculate_checksum_iterator_init(&i);
```
avagin, line 1651: IMHO: "i" isn't a good name here. i is an offset in a file, so I would use off or offset in this case.

criu/files-reg.c
AnorexicAtticusFinch: I thought it would be better to inform the user that the checksum couldn't be obtained and only file size validation is being used for a particular file and proceed, instead of throwing an error and stopping. Should I change this?

criu/files-reg.c
AnorexicAtticusFinch: I don't think you can make it any faster (Unrolling the for loop would help a bit I guess) without using a lookup table, which is what the Linux kernel does.

avagin: why do we not use the lookup table?

criu/files-reg.c
avagin: Yes, I think you should. If a user asks to check a checksum, they want to be sure that we will do this.

criu/files-reg.c
AnorexicAtticusFinch: Since the checksum method is more like a fallback for the build-id method, I thought it seemed unnecessary. The lookup table is a necessity for the Linux kernel since it has to run for every network packet sent and received.

Regardless of this, I've mentioned it under "Possible Improvements" in the document I sent you for the wiki page.

avagin: Do we do validations for all files (not only mapped files)? I think the answer is yes and it isn't just a fallback in this case. Right now, this checksum code is very slow. We decided that you don't need to fix it right now. If you will have a wish, you can rework it after the GSOC program.

avagin:
Can you move file validation helpers in a separate file? files-reg.c is too big already.

AnorexicAtticusFinch:
> Can you move file validation helpers in a separate file? files-reg.c is too big already.

Yeah, I've moved the functions to files-validation.c.

xiongzubiao:
If I understand correctly, this change is not merged in. Anyone knows why? CRIU's wiki has it though... https://criu.org/index.php?title=Validate_files_on_restore

Snorch:
@xiongzubiao @avagin @checkpoint-restore/maintainers

The PR has unresolved conversation https://github.com/checkpoint-restore/criu/pull/1148#discussion_r472275419 about "how fast is the algorithm?", it seem to only work fast for files with BUILDID, checksumming is not so fast obviously.

So it seems we need to update wiki that this feature is not available.

Issue #1149, opened 2020-07-22
CRIU can't find the build-id note


avagin:
```
$ python test/zdtm.py run -t zdtm/static/env00  -f h
=== Run 1/1 ================ zdtm/static/env00
========================== Run zdtm/static/env00 in h ==========================
Start test
./env00 --pidfile=env00.pid --outfile=env00.out --envname=ENV_00_TEST
Run criu dump
=[log]=> dump/zdtm/static/env00/44/1/dump.log
------------------------ grep Error ------------------------
b'(00.000000) vma 405000 borrows vfi from previous 402000'
b'(00.000000) vma 407000 borrows vfi from previous 405000'
b'(00.000000) vma 408000 borrows vfi from previous 407000'
b'(00.000000) Dumping path for -3 fd via self 12 [/usr/lib64/libc-2.30.so]'
b"(00.000000) Warn  (criu/files-reg.c:1500): Couldn't find the build-id note for file with fd 15"
b'(00.000000) vma 7efe98f68000 borrows vfi from previous 7efe98f1e000'
b'(00.000000) vma 7efe98f69000 borrows vfi from previous 7efe98f68000'
b'(00.000000) vma 7efe98f6c000 borrows vfi from previous 7efe98f69000'
b'(00.000000) Dumping path for -3 fd via self 12 [/usr/lib64/ld-2.30.so]'
b"(00.000000) Warn  (criu/files-reg.c:1500): Couldn't find the build-id note for file with fd 15"
------------------------ ERROR OVER ------------------------
Run criu restore
Send the 15 signal to  44
Wait for zdtm/static/env00(44) to die for 0.100000
Removing dump/zdtm/static/env00/44
========================= Test zdtm/static/env00 PASS ==========================
```

```
# readelf -a /usr/lib64/ld-2.30.so | grep -B 10 -A 10 gnu.build
  Number of section headers:         30
  Section header string table index: 29

Section Headers:
  [Nr] Name              Type             Address           Offset
       Size              EntSize          Flags  Link  Info  Align
  [ 0]                   NULL             0000000000000000  00000000
       0000000000000000  0000000000000000           0     0     0
  [ 1] .note.gnu.propert NOTE             00000000000002a8  000002a8
       0000000000000020  0000000000000000   A       0     0     8
  [ 2] .note.gnu.build-i NOTE             00000000000002c8  000002c8
       0000000000000024  0000000000000000   A       0     0     4
[...]

PR #1151, opened 2020-07-24
Bug fixes for the build-ID validation method

AnorexicAtticusFinch:
This addresses [issue #1149](https://github.com/checkpoint-restore/criu/issues/1149). Even if the ELF file contains multiple PT_NOTE program headers, only the first header was being checked for the build-id.
Also addresses a bug where the RegFileEntry n_build_id field would be populated with incorrect data, causing dump to crash at times (Sometimes affected job 7 on Travis).

Mentors: @avagin @0x7f454c46

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -1358,7 +1358,7 @@ static int get_build_id_32(Elf32_Ehdr *file_header, unsigned char **build_id,
 	int size, num_iterations;
 	size_t file_header_end;
 	Elf32_Phdr *program_header, *program_header_end;
-	Elf32_Nhdr *note_header, *note_header_end;
+	Elf32_Nhdr *note_header_end, *note_header = NULL;
 
 	file_header_end = (size_t) file_header + mapped_size;
 	if (sizeof(Elf32_Ehdr) > mapped_size)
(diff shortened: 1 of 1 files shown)


AnorexicAtticusFinch:
@avagin Should I keep the fix as a separate commit or should I squash it with the "add build-id validation functionality" commit and force-push?

criu/files-reg.c
```diff
@@ -1384,30 +1384,42 @@ static int get_build_id_32(Elf32_Ehdr *file_header, unsigned char **build_id,
...
-		note_header = (Elf32_Nhdr *) ((char *) note_header + sizeof(Elf32_Nhdr) +
-						ALIGN(note_header->n_namesz, 4) +
-						ALIGN(note_header->n_descsz, 4));
+		/* The note type for the build-id is NT_GNU_BUILD_ID. */
+		while (note_header <= note_header_end && note_header->n_type != NT_GNU_BUILD_ID)
+			note_header = (Elf32_Nhdr *) ((char *) note_header + sizeof(Elf32_Nhdr) +
+							ALIGN(note_header->n_namesz, 4) +
+							ALIGN(note_header->n_descsz, 4));
+		
+		if (note_header >= note_header_end) {
```
avagin, line 1410: note_header = note_header_end is valid, isn't it?

criu/files-reg.c
```diff
@@ -1384,30 +1384,42 @@ static int get_build_id_32(Elf32_Ehdr *file_header, unsigned char **build_id,
...
+			program_header++;
+			note_header = NULL;
+			continue;
+		}
+	}
```
avagin, line 1415: you will go to the next loop iteration with the same  program_header. I think you need yo add break right before this line.

criu/files-reg.c
```diff
@@ -1384,30 +1384,42 @@ static int get_build_id_32(Elf32_Ehdr *file_header, unsigned char **build_id,
...
 
-	if (note_header >= note_header_end) {
+	if (!num_iterations || program_header >= program_header_end) {
```
avagin, line 1417: program_header = program_header_end is valid
num_iterations = 0 is valid

criu/files-reg.c
AnorexicAtticusFinch: Yes that's valid. I've made the change.

criu/files-reg.c
AnorexicAtticusFinch: I've added the break statement.

criu/files-reg.c
AnorexicAtticusFinch: I've made the change for program_header.
`num_iterations = 0` is invalid since num_iterations is initialized to `file_header->e_phnum + 1`.

criu/files-reg.c
```diff
@@ -1384,30 +1384,43 @@ static int get_build_id_32(Elf32_Ehdr *file_header, unsigned char **build_id,
...
 	 */
-	while (num_iterations-- && program_header <= program_header_end &&
-			program_header->p_type != PT_NOTE)
-		program_header++;
-
-	if (!num_iterations || program_header >= program_header_end) {
-		pr_warn("Couldn't find the note program header for file with fd %d\n", fd);
-		return -1;
-	}
+	while (num_iterations-- && program_header <= program_header_end) {
```
avagin, line 1387: If e_phnum is 1, this loop will do two cycles, what is wrong, isn't it?

```
avagin@avagin-glaptop:~$ cat test.c
#include <stdio.h>
#include <unistd.h>

int main()
{
	int num_iters = 2;

	while (num_iters--) {
		printf("%d\n", num_iters);
	}
	return 0;
}
avagin@avagin-glaptop:~$ gcc test.c
avagin@avagin-glaptop:~$ ./a.out 
1
0
```

criu/files-reg.c
avagin: hmmm. In this case, while (num_iterations-- ...) looks wrong. This code is not easy to read. Maybe it is better to rewrite it like this:

```
for (num_iteration = 0; num_interation < file_header->e_phnum; num_iteration++, program_header++) {
  if program_header > program_header_end {
    break;
  }
  if (program_header->p_type != PT_NOTE)
    continue;
  ....
}
```

AnorexicAtticusFinch: I've made the change.

criu/files-reg.c
AnorexicAtticusFinch: Yeah, you're right. I've changed this.

criu/files-reg.c
```diff
@@ -1604,15 +1614,16 @@ static int store_validation_data_build_id(RegFileEntry *rfe, int lfd,
 	if (!build_id || build_id_size == -1)
 		return 0;
 
-	rfe->build_id = xmalloc(round_up(build_id_size, sizeof(uint32_t)));
+	allocated_size = round_up(build_id_size, sizeof(uint32_t));
+	rfe->build_id = xmalloc(allocated_size);
```
avagin, line 1618: should we use xzalloc here? allocated_size can be bigger than build_id_size, but memcpy is called for build_id_size only.

avagin:
> Also addresses a bug where the RegFileEntry n_build_id field would be populated with incorrect data, causing dump to crash at times.

You need to create two separate commits for both fixes.

criu/files-reg.c
AnorexicAtticusFinch: Yeah, you're right. I've made the change.

PR #1152, opened 2020-07-24
Reserve images fields for new virtuozzo features

Snorch:
We've prepared several CRIU restore improvements for Virtuozzo criu to make it possible dumping Virtuozzo containers with Docker containers inside in our next update. So to maintain images format compatibility between mscriu and vzcriu lets reserve field ids in images.

### DIFF
--- a/images/pstree.proto
+++ b/images/pstree.proto
@@ -6,4 +6,11 @@ message pstree_entry {
 	required uint32			pgid		= 3;
 	required uint32			sid		= 4;
 	repeated uint32			threads		= 5;
+/*
+ * These fields are reserved for Virtuozzo criu nested pidns feature
+ *	repeated uint32			ns_pid		= 6;
+ *	repeated uint32			ns_pgid		= 7;
+ *	repeated uint32			ns_sid		= 8;
+ *	repeated ns_tid			tids		= 9;
+ */
 }
--- a/images/core.proto
+++ b/images/core.proto
@@ -73,6 +73,10 @@ message task_kobj_ids_entry {
 	optional uint32			user_ns_id	= 10;
 	optional uint32			cgroup_ns_id	= 11;
 	optional uint32			time_ns_id	= 12;
+/*
+ * These field is reserved for Virtuozzo criu nested pidns feature
+ *	optional uint32			pid_for_children_ns_id	= 13;
+ */
 }
 
 message thread_sas_entry {
--- a/images/mnt.proto
+++ b/images/mnt.proto
@@ -30,6 +30,10 @@ enum fstype {
 	// NFS4			= 22;
 
 	CGROUP2			= 23;
+/*
+ * These field is reserved for Virtuozzo criu new mounts engine
+ * 	NSFS			= 24;
+ */
 };
 
 message mnt_entry {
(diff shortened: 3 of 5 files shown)


rppt:
> We've prepared several CRIU restore improvements for Virtuozzo criu to make it possible dumping Virtuozzo containers with > Docker containers inside in our next update. So to maintain images format compatibility between mscriu and vzcriu lets reserve field ids in images.

`mscriu` as in `MS-Word`? ;-)

A general comment for all patches, I think it'll look better if you use "future feature" rather than "virtuozzo feature" in the hope `mscriu` will see these features at some point.

Snorch:
>mscriu as in MS-Word? ;-)

The "mscriu" stants for "Mainstream CRIU" similar to "vzcriu" stants for "Virtuozzo CRIU".

> I think it'll look better if you use "future feature" rather than "virtuozzo feature" in the hope mscriu will see these features at some point.

It can cause confusion, let's better leave them as Virtuozzo criu features and change it when/if features will be ported.

images/pstree.proto
```diff
@@ -6,4 +6,11 @@ message pstree_entry {
 	required uint32			pgid		= 3;
 	required uint32			sid		= 4;
 	repeated uint32			threads		= 5;
+/*
+ * These fields are reserved for Virtuozzo criu nested pidns feature
+ *	repeated uint32			ns_pid		= 6;
+ *	repeated uint32			ns_pgid		= 7;
+ *	repeated uint32			ns_sid		= 8;
+ *	repeated ns_tid			tids		= 9;
+ */
```
0x7f454c46, line 15: I don't mind, but still wondering - can you use in vz-criu some upper number instead?
Like:
```
 	repeated uint32			ns_pid		= 1021;
 	repeated uint32			ns_pgid		= 1022;
 	repeated uint32			ns_sid		= 1023;
 	repeated ns_tid			tids		= 1024;
```
That way you wouldn't need to reserve any fields..

avagin: +1

Snorch: If we port one vzcriu feature to mainstream (e.g. multiple unix socket bindmounts) we will have to port those "upper" number ids to mainstream too (so that vzcriu can be re-based on top).

If such a big numbers are fine in mainstream criu and nobody exept Virtuozzo will use them we can follow this convention.

adrianreber: Using higher numbers for vzcriu features sounds indeed like a good idea.

> If we port one vzcriu feature to mainstream (e.g. multiple unix socket bindmounts) we will have to port those "upper" number ids to mainstream too (so that vzcriu can be re-based on top).

I do not think this is a big problem. The protobuf numbers do not have to consecutive. We should probably have somewhere a hint why there are high protobuf field numbers.

mihalicyn: We doing all possible to port all features from vzcriu to main the main CRIU project. The idea to make all Virtuozzo-specific image fields numbers start from `1000` is good as for me. But still remains the problem with names of fields in protobuf. Because we already meet name collisions in the past (`mnt_id` in `sk-unix.proto` as example). And I think that even after we start numbering from `1000` in any way we will have to prepare reservations for new field names. Is it OK?

xemul:
This is getting a bit messy. Gents, I propose to introduce a "vendor" sub-object for ... at least those that need it (but probably should be for all images). Like this

````
message pstree_entry {
        required uint32                 pid             = 1;
        required uint32                 ppid            = 2;
        required uint32                 pgid            = 3;
        required uint32                 sid             = 4;
        repeated uint32                 threads         = 5;
        optional vendor_pstree_entry vendor = 1024;
}

````
the mainstream criu should keep all vendor stuff empty, all vendos may patch them internally the way they like. Hopefully protonuf doesn't care the gaps in field IDs.

Snorch:
Updated PR https://github.com/checkpoint-restore/criu/pull/1174 with vendor image fields.

PR #1155, opened 2020-07-25
 Allow CRIU to be used as non-root with CAP_CHECKPOINT_RESTORE or CAP_SYS_ADMIN

adrianreber:
With the almost finished discussion about CAP_CHECKPOINT_RESTORE I reworked my test code to be ready for review.

This contains the changes to run CRIU as non-root with either CAP_SYS_ADMIN or CAP_CHECKPOINT_RESTORE. As Travis does not yet support CAP_CHECKPOINT_RESTORE I am using CAP_SYS_ADMIN during Travis testing, but locally it works just the same with CAP_CHECKPOINT_RESTORE.

This PR checks if the euid is non-zero and if it is non-zero it checks if CRIU has either CAP_SYS_ADMIN or CAP_CHECKPOINT_RESTORE. If that is true CRIU remembers the euid and decides at different places in the code to not do certain things which do not work as non-root.

This is a minimal start of non-root CRIU and it can be used with zdtm to test with env00. As seen in the patch.

This is based on @nviennot and my test branches for the lkml CAP_CHECKPOINT_RESTORE discussions.


test/zdtm.py
```diff
@@ -2604,6 +2620,9 @@ def clean_stuff(opts):
                 help="Use splice or read mode of pre-dumping",
                 choices=['splice', 'read'],
                 default='splice')
+rp.add_argument("--non-root",
```
avagin, line 2623: we already have the --user option. How is this one different?

adrianreber: Not at all. I was not aware of `--user`. I have updated this PR to use `--user` instead of `--non-root`.

nviennot: Can we rename --user to --non-root? user could mean anything. (criU, user namespaces). Not at all obvious of what it would do. --unprivileged is pretty excplicit too. I'd argue that Adrian might have seen the option if the naming of the option was unambiguous. Our users might miss it too.

test/zdtm/lib/test.c
```diff
@@ -238,34 +241,36 @@ void test_init(int argc, char **argv)
...
-		tok = strtok(val, " ");
-		while (tok) {
-			size++;
-			groups[size - 1] = atoi(tok);
-			tok = strtok(NULL, " ");
+	if (!uid) {
+		val = getenv("ZDTM_GROUPS");
+		if (val) {
+			char *tok = NULL;
+			unsigned int size = 0, groups[NGROUPS_MAX];
```
mihalicyn, line 248: nit: `NGROUPS_MAX` is `65536`
Isn't it will be better to allocate memory from heap?

adrianreber: This is just me moving code behind an `if` if running as non-root. I have not changed the actual code. If this should be changed it should happen in another PR.

test/zdtm.py
adrianreber: What is the history behind `--user`? I see now that is has been used in one of the jenkins jobs which I am not familiar with.

Is `--user` already testing exactly the same I wanted to test with this PR? Running CRIU with CAP_CHECKPOINT_RESTORE/CAP_SYS_ADMIN?

test/zdtm/lib/test.c
mihalicyn: > This is just me moving code behind an `if` if running as non-root. I have not changed the actual code. If this should be changed it should happen in another PR.

no problem. It's just comment.

test/zdtm_ct.c
```diff
@@ -71,46 +71,52 @@ static int create_timens()
 
 int main(int argc, char **argv)
 {
+	uid_t uid;
 	pid_t pid;
 	int status;
 
+	uid = getuid();
```
rst0git, line 78: `geteuid()`?

adrianreber:
>     1. We need solid testing without CAP_SYS_ADMIN, where we run with just with CAP_CHECKPOINT_RESTORE and CAP_PTRACE, as advertised. Otherwise, we can't develop properly. This may require us to use a custom kernel (could be compiled and put as a binary blob in some repo we own) that we run with qemu.

Thanks to @brauner the CAP_CHECKPOINT_RESTORE changes have been merged and should be available with RC1 (https://github.com/torvalds/linux/commit/74858abbb1032222f922487fd1a24513bbed80f9). As we are already using Vagrant to test the latest Fedora release we can easily add another Vagrant based test and install the latest RC1 kernel via https://fedoraproject.org/wiki/Kernel_Vanilla_Repositories which provides all RC releases:

https://repos.fedorapeople.org/repos/thl/kernel-vanilla-mainline/fedora-32/x86_64/

>     2. CLI options: should we require users to manually disable features (like network), or automatically disable features when we see we don't have the right capability. Another way to look at this is to be tolerant as much as we can to EPERM errors.

Any reason to not automatically disable features if we already detect that it will not work?

>     3. CRIU should not look at uid. Rather, it should look at its effective capabilities (and perhaps whether it is running in a user namespace, as it makes all the `capable(CAP_*)` tests fail in the kernel, which matters for things like `SO_SNDBUFFORCE`).

Sounds good. Like you mentioned above looking at `/proc/self/status` sounds like a good approach.

>     4. At some point, we should motivate the docker/kubernetes people to add CAP_CHECKPOINT_RESTORE in the set of bounded capabilities (so users don't have to add it manually).
> 
> 
> I can help

Thanks. In your review you mentioned a lot of different capabilities we should test and this sounds like a good approach. I would start with reading `/proc/self/status` and only focus on CAP_CHECKPOINT_RESTORE for now before looking at what other capabilities are needed. This could be done in follow up pull requests.

nviennot:
1. Wonderful

> Any reason to not automatically disable features if we already detect that it will not work?

Sometimes, failing can be preferable:
1) Let say that one is using CRIU for some time with some settings that they like in a managed kubernetes environment.
2) The hosting provider changes the kubernetes environment, changing the default capabilities that a container has
3) CRIU will then skip certain things to checkpoint (or restore) as a result. However, the user will not know about the change of behavior. This can potentially lead to an silent application failure.

So I'd argue that we want a deterministic behavior given a set of CLI options.

I see two approaches:
a) We can skip individual features: So we can throw errors like "CAP_SYS_ADMIN is not detected. Please retry with --skip-lsm --skip-X -skip-Y" where X and Y are features that need CAP_SYS_ADMIN to work.
b) We can supply the list of capabilities that should not be used by CRIU via the CLI, e.g. "--no-cap-sys-admin". With the list of caps that we have and don't have, we can disables all non-compatible features on boot.

b) is a superset of a) and would be easy to use for users. Most importantly, the behavior of CRIU is deterministic given a list of CLI options.

criu/cr-check.c
```diff
@@ -1551,3 +1553,125 @@ static char *feature_name(int (*func)(void))
...
+	int ret = prctl(PR_SET_DUMPABLE, value, 0, 0, 0);
+	if (ret < 0)
+		pr_perror("Unable to set PR_SET_DUMPABLE");
+	return ret;
+}
+
+int check_uid(void)
+{
+	const char *attribute = "security.capability";
+	const char *criu = "/proc/self/exe";
```
brauner, line 1595: I have to lol at this. Didn't we just argue that it's a bad idea to rely on exe links. :D

criu/cr-check.c
```diff
@@ -1551,3 +1553,125 @@ static char *feature_name(int (*func)(void))
...
+
+	if (uid == 0)
+		return 0;
+
+	/*
+	 * If running as non-root CRIU either needs CAP_SYS_ADMIN or
+	 * CAP_CHECKPOINT_RESTORE and CAP_PTRACE.
+	 */
+
+	fd = open(criu, O_RDONLY);
```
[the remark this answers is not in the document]
brauner: criu already should have a helper to test whether or not it is running a user namespace. If not, what should work is sm like:
```c
inline static bool in_userns(void)
{
	__do_fclose FILE *f = NULL;
	uid_t user, host, count;
	int ret;

	/* Now: are we in a user namespace? Because then we're also
	 * unprivileged.
	 */
	f = fopen("/proc/self/uid_map", "re");
	if (!f)
		return false;

	ret = fscanf(f, "%u %u %u", &user, &host, &count);
	if (ret != 3)
		return false;

	return user != 0 || host != 0 || count != UINT32_MAX;
}
```

adrianreber:
I created a vagrant based test using a 5.8.0 kernel for now. Once the Fedora vanilla kernel mainline repository picks up 5.9.0-rc1 the test should work. Right now I added the test more as a proof of concept. To see how it could look like.

https://travis-ci.org/github/checkpoint-restore/criu/jobs/715943265

criu/fdstore.c
```diff
@@ -46,8 +49,16 @@ int fdstore_init(void)
 		return -1;
 	}
 
-	if (setsockopt(sk, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0], sizeof(buf[0])) < 0 ||
-	    setsockopt(sk, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1], sizeof(buf[1])) < 0) {
+	if (opts.uid) {
```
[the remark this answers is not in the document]
avagin: I agree that we need to check CAP_NET_ADMIN here. uid can be 0 but setsockopt(SO_RCVBUFFORCE) will fail without CAP_NET_ADMIN.

criu/kerndat.c
```diff
@@ -829,6 +829,9 @@ static int kerndat_try_load_cache(void)
 {
 	int fd, ret;
 
+	if (opts.uid)
```
avagin, line 832: pls add a comment why we can't load cache in this case.

criu/kerndat.c
```diff
@@ -1065,7 +1071,8 @@ int kerndat_init(void)
 	memset(&kdat, 0, sizeof(kdat));
 
 	preload_socket_modules();
-	preload_netfilter_modules();
+	if (!opts.uid)
+		preload_netfilter_modules();
```
avagin, line 1075: CAP_SYS_MODULE or CAP_NET_ADMIN

criu/namespaces.c
```diff
@@ -1591,10 +1591,12 @@ int collect_namespaces(bool for_dump)
 
 int prepare_userns_creds(void)
 {
-	/* UID and GID must be set after restoring /proc/PID/{uid,gid}_maps */
-	if (setuid(0) || setgid(0) || setgroups(0, NULL)) {
-		pr_perror("Unable to initialize id-s");
-		return -1;
+	if (!opts.uid) {
```
avagin, line 1594: CAP_SETUID

adrianreber:
Test is now running on Fedora with 5.9.0-rc0 kernel and CAP_CHECKPOINT_RESTORE

criu/kerndat.c
adrianreber: This actually fails with
```
openat(AT_FDCWD, "/run/xtables.lock", O_RDONLY|O_CREAT, 0600) = -1 EACCES (Permission denied)
```
so checking for uid 0 seems to be the best approach.

adrianreber:
I rebased this and pushed a new version in which I tried to switch to using /proc/pid/status.

adrianreber:
Pushed again to fix a CI error with usernsd.

criu/kerndat.c
avagin: Will modules be loaded without CAP_SYS_MODULE or CAP_NET_ADMIN?

adrianreber: No. On my system `preload_netfilter_modules()` runs iptables and iptables will fail while trying to access `/run/xtables.lock`. I do not think any capability will make this work.

adrianreber:
@nviennot Any comments from your side to the current patches?

criu/util.c
```diff
@@ -1334,6 +1334,9 @@ void rlimit_unlimit_nofile(void)
 {
 	struct rlimit new;
 
+	if (opts.uid)
+		return;
+
```
rst0git, line 1339: This change will cause `pre-dump` to fail with the following error for processes with larger (e.g., 1GB) memory state.  
```
Error (criu/page-pipe.c:120): page-pipe: Can't make pipe for page-pipe: Too many open files
```

rst0git:
@adrianreber running `pre-dump` as non root user fails with:
```
(00.023457) Error (criu/mem.c:44): BUG at criu/mem.c:44
```

adrianreber:
@rst0git What does happen if you run pre-dump as non-root before my patches? I am aware that my patches are only touching a small part of what needs to be done to have the complete code base ready for CAP_CHECKPOINT_RESTORE. But it at least provides a working checkpoint and restore implementation as non-root.

I would prefer an incremental approach on the `criu-dev` branch. If the patches in this PR are correct I would rather see this merged and then look at the `pre-dump` use case.

rst0git (approved):

rst0git:
> @rst0git What does happen if you run pre-dump as non-root before my patches?

@adrianreber I can confirm that the same error occurs before applying the patches in this PR and I agree with you that incremental approach on the `criu-dev` branch would be better.

nviennot:
I'll take a look next week

avagin:
@adrianreber @nviennot l What is going on with this PR. Let's make it ready for the next round of reviews.

scripts/ci/vagrant.sh
```diff
@@ -49,4 +49,15 @@ fedora-no-vdso() {
...
+	ssh default 'curl -s https://repos.fedorapeople.org/repos/thl/kernel-vanilla.repo | sudo tee /etc/yum.repos.d/kernel-vanilla.repo'
+	ssh default 'sudo dnf -y --enablerepo=kernel-vanilla-mainline update kernel-core; sudo dnf -y install glibc-static'
+	vagrant reload
+	ssh default uname -a
+	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; make -j 4'
+	ssh default 'cd /vagrant; git clone https://git.kernel.org/pub/scm/libs/libcap/libcap.git; cd libcap; make -C progs'
+	ssh default 'sudo /vagrant/libcap/progs/setcap cap_checkpoint_restore+eip /vagrant/criu/criu/criu'
+	ssh default 'cd /vagrant/criu; sudo ./test/zdtm.py run -t zdtm/static/env00 -f h --ignore-taint; sudo chmod 777 test test/zdtm/static/ test/dump/zdtm/static/env00; sudo ./test/zdtm.py run -t zdtm/static/env00 -f h --user --ignore-taint'
+}
+
```
rst0git, line 62: @adrianreber the Vagrant tests are currently not running in CI and it looks like the Ubuntu GitHub actions environment does not support nested virtualization (https://github.com/actions/virtual-environments/issues/183). However, it is possible to use macOS to run Vagrant VM (e.g. https://github.com/luxonis/depthai-docs-website/commit/f24e690c13c9d2fa5880b60fac2e327a527589d9).

Also, the Fedora 33 kernel provides CAP_CHECKPOINT_RESTORE.

adrianreber: I have working vagrant using cirrus ci: #1272

nviennot:
Sorry, I'm swamped right now. I didn't find time to look at this. Hopefully this month I'll make some progress. Sorry.

PavloMykhailyshyn:
People are waiting for this feature to arrive as soon as possible. It is very important to be able to run the CRIU as a non-root user.
@nviennot please take into considaration.

nviennot:
Yes, I'm sorry I've been quite unresponsive on this issue. I'm working on clearing some space for tackling this.

avagin:
@adrianreber could you rebase this pr and close comments that have been resolved. Thanks!

adrianreber:
So, the non-root test works, but is seems I broke a couple of other things. Need to investigate.

adrianreber:
I updated the PR and now all tests are green. I also added two zdtm test cases (env00 and pthread00) to be tested as non-root.

This is still a limited implementation of non-root CRIU but it works (for the two tests at least). It is a proof of concept.

I tried to use capabilities were possible. Unfortunately there are a couple of places where it does not make sense to check for capabilities. Right now the kerndat cache file is not written at all. We would need to probably just write it to the user's home-directory if `uid != 0`.

criu/cgroup.c
```diff
@@ -732,6 +732,9 @@ int dump_task_cgroup(struct pstree_item *item, u32 *cg_id, struct parasite_dump_
 	unsigned int n_ctls = 0;
 	struct cg_set *cs;
 
+	if (opts.uid)
```
rppt, line 735: Maybe add a wrapper like `has_caps_*()` for this test as well?

criu/cr-restore.c
```diff
@@ -1809,6 +1810,7 @@ static int restore_task_with_children(void *_arg)
 		current->pid->real = atoi(buf);
 		pr_debug("PID: real %d virt %d\n",
 				current->pid->real, vpid(current));
+
```
rppt, line 1813: Does not seem related ;-)

criu/kerndat.c
```diff
@@ -1099,10 +1119,32 @@ int kerndat_init(void)
...
+			ret = -1;
+		}
+		if (!ret && kerndat_link_nsid()) {
+			pr_err("kerndat_link_nsid failed when initializing kerndat.\n");
+			ret = -1;
+		}
+		if (!ret && kerndat_socket_netns()) {
+			pr_err("kerndat_socket_netns failed when initializing kerndat.\n");
+			ret = -1;
+		}
```
rppt, line 1146: I think it's worth splitting the tests that require root into a helper function

criu/pie/util.c
```diff
@@ -86,3 +86,18 @@ bool has_cap_checkpoint_restore(u32 *cap_eff)
...
+
+bool has_cap_sys_chroot(u32 *cap_eff)
+{
+	return has_capability(CAP_SYS_CHROOT, cap_eff);
+}
+
+bool has_cap_setuid(u32 *cap_eff)
+{
+	return has_capability(CAP_SETUID, cap_eff);
+}
```
rppt, line 103: Can those be inline?

rppt:
> Right now the kerndat cache file is not written at all. We would need to probably just write it to the user's home-directory if `uid != 0`.

Well, `~/.criu.kdat ` does not sound bad to me.

criu/cgroup.c
adrianreber: What do you exactly mean? A capability check for CGROUP or a functions which checks the for uid 0?

criu/cr-restore.c
adrianreber: No, right. Will remove it.

criu/kerndat.c
adrianreber: Yes, good idea.

criu/pie/util.c
adrianreber: Sure.

adrianreber:
> > Right now the kerndat cache file is not written at all. We would need to probably just write it to the user's home-directory if `uid != 0`.
> 
> Well, `~/.criu.kdat ` does not sound bad to me.

I also thought about it and the current design, having it on a `tmpfs`, makes sure the file is gone after a reboot. Maybe we could track CLOCK_MONOTONIC to see if we should recreate `.criu.kdat`. Hopefully no one is running CRIU in a time namespace :wink: - that would break CLOCK_MONOTONIC as an indicator if the system has been rebooted.

criu/cgroup.c
rppt: I meant to have a wrapper function for checking `opts.uid`. E.g. `god_mode()` or more modest `are_we_root()` :)
`

rppt:
> > Well, `~/.criu.kdat ` does not sound bad to me.
> 
> I also thought about it and the current design, having it on a `tmpfs`, makes sure the file is gone after a reboot. Maybe we could track CLOCK_MONOTONIC to see if we should recreate `.criu.kdat`. Hopefully no one is running CRIU in a time namespace wink - that would break CLOCK_MONOTONIC as an indicator if the system has been rebooted.

There is a related issue #1441 about detecting system reboots. I didn't check the details, maybe it does not help here.

adrianreber:
> > > Well, `~/.criu.kdat ` does not sound bad to me.
> > 
> > 
> > I also thought about it and the current design, having it on a `tmpfs`, makes sure the file is gone after a reboot. Maybe we could track CLOCK_MONOTONIC to see if we should recreate `.criu.kdat`. Hopefully no one is running CRIU in a time namespace wink - that would break CLOCK_MONOTONIC as an indicator if the system has been rebooted.
> 
> There is a related issue #1441 about detecting system reboots. I didn't check the details, maybe it does not help here.

I do not think that #1441 can help here. It is more for processes being migrated than for CRIU.

criu/cgroup.c
adrianreber: There is already a function `is_root_user()` but it does a geteuid() every time. Using `opts.uid` I do not need to do a syscall every time. Not sure how important that is. I also looked into putting `if (opts.uid)` into a function but it felt a bit strange to replace `if (opts.uid)` with a function. Seems unnecessary.

criu/pie/util.c
adrianreber: inlined

criu/cr-restore.c
adrianreber: done

criu/kerndat.c
adrianreber: done

adrianreber:
I tried to implement most of the changes from @rppt's review. I did not move the uid 0 test to a function because it felt weird to replace one line of code with a function.

The biggest changes in this update is that now `criu check` should work as non-root. I also added a run of `criu check` to the CI test case. I also changed the CI test case to completely run as non-root. Only the `setcap` call runs as root. Building and testing is all done without root.

Let's see if CI likes my latest changes.

criu/cgroup.c
rppt: Well, my feeling was that `opts.uid` is kinda an instruction from user and a function like `is_root_user()` says what is our
state. 
But I won't insist, we can use `opts.uid`

mbien:
I did some manual testing with this PR applied and took some notes. In case this could be of any help:
- rootless required cap_sys_ptrace additionally to cap_checkpoint_restore on my system
- rootless dump using criu with --shell-job works but restore doesn't, programs have to be completely detached from shells right now ((criu/tty.c:846): tty: Can't set tty params on 0x8: Operation not permitted). Capabilities were set to criu.
- the libcriu C RPC API doesn't work rootless and fails very early ((criu/image.c:550): Can't open dir /proc/4160/fd/3: Permission denied). Capabilities were set to the c program.
- all experiments worked with root, even a self checkpointing JVM using libcriu with tcp established etc (although i found other bugs regarding libcriu, but this would be off topic)

adrianreber:
> I did some manual testing with this PR applied and took some notes. In case this could be of any help:

Thanks a lot for testing.

>     * rootless required cap_sys_ptrace additionally to cap_checkpoint_restore on my system

Either that or `echo 0 > /proc/sys/kernel/yama/ptrace_scope`. The `ptrace_scope` value seems to be different depending on the used distribution.

>     * rootless dump using criu with --shell-job works but restore doesn't, programs have to be completely detached from shells right now ((criu/tty.c:846): tty: Can't set tty params on 0x8: Operation not permitted). Capabilities were set to criu.

I have seen that. Not sure if that can be fixed. non-root checkpoint/restore will have limitations. Maybe that is one of them.

>     * the libcriu C RPC API doesn't work rootless and fails very early ((criu/image.c:550): Can't open dir /proc/4160/fd/3: Permission denied). Capabilities were set to the c program.

Hmm. Never tried it and I do not expect it to work, but capabilities need to be set on the CRIU binary as libcriu is just wrapper around the CRIU binary.

>     * all experiments worked with root, even a self checkpointing JVM using libcriu with tcp established etc (although i found other bugs regarding libcriu, but this would be off topic)

mbien:
> > * the libcriu C RPC API doesn't work rootless and fails very early ((criu/image.c:550): Can't open dir /proc/4160/fd/3: Permission denied). Capabilities were set to the c program.
> 
> Hmm. Never tried it and I do not expect it to work, but capabilities need to be set on the CRIU binary as libcriu is just wrapper around the CRIU binary.

i tried that too actually, just wanted to keep the post short. If i set caps to criu without giving the c program any i get this during checkpoint:

```
(00.000023) Version: 3.15 (gitid v1.5-5425-gb19af3c6)
(00.000021) Running on longbow Linux 5.10.30-1-MANJARO #1 SMP Wed Apr 14 08:07:27 UTC 2021 x86_64
(00.000028) Warn  (criu/kerndat.c:869): Can't load /run/criu.kdat
(00.000038) sockets: Probing sock diag modules
(00.000074) sockets: Done probing
(00.001256) Error (criu/util.c:643): exited, status=4
(00.002420) Error (criu/util.c:643): exited, status=4
(00.002464) Error (criu/kerndat.c:54): Can't open 0/pagemap on procfs: Permission denied
(00.002471) Error (criu/kerndat.c:1192): check_pagemap failed when initializing kerndat.
(00.002494) Found mmap_min_addr 0x10000
(00.002504) files stat: fs/nr_open 1073741816
```

test/zdtm.py
avagin: > What is the history behind --user?

It was Pavel's attempt to implement criu dump that can't be executed by non-root users to dump their processes.

criu/cr-check.c
```diff
@@ -1562,3 +1563,47 @@ static char *feature_name(int (*func)(void))
...
+		pr_perror("Unable to set PR_SET_DUMPABLE");
+	return ret;
+}
+
+int check_caps(void)
+{
+	struct proc_status_creds creds;
+	int exit_code = -1;
+
+	if (parse_pid_status(PROC_SELF, &creds.s, NULL))
```
avagin, line 1588: why we can't use capget here?

adrianreber: I don't know.

criu/files.c
```diff
@@ -1335,6 +1335,8 @@ int prepare_fds(struct pstree_item *me)
 
 static int fchroot(int fd)
 {
+	if (!has_cap_sys_chroot(opts.cap_eff))
+		return 0;
```
avagin, line 1339: Hmmm. I don't think that it is a good idea to return 0 if we can't do chroot. I think we need to inspect each place where we call fchroot and decide what we need to do if we don't have caps.

criu/include/kerndat.h
```diff
@@ -69,6 +69,17 @@ struct kerndat_s {
...
+	 * For non-root CRIU we read CLOCK_MONOTONIC and assume
+	 * that it can only move forward to detect a reboot.
+	 * Hopefully no one is using non-root CRIU in a timens.
+	 * Usually kdat cache is stored on tmpfs, for the non-root
+	 * use case we probably cannot access /run/. So we store it
+	 * ~/.criu.kdat and if the clock moves backward we assume
+	 * the system has been rebooted and the kdat cache needs
+	 * to be recreated.
+	 */
+	time_t clock_monotonic;
```
avagin, line 82: clock_monotonic is started from 0 after reboot, isn't it?

criu/include/util.h
```diff
@@ -383,4 +383,13 @@ static inline void print_stack_trace(pid_t pid) {}
 
 extern int mount_detached_fs(const char *fsname);
 
+#define cleanup_free __attribute__ ((cleanup (cleanup_freep)))
+
+static inline void
```
avagin, line 388: static inline void cleanup_freep (void *p)

you don't need to split this line.

criu/util.c
```diff
@@ -1367,6 +1367,9 @@ void rlimit_unlimit_nofile(void)
 {
 	struct rlimit new;
 
+	if (opts.uid)
```
avagin, line 1370: imho: this check should be out of this function. I mean it should be in a place where this function is called and we need to support it with a comment that explains why we can skip it if we don't have enough permissions.

avagin, line 1370: prlimit checks CAP_SYS_RESOURCE.

criu/util.c
avagin: @adrianreber you marked it as resolved, but for me it is unclear how it was resolved.

criu/cr-restore.c
```diff
@@ -1846,6 +1847,11 @@ static int restore_task_with_children(void *_arg)
 				goto err;
 		}
 
+		if (parse_pid_status(PROC_SELF, &creds.s, NULL))
+			goto err;
+
+		memcpy(&opts.cap_eff, &creds.cap_eff, sizeof(u32) * PROC_CAP_SIZE);
```
avagin, line 1853: what is going on here? why can we not use capget? Pls write a comment here.

adrianreber: I really don't know. I just used what CRIU already provided. If capget() is better, I can use that.

criu/util.c
adrianreber: It's about pre-dump which is out of scope of this PR.

test/zdtm.py
kolyshkin: Can I suggest `--rootless`? It is an established term and the meaning is very clear.

adrianreber: My first patch series was using `--non-root`, now I am currently using `--user`. `--rootless` is indeed a good idea, but I do not really care which name it is.

Should we re-use `--user` or give it a new name?

criu/cr-restore.c
adrianreber: Seeing the old comments I see that @nviennot suggested reading /proc/self/status and as there is a function for that (`parse_pid_status()`) I was using that. I have no strong opinion which was one to use. `capget()` is nicer as it does not require string parsing, but just let me know what is recommended.

test/zdtm.py
avagin: The idea of the --user option is to allow non-root users dumping processes even when criu doesn't have any capabilities. We want to save this functionality.

criu/cr-restore.c
avagin: capget() is much faster and more reliable.

criu/include/kerndat.h
avagin: I don't understand how it can help us to detect the old kdat file...

avagin:
>  If that is true CRIU remembers the euid and decides at different places in the code to not do certain things which do not work as non-root.

In such cases, we need to get approval from a user to work in this mode. I mean we need to add a new criu options and describe in the documentation what certain things will not be made in this case.

test/zdtm.py
adrianreber: Renamed my changes to `--rootless`.

criu/cr-check.c
adrianreber: changed to capget

criu/include/kerndat.h
adrianreber: It doesn't help. This is just wrong. Any recommendations how to detect a reboot to reload the cache file?

criu/include/util.h
adrianreber: done

criu/util.c
adrianreber: okay, will change that

criu/cr-restore.c
adrianreber: switched to capget

adrianreber:
> > If that is true CRIU remembers the euid and decides at different places in the code to not do certain things which do not work as non-root.
> 
> In such cases, we need to get approval from a user to work in this mode. I mean we need to add a new criu option and describe in the documentation what certain things will not be made in this case.
> 
> And in the code, we will check the option flag rather than opts.uid. IMHO this will look more cleaner. Right now, this series checks uid when we need to check capabilities (CAP_SYS_RESOURCE, CAP_NET_ADMIN, CAP_SETGID, CAP_SETUID, etc). The uid check doesn't take into account a case when a user want to dump processes in a separate user namespace where criu has all set of capabilities and its uid is 0.

Can you be more specific of what you want to have? Not clear to me.

criu/files.c
adrianreber: Can you be more specific what changes you want. Not sure what to do here?

criu/util.c
adrianreber: switched to cap checking

avagin:
> > > If that is true CRIU remembers the euid and decides at different places in the code to not do certain things which do not work as non-root.
> > 
> > 
> > In such cases, we need to get approval from a user to work in this mode. I mean we need to add a new criu option and describe in the documentation what certain things will not be made in this case.
> > And in the code, we will check the option flag rather than opts.uid. IMHO this will look more cleaner. Right now, this series checks uid when we need to check capabilities (CAP_SYS_RESOURCE, CAP_NET_ADMIN, CAP_SETGID, CAP_SETUID, etc). The uid check doesn't take into account a case when a user want to dump processes in a separate user namespace where criu has all set of capabilities and its uid is 0.
> 
> Can you be more specific of what you want to have? Not clear to me.

I think the non-root mode in CRIU has to be enabled explicitly and users have to understand its limitation. So we need to add the --non-root option in CRIU and describe how it works in documentation with all details.

rst0git:
>So we need to add the --non-root option in CRIU and describe how it works in documentation with all details.

Could use `--rootless` instead of `--non-root` (for consistency)?
```
--rootless              running as an unprivileged user
```

nviennot:
`--unprivileged` might be better because root != CAP_SYS_ADMIN and all

criu/include/kerndat.h
adrianreber: switched stale detection to release and nodename from `uname()`.

adrianreber:
I am not really convinced about the parameter when running as non-root. But I added now `--unprivileged`. I hope I understood the intention correctly.

I added a section to the man page about running CRIU as non-root and I switched the stale detection to use `uname()`'s release and nodename information.

avagin:
> I am not really convinced about the parameter when running as non-root. 

If you are not convinced, we need to continue discussing before we will not find a consensus.

> This PR checks if the euid is non-zero and if it is non-zero it checks if CRIU has either CAP_SYS_ADMIN or CAP_CHECKPOINT_RESTORE. If that is true CRIU remembers the euid and decides at different places in the code to not do certain things which do not work as non-root.

My idea is that it is okay to "not do certain things which do not work as non-root", but an user has to confirm that CRIU will be run in a restricted mode. There can be a situation when an user wants to do a full dump but call CRIU without root privileges by mistake and I think it has to receive an error in such case.

rst0git:
>There can be a situation when an user wants to do a full dump but call CRIU without root privileges by mistake and I think it has to receive an error in such case.

If I recall correctly, most of the "full dump" functionality already works even when CRIU runs as unprivileged process and [`CAP_CHECKPOINT_RESTORE`](https://github.com/torvalds/linux/commit/124ea650d3072b005457faed69909221c2905a1f) was introduced mainly to help with restoring as non-root. Of course, one unprivileged user is not able to checkpoint a process tree owned by another, but an example use-case for `--unprivileged` could be checkpoint/restore of rootless containers.

avagin:
> If I recall correctly, most of the "full dump" functionality already works even when CRIU runs as unprivileged process 

We can't dump anon shared memory mappings, because we use /proc/pid/map_files for that.

adrianreber:
Rebased to solve a merge conflict.

So what is still missing to merge this?

criu/cgroup.c
avagin, line 735: should it be "if (opts.unprivileged)" ?

criu/cgroup.c
```diff
@@ -989,6 +992,9 @@ int dump_cgroups(void)
 	CgroupEntry cg = CGROUP_ENTRY__INIT;
 	int ret = -1;
 
+	if (opts.uid)
```
avagin, line 995: opts.unprivileged  and we need to print a warning that cgroups are not dumped.

criu/files.c
avagin: fchroot can't return 0 if it has done nothing. I think we need to find all places where it is called and see when we can skip it in the unprivileged mode...

criu/image.c
```diff
@@ -211,7 +211,8 @@ int prepare_inventory(InventoryEntry *he)
 	if (get_task_ids(&crt.i))
 		return -1;
 
-	he->has_root_cg_set = true;
+	if (!opts.uid)
+		he->has_root_cg_set = true;
```
avagin, line 215: This looks very very suspicious. We set has_root_cg_set only if uid is zero, but we always call dump_task_cgroup. If here is any reasons for this, we need to write a comment here.

When I suggested to add the --unprivileged options, I mean that we will check opts.unprivileged instead of opts.uid.

criu/kerndat.c
```diff
@@ -822,19 +823,51 @@ static int kerndat_x86_has_ptrace_fpu_xsave_bug(void)
...
+static int get_kerndat_filename(char **kdat_file)
+{
+	int ret;
+
+	/*
+	 * Running as non-root, even with CAP_CHECKPOINT_RESTORE, does not
+	 * allow to write to KDAT_RUNDIR which usually is only writable by root.
+	 * Let's write .criu.kdat file to the home directory for non-root cases.
+	 */
+	if (opts.uid) {
```
avagin, line 838: opts.unprivileged. opts.uid doesn't guarantee that criu has been executed with a full set of capabilities.

criu/kerndat.c
```diff
@@ -847,24 +880,48 @@ static int kerndat_try_load_cache(void)
 
 	close(fd);
 
+	/*
+	 * For the non-root use case we use uname(2) to decide
+	 * if a reboot happened.
```
avagin, line 885: How can uname help up to detect a reboot? Do you mean a kernel update?

adrianreber:
Hello @avagin,

I think it would make more sense if you take over this PR. You have a much clearer understanding and vision about what needs to be done than myself. Just take what already exist and make the necessary changes. You can close this PR or modify it. Whatever works best for you.

rst0git:
The changes in this PR have been used to enable rootless checkpoint/restore support in https://github.com/OpenLiberty/open-liberty/pull/20683

PR #1166, opened 2020-08-04
Fixes #1165

### LINKED ISSUE #1165: pb2dict.py crashes in string conversion.
The file `pb2dict.py` crashes when part of the value in an image file is a byte string. 
To fix this change line 360 from
```
d[field.name] = d_val
```
to
```
d[field.name] = d_val.decode() if type(d_val) == bytes else d_val
```

derpsteb:
This PR fixes #1165. 

We were experiencing cases where parsed snapshot images included byte string in the resulting dictionaries. These byte strings where not handled correctly and are properly decoded with this change.

Reported by: liikt

### DIFF
--- a/lib/py/images/pb2dict.py
+++ b/lib/py/images/pb2dict.py
@@ -357,7 +357,7 @@ def pb2dict(pb, pretty=False, is_hex=False):
         else:
             d_val = _pb2dict_cast(field, value, pretty, is_hex)
 
-        d[field.name] = d_val
+        d[field.name] = d_val.decode() if type(d_val) == bytes else d_val
     return d
 
 


adrianreber:
Thanks a lot. Do you have re-producer for us? Is this with python 2 or python 3? This looks a bit like a python 2 to python 3 conversion left-over.

avagin:
> Fixes #1165
>
> python: Handle byte strings when converting protobuf to dict

The subject of the commit message has to describe the patch. "Fixes #1165" has to be in the commit message body. I think you can swap lines in your commit message.

```
Subject: python: Handle byte strings when converting protobuf to dict

Fixes #1165
```

derpsteb:
Sorry about the commit message, misinterpreted what I read in the Contribution guide.

I just force pushed to the branch to update the commit message - is that alright or should I open another PR?
I will try to look for a small file that produced the issue.

The code was written by hand and we are using a python 3 script to run `crit decode`. However I just ran the same script using python 2 and it worked as well.

adrianreber:
Force pushing works. (It might depend a bit on who reviewed the PR :smile:)

Could you attach an image which triggers this bug in crit?

adrianreber:
> Sorry about the commit message, misinterpreted what I read in the Contribution guide.

Which part of the Contribution guide was that? Maybe we could reword it.

rppt:
> Force pushing works. (It might depend a bit on who reviewed the PR smile)

We, dinosaurs, are not that scary ;-)

derpsteb:
So turns out `crit-python2` doesn't have the problem :D. 
But `crit-python2` still works if the change is applied and then `crit-python3` also works.

To test, run one of the following commands on the attached file:
`<path>/criu/crit/crit-python2 decode -i ./testcase.img --pretty -o ./file`
`<path>/criu/crit/crit-python3 decode -i ./testcase.img --pretty -o ./file`

[reproducer.zip](https://github.com/checkpoint-restore/criu/files/5028321/reproducer.zip)

The part that led me to the original commit message was the last bit [here](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes):
"If your change address an issue listed in GitHub, please use Fixes: tag with the number of the issue."

rppt:
> The part that led me to the original commit message was the last bit [here](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes):
> "If your change address an issue listed in GitHub, please use Fixes: tag with the number of the issue."

I see what you mean. We definitely can improve here :)

adrianreber:
Thanks for the reproducer. Can confirm that this works now with Python 3. From my side ready to be merged.

If you could add the Python 3 traceback to the commit message (above or below the 'Fixes' line), that would be great to understand what is fixed even without accessing github. This is not something I will insist on, but would be nice to have.

derpsteb:
Added the traceback to the commit message.

PR #1168, opened 2020-08-05
CONTRIBUTING.md: clarify placement of Fixes: tags

rppt:
The description of the Fixes: tags could be misleading and may be
understood as if "Fixes: " should be the commit summary.

Add a sentence with explicit description of Fixme: tags placement.

Reported-by: Otto Bittner <otto-bittner@gmx.de>
Signed-off-by: Mike Rapoport <rppt@linux.ibm.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -124,6 +124,8 @@ If your change address an issue listed in GitHub, please use `Fixes:` tag with t
 	Fixes: #339
 ```
 
+The `Fixes:` tags should be put at the end of the detailed description.
+
 You may refer to [How to Write a Git Commit
 Message](https://chris.beams.io/posts/git-commit/) article for
 recommendations for good commit message.


CONTRIBUTING.md
```diff
@@ -124,6 +124,8 @@ If your change address an issue listed in GitHub, please use `Fixes:` tag with t
 	Fixes: #339
 ```
 
+The `Fixes:` tags should be put in the end of the detailed description.
+
```
adrianreber, line 128: I think it should be `at the end` and not `in the end`.

rppt: Probably you are right, will re-push soon

Snorch: Maybe also clarify "detailed description" to "detailed commit description".

rppt: Would too much IMHO. It's in the section about the commit descriptions and line after this sentence there is a pointer to an article about good commit messages.

Snorch: Ok, agreed.

PR #1169, opened 2020-08-05
travis: loop over apt-get to recover from errors

adrianreber:
One of the most common CI errors we see is that package install fails due to some hashsum mismatch or some DNS errors.

This adds a loop around each apt-get install call to do a clean, update and install and if one of the steps fails it repeats it up to 10 times.

### DIFF
--- a/scripts/travis/travis-tests
+++ b/scripts/travis/travis-tests
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -x -e
 
 TRAVIS_PKGS="protobuf-c-compiler libprotobuf-c-dev libaio-dev python-future
@@ -19,6 +19,27 @@ if [ "$UNAME_M" != "x86_64" ]; then
 	SKIP_TRAVIS_TEST=1
 fi
 
+install_retry_counter=0
+max_apt_retries=5
+
+# This function loops a couple of times over apt-get, hoping to
+# avoid CI errors due to errors during apt-get
+# hashsum mismatches, DNS errors and similar things
+apt_install () {
+	while true; do
+		let install_retry_counter+=1
+		if [ $install_retry_counter -gt $max_apt_retries ]; then
+			exit 1
+		fi
+		apt-get clean -qqy || continue
+		apt-get update -qqy || continue
+		apt-get install -qqy --no-install-recommends $@ && break
+
+		# In case it is a network error let's wait a bit.
+		sleep $install_retry_counter
+	done
+}
+
 travis_prep () {
 	[ -n "$SKIP_TRAVIS_PREP" ] && return
 
@@ -36,8 +57,7 @@ travis_prep () {
 
 	[ -n "$GCOV" ] && {
 		apt-add-repository -y "ppa:ubuntu-toolchain-r/test"
-		apt-get update -yq
-		apt-get -yq --no-install-suggests --no-install-recommends --force-yes install g++-7
+		apt_install --no-install-suggests g++-7
 		CC=gcc-7
 	}
 
@@ -56,8 +76,7 @@ travis_prep () {
 		TRAVIS_PKGS="$TRAVIS_PKGS $X86_64_PKGS"
 	fi
 
-	apt-get update -qq
-	apt-get install -qq --no-install-recommends $TRAVIS_PKGS
+	apt_install $TRAVIS_PKGS
 	chmod a+x $HOME
 }
 
@@ -111,7 +130,7 @@ if [ "${COMPAT_TEST}x" = "yx" ] ; then
 		IA32_PKGS="$IA32_PKGS $i:i386"
 	done
 	apt-get remove $INCOMPATIBLE_LIBS
-	apt-get install --no-install-recommends $IA32_PKGS
+	apt_install $IA32_PKGS
 	mkdir -p /usr/lib/x86_64-linux-gnu/
 	mv "$REFUGE"/* /usr/lib/x86_64-linux-gnu/
 fi


scripts/travis/travis-tests
```diff
@@ -19,6 +19,35 @@ if [ "$UNAME_M" != "x86_64" ]; then
...
+                        continue
+                fi
+                apt-get update -qqy
+                if [ $? -ne 0 ]; then
+                        continue
+                fi
+                apt-get install -qqy --no-install-recommends $@
+                if [ $? -eq 0 ]; then
+                        break
+                fi
```
rppt, line 46: Maybe also 
```
apt-get -f install
```
just in case?

scripts/travis/travis-tests
```diff
@@ -19,6 +19,35 @@ if [ "$UNAME_M" != "x86_64" ]; then
 	SKIP_TRAVIS_TEST=1
 fi
 
+install_retry_counter=0
+max_apt_retries=10
```
rppt, line 23: Do you think a smaller number of retries won't be sufficient?

scripts/travis/travis-tests
adrianreber: sure

scripts/travis/travis-tests
adrianreber: Initially I had 5 retries. 10 is indeed a lot and if it does not work after the third or fourth it will probably fail the tenth time.

I can change it to 5. Maybe I should add a `sleep $install_retry_counter` to give it some time to recover from errors.

scripts/travis/travis-tests
```diff
@@ -19,6 +19,35 @@ if [ "$UNAME_M" != "x86_64" ]; then
...
+
+# This function loops a couple of times over apt-get, hoping to
+# avoid CI errors due to errors during apt-get
+# hashsum mismatches, DNS errors and similar things
+apt_install () {
+        set +e
+        while true; do
+                install_retry_counter=$(expr $install_retry_counter + 1)
```
avagin, line 31: let  install_retry_counter++

scripts/travis/travis-tests
```diff
@@ -19,6 +19,35 @@ if [ "$UNAME_M" != "x86_64" ]; then
...
 fi
 
+install_retry_counter=0
+max_apt_retries=10
+
+# This function loops a couple of times over apt-get, hoping to
+# avoid CI errors due to errors during apt-get
+# hashsum mismatches, DNS errors and similar things
+apt_install () {
+        set +e
```
avagin, line 29: you don't need to do this

scripts/travis/travis-tests
```diff
@@ -19,6 +19,35 @@ if [ "$UNAME_M" != "x86_64" ]; then
...
+        while true; do
+                install_retry_counter=$(expr $install_retry_counter + 1)
+                if [ $install_retry_counter -gt $max_apt_retries ]; then
+                        exit 1
+                fi
+                apt-get clean -qqy
```
avagin, line 35: apt-get clean -qqy || continue

scripts/travis/travis-tests
adrianreber: `let` needs `/bin/bash`. Initially I had `let` but had to change it to `expr`. We can, of course, change the whole script to `/bin/bash`.

rppt: > let install_retry_counter++

Won't work on `dash` which might be used as `/bin/sh` on Ubuntu.
Maybe we want to make travis-tests to explicitly use bash and then `let cnt++` or `((cnt++))` would work.

```
diff --git a/scripts/travis/travis-tests b/scripts/travis/travis-tests
index e9c2d5c6a..d159bb71f 100755
--- a/scripts/travis/travis-tests
+++ b/scripts/travis/travis-tests
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -x -e
 
 TRAVIS_PKGS="protobuf-c-compiler libprotobuf-c-dev libaio-dev python-future
```

avagin: I vote to force using bash for all our shell scripts, but @adrianreber it is up to you

adrianreber:
Reworked and force pushed.

 * Switch to `/bin/bash` as it is probably anyway `/bin/bash` on most systems
 * Removed `set +e` and `set -e`
 * Added a `sleep` if it fails
 * Using `let` instead of `expr`

scripts/travis/travis-tests
adrianreber: done

scripts/travis/travis-tests
adrianreber: done

scripts/travis/travis-tests
adrianreber: switched to `let` and bash

scripts/travis/travis-tests
```diff
@@ -19,6 +19,27 @@ if [ "$UNAME_M" != "x86_64" ]; then
...
+		if [ $install_retry_counter -gt $max_apt_retries ]; then
+			exit 1
+		fi
+		apt-get clean -qqy || continue
+		apt-get update -qqy || continue
+		apt-get install -qqy --no-install-recommends $@ && break
+
+		# In case it is a network error let's wait a bit.
+		sleep $install_retry_counter
+	done
```
Snorch, line 40: @adrianreber Hi, shouldn't we reset install_retry_counter to 0 at the end?

adrianreber: Yes, right. Thanks. I will fix it in a follow up PR. I was planning to move it to its own script and also use it in all the Dockerfiles.

PR #1174, opened 2020-08-10
Images: kindly ask not to use fields with id 18 in unix_sk_entry 

Snorch:
The id 18 in unix_sk_entry is already used in Virtuozzo criu for a long time so we can't easily change it to new ```vz_<field> 1000+<id>``` model so let's kindly ask not to use it.

Previous PR: https://github.com/checkpoint-restore/criu/pull/1152

v2: instead of reserving new fields in original image, lets reserve special vendor message field with id 1024, so that all extra fields are put in it
v3: We will use `vz_<field> 1000+<id>` semantics in Virtuozzo so that it would be easier to handle porting to mainstream and rebasing on mainstream with the ported id. Only thing which is left is id 18 in unix_sk_entry, for which we would kindly ask not to use it.

### DIFF
--- a/images/sk-unix.proto
+++ b/images/sk-unix.proto
@@ -51,4 +51,5 @@ message unix_sk_entry {
 
 	optional uint32			ns_id		= 16;
 	optional sint32			mnt_id		= 17 [default = -1];
+	/* Please, don't use field with number 18. */
 }


images/mnt.proto
```diff
@@ -32,29 +32,34 @@ enum fstype {
...
 
-	optional bool		deleted			= 16;
-	optional uint32		sb_flags		= 17 [(criu).hex = true];
+	optional bool			deleted			= 16;
+	optional uint32			sb_flags		= 17 [(criu).hex = true];
 	/* user defined mapping for external mount */
-	optional string		ext_key			= 18;
+	optional string			ext_key			= 18;
+
+	optional vendor_mnt_entry	vendor			= 1024;
```
rppt, line 68: I won't insist, but I'd prefer an unaligned `vendor_mnt_entry` rather than all this indentation churn.

xemul: Indentation churn may go in separate patch. And, while we're all here, 4 of these patches can be merged together -- no need to add vendor field to different images with different patches.

0x7f454c46:
I totally don't mind this PR, but just thinking @xemul's idea further: do we need to reserve anything?
Like, can vz-criu reserve `optional vz_mnt_entry	virtuozzo			= 1024;` in vz-criu?
The number is big enough not to intersect and the name is unique so it won't conflict.
The only downside will be if another vendor adds internal patch with the same number, but it's the same in this PR.

Again I don't mind if this version gets merged, just trying to minimize some routines that don't add much value to mainstream criu.
I hope you don't get me wrong: it's not an attempt to complicate your work, rather a way to simplify and possibly even stop this toil work for both parties.

0x7f454c46:
Another unfortunate downside comparing to #1152 is that it will be harder to upstream changes from vz-criu.
As after upstreaming, you will have to accept both images formats, which will add legacy code in vz-criu.
So kinda question is to find a trade-off in a discussion.

Probably, worth to write a wiki page about vendor images, the way we deal with them and what was considered.
If we stop on the original #1152 PR than I think we need some rules about the reservation process: i.e., when to remove it? (a year without any attempt to mainstream usage for reserved fields?) That would minimize legacy code in vz-criu and make upstreaming easier, but it will require some job to be done to upstream changes. So the reservation comments would have a value in the end.

xemul:
> As after upstreaming, you will have to accept both 
> images formats, which will add legacy code in vz-criu

This is one the things that encourages people not to develop intrusive things close-source ;)

xemul:
> The number is big enough not to intersect

With this argument people can put vendor stuff directly into images. With big enough numbers and $vendor_ prefix to name.

I think I know one tiny benefit mainstream criu can get from it -- it can check for the vendor field being set and refuse to proceed with restore :P

xemul:
> number is big enough

Actually, numbers grow not only because new fields are added, but also because old stuff gets removed because of better replacement (I know we've not been doing this actively so far).

0x7f454c46:
> With this argument people can put vendor stuff directly into images. With big enough numbers and $vendor_ prefix to name.

Agree. That sounds like a possible solution.

The reason why I wanted the discussion is that I find unjust the way we have been reserving fields.
In particular:
```
	// Reserved for container relative start time
	//optional uint64		start_time	= 19;
```
As far as I've checked the history (a while ago, though) - previous attempts to add the functionality to Torvald's kernel resulted in security concerns.
So, we don't know if it will ever be possible to restore start time for a process. And if it will be - that API kernel provides won't differ.
In the end, it means that if something gets into upstream there's no guarantee that reserved fields will get used and have the same type.

It means that it adds possible legacy to mainstream criu. While I believe we want to support upstreaming of vendor's functionality, the price for adding functionality around mainstream kernel should be on vendor.

If the reason is that a particular feature is needed now at this customer's release ahead of acceptance by mainstream - we can go the #1152 way. But there should be some rules about it.
If the reason is to add value to vendor-specific criu to justify product sales (something that you can't make upstream and upstreaming is not planned) - I think, it's unjust to reserve fields here.

images/mnt.proto
Snorch: @rppt
I can get rid of proper indentation, not a problem.

Snorch: @xemul I will merge 3 (not 4, please see my reply in images/sk-unix.proto) patches.

Snorch:
Thank you all for comments!

First, I've tried to think how vendor would be integrated from vz-criu side, and here it is what I've managed to do:

https://src.openvz.org/projects/OVZ/repos/criu/commits/6d0196739c07

It looks like handling proper allocation of ->vendor fields brings more pain than vendor fields bring help.

Second, I understand that spoiling criu images with will-never-be-used fields is not a beautiful solution.

Third, about injustice of crutches for out of tree code, let me defend Virtuozzo a little bit. Virtuozzo had an attempt to mainstream nested namespaces but it was rejected. Me personally had several attempts to improve mounts code and they were rejected too. There would always be some parts of code Virtuozzo has out of tree, which are required for us to support our customer needs and fix Virtuozzo case specific bugs. Please don't blame us for it we try to mainstream all we have (And we are not "close-source" you can always take a look on https://src.openvz.org/projects/OVZ/repos/criu/browse =) ).

So to conclude it looks like a less pain way for us is to put new fields to images directly but adding vz_ prefix and choosing big (1000+) ids. In this case we would only need to handle name switch when rebasing to mainstream criu (containing the feature) for backward compatibility.

About reservation unix_sk_entry->vz_mnt_id [18] I would ask for as a last exception. It would be hard to handle unix socket images rebase anyway because unix socket bindmounts have possible bugs, but it would make my life little bit easier =) But I don't insist.

0x7f454c46:
> So to conclude it looks like a less pain way for us is to put new fields to images directly but adding vz_ prefix and choosing big (1000+) ids. 

Good, glad we came up to something that will stop growing legacy in mainstream criu.

> About reservation unix_sk_entry->vz_mnt_id [18] I would ask for as a last exception. It would be hard to handle unix socket images rebase anyway because unix socket bindmounts have possible bugs, but it would make my life little bit easier =) But I don't insist.

I'm up for reserving this and nested-namespaces fields. Those are already used in production, so in my POV new ruling shouldn't apply to something that existed before. Ack from my side, unsure if everyone agrees though.

Snorch:
@0x7f454c46  nested-namespaces fields are fine though implemented long time ago, because (sadly) they were disabled in Virtuozzo criu too, and waited for a better time (which is now I hope).

xemul:
> First, I've tried to think how vendor would be integrated from vz-criu side, and here it is what I've managed to do:
> 
> https://src.openvz.org/projects/OVZ/repos/criu/commits/6d0196739c07
> 
> It looks like handling proper allocation of ->vendor fields brings more pain than vendor fields bring help.

Well, I'd say the patch you've mention is not that awful :)

> Second, I understand that spoiling criu images with will-never-be-used fields is not a beautiful solution.

Yes.

> Third, about injustice of crutches for out of tree code, let me defend Virtuozzo a little bit. Virtuozzo had an attempt to mainstream nested namespaces but it was rejected. Me personally had several attempts to improve mounts code and they were rejected too. There would always be some parts of code Virtuozzo has out of tree, which are required for us to support our customer needs and fix Virtuozzo case specific bugs. Please don't blame us for it we try to mainstream all we have (And we are not "close-source" you can always take a look on https://src.openvz.org/projects/OVZ/repos/criu/browse =) ).

Pasha, Virtuozzo created criu in the first place, there's no need in defending the company here.

> So to conclude it looks like a less pain way for us is to put new fields to images directly but adding vz_ prefix and choosing big (1000+) ids. In this case we would only need to handle name switch when rebasing to mainstream criu (containing the feature) for backward compatibility.

Keeping a code that handles two different versions of input data is always painful, but the good news is that older versions wash away sooner than we expect, so this period will end some day.

> About reservation unix_sk_entry->vz_mnt_id [18] I would ask for as a last exception. It would be hard to handle unix socket images rebase anyway because unix socket bindmounts have possible bugs, but it would make my life little bit easier =) But I don't insist.

I already agreed with the exception, but not in form of a commented-out field with exact name and number (you will patch this place anyway) but in form of a single comment saying smth like "please, don't use field with number 19".

Snorch:
I've updated the PR with force-push leaving only the patch related to unix_sk_entry field id 18.

Snorch:
Hi, @checkpoint-restore/maintainers can I have it merged? (also updated pr description)

rppt (approved):

0x7f454c46 (approved):
Thanks!

Snorch:
Great Thanks to everybody involved!

PR #1175, opened 2020-08-11
criu: the type of a socket inode has to be "unsigned int"

avagin:
```
(00.015271) unix: 	Add a peer: ino 2203289876 peer_ino 2203289875 family    1 type    1 state  1 name /mnt/test/zdtm/static/sockets03.test
(00.015277) Warn  (criu/sk-unix.c:475): unix: Shutdown mismatch -2091677421:1 -> -2091677420:0
```

### DIFF
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -116,7 +116,7 @@ static void show_one_inet_img(const char *act, const InetSkEntry *e)
 		src_addr);
 }
 
-static int can_dump_ipproto(int ino, int proto, int type)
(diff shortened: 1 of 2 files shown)


mihalicyn:
LGTM

But, could you, please, also handle:
https://github.com/checkpoint-restore/criu/blob/28f683df2afb522395332d66b100417612193306/criu/sk-unix.c#L1148
https://github.com/checkpoint-restore/criu/blob/28f683df2afb522395332d66b100417612193306/criu/sk-unix.c#L1828

PR #1178, opened 2020-08-12
github: disable daily cross-compile for mips

adrianreber:
The daily cross compile for mips failed on master, because master has no mips support yet. This removes mips for master and criu-dev for now.

Cross compiles for mips will still happen for pull requests and pushes.

### DIFF
--- a/.github/workflows/cross-compile-daily.yml
+++ b/.github/workflows/cross-compile-daily.yml
@@ -10,7 +10,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        target: [armv7-cross, aarch64-cross, ppc64-cross, mips64el-cross]
+        target: [armv7-cross, aarch64-cross, ppc64-cross]
         branches: [criu-dev, master]
 
     steps:


rppt:
So, if I understand correctly, the daily builds for mips won't run neither for master nor for criu-dev but they still will run for both branches for PRs and pushes. 
Can't we just disable mips on master and leave it on criu-dev?

adrianreber:
> So, if I understand correctly, the daily builds for mips won't run neither for master nor for criu-dev but they still will run for both branches for PRs and pushes. 

Yes.

> Can't we just disable mips on master and leave it on criu-dev?

We could with two different files. At least as far as I understand GitHub actions it cannot be done in a single file.

Once we have mips on master it can be enabled again.

rppt:
> > So, if I understand correctly, the daily builds for mips won't run neither for master nor for criu-dev but they still will run for both branches for PRs and pushes.
> 
> Yes.

That would mean that any PR to master branch that is not mips will fail, wouldn't it?
 
> > Can't we just disable mips on master and leave it on criu-dev?
> 
> We could with two different files. At least as far as I understand GitHub actions it cannot be done in a single file.
> 
> Once we have mips on master it can be enabled again.

Maybe we'd split out mips to a separate files for a while?
E.g. https://github.com/rppt/criu/tree/cross-mips

adrianreber:
> > > So, if I understand correctly, the daily builds for mips won't run neither for master nor for criu-dev but they still will run for both branches for PRs and pushes.
> > 
> > Yes.
> 
> That would mean that any PR to master branch that is not mips will fail, wouldn't it?

Hmm... Could be. Not sure how actions really work. We have the definition on criu-dev mentioning master. But master does not yet have the action files. So... Could be. As we do not really open PRs on master we have not yet seen it in "action" 😆


> > > Can't we just disable mips on master and leave it on criu-dev?
> > 
> > We could with two different files. At least as far as I understand GitHub actions it cannot be done in a single file.
> > 
> > Once we have mips on master it can be enabled again.
> 
> Maybe we'd split out mips to a separate files for a while?
> E.g. https://github.com/rppt/criu/tree/cross-mips

Yes, that looks correct. Will you open a PR? I can then close this one.

rppt:
> Yes, that looks correct. Will you open a PR? I can then close this one.

done

PR #1180, opened 2020-08-14
CI: switch to loop based apt-get

adrianreber:
The previously introduced apt_install loop function to make package install more robust against network errors is now moved to its own script used in multiple places.

### DIFF
--- a/scripts/travis/apt-install
+++ b/scripts/travis/apt-install
@@ -1,0 +1,23 @@
+#!/bin/bash
+
+set -e -x
+
+export DEBIAN_FRONTEND=noninteractive
+
+install_retry_counter=0
+max_apt_retries=5
+
+# This function loops a couple of times over apt-get, hoping to
+# avoid CI errors due to errors during apt-get
+# hashsum mismatches, DNS errors and similar things
+while true; do
+	let install_retry_counter+=1
+	if [ ${install_retry_counter} -gt ${max_apt_retries} ]; then
+		exit 1
+	fi
+	apt-get clean -qqy && apt-get update -qqy && apt-get install -qqy --no-install-recommends $@ && break
+
+	# In case it is a network error let's wait a bit.
+	echo "Retrying attempt ${install_retry_counter}"
+	sleep ${install_retry_counter}
+done
--- a/scripts/build/Dockerfile.aarch64-cross
+++ b/scripts/build/Dockerfile.aarch64-cross
@@ -1,11 +1,13 @@
 FROM dockcross/base:latest
 
+COPY scripts/travis/apt-install /bin/apt-install
+
 # Add the cross compiler sources
 RUN echo "deb http://ftp.us.debian.org/debian/ buster main" >> /etc/apt/sources.list && \
   dpkg --add-architecture arm64 && \
-  apt-get install emdebian-archive-keyring
+  apt-install emdebian-archive-keyring
 
-RUN apt-get update && apt-get install -y \
+RUN apt-install \
 	crossbuild-essential-arm64	\
 	libc6-dev-arm64-cross		\
 	libc6-arm64-cross		\
--- a/scripts/build/Dockerfile.armv7-cross
+++ b/scripts/build/Dockerfile.armv7-cross
@@ -1,11 +1,13 @@
 FROM dockcross/base:latest
 
+COPY scripts/travis/apt-install /bin/apt-install
+
 # Add the cross compiler sources
 RUN echo "deb http://ftp.us.debian.org/debian/ buster main" >> /etc/apt/sources.list && \
   dpkg --add-architecture armhf && \
-  apt-get install emdebian-archive-keyring
+  apt-install emdebian-archive-keyring
 
-RUN apt-get update && apt-get install -y \
+RUN apt-install \
 	crossbuild-essential-armhf	\
 	libbz2-dev:armhf		\
 	libexpat1-dev:armhf		\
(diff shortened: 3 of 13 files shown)


adrianreber:
Just a draft to see what the CI says about this.

scripts/travis/apt-install
```diff
@@ -0,0 +1,22 @@
...
+	let install_retry_counter+=1
+	if [ $install_retry_counter -gt $max_apt_retries ]; then
+		exit 1
+	fi
+	apt-get clean -qqy || continue
+	apt-get update -qqy || continue
+	apt-get install -qqy --no-install-recommends $@ && break
+
+	# In case it is a network error let's wait a bit.
+	sleep $install_retry_counter
```
0x7f454c46, line 21: Actually, maybe print something here `Retrying attempt ${install_retry_counter}`?
Probably will make logs easier to understand.

adrianreber: Good idea. I added that.

scripts/travis/apt-install
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+set -e -x
+
+export DEBIAN_FRONTEND=noninteractive
+
+install_retry_counter=0
+max_apt_retries=5
```
rst0git, line 8: It looks like `apt-get` has support for `APT::Acquire::Retries`, which allows to set
> Number of retries to perform. If this is non-zero APT will retry failed files the given number of times. 

Additionally, we could set a few other options as well, for example
```sh
$ cat scripts/travis/80-apt-conf 
APT::Get::Install-Recommends "false";
APT::Get::Install-Suggests "false";
APT::Get::Assume-Yes "true";
APT::Acquire::Retries 5;
```

```
COPY scripts/travis/80-apt-conf /etc/apt/apt.conf.d/
```

adrianreber: Interesting. Would be good to know which way, my loop or the apt settings, performs more reliable in Travis. Probably hard to tell. 

Any other comments/ideas which approach is better suited for us.

rst0git: I haven't thoroughly evaluated the robustness of the `Retries` option but the [implementation](https://github.com/Debian/apt/blob/289ee74dd23cba7e08b08c6c3602bcf4bf8167bc/apt-pkg/acquire-worker.cc#L610) and [test](https://github.com/Debian/apt/blob/289ee74dd23cba7e08b08c6c3602bcf4bf8167bc/test/integration/test-bug-869859-retry-downloads#L24)  in apt-get look reasonable to me. I guess the advantage of using the built-in functionality may be an improved performance as on failure apt-get will retry to connect rather than restating the whole process.

adrianreber: I would really prefer to use the built-in method instead of my shell script around apt-get. I expect, however, that the current shell loop based variant is more robust. I also found https://travis-ci.community/t/apache-arrow-sporadic-build-failures-due-to-hash-sum-mismatch-errors/2563 which seems to be an error like we also have seen. Especially with the sleep in between tries we might have a better chance to recover from a network error.

But using a built-in solution instead of a shell script would be nice. Not sure what is better.

scripts/travis/apt-install
```diff
@@ -0,0 +1,25 @@
...
+
+# This function loops a couple of times over apt-get, hoping to
+# avoid CI errors due to errors during apt-get
+# hashsum mismatches, DNS errors and similar things
+while true; do
+	sleep ${install_retry_counter}
```
avagin, line 14: sleep has to be after apt-get install

Snorch (approved):
Looks nice.

scripts/travis/apt-install
adrianreber: I moved here so that even after:
```
	apt-get clean -qqy || continue
	apt-get update -qqy || continue
```
there will be a sleep. Especially if one of the commands fails because of temporary network problems, a sleep might help to resolve those problems. At least that is one of the ideas behind this PR. If the sleep is at the end it will only run if `apt-get install` fails not if `clean` (okay, really unnecessary) or `update` fails. Therefore I moved it to the beginning. As the `install_retry_counter` is zero it will not introduce a delay for the first run if everything works.

avagin: oh. you call sleep with the zero timeout. I agree, this will work, but I think the code will look better and will be more readable if we move this sleep after apt calls and do this change:

apt-get clean -qqy && apt-get update -qqy && apt-get install -qqy --no-install-recommends $@ && break

Let me know if you don't think so and I will merge it as is.

adrianreber: Will
```
apt-get clean -qqy && apt-get update -qqy && apt-get install -qqy --no-install-recommends $@ && break
```
work in combination with `set -e`? Or will the `set -e` abort the whole script?

scripts/travis/apt-install
```diff
@@ -0,0 +1,25 @@
...
+	let install_retry_counter+=1
+	if [ ${install_retry_counter} -gt ${max_apt_retries} ]; then
+		exit 1
+	fi
+	apt-get clean -qqy || continue
+	apt-get update -qqy || continue
+	apt-get install -qqy --no-install-recommends $@ && break
+
+	# In case it is a network error let's wait a bit.
```
kolyshkin, line 23: IMHO `sleep` would belong here as it would be in-line with the "let's wait" comment. Otherwise we have this comment detached from the actual code.

scripts/travis/apt-install
adrianreber: I looked at the documentation and `set -e` does the right thing here. I changed it to be what you suggested.

scripts/travis/apt-install
adrianreber: moved the sleep to the bottom again

adrianreber:
@avagin Can this also be merged?

rst0git (approved):

PR #1181, opened 2020-08-14
contrib: Add python-future to Debian packages

xemul:
This one contains builtins module from which zdtm.py imports.

Signed-off-by: Pavel Emelyanov <xemul@openvz.org>

### DIFF
--- a/contrib/debian/dev-packages.lst
+++ b/contrib/debian/dev-packages.lst
@@ -18,3 +18,4 @@ libcap-dev
 libaio-dev
 python-yaml
 libnl-route-3-dev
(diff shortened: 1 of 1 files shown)


rst0git:
Hmm, AFAIK `python-future` is used for Python 2/3 compatibility, however, since Python 2 is now deprecated it would be better to Python 3 only (at least for `zdtm.py`). @xemul what do you think?

xemul:
Switching to python 3 is the way to go for sure, but current state of zftm.py needs this package, at least on my Debian laptop :)

PR #1183, opened 2020-08-15
travis: add a focal based test run

adrianreber:
It seems travis supports now Ubuntu 20.04. Let's run at least one test also on 20.04 (focal).

This depends on #1074 because 20.04 seems to have limited Python 2 support. 

Tests will fail without #1074. Only opening as a draft until  #1074 is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,10 @@ env:
 jobs:
   include:
     - os: linux
+      arch: amd64
+      env: TR_ARCH=local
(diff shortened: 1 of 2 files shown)


adrianreber:
Assuming tests are happy this is ready to be added. It adapts `travis-tests` to work on *focal* (20.04) and excludes the test *sched_policy00* because it fails on 20.04 (focal) with "Operation not permitted".

PR #1191, opened 2020-08-21
eventpoll: resort toff back by idx

Snorch:
In dump_one_eventpoll we call find_tfd_bsearch with arguments
"e->tfd[i]->tfd, toff[i].off". By this we want to check if
"toff[i].off"-th inclusion of target "e->tfd[i]->tfd" in an eventpoll
coresponds to our fd "e->tfd[i]->tfd" in terms of kcmp.

But because of toff was sorted to calculate off, indexes in e->tfd and
toff does not address the same eventpoll target at this point.

Let's sort toff back to original state to make indexes consistent.

### DIFF
--- a/test/zdtm/static/epoll01.c
+++ b/test/zdtm/static/epoll01.c
@@ -1,0 +1,121 @@
+#include <unistd.h>
+
+#include <errno.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include <sys/eventfd.h>
+#include <sys/ioctl.h>
+#include <sys/epoll.h>
+#include <sys/types.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <inttypes.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc	= "Check another case of epoll: This adds three epoll "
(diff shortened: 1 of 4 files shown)


cyrillos:
Ack. Thanks!

Snorch:
@cyrillos Thanks for a review! But it looks like I need some more... When preparing a test I've found another related problem in dump_one_eventpoll.

So I've added another fix and a test to differentiate the problem.

test/zdtm/static/epoll01.c
```diff
@@ -0,0 +1,119 @@
...
+#include <sys/types.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <stdint.h>
+#include <inttypes.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc	= "Check for epoll";
```
avagin, line 22: Could you add a comment that explains what this test checks? It can be the description from the commit message.

Snorch: I've already put to commit message:
```
This adds three epoll targets on tfd 702 and then adds two epoll targets
on tfd 701. This test is for off calculation in dump_one_eventpoll, the
reverse order makes qsort to actually work.
```
Isn't this enough?

cyrillos (approved):

test/zdtm/static/epoll01.c
Snorch: After rereading your message once again, look like you just ask to add description as a comment in code. Sure - no problem. Done.

PR #1192, opened 2020-08-23
[GSoC: 2020] Optimize logging engine

prakritigoyal19:
This pr is a follow up of the pr #1145. 

This makes changes to the existing logging system and adds an additional binary logger. The user now has an option to log in either binary or via the existing printer. The zdtm.py completely logs in binary and the decoder is used to convert these logs to human-readable format before printing them.

The changes that are still needed in this pr are:
- Fixing the _Generic macro for centos: Using P99_GENERIC gives a gcc internal compiler error.
- The default for _Generic has been made 0: It needs to be replaced with a way to determine if the input passed is a struct pointer or a function pointer.
- The decoder fails for arm64 architecture: The va_list is passed as (void *), and this is accepted by all other architectures but is not acceptable in arm64.


xemul:
Awesome :) it's a pity that aarch64 is the blocker. Is it possible to teach zdtm.py not to use binary logging on this arch? If yes, we could merge the pr into dev.

crit/decoder.c
```diff
@@ -0,0 +1,95 @@
...
+	size_t i, ret;
+	char *fmt, *b;
+	int size, n;
+
+	while (1) {
+		ret = read(fdin, mbuf, sizeof(m));
+
+		if (ret == 0)
+			break;
+		if (ret < 0) {
```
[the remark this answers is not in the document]
avagin: EINTR has to be handled only if a process sets a signal handler without the SA_RESTART flag.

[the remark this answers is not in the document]
mihalicyn: @avagin thank you!

crit/decoder.c
```diff
@@ -0,0 +1,95 @@
...
+		size += buf_off;
+		ret = write(fdout, buffer, size );
+	}
+	return 0;
+}
+
+int main(int argc, char *argv[]){
+	int fdin, fdout;
+	fdin = open(argv[1], O_RDONLY);
+	if(fdin<0){
```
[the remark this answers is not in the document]
prakritigoyal19: Thank you for pointing these out! I have made the changes.

prakritigoyal19:
@xemul  Since this error is arising during compilation time of aarch64 architecture and arm architecture, I disabled binary logging for these 2 architectures. For all the others, it uses the flog printer, but for these 2, it uses the regular printer for logging. 
The decoder that is called by zdtm.py just copies the contents of one file to another and contents are printed from that file.

criu/log.c
```diff
@@ -351,7 +350,7 @@ static void early_vprint(const char *format, unsigned int loglevel, va_list para
 	early_log_buf_off += log_size;
 }
 
-static void vprint_on_level(unsigned int loglevel, const char *format, va_list params)
+void vprint_on_level(unsigned int loglevel, const char *format, va_list params)
```
rppt, line 353: I think this change should have been done in the third patch, otherwise it would break the build and bisect

xemul:
@prakritigoyal19 , the build on centos environment fais:
https://travis-ci.org/github/checkpoint-restore/criu/jobs/721650710

criu/config.c
```diff
@@ -528,6 +528,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{ "status-fd",			required_argument,	0, 1088 },
 		BOOL_OPT(SK_CLOSE_PARAM, &opts.tcp_close),
 		{ "verbosity",			optional_argument,	0, 'v'	},
+		{ "binary-file",		no_argument,	0, 'b' },
```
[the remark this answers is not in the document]
prakritigoyal19: I will make the change of binary-file to binary-log.

criu/flog.c
```diff
@@ -10,6 +10,9 @@
```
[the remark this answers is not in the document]
prakritigoyal19: Thank you, I will make this change.

criu/arch/ppc64/flog.c
```diff
@@ -0,0 +1,238 @@
...
+				m->args[i] = (long)(mbuf + m->args[i]);
+			}
+		}
+
+		if (ffi_prep_cif(&cif, FFI_DEFAULT_ABI, m->nargs + 2,
+				 &ffi_type_sint, args) == FFI_OK)
+			ffi_call(&cif, FFI_FN(dprintf), &rc, values);
+	}
+	return 0;
+}*/
```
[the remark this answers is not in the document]
prakritigoyal19: No sir I have used it in the crit/decoder.c. I will remove the commented code from here.

prakritigoyal19:
> @prakritigoyal19 , the build on centos environment fais:
> https://travis-ci.org/github/checkpoint-restore/criu/jobs/721650710

Yes sir, the _Generic macro is supported from C11 and forth, and centos doesn't support this version yet, thus the error is generated. I had asked about this issue in the gitter community group. I was then suggested to try using the P99_GENERIC mentioned here: https://gustedt.wordpress.com/2012/01/02/emulating-c11-compiler-features-with-gcc-_generic/
As I have even mentioned in the pr message, when I try working with this in the code, it leads to internal compiler error and I am having issues trying to understand why this is occurring. I have been looking for other alternatives but couldn’t find a workable solution.
Could you please suggest any other alternatives for me to try out?

xemul:
Then we should just not compile in the support for flog in environments, that do not support _Generic

crit/arch/aarch64/decoder.c
```diff
@@ -0,0 +1,29 @@
...
+		fprintf(stderr, "Unable to open file: %m");
+		return -1;
+	}
+	fdout = open(argv[2], O_CREAT|O_TRUNC|O_WRONLY|O_APPEND);
+	if (fdout<0){
+		fprintf(stderr, "Unable to open file: %m");
+		return -1;
+	}
+	read(fdin, c, 1);
+	while (c != EOF)
```
xemul, line 23: That's a pity, that decoder code is duplicated across arches.

criu/arch/aarch64/include/asm/vdso.h
```diff
@@ -9,6 +9,7 @@
  * This is a minimal amount of symbols
  * we should support at the moment.
  */
+#define no_binary_log
```
xemul, line 12: @cyrillos , @0x7f454c46 , is vdso.h the correct place for such a thing?

criu/log.c
xemul: @prakritigoyal19 , is it done in the new version of the pr?

criu/arch/aarch64/include/asm/vdso.h
0x7f454c46: I don't think so. Also, double-negative sounds hard to read:
`#ifndef no_binary_log`
And as it's a macro it should be in CAPITAL LETTERS.
If I would do something like that, I'd add:
```
ifeq ($(ARCH),x86)
        LDARCH		:= i386:x86-64
        DEFINES		:= -DCONFIG_X86_64 -DARCH_HAS_FLOG
endif
```
And so on to the supported arches. That way one can disable compilation of something that's not working anyway.

Furthermore, I don't understand what does `aarch64 and arm architectures are not compatible with the decoder`.
Is it something fixable? What means `not compatible`? What are the errors if you try enabling it? Is there a plan to support it?
Probably, with such definition should go a descriptive comment..

criu/log.c
prakritigoyal19: Yes sir, I have moved this change to patch 3 - flog: add option to fallback to default textual logging

criu/arch/aarch64/include/asm/vdso.h
prakritigoyal19: Thank you, I will make these changes.

The line that causes issues with the decoder is this:
`size  = vsnprintf(buffer + buf_off, sizeof buffer - buf_off, fmt, (void *)values);`
 
The other architectures accept va_list in type (void *) but these arm and aarch64 architectures produce errors of an incompatible type. To fix this issue, there needs to be another way to feed the printf family function a variable number of variables in a way that is compatible with these architectures. 

I will also add this description to the commit message.

0x7f454c46:
>  flog/include/compiler.h → criu/include/compiler.h 
Do you need this?
I don't see anything new there from `git diff flog/include/compiler.h include/common/compiler.h`

criu/arch/aarch64/include/asm/vdso.h
0x7f454c46: Hey, I've looked at the result..
Let me try to help you a bit here from the makefile side:
https://github.com/0x7f454c46/criu/commit/6a659be02e56a06e84587a1937e594cb2eeee7ea
Please, reuse/integrate the changes in your commits: this way arm/aarch64 will skip compiling criu/flog.o and you won't need those scary undefs. I've also removed in that commit compiling crit/decoder for unsupported case, but if you feel like keeping it for consistency - I don't insist, just return ifdef CONFIG_HAS_FLOG there.
Thanks!

crit/Makefile
```diff
@@ -1,5 +1,12 @@
 
-all-y	+= crit
+all-y	+= crit decoder
+
+CC	= gcc
+CFLAGS  = -g -Wall
+
+decoder:
+	$(Q) $(CC) $(DEFINES) crit/decoder.c -o crit/decoder
```
0x7f454c46, line 8: Please, add it to .gitignore so that git gets to know that it doesn't need track changes in the binary.
```
diff --git a/.gitignore b/.gitignore
index 23cd703be2f5..83368327de5d 100644
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ images/google/protobuf/*.h
 .gitid
 criu/criu
 crit/crit
+crit/decoder
 criu/arch/*/sys-exec-tbl*.c
 # x86 syscalls-table is not generated
 !criu/arch/x86/sys-exec-tbl.c
```

criu/flog.c
```diff
@@ -150,12 +153,30 @@ int flog_close(int fdout)
...
+	if (!opts.log_in_binary){
+		va_start(argptr, format);
+		vprint_on_level(loglevel, format, argptr);
+		va_end(argptr);
+		return 0;
+	}
+
+	if (unlikely(loglevel == LOG_MSG)) {
+		fdout = STDOUT_FILENO;
+	}else {
```
0x7f454c46, line 116: I'm not sure I read it correctly: does it mean that for LOG_MSG messages we get binary output out of criu?
So that criu-check and stuff like that will break tty setup for a user?
Should those messages just get printed instead?

criu/flog.c
```diff
@@ -150,12 +153,30 @@ int flog_close(int fdout)
...
+		current_loglevel = log_get_loglevel();
+		if (loglevel > current_loglevel)
+			return 0;
+		fdout = log_get_fd();
```
0x7f454c46, line 120: The same question is about stderr: I see you print messages there unencoded (in flog.c), should you do the same for criu if the destination is stderr?

criu/flog.c
```diff
@@ -150,12 +153,30 @@ int flog_close(int fdout)
 	return 0;
 }
 
-int flog_encode_msg(int fdout, unsigned int nargs, unsigned int mask, const char *format, ...)
+int flog_encode_msg(int loglevel, unsigned int nargs, unsigned int mask, const char *format, ...)
```
0x7f454c46, line 98: Looking at this function, I think it makes sense to integrate common functionality with `vprint_on_level()` into some new function. At this moment with binary log you won't know when the logging is set up by criu and text logging uses an early log for such case which here is omitted and stdout/stderr may get some binary message out.
Also, it may be worth to still have LOG_TIMESTAMP functionality.

0x7f454c46: Having one common function would also take the burden of defining all pr_*() macro twice for flog_encode() and for print_on_level()

criu/flog.c
```diff
@@ -20,64 +20,9 @@ static char *fbuf;
...
-		values[1] = &fmt;
-
-		for (i = 0; i < m->nargs; i++) {
-			values[i + 2] = (void *)&m->args[i];
-			if (m->mask & (1u << i)) {
-				m->args[i] = (long)(mbuf + m->args[i]);
-			}
-		}
-
-		if (ffi_prep_cif(&cif, FFI_DEFAULT_ABI, m->nargs + 2,
```
0x7f454c46, line 71: There's another one after all changes:
```
/*extern char *rodata_start;
extern char *rodata_end;
*/
```

criu/flog.c
```diff
@@ -20,64 +20,9 @@ static char *fbuf;
 static uint64_t fsize;
```
0x7f454c46, line 20:
```
	if (fsize == 0)
		fsize += BUF_SIZE;
	fsize += BUF_SIZE;
```
Please, just static-initialize with BUF_SIZE.

0x7f454c46:
Looking at the end result, I have some questions:
1. You don't integrate it with parasite/restorer as I can see, right? Is it planned? How do they work right now: do they write text data into binary log file?
2. I don't see any synchronization in the buffer code. I doubt we always run in single-threaded mode even dropping the restorer: at least swrk mode and userns-helpers come in my mind that fork()s out and will have the same shared log buffer. I probably miss something, but won't they override each other?
3. As I understand the binary logging is for the performance. Could you also gather some numbers to see what's the speed-up?
4. The testing. Probably, there's some plan for testing it? What's on my mind: can we run C/R twice and compare the log output for text/binary forms? I think, numbers may differ (PIDs/FDs/etc), so we can cut the numbers from logs and compare that all the text is the same. Probably, when PIEs support gets added, messages may appear in a different sequence, but that's also solvable by sorting the results. This will verify that both logging engines keep working and provide the same result. Practically, it can be even possible to run such testing for all zdtm tests in a separate Travis job.
Thanks for working on it,
Dmitry

criu/flog.c
prakritigoyal19: I have been looking for ways to identify if it is printed in the stderr, but since the fd value that we get via the log_get_fd() doesn't match the standard stderr fd value, I am not able to differentiate between the one being printed in the file and one being printed on stdout/stderr. 
I am thinking of making the -b option available exclusively with the -o option by introducing variables in the parse_options() function. This variable will tell us if it's being logged in a file or not.

criu/flog.c
prakritigoyal19: As per my understanding, you are asking me to eliminate the need for print_on_level() and transfer all log calls to flog_encode() for all the architectures. The issue arises with the centos architecture, it doesn’t recognize _Generic and throws error upon encountering it. This is the main reason for having 2 definitions of pr_*() marcos.
If I try to have an intermediate function to segregate the request for binary logging compatible architectures and incompatible architectures, the issue is sending the arguments from one function to another. The flog needs all the arguments in its original form and I have not been able to find a way to do it from one function to another.
Also, I had faced issues due to the early logging but hadn’t realized then that the messages in the terminal were being printed because of this. I will try to introduce an early logging facility, similar to the normal printer, for the binary printer too.
I also plan to integrate LOG_TIMESTAMP and PID and make the difference between the normal printer and the binary printer, after decoding, minimal. 
Please let me know if I misunderstood and you had meant something different.

prakritigoyal19:
> Looking at the end result, I have some questions:
> 
> 1. You don't integrate it with parasite/restorer as I can see, right? Is it planned? How do they work right now: do they write text data into binary log file?

Currently, the logs from compel are just printed as text data into the binary file. The decoder is built in such a way that whenever it encounters such messages, it just copies these messages to the output file as it is. 
Before, when the decoder wasn’t able to parse these messages, there were plans of using register_logging for the libraries. Are we continuing with that plan or are we introducing the binary logging in compel too?

> 2. I don't see any synchronization in the buffer code. I doubt we always run in single-threaded mode even dropping the restorer: at least swrk mode and userns-helpers come in my mind that fork()s out and will have the same shared log buffer. I probably miss something, but won't they override each other?

When I ran this code in my system, during restoring I could see multiple forks, and when I compared it with the logs using the regular printer, the output produced in both were similar. I am looking at the documentation of how to run criu in swrk mode, but I am not completely clear on how to do that. 

> 3. As I understand the binary logging is for the performance. Could you also gather some numbers to see what's the speed-up?

I have compared the time taken by only a simple loop. When I compare the time taken by it to dump and restore. Taken an average over a few trials, these are the values I receive. 
For dumping without binary: 53s
Restore without binary: 20s
Dump with binary: 42s
Restore with binary: 18s
This is a ~20% increase in the dump and ~10% in the restore. Are there any other programs you would like me to run and collect these numbers?

> 4. The testing. Probably, there's some plan for testing it? What's on my mind: can we run C/R twice and compare the log output for text/binary forms? I think, numbers may differ (PIDs/FDs/etc), so we can cut the numbers from logs and compare that all the text is the same. Probably, when PIEs support gets added, messages may appear in a different sequence, but that's also solvable by sorting the results. This will verify that both logging engines keep working and provide the same result. Practically, it can be even possible to run such testing for all zdtm tests in a separate Travis job.

I did the above-mentioned testing manually and compared the output produced by the normal logger and the binary logger. They were similar. Would you like me to try and write a script to automate this checking process? 

> 
> Thanks for working on it,
> Dmitry

criu/flog.c
0x7f454c46: > I am thinking of making the -b option available exclusively with the -o option

Sounds like a good idea to me. You can also try using `isatty()` as an additional check.

criu/flog.c
0x7f454c46: Hmm, I'll try to help you with this when you'll have the next version of PR.
At this moment, instead of redefining all pr_*() functions, could you try to rather redefine `print_on_level()` as a macro under `ifdef CONFIG_HAS_FLOG`?

criu/arch/aarch64/include/asm/vdso.h
prakritigoyal19: Thanks a lot! I will integrate these changes and send the update by tonight.
I believe the crit/decoder would be needed for the other unsupported cases as well because zdtm.py generates calls for them. Would you suggest changing zdtm.py for unsupported cases, or is it better to retain the decoder?

compel/plugins/std/log.c
```diff
@@ -355,6 +355,27 @@ void print_on_level(unsigned int loglevel, const char *format, ...)
 	sbuf_log_flush(&b);
 }
 
+int flog_encode_msg(int loglevel, unsigned int nargs, unsigned int mask, const char *format, ...)
```
Snorch, line 358: Probably copying print_on_level code here is not a good thing.

test/zdtm.py
```diff
@@ -897,6 +897,8 @@ def __set_opts(criu, args, ctx):
             arg = args.pop(0)
             if "-v4" == arg:
                 criu.opts.log_level = 4
+            elif "-b" == arg:
+                continue
```
Snorch, line 901: Do we plan to add binary logging setup via rpc?

Snorch (commented):
Probably we should move "Add error message for using binary logger without an output file" after "Add option to convert logs to binary".

I mean if one checkouts to intermediate state of your patchset they should not be able to use binary logging without output file. Or am I missing something?

compel/plugins/std/log.c
prakritigoyal19: Without including that function, I received an error: external symbol in PIE. I had copied the print_on_level code as I didn’t want to lose any log messages. I am also not sure if we would include binary logging in compel or if we would be choosing some other logging technique for it. 
 
Rather than copying the print_on_level code, should I make it return without printing, or should I make both the functions point to a goto label, or is there some other alternative that I could use to solve this?

prakritigoyal19:
In the commit “flog: add option to fallback to default textual logging” I have added a check to see if the messages are being printed into the terminal. If the fd is pointed at the terminal then they’ll be printed using the regular printer. I added this commit as an extra check. Should still I make the change?

Snorch:
If I run simple env00 with this series applied:

```
/test/zdtm.py run -t zdtm/static/env00 -f h -k always
```

"decode" hangs on decoding for too much time at 100% cpu load:

```
crit/decoder test/dump/zdtm/static/env00/48/1/dump.log /proc/self/fd/0
Version: 3.14 (gitid v3.14-130-g6aa7d492e)
Running on snorch Linux 5.8.7-200.fc32.x86_64 #1 SMP Mon Sep 7 15:26:10 UTC 2020 x86_64
Loaded kdat cache from /run/criu.kdat
========================================
Dumping processes (pid: 48)
========================================
rlimit: RLIMIT_NOFILE unlimited for self
Running pre-dump scripts
irmap: Searching irmap cache in work dir
No irmap-cache image
irmap: Searching irmap cache in parent
irmap: No irmap cache

...
```
Spinning endlessly in decode_all while loop on same mbuf:

```
(gdb) p *mbuf@200
$25 = "(00.0001\000\000\000\000\000\000\000\000\030\000\000\000\000\000\000\000irmap: No irmap cache\n\000cache in parent\n\000\n\000iu.kdat\000 %s %s %s %s %s\n\000snorch\000Linux\000\065.8.7-200.fc32.x86_64\000#1 SMP Mon Sep 7 15:26:10 UTC 2020\000x86_64", '\000' <repeats 32 times>
```

crit/decoder.c
```diff
@@ -0,0 +1,125 @@
...
+
+		if (ret == 0)
+			break;
+		if (ret < 0) {
+			fprintf(stderr, "Unable to read a message: %m");
+			return -1;
+		}
+		if (m->magic != MAGIC) {
+			lseek(fdin, -sizeof(ret), SEEK_CUR);
+			read(fdin, b, 1);
```
Snorch, line 51: You need to check return codes of ALL functions.

Snorch:
>In the commit “flog: add option to fallback to default textual logging” I have added a check to see if the messages are being printed into the terminal. If the fd is pointed at the terminal then they’ll be printed using the regular printer. I added this commit as an extra check. Should I still make the change?

Yes you should. "Tty file" and "no file at all" are different things as for me.

prakritigoyal19:
> > In the commit “flog: add option to fallback to default textual logging” I have added a check to see if the messages are being printed into the terminal. If the fd is pointed at the terminal then they’ll be printed using the regular printer. I added this commit as an extra check. Should I still make the change?
> 
> Yes you should. "Tty file" and "no file at all" are different things as for me.

Okay sir, I will make this change

prakritigoyal19:
> If I run simple env00 with this series applied:
> 
> ```
> /test/zdtm.py run -t zdtm/static/env00 -f h -k always
> ```
> 
> "decode" hangs on decoding for too much time at 100% cpu load:
> 
> ```
> crit/decoder test/dump/zdtm/static/env00/48/1/dump.log /proc/self/fd/0
> Version: 3.14 (gitid v3.14-130-g6aa7d492e)
> Running on snorch Linux 5.8.7-200.fc32.x86_64 #1 SMP Mon Sep 7 15:26:10 UTC 2020 x86_64
> Loaded kdat cache from /run/criu.kdat
> ========================================
> Dumping processes (pid: 48)
> ========================================
> rlimit: RLIMIT_NOFILE unlimited for self
> Running pre-dump scripts
> irmap: Searching irmap cache in work dir
> No irmap-cache image
> irmap: Searching irmap cache in parent
> irmap: No irmap cache
> 
> ...
> ```
> 
> Spinning endlessly in decode_all while loop on same mbuf:
> 
> ```
> (gdb) p *mbuf@200
> $25 = "(00.0001\000\000\000\000\000\000\000\000\030\000\000\000\000\000\000\000irmap: No irmap cache\n\000cache in parent\n\000\n\000iu.kdat\000 %s %s %s %s %s\n\000snorch\000Linux\000\065.8.7-200.fc32.x86_64\000#1 SMP Mon Sep 7 15:26:10 UTC 2020\000x86_64", '\000' <repeats 32 times>
> ```
> 
> Decode requires a good testing.

Do you have any suggestions on how I should go about testing it

crit/decoder.c
prakritigoyal19: I will make this change and return the appropriate error messages for the same.

Snorch:
> Do you have any suggestions on how I should go about testing it?

Probably you can run your script manually for all the logs zdtm can generate with `./test/zdtm.py run -a -k always` to check that at least we would be able to read zdtm logs after your change.

And also I think you need to prepare a fake test log which covers binary log messages mixed with printf log messages and binary log messages for different types of variables, and all other cases you cover (you should know them better than me =)) And add a bash script to check it, so that it can be run in jenkins/travis/github-checks.

PR #1194, opened 2020-08-26
dump: at exit do not call timing_stop if stats are not initialized

valeriyvdovin:
Stats often call 'get_timing' function which has a BUG() assertion to
catch cases when stats failed to initialize correctly.
If stats haven't initialized yet assertion will also be triggered.
We dont want the trigger to happen in a case when criu fails at early
steps before initializing stats, but this can happen in the following
case:
  - at cr_dump_tasks criu can catch error before the call to init_stats.
  - it then decides to gracefully quit with error and
    calls cr_dump_finish.
  - cr_dump_finish will call timing_stop -> get_timing
    and BUG() gets triggered

But because criu is already quitting gracefully, BUG() is not needed.
In this code path we can call timing_stop under proper condition.

Signed-off-by: Valeriy Vdovin <valeriy.vdovin@virtuozzo.com>


valeriyvdovin:
As an alternative way of solving this problem we could possibly get rid of "BUG();" line in get_timing and just return NULL, but:
The callers of get_timing are 
- timing_start
- timing_stop
- encode_time
all of which have return type void.
This brings me to the idea that removing BUG line will actually lead to heavy refactoring because we would need to handle error cases in all code locations where these three are called. Only timing_stop is called in 15 different places.

Snorch (approved):

criu/stats.c
```diff
@@ -224,7 +224,11 @@ int init_stats(int what)
 		dstats = shmalloc(sizeof(*dstats));
 		return dstats ? 0 : -1;
 	}
-
```
avagin, line 227: this should not be in this patch

avagin:
> at cr_dump_tasks criu can catch error before the call to init_stats.

can we call init_stats earlier?

valeriyvdovin:
> > at cr_dump_tasks criu can catch error before the call to init_stats.
> 
> can we call init_stats earlier?

We could do that, actually i wanted to do that, but it's too implicit. Rearranging series of init calls can lead to subtle problems of same type that we solve here. I'm afraid that we might not see them immediately, instead some strange bugs will appear later. There is also a view from the opposite side of this fix. Suppose we have reordered now and fixed our problem, but later  some other bug would also be fixed by reordering init calls in the other direction and cause our current problem to return.
For the developer doing that fix and reviewers of it's PR  there won't be any clear and explicit written clues that reordering will reintroduce this exact bug.
That is why i think that being more or less independent from init order is a good thing in this situation.

avagin:
This change has been pushed in 41b486790a63d37a22791fbd777f2b053171e0ca.

PR #1196, opened 2020-08-28
[RFC] Add clang-format, checkpatch to tree

mihalicyn:
Hi!

1. This is just a proposal to add format clang-format to declare our code style and simplify the code style checking process.
2. Also, I've tried to adopt checkpatch.pl from the Linux kernel tree. I made a few changes to make it work with the CRIU tree
without problems. But it still makes sense to go through checkpatch.pl code and comment all unnecessary checks (I've commented out the most obvious places.) I will do that If we decide that these changes is useful for our project.
3. In a bright future, we can take a look on https://github.com/marketplace/actions/clang-format-check
4. We can add to checkpatch.pl checks like `pr_err/debug/warn` without `\n` at the end, and `pr_perror` with unnecessary `\n` at the end, and so on and so forth

Regards, Alex


rppt:
I like the idea very much, but I'm a bit concerned about maintaining our own copy of `checkpatch.pl`. The initial difference does not look encouraging in that respect ;-) 

```
$ diff linux/scripts/checkpatch.pl criu/scripts/checkpatch.pl  | diffstat
 unknown |  778 +++++++++++++++++++++++++++++++++++-----------------------------
 1 file changed, 436 insertions(+), 342 deletions(-)
```

Did you check the possibility to have a simple wrapper that will use the Linux's version as is with carefully chosen `--types` and `--ignore`?

Another thing, if we add some coding style checks, they should be documented in CONTRIBUTING.md

mihalicyn:
Hi, Mike,

> Did you check the possibility to have a simple wrapper that will use the Linux's version as is with carefully chosen
Hm, interesting idea. I will take a look on that. If it doesn't work for us, then we may try to write "patch script" which will be applied to original checkpatch.pl and generate our-own version. Most of my changes are trivial and can be done with `sed` or something similar.

Thank you!

adrianreber:
I also like this, especially if we can tie it in into CI. I also see it as a problem to maintain our own copy of `checkpatch.pl`.

Couldn't we add the format-check to CI as a test which is allowed to fail and use that as basis to getting it fixed. I guess my idea is if this is included in CI and at some point blocking for a merge we probably do not really need `checkpatch.pl`.

I just tried it out and it reports a lot of wrong lines, so currently it seems that the existing code is far away from the definition in `.clang-format`. Maybe we would need a check during CI to see that new code follows the existing coding guidelines.

0x7f454c46:
> Did you check the possibility to have a simple wrapper that will use the Linux's version as is with carefully chosen `--types` and `--ignore`?

+1. If there's not enough options to ignore the differences between kernel and criu, you could possibly propose a patch to the original script that will have a needed option(s)?

mihalicyn:
Hi guys!

Thank you very much for your comments and suggestions.

1. About simple wrapper. Looks like as a first and easy solution we can just use original `checkpatch.pl` without *any* changes if we add `--no-tree` as command line parameter. This will skip source tree tests (which files/directories must be present in the valid kernel tree).

2. About `--types` and `--ignore`. Great idea! I've analyzed message types and built this list:
`--ignore SPDX_LICENSE_TAG,DEFINE_ARCH_HAS,FILE_PATH_CHANGES,NEW_TYPEDEFS,GLOBAL_INITIALISERS,GIT_COMMIT_ID,SWITCH_CASE_INDENT_LEVEL,FUNCTION_ARGUMENTS,SSCANF_TO_KSTRTO,MINMAX`

Looks like all warning/errors except this are valid on last 1000 commits.
I've tested using:
```
perl ./scripts/checkpatch.pl --no-tree --show-types --terse --ignore SPDX_LICENSE_TAG,DEFINE_ARCH_HAS,FILE_PATH_CHANGES,NEW_TYPEDEFS,GLOBAL_INITIALISERS,GIT_COMMIT_ID,SWITCH_CASE_INDENT_LEVEL,FUNCTION_ARGUMENTS,SSCANF_TO_KSTRTO,MINMAX -g HEAD~1000..HEAD > types.txt
cat types.txt | grep -Pv '(LONG_LINE|SPAC|COMMIT_MESSAGE|total|INITIALISED_STATIC|AVOID_BUG|NO_AUTHOR_SIGN_OFF|BAD_SIGN_OFF|CODE_INDENT|UNSPECIFIED_INT|EMBEDDED_FUNCTION_NAME|EXECUTE_PERMISSIONS|TRAILING|BLOCK_COMMENT_STYLE|FUNCTION_WITHOUT_ARGS|BRACES)' > types1.txt
less types1.txt
```

3. About CI. I think we can try https://github.com/marketplace/actions/clang-format-check

Regards,
Alex

Snorch:
Nice Idea, I vote for "checkpatch with --ignore"!

SWITCH_CASE_INDENT_LEVEL saves us from extra indentation levels. e.g.:

```
switch(value) {
case 1:
        printf("1\n");
        break;  
case 2:
        printf("2\n");
        break;  
default:
        BUG();  
}
```

is probably better than:

```
switch(value) {
        case 1:
                printf("1\n");
                break;
        case 2:
                printf("2\n");
                break;
        default:
                BUG();
}
```

But I don't insist.

I think we does not need to fix all errors in current codebase, only _new_ patches should pass checkpatch.

rppt:
> SWITCH_CASE_INDENT_LEVEL saves us from extra indentation levels. e.g.:

We already have this mixed, so for the new code we can require  `SWITCH_CASE_INDENT_LEVEL`

> I think we does not need to fix all errors in current codebase, only _new_ patches should pass checkpatch.

I agree that we don't need to fix all errors in the current code, but we need to decide whether we accept coding style fixes and document it.

rppt:
>     1. About CI. I think we can try https://github.com/marketplace/actions/clang-format-check

It seems that clang-format-check requires all the source files to be compliant. 

> > This action checks all C files (.c and .h) in the GitHub workspace are formatted correctly using clang-format.

We still can use checkpatch.pl for pull requests, IMHO

adrianreber:
Just saw that there is a github action based on checkpatch.pl: https://github.com/webispy/checkpatch-action

Maybe it would make sense to integrate this as a first step to make sure new code follows the coding guidelines.

adrianreber:
Now that we have the code reformatted on clang-format from #1546 I am closing this. Thanks @mihalicyn for the initial work and inspiration.

PR #1197, opened 2020-08-30
sk-unix: add fake queuers only for sockets without peer

mihalicyn:
Fake queuers needed for sockets which are connected
but has closed peer (see zdtm/static/socket_close_data(01)).
We support this for both STREAM and DRGAM sockets, but
for DGRAM sockets our check in add_fake_unix_queuers
isn't correct - we should add fake queuer only and only if
DGRAM socket has no peer.

This change not fix real visible bug, but remove incorrect and
unnecessary actions during sockets restore which may follow bug
in the future.

This code covered by zdtm/static/socket_close_data(01).c testcases.


criu/sk-unix.c
```diff
@@ -2225,8 +2225,8 @@ int add_fake_unix_queuers(void)
 	list_for_each_entry(ui, &unix_sockets, list) {
 		if ((ui->ue->uflags & (USK_EXTERN | USK_CALLBACK)) || ui->queuer)
 			continue;
-		if (!(ui->ue->state == TCP_ESTABLISHED && !ui->peer) &&
-		     ui->ue->type != SOCK_DGRAM)
+		if (!(ui->ue->type == SOCK_STREAM && ui->ue->state == TCP_ESTABLISHED && !ui->peer) &&
+		    !(ui->ue->type == SOCK_DGRAM  && !ui->peer))
```
avagin, line 2229: Where is SOCK_SEQPACKET?
do we need to add a fake queuer if there is nothing in a recv queue?

criu/sk-unix.c
```diff
@@ -2225,8 +2225,8 @@ int add_fake_unix_queuers(void)
 	list_for_each_entry(ui, &unix_sockets, list) {
 		if ((ui->ue->uflags & (USK_EXTERN | USK_CALLBACK)) || ui->queuer)
 			continue;
```
avagin, line 2227: here we check that queuer isn't nil, why do we need to check it again below?

Snorch: note: queuer != peer

Snorch:
We probably should add:

```
Fixes: 254958ce7 ("unix: Add fake queuer for standalone dgram sockets")
```

criu/sk-unix.c
Snorch: > Where is SOCK_SEQPACKET?
note: we don't support established non-stream: https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/sk-unix.c#L1828

PR #1201, opened 2020-09-09
socket/tcp: shutdown tcp sockets if the tcp-close option is set

### LINKED ISSUE #1134: redis-server takes a lot of CPU after restore with --tcp-close
Environment:
OS: CentOS 8.1 kernel 4.18
CRIU: 3.12
redis: 5.0.8

Launch a redis-server just with the default setting and then launch redis-cli to connect to the server. Make sure there is an established TCP connection.
```
redis-ser 7896 root    8u     IPv4 4888612      0t0      TCP localhost:redis->localhost:53634 (ESTABLISHED)
```

Dump redis-server with `--tcp-established`:
```
criu dump -t 7896 -D redis/ -o dump.log -v4 --tcp-established -j
```

Then restore it with `--tcp-close`:
```
criu restore -D redis/ -o restore.log -d --tcp-close -j -v4
```

You can see that redis-server takes a log of CPU.
```
PID USER      PRI  NI  VIRT   RES   SHR S CPU% MEM%   TIME+  Command
7896 root       20   0 59160  9096     0 R 49.3  0.5  4:20.12 redis-server *:6379
```

From `lsof` you can see that the originally established TCP socket was not closed correctly. 
```
redis-ser 7896 root    8u     sock     0,9      0t0  4889353 protocol: TCP
```

From `strace`, we can see redis-server keeps getting `EPOLLHUP`.
```
epoll_wait(5, [{EPOLLHUP, {u32=8, u64=8}}], 10128, 85) = 1
```

We can argue that this might be considered as a redis bug, since it should handle the `EPOLLHUP` event and close the socket fd. But I found if I applied this patch in CRIU, it also works well:

```
 diff --git a/criu/sk-tcp.c b/criu/sk-tcp.c
 index 7ee60381..b4f49fdd 100644
 --- a/criu/sk-tcp.c
 +++ b/criu/sk-tcp.c
 @@ -432,6 +432,9 @@ int restore_one_tcp(int fd, struct inet_sk_info *ii)

   if (opts.tcp_close &&
     ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {
 +   if (shutdown(fd, SHUT_RD) && errno != ENOTCONN) {
 +       pr_perror("Failed to shudown socket %d\n", fd);
 +   }
     return 0;
   }
```

Although I always got `ENOTCONN` here, kernel still eventually set the socket to `RCV_SHUTDOWN`, then it will generate event `EPOLLIN | EPOLLRDNORM | EPOLLRDHUP` instead of just `EPOLLHUP`.

```
   if (sk->sk_shutdown == SHUTDOWN_MASK || state == TCP_CLOSE)
     mask |= EPOLLHUP;
   if (sk->sk_shutdown & RCV_SHUTDOWN)
     mask |= EPOLLIN | EPOLLRDNORM | EPOLLRDHUP;
```

Then redis-server can handle it correctly.

I'm wondering is that something we can always do in CRIU when we have `--tcp-close` to avoid this kind of problems? Shall I submit a PR for this?

Thanks,

avagin:
A user process can try to write in a tcp socket and it has to get a
error in this case.

Fixes #1134


adrianreber (approved):
The `--tcp-close` logic changes look correct. Cannot say much about the use of `shutdown()`, but if it solves the reported problem and does not introduce new errors in CI I am :+1:

rst0git:
@avagin Could you please add something like this to the man page:
```diff --git a/Documentation/criu.txt b/Documentation/criu.txt
index ff6886d9..2953f618 100644
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -327,6 +327,10 @@ mount -t cgroup -o devices,freezer none devices,freezer
 *--tcp-established*::
     Checkpoint established TCP connections.
 
+*--tcp-close*::
+    Don't dump the state, or block, established tcp connections.
+    This is useful when tcp connections are not going to be restored.
+
 *--skip-in-flight*::
     This option skips in-flight TCP connections. If any TCP connections
     that are not yet completely established are found, *criu* ignores
```

Snorch (approved):
Looks good.

criu/sk-tcp.c
```diff
@@ -431,7 +435,11 @@ int restore_one_tcp(int fd, struct inet_sk_info *ii)
 	pr_info("Restoring TCP connection\n");
 
 	if (opts.tcp_close &&
-		ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {
+	    ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {
```
geminijun, line 438: I met a case that "state = TCP_CLOSE" for a socket, probably an unconnected socket dumped by function `gen_uncon_sk`

In this case, shall we just return 0 here when `state == TCP_CLOSE`, although we don't need to call `shutdown`.

Currently, it falls to `restore_tcp_conn_state` function below, which will look for the "tcp-stream.img", and failed because we don't have that image.

PR #1213, opened 2020-09-28
Trying to fix many coverity warnings

adrianreber:
I went through all open coverity warnings and tried to fix them.

### DIFF
--- a/criu/action-scripts.c
+++ b/criu/action-scripts.c
@@ -162,10 +162,18 @@ int add_script(char *path)
 
 int add_rpc_notify(int sk)
 {
+	int fd;
+
+	fd = dup(sk);
+	if (fd < 0) {
+		pr_perror("dup() failed");
+		return -1;
+	}
+
 	BUG_ON(scripts_mode == SCRIPTS_SHELL);
 	scripts_mode = SCRIPTS_RPC;
 
-	if (install_service_fd(RPC_SK_OFF, dup(sk)) < 0)
+	if (install_service_fd(RPC_SK_OFF, fd) < 0)
 		return -1;
 
 	return 0;
--- a/criu/cgroup-props.c
+++ b/criu/cgroup-props.c
@@ -319,6 +319,7 @@ static int cgp_parse_stream(char *stream, size_t len)
 					  (cgp_entry->cgp.nr_props + 1) * sizeof(char *))) {
 				pr_err("Can't allocate property for controller %s\n",
 				       cgp_entry->cgp.name);
+				xfree(p);
 				goto err_parse;
 			}
 
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1906,6 +1906,7 @@ static int rewrite_cgsets(CgroupEntry *cge, char **controllers, int n_controller
 						    cg->path + dirlen + 1);
 				if (!cg->path) {
 					cg->path = prev;
+					xfree(dirnew);
 					return -ENOMEM;
 				}
 				xfree(prev);
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1884,6 +1884,7 @@ static int rewrite_cgsets(CgroupEntry *cge, char **controllers, int n_controller
 				cg->path = xsprintf("%s%s", newroot, cg->path + cg->cgns_prefix);
 				if (!cg->path) {
 					cg->path = prev;
+					xfree(dirnew);
 					return -ENOMEM;
 				}
 				xfree(prev);
--- a/criu/filesystems.c
+++ b/criu/filesystems.c
@@ -656,7 +656,6 @@ static int dump_empty_fs(struct mount_info *pm)
 		return fd;
 
 	ret = is_empty_dir(fd);
-	close(fd);
 	if (ret < 0) {
 		pr_err("%s isn't empty\n", pm->fstype->name);
 		return -1;
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -1484,7 +1484,7 @@ static __maybe_unused int add_cr_time_mount(struct mount_info *root, char *fsnam
 
 	mi->mountpoint = xmalloc(len + strlen(path) + 1);
 	if (!mi->mountpoint)
-		return -1;
+		goto err;
 	mi->ns_mountpoint = mi->mountpoint;
 	if (!add_slash)
 		sprintf(mi->mountpoint, "%s%s", root->mountpoint, path);
@@ -1497,7 +1497,7 @@ static __maybe_unused int add_cr_time_mount(struct mount_info *root, char *fsnam
 	mi->source = xstrdup(fsname);
 	mi->options = xstrdup("");
 	if (!mi->root || !mi->fsname || !mi->source || !mi->options)
-		return -1;
+		goto err;
 	mi->fstype = find_fstype_by_name(fsname);
 
 	mi->s_dev = mi->s_dev_rt = s_dev;
@@ -1523,6 +1523,10 @@ static __maybe_unused int add_cr_time_mount(struct mount_info *root, char *fsnam
 	pr_info("Add cr-time mountpoint %s with parent %s(%u)\n",
 		mi->mountpoint, parent->mountpoint, parent->mnt_id);
 	return 0;
+
+err:
+	mnt_entry_free(mi);
+	return -1;
 }
 
 /* Returns 1 in case of success, -errno in case of mount fail, and 0 on other errors */
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -3585,6 +3585,10 @@ int mntns_get_root_fd(struct ns_id *mntns)
 {
 	if (!(root_ns_mask & CLONE_NEWNS))
 		return __mntns_get_root_fd(0);
+
+	if (!mntns)
+		return -1;
+
 	/*
 	 * All namespaces are restored from the root task and during the
 	 * CR_STATE_FORKING stage the root task has two file descriptors for
--- a/criu/net.c
+++ b/criu/net.c
@@ -1915,7 +1915,12 @@ static inline int dump_nftables(struct cr_imgset *fds)
 		return -1;
 
 	img = img_from_set(fds, CR_FD_NFTABLES);
-	img_fd = dup(img_raw_fd(img));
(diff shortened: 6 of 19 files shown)


criu/mount.c
```diff
@@ -3585,6 +3585,10 @@ int mntns_get_root_fd(struct ns_id *mntns)
 {
 	if (!(root_ns_mask & CLONE_NEWNS))
 		return __mntns_get_root_fd(0);
+
+	if (!mntns)
```
Snorch, line 3589: Do we know where can it be called with NULL? Or it is just a precaution?

Snorch (commented):
"mount: Resource leak (RESOURCE_LEAK)" Looks good.

criu/mount.c
adrianreber:
```
3632int mntns_get_root_by_mnt_id(int mnt_id)
3633{
    1. assign_zero: Assigning: mntns = NULL.
3634        struct ns_id *mntns = NULL;
3635
    2. Condition root_ns_mask & 131072, taking false branch.
3636        if (root_ns_mask & CLONE_NEWNS) {
3637                mntns = lookup_nsid_by_mnt_id(mnt_id);
3638                BUG_ON(mntns == NULL);
3639        }
3640
    CID 181217 (#1 of 1): Explicit null dereferenced (FORWARD_NULL)3. var_deref_model: Passing null pointer mntns to mntns_get_root_fd, which dereferences it. [show details]
3641        return mntns_get_root_fd(mntns);
3642}
```
That is the actual coverity warning. So it looks like it could theoretically be called with `NULL`. Not sure if it will actually happen.

Snorch: In case in mntns_get_root_by_mnt_id we take true branch of `if (root_ns_mask & CLONE_NEWNS)`, when mntns is not null. And in mntns_get_root_fd we have special branch for `if (!(root_ns_mask & CLONE_NEWNS))`:

```
if (!(root_ns_mask & CLONE_NEWNS))
        return __mntns_get_root_fd(0);
```

So it's a false positive.

But probably it is still a good thing to make coverity happy.

adrianreber: Sounds good. Thanks for the review. I kind of expected it to be a false positive, but the change to make coverity happy is not that big and also not wrong.

criu/net.c
```diff
@@ -1915,6 +1915,11 @@ static inline int dump_nftables(struct cr_imgset *fds)
 		return -1;
 
 	img = img_from_set(fds, CR_FD_NFTABLES);
+	img_fd = img_raw_fd(img);
+	if (img_fd < 0) {
+		pr_perror("Getting raw FD failed");
```
avagin, line 1920: you probably need to use pr_err in this case.

avagin (approved):

criu/cgroup.c
```diff
@@ -1906,6 +1906,8 @@ static int rewrite_cgsets(CgroupEntry *cge, char **controllers, int n_controller
 						    cg->path + dirlen + 1);
 				if (!cg->path) {
 					cg->path = prev;
+					if (dirnew)
+						xfree(dirnew);
```
avagin, line 1912: xfree can handler NULL, so you don't need if (dirnew) here

criu/pagemap.c
```diff
@@ -414,10 +414,16 @@ static int maybe_read_page_img_streamer(struct page_read *pr, unsigned long vadd
...
 {
 	unsigned long len = nr * PAGE_SIZE;
-	int fd = img_raw_fd(pr->pi);
+	int fd;
 	int ret;
 	size_t curr = 0;
 
+	fd = img_raw_fd(pr->pi);
+	if (fd < 0) {
+		pr_perror("Getting raw FD failed");
```
avagin, line 423: pr_err

criu/cgroup.c
```diff
@@ -1884,6 +1884,8 @@ static int rewrite_cgsets(CgroupEntry *cge, char **controllers, int n_controller
 				cg->path = xsprintf("%s%s", newroot, cg->path + cg->cgns_prefix);
 				if (!cg->path) {
 					cg->path = prev;
+					if (dirnew)
```
avagin, line 1887: don't need if here

criu/net.c
adrianreber: fixed

criu/cgroup.c
adrianreber: removed if

criu/cgroup.c
adrianreber: removed if

criu/pagemap.c
adrianreber: changed to pr_err

adrianreber:
I did all the changes as requested. Thanks for pushing the branch to coverity again, coverity found one bug in my changes which I also fixed.

criu/uffd.c
```diff
@@ -1466,20 +1466,28 @@ int cr_lazy_pages(bool daemon)
 	 */
 	nr_fds = task_entries->nr_tasks + (opts.use_page_server ? 2 : 1);
 	epollfd = epoll_prepare(nr_fds, &events);
-	if (epollfd < 0)
+	if (epollfd < 0) {
+		if (events)
+			xfree(events);
```
Snorch, line 1471: Looks like double free. On error epoll_prepare already frees events, isn't it?

criu/filesystems.c
```diff
@@ -656,9 +656,9 @@ static int dump_empty_fs(struct mount_info *pm)
 		return fd;
 
 	ret = is_empty_dir(fd);
-	close(fd);
 	if (ret < 0) {
 		pr_err("%s isn't empty\n", pm->fstype->name);
+		close(fd);
```
Snorch, line 661: Probably making is_empty_dir to always close an fd (even on error) and adding a comment about it would be better, in terms of code readability.

criu/cgroup-props.c
```diff
@@ -319,6 +319,7 @@ static int cgp_parse_stream(char *stream, size_t len)
 					  (cgp_entry->cgp.nr_props + 1) * sizeof(char *))) {
 				pr_err("Can't allocate property for controller %s\n",
 				       cgp_entry->cgp.name);
+				xfree(p);
 				goto err_parse;
```
Snorch, line 323: Above in the same function we have two consequent `p = get_quoted` without any free, probably it's also a problem.

criu/proc_parse.c
```diff
@@ -1372,6 +1372,8 @@ static int parse_mountinfo_ent(char *str, struct mount_info *new, char **fsname)
 	cure_path(new->root);
 
 	root_link.len = strlen(new->root);
+	if (root_link.len >= PATH_MAX - 1)
+		goto err;
 	strcpy(root_link.name, new->root);
```
Snorch, line 1377: We can print some meaningful error here about long root.

Also probably we should use strlcpy as it is "more overflow safe"? But I don't insist.

And please replace PATH_MAX with sizeof(root_link.name) which would be more robust.

criu/action-scripts.c
```diff
@@ -162,10 +162,16 @@ int add_script(char *path)
 
 int add_rpc_notify(int sk)
 {
+	int fd;
+
+	fd = dup(sk);
+	if (fd < 0)
```
Snorch, line 168: Please add an error message.

criu/util.c
```diff
@@ -334,7 +334,11 @@ void close_proc(void)
 
 int set_proc_fd(int fd)
 {
-	if (install_service_fd(PROC_FD_OFF, dup(fd)) < 0)
+	int _fd = dup(fd);
+
+	if (_fd < 0)
+		return -1;
```
Snorch, line 340: Please split _fd initialization and declaration. It does not look readable when we check error of the call from variable declarations.

Also please add proper error message here.

Snorch (changes requested):
General comment - Instead of

```
-       close(fd);
+       if (fd >= 0)
+               close(fd);
```

We can use close_safe everywhere.

criu/uffd.c
adrianreber: Yes, I guess the confusion here is in `epoll_prepare()` it is compared to '-1' and here it is '< 0'. Reading the man-page this is correct, but probably something coverity does not know about.

criu/cgroup-props.c
adrianreber: I already had a closer look at this function and I think this is now correct. Once `p` is assigned to `cgp_entry->cgp.name` it will be cleaned up using `cgp_free()`.

criu/proc_parse.c
adrianreber: I will soon push a new version. Please have a look at the error message. If it makes any sense. Happy to change it to something else.

adrianreber:
@Snorch Thanks for all the comments. I tried to address all of them.

criu/cgroup-props.c
Snorch: I missed it, then it looks good.

criu/util.c
```diff
@@ -709,26 +716,32 @@ int is_root_user(void)
...
+		goto out_closedir;
 	}
 
 	ret = 1;
-out:
+
+out_closedir:
 	closedir(fdir);
+out:
+	close_safe(&dirfd);
```
Snorch, line 744: Now you have double close. What I was thinking of is:

```
+/*
+ * is_empty_dir will always close the FD dirfd: either implicitly
+ * via closedir or explicitly in case fdopendir had failed
+ */
 int is_empty_dir(int dirfd)
 {
        int ret = 0;
@@ -716,8 +727,10 @@ int is_empty_dir(int dirfd)
        struct dirent *de;
 
        fdir = fdopendir(dirfd);
-       if (!fdir)
+       if (!fdir) {
+               close_safe(&dirfd);
                return -1;
+       }
 
        while ((de = readdir(fdir))) {
                if (dir_dots(de))
```

criu/proc_parse.c
Snorch: Thanks! The message looks good.

criu/util.c
adrianreber: Thanks. Changed and pushed.

Snorch (approved):

avagin:
Applied. Thanks!

PR #1214, opened 2020-09-29
travis: add CentOS 8 based CI run

adrianreber:
Our CentOS based CI run is based on CentOS 7. CentOS 8 exists already for some time and CentOS 7 will probably go end of life at some point.

This adds a CentOS 8 based CI run to be prepared for the time CentOS 7 goes away.


rppt (approved):

scripts/build/Makefile
```diff
@@ -1,4 +1,4 @@
-ARCHES := x86_64 fedora-asan fedora-rawhide centos armv7hf
+ARCHES := x86_64 fedora-asan fedora-rawhide centos armv7hf centos8
```
rst0git, line 1: it may be good to rename the previous CentOS target to `centos7`

rst0git (approved):

adrianreber:
Added a commit to rename centos to centos7.

.travis.yml
```diff
@@ -81,7 +81,7 @@ jobs:
       dist: bionic
     - os: linux
       arch: amd64
-      env: TR_ARCH=centos
+      env: TR_ARCH=centos7
```
avagin, line 88: do we still need to test centos7? Can we just change the version in the Dockerfile?

adrianreber: > do we still need to test centos7?

Well... It is our only test OS which still uses Python 2 as default. Just today I saw an article about the RHEL 7.9 release (which is the last release in the RHEL 7 timeframe) which mentioned that it is supported until 2024. Although CentOS 7 is old it will stay for some time and I think there will still be users on that platform for a couple of years. It is also not broken. I think if we find something that does just not work on CentOS 7 we can think about not testing it on CentOS 7 any more, but right now, except for Travis test time, it does not really have a negative effect on us.

> Can we just change the version in the Dockerfile?

No. Because of Python 3, not all necessary packages are available in EPEL, we need to enable the Powertools repository for certain packages (and probably some other things I forgot).

adrianreber:
Related to #1257. I rebased and moved both CentOS based CI runs to Github Actions.

PR #1218, opened 2020-10-01
Enable codeql security analysis

adrianreber:
Now that github announced its code analysis feature, this tries to enable it.

As far as I understand it, it is based on pre-loading libraries, so we need to build it as non-root to make sure it can actually analyse our files during the build process.

Once this is enabled the possible security related messages will be seen under 'Security' in github and once it is part of 'criu-dev' each PR will get the information if new security related problems might have been introduced.

The initial scan on my repository shows 6 possible security warnings.


adrianreber:
See https://github.com/checkpoint-restore/criu/security/code-scanning?query=is%3Aopen+ref%3Arefs%2Fpull%2F1218%2Fhead for results.

scripts/travis/vagrant.sh
```diff
@@ -6,29 +6,57 @@
...
 		dpkg -i /tmp/vagrant.deb
 
-	./apt-install libvirt-bin libvirt-dev qemu-utils qemu
-	systemctl restart libvirt-bin
+	./apt-install libvirt-clients libvirt-daemon-system libvirt-dev qemu-utils qemu \
+		ruby build-essential libxml2-dev qemu-kvm rsync ebtables dnsmasq-base \
+		openssh-client
+	systemctl restart libvirtd || :
+	/usr/sbin/libvirtd -d
+	/usr/sbin/virtlogd -d
```
rst0git, line 32: First, `systemctl restart libvirtd` would start `libvirtd` with `Requires=virtlogd.socket`.
Second, `virtlogd -d` should be before `libvirtd -d`.

adrianreber:
I forgot about this and pushed random commits here. I will close it and reopen it at #1261

PR #1219, opened 2020-10-01
Changes for v3.15

avagin:
I decided to move everything except criu-ns from the crit-dev branch.

@abhishekvijeev should we release the BPF support or should we postpone it to the v3.16?


abhishekvijeev:
Hi,

To the best of my knowledge, CRIU is fully capable of performing a c/r of map types BPF_MAP_TYPE_HASH and BPF_MAP_TYPE_ARRAY. This includes the map's data and all information read in from procfs as well as using BPF_OBJ_GET_INFO_BY_FD. The fields in struct bpf_map_info (returned by BPF_OBJ_GET_INFO_BY_FD) that are not used while restoring maps are as follows:

(1) __u32 btf_vmlinux_value_type_id
(2) __u64 netns_dev
(3) __u64 netns_ino
(4) __u32 btf_id
(5) __u32 btf_key_type_id
(6) __u32 btf_value_type_id

(2) netns_dev and (3) netns_ino are ignored because there is no way to specify these while creating BPF maps. As far as I know, these are internally allocated by the kernel and are not programmable by users. 

The remaining fields are related to BTF (BPF Type Format) - metadata that encodes debug info about maps. I am yet to understand BTF in detail and explore the possibility of c/r for BTF information. 

Therefore, unless there is a need for additional features (such as support for BTF, more map types, etc.) or any other reasons that the community finds important (because of which it would be better to postpone merging) before adding BPF support to mainline CRIU, I think we're good to go.

Thank you.

avagin:
@abhishekvijeev  can I ask you to create a wiki page to describe the current state of the BPF support? It has to be in the Under_the_hood category: 
https://criu.org/Category:Under_the_hood

abhishekvijeev:
Certainly, it would be my pleasure!
A small request - I'm settling down into a new role at work and will hence be quite busy for a few weeks. Is it alright if I start working on the documentation after a couple of weeks?

rst0git (approved):

.github/workflows/cross-compile-daily-mips.yml
```diff
@@ -0,0 +1,22 @@
+name: Daily Cross Compile Tests
```
rppt, line 1: This change can be dropped since we merge mips into master

rppt (approved):

adrianreber:
Why not `criu-ns`? There were people asking for `criu-ns` in our Ubuntu package. So just including it in releases would be nice for those people and less difference between our branches.

.github/workflows/cross-compile-daily-mips.yml
adrianreber: We would need a fix up commit on `master` to include `mips64el-cross` in `.github/workflows/cross-compile-daily.yml`. Probably not really important as we know it works on `criu-dev` and the daily build will happen on `criu-dev`.

avagin:
> Why not criu-ns? There were people asking for criu-ns in our Ubuntu package. So just including it in releases would be nice for those people and less difference between our branches.

I remember it has an issue. It forks criu into a new pid namespace, but criu can get a PID which is used by one of restored processes and criu restore will fail in this case. I think this issue has not been fixed. And we need to have a test for it.

rst0git:
> It forks criu into a new pid namespace, but criu can get a PID which is used by one of restored processes and criu restore will fail in this case. I think this issue has not been fixed.

I opened https://github.com/checkpoint-restore/criu/pull/1222 but I'm not sure how likely it is to checkpoint/restore a process with PID=2.

avagin:
@abhishekvijeev Do you want to suggest a name for this release?
https://criu.org/Download/criu

The name format is material + bird breed

abhishekvijeev:
Sure, thank you. How about 'Titanium Falcon'?

rst0git:
falcon
(Allen the peregrine falcon)

abhishekvijeev:
The color meshes perfectly with that of titanium as well!

adrianreber:
@abhishekvijeev Could you describe in one sentence what the BPF support includes (for the release notes)?

abhishekvijeev:
C/R of BPF hash maps (BPF_MAP_TYPE_HASH) and array maps (BPF_MAP_TYPE_ARRAY).

avagin:
flog has been merged by mistake. It isn't ready yet, is it?

Snorch (approved):

xemul:
@avagin , flog is not ready

mihalicyn (approved):

abhishekvijeev (approved):

PR #1222, opened 2020-10-06
criu-ns: Use PID 1 on restore

rst0git:
`criu-ns` performs double fork, which results in `criu restore` using PID 2. Thus, if a user is trying to restore a process with that PID, the restore will fail.

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2192,6 +2192,17 @@ static int write_restored_pid(void)
 	return 0;
 }
 
+static void reap_zombies(void) {
+	while (1) {
+		pid_t pid = wait(NULL);
+		if (pid == -1) {
+			if (errno != ECHILD)
+				pr_perror("Error while waiting for pids");
+			return;
+		}
+	}
+}
+
 static int restore_root_task(struct pstree_item *init)
 {
 	enum trace_flags flag = TRACE_ALL;
@@ -2445,8 +2456,9 @@ skip_ns_bouncing:
 	if (ret != 0)
 		pr_err("Post-resume script ret code %d\n", ret);
 
-	if (!opts.restore_detach && !opts.exec_cmd)
-		wait(NULL);
+	if (!opts.restore_detach && !opts.exec_cmd) {
+		reap_zombies();
+	}
 
 	return 0;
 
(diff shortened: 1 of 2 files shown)


scripts/criu-ns
```diff
@@ -83,44 +79,22 @@ def wrap_restore():
...
-		os.close(w_pipe)
-
-		if status != 0:
-			sys.exit(status)
-
-		while True:
-			try:
-				os.wait()
-			except OSError:
-				break
```
avagin, line 111: The init process has to reap any zombies and this loop did this work. Who will do it now?

avagin:
If we run CRIU as an init process, this means it should not exit after restore and should be a child reaper.

scripts/criu-ns
rst0git: `criu` does that by default (unless `--restore-detached` is used)?
```bash
# criu/cr-restore.c:2448
	if (!opts.restore_detach && !opts.exec_cmd)
		wait(NULL);
```

criu/cr-restore.c
```diff
@@ -2192,6 +2192,29 @@ static int write_restored_pid(void)
 	return 0;
 }
 
+static void reap_zombies(pid_t child_pid) {
+	pid_t current_pid;
+	int current_status;
+
+	while (1) {
+		current_pid = waitpid(-1, &current_status, WNOHANG);
```
avagin, line 2200: I don't think that we need WNOHANG here. Otherwise it will be heating the planet;).

criu/cr-restore.c
```diff
@@ -2192,6 +2192,29 @@ static int write_restored_pid(void)
...
+			if (errno == ECHILD)
+				break;  /* No child to wait */
+			pr_perror("Error while waiting for pids");
+			return;
+		}
+		if (current_pid == 0)
+			continue;  /* No child to reap */
+		if (current_pid == child_pid) {
+			if (!WIFEXITED(current_status) && !WIFSIGNALED(current_status)) {
+				/* Main child process exited for unknown reason */
```
avagin, line 2211: I really don't understand what is going on here.

criu/cr-restore.c
```diff
@@ -2192,6 +2192,29 @@ static int write_restored_pid(void)
 	return 0;
 }
 
+static void reap_zombies(pid_t child_pid) {
+	pid_t current_pid;
+	int current_status;
```
avagin, line 2197: You can drop the "current_" prefix, we don't have any other variables here. And you can move declarations in the loop.

avagin:
Applied with a small coding style fix. Thanks.

rst0git:
Thanks!

PR #1225, opened 2020-10-06
scripts/build: add a docker file for archlinux

avagin:


scripts/build/Dockerfile.archlinux
```diff
@@ -0,0 +1,45 @@
+FROM archlinux
+ARG CC=gcc
+ARG ENV1=FOOBAR
+
+RUN pacman -Sy --noconfirm \
```
rst0git, line 5: The Arch wiki recommends using `-Syu` to avoid dependency issues:
https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported

avagin: Fixed. Thanks!

rst0git (approved):

adrianreber:
Archlinux based testing was added in #1561

PR #1227, opened 2020-10-07
compel: Calculate sh_addr if not provided by linker

devdri:
GNU ld precalculates this information but lld does not. With this change, handle-elf.c calculates those addresses on its own.

When calculating addresses sections with SHF_ALLOC bit are put one after another, respecting their alignment requirements. This matches the way how the blob is constructed by copying section contents.

This solves the issue with lld that was discussed in https://github.com/checkpoint-restore/criu/pull/1121 but this time in a way that handles all known cases.

Signed-off-by: Wojciech Marczenko <marczenko@google.com>

### DIFF
--- a/compel/src/lib/handle-elf.c
+++ b/compel/src/lib/handle-elf.c
@@ -200,6 +200,27 @@ int __handle_elf(void *mem, size_t size)
 		}
 #endif
 	}
+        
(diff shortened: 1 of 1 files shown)


cyrillos (approved):

avagin:
Applied, thanks!

PR #1229, opened 2020-10-13
compel: don't mmap parasite as RWX

mclapinski:
Some kernels have W^X mitigation, which means they won't execute memory blocks if that memory block is also writable or ever was writable. This patch enables CRIU to run on such kernels.

1. Align .data section to a page.
2. mmap a memory block for parasite as RX.
3. mprotect everything after .text as RW.

I'm not sure if this is the best way to achieve this goal. If anyone has a better idea, feel free to share.

v2: add alignment for ppc64 and s390

### DIFF
--- a/compel/src/lib/handle-elf.c
+++ b/compel/src/lib/handle-elf.c
@@ -155,6 +155,7 @@ int __handle_elf(void *mem, size_t size)
 	int64_t toc_offset = 0;
 #endif
 	int ret = -EINVAL;
+	unsigned long data_off = 0;
 
 	pr_debug("Header\n");
 	pr_debug("------------\n");
@@ -698,6 +699,9 @@ int __handle_elf(void *mem, size_t size)
 				pr_out("\n\t");
 			pr_out("0x%02x,", shdata[j]);
 		}
+
+		if (!strcmp(&secstrings[sh->sh_name], ".data"))
+			data_off = sh->sh_addr;
 	}
 	pr_out("};\n");
 	pr_out("\n");
@@ -722,6 +726,7 @@ int __handle_elf(void *mem, size_t size)
 	pr_out("\tpbd->hdr.args_ptr_off	= %s_sym__export_parasite_service_args_ptr;\n", opts.prefix);
 	pr_out("\tpbd->hdr.got_off	= round_up(pbd->hdr.bsize, sizeof(long));\n");
 	pr_out("\tpbd->hdr.args_off	= pbd->hdr.got_off + %zd*sizeof(long);\n", nr_gotpcrel);
+	pr_out("\tpbd->hdr.data_off	= %#lx;\n", data_off);
 	pr_out("\tpbd->hdr.relocs		= %s_relocs;\n", opts.prefix);
 	pr_out("\tpbd->hdr.nr_relocs	= "
 			"sizeof(%s_relocs) / sizeof(%s_relocs[0]);\n",
--- a/compel/src/lib/infect.c
+++ b/compel/src/lib/infect.c
@@ -690,12 +690,11 @@ static int parasite_start_daemon(struct parasite_ctl *ctl)
 	return 0;
 }
 
-static int parasite_mmap_exchange(struct parasite_ctl *ctl, unsigned long size)
+static int parasite_mmap_exchange(struct parasite_ctl *ctl, unsigned long size, int remote_prot)
 {
 	int fd;
 
-	ctl->remote_map = remote_mmap(ctl, NULL, size,
-				      PROT_READ | PROT_WRITE | PROT_EXEC,
+	ctl->remote_map = remote_mmap(ctl, NULL, size, remote_prot,
 				      MAP_ANONYMOUS | MAP_SHARED, -1, 0);
 	if (!ctl->remote_map) {
 		pr_err("Can't allocate memory for parasite blob (pid: %d)\n", ctl->rpid);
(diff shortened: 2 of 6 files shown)


compel/src/lib/infect.c
```diff
@@ -856,15 +854,47 @@ void compel_relocs_apply(void *mem, void *vbase, struct parasite_blob_desc *pbd)
...
+
 static int compel_map_exchange(struct parasite_ctl *ctl, unsigned long size)
 {
-	int ret;
+	int ret, remote_prot;
 
-	ret = parasite_memfd_exchange(ctl, size);
+	if (ctl->pblob.hdr.data_off)
+		remote_prot = PROT_READ | PROT_EXEC;
+	else
```
rppt, line 878: Will it ever take the `else` arm?
`data_off` will be set to offset of `.data` which is always present, isn't it?

compel/src/lib/infect.c
```diff
@@ -856,15 +854,47 @@ void compel_relocs_apply(void *mem, void *vbase, struct parasite_blob_desc *pbd)
...
+		if (ret)
+			return ret;
 	}
+
+	if (!ctl->pblob.hdr.data_off)
+		return 0;
+
+	ret = remote_mprotect(ctl, ctl->remote_map + ctl->pblob.hdr.data_off,
+			      size - ctl->pblob.hdr.data_off,
+			      PROT_READ | PROT_WRITE);
```
rppt, line 894: I afraid this will break all architectures but x86. None of `*lds.S` align `.data` section of a page boundary, and that will casuse `mprotect()` to fail.

We either need to make sure `.data` is always page aligned or maybe add a dummy section in the linker script. This section will be page aligned and whenever `__handle_elf` sees such section, it will set `.data_off` to the address of that section.

compel/src/lib/infect.c
mclapinski: Yes, for example on arm and aarch64 the `.data` section is inside another section (`.crblob`), so the `handle-elf.c` doesn't find it.  On mips it looks like `.data` is inside `.text`? ppc64, s390 and x86 looks like I'd expect them to. So I could align it on ppc64, s390 and x86 while on the other architectures it would work as before. What do you think?

0x7f454c46:
JFYI: Last time I've tried to do it - I ended with:
https://patchwork.criu.org/patch/3680/
But that still wasn't working on ppc, so this will need a run of zdtm tests on non-x86 architectures.

compel/src/lib/infect.c
rppt: I think it could work, just ppc64 would need `.data` aligned at 64Kb as this is the default page size on ppc64.
Yet, arm and arm64 may also have kernels with W^X enabled, so it would be nice to convert them as well :)

rppt:
> But that still wasn't working on ppc

I believe that with

```
. = ALIGN(0x10000);
```
it would work on ppc64

> so this will need a run of zdtm tests on non-x86 architectures.

+1

adrianreber:
I can run zdtm on all architectures except MIPS and arm32 if necessary. Let me know if that would help.

compel/src/lib/handle-elf.c
```diff
@@ -698,6 +699,9 @@ int __handle_elf(void *mem, size_t size)
 				pr_out("\n\t");
 			pr_out("0x%02x,", shdata[j]);
 		}
+
+		if (!strcmp(&secstrings[sh->sh_name], ".data"))
```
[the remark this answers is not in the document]
mclapinski: I'm not sure what you mean. Would you prefer `strncmp(&secstrings[sh->sh_name], ".data", 6)`? But a string literal is always null terminated.

compel/src/lib/infect.c
mclapinski: Done for ppc64 and s390. Unfortunately I don't understand this code well enough to implement it for arm (especially without an access to it).

rppt (approved):
Looks ok to me.
Let's wait for  @adrianreber  to confirm this does not break ppc64 or s390

adrianreber:
Let me set up zdtm runs on those arches. It takes some time as it is all manual.

rppt:
@0x7f454c46 do you remember why is there .crblob at all on arm/arm64?
It does not seem necessary and I think both arm's linker scripts can be made inline with x86/ppc/s390

0x7f454c46:
Looking at the history, it was .crblob for all arches before 0be8beb666ab.
I think it can and maybe should be aligned with x86/ppc/s390.

0x7f454c46:
See also historical 662bfd5fbc62

adrianreber:
Without the patch I see on aarch64, s390x and ppc64le all RHEL 8.2:
```
################### 2 TEST(S) FAILED (TOTAL 390/SKIPPED 37) ####################
 * zdtm/static/pty00(h)
 * zdtm/static/netns-nft(uns)
```
I guess the pty00 error might come from running it through ansible and the nft error is because there is no libnft-devel installed.

With the patch applied I get exactly the same result from ZDTM. So this patch does not cause any ZDTM regressions on aarch64, s390x and ppc64le.

mclapinski:
@adrianreber thanks so much!

0x7f454c46:
@mclapinski thanks! That looks good, merged.
I'll open a github issue to align arm/arm64.
Also, it may make sense to mprotect() RO parasite .text in local (criu) mapping (after applying relocations).

PR #1235, opened 2020-10-14
Even more static code analysis fixes

adrianreber:
This tries to fix even more reports from different static code analyzers (coverity, cppcheck, scan-build).

Some of the 'fixes' are just to mark certain things as false positive, so that the checkers can ignore it the next time.

CI seems happy with the changes, but as it touches many different parts I hope I did not break something.

### DIFF
--- a/compel/src/lib/infect.c
+++ b/compel/src/lib/infect.c
@@ -355,6 +355,7 @@ static int gen_parasite_saddr(struct sockaddr_un *saddr, int key)
 	int sun_len;
 
 	saddr->sun_family = AF_UNIX;
+	/* cppcheck-suppress nullPointer */
 	snprintf(saddr->sun_path, UNIX_PATH_MAX,
 			"X/crtools-pr-%d", key);
 
--- a/criu/autofs.c
+++ b/criu/autofs.c
@@ -1114,6 +1114,7 @@ close_pipe:
 free_info:
 	free(info);
 umount:
+	/* coverity[toctou] */
 	if (umount(mi->mountpoint) < 0)
 		pr_perror("Failed to umount %s", mi->mountpoint);
 	goto close_pipe;
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -3633,7 +3633,13 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 	 * it gets unmapped at the very end of __export_restore_task
 	 */
 
-	task_args->proc_fd = dup(get_service_fd(PROC_FD_OFF));
+	i = get_service_fd(PROC_FD_OFF);
+	if (i < 0) {
+		pr_err("Cannot get PROC_FD_OFF fd\n");
+		goto err;
+	}
+
+	task_args->proc_fd = dup(i);
 	if (task_args->proc_fd < 0) {
 		pr_perror("can't dup proc fd");
 		goto err;
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -3240,7 +3240,7 @@ static bool groups_match(gid_t* groups, int n_groups)
 	n = getgroups(0, NULL);
 	if (n == -1) {
 		pr_perror("Failed to get number of supplementary groups");
-		ret = false;
+		return false;
 	}
 	if (n != n_groups)
 		return false;
--- a/criu/fdstore.c
+++ b/criu/fdstore.c
@@ -110,6 +110,11 @@ int fdstore_get(int id)
 	int sk = get_service_fd(FDSTORE_SK_OFF);
 	int fd;
 
+	if (sk < 0) {
+		pr_err("Cannot get FDSTORE_SK_OFF fd\n");
+		return -1;
+	}
+
 	mutex_lock(&desc->lock);
 	if (setsockopt(sk, SOL_SOCKET, SO_PEEK_OFF, &id, sizeof(id))) {
 		mutex_unlock(&desc->lock);
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -341,6 +341,7 @@ static int kerndat_get_dirty_track(void)
 
 	map[0] = '\0';
 
+	/* coverity[check_return] */
 	lseek(pm2, (unsigned long)map / PAGE_SIZE * sizeof(u64), SEEK_SET);
 	ret = read(pm2, &pmap, sizeof(pmap));
 	if (ret < 0)
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -2781,6 +2781,7 @@ err_tmpfs:
 	}
 
 err_root:
+	/* coverity[toctou] */
 	if (tmp_dir && rmdir(put_root)) {
 		pr_perror("Can't remove the directory %s", put_root);
 		return -1;
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -2852,6 +2852,7 @@ static int create_mnt_roots(void)
 		mnt_roots = NULL;
 		goto out;
 	}
+	/* coverity[check_return] */
 	chmod(mnt_roots, 0777);
 
 	exit_code = 0;
--- a/criu/namespaces.c
+++ b/criu/namespaces.c
@@ -526,7 +526,7 @@ static int open_ns_fd(struct file_desc *d, int *new_fd)
 	struct ns_desc *nd = NULL;
 	struct ns_id *ns;
 	int nsfd_id, fd;
-	char path[64];
+	char path[64] = "";
 
 	for (ns = ns_ids; ns != NULL; ns = ns->next) {
 		if (ns->id != nfi->nfe->ns_id)
--- a/criu/namespaces.c
+++ b/criu/namespaces.c
@@ -1366,6 +1366,10 @@ int __userns_call(const char *func_name, uns_call_t call, int flags,
 		return call(arg, fd, getpid());
 
 	sk = get_service_fd(USERNSD_SK);
+	if (sk < 0) {
+		pr_err("Cannot get USERNSD_SK fd\n");
+		return -1;
+	}
 	pr_debug("uns: calling %s (%d, %x)\n", func_name, fd, flags);
 
 	if (!async)
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -151,6 +151,7 @@ static void skip_pagemap_pages(struct page_read *pr, unsigned long len)
 static int seek_pagemap(struct page_read *pr, unsigned long vaddr)
 {
 	if (!pr->pe)
+		/* coverity[branch_past_initialization] */
 		goto adv;
 
 	do {
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -1381,6 +1381,7 @@ static int parse_mountinfo_ent(char *str, struct mount_info *new, char **fsname)
 	}
 	strcpy(root_link.name, new->root);
 	if (strip_deleted(&root_link)) {
+		root_link.name[sizeof(root_link.name) - 2] = 0;
 		strcpy(new->root, root_link.name);
 		new->deleted = true;
 	}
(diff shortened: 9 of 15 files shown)


criu/vdso-compat.c
```diff
@@ -13,6 +13,9 @@ static void exit_on(int ret, int err_fd, char *reason)
 	if (ret) {
 		syscall(__NR_write, err_fd, reason, strlen(reason));
 		syscall(__NR_exit, ret);
+#ifdef __COVERITY__
+		__coverity_panic__();
+#endif
```
0x7f454c46, line 18: Can we use `__noreturn` in the function definition instead?

adrianreber: I just tried to add `__attribute__((__noreturn__))` and gcc says:
```
  CC       criu/vdso-compat.o
criu/vdso-compat.c: In function ‘exit_on’:
criu/vdso-compat.c:22:1: error: ‘noreturn’ function does return [-Werror]
   22 | }
      | ^
```

Any other ways to use `__noreturn__`?

0x7f454c46: Ah, yeah, I've missed that it returns on `(!ret)`, sorry.
Maybe like in kenel `unreachable()` rather than this coverity iffdeffery, what do you think?

adrianreber: Using `__builtin_unreachable();` seems to work.

include/common/lock.h
```diff
@@ -9,10 +9,12 @@
 #include "common/asm/atomic.h"
 #include "common/compiler.h"
 
+#ifndef __clang_analyzer__
```
avagin: need to add a comment that explains why we need this.

compel/src/lib/infect.c
```diff
@@ -105,7 +105,7 @@ static int parse_pid_status(int pid, struct seize_task_status *ss, void *data)
 int compel_stop_task(int pid)
 {
 	int ret;
-	struct seize_task_status ss;
+	struct seize_task_status ss = {0};
```
avagin, line 108: ss = {};

criu/bfd.c
```diff
@@ -198,7 +198,6 @@ char *breadchr(struct bfd *f, char c)
 
 		if (b->sz == BUFSIZE) {
 			pr_err("The bfd buffer is too small\n");
-			ERR_PTR(-EIO);
```
avagin, line 201: Looks like it should be return ERR_PTR(-EIO);

criu/autofs.c
```diff
@@ -331,12 +331,14 @@ static int autofs_revisit_options(struct mount_info *pm)
 {
 	FILE *f;
 	char *str;
+	char *org;
```
avagin, line 334: fgets should get a pointer to the whole buffer:

char *buf;
....
while (fgets(buf, 1024, f)) {
  char *str = buf;

criu/kerndat.c
```diff
@@ -341,6 +341,7 @@ static int kerndat_get_dirty_track(void)
 
 	map[0] = '\0';
 
+	/* coverity[check_return] */
 	lseek(pm2, (unsigned long)map / PAGE_SIZE * sizeof(u64), SEEK_SET);
```
avagin, line 345: Here we probably need to check a return code or we need to add a comment why we don't need to do this here

criu/mount.c
```diff
@@ -2850,6 +2852,7 @@ static int create_mnt_roots(void)
 		mnt_roots = NULL;
 		goto out;
 	}
+	/* coverity[check_return] */
 	chmod(mnt_roots, 0777);
```
avagin, line 2857: need to check the return code

avagin:
> autofs: ignore CHECKED_RETURN

It isn't only about autofs

criu/cr-dump.c
```diff
@@ -946,6 +946,7 @@ static int dump_signal_queue(pid_t tid, SignalQueueEntry **sqe, bool group)
 			} else
 				pr_perror("ptrace");
 
+			xfree(si);
```
avagin, line 951: there are two other places where we need to call xfree(si).

criu/protobuf.c
```diff
@@ -66,6 +66,10 @@ int do_pb_read_one(struct cr_img *img, void **pobj, int type, bool eof)
 		ret = 0;
 	else
 		ret = bread(&img->_x, &size, sizeof(size));
+
+	/* This is to make coverity happy and to not report TAINTED_SCALAR. */
+	size &= UINT32_MAX;
```
avagin, line 71: This is not a false positive, but I am not sure that we want to care about this. But I don't like this hack in the code and I think we can just ignore this issue for now.

criu/pstree.c
```diff
@@ -674,12 +674,12 @@ static int prepare_pstree_ids(pid_t pid)
 		leader = pstree_item_by_virt(item->sid);
 		BUG_ON(leader == NULL);
 		if (leader->pid->state != TASK_UNDEF) {
-			pid_t pid;
+			pid_t __pid;
 
-			pid = get_free_pid();
-			if (pid < 0)
+			__pid = get_free_pid();
```
avagin, line 679: IMHO __pid doesn't look nice in this case.  helper_pid will look better.

criu/pstree.c
```diff
@@ -771,15 +771,15 @@ static int prepare_pstree_ids(pid_t pid)
 
 	/* Add a process group leader if it is absent  */
 	for_each_pstree_item(item) {
-		struct pid *pid;
+		struct pid *__pid;
```
avagin, line 774: maybe pgid?

criu/uffd.c
```diff
@@ -1485,6 +1485,5 @@ int cr_lazy_pages(bool daemon)
 
 	tls_terminate_session();
 
-	xfree(events);
```
avagin, line 1488: could you explain in the commit message why it is wrongly added?

I see it is allocated in epoll_prepare, but I don't see where we release it.

criu/cr-restore.c
```diff
@@ -3240,7 +3240,7 @@ static bool groups_match(gid_t* groups, int n_groups)
 	n = getgroups(0, NULL);
 	if (n == -1) {
 		pr_perror("Failed to get number of supplementary groups");
-		ret = false;
+		return false;
```
avagin, line 3243: should we return an error in this case?

criu/sk-unix.c
```diff
@@ -1170,6 +1170,7 @@ static int revert_unix_sk_cwd(struct unix_sk_info *ui, int *prev_cwd_fd, int *ro
 	if (*ns_fd >= 0 && restore_ns(*ns_fd, &mnt_ns_desc))
 		ret = -1;
 	if (*root_fd >= 0) {
+		/* coverity[chroot_call] */
```
avagin, line 1174: I don't understand what warnings  mean in the commit message. Could you add a more detailed explanation and add a link to the coverity issue? Thanks!

criu/namespaces.c
```diff
@@ -526,7 +526,7 @@ static int open_ns_fd(struct file_desc *d, int *new_fd)
 	struct ns_desc *nd = NULL;
 	struct ns_id *ns;
 	int nsfd_id, fd;
-	char path[64];
+	char path[64] = "";
```
avagin, line 529: I think it would be better to print a proper error if fdstore_get fails and remove the check_open label.

criu/pagemap.c
```diff
@@ -246,6 +246,10 @@ static int read_local_page(struct page_read *pr, unsigned long vaddr,
 	ssize_t ret;
 	size_t curr = 0;
 
+	if (fd < 0) {
```
avagin, line 250: fd = img_raw_fd(pr->pi);
if (fd < 0) {

criu/proc_parse.c
```diff
@@ -853,7 +853,7 @@ int parse_pid_stat(pid_t pid, struct proc_pid_stat *s)
 	if (fd < 0)
 		return -1;
 
-	n = read(fd, buf, BUF_SIZE);
+	n = read(fd, buf, BUF_SIZE - 1);
```
avagin, line 858: buf[BUF_SIZE] is always 0. Pls look at its declaration:

struct buffer {
        char buf[BUF_SIZE];
        char end; /* '\0' */
};

static struct buffer __buf;
static char *buf = __buf.buf;

criu/cr-dump.c
adrianreber: I already looked at this one a lot. The two other places are actually only one place because the first break (line 970) will also go through the second break (line 982). The reason why I did not include it is that the for loop (line 963) can end any time but the pointer to `si` is being changed by the loop. As this seems to be error code path it means we should probably run `free()` on the original pointer and not on the one after `si++`, right?

criu/cr-restore.c
adrianreber: Good question. I do not know. If we already print out errno, so it seems it is an error. The man-page says:
`On error, -1 is returned, and errno is set appropriately.`. This was just recently added:
```
commit 0d6b8be72521232261ec1872522fb6977e0a4085
Author: Angie Ni <avtni@google.com>
Date:   Fri Jun 12 10:20:28 2020 -0600

    When groups already have desired values, we skip calling setgroups.
    Furthermore, this check allows us to restore when in an
    unprivileged user namespace that cannot use setgroups.
```

criu/uffd.c
adrianreber: It's complicated. In certain error cases `events` could have been already free'd in xrealloc (, but it is very unlikely. It does not seem to be able to handle this at this level for error cases correctly. I will just drop this one.

Re-reading the man-page of `realloc` this patch might indeed be wrong, because it says, that if it fails it will not change `events`. So it can be dropped.

adrianreber:
@avagin I tried to address all your comments, by changing the code, the commit message, dropping it or trying to answer open questions.

criu/kerndat.c
adrianreber: If I check the return code, CRIU fails with:
```
+ ./criu/criu check
Error (criu/kerndat.c:346): Seek pmap err!: No such file or directory
Error (criu/kerndat.c:1084): kerndat_get_dirty_track failed when initializing kerndat.
Error (criu/crtools.c:213): Could not initialize kernel features detection.
```
Not sure why we do it if it fails.

avagin:
@adrianreber could you rebase this pr? I pushed some of these commits.

adrianreber:
@avagin Good idea to merge some commits. I rebased the PR.

adrianreber:
Rebased and force pushed

criu/fdstore.c
```diff
@@ -110,6 +110,11 @@ int fdstore_get(int id)
 	int sk = get_service_fd(FDSTORE_SK_OFF);
 	int fd;
 
+	if (sk < 0) {
```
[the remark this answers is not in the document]
adrianreber: Sure. I think it that change already in some other location. Will change it.

criu/uffd.c
avagin:
```
index 670ef158d..2aeb53b43 100644
--- a/criu/util.c
+++ b/criu/util.c
@@ -1312,6 +1312,7 @@ int epoll_prepare(int nr_fds, struct epoll_event **events)
 
 free_events:
        xfree(*events);
+       *events = NULL;
        return -1;
 }
 
```

criu/cr-restore.c
```diff
@@ -3633,7 +3633,13 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 	 * it gets unmapped at the very end of __export_restore_task
 	 */
 
-	task_args->proc_fd = dup(get_service_fd(PROC_FD_OFF));
+	i = get_service_fd(PROC_FD_OFF);
```
avagin, line 3636: pls don't use random variables here. This doesn't look right. It is better to wrap this in a function.

adrianreber:
I will close this and open a PRs for each patch for easier review.

PR #1248, opened 2020-10-28
Fixup external pidns functionality and make code more consistent

Snorch:


Snorch:
@adrianreber please take a look, as it touches external pidnses and clone3(set_tid) paths.

adrianreber:
This looks so much nicer than what I did. Thanks. Trying to restore a process out of PID namespaces into the host PID namespace fails now, however, with:
```
Error (criu/cr-restore.c:1414): Unable to restore into an empty PID namespace
```
Which sounds wrong, because I try to restore into the host namespace:
```
# criu/criu dump -t `pidof minimal` -D /tmp/ -j --external pid[4026532253]:exti
# exec 33< /proc/self/ns/pid
# criu/criu restore -D /tmp -v4 -j --inherit-fd fd[33]:exti
```

Snorch:
Hi, Adrian! Thanks for your response!

I tried to do:

```
[root@snorch criu]# cat check.sh 
#!/bin/bash
killall -9 sleep
setsid sleep 1000 &>/dev/null </dev/null &
PID=$(pidof sleep)
PIDNS=$(readlink /proc/$PID/ns/pid)
exec 100<&-
exec 100< /proc/self/ns/pid

rm -rf /tmp/sleep-dump
mkdir /tmp/sleep-dump

criu/criu dump -t $PID -D /tmp/sleep-dump/ --external $PIDNS:exti
criu/criu restore -D /tmp/sleep-dump/ -v4 --restore-detached --inherit-fd fd[100]:exti

cat /proc/$PID/stack
```

And everything works fine. What exact program you dump?

```
...
(00.003757) Running pre-resume scripts
(00.003764) Restore finished successfully. Tasks resumed.
(00.003776) Writing stats
(00.003818) Running post-resume scripts
[<0>] hrtimer_nanosleep+0xbc/0x190
[<0>] common_nsleep+0x40/0x50
[<0>] __x64_sys_clock_nanosleep+0xb0/0x110
[<0>] do_syscall_64+0x52/0x90
[<0>] entry_SYSCALL_64_after_hwframe+0x44/0xa9
```

According to the error

> Error (criu/cr-restore.c:1414): Unable to restore into an empty PID namespace

It looks like you try to restore process with pid==1 in host's pidns and this pid is already used, so it is ok to error.

---
Note: I will fix "cr-restore: setup external pidns only for root task" a little bit, because I need to also move under (!current) two other ext_key checks.

adrianreber:
You are right. I tested it wrongly. I am now using:
```
# unshare -p -f sh -c 'echo -n 30000 > /proc/sys/kernel/ns_last_pid; setsid sleep infinity'
# criu/criu dump -t `pidof sleep` -D /tmp/ -j --external pid[4026532253]:exti
# exec 33< /proc/self/ns/pid
# criu/criu restore -D /tmp -v4 -j --inherit-fd fd[33]:exti
```
Now it works.

adrianreber (approved):
Really nice code flow changes.

Snorch:
More over even nested pidns to root pidns case works fine for me:

```
[root@snorch criu]# cat pidns-check.sh 
#!/bin/bash
killall -9 sleep

rm -rf /tmp/sleep-dump
mkdir /tmp/sleep-dump /tmp/sleep-dump/proc

unshare -p --fork sh -c 'mount -t proc proc /tmp/sleep-dump/proc;
			 echo 100000 > /tmp/sleep-dump/proc/sys/kernel/ns_last_pid;
			 umount /tmp/sleep-dump/proc;
			 setsid sleep 1000 &>/dev/null </dev/null' &>/dev/null </dev/null &

while :; do
	PID=$(pidof sleep)
	if [ ! -z "$PID" ]; then
		break
	fi
done
echo "PID=$PID"

PIDNS=$(readlink /proc/$PID/ns/pid | sed 's/://')
exec 100<&-
exec 100< /proc/$$/ns/pid

BEFORE=$(grep NSpid /proc/$PID/status)

criu/criu dump -t $PID -D /tmp/sleep-dump/ --external $PIDNS:exti
criu/criu restore -D /tmp/sleep-dump/ --restore-detached --inherit-fd fd[100]:exti

echo $BEFORE
grep NSpid /proc/$(pidof sleep)/status
```
```
[root@snorch criu]# bash pidns-check.sh 
PID=94681
Warn  (compel/arch/x86/src/lib/infect.c:281): Will restore 94681 with interrupted system call
NSpid: 94681 100002
NSpid:	100002
```

Though it took quiet a long time for me to write it in bash =)

adrianreber:
> More over even nested pidns to root pidns case works fine for me:

Cool. Thanks for verifying.
 
> Though it took quiet a long time for me to write it in bash =)

I can totally understand that :wink:. Maybe it would make sense to include your test script also in `test/other/ns_ext`.

Snorch:
Looking on `test/others/ns_ext/run.sh` I see that the only major difference between my script(pidns-check.sh) and this one is that I restore in host's pidns, but I think restoring in host pidns even with BIG pid is quiet racy, and your test covers almost exactly the same case, so probably we don't need my version. Though we can just add the script in the directory without adding automatics to it just for memory if you like, so I did it.

adrianreber:
Sounds like a good solution.

Snorch:
Added an issue for this strange test fail https://github.com/checkpoint-restore/criu/issues/1280

Snorch:
I've updated to a new criu-dev, as it now has fixes/debugs to a triggered travis problem.

Snorch:
Another test error:
```
=================== Run zdtm/static/socket-tcp6-local in uns ===================
Start test
./socket-tcp6-local --pidfile=socket-tcp6-local.pid --outfile=socket-tcp6-local.out
Run criu dump
Run criu restore
5: Old files lost: set([u'1', u'0', u'3', u'2', u'5', u'4'])
5: New files appeared: set([])
############ Test zdtm/static/socket-tcp6-local FAIL at fds compare ############
```
It could be c/r restored fds wrong, or one of socket-tcp6-local processes (pid 5) died unexpectedly after c/r. Need full logs of criu and test to investigate it, it does not reproduce on my machine in multiple runs.

avagin:
Overall this looks good for me, but we need to improve the test:

```
before c/r: NSpid:	4707	100003
ok
ok
CRIU restore failed
FAIL
umount: pidns_proc: no mount point specified.
```

avagin:
@Snorch Here is a draft fix for this test https://github.com/avagin/criu/commit/001801363898145917ea1e8fd2b879404f033362

* use pipe-s to synchronize processes
* run the test in a pid namespace to avoid pid conflicts in the host pid namespace
* grep errors in restore.log

Snorch:
>@Snorch Here is a draft fix for this test avagin@0018013
>
>    use pipe-s to synchronize processes
>    run the test in a pid namespace to avoid pid conflicts in the host pid namespace
>    grep errors in restore.log

Your changes are nice, will merge them, Thanks!

Snorch:
I've changed 100000 to 1000 for ns_last_pid, as the bigger does not work on all environments (especially the ones with low pid_max).

https://travis-ci.org/github/Snorch/criu/jobs/748360070

```
unshare -fp -m --mount-proc --propagation private ./run_pidns.sh
_run_pidns.sh: line 13: echo: write error: Invalid argument
```

Snorch:
Thanks!

PR #1254, opened 2020-11-03
Distribute CI jobs between Travis and GitHub Actions

adrianreber:
This renames 'travis' to 'ci' all over the source tree. The motivation for this is that we already use Travis and GitHub Actions so that the name 'travis' is already misleading.

This PR also moves then four more CI runs from Travis to GitHub Actions.

This PR is also motivated by https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing. It is not totally clear to me if this will be relevant for CRIU, but it sounds like, if I am not too confused about the difference between travis-ci.org and travis-ci.com, the number of minutes per project per month might soon be limited, so that we need to look at additional CI services to distribute our load.

I already tried to run CI on drone.io and circleci.com and it seems to be possible to run CI jobs on all four platforms (Travis, GitHub Actions, CircleCI and Drone) for CRIU.

It still seems Travis has the best setup for CRIU tests as it provides bare metal access, but I was also able to run tests using Circle CI on bare metal hosts. Drone only seems to provide container based testing, but also for ARM.

I also tried a self-hosted GitHub Action runner, but it is not recommended for public projects, because each PR can result in running arbitrary code (whatever is in the PR) on the self hosted runner and there is no real easy way to clean up self-hosted runners after each CI run.


0x7f454c46:
There is also cirrus-ci.org which seems to provide VMs free of charge for opensource.

adrianreber:
> There is also cirrus-ci.org which seems to provide VMs free of charge for opensource.

Thanks for the pointer. After about 20 test runs I am able to boot a vagrant based Fedora 32 VM on Cirrus CI. So it seems we can move all of our tests somewhere else. Until one of the other services decides to reduce the free minutes for FOSS.

PR #1263, opened 2020-11-07
criu-ns: Remove unreachable statement

rst0git:


adrianreber (approved):
I saw this also in a run of lgtm.com. Thanks for fixing.

PR #1264, opened 2020-11-08
uffd: fix 'double free detected in tcache 2'

adrianreber:
This xfree() which was added during static code analysis fixes breaks one of the tests with a double free error message:

free(): double free detected in tcache 2

### DIFF
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -1254,18 +1254,18 @@ static void lazy_pages_summary(struct lazy_pages_info *lpi)
 #endif
 }
 
-static int handle_requests(int epollfd, struct epoll_event *events, int nr_fds)
+static int handle_requests(int epollfd, struct epoll_event **events, int nr_fds)
 {
 	struct lazy_pages_info *lpi, *n;
 	int poll_timeout = -1;
 	int ret;
 
 	for (;;) {
-		ret = epoll_run_rfds(epollfd, events, nr_fds, poll_timeout);
+		ret = epoll_run_rfds(epollfd, *events, nr_fds, poll_timeout);
 		if (ret < 0)
 			goto out;
 		if (ret > 0) {
-			ret = complete_forks(epollfd, &events, &nr_fds);
+			ret = complete_forks(epollfd, events, &nr_fds);
 			if (ret < 0)
 				goto out;
 			if (restore_finished)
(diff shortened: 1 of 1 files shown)


rppt (approved):
I don't see why there would be a double free, but since return from `cr_lazy_pages()` is essentially `exit()` we don't really have to call free there.

avagin:
complete_forks calls xrealloc, so it can change the events pointer, but we will not see a new value in complete_forks.

BTW: complete_forks leaks events if xrealloc fails.

criu/uffd.c
```diff
@@ -1485,6 +1485,5 @@ int cr_lazy_pages(bool daemon)
 
 	tls_terminate_session();
 
-	xfree(events);
```
avagin, line 1488: I think it would be good to have a comment to explain why we don't free events after handle_requests.

nit: we can move this xfree in handle_requests.

adrianreber:
I changed this PR to pass `&events` to `handle_requests()` instead of `events`. As `xrealloc()` can change `events` it can now be correctly freed by `xfree()`.

adrianreber:
The re-enabled lazy tests are now failing (only on alpine) with:
```
b'(00.075306) Error (criu/tls.c:147): tls: Failed receiving data: Error in the pull function.'
b"(00.075323) Error (criu/page-xfer.c:1225): page-xfer: Can't read pagemap from socket: I/O error"
```
@rst0git any ideas why this is happening?

rst0git:
Hi Adrian, I can see the error in the Travis logs but I can't reproduce it on my local machine.
```
b'(00.000652) Running status-ready scripts'
b'(00.029320) Accepted connection from 127.0.0.1:53068'
b'(00.050112) tls: TLS handshake completed'
```
This implies that the TLS session was successfully established.

```
Error (criu/tls.c:147): tls: Failed receiving data: Error in the pull function.'
```
This error implies that `recv()`, invoked within `_tls_pull_cb()`, failed for some reason.

I restarted the test in Travis and it is passing now.

rst0git (approved):

adrianreber:
@rst0git Thanks for taking a look at it.

adrianreber:
@avagin Can this be merged. CI is currently broken on a couple of targets.

avagin:
And we need to fix this:
https://github.com/checkpoint-restore/criu/blob/b9f980418ed8e6868924a5f2a0cebda193363f6b/criu/uffd.c#L1141

PR #1272, opened 2020-11-12
Move first job to Cirrus CI

adrianreber:
This includes the changes to run the Vagrant based CI run on Cirrus CI. Cirrus CI seems to be one of the few (only) services which enable nested virtualization besides Travis. With the upcoming shutdown of travis-ci.org (end of 2020) and the necessary move the travis-ci.com I am trying to move all long running jobs away from Travis. To activate the Cirrus CI based job the Cirrus CI integration into checkpoint-restore/criu is necessary. Let me know if I should do that. Once we know it works on Cirrus CI I would disable it on Travis.

This is also includes a commit to prepare our aarch64 based run to work on aws-graviton2 based system. That would be the first non x86_64 based Travis CI run doing actual tests instead of just building CRIU. This is only available on travis-ci.com. If we actually enable this depends if we have enough minutes on travis-ci.com once we do the switch.

### DIFF
--- a/scripts/ci/vagrant.sh
+++ b/scripts/ci/vagrant.sh
@@ -6,29 +6,55 @@
 set -e
 set -x
 
-VAGRANT_VERSION=2.2.7
+VAGRANT_VERSION=2.2.10
 FEDORA_VERSION=32
 FEDORA_BOX_VERSION=32.20200422.0
 
 setup() {
-	apt-get -qq update
-	# Load the kvm modules for vagrant to use qemu
-	modprobe kvm kvm_intel
+	if [ -n "$TRAVIS" ]; then
+		# Load the kvm modules for vagrant to use qemu
+		modprobe kvm kvm_intel
+	fi
+	if [ -n "$CIRRUS_CI" ]; then
+		# Running modprobe is not possible on Cirrus, because
+		# we are running in a container with potentially other
+		# modules than the host.
+		# Vagrant can still use /dev/kvm later if we do
+		chmod 666 /dev/kvm
+	fi
 
 	# Tar up the git checkout to have vagrant rsync it to the VM
 	tar cf criu.tar ../../../criu
-	wget https://releases.hashicorp.com/vagrant/${VAGRANT_VERSION}/vagrant_${VAGRANT_VERSION}_"$(uname -m)".deb -O /tmp/vagrant.deb && \
+	# Cirrus has problems with the following certificate.
+	wget --no-check-certificate https://releases.hashicorp.com/vagrant/${VAGRANT_VERSION}/vagrant_${VAGRANT_VERSION}_"$(uname -m)".deb -O /tmp/vagrant.deb && \
 		dpkg -i /tmp/vagrant.deb
 
-	./apt-install libvirt-bin libvirt-dev qemu-utils qemu
-	systemctl restart libvirt-bin
+	./apt-install libvirt-clients libvirt-daemon-system libvirt-dev qemu-utils qemu \
+		ruby build-essential libxml2-dev qemu-kvm rsync ebtables dnsmasq-base \
+		openssh-client
+	if [ -n "$CIRRUS_CI" ]; then
+		# On Cirrus systemctl does not work, because we are running in
+		# a container without access to systemd
+		/usr/sbin/virtlogd -d
+		/usr/sbin/libvirtd -d
+	else
+		systemctl restart libvirtd
+	fi
 	vagrant plugin install vagrant-libvirt
 	vagrant init fedora/${FEDORA_VERSION}-cloud-base --box-version ${FEDORA_BOX_VERSION}
 	# The default libvirt Vagrant VM uses 512MB.
 	# Travis VMs should have around 7.5GB.
 	# Increasing it to 4GB should work.
 	sed -i Vagrantfile -e 's,^end$,  config.vm.provider :libvirt do |libvirt|'"\n"'    libvirt.memory = 4096;end'"\n"'end,g'
-	vagrant up --provider=libvirt
+	if [ -n "$CIRRUS_CI" ]; then
+		# Work around for:
+		# Error while activating network: Call to virNetworkCreate failed: internal error:
+		# Failed to apply firewall rules /usr/sbin/ip6tables --table filter --list-rules: modprobe: FATAL: Module ip6_tables not found in directory /lib/modules/5.4.0-1025-gcp
+		# On cirrus-ci.com. Running in a container without access to the host's kernel modules
+		rm -f /sbin/ip6tables
+		cp /bin/true /sbin/ip6tables
+	fi
+	vagrant up --provider=libvirt > /dev/null
 	mkdir -p /root/.ssh
 	vagrant ssh-config >> /root/.ssh/config
 	ssh default sudo dnf install -y gcc git gnutls-devel nftables-devel libaio-devel \
@@ -45,7 +71,6 @@ fedora-no-vdso() {
 	vagrant reload
 	ssh default cat /proc/cmdline
 	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; make -j 4'
-	# Excluding two cgroup tests which seem to fail because of cgroup2
 	ssh default 'cd /vagrant/criu/test; sudo ./zdtm.py run -a --keep-going'
 }
 
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,16 @@
+environment:
+  CIRRUS_WORKING_DIR: "/tmp/criu"
+container:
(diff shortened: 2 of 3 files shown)


rppt:
Not strictly related to this PR, but do we really want to move travis-ci.com? 
Citing travis-ci.com authorization request:
"This application will be able to read and write all public and private repository data."
It seems to me that *write* is too much

adrianreber:
> Not strictly related to this PR, but do we really want to move to travis-ci.com?

If we want to have access to arm32 (via aarch64 setarch), ppc64le and s390x to do native builds then yes.

> Citing travis-ci.com authorization request:
> "This application will be able to read and write all public and private repository data."
> It seems to me that **write** is too much

Yes, I am also always unhappy when I have to accept that. But looking closer at my current settings it says:

> Read and write access to checks, commit statuses, deployments, and repository hooks

That sounds reasonable.

rppt:
> If we want to have access to arm32 (via aarch64 setarch), ppc64le and s390x to do native builds then yes.

Yeah, that would be nice.
 
> > Citing travis-ci.com authorization request:
> > "This application will be able to read and write all public and private repository data."
> > It seems to me that **write** is too much
> 
> Yes, I am also always unhappy when I have to accept that. But looking closer at my current settings it says:
> 
> > Read and write access to checks, commit statuses, deployments, and repository hooks
>
> That sounds reasonable.

Agree, but then what does the long list (below) on the github authorization page mean?
If we grant the write permission does it mean that travis will be eventually able to "contriubte" to CRIU? ;-)

```
This application will be able to read and write all public and private repository data. This includes the following:
 *   Code
 *   Issues
 *   Pull requests
 *   Wikis
 *   Settings
 *   Webhooks and services
 *   Deploy keys
 *   Collaboration invites
```

0x7f454c46:
> > If we want to have access to arm32 (via aarch64 setarch), ppc64le and s390x to do native builds then yes.
> 
> Yeah, that would be nice.

I'm actually wondering how much does it gives us comparing to cross-compilation.
Note that there is also Jenkins to naively test things on different architectures.
So, the question is more not "do we need to test native build?" but rather "should native build be in every PR?".

0x7f454c46:
And yes, write permission from CI?
Yuck!

adrianreber:
> > > If we want to have access to arm32 (via aarch64 setarch), ppc64le and s390x to do native builds then yes.
> > 
> > 
> > Yeah, that would be nice.
> 
> I'm actually wondering how much does it give us comparing to cross-compilation.
> Note that there is also Jenkins to naively test things on different architectures.
> So, the question is more not "do we need to test native build?" but rather "should native build be in every PR?".
> Not that I'm sure we don't need it, but just to mention how the question stands.

I also thought about if native building brings us anything over cross-compiling. No sure. I guess it gives a bit more confidence that it actually works on the destination platform.

> And yes, write permission from CI?
> Yuck!

No, not to the code:

 * Read access to code, metadata, and pull requests
 * Read and write access to checks, commit statuses, deployments, and repository hooks 

This sounds reasonable for a CI system.

rppt:
> > > > If we want to have access to arm32 (via aarch64 setarch), ppc64le and s390x to do native builds then yes.
> > > 
> > > Yeah, that would be nice.
> > 
> > I'm actually wondering how much does it give us comparing to cross-compilation.
> > Note that there is also Jenkins to naively test things on different architectures.

We have ppc and s390 there, but we don't have arm(64).

> > So, the question is more not "do we need to test native build?" but rather "should native build be in every PR?".
> > Not that I'm sure we don't need it, but just to mention how the question stands.
> 
> I also thought about if native building brings us anything over cross-compiling. No sure. I guess it gives a bit more confidence that it actually compiles on the destination platform.

I perceived "native build" as building and running tests. Cross-compilation only apparently does not add much, but having the tests run on different platforms has added value as there are arch-specific code paths.

> > And yes, write permission from CI?
> > Yuck!
> 
> No, not to the code:
> 
>     * Read access to code, metadata, and pull requests
>     * Read and write access to checks, commit statuses, deployments, and repository hooks
> 
> This sounds reasonable for a CI system.

True.
I just wonder, what rights we grant when we authorize an application. Could it happen that some error or malfunction on travis side can corrupt our repo?

adrianreber:
> > > > > If we want to have access to arm32 (via aarch64 setarch), ppc64le and s390x to do native builds then yes.
> > > > 
> > > > 
> > > > Yeah, that would be nice.
> > > 
> > > 
> > > I'm actually wondering how much does it give us comparing to cross-compilation.
> > > Note that there is also Jenkins to naively test things on different architectures.
> 
> We have ppc and s390 there, but we don't have arm(64).
> 
> > > So, the question is more not "do we need to test native build?" but rather "should native build be in every PR?".
> > > Not that I'm sure we don't need it, but just to mention how the question stands.
> > 
> > 
> > I also thought about if native building brings us anything over cross-compiling. No sure. I guess it gives a bit more confidence that it actually compiles on the destination platform.
> 
> I perceived "native build" as building and running tests. Cross-compilation only apparently does not add much, but having the tests run on different platforms has added value as there are arch-specific code paths.

Unfortunately, Travis s390x, aarch64 and ppc64le is based on unprivileged LXD which blocks many interfaces CRIU needs. So far it has therefore only been compile tested there.

If the non-root changes are merged at some point it might be possible, with the help of CAP_SYS_ADMIN to run the non-root tests on those platforms and it might work. I tested it at some point and the only missing thing, if I understood it correctly, is accessing /proc/PID/map_files, which does not work in a user namespaces, even with CAP_SYS_ADMIN. If CRIU would gracefully fall back to /proc/PID/maps, I think it might be possible to run as non-root, with CAP_SYS_ADMIN on those Travis architectures. Running at least one test should be possible.

With the move to travis-ci.com it is also possible to have bare metal aarch64 access to finally run tests there.

Bit right now a 'native build' is not much more than a cross compile. But it is really fast, usually less than a minute.

> > > And yes, write permission from CI?
> > > Yuck!
> > 
> > 
> > No, not to the code:
> > ```
> > * Read access to code, metadata, and pull requests
> > * Read and write access to checks, commit statuses, deployments, and repository hooks
> > ```
> > 
> > 
> > This sounds reasonable for a CI system.
> 
> True.
> I just wonder, what rights we grant when we authorize an application. Could it happen that some error or malfunction on travis side can corrupt our repo?

Could? I guess it *could* happen. I have not heard of any case were a CI system changed the code of a open source project. Thanks to git it should be easy to detect and to revert. But so far we have trusted Travis and GitHub Actions and nobody detected any changes from their side.

adrianreber:
@avagin Do we want to enable this? Now that Travis is no longer an option.

adrianreber:
@avagin This is also ready for merge. This enables us to start up a VM and test against the latest kernel or without VDSO as in this case. See output at https://cirrus-ci.com/build/6541687033954304

scripts/ci/vagrant.sh
```diff
@@ -6,29 +6,55 @@
 set -e
 set -x
 
-VAGRANT_VERSION=2.2.7
+VAGRANT_VERSION=2.2.10
 FEDORA_VERSION=32
```
rst0git, line 10: Should we also update to Fedora 33?

scripts/ci/vagrant.sh
```diff
@@ -6,29 +6,55 @@
...
-	vagrant up --provider=libvirt
+	if [ -n "$CIRRUS_CI" ]; then
+		# Work around for:
+		# Error while activating network: Call to virNetworkCreate failed: internal error:
+		# Failed to apply firewall rules /usr/sbin/ip6tables --table filter --list-rules: modprobe: FATAL: Module ip6_tables not found in directory /lib/modules/5.4.0-1025-gcp
+		# On cirrus-ci.com. Running in a container without access to the host's kernel modules
+		rm -f /sbin/ip6tables
+		cp /bin/true /sbin/ip6tables
+	fi
+	vagrant up --provider=libvirt > /dev/null
```
rst0git, line 57: There is a [`--no-tty`](https://github.com/hashicorp/vagrant/pull/11465) option that can make `vagrant up` less verbose.

scripts/ci/vagrant.sh
adrianreber: We can do that, yes, but it is not really important to get this PR initially merged.

scripts/ci/vagrant.sh
adrianreber: Good idea, but could also be fixed up later.

adrianreber:
@avagin There are currently, at least, three CI related pull requests open. If you think the changes are not necessary or if you like it done differently, please just close the PRs. Some of these three PRs are open longer than two months. There also a lot of CI runs still defined for Travis which could be moved but with the current CI PRs not being merged it is unnecessary difficult due to merge conflicts.

PR #1275, opened 2020-11-12
CI: run aarch64 compile tests on Drone

adrianreber:
Besides Travis CI Drone CI seems to be only service providing ARM based builds. This switches the aarch64 and arm32 builds to drone.io.

Because Drone CI is running in a Docker container we cannot use 'setarch linux32' as it requires the blocked syscall 'personality(2)'.

But Drone CI provides an 'arch: arm' which gives the same architecture as 'setarch linux32' on Travis aarch64: armv8l

The goal from all this CI changes it to be able to move as much away from Travis as possible. Except maybe ppc64le and s390x (and maybe bare metal graviton2 based aarch64).

Using so many different CI systems is on the one hand not the nicest solution, but if one of the free CI systems decides to go the same way as Travis does we already are comfortable with setting up runs on other CI systems.

If this get merged I will remove the corresponding Travis runs.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,65 @@
+---
+kind: pipeline
+type: docker
+name: aarch64 build GCC (native)
+
+platform:
+  os: linux
+  arch: arm64
+
+steps:
+- name: build
+  image: ubuntu:focal
+  commands:
+  - scripts/ci/apt-install make
+  - make -C scripts/ci local
+
+---
+kind: pipeline
+type: docker
+name: aarch64 build CLANG (native)
+
+platform:
+  os: linux
+  arch: arm64
+
+steps:
+- name: build
+  image: ubuntu:focal
+  commands:
+  - scripts/ci/apt-install make
+  - make -C scripts/ci local CLANG=1
+
+---
+kind: pipeline
+type: docker
+name: armhf build GCC (native)
+
+platform:
+  os: linux
+  arch: arm
+
+steps:
+- name: build
+  # At the time of setting up focal did not work
+  image: ubuntu:bionic
+  commands:
+  - scripts/ci/apt-install make
+  - make -C scripts/ci local
+
+---
+kind: pipeline
+type: docker
+name: armhf build CLANG (native)
+
+platform:
+  os: linux
+  arch: arm
+
+steps:
+- name: build
+  # At the time of setting up focal did not work
+  image: ubuntu:bionic
+  commands:
+  - scripts/ci/apt-install make
+  - make -C scripts/ci local CLANG=1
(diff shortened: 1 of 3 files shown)


adrianreber:
@rst0git I again see the failure regarding TLS: https://travis-ci.org/github/checkpoint-restore/criu/jobs/743236941

adrianreber:
Rebased and force-push. @avagin this is also a simple change just moving the CI runs from one platform to another platform.

Results can be seen here: https://cloud.drone.io/adrianreber/criu

adrianreber:
@avagin: This is rebased and ready for merge to get native aarch64 and arm builds again.

See Drone CI results here: https://cloud.drone.io/adrianreber/criu/31

.drone.yml
```diff
@@ -0,0 +1,63 @@
...
+platform:
+  os: linux
+  arch: arm
+
+steps:
+- name: build
+  image: ubuntu:bionic
+  commands:
+  - scripts/ci/apt-install make gcc pkg-config git perl-modules
+  - make -C scripts/ci local CLANG=1
```
rst0git, line 63: hmm, why is called `CLANG (native)` if it installs `gcc` as a dependency?

adrianreber: Ah, that is just copy paste error. The actual packages are installed using our main CI script. I will clean this up.

adrianreber:
@rst0git Removing gcc from the CI environment uncovered that we always build compel-host-bin with gcc even when CC=clang because compel-host-bin uses HOSTCC and not CC as a compiler. Now I am also setting HOSTCC to clang and now everything is built using clang.

https://cloud.drone.io/adrianreber/criu/37

@avagin ready after reworking to not blindly install gcc in every situation

rst0git (approved):
Well done!

scripts/ci/run-ci-tests.sh
```diff
@@ -51,12 +52,17 @@ ci_prep () {
...
-	CC=gcc
-	# clang support
 	if [ "$CLANG" = "1" ]; then
-		CI_PKGS="$CI_PKGS clang"
+		# clang support
 		CC=clang
+		# If this is running in an environment without gcc installed
+		# compel-host-bin will fail as it is using HOSTCC. Also
+		# set HOSTCC to clang to build compel-host-bin with it.
+		export HOSTCC=clang
```
avagin, line 61: I think this should be in a separate commit. It isn't about Drone...

adrianreber: Created a separate commit.

0x7f454c46 (approved):
Looks sane to me, thanks for working on it, @adrianreber

PR #1277, opened 2020-11-17
ci: disable --tls tests on alpine

adrianreber:
We see intermittent CI failures only on alpine with:
```
(00.038789) Error (criu/tls.c:147): tls: Failed receiving data: Error in the pull function.'
```
Disabling alpine tls tests for now.

I am not able to reproduce it locally. We do not know when it actually broke, because these tests have not been running for some time due to other problems in the CI script.

See: https://travis-ci.org/github/checkpoint-restore/criu/jobs/744134305#L29861

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -181,8 +181,14 @@ LAZY_OPTS="-p 2 -T $LAZY_TESTS $LAZY_EXCLUDE $ZDTM_OPTS"
 ./test/zdtm.py run $LAZY_OPTS --lazy-pages
 # shellcheck disable=SC2086
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages
-# shellcheck disable=SC2086
-./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
+if [ ! -f /etc/alpine-release ]; then
+	# This fails only on alpine. Not clear why.
+	# (00.038789) Error (criu/tls.c:147): tls: Failed receiving data: Error in the pull function.
+	# Skipping it on alpine to avoid broken CI
+
+	# shellcheck disable=SC2086
+	./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
+fi
 
 bash ./test/jenkins/criu-fault.sh
 bash ./test/jenkins/criu-fcg.sh


adrianreber:
Now there is an uffd error:

https://travis-ci.org/github/checkpoint-restore/criu/jobs/744163176#L17209

@rppt any ideas?

rppt:
> Now there is an uffd error:
> 
> https://travis-ci.org/github/checkpoint-restore/criu/jobs/744163176#L17209
> 
> @rppt any ideas?

My guess is that there are unprocessed events in one of uffd queues after the task exits, but I cannot say for sure without reproducing this locally.

I've pushed a possible workaround here
https://github.com/rppt/criu/tree/lazy-fix
Let's see how travis will react to it.

rst0git:
I was able to reproduce the error locally, with an alpine container. It looks like Alpine Linux ships newer version of `gnutls` (3.6.15) but I've not been able to narrow down the exact reason for the problem yet.

```sh
sudo podman run -it alpine sh
```

```sh
apk update && apk add \
gcc \
bash \
build-base \
ccache \
coreutils \
git \
gnutls-dev \
libaio-dev \
libcap-dev \
libnet-dev \
libnl3-dev \
nftables \
nftables-dev \
pkgconfig \
protobuf-c-dev \
protobuf-dev \
py3-pip \
py3-protobuf \
python3 \
sudo \
ip6tables \
iptables \
nftables \
iproute2 \
tar \
bash \
go \
e2fsprogs \
py-yaml \
py3-flake8 \
asciidoctor

pip3 install junit_xml

git clone https://github.com/checkpoint-restore/criu
cd criu && make -j4

./test/zdtm.py run -p 2 -t zdtm/transition/lazy-thp --remote-lazy-pages --tls
```

Snorch:
@rppt 
> I've pushed a possible workaround here
> https://github.com/rppt/criu/tree/lazy-fix
> Let's see how travis will react to it.

Lookin on read error handling in handle_uffd_event, it seems for me that it is badly broken. Why do we check errno for not ret==-1 case it can be some random errno from some random previous syscall, isn't it?

rppt:
> Lookin on read error handling in handle_uffd_event, it seems for me that it is badly broken. 

Yeah, looks like it is.

> Why do we check errno for not ret==-1 case it can be some random errno from some random previous syscall, isn't it?

Right, something piled up wrongly there.

adrianreber:
I have just seen the TLS error also on Fedora rawhide, so it is no longer an Alpine only problem.

rst0git:
> I have just seen the TLS error also on Fedora rawhide, so it is no longer an Alpine only problem.

I think it may occur because the CRIU process exits (and closes the socket) before the TLS session has been terminated.
https://github.com/checkpoint-restore/criu/blob/0ecec9966c15ad11d8557a36452f9b8648554e39/criu/tls.c#L43
Setting `SO_LINGER` on the socket may help to resolve this issue.

scripts/ci/run-ci-tests.sh
```diff
@@ -181,8 +181,14 @@ LAZY_OPTS="-p 2 -T $LAZY_TESTS $LAZY_EXCLUDE $ZDTM_OPTS"
 ./test/zdtm.py run $LAZY_OPTS --lazy-pages
 # shellcheck disable=SC2086
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages
-# shellcheck disable=SC2086
-./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
+if [ ! -f /etc/alpine-release ]; then
+	# This fails only on alpine. Not clear why.
```
criupatchwork, line 185: Please file an issue and add TODO in the comment:
TODO(issue/XXX): This fails only on alpine.

adrianreber:
Closing. TLS test failures have also been seen on Fedora rawhide. This needs to be fixed differently.

PR #1281, opened 2020-11-19
Replace occurrences of python with python3

rst0git:
Some distributions have no /usr/bin/python executable (e.g., on Ubuntu 20.04, where Python2 was never installed).

This patch allows running on a system that has no `/usr/bin/python` executable at all, only python3.

### DIFF
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -48,7 +48,4 @@ RUN adduser -u 1000 -D test
 
 RUN pip3 install junit_xml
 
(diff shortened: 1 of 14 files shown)


adrianreber:
I am undecided about this change. CentOS 7 tests break because of this. You would need to install python3 on CentOS 7 which should be easy, but this means that we will no longer test on python2 at all. CentOS 7 should receive updates until 2024 so we have a platform to test with python2. Not sure if it is really necessary to test with Python 2 at all, now that Python 3 is available everywhere.

PR #1285, opened 2020-11-20
vim: remove wrong 8-space tabs indent from python files

Snorch:
Probably all vim users can setup their desired indent in their vimrc by
themselfs.


adrianreber (approved):
Since we tried to change most Python code to use the official Python coding style having these vim annotations is even wrong. This change makes totally sense.

Issue #1280, opened 2020-11-19
"FAIL at criu lazy-pages/page-server exited with" with no proper error message visible


Snorch:
Reported by Travis https://travis-ci.org/github/checkpoint-restore/criu/jobs/744605516

```
====================== Run zdtm/transition/lazy-thp in ns ======================
Start test
./lazy-thp --pidfile=lazy-thp.pid --outfile=lazy-thp.out
Run criu dump
=[log]=> dump/zdtm/transition/lazy-thp/96/1/dump.log
------------------------ grep Error ------------------------
b'(00.028384)'
b'(00.028445) Collecting mappings (pid: 96)'
b'(00.028451) ----------------------------------------'
b'(00.028699) Dumping path for -3 fd via self 12 [/zdtm/transition/lazy-thp]'
b"(00.028782) Warn  (criu/files-reg.c:1510): Couldn't find the build-id note for file with fd 14"
b'(00.028924) vma 563c2f43f000 borrows vfi from previous 563c2f43b000'
b'(00.028949) vma 563c2f441000 borrows vfi from previous 563c2f43f000'
b'(00.028969) vma 563c2f442000 borrows vfi from previous 563c2f441000'
b'(00.029038) Dumping path for -3 fd via self 12 [/lib/ld-musl-x86_64.so.1]'
b"(00.029071) Warn  (criu/files-reg.c:1510): Couldn't find the build-id note for file with fd 15"
b'(00.036367) Parasite syscall_ip at 0x563c2f43b000'
b'(00.037507) Set up parasite blob using memfd'
b'(00.037538) Putting parasite blob into 0x7f766e88e000->0x7f48f75d8000'
b'(00.037593) Dumping general registers for 99 in native mode'
b'(00.037597) Warn  (compel/arch/x86/src/lib/infect.c:281): Will restore 99 with interrupted system call'
------------------------ ERROR OVER ------------------------
Run criu page-server
Run criu lazy-pages
Run criu restore
Send the 15 signal to  117
Wait for zdtm/transition/lazy-thp(117) to die for 0.100000
=[log]=> dump/zdtm/transition/lazy-thp/96/1/page-server.log
------------------------ grep Error ------------------------
b'(00.001667) Running status-ready scripts'
b'(00.178545) Accepted connection from 127.0.0.1:54844'
b'(00.233720) tls: TLS handshake completed'
b'(00.235191) page-pipe: Created tee pipe size 4194304'
[...]

PR #1286, opened 2020-11-20
Zdtm add more verbosity to fail with no error case

Snorch:
zdtm: on fail with no error also print the tail of the log

We see strange cases there page-server or lazy-pages are exiting with
non-zero but print no errors, probably the tail of the log can help us
to understand what happened.

There are some other uses of grep_errors but let's only change cases
where we explicitly through an exeption on bad ret. For others I'm not
sure if we need extra output, e.g. for validly failing fault injections.

To debug #1280


adrianreber (approved):
Sounds like a good idea. It was always hard to tell why something failed in cases like this.

PR #1287, opened 2020-11-20
criu-ns: Fix --restore-detached option

### LINKED ISSUE #807: Impossible to restore any process due to PID increment rates.
How to reproduce:
1. Dump a process with lots of RAM usage and many threads (like ZPAQ).
2. Reboot.
3. When trying to restore, it fails with "Pid $PID do not match expected $PID".

The `unshare` method does not apply since it does not cover shell jobs.
The problem is that many thing in the OS creates new processes frequently and take up the PID desired.
They create new processes every few seconds, and they are too important (or simply impossible) to kill.
For example, "systemd-udevd" now frequently ends and spawns child processes after an update.
BTRFS also frequently creates kernel threads.
Even worse, CRIU now set the starting point of new PIDs to the restored process' PID, so no matter how many times you try, it will always fail because the next few PIDs are quickly taken.
I think that a workaround is to create empty threads in advance so the PIDs won't be taken during restoration, but I am not sure.

This is a fundamental problem. You tried to restore processes to their original PIDs, which is impossible in reality.

### LINKED ISSUE #1278: target process failed to start after  criu-ns restore  with “-d”
I used criu-ns to restore with "-d", while target process didnot restart  after criu detach  as  follwing log.

#./criu-ns  restore  -j -D ./graph500-ckpt/ --tcp-established -d -v4

Whether criu-ns  cannot work with "-d"?

(01.380027) 73655 (native) is going to execute the syscall 215, required is 215
(01.380061) 73655 was stopped
(01.380087) 73647 was trapped
(01.380092) 73647 (native) is going to execute the syscall 215, required is 215
(01.380127) 73647 was stopped
(01.380153) 73636 was trapped
(01.380159) 73636 (native) is going to execute the syscall 215, required is 215
(01.380192) 73636 was stopped
(01.380218) 73625 was trapped
(01.380223) 73625 (native) is going to execute the syscall 215, required is 215
(01.380256) 73625 was stopped
(01.380282) 73615 was trapped
(01.380288) 73615 (native) is going to execute the syscall 215, required is 215
(01.380319) 73615 was stopped
(01.380345) 73607 was trapped
(01.380350) 73607 (native) is going to execute the syscall 215, required is 215
(01.380384) 73607 was stopped
(01.380408) 73605 was trapped
(01.380414) 73605 (native) is going to execute the syscall 215, required is 215
(01.380456) 73605 was stopped
(01.380481) 73601 was trapped
(01.380487) 73601 (native) is going to execute the syscall 215, required is 215
(01.380520) 73601 was stopped
(01.380528) Running pre-resume scripts
(01.381344) Restore finished successfully. Tasks resumed.
(01.381354) Writing stats
(01.381672) Running post-resume scripts

rst0git:
The `criu-ns` script creates a new PID namespace where CRIU runs as the "init" process.

When using the `--restore-detached` option with `criu-ns`, users expect `criu-ns` to exit without killing the restored process tree. Thus, `criu-ns` should exit instead of waiting for CRIU's status and it should not pass the `--restore-detached` to CRIU to prevent terminating the init process.

Resolves #1278
Resolves #807

### DIFF
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -47,14 +47,6 @@ else:
     ]
     _mount.restype = ctypes.c_int
 
-try:
-    _umount = _libc.umount
-except AttributeError:
-    raise OSError(errno.EINVAL, "umount is not supported on this platform")
-else:
-    _umount.argtypes = [ctypes.c_char]
-    _umount.restype = ctypes.c_int
-
 
 def _mount_new_proc():
     """
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -18,190 +18,190 @@ MS_SLAVE = 1 << 19
 _libc = ctypes.CDLL(ctypes.util.find_library("c"), use_errno=True)
 
 try:
-	_unshare = _libc.unshare
+    _unshare = _libc.unshare
 except AttributeError:
-	raise OSError(errno.EINVAL, "unshare is not supported on this platform")
+    raise OSError(errno.EINVAL, "unshare is not supported on this platform")
 else:
-	_unshare.argtypes = [ ctypes.c_int ]
-	_unshare.restype = ctypes.c_int
+    _unshare.argtypes = [ ctypes.c_int ]
+    _unshare.restype = ctypes.c_int
 
 try:
-	_setns = _libc.setns
+    _setns = _libc.setns
 except AttributeError:
-	raise OSError(errno.EINVAL, "setns is not supported on this platform")
+    raise OSError(errno.EINVAL, "setns is not supported on this platform")
 else:
-	_setns.argtypes = [ ctypes.c_int, ctypes.c_int ]
-	_setns.restype = ctypes.c_int
+    _setns.argtypes = [ ctypes.c_int, ctypes.c_int ]
+    _setns.restype = ctypes.c_int
 
 try:
-	_mount = _libc.mount
+    _mount = _libc.mount
 except AttributeError:
-	raise OSError(errno.EINVAL, "mount is not supported on this platform")
+    raise OSError(errno.EINVAL, "mount is not supported on this platform")
 else:
-	_mount.argtypes = [
-		ctypes.c_char_p,
-		ctypes.c_char_p,
-		ctypes.c_char_p,
-		ctypes.c_ulong,
-		ctypes.c_void_p
-	]
-	_mount.restype = ctypes.c_int
+    _mount.argtypes = [
+        ctypes.c_char_p,
+        ctypes.c_char_p,
+        ctypes.c_char_p,
+        ctypes.c_ulong,
+        ctypes.c_void_p
+    ]
+    _mount.restype = ctypes.c_int
 
 try:
-	_umount = _libc.umount
+    _umount = _libc.umount
 except AttributeError:
-	raise OSError(errno.EINVAL, "umount is not supported on this platform")
+    raise OSError(errno.EINVAL, "umount is not supported on this platform")
 else:
-	_umount.argtypes = [ctypes.c_char]
-	_umount.restype = ctypes.c_int
+    _umount.argtypes = [ctypes.c_char]
+    _umount.restype = ctypes.c_int
 
 
 def run_criu():
-	print(sys.argv)
-	os.execlp('criu', *['criu'] + sys.argv[1:])
+    print(sys.argv)
+    os.execlp('criu', *['criu'] + sys.argv[1:])
 
 
 def wrap_restore():
-	# Unshare pid and mount namespaces
-	if _unshare(CLONE_NEWNS | CLONE_NEWPID) != 0:
-		_errno = ctypes.get_errno()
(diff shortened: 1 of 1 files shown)


scripts/criu-ns
```diff
@@ -61,39 +61,87 @@ def run_criu():
...
+	Create an init process to keep the new PID namespace
+	after CRIU exits.
+	"""
+	(r_pipe, w_pipe) = os.pipe()
+	# Spawn the init
+	if os.fork() == 0:
+		os.close(r_pipe)
+		_mount_new_proc()
+		# Spawn CRIU binary
+		criu_pid = os.fork()
```
avagin: Hmmm. I think you are going to reintroduce the bug that you fixed a month ago:
https://github.com/checkpoint-restore/criu/commit/265569a40f04326dc2b310eb5ddbac9c1cb989e3

rst0git: As far as understand when using `--restore-detached`, CRIU exits after restore, therefore we need another process to keep running within the namespace as PID=1 after restore. As illustrated in https://criu.org/Tree_after_restore#Detached

When CRIU is used without `--restore-detached`, it will run as PID=1.

scripts/criu-ns
```diff
@@ -18,190 +18,238 @@ MS_SLAVE = 1 << 19
...
-	os.execlp('criu', *['criu'] + sys.argv[1:])
+    print(sys.argv)
+    os.execlp('criu', *['criu'] + sys.argv[1:])
+
+
+def _mount_new_proc():
+    """
+    Mount new /proc filesystem.
+    """
+    if _mount(None, b"/", None, MS_SLAVE|MS_REC, None):
```
Snorch, line 68: Why not MS_PRIVATE? why do you need MS_SLAVE, for which exact propagation?

scripts/criu-ns
Snorch: Probably https://github.com/checkpoint-restore/criu/commit/265569a40f04326dc2b310eb5ddbac9c1cb989e3 could've been fixed with just forking criu with some unused pid which will not intersect with restored pids, and this way we would've hit two birds with one stone.

scripts/criu-ns
Snorch: I'm not strictly against it as it's code copying, but this propagation is useless FMPOV.

avagin: If we create a mount namespace just for mounting /proc, MS_SLAVE is the best choice. All changes in the host mount namespace will be propagated into this namespace.

Snorch: > If we create a mount namespace just for mounting /proc, MS_SLAVE is the best choice. All changes in the host mount namespace will be propagated into this namespace.

*offtop*

@avagin Note for the case of "umount a file system and detach its device from the system" probably a more robust way would be not to rely on propagation but to create some new syscall umount3 (or smth. like this) to just umount all mounts of the superblock.

It would probably also help with autofs problems: now when autofs service is restarted it tries to umount autofs and mount a new one in it's place, if you clone mountns and make it rprivate, bindmount of old autofs mount in it will remain but be in some kind of broken state. This new syscall can give an ability to umount all kinds of leftover mounts.

scripts/criu-ns
criupatchwork: @rst0git could you explain how you resolved this comment?

rst0git: As reported in https://github.com/checkpoint-restore/criu/issues/1278, `-d` when used with `criu-ns` is currently broken. It does not work because CRIU exists after restore to detach itself from the process tree, and if  CRIU runs as PID=1 all processes in the PID namespace will be terminated.
This issue was introduced with https://github.com/checkpoint-restore/criu/commit/265569a40f04326dc2b310eb5ddbac9c1cb989e3. 

> could you explain how you resolved this comment?

To fix the broken `-d` functionality, I'm reusing the previous code/behaviour before the commit mention above. Although it has the limitation mentioned in this comment, it works.

In other words, I haven't (yet) resolved the problem mentioned in this comment. It remains a limitation of `criu-ns`.

scripts/criu-ns
```diff
@@ -18,190 +18,195 @@ MS_SLAVE = 1 << 19
...
+        raise OSError(_errno, errno.errorcode[_errno])
 
-	return (None, None)
+    restore_detached = False
+    if '-d' in restore_args:
+        restore_detached = True
+        restore_args.remove('-d')
+    if '--restore-detached' in restore_args:
+        restore_detached = True
+        restore_args.remove('--restore-detached')
```
Snorch, line 101: A bit funny that to restore "detached" we remove "--restore-detached". =)

Snorch (approved):
This looks good for me!

scripts/criu-ns
rst0git: Yes, it is a bit funny and paradoxical, but I hope the commit message is sufficient to explain why.

rst0git:
@Snorch Thank you for the code review!

rst0git:
Thanks!

PR #1296, opened 2020-11-25
uffd: check for exited task when reading uffd_msg

rppt:
Sometimes there are uffd messages in a queue of a dying task and by the
time these messages are processed in handle_request, the uffd is no
longer valid and reading from it causes errors.

Add processing of EBADF in handle_uffd_event() to gracefully handle such
situation.

Signed-off-by: Mike Rapoport <rppt@linux.ibm.com>


rppt:
This could be symptom fix, but it's better than nothing from one side and we'll get some more information about what's going on there as nobody was able to reproduce this locally.

Snorch (approved):
Looks reasonable.

PR #1297, opened 2020-11-26
dump: check the --file-locks flags when dumping.

criupatchwork:
This PR is autogenerated from: https://patchwork.criu.org/series/4494


Snorch:
If we exclude unix SCMs file locks are simple as:

(note call stacks are in reverse order the leftmost function is deepest in
stack)

# here we get info about locks on open files from fdinfo and setup
# file_lock_list

  +-< parse_fdinfo_pid_s
    +-< parse_fdinfo_pid
      +-< fill_fd_params
        +-< dump_one_file (2)
          +-< dump_task_files_seized
            +-< dump_one_task
              +-< cr_dump_tasks (1)

# or here if there are no fdinfo locks

  +-< parse_file_locks
    +-< collect_file_locks # before dump_one_task
      +-< cr_dump_tasks (1)

# here we detect that file lock is held and fail if no option
# opts.handle_file_locks provided

  +-< note_file_lock # after fill_fd_params
    +-< dump_one_file (2)

# so if we got to:

  +-< dump_file_locks # after dump_one_task
    +-< cr_dump_tasks (1)

# we either have empty file_lock_list or have opts.handle_file_locks set or we
# have some "fantom" locks which are held but not on files of processes we
# dump, so probably we should not care about them.

To conclude, in my understanding your patch is excess.

avagin:
https://github.com/checkpoint-restore/criu/commit/7a64074c0b899b98e0d5f438e8d3fa2f7706ed3b

PR #1298, opened 2020-11-26
criu: dump and restore cpu affinity of each thread

snoweay:
Criu should dump and restore threads' or processes'
cpu affinity.

Add one entry of thread_cpuallow_entry into
thread_core_entry to save cpu affinity info.

Restore it after threads restored but before running.

Signed-off-by: Sang Yan <sangyan@huawei.com>


avagin:
you need to implement a zdtm test.
Here are our static tests:
https://github.com/checkpoint-restore/criu/tree/criu-dev/test/zdtm/static

This is an example of how to run them:
python test/zdtm.py run -t zdtm/static/env00

criu/cr-restore.c
```diff
@@ -3225,6 +3229,23 @@ static int prepare_signals(int pid, struct task_restore_args *ta, CoreEntry *lea
 	return ret;
 }
 
+static int prepare_alloweds(int pid, struct task_restore_args *ta, CoreEntry *leader_core)
```
rppt, line 3232: I'd suggest to use `prepare_allowed_cpus` as the name of this function

criu/include/restorer.h
```diff
@@ -162,6 +163,8 @@ struct task_restore_args {
 	siginfo_t			*siginfo;
 	unsigned int			siginfo_n;
 
+	cpu_set_t			*cpualloweds;
```
rppt, line 166: Maybe `allowed_cpus`?

criu/pie/restorer.c
```diff
@@ -432,6 +432,31 @@ static int restore_signals(siginfo_t *ptr, int nr, bool group)
 	return 0;
 }
 
+static int restore_cpuallowed(struct task_restore_args *args)
```
rppt, line 435: Here as well, `restore_allowed_cpus`

rppt (changes requested):

criu/include/restorer.h
snoweay: > you need to implement a zdtm test.
> Here are our static tests:
> https://github.com/checkpoint-restore/criu/tree/criu-dev/test/zdtm/static
> 
> This is an example of how to run them:
> python test/zdtm.py run -t zdtm/static/env00

OK，I will implement it later, in this week.

avagin:
How does it work when we restore processes on a host with more or less cpu-s?

snoweay:
> How does it work when we restore processes on a host with more or less cpu-s?

snoweay:
> How does it work when we restore processes on a host with more or less cpu-s?

There's probably no good way to resolve it on a host with more or less cpu-s.
I propose to add a options for this function.

snoweay:
@avagin How many cpus in the zdtm/test CI environmet ?  I added one test case of zdtm/static/cpu-affinity0 with 4 cores affinity dump and restore, and it failed.
Now I tested only one cpu's affinity restore ok.

Please review the commit.

Thanks.

criu/include/cr_options.h
```diff
@@ -174,6 +174,8 @@ struct cr_options {
 
 	/* This stores which method to use for file validation. */
 	int 			file_validation_method;
+
+	int			with_cpu_affinity;	/* restore cpu affinity */
```
adrianreber, line 178: CRIU usually has comments in the line before and not in the same line after the code like here. Just move the the comment to the line above and drop the empty line.

adrianreber:
> @avagin How many cpus in the zdtm/test CI environmet ? I added one test case of zdtm/static/cpu-affinity0 with 4 cores affinity dump and restore, and it failed.
> Now I tested only one cpu's affinity restore ok.

Each CI environment has a different number of CPUs. You need to figure out the number of available CPUs in the test dynamically.

Please also update the man-page and add your new CLI option to the RPC and libcriu interface.

criu/cr-restore.c
```diff
@@ -3223,6 +3227,27 @@ static int prepare_signals(int pid, struct task_restore_args *ta, CoreEntry *lea
 	return ret;
 }
 
+static int prepare_allowed_cpus(int pid, struct task_restore_args *ta, CoreEntry *leader_core)
+{
+	int i;
+	int *need_cpu_affinity;
+	cpu_set_t *cpumaks;
+
```
adrianreber, line 3235: This should probably be cpumask and not cpumaks

criu/pie/restorer.c
```diff
@@ -432,6 +432,40 @@ static int restore_signals(siginfo_t *ptr, int nr, bool group)
...
+{
+	int i;
+	int pid;
+	int ret;
+	int *need_cpu_affinity;
+	cpu_set_t *cpumask;
+	cpu_set_t *allowed_cpus;
+
+	need_cpu_affinity = (int *)args->allowed_cpus;
+	if (!*need_cpu_affinity) {
```
adrianreber, line 445: It seems like you are storing multiple things in 'allowed_cpus'. A comment would be nice describing what you exactly store in 'allowed_cpus'. It took me some time that it included an integer with the information if this is turned on or off and after that the actual information.

test/zdtm/static/cpu-affinity0.c
```diff
@@ -0,0 +1,42 @@
...
+	test_init(argc, argv);
+
+	CPU_ZERO(&old);
+	CPU_ZERO(&new);
+
+	/* test only 0 core because of CI test env limited */
+	CPU_SET(0, &old);
+
+	if (sched_setaffinity(getpid(), sizeof(old), &old) < 0) {
+		pr_perror("Can't set old cpu affinity! errno: %d", errno);
```
rst0git, line 24: `pr_perror()` is defined in `test/zdtm/lib/zdtmtst.h` as
```C
#define pr_perror(format, arg...)       \
        test_msg("ERR: %s:%d: " format " (errno = %d (%s))\n", \
                __FILE__, __LINE__, ## arg, errno, strerror(errno))
```

Note that it already includes information about errno.

adrianreber:
I would prefer to not merge this before we have more CI tests running again.

test/zdtm/static/cpu-affinity0.c
```diff
@@ -0,0 +1,42 @@
...
+		exit(1);
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	if (sched_getaffinity(getpid(), sizeof(new), &new) < 0) {
+		pr_perror("Can't get new cpu affinity! errno: %d", errno);
```
rst0git, line 32: same comment here. No need to add `errno`.

images/core.proto
```diff
@@ -81,6 +81,10 @@ message thread_sas_entry {
 	required uint32			ss_flags	= 3;
 }
 
+message thread_allowedcpus_entry {
+	repeated uint64			cpumask		= 1;
```
rst0git, line 85: For historical reasons, repeated fields of scalar numeric types aren't encoded as efficiently as they could be. New code should use the special option `[packed=true]` to get a more efficient encoding. For example:
```
repeated int32 samples = 4 [packed=true];
```
https://developers.google.com/protocol-buffers/docs/proto#optional

images/core.proto
```diff
@@ -99,6 +103,7 @@ message thread_core_entry {
 
 	optional string			comm		= 13;
 	optional uint64			blk_sigset_extended	= 14;
+	required thread_allowedcpus_entry	allowed_cpus	= 15;
```
rst0git, line 106: Is there a reason why this field is not `optional`?

> **Required Is Forever** You should be very careful about marking fields as required. If at some point you wish to stop writing or sending a required field, it will be problematic to change the field to an optional field – old readers will consider messages without this field to be incomplete and may reject or drop them unintentionally.

https://developers.google.com/protocol-buffers/docs/proto#specifying_field_rules

criu/cr-restore.c
```diff
@@ -3223,6 +3227,27 @@ static int prepare_signals(int pid, struct task_restore_args *ta, CoreEntry *lea
...
+	ta->allowed_cpus = (char *)rst_mem_align_cpos(RM_PRIVATE);
+
+	need_cpu_affinity = rst_mem_alloc(sizeof(int), RM_PRIVATE);
+	*need_cpu_affinity = opts.with_cpu_affinity;
```
rst0git, line 3239: `rst_mem_alloc` could return `NULL`

rppt: Why is `need_cpu_affinity` is a pointer?
And, it seems unused...

criu/pie/restorer.c
rppt: I think it would be better to have `need_cpu_affinity` as a separate field in `struct task_restore_args`

rppt (changes requested):
Maybe I'm missing something, but I don't see where is the check that the number of CPUs on the restore side is not less than on the dump side.

test/zdtm/static/cpu-affinity0.desc
```diff
@@ -0,0 +1 @@
+{'dopts': '', 'ropts': '--with-cpu-affinity', 'flags': 'reqrst '}
```
Snorch, line 1: If dump opts are empty probably we don't need them at all? And one more thing - Why do you need reqrst flag? What happens if you remove it and run the test with --norst?

rst0git:
Closing in favour of https://github.com/checkpoint-restore/criu/pull/1969

PR #1299, opened 2020-11-27
dump: check the --file-locks flag when dumping

dracoding:
When the --file-locks flag is not enabled, it still dump the file-locks msg
to image file. We should check the --file-locks flags when dumping.

Signed-off-by: Xiaoguang Li <lixiaoguang2@huawei.com>

### DIFF
--- a/criu/file-lock.c
+++ b/criu/file-lock.c
@@ -95,6 +95,9 @@ int dump_file_locks(void)
 	struct file_lock *fl;
 	int	ret = 0;
 
+	if (!opts.handle_file_locks)
+		return 0;
+
 	pr_info("Dumping file-locks\n");
 
 	list_for_each_entry(fl, &file_lock_list, list) {


avagin:
Are you sure that it will dump file locks? I think it should fail here:
https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/file-lock.c#L362

dracoding:
> Are you sure that it will dump file locks? I think it should fail here:
> https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/file-lock.c#L362

yes, it dump the file locks. It returned from this line, not going into the loops.
https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/file-lock.c#L351

criu/file-lock.c
```diff
@@ -95,6 +95,9 @@ int dump_file_locks(void)
 	struct file_lock *fl;
 	int	ret = 0;
 
+	if (!opts.handle_file_locks)
+		return 0;
```
avagin, line 99: We can't ignore file locks. If --file-locks isn't set, we need to log an error and exit with a non-zero code.

Snorch: https://github.com/checkpoint-restore/criu/pull/1297#issuecomment-740504224 I didn't initially mentioned this pr, so I posted it there.

It looks like in @dracoding has a "fantom" lock - the one which shows up in /proc/locks but there is no corresponding open fd in dumped processes subtree.

Probably it's ok to ignore them, and we already do it. (but though we put them to images)

dracoding: @Snorch It is not a "fantom" lock. In the dumped process subtree, there is a opened fd that has a lock, so the file_lock_list is not empty. The fdinfo of fd 8 of the dumped process looks like the following:

> pos:    0
flags:  02400002
mnt_id: 1
lock:   1: FLOCK  ADVISORY  WRITE 18514 00:02:142112 0 EOF

If there is no option of opts.handle_file_locks provided,  i think it should be fail from here: https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/file-lock.c#L362, as @avagin said. But it's not return failed. so we will put file locks to images whether or not opts.handle_file_locks is provided. 

I have another question. if multiple users compete for a specific file lock. the recommend solution is assume all file lock users are taken into dump, but if we can’t guarantee all users taken into dump, is there other suggestion to solve this problem?

avagin: @snorch how can we distinguish between "fantom" and real locks? I still don't understand what fantom lock means and why it is there?

avagin: > I have another question. if multiple users compete for a specific file lock. the recommend solution is assume all file lock users are taken into dump, but if we can’t guarantee all users taken into dump, is there other suggestion to solve this problem?

I am not sure that that I understand your question. We request user approval to dump file locks because only the user can guarantee that a file system will be in a consistent state to a moment when a restore will be done. Each time when criu meets any locks on dump without the --file-locks options, it has to fail.

Snorch: > @Snorch how can we distinguish between "fantom" and real locks? I still don't understand what fantom lock means and why it is there?

Sorry, I missed that we only dump locks with fl->real_owner set, so only locks of dumped processes are dumped anyway.

(My idea was that we somehow put locks of unrelated processes to image and I called such locks "fantom". Now I see that I was wrong.)

Snorch:
My opinion is that this patch is a good thing at last.

With this patch we at least:
1) don't print "Dumping file-locks" in dump log in case of no "--file-locks" option;
2) don't do excess iterations over file_lock_list in dump_file_locks, which only skips unrelated locks. (only for !kdat.has_fdinfo_lock case)

The dump_file_locks has no actual checks which should fail criu so it should be safe to disable it in no "--file-locks" case.

But we should change the commit message accordingly.

avagin:
https://github.com/checkpoint-restore/criu/commit/7a64074c0b899b98e0d5f438e8d3fa2f7706ed3b

Issue #1301, opened 2020-12-02
compilling with rpmbuild default flags can fail


Snorch:
If I run (options are taken from default rpmbuild `%{optflags}` on VZ7 which should be the same to Centos7):
```
CFLAGS="-O2 -g -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -fexceptions -fstack-protector-strong -grecord-gcc-switches -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection" make V=1 WERROR=0 -j5
```

I get (fedora 32):
```
ld  -r -z noexecstack -T ./compel/arch/x86/scripts/compel-pack.lds.S  -o criu/pie/parasite.built-in.o  criu/pie/parasite.o criu/pie/pie.lib.a ./compel/plugins/std.lib.a
./compel/compel-host hgen -f criu/pie/parasite.built-in.o -o criu/pie/parasite-blob.h
Error (compel/src/lib/handle-elf-host.c:216): Unexpected precalculated address of section (section .rodata addr 0x80 expected 0x5080)
make[2]: *** [criu/pie/Makefile:58: criu/pie/parasite-blob.h] Error 255
make[2]: *** Waiting for unfinished jobs....
make[1]: *** [criu/Makefile:59: pie] Error 2
make: *** [Makefile:259: criu] Error 2
```

Short reproduce:
```
CFLAGS="-specs=/usr/lib/rpm/redhat/redhat-annobin-cc1" make -j5
```

And a criu/pie/parasite.built-in.o file actually has addres 0x80 after linking:
```
[snorch@snorch criu]$ readelf -WS criu/pie/parasite.built-in.o
Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 1] .text             PROGBITS        0000000000000000 000040 00400a 00  AX  0   0 16
  [87] .data             PROGBITS        0000000000000000 005000 000068 00  WA  0   0 4096
  [88] .rodata           PROGBITS        0000000000000080 005080 001016 00   A  0   0 32
```

Probably introduced by 749eb33a9 ("compel: Calculate sh_addr if not provided by linker"), cc @devdri.

[...]

PR #1302, opened 2020-12-03
compel/handle-elf: override unexpected precalculated addresses

Snorch:
We've seen addresses in parasite.built-in.o precalculated by linker but
in some unexpected manner:

readelf -WS criu/pie/parasite.built-in.o
Section Headers:
  [Nr] Name              Type            Address          Off    Size   ES Flg Lk Inf Al
  [ 1] .text             PROGBITS        0000000000000000 000040 00400a 00  AX  0   0 16
  [87] .data             PROGBITS        0000000000000000 005000 000068 00  WA  0   0 4096
  [88] .rodata           PROGBITS        0000000000000080 005080 001016 00   A  0   0 32

(Notes: All other sections does not have SHF_ALLOC or are of size 0, so I
skip them. Need to add "-specs=/usr/lib/rpm/redhat/redhat-annobin-cc1" to
CFLAGS to reproduce.)

Section 88 has address 0x80 in elf file but in our "consequent"
addresses precalculation algorithm it should be at 0x5080:

  addr(.text) == 0x0
  addr(.data) == 0x400a + (0x1000 - 0x400a % 0x1000) + 0x68 == 0x5068
  addr(.rodata) == 0x5068 + (0x20 - 0x5068 % 0x20) == 0x5080

Probably the linker advises us to move 4096 aligned section to the
beginning to save some space, but it's just a guess.

So probably we should be ready to "non-consequent" alignments
precalculated and just override them.

https://github.com/checkpoint-restore/criu/issues/1301

### DIFF
--- a/compel/src/lib/handle-elf.c
+++ b/compel/src/lib/handle-elf.c
@@ -212,13 +212,11 @@ int __handle_elf(void *mem, size_t size)
 		if (sh->sh_addralign > 0 && k % sh->sh_addralign != 0) {
 			k += sh->sh_addralign - k % sh->sh_addralign;
 		}
-		if (sh->sh_addr && sh->sh_addr != k) {
-			pr_err("Unexpected precalculated address of section (section %s addr 0x%lx expected 0x%lx)\n",
-				   &secstrings[sh->sh_name],
-				   (unsigned long) sh->sh_addr,
-				   (unsigned long) k);
-			goto err;
-		}
+		if (sh->sh_addr && sh->sh_addr != k)
+			pr_info("Overriding unexpected precalculated address of section (section %s addr 0x%lx expected 0x%lx)\n",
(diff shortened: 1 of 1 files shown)


0x7f454c46:
Looks alright to me. Need to be cautious to patch at `sh_offset` while doing relocations (not on `sh_addr`), but that's what the code does right now. `sh_addr` we can re-calculate the way we want (with the specified align)
I think it can be `pr_info()` rather than `pr_warn()`, but I don't insist.

Snorch:
@0x7f454c46 Thanks for taking a look on it!

> I think it can be pr_info() rather than pr_warn(), but I don't insist.

I'm also ok with changing it to pr_info, but probably there is no much difference as it is only compilation warning. -> will do.

0x7f454c46:
@Snorch Thanks!

Issue #1217, opened 2020-10-01
Assembler warning message


adrianreber:
I see following warnings when compiling CRIU on Fedora 33:
```
include/common/asm/bitops.h: Assembler messages:
include/common/asm/bitops.h:37: Warning: no instruction mnemonic suffix given and no register operands; using default for `bt'
include/common/asm/bitops.h: Assembler messages:
include/common/asm/bitops.h:63: Warning: no instruction mnemonic suffix given and no register operands; using default for `bts'
```

adrianreber:
I see @cyrillos and @tkhai as authors. Any ideas how this could be fixed?

cyrillos:
They have been taken from linux kernel code mostly. Take a look please into kernel x86 code, should be fixed there.

tkhai:
Yes, commit 919df37c2

"bits: Add test_and_set_bit()
    
Borrowed from Linux Kernel".

I have no a better idea than borrow a new version of this functions from kernel, if this helps.

adrianreber:
Thanks. I had a look but the linux kernel code has become much more complicated than the versions we have.

I tried to add `q` (64bit values) at the end of both assembler statements and that fails. Adding a `l` (32bit values) seems to work. Which probably makes sense as we are not dealing with 64bit values in both functions. Both are using `int`.

I will open a PR to see what CI things about it.

rppt:
On Wed, Oct 7, 2020 at 4:34 PM Adrian Reber <notifications@github.com> wrote:
>
> Thanks. I had a look but the linux kernel code has become much more complicated than the versions we have.
>
> I tried to add q (64bit values) at the end of both assembler statements and that fails. Adding a l (32bit values) seems to work. Which probably makes sense as we are not dealing with 64bit values in both functions. Both are using int.

We are dealing with 64-bit values for the first argument of `bt` and `bts`, so I'd expect q suffix to work :/

I've tried to adopt the latest version from Linux and it compiles fine
on f31. Can you please check if f33 would be happy with this?

diff --git a/include/common/arch/x86/asm/bitops.h b/include/common/arch/x86/asm/bitops.h
index b60ead7fa..8a0bc5807 100644
--- a/include/common/arch/x86/asm/bitops.h
+++ b/include/common/arch/x86/asm/bitops.h
@@ -1,9 +1,18 @@
 #ifndef __CR_BITOPS_H__
 #define __CR_BITOPS_H__
 
+#include <stdbool.h>
[...]

Issue #127, opened 2016-03-04
Need test for /dev/kmsg


xemul:
In files.c there's special handling for this device in get_mem_dev_ops

PR #1303, opened 2020-12-03
x86/asm: fix compile error in bitops.h

anatasluo:
When I try to compile criu in ubuntu amd64 with command "make DEBUG=1", it produces following errors:

include/common/asm/bitops.h: Assembler messages:
include/common/asm/bitops.h:71: Error: incorrect register `%edx' used with `q' suffix

The problem is caused by following code:
`
static inline bool test_and_set_bit(int nr, volatile unsigned long *addr)
{
	bool oldbit;

	asm("btsq %2,%1"
	    CC_SET(c)
	    : CC_OUT(c) (oldbit)
	    : "m" (*(unsigned long *)addr), "Ir" (nr) : "memory");
	return oldbit;
}
`

since nr is a int, it can not be used with 'q' suffix.

### DIFF
--- a/include/common/arch/x86/asm/asm.h
+++ b/include/common/arch/x86/asm/asm.h
@@ -1,0 +1,28 @@
+#ifndef __CR_ASM_H__
+#define __CR_ASM_H__
+
+#ifdef __GCC_ASM_FLAG_OUTPUTS__
+# define CC_SET(c) "\n\t/* output condition code " #c "*/\n"
+# define CC_OUT(c) "=@cc" #c
+#else
+# define CC_SET(c) "\n\tset" #c " %[_cc_" #c "]\n"
+# define CC_OUT(c) [_cc_ ## c] "=qm"
+#endif
+
+#ifdef __ASSEMBLY__
+# define __ASM_FORM(x)	x
+#else
+# define __ASM_FORM(x)	" " #x " "
+#endif
+
+#ifndef __x86_64__
+/* 32 bit */
+# define __ASM_SEL(a,b) __ASM_FORM(a)
+#else
+/* 64 bit */
+# define __ASM_SEL(a,b) __ASM_FORM(b)
+#endif
+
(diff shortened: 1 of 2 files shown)


adrianreber:
@rppt We did just recently discuss something like this in #1217, right?

adrianreber:
@anatasluo Just for completeness. Which version of Ubuntu and which compiler and version?

anatasluo:
@adrianreber Sorry, I should add these infomation at first.

About compile environment:
Linux version 4.15.0-124-generic (buildd@lgw01-amd64-029) (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #127-Ubuntu SMP Fri Nov 6 10:54:43 UTC 2020

rppt:
> @rppt We did just recently discuss something like this in #1217, right?

Yep, but apparently we only fixed it halfway :)

rppt (commented):
Please add a signed-of-by to the commit message as described in [CONTRIBUTING.md](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md)

rppt (approved):
I presume these changes follow the implementation in Linux, so with added Signed-off-by this looks Ok to me.

Reviewed-by: Mike Rapoport <rppt@linux.ibm.com>

anatasluo:
@rppt yes, I follow the implementation in Linux. I have added the Signed-off-by message in my commit, hope this time is right.

adrianreber:
Looks good. Travis is still not running, so that we are missing most of our testing, but I see no errors in the remaining, non Travis, tests.

0x7f454c46:
Yeah, looks alright.
I think the main fix is in changes from `int nr` to `long nr` because so far we don't support native i386.
IOW, `#ifndef __x86_64__` is always false.
But I don't mind, let's align it a bit more with Linux/kernel sources.

0x7f454c46:
Thanks!

Issue #336, opened 2017-06-29
Move sigact dump/restore code into separate file


xemul:
In the `parasite-syscall.c` file there's a function `parasite_dump_sigacts_seized()`. In the `cr-restore.c` file there's a `prepare_sigactions()` call (and whatever it calls).

Both are better to be moved into a .c file in the criu/ dir.

prakritigoyal19:
@xemul Hi! I would like to work on this issue. Can this be assigned to me please?

xemul:
Done

prakritigoyal19:
Thanks!
I am planning to move those functions from their original files along with all the necessary imports, to another file, created in the criu dir. Then importing this newly created file to the original files (parasite-syscall.c , cr-restore.c).
Please correct me if I'm wrong.

prakritigoyal19:
Hey, I have mailed my patches to the CRIU mailing list. Please let me know if I should make any changes. 

The repo link for the same is https://github.com/prakritigoyal19/criu/tree/sigactions.

Thank You!

PR #1308, opened 2020-12-12
Move sigact dump/restore code into separate file

rakki-18:
This tries to solve the issue #336.

Changes introduced in this pull request:
- `parasite_dump_sigacts_seized`  in parasite-syscall.c is moved to sigacts-dump.c
- `prepare_sigactions` and the other functions that it calls are also moved to sigacts- dump.c
- sigacts-dump.o is added to makefiles.
- sigacts-dump.h is created and this header is included in cr-dump.c and cr-restore.c


rakki-18:
can someone please review this ?

xemul:
@rakki-18 , I've a feeling that your PR is not git-bisect safe. If this is just in moving the code from one file to another, why do we need so many patches at all? Isn't it better just to squash all the changes into a single patch?

rakki-18:
@xemul I have squashed all the changes into one commit.

Snorch:
@rakki-18 You've lost newlines around restore_one_sigaction also it would be nice to have them above sa_inherited and below parasite_dump_sigacts_seized. Except that functions diff looks fine.

criu/sigacts-dump.c
```diff
@@ -0,0 +1,356 @@
...
+#include <sys/wait.h>
+#include <sys/mman.h>
+
+#include "common/config.h"
+#include "common/compiler.h"
+#include "types.h"
+#include "protobuf.h"
+#include "images/sa.pb-c.h"
+#include "images/timer.pb-c.h"
+#include "images/vma.pb-c.h"
```
Snorch, line 14: Probably the old story about headers (https://github.com/checkpoint-restore/criu/pull/990#discussion_r407265715) can be continued here, why do you need vma.pb-c.h ?

rakki-18: Thanks for reviewing. I will try and remove the unnecessary headers.

rakki-18: @Snorch I have removed the unnecessary headers.

Snorch:
>> @rakki-18 , I've a feeling that your PR is not git-bisect safe. If this is just in moving the code from one file to another, why do we need so many patches at all? Isn't it better just to squash all the changes into a single patch?

> @xemul I have squashed all the changes into one commit.

@rakki-18 again. Please think twice before you push =)

criu/sigacts-dump.c
Snorch: You do all the same mistakes again and again https://github.com/checkpoint-restore/criu/pull/990#discussion_r401401455

All external calls should have their headers, all headers for which you don't have any external call should be removed.

rakki-18: @Snorch Oh Sorry, I will try and correct it.
The sigpending.c test is failing as the value of numsig is coming out to be one less than sent_sigs. any idea why ?

rakki-18:
@Snorch Can you please check now ?

rst0git:
@rakki-18 nice work!

Just a small piece of advice on your commit message:

>  Move parasite_dump_sigacts_seized and prepare_sigactions to sigacts-dump
> 
> Moved parasite_dump_sigacts_seized() to sigacts-sump.c from parasite-syscall.c.
> Moved prepare_sigactions() and the other functions it calls from cr-restore.c to cr-dump.c
> Added sigacts-dump.h. sigacts-dump.o was added to makefiles.

It would be good to have a shorter subject line and explain *why* this change is needed. I would recommend this blog post for more information: https://chris.beams.io/posts/git-commit/

rakki-18:
@rst0git Thanks for the advice! I will try to make better commit messages from now.

Snorch:
> https://chris.beams.io/posts/git-commit/

It's really nice, except for paragraph 3, if you do "git log --oneline" you will see that we don't capitalize in commit subject, so @rakki-18  please fix it in you commit.

Snorch:
Moving some interface to a separate file, but still leaving some global variable `sigchld_act` which is used in both files, looks wrong for me.

More over the linker on fedora detects multiple definition on your code:

```
make -j5
...
ld: criu/sigacts-dump.o:/home/snorch/devel/criu/criu/sigacts-dump.c:63: multiple definition of `sigchld_act'; criu/cr-restore.o:(.bss+0x20): first defined here
make[2]: *** [/home/snorch/devel/criu/scripts/nmk/scripts/build.mk:164: criu/built-in.o] Error 1

```

rakki-18:
@Snorch I have fixed the capitalisation of the commit subject.

When I had the static declarations  of  ```sigchld_act``` in both the files, it was failing the sigpending.c test as  the variable ```numsig``` in sigpending was lesser than the variable```sent_sigs``` by one.  Moving the interface to a separate file and keeping the global declarations of ```sigchld_act```  fixed the error.
any idea on how else I can solve this error ?

PR #1309, opened 2020-12-14
GitHub Actions: Fix Centos8 build

0x7f454c46:
Recently started to fail on criu-dev


adrianreber (approved):
I also saw the centos 8 failure and wanted to fix it. Thanks for doing it.

PR #1311, opened 2020-12-15
Make proc/self a service fd and use it for memfd

Snorch:
This is an attempt to cleanup service fds and put "self" to service fds. Also now we can use it instead of getpid() variant in memfd_open, it is a more general way because in case we would support nested pid namespaces opening by getpid is not an option.

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -418,6 +418,10 @@ static int populate_pid_proc(void)
 		pr_err("Can't open PROC_SELF\n");
 		return -1;
 	}
+	if (open_pid_proc(PROC_SELF) < 0) {
+		pr_err("Can't open PROC_SELF\n");
+		return -1;
+	}
 	return 0;
 }
 
--- a/criu/util.c
+++ b/criu/util.c
@@ -281,15 +281,18 @@ int move_fd_from(int *img_fd, int want_fd)
 
 static pid_t open_proc_pid = PROC_NONE;
 static pid_t open_proc_self_pid;
-static int open_proc_self_fd = -1;
 
-void set_proc_self_fd(int fd)
+int set_proc_self_fd(int fd)
 {
-	if (open_proc_self_fd >= 0)
-		close(open_proc_self_fd);
+	int ret;
 
-	open_proc_self_fd = fd;
+	if (fd < 0)
+		return close_service_fd(PROC_SELF_FD_OFF);
+
 	open_proc_self_pid = getpid();
+	ret = install_service_fd(PROC_SELF_FD_OFF, fd);
+
+	return ret;
 }
 
 static inline int set_proc_pid_fd(int pid, int fd)
@@ -308,10 +311,18 @@ static inline int set_proc_pid_fd(int pid, int fd)
 static inline int get_proc_fd(int pid)
 {
 	if (pid == PROC_SELF) {
-		if (open_proc_self_fd != -1 && open_proc_self_pid != getpid()) {
-			close(open_proc_self_fd);
+		int open_proc_self_fd;
+
+		open_proc_self_fd = get_service_fd(PROC_SELF_FD_OFF);
+		/**
+		 * FIXME in case two processes from different pidnses have the
+		 * same pid from getpid() and one inherited service fds from
+		 * another or they share them by shared fdt - this check will
+		 * not detect that one of them reuses /proc/self of another.
+		 * Everything proc related may break in this case.
+		 */
+		if (open_proc_self_fd >= 0 && open_proc_self_pid != getpid())
 			open_proc_self_fd = -1;
-		}
 		return open_proc_self_fd;
 	} else if (pid == open_proc_pid)
 		return get_service_fd(PROC_PID_FD_OFF);
@@ -402,7 +413,7 @@ inline int open_pid_proc(pid_t pid)
 	}
 
 	if (pid == PROC_SELF)
-		set_proc_self_fd(fd);
+		fd = set_proc_self_fd(fd);
 	else
 		fd = set_proc_pid_fd(pid, fd);
 
--- a/criu/files.c
+++ b/criu/files.c
@@ -1256,6 +1256,14 @@ int close_old_fds(void)
 	struct dirent *de;
 	int fd, ret;
 
+	/**
+	 * Close previous /proc/self/ service fd, as we don't wan't to reuse it
+	 * from a different task. Also there can be some junk fd in it's place
+	 * after we've moved our service fds (e.g. from other task of parents
+	 * shared fdtable), we need to close it before opendir_proc() below.
+	 */
+	__close_service_fd(PROC_SELF_FD_OFF);
+
 	dir = opendir_proc(PROC_SELF, "fd");
 	if (dir == NULL)
 		return -1;
@@ -1297,7 +1305,6 @@ int prepare_fds(struct pstree_item *me)
 	sfds_protected = false;
 	close_service_fd(CGROUP_YARD);
 	sfds_protected = true;
-	set_proc_self_fd(-1); /* flush any proc cached fds we may have */
 
 	if (rsti(me)->fdt) {
 		struct fdt *fdt = rsti(me)->fdt;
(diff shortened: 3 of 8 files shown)


avagin:
> Currently we check that base is the same and shared fdt ids are the same, but there also can be situations where different bases with different shared fdt ids still give the same offset sum (running zdtm in Virtuozzo CT we've seen such a case where service_fd_base=512, new_base=128, service_fd_id=24, id=0, SERVICE_FD_MAX=16).

What test was it? It looks like we need to fork 25 processes with CLONE_FD to get this case...

Snorch:
> > Currently we check that base is the same and shared fdt ids are the same, but there also can be situations where different bases with different shared fdt ids still give the same offset sum (running zdtm in Virtuozzo CT we've seen such a case where service_fd_base=512, new_base=128, service_fd_id=24, id=0, SERVICE_FD_MAX=16).
> 
> What test was it? It looks like we need to fork 25 processes with CLONE_FD to get this case...

Thanks for a good question! The answer is: not some special test but the way how we run all tests in VZT lead to this situation.

```
[snorch@snorch PrlProblemReport-2020.11.25-16.31.42.847]$ grep Error restore.log
(13.026824)   1188: Error (criu/servicefd.c:236): sfd: FDSTORE_SK_OFF unable to clone 114->114: Bad file descriptor

[snorch@snorch PrlProblemReport-2020.11.25-16.31.42.847]$ grep "Attach.*1188" restore.log -B1
(00.028528) Add a helper 317 for restoring SID 1171
(00.028637) Attach 1188 to the temporary task 317

[snorch@snorch PrlProblemReport-2020.11.25-16.31.42.847]$ grep "Process.*\<1188\>" dump.log 
(56.448429) Process: 1188(38839)

[snorch@snorch PrlProblemReport-2020.11.25-16.31.42.847]$ grep "SEIZE 38839" dump.log
(00.867513) SEIZE 38839 (comm socket-tcp6-clo): success
```

On the first look socket-tcp6-close-wait is guilty, but it's not.

We run all zdtm tests all at the same time in single VZ ct and then c/r the ct. We run each test, roughly, like this:

```
make -C ./test/zdtm/static/ socket-tcp6-close-wait.pid
```

So we get (this exampe is done on F32 + criu-dev not on VZ for simplicity):

```
[snorch@snorch criu]$ ps auxf | grep socket-tcp6-close-wait
snorch   2496506  0.0  0.0   2224    80 pts/18   S    10:03   0:00  \_ ./socket-tcp6-close-wait --pidfile=socket-tcp6-close-wait.pid --outfile=socket-tcp6-close-wait.out
snorch   2496507  0.0  0.0   2372    88 ?        Ss   10:03   0:00  \_ ./socket-tcp6-close-wait --pidfile=socket-tcp6-close-wait.pid --outfile=socket-tcp6-close-wait.out

[snorch@snorch criu]$ cat /proc/2496506/status | grep "^Pid\|^PPid\|^NSsid"
Pid:	2496506
PPid:	1611
NSsid:	2492027

[snorch@snorch criu]$ cat /proc/2496507/status | grep "^Pid\|^PPid\|^NSsid"
Pid:	2496507
PPid:	1611
NSsid:	2496507

[snorch@snorch criu]$ cat /proc/2492027/comm 
bash
```

So one test process has session 2492027 and then bash dies after all tests are started, so we don't have session leader for this session.

We also have other session leaders missing from different session0x tests and probably something else:

```
[snorch@snorch PrlProblemReport-2020.11.25-16.31.42.847]$ grep "Add a helper" restore.log  | wc -l
35
```

It looks like at least 24 session helpers were created for container init pidns. And you know that we clone helpers with CLONE_FILES to make them more lightweight.

So everithing is obvious now - all those helper tasks to restore sessions share fdt between each other and with pidns init. When we fork some task to inherit session from them we need patches from this PR.

criu/cr-restore.c
```diff
@@ -1107,21 +1107,22 @@ static int restore_one_zombie(CoreEntry *core)
 
 static int setup_newborn_fds(struct pstree_item *me)
 {
-	if (clone_service_fd(me))
+	bool cleanup_parent_sfdt = !me->parent ||
+				   (rsti(me->parent)->fdt &&
+				    !(rsti(me)->clone_flags & CLONE_FILES));
+
+	if (clone_service_fd(me, cleanup_parent_sfdt))
```
avagin, line 1118: I understand what you want to do in this patch, but this looks like a hack.

I think we can make it simpler and write a comment that other descriptors will be cleaned up in close_old_fds.

```
diff --git a/criu/servicefd.c b/criu/servicefd.c
index 1661e12af..89138028f 100644
--- a/criu/servicefd.c
+++ b/criu/servicefd.c
@@ -294,6 +294,14 @@ int clone_service_fd(struct pstree_item *me)
        if (service_fd_base == new_base && service_fd_id == id)
                return 0;
 
+
+       {
+               int fd = __get_service_fd(PROC_SELF_FD_OFF, id);
+               close(fd);
+               clear_bit(PROC_SELF_FD_OFF, sfd_map);
+       }
+
+
        /* Dup sfds in memmove() style: they may overlap */
        if (get_service_fd(LOG_FD_OFF) < new_base - LOG_FD_OFF - SERVICE_FD_MAX * id)
                for (i = SERVICE_FD_MIN + 1; i < SERVICE_FD_MAX; i++)

```

avagin: we call clone_service_fd in a new process and we have to close PROC_SELF_FD_OFF anyway.

criu/cr-restore.c
```diff
@@ -418,6 +418,10 @@ static int populate_pid_proc(void)
 		pr_err("Can't open PROC_SELF\n");
 		return -1;
 	}
+	if (open_pid_proc(PROC_SELF) < 0) {
```
avagin, line 421: Hmmm.  (open_pid_proc(vpid(current) does the same thing, doesn't it?

criu/util.c
```diff
@@ -308,10 +311,18 @@ static inline int set_proc_pid_fd(int pid, int fd)
...
-			close(open_proc_self_fd);
+		int open_proc_self_fd;
+
+		open_proc_self_fd = get_service_fd(PROC_SELF_FD_OFF);
+		/**
+		 * FIXME in case two processes from different pidnses share
+		 * both files and pids - this check will not detect that we
+		 * reuse /proc/self of a different process and everything proc
+		 * related will break. Note that service_fd_id does not help
+		 * here either.
```
avagin, line 322: if you will close PROC_SELF_FD_OFF in clone_service_fd, you will not have this problem.

criu/cr-restore.c
Snorch: Yes I think it does. But "open_pid_proc(PROC_SELF)" is more general as it does the thing in any pidns-proc host/criu/dumpie which contains the process. AFAICR there are some places where we swap proc service fd to criu/dumpie pidns-proc so one can be confused.

criu/util.c
Snorch: Function clone_service_fd() is called from:
```
  +-< clone_service_fd
    +-< setup_newborn_fds
    | +-< restore_task_with_children
```
So for restored tasks/helpers it helps. But there are different call_in_child_process() or just fork() kind of processes for which this problem is probably still there.

Snorch:
Changes in v2:

Added `__close_service_fd(PROC_SELF_FD_OFF);` just before `opendir_proc(PROC_SELF...)`. Which looks like a best place to solve problem of junk fd in place of proc self service fd preventing us from open.

Removed "servicefd: fully cleanup parent's shared fd table on service fd update" as now it is useless.

Improved FIXME comment, I still think that this one is needed (though probably in quite a rare case).

Snorch:
Rebase to fix circleci missing config.

criu/cr-restore.c
avagin: Why do we need to open vpid(current) and PROC_SELF here. Why is PROC_SELF not enough?

Snorch: Probably I'm to late here with answer, but for history:

We open both PROC_SELF_FD_OFF and PROC_PID_FD_OFF both here so that one would be able to reopen them in service fd protected context. (in sfd protected context we don't allow to open new service fd - only replace if it was already open)

Mostly we use PROC_PID_FD_OFF for opening vpid(current)'s proc, but in general it can be used to open some other pid's proc, so probably we still need it.

Snorch:
Thanks!

PR #1312, opened 2020-12-16
CI: Add Circle CI definition

adrianreber:
Circle CI provides bare metal test systems which are a very good environment for the CRIU test cases. This adds two CI runs on Circle CI.

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -97,6 +97,12 @@ test_stream() {
 
 ci_prep
 
+if [ "$CLANG" = "1" ]; then
+	# Needed for clang on Circle CI
+	LDFLAGS="$LDFLAGS -Wl,-z,now"
+	export LDFLAGS
+fi
+
 export GCOV
 $CC --version
 time make CC="$CC" -j4
@@ -204,7 +210,7 @@ if [ -z "$SKIP_EXT_DEV_TEST" ]; then
 fi
 #make -C test/others/exec/ run
 make -C test/others/make/ run CC="$CC"
-if [ -n "$TRAVIS" ]; then
+if [ -n "$TRAVIS" ] || [ -n "$CIRCLECI" ]; then
        # GitHub Actions does not provide a real TTY and CRIU will fail with:
        # Error (criu/tty.c:1014): tty: Don't have tty to inherit session from, aborting
        make -C test/others/shell-job/ run
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -95,6 +95,36 @@ test_stream() {
 	./test/zdtm.py run --stream -p 2 --keep-going -T "$STREAM_TEST_PATTERN" $ZDTM_OPTS
 }
 
+print_header() {
+	echo "############### $1 ###############"
+}
+
+print_env() {
+	# As this script can run on multiple different CI systems
+	# the following lines should give some context to the
+	# evnvironment of this CI run.
+	print_header "Environment variables"
+	printenv
+	print_header "uname -a"
+	uname -a || :
+	print_header "Mounted file systems"
+	mount || :
+	print_header "Kernel command line"
+	cat /proc/cmdline || :
+	print_header "Distribution information"
+	[ -e /etc/lsb-release ] && cat /etc/lsb-release
+	[ -e /etc/redhat-release ] && cat /etc/redhat-release
+	[ -e /etc/alpine-release ] && cat /etc/alpine-release
+	print_header "ulimit -a"
+	ulimit -a
+	print_header "Available memory"
+	free -h
+	print_header "Available CPUs"
+	lscpu || :
+}
+
+print_env
+
 ci_prep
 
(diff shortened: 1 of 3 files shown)


rst0git (approved):

scripts/ci/run-ci-tests.sh
```diff
@@ -191,7 +191,9 @@ LAZY_OPTS="-p 2 -T $LAZY_TESTS $LAZY_EXCLUDE $ZDTM_OPTS"
 # shellcheck disable=SC2086
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
 
-bash -x ./test/jenkins/criu-fault.sh
+# This breaks all of a sudden everywhere
+#bash -x ./test/jenkins/criu-fault.sh
```
avagin, line 195: We need to investigate and fix these tests. I am not going to disable them without knowing the reason.

avagin: https://ci.openvz.org/job/CRIU/job/CRIU-fault-injections/job/criu-dev/

test/zdtm/transition/thread-bomb.c
```diff
@@ -34,7 +34,7 @@ static void *thread_fn(void *arg)
 
 int main(int argc, char **argv)
 {
-	int max_nr = 1024, i;
+	int max_nr = 768, i;
```
avagin, line 37: We need to find why it fails with 1024 and add a comment that explains this.

adrianreber: I also saw failures with 768. So this does not help at all. It is really strange. The following error can be seen during tests:
```
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7f18252e8700 (LWP 3515))]
(gdb) #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f1851a37859 in __GI_abort () at abort.c:79
#2  0x00007f1851a37729 in __assert_fail_base (
    fmt=0x7f1851bcd588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x7f1851c1c1cc "freesize < size", 
    file=0x7f1851c1c1a6 "allocatestack.c", line=389, function=<optimized out>)
    at assert.c:92
#3  0x00007f1851a48f36 in __GI___assert_fail (
    assertion=assertion@entry=0x7f1851c1c1cc "freesize < size", 
    file=file@entry=0x7f1851c1c1a6 "allocatestack.c", line=line@entry=389, 
    function=function@entry=0x7f1851c1c260 <__PRETTY_FUNCTION__.13152> "advise_stack_range") at assert.c:101
#4  0x00007f1851c0d8ce in advise_stack_range (guardsize=<optimized out>, 
    pd=139741679748864, size=<optimized out>, mem=0x7f18252d8000)
    at allocatestack.c:389
#5  start_thread (arg=<optimized out>) at pthread_create.c:567
#6  0x00007f1851b34293 in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```

But, I think, I am only seeing this with clang. Let's handle this PR as the last and try to get the other CI changes merged.

scripts/ci/run-ci-tests.sh
adrianreber: Having no tests for a few weeks resulted in the current situation. We should rather enable all other tests and then figure out why the injection fails instead of skipping tests even longer. But I think the Circle CI PR, this one, should be the last to look closer at. Let's try to get #1313 merged first. #1313 includes the same change to fault injection as this one.

adrianreber:
I removed most commits and changed this to a draft as other CI related PRs should be merged first.

adrianreber:
@avagin This would also be ready. See #1317 for a ticket about the previous failure encountered. Fixed with adding `-Wl,-z,now`.

criupatchwork:
> I removed most commits and changed this to a draft as other CI related PRs should be merged first.

Pls specify all PR-s in the pr description that have to be merged first.

adrianreber:
> > I removed most commits and changed this to a draft as other CI related PRs should be merged first.
> 
> Pls specify all PR-s in the pr description that have to be merged first.

Currently the PRs are independent of each other. Before figuring out the problem with clang and `-Wl,-z,now` I had some pseudo fixes in another PR. Right now every PR is on its own. This one and #1313 #1275 #1272. Maybe first this one and then #1313, I can then rebase #1275 and #1272.

adrianreber:
Ah, I was too slow, you already merged it.

scripts/ci/run-ci-tests.sh
```diff
@@ -97,6 +97,12 @@ test_stream() {
 
 ci_prep
 
+if [ "$CLANG" = "1" ]; then
+	# Needed for clang on Circle CI
```
avagin, line 131: The comment has to describe what it means and why we need this.

PR #1313, opened 2020-12-17
Switch Alpine to GitHub Actions

adrianreber:
I had to disable the fault injection tests as they do no longer seem to work. Something during the last weeks, when Travis was not running broke fault injection.

### DIFF
--- a/.github/workflows/alpine-test.yml
+++ b/.github/workflows/alpine-test.yml
@@ -1,0 +1,15 @@
+name: Alpine Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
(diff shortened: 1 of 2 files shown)


adrianreber:
Disabling the injection test seems useless as it should be skipped anyway when running in an environment that has SKIP_CI_PREP set, which alpine has set.

I am running it now without that change to see if it still fails.

adrianreber:
@avagin This is just the the switch from Travis to GitHub Actions for Alpine based tests without any other changes. Let's merge this first.

rst0git (approved):

PR #1315, opened 2020-12-18
CONTRIBUTING.md: add component prefix to the subject example

Snorch:
If one will do "git log --oneline", it is quite easy to see that we
don't begin subject lines with a capital letter. We start subjects with
component prefixes where components (mostly) are lowercase. So let's fix
it in contribution guide not to mislead newcomers.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -126,6 +126,19 @@ If your change address an issue listed in GitHub, please use `Fixes:` tag with t
 
 The `Fixes:` tags should be put at the end of the detailed description.
 
+Please add a prefix to your commit subject line describing the part of the
+project your change is related to. This can be either the name of the file or
+directory you changed, or just a general word. If your patch is touching
+multiple components you may separate prefixes with "/"-es. Here are some good
+examples of subject lines from git log:
+
+```
+criu-ns: Convert to python3 style print() syntax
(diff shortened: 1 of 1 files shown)


adrianreber (approved):
Sounds good to me. I am sometimes using upper-case and sometimes lower-case, but having a clear rule is a good idea.

I also like that it says 'please' so it is a SHOULD rule and not required as it not always makes sense or it is not always possible.

CONTRIBUTING.md
```diff
@@ -128,7 +128,12 @@ The `Fixes:` tags should be put at the end of the detailed description.
 
 You may refer to [How to Write a Git Commit
 Message](https://chris.beams.io/posts/git-commit/) article for
-recommendations for good commit message.
+recommendations for good commit message. Please also add component (for
+instance name of the file you change) prefixes to your subject line like:
+
+```
+crns.py: New attempt to have --unshare option
```
rst0git, line 135: I also think that it is a good idea to mention this but it is common for git commits to change multiple files. Using `crns.py` as an example for a prefix could be misleading.

Snorch: I will add one another example here, with non-file prefix.

Snorch:
I've updated pr with better examples and moved the explanation a bit up with addition explanation about component prefixes.

rst0git (approved):

PR #1320, opened 2020-12-23
zdtm/transition/thread-bomb: allocate stack for each thread

avagin:
A thread stack can be set by pthread_attr_setstack.

A per-thread stack can be allocated by the pthread_create call, but the test
tiggers a bug there:

    assertion=0x7f1851c1c1cc "freesize < size",
    file=0x7f1851c1c1a6 "allocatestack.c", line=389, function=<optimized out>)
    at assert.c:92

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -127,12 +127,6 @@ print_env
 
 ci_prep
 
-if [ "$CLANG" = "1" ]; then
-	# Needed for clang on Circle CI
-	LDFLAGS="$LDFLAGS -Wl,-z,now"
-	export LDFLAGS
-fi
-
 export GCOV CC
 $CC --version
 time make CC="$CC" -j4 V=1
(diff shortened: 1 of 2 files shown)


adrianreber:
Did you disable Circle CI? I do not see it running Would be nice to see the results with your patch.

If this fixes the CI problems great, but do you know why we only see the error when using clang? Is the thread-bomb test just bad code which only works because we are lucky with gcc?

Is this PR a workaround hiding a CRIU bug or is thread-bomb just lucky to work in combination with gcc?

avagin:
@adrianreber 

>  Is the thread-bomb test just bad code which only works because we are lucky with gcc?

I see nothing bad in the thread-bomb code.

> Is this PR a workaround hiding a CRIU bug or is thread-bomb just lucky to work in combination with gcc?

I am not sure that this is due to CRIU. Could you try to run thread-bomb without CRIU?

$ cd test/zdtm/transition/
$ make CC=clang thread-bomb.cleanout
$ make CC=clang thread-bomb.pid
$ sleep 5
$ make CC=clang thread-bomb.out
$ cat thread-bomb.out

adrianreber:
I tried it on Fedora 33 and Ubuntu 18.04 locally with clang 10 and clang 11 and always get PASS in `thread-bomb.out` (without your patch).

Should I see an error message?

avagin:
I thought that this is an issue in glibc, but I found that it is in CRIU.

PR #1326, opened 2020-12-31
github: auto-close stale issues and pull requests

adrianreber:
This is a copy of stale.yml from the Podman repository to automatically close tickets after 365 days. After 30 days tickets and issues are marked as stale and after 365 tickets and issues are finally closed.

We have lots of old open issues and pull requests in GitHub. I think auto-closing, as done in this PR, issues and pull requests after 365 days of inactivity would make sense. I do not think any of our old issues and pull requests will every get any attention.

There are currently over 300 open issues against CRIU and I think it is difficult to see which of those are still relevant. 365 days is a long enough period for someone interested in a issue or pull request to still work on it, but if after 365 days no one is doing anything on an issue or pull request it makes more sense to close it. Issues and pull requests can still be re-opened if necessary and we can also add special labels to issues and pull requests which should be exempt from auto-close.

Based on: https://github.com/marketplace/actions/close-stale-issues

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,27 @@
+name: Mark stale issues and pull requests
+
+# Please refer to https://github.com/actions/stale/blob/master/action.yml
+# to see all config knobs of the stale action.
+
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'A friendly reminder that this issue had no activity for 30 days.'
+        stale-pr-message: 'A friendly reminder that this PR had no activity for 30 days.'
+        stale-issue-label: 'stale-issue'
+        stale-pr-label: 'stale-pr'
+        days-before-stale: 30
+        days-before-close: 365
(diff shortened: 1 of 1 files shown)


0x7f454c46:
I like it. Though, I wish it was possible to differ all open issues from the ones where the root cause is known.
IOW, could it keep some issues open, say with a specific tag?
We have some issues that are improvements and others where the root-cause is known and they are just waiting to be fixed.
Of course, I have my private TODO-list, but those are kind of open-to-public-TODO-list.

adrianreber:
> I like it. Though, I wish it was possible to differ all open issues from the ones where the root cause is known.
> IOW, could it keep some issues open, say with a specific tag?
> We have some issues that are improvements and others where the root-cause is known and they are just waiting to be fixed.
> Of course, I have my private TODO-list, but those are kind of open-to-public-TODO-list.

I added
```
+        exempt-pr-labels: 'no-auto-close'
+        exempt-issue-labels: 'no-auto-close'
```
and force pushed, which should exempt issues and pull requests with the label 'no-auto-close'. At least according to the documentation.

It can also be a list of labels which would exempt issues and pull requests from the close automation. I thought 'no-auto-close' is generic enough for everything, but we can add additional labels whenever we feel it is necessary.

0x7f454c46:
Thanks! I'll go through the issues I'm interested to keep open, once it merges.
LGTM

0x7f454c46 (approved):

avagin:
Thanks a lot!

PR #1327, opened 2020-12-31
sk-inet:fix socket which bind to vrf connect back error

dracoding:
It will connect back error when restore tcp socket
which bind to vrf. We should restore the opt of 
SO_BINDTODEVICE before calling the connect function.

Signed-off-by: Xiaoguang Li <lixiaoguang2@huawei.com>


criu/sk-inet.c
```diff
@@ -875,6 +875,8 @@ static int open_inet_sk(struct file_desc *d, int *new_fd)
 		goto err;
 	if (restore_opt(sk, SOL_SOCKET, SO_REUSEPORT, &yes))
 		goto err;
+	if (restore_bound_opts(sk, ie->opts))
```
avagin, line 878: pls you add a comment that explains why restore_bound_opts has to be called before connect

criu/sockets.c
```diff
@@ -605,14 +605,18 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
...
 
-	ret |= restore_bound_dev(sk, soe);
 	ret |= restore_socket_filter(sk, soe);
 
 	/* The restore of SO_REUSEADDR depends on type of socket */
 
 	return ret;
 }
 
+int restore_bound_opts(int sk, SkOptsEntry *soe)
```
avagin, line 615: Why do we need to introduce a new function here? Maybe it is better to call restore_bound_dev directly?

criu/sockets.c
```diff
@@ -605,14 +605,18 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 	tv.tv_usec = soe->so_rcv_tmo_usec;
 	ret |= restore_opt(sk, SOL_SOCKET, SO_RCVTIMEO, &tv);
 
-	ret |= restore_bound_dev(sk, soe);
```
avagin, line 608: restore_socket_opts was called for packet sockets too. Do we need to call restore_bound_dev() for them with your change?

criu/sk-inet.c
dracoding: @avagin 
My runing environmet configuration:
> #ip route show table 40

> 127.127.127.0/28 dev eth0 proto kernel scope link src 127.127.127.10

> route 
> Kernel IP routing table
> Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
> default         8.44.32.1       0.0.0.0         UG    0      0        0 eth4

when restore, kernel find the output interface of eth4. 

`@avagin 
My runing environmet configuration:
> #ip route show table 40

> 127.127.127.0/28 dev eth0 proto kernel scope link src 127.127.127.10

> route 
> Kernel IP routing table
> Destination     Gateway         Genmask         Flags Metric Ref    Use Iface
> default         8.44.32.1       0.0.0.0         UG    0      0        0 eth4

when restore, kernel find the output interface of eth4, then calling __mkroute_output method to add the route table,
it will return -22 from here.

`        if (likely(!IN_DEV_ROUTE_LOCALNET(in_dev)))
        {
                if (ipv4_is_loopback(fl4->saddr) &&
                    !(dev_out->flags & IFF_LOOPBACK) &&
                    !netif_is_l3_master(dev_out))
                {
                        return ERR_PTR(-EINVAL);
                }
        }
`

So I thidk if we restore the bound device before connect, it will find the output device of eth0.

criu/sockets.c
dracoding: restore_bound_dev is a static method, can't be called directly for sk-inet.c

criu/sockets.c
dracoding: restore_bound_dev() method should be retained here, it's also need by packet socket.

Issue #1325, opened 2020-12-29
pps64le: ERR: poll.c:93: Poll-2 returned -1 (events?!) (errno = 14 (Bad address))


avagin:
```
========================== Run zdtm/static/poll in h ===========================
Start test
./poll --pidfile=poll.pid --outfile=poll.out
Run criu dump
Run criu restore
Send the 15 signal to  52
################## Test zdtm/static/poll FAIL at result check ##################
Test output: ================================
02:16:22.503:    52:  0) fd:  6 events  1 revents  0
02:16:22.503:    52:  1) fd:  7 events  1 revents  0
02:16:22.503:    52:     Init: sec           1609035382 nsec               503704
02:16:22.503:    53:    Start: sec           1609035382 nsec               503824
02:16:23.565:    53:  0) fd:  6 events  1 revents  0
02:16:23.565:    53:  1) fd:  7 events  1 revents  0
02:16:23.565:    53: ERR: poll.c:93: Poll-2 returned -1 (events?!) (errno = 14 (Bad address))
02:16:23.567:    52: FAIL: poll.c:132: Child exited with error (errno = 11 (Resource temporarily unavailable))

 <<< ================================
```

```
     2.620 ( 0.004 ms): futex(uaddr: 0x2008c, op: WAKE, val: 2147483647)                      = 0
     2.628 ( 0.006 ms): futex(uaddr: 0x7000c, op: WAKE, val: 2147483647)                      = 1
     2.638 ( 0.021 ms): futex(uaddr: 0x70010, val: 5)                                         = -1 (unknown) (INTERNAL ERROR: strerror_r(512, [buf], 128)=22)
     2.744 ( 0.004 ms): futex(uaddr: 0x70010, val: 5)                                         = -1 EAGAIN (Resource temporarily unavailable)
     4.718 ( 0.007 ms): close(fd: 13)                                                         = 0
     4.816 ( 0.004 ms): close(fd: 127</root/criu/test/dump/zdtm/static/poll/56/1/restore.log>) = 0
     4.906 ( 0.042 ms): munmap(addr: 0x40000, len: 327680)                                    = 0
     5.036 ( 0.006 ms): rt_sigreturn()                                                        = 0
     5.132 ( 0.022 ms): munmap(addr: 0x10000, len: 655360)                                    = 0
     6.049 ( 0.020 ms): poll(ufds: 0x204, nfds: 2, timeout_msecs: 5000)                       = -1 EFAULT (Bad address)
[...]

PR #1328, opened 2020-12-31
ppc64,arm,mips: sanitize the ERESTART_RESTARTBLOCK -> EINTR transition

avagin:
1. The -ERESTART_RESTARTBLOCK case in get_task_regs() depends on kernel
   internals too much, and for no reason. We shouldn't rely on fact that
   a) we are going to do sigreturn() and b) restore_sigcontext() always
   sets restart_block->fn = do_no_restart_syscall which returns -EINTR.

   Just change this code to enforce -EINTR after restore, this is what
   we actually want until we teach criu to handle ERESTART_RESTARTBLOCK.

2. Add pr_warn() to make the potential bug-reports more understandable,
   a sane application should handle -EINTR correctly but this is not
   always the case.

Fixes: #1325

### DIFF
--- a/compel/arch/arm/src/lib/infect.c
+++ b/compel/arch/arm/src/lib/infect.c
@@ -4,6 +4,8 @@
 #include <string.h>
 #include <compel/plugins/std/syscall-codes.h>
 #include <compel/asm/processor-flags.h>
+#include <errno.h>
+
 #include "common/page.h"
 #include "uapi/compel/asm/infect-types.h"
 #include "log.h"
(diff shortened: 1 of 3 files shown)


0x7f454c46 (approved):
LGTM, and I think it makes sense.
The only concern not directly related to this change is what's going on with syscall restart on aarch64?
Probably, we need a zdtm test for this handling (though, I understand that we don't do it properly with adjusting timeout yet).
Anyway, it's outside of this PR.

avagin:
> The only concern not directly related to this change is what's going on with syscall restart on aarch64?

On arm64, ptrace returns registers that are already adjusted to restart a system call:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/kernel/signal.c#n864

PR #1336, opened 2021-01-11
unix socket name resolution

azhadchenko:
Currently criu is resolving unix socket name by concatenating process cwd or root with unix socket relative name. Unfortunately this approach fails in some cases as overmounting, socket removal and replacement with other file, cwd change, etc.
However we can use SIOCUNIXFILE ioctl which opens unix file fd as descriptor with O_PATH modifier, so it can be readlink(). This solves all problems mentioned above.
Several new test are added to cover this problems and new functionality, old overmount_sock has 'crfail' option no longer, since it is working now.

In this patch series I also rework strip_deleted in order to use it without files-reg.h header, but util.h


criu/files-reg.c
```diff
@@ -1301,9 +1267,11 @@ static int check_path_remap(struct fd_link *link, const struct fd_parms *parms,
 		 * name.
 		 */
 
-		if (errno == ENOENT)
+		if (errno == ENOENT) {
+			link_strip_deleted(link);
```
Snorch, line 1271: Probably you should port https://src.openvz.org/projects/OVZ/repos/criu/commits/f5768e79fd ("files: Don't forget on stipping deleted postfix on linked files") as a separate patch instead of implicitly adding link_strip_deleted here, for the sake of better git history.

Snorch (approved):
Rebase from vzcriu (except for a small comment) looks good so I confirm my review-by.

avagin:
> However we can use SIOCUNIXFILE ioctl which opens unix file fd as descriptor with O_PATH modifier, so it can be readlink(). This solves all problems mentioned above.

When was this ioctl added? I have never heard about it before.

criu/sk-unix.c
```diff
@@ -577,71 +504,60 @@ static int unix_resolve_name(int lfd, uint32_t id, struct unix_sk_desc *d,
...
 	}
 
-	if (ue->mnt_id >= 0)
-		ns = lookup_nsid_by_mnt_id(ue->mnt_id);
-	else
-		ns = lookup_ns_by_id(root_item->ids->mnt_ns_id, &mnt_ns_desc);
-	if (!ns) {
-		ret = -ENOENT;
-		goto out;
+	fd = ioctl(lfd, SIOCUNIXFILE);
```
avagin, line 510: SIOCUNIXFILE was added in 4.10. Should we care about older kernels?

rst0git: If `ioctl` fails, we could fallback to `resolve_rel_name`, and include a comment that `SIOCUNIXFILE` is not supported in older kernels.

azhadchenko:
Added new commit by Cyrill Gorcunov introducing one more strip_deleted (was previously meld into my commit with strip_deleted rework).
Preserved old unix_resolve_name as fallback.

Issue #1339, opened 2021-01-19
ZDTM rpc unexpected behavior 


ZeyadYasser:
I was working on this issue https://github.com/checkpoint-restore/criu/issues/717.

I followed the approach suggested by @Snorch of opening a socket pair to transfer pidfds between pre-dump/dump iterations for pidfd based pid reuse detection.
I got the idea that I could open the socket pair implicitly in service mode right before the pre-dump loop https://github.com/checkpoint-restore/criu/blob/0e7321c2fcc5626b45266290b17d9e690d526479/criu/cr-service.c#L874 since CRIU forces the pre-dump to go through the pre-dump loop until the last dump.
I tried this approach with P.Haul and it worked as expected without adding new options for the sockets fds to be passed which was a welcomed plus as I didn't have to change anything in P.Haul.

When running the ZDTM pre-dump rpc test, it failed.
```test/zdtm.py run -t zdtm/static/env00 --rpc --pre 2```

This got me digging into ZDTM, PyCriu and the pre-dump logs.
It turns out that when ZDTM runs in rpc mode, it opens CRIU in SWRK mode, sends the request and then closes the socket directly and opens CRIU in SWRK mode for next pre-dump and so on unlike P.Haul https://github.com/checkpoint-restore/criu/blob/0e7321c2fcc5626b45266290b17d9e690d526479/lib/py/criu.py#L228 which causes CRIU to crash the service worker due to closing the socket in the middle of the pre-dump loop https://github.com/checkpoint-restore/criu/blob/0e7321c2fcc5626b45266290b17d9e690d526479/criu/cr-service.c#L882 and not continue the pre-dump loop.

This caused the side effect that the initialization of the socket pair I did just before the pre-dump loop starts does not persist between pre-dump iterations in the ZDTM rpc tests.

Is this the expected behavior? Should CRIU support pre-dumps in service mode that are not part of the same pre-dump loop?

If this behavior is expected I think I should go with @Snorch 's suggestion of passing the socket pair fds as an option to CRIU instead of the implicit initialization when running in service mode, but this would lead to adding a new option to CRIU, the rpc protobuf definition and updating P.Haul to pass the socket pair when preparing CRIU .

Thank you

adrianreber:
> Is this the expected behavior? Should CRIU support pre-dumps in service mode that are not part of the same pre-dump loop?

[...]

PR #1341, opened 2021-01-22
Add pidfd based pid reuse detection

### LINKED ISSUE #717: RFE: Replace detect_pid_reuse() check from using task's creation time to keeping pidfd on pre-dump
It's a raw arguable idea, but we might want to use pidfd instead of /proc/<pid>/stat creation time.

ZeyadYasser:
This PR closes #717.

This increases the reliability of pid reuse detection using pidfds, mainly during migration like in P.Haul.
A unix domain socket pair is passed to criu (--pidfd-in-fd and --pidfd-in-fd) in CLI or RPC mode.
If those options are set, criu sends tasks pidfds to pidfd_out_fd in send_pidfd_entry() and receives them in the next pre-dump/dump iteration from pidfd_in_fd in init_pidfd_hash().
These pidfds will be used later in detect_pid_reuse().

When pidfd based pid reuse detection is used, the output of detect_pid_reuse() is deterministic hence it returns -1 if pid reuse is detected (task associated with pidfd is closed) which causes criu to exit.

How it should be used in migration tools like P.Haul:
	- Open unix domain socket pair
	- Launch criu in SWRK mode so that it inherits the socket pair fds
	- Pass the socket pair fds in the RPC opts pidfd_in_fd and pidfd_out_fd when doing a pre-dump or dump

This will fail if the kernel does not support pidfd_open syscall, so those options should not be set if the kernel does not support pidfd_open.
This could be checked with: ```criu check --feature pidfd_open```

I have already tried this with P.Haul and it is working as expected. As soon as this gets merged I will open a PR for P.Haul to enable this feature.

Thanks @Snorch  for giving me an idea on how to approach this.


Snorch:
Hi, sorry for a late response! From the first look it seems nice. I will look at it more closer, it may take some time.

Snorch (changes requested):
1) Please split " criu: add pidfd based pid reuse detection" to several separate patches for easier code review.

See more info https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#separate-your-changes

You can split the "add new options" part from actual use of opts.pidfd_out_fd by adding just a "stab" options first and then using them in separate patch.

You can also split actual detect_pid_reuse part from part which fills the hash list and probably criu rpc part can be separated too.

2) I think we can also improve the overall design in two things:

- Why use two sockets? - We can do the same with one unix socket, see how fdstore works (see fdstore_get, fdstore_add and MSG_PEEK). We can read old fds from socket without removing them from it we can add new pidfds to socket and replace old ones in case of pid-reuse.

- Inheriting sockets works badly with criu swrk because we can use one criu swrk to dump multiple containers and it would be better to be able to change pid-reuse store socket on each request. You can take a look on how work_dir_fd is passed over rpc request.

ZeyadYasser:
@Snorch  Thanks for such a detailed review.

I really like the elegance of fdstore_get/fdstore_add utilizing the same socket but I faced a small problem.
P.Haul would create this unix socket and pass the fd to criu just like work_dir_fd, but I found that unlike normal files and folders, sockets of other processes cannot be accessed from /proc.
Luckily there is new syscall for getting fds from other processes [pidfd_getfd()](https://kernelnewbies.org/Linux_5.6#A_new_pidfd_syscall.2C_pidfd_getfd.282.29).
This comes with a minor downside of requiring kernel version >= 5.6.
The other alternative is just using the socket pair and doing the filling and draining of the hashtable with every iteration where criu uses the socket pair as normal client.

I am inclined towards using pidfd_getfd() given the elegance it would provide similar to what is done in fdstore_get/fdstore_add.
What is your opinion?

Snorch:
> sockets of other processes cannot be accessed from /proc (at least not
> without workarounds using ptrace, please correct me if I am wrong).

It looks like you are right about this problem, sorry I didn't thought about
this.

1) Your pidfd_getfd way is probably ok but I can't be sure here.

Some more options:

1) If we can't open socket of other process, we can probably just send this fd
(with pidfds inside already) alongside with rpc request, via same socket we use
to receive requests in cr_service_work.

But probably it's a bit over-complicated.

2) We can provide a socket via a path to a socket file on the filesystem in a
request and bind() to it from criu swrk and read pidfds.

3) Probably pre_dump_loop() can just create and maintain a pidfd storage socket
itself. Without any need from p.haul to pass socket fd to it?

Probably (3) is the most promissing one, please consider =)

criu/kerndat.c
```diff
@@ -1052,6 +1052,24 @@ static bool kerndat_has_clone3_set_tid(void)
...
+	ret = syscall(SYS_pidfd_open, getpid(), 0);
+
+	if (ret == -1) {
+		kdat.has_pidfd_open = false;
+	} else {
+		kdat.has_pidfd_open = true;
+	}
+
+	if (close(ret))
+		return -1;
```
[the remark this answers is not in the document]
rst0git: In addition, the code could be simplified to:
```C
int pidfd = syscall(SYS_pidfd_open, getpid(), 0);
if (pidfd == -1) {
    kdat.has_pidfd_open = false;
    return -1;
}

kdat.has_pidfd_open = true;
close_safe(pidfd);
return 0;
```

ZeyadYasser:
Sorry for such a late response. I am working on it. I am in the middle of my exams so It might take some time.

> 3. Probably pre_dump_loop() can just create and maintain a pidfd storage socket
>    itself. Without any need from p.haul to pass socket fd to it?
> 
> Probably (3) is the most promissing one, please consider =) (upd: ah I see #1339, that is sad...)

Yes, That would have been great doing that without modifying p.haul, It was what I started with at first but unfortunately it wasn't feasible.

ZeyadYasser:
Sorry for taking too long. I hope I was able to address all the issues in the previous version.

I ended up using a single socket to store the pidfds as @Snorch  recommended, and now this pidfd_store socket is passed as an RPC option instead of being inherited in swrk mode which comes with the unwanted side effects that @Snorch  pointed out. A new syscall called pidfd_getfd allows criu to steal the pidfd_store socket while still keeping the socket alive if criu is closed as it still referenced by a fd from the migration tool.

Any feedback would be highly appreciated. Thanks

criu/mem.c
```diff
@@ -30,10 +30,257 @@
...
+
+	/* Steal pidfd store socket from RPC client */
+	pidfd = syscall(SYS_pidfd_open, pid, 0);
+	if (pidfd == -1) {
+		pr_perror("Can't get pidfd of (pid: %d)", pid);
+		goto err;
+	}
+
+	pidfd_store_sk = syscall(SYS_pidfd_getfd, pidfd, sk, 0);
+	close(pidfd); /* Not needed anymore */
```
Snorch, line 77: This close overrides the errno of syscall. And the pr_perror below would print wrong errno. Probably it's better to instead do:

```
+       pidfd_store_sk = syscall(SYS_pidfd_getfd, pidfd, sk, 0);
+       if (pidfd_store_sk == -1) {
+               pr_perror("Can't steal fd %d using pidfd_getfd", sk);
+               close(pidfd);
+               goto err;
+       }
+       close(pidfd);
```

criu/mem.c
```diff
@@ -30,10 +30,257 @@
...
+	}
+
+	/* Steal pidfd store socket from RPC client */
+	pidfd = syscall(SYS_pidfd_open, pid, 0);
+	if (pidfd == -1) {
+		pr_perror("Can't get pidfd of (pid: %d)", pid);
+		goto err;
+	}
+
+	pidfd_store_sk = syscall(SYS_pidfd_getfd, pidfd, sk, 0);
```
Snorch, line 76: I would prefer doing close_safe(&pidfd_store_sk) before opening it, because we can have some leftover from pervious iterations of pre-dump/dump, I see that you do close_safe(&pidfd_store_sk); in free_pidfd_store() but it is only done in the last patch, so after "cr-service: add pidfd_store_sk option to rpc.proto" but before "criu: add pidfd based pid reuse detection for RPC clients" we can see situations where this pidfd_store_sk is not -1 and is open fd...

criu/mem.c
```diff
@@ -30,10 +30,257 @@
...
+	 * When there is data to read from the pidfd, it means
+	 * that the task associated with this pidfd is closed.
+	 */
+	pollfd.fd = entry->pidfd;
+	pollfd.events = POLLIN;
+
+	while (1) {
+		ret = poll(&pollfd, 1, 0);
+		if (errno == EINTR)
+			continue; /* restart polling */
```
Snorch, line 229: Probably some reasonable limit on polling restarts would be nice here.

criu/mem.c
```diff
@@ -30,10 +30,257 @@
...
+int send_pidfd_entry(pid_t pid)
+{
+	int pidfd, entry_state;
+	struct pidfd_entry *entry;
+
+	/* Skip sending if pidfd_store_sk is not initialized */
+	if (pidfd_store_sk == -1)
+		return 0;
+
+	/*
```
Snorch, line 244: Space at the end of line.

criu/mem.c
```diff
@@ -30,10 +30,257 @@
...
+		if (pidfd == -1) {
+			pr_perror("Can't get pidfd of pid %d", pid);
+			return -1;
+		}
+	} else {
+		pidfd = entry->pidfd;
+	}
+
+	if (send_fds(pidfd_store_sk, NULL, 0, &pidfd, 1, &pid, sizeof(pid_t))) {
+		pr_perror("Can't send pidfd %d of pid %d", pidfd, pid);
```
Snorch, line 277:
```
if (!entry)
    close(pidfd);
```

criu/mem.c
```diff
@@ -30,10 +30,257 @@
...
+		return -1;
+	}
+
```
Snorch, line 280:
```
if (!entry)
    close(pidfd);
```
Else you would leak it.

Snorch (commented):
@ZeyadYasser Nice work! Please see inline comments.

One more thing, do we have a test for it? Can we provide pidfd_store_sk in zdtm with iterations.

ZeyadYasser:
@Snorch Thank you for such detailed review. I hope I have addressed the issues appropriately.
Also I noticed something and I don't know what is the best action for it.
https://github.com/ZeyadYasser/criu/blob/11d367dee58b31dae06c6383164bfa29eb643d39/criu/mem.c#L580
If we reached this, it means either CRIU didn't send the pidfd in the last iteration, or (most likely) the RPC client sends a new empty pidfd_store socket each time. Should we exit with error in this situation, or just mark it as pid reuse to make a full memory dump?
I think that if we enter detect_pid_reuse(), it means that we found a pagemap image for that pid, so reaching this code path means something unexpected happened.
Your feedback would be much appreciated.

ZeyadYasser:
I just noticed that the new test was never triggered in CI because it requires a recent kernel (>= 5.6) so I added the test to the Fedora 33 vagrant test because it needs to be run manually for RPC mode to work. I am not sure if I could tell the test to run in RPC mode in the .desc file.

Snorch:
>If we reached this, it means either CRIU didn't send the pidfd in the last iteration, or (most likely) the RPC client sends a new empty pidfd_store socket each time. Should we exit with error in this situation, or just mark it as pid reuse to make a full memory dump?

I think that both are ok:

a) If criu didn't send the pidfd in the last iteration, it probably does not see this pid at that time e.g.: a new task forked in migrated container.

b) If someone sends empty pidfd_store on non-first iteration - yes that's probably a problem, but it's on the sender side.. Though probably in this case we can fail in init_pidfd_store_hash() if the store is completely empty and at the same time criu has previous images dir option (--prev-images-dir) set. (Or at least print warning.)

Snorch (approved):
All comments fixed, and test is working right. Thanks!

ZeyadYasser:
> b) If someone sends empty pidfd_store on non-first iteration - yes that's probably a problem, but it's on the sender side.. Though probably in this case we can fail in init_pidfd_store_hash() if the store is completely empty and at the same time criu has previous images dir option (--prev-images-dir) set. (Or at least print warning.)

I added a check that fails when the pidfd store is empty after the first pre-dump.
I believe it would cause slow full page dumps as all pids will be classified as a pid reuse which is against the purpose pid reuse detection in the first place.

Snorch (approved):

criu/mem.c
```diff
@@ -30,10 +30,138 @@
...
+static struct hlist_head pidfd_hash[PIDFD_HASH_SIZE];
+
+void free_pidfd_hash(void)
+{
+	int i;
+	struct pidfd_entry *entry;
+
+	for (i = 0; i < PIDFD_HASH_SIZE; i++) {
+		hlist_for_each_entry(entry, &pidfd_hash[i], hash)
+			xfree(entry);
```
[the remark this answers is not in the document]
avagin: @ZeyadYasser could you address this comment?

criu/mem.c
```diff
@@ -131,6 +140,171 @@ int init_pidfd_store_sk(pid_t pid, int sk)
...
+
+	/*
+	 * When there is data to read from the pidfd, it means
+	 * that the task associated with this pidfd is closed.
+	 */
+	pollfd.fd = entry->pidfd;
+	pollfd.events = POLLIN;
+
+	while (1) {
+		ret = poll(&pollfd, 1, 0);
```
avagin, line 244: poll with the zero timeout can't return EINTR, can it?

criu/mem.c
```diff
@@ -131,6 +140,171 @@ int init_pidfd_store_sk(pid_t pid, int sk)
...
+	if (pidfd_store_sk == -1)
+		return 0;
+
+	/*
+	 * Drain all queued pidfd entries in pidfd_store_sk from
+	 * the last predump into pidfd_hash.
+	 */
+	cnt = 0;
+	while (1) {
+		entry = xmalloc(sizeof(struct pidfd_entry));
```
avagin, line 177: if (entry == NULL) ...

ZeyadYasser: I don't understand what is the problem, could you please clarify?

criu/mem.c
ZeyadYasser: @avagin I encountered this behavior before once when polling with zero timeout in Golang, This was most likely due to the Go runtime bombarding the thread with ```SIGURG```. I am not sure if there is a scenario where CRIU would receive a high frequency of signals that would cause zero timeout poll to interrupt, so I went with the safe route.

Your perspective on this would be highly valuble.

criu/mem.c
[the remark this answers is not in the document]
ZeyadYasser: @avagin I not sure why github didn't reflect the change in the conversation, but I updated that part using ```hlist_for_each_entry_safe```. Also I just noticed that I need to reset the heads of each hlist ```INIT_HLIST_HEAD```,  thanks for pointing it out.
https://github.com/checkpoint-restore/criu/pull/1341/files#diff-b6aa45df2259d449d7828c7a5ed410a3282d977aa73607e6111808016dd3e2d3R150

criu/mem.c
Snorch: @avagin Looks like it actually can return EINTR.

```
do_poll()
{
...
                if (!count) {
                        count = wait->error;
                        if (signal_pending(current))
                                count = -ERESTARTNOHAND;
                }
}
```

If signal_pending is true. I don't see any explicit protection from it.

criu/mem.c
Snorch: @ZeyadYasser you should check error on allocation.

rst0git: e.g., something like this
```C
entry = xmalloc(sizeof(struct pidfd_entry));
if (entry == NULL) {
    pr_perror("Can't allocate entry");
    return -1;
}

criu/mem.c
avagin: Ok, I had to be more detailed. I actually mean that it will not return EINTR if there is any event.

avagin:
@ZeyadYasser I decided to merge this PR in criu-dev. Thank you for this work!

We need to do a few actions to merge it into the master branch: https://criu.org/New_features_checklist. 

As for the code, I think it would be better to move it into a separate file.

criu/mem.c
Snorch: @ZeyadYasser Anyway, the code in check_pidfd_entry_state() is badly broken. Shame on me that I initially missed it...

!!!!One should never rely on errno if libc syscall is successful!!!!

Imagine on the first iteration you get no event and a signal and thus get errno == EINTR or errno was already EINTR before we even started poll-ing because of previous calls, then all later iterations would be successful but errno would still continue to be EINTR, and we would cycle 10 times for no reason.

Also it is known practice for glibc to set errno to some error even in case of successful glibc syscall, because it can do "retries" inside glibc and first try may set errno, while second try would be successful.

Probably it was the case with Golang too, I've seen places in docker code where they forget to check retcode and wrongly rely on errno =)

So we should probably remove the while loop.

ZeyadYasser: My bad, I didn't know that a call could be successful and errno not changing.
Thank you

Issue #1346, opened 2021-01-25
(zdtm/thp_disable) get_smaps_bits.c: VmFlags not found


rst0git:
The `zdtm/thp_disable` test fails with the following error:
```
======================= Run zdtm/static/thp_disable in h =======================
Start test
./thp_disable --pidfile=thp_disable.pid --outfile=thp_disable.out
make: *** [Makefile:426: thp_disable.pid] Error 1
 Test zdtm/static/thp_disable FAIL at ['make', '--no-print-directory', '-C', 'zdtm/static', 'thp_disable.pid'] 
Test output: ================================
09:47:26.130:    52: Fetch existing flags/adv
09:47:26.130:    52: ERR: get_smaps_bits.c:122: VmFlags not found for 7f08ef89d000 (errno = 2 (No such file or directory))
09:47:26.130:    51: ERR: test.c:316: Test exited unexpectedly with code 255
```
By adding a debug print statement, I noticed that the mmap address gets rounded to 4K.
```diff
--- a/test/zdtm/static/get_smaps_bits.c
+++ b/test/zdtm/static/get_smaps_bits.c
@@ -109,6 +109,7 @@ int get_smaps_bits(unsigned long where, unsigned long *flags, unsigned long *mad
        while (fgets(buf, sizeof(buf), smaps)) {
                is_vma_range_fmt(buf, &start, &end);
 
+               printf("%lu %lu | %s", start, where, buf);
                if (!strncmp(buf, "VmFlags: ", 9) && start == where) {
                        found = 1;
                        parse_vmflags(buf, flags, madv);

```
```
(start)        (where)
-------------------------------------------------------------------------
139894293479424 139894293483520 | 7f3badab4000-7f3badcb5000 rw-p 00000000 00:00 0 
139894293479424 139894293483520 | Size:               2052 kB
139894293479424 139894293483520 | KernelPageSize:        4 kB
139894293479424 139894293483520 | MMUPageSize:           4 kB
139894293479424 139894293483520 | Rss:                   4 kB
139894293479424 139894293483520 | Pss:                   4 kB
139894293479424 139894293483520 | Shared_Clean:          0 kB
139894293479424 139894293483520 | Shared_Dirty:          0 kB
139894293479424 139894293483520 | Private_Clean:         0 kB
139894293479424 139894293483520 | Private_Dirty:         4 kB
139894293479424 139894293483520 | Referenced:            4 kB
139894293479424 139894293483520 | Anonymous:             4 kB
[...]

Issue #1317, opened 2020-12-21
thread-bomb: allocatestack.c:389: advise_stack_range: Assertion `freesize < size' failed


adrianreber:
Trying to port CRIU's CI setup partly to Circle CI (#1312) I have encountered following error:
```
Program terminated with signal SIGABRT, Aborted.
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
50	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
[Current thread is 1 (Thread 0x7f18252e8700 (LWP 3515))]
(gdb) #0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:50
#1  0x00007f1851a37859 in __GI_abort () at abort.c:79
#2  0x00007f1851a37729 in __assert_fail_base (
    fmt=0x7f1851bcd588 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x7f1851c1c1cc "freesize < size", 
    file=0x7f1851c1c1a6 "allocatestack.c", line=389, function=<optimized out>)
    at assert.c:92
#3  0x00007f1851a48f36 in __GI___assert_fail (
    assertion=assertion@entry=0x7f1851c1c1cc "freesize < size", 
    file=file@entry=0x7f1851c1c1a6 "allocatestack.c", line=line@entry=389, 
    function=function@entry=0x7f1851c1c260 <__PRETTY_FUNCTION__.13152> "advise_stack_range") at assert.c:101
#4  0x00007f1851c0d8ce in advise_stack_range (guardsize=<optimized out>, 
    pd=139741679748864, size=<optimized out>, mem=0x7f18252d8000)
    at allocatestack.c:389
#5  start_thread (arg=<optimized out>) at pthread_create.c:567
#6  0x00007f1851b34293 in clone ()
    at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
```
This happens during the thread-bomb test, but only with clang. If I understand this special case correctly, the following happens:
```
./test/zdtm.py run -t zdtm/transition/thread-bomb -f h --keep-going --report report --freezecg zdtm:f --norst
```
1. thread-bomb is started
2. we dump it using the freezer cgroup
3. thread-bomb continues to run and then fails

[...]

PR #1349, opened 2021-01-27
ci/gha: consolidate jobs

kolyshkin:
~~This commit stems from fixing the issue of running github actions ci
for each pull request. This happens because the configuration states
that those jobs need to run on push and pull_request. As a result,
each PR have twice the amount of CI jobs.~~

~~The fix to the above problem is to limit the CI for pushes only
to main and criu-dev branches, and tags starting with v.~~

~~Now, instead of modifying this is 8 different files, let's consolidate
those jobs into one single file.~~

Consolidate ci jobs and perform some cleanups:
 - make job names more compact
 - s/ubuntu-latest/ubuntu-20.04/ to avoid warnings
 - dedup x86_64 jobs by using matrix
 - replace matrix.target name where it is not a make target


adrianreber:
Thanks. I am trying to remember if there was a reason why I created a file for each test instead one file like you did in this PR. But I cannot remember. So maybe there was no reason :grin: 

Looks good. It seems this is also triggering the problem described in https://github.com/checkpoint-restore/criu/issues/1346 and https://github.com/checkpoint-restore/criu/issues/1317 which are unrelated to this PR.

I would say this can be merged.

adrianreber:
Why the limitation to run only on `master` and `criu-dev`? If I create some random branch on my checkout which I would like to use to create a PR later, I really like that CI runs, locally on my account before opening the PR. That way I can fix CI errors before opening a PR.

Don't you think that is useful? Just curious if my approach is unusual?

kolyshkin:
> If I create some random branch on my checkout which I would like to use to create a PR later, I really like that CI runs, locally on my account before opening the PR. That way I can fix CI errors before opening a PR.

It's a matter of personal taste I guess. I would create a draft / WIP PR to run CI, or, if I do not intend at all to merge it, I'd create a test PR in my own fork (not in this repo), as **I am** the only one interested in CI at this point. On the contrary, **everyone** is (or should be) interested in CI results from `master` and `criu-dev` branches, this is why we have those.

Surely, if a few people are working together on a feature and they have a branch for it in this repo, it makes sense to have CI for that branch. As I see we have some stable* branches here, so yeah you're probably right, let's do CI for all branches (at least while it's free).

kolyshkin:
> Why the limitation to run only on master and criu-dev?

Oh, OK, now I know why. I created a PR against my own fork of repo (https://github.com/kolyshkin/criu/pull/1) and noticed that every job is run twice.

Now I realize this is not the same for this repo (unless someone create a branch in here and then a PR to merge it), and it's not common to create a PR in a forked repo (unless someone will fork criu for real, rather than for development).

So I am not sure what to do here. Perhaps we can add `*stable` and `wip/*` to the list, to avoid CI running twice as per above scenario. WDYT @adrianreber ?

kolyshkin:
>  Perhaps we can add `*stable` and `wip/*` to the list

I took a look and these branches were last pushed to in 2018, so I guess it does not make sense to add those.

In any case, I think if there's a need to test more branches, we can trivially add it later, and if there's a need for a CI run, anyone can create a WIP/DNM/draft PR either here on in their own fork.

adrianreber:
> > Perhaps we can add `*stable` and `wip/*` to the list
> 
> I took a look and these branches were last pushed to in 2018, so I guess it does not make sense to add those.
> 
> In any case, I think if there's a need to test more branches, we can trivially add it later, and if there's a need for a CI run, anyone can create a WIP/DNM/draft PR either here on in their own fork.

I don't like this change. I am happy with my current workflow. Push to a branch on my fork, wait for CI, create a PR if everything is green. Opening a PR against my fork seems unnecessary complicated. But that is just my workflow and I can deal with it if we want to change it. I am not in favour of this, but do not not merge this because of me. Maybe others agree that this is the right thing to do.

kolyshkin:
> I don't like this change. I am happy with my current workflow. Push to a branch on my fork, wait for CI, create a PR if everything is green. Opening a PR against my fork seems unnecessary complicated. But that is just my workflow and I can deal with it if we want to change it. I am not in favour of this, but do not not merge this because of me. Maybe others agree that this is the right thing to do.

Sounds reasonable; PR updated; PTAL @adrianreber

kolyshkin:
Fixed PR description and commit message. @adrianreber PTAL

adrianreber:
@kolyshkin I really feel bad for so much discussion about this, but now CI does not run on push. If I understand GitHub Actions correctly the following means:
```
on:
  push:
    tags:
      - v*
    branches:
  pull_request:
```

Build on any pull request and only on pushes to tags starting with `v*`. This does not run on pushes to any other branch. Also not on pushes to `criu-dev`.

I would like to keep the existing:
```
on: [push, pull_request]
```

As far as I understood this PR, the change to not build on any push was your initial motivation for this PR and that would be completely gone.

kolyshkin:
@adrianreber you're right, my initial motivation was to fix when the jobs run, and it seems this does not need fixing. With that gone, this is just a consolidation and cleanup. There's little sense in consolidation per se; as for cleanups, I'll do it separately.

Closing, and sorry for wasting your time.

PR #1351, opened 2021-01-29
Tell podman to use vfs as storage-driver

adrianreber:
Just saw that the podman test is failing. This is still caused by the broken overlayfs in the Ubuntu kernel. Hopefully this will be fixed at some point in the future.

It seems each podman command now needs to be told to use the storage-driver vfs

### DIFF
--- a/scripts/ci/podman-test.sh
+++ b/scripts/ci/podman-test.sh
@@ -35,6 +35,7 @@ make install
 
 # overlaysfs behaves differently on Ubuntu and breaks CRIU
 # https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
+export STORAGE_DRIVER=vfs
 podman --storage-driver vfs info
 
 criu --version


adrianreber:
https://github.com/containers/podman/issues/9162

scripts/ci/podman-test.sh
```diff
@@ -35,50 +35,51 @@ make install
 
 # overlaysfs behaves differently on Ubuntu and breaks CRIU
 # https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
-podman --storage-driver vfs info
+PODMAN="podman --storage-driver vfs"
```
rst0git, line 38: Setting the storage-driver with an environment variable ([`export STORAGE_DRIVER=vfs`](http://docs.podman.io/en/latest/markdown/podman.1.html#storage-driver-value)), instead of adding the command-line option everywhere, might be a cleaner solution.

adrianreber: Yes, much simpler. Thanks. Changed.

adrianreber:
@avagin This fixes the Podman test failures we are seeing.

rst0git (approved):

Issue #1354, opened 2021-01-30
bpf test are failing on 5.10 Fedora 33


adrianreber:
The vagrant based tests using Fedora 33 on 5.10.10-200.fc33.x86_64 are failing in:
```
################### 2 TEST(S) FAILED (TOTAL 392/SKIPPED 32) ####################
 * zdtm/static/bpf_hash(ns)
 * zdtm/static/bpf_array(ns)
##################################### FAIL #####################################
```
with
```
======================== Run zdtm/static/bpf_hash in ns ========================
Start test
Test is SUID
./bpf_hash --pidfile=bpf_hash.pid --outfile=bpf_hash.out
The test failed (0, 1)
make: *** [Makefile:426: bpf_hash.pid] Error 1
 Test zdtm/static/bpf_hash FAIL at ['make', '--no-print-directory', '-C', 'zdtm/static', 'bpf_hash.pid'] 
Test output: ================================
07:18:42.761:     4: ERR: bpf_hash.c:27: Can't load key-value pairs to BPF map (errno = 9 (Bad file descriptor))
07:18:42.761:     4: FAIL: bpf_hash.c:170:  (errno = 9 (Bad file descriptor))
07:18:42.762:     3: ERR: test.c:316: Test exited unexpectedly with code 1

 <<< ================================

======================= Run zdtm/static/bpf_array in ns ========================
Start test
Test is SUID
./bpf_array --pidfile=bpf_array.pid --outfile=bpf_array.out
The test failed (0, 1)
make: *** [Makefile:426: bpf_array.pid] Error 1
 Test zdtm/static/bpf_array FAIL at ['make', '--no-print-directory', '-C', 'zdtm/static', 'bpf_array.pid'] 
Test output: ================================
07:20:15.380:     4: ERR: bpf_array.c:96: Can't create BPF map (errno = 1 (Operation not permitted))
07:20:15.380:     4: FAIL: bpf_array.c:171:  (errno = 1 (Operation not permitted))
07:20:15.380:     3: ERR: test.c:316: Test exited unexpectedly with code 1

 <<< ================================
```

This could be related to SELinux. Who can look into that? Either disable tests when running in enforcing mode or work towards getting the corresponding policies into SELinux? Or figure out if it is something else.

adrianreber:
Fixed in #1673

PR #1355, opened 2021-01-31
ci: skip bpf tests on vagrant

adrianreber:
See: https://github.com/checkpoint-restore/criu/issues/1354

Until we have a fix I think it would be better to skip tests which are known to fail.


adrianreber:
@avagin With the updated kernel in the vagrant tests the bpf tests are starting to fail (see #1354). This disables those tests until we have a real solution. On the 5.6 based Fedora kernel this error did not appear. On the 5.10 based Fedora bpf_hash and bpf_array fail. If there is a quick fix for #1354 we can also just drop this.

PR #1357, opened 2021-02-01
dump: dump has to fail if there is locks and --file-locks isn't set

avagin:
If criu finds a file lock and the --file-locks option isn't set, it
stops dumping processes, resumes them and exits with an error.

Cc: @dracoding

### DIFF
--- a/criu/file-lock.c
+++ b/criu/file-lock.c
@@ -108,6 +108,12 @@ int dump_file_locks(void)
 			continue;
 		}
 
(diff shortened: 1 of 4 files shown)


Snorch (approved):
Wow, thanks for the test, only now I see the problem clearly.

Looks good for me.

Note that we also can fail earlier in parse_fdinfo_pid_s as we add the lock to the file_lock_list there and set real_owner there too. But probably your way is cleaner.

PR #1361, opened 2021-02-01
ci: move asan and image streamer test to github

adrianreber:
This moves two more tests from Travis to GitHub Actions.

At Travis we now only have the aarch64 Graviton2, ppc64le and s390x tests remaining. As far as I know there is no CI system supporting those. We could, if we have access to any of those architectures, try to work with self-hosted GitHub Actions runners.

There are also still the Docker based tests. My last information was that Docker was broken. Not sure. @rst0git do you know if Docker works again with checkpoint/restore? Can you look into moving those tests to GitHub Actions?

### DIFF
--- a/.github/workflows/fedora-asan-test.yml
+++ b/.github/workflows/fedora-asan-test.yml
@@ -1,0 +1,12 @@
+name: Fedora ASAN Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ubuntu-20.04
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run Fedora ASAN Test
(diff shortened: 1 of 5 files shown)


rst0git (approved):

rst0git:
> There are also still the Docker based tests. My last information was that Docker was broken. Not sure. @rst0git do you know if Docker works again with checkpoint/restore?

As far as I know, the problem is with containerd, not Docker (moby). The latest version that works with checkpoint/restore is containerd [v1.2.14](https://github.com/containerd/containerd/releases/tag/v1.2.14).

> Can you look into moving those tests to GitHub Actions?

I can move the tests to GitHub Actions using that version.

PR #1362, opened 2021-02-04
ci: move Travis CI Docker tests to GitHub Actions

rst0git:


.github/workflows/docker-test.yml
```diff
@@ -0,0 +1,14 @@
+name: Docker Test
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-18.04, ubuntu-20.04]
```
adrianreber, line 10: Not sure we need to run it on 18.04 and 20.04. The reason we had 16.04 is that it still had a working kernel in combination with overlayfs. Not sure we gain anything with running it on 18.04 and 20.04.

On the other hand it does not really hurt to have it on both releases. My hope is that Ubuntu will no longer have the problems with 22.04 regarding overlayfs and then we can easily replace 18.04 with 22.04.

Now that I am at the end of my comment it feels unnecessary at all :grin:

adrianreber (approved):
Looks good. Thanks. As this was the last test using 16.04 I think we can finally drop the 16.04 conditionals once this is merged.

adrianreber:
@avagin Can you merge this one. This is the last CI missing from Travis.

PR #1363, opened 2021-02-05
show relative timestamps in criu logs

rakki-18:
Aims to solve #341 (showing relative timestamps in criu logs).
Introduces a new timeval variable that stores the time of the previous log.
Time difference in logs is calculated as the difference between current time and the previous log time.


rakki-18:
@xemul can you please review this?

avagin:
Can you add the --relative-timestamps option? I think we want to have absolute timestamps by default.

rakki-18:
@avagin I have added the --relative-timestamps option. Can you please review it?

compel/plugins/std/log.c
```diff
@@ -54,10 +60,19 @@ static void sbuf_log_init(struct simple_buf *b)
...
 
 		std_gettimeofday(&now, NULL);
+
+		if(relative_timestamps)
+			temp = now;
+
 		timediff(&start, &now);
 
+		if(relative_timestamps)
+			start = temp;
```
avagin, line 74: Is this function called from a few threads concurrently? If I remember correctly, this can happen when threads are restored. It means that we need to use locks or atomic operations here. Or in an ideal case, we can think about how to make per-thread relative timestamps.

rakki-18: My implementation of `relative_timestamp` option is similar to `cur_loglevel`. But there were no locks used for `cur_loglevel`.
So, I don't think `relative_timestamps` is being called concurrently here.

compel/src/lib/log.c
```diff
@@ -22,6 +23,12 @@ unsigned int compel_log_get_loglevel(void)
 	return current_loglevel;
 }
 
+void compel_relativetimestamps_init(compel_log_fn log_fn, int timestamps)
+{
+	logfn = log_fn;
```
Snorch, line 28: "logfn = log_fn;" looks excess.

criu/config.c
```diff
@@ -480,6 +481,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{ "close",			required_argument,	0, 1043	},
 		BOOL_OPT("log-pid", &opts.log_file_per_pid),
 		{ "version",			no_argument,		0, 'V'	},
+		BOOL_OPT("relative-timestamps", &opts.relative_timestamps),
```
Snorch, line 484: Boolopt will obviously have no short option 'T', which is used in next hunk.

criu/config.c
```diff
@@ -281,6 +281,7 @@ void init_opts(void)
 	opts.log_level = DEFAULT_LOGLEVEL;
 	opts.pre_dump_mode = PRE_DUMP_SPLICE;
 	opts.file_validation_method = FILE_VALIDATION_DEFAULT;
+	opts.relative_timestamps = 0;
```
Snorch, line 284: You should add DEFAULT_RELATIVETIMESTAMPS for this assignment.

criu/log.c
```diff
@@ -69,9 +69,17 @@ static void timediff(struct timeval *from, struct timeval *to)
 static void print_ts(void)
 {
 	struct timeval t;
+	/* temporary variable to store t*/
+	struct timeval temp;
 
 	gettimeofday(&t, NULL);
+
```
Snorch, line 76:
```
s/\s\s*$//
```

Please follow the codding style, empty-line should not contain spaces.

criu/log.c
```diff
@@ -217,7 +217,6 @@ int log_init(const char *output)
 {
 	int new_logfd, fd;
 
-	gettimeofday(&start, NULL);
```
Snorch: You should follow CONTRIBUTING.md guide. No fixup commits allowed.

rakki-18:
@Snorch I have made the changes. Can you check now?

compel/plugins/std/log.c
avagin: cur_loglevel is never changed, so we can read it concurrently without locks. This case is different...

Look at the code where we fork threads:
https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/pie/restorer.c#L1811

rakki-18: Ok, I will try to make the changes.

rakki-18:
I was not calling `compel_relativetimestamps_init()` in `criu/` before. I have added it now.

Snorch (changes requested):
You should add some words about WHAT you do and HOW into the commit description. So that reviewer and users of your option does not need to guess what is "relative" and relative to what.

Adding some words into Documentation and usage would also be nice.

compel/plugins/std/log.c
```diff
@@ -40,6 +41,11 @@ static inline void pad_num(char **s, int *n, int nr)
 	}
 }
 
+void std_log_set_relativetimestamps(enum __compel_relative_timestamps timestamps)
```
Snorch, line 44: Don't see where it's used. Excess?

compel/include/uapi/loglevels.h
```diff
@@ -17,4 +17,9 @@ enum __compel_log_levels
 	COMPEL_DEFAULT_LOGLEVEL	= COMPEL_LOG_WARN
 };
 
+enum __compel_relative_timestamps
+{
+	COMPEL_DEFAULT_RELATIVETIMESTAMPS
+};
```
Snorch, line 23: I don't like this, and would prefer a define to 0.

rakki-18:
@Snorch I will try to make the changes.

PR #1342, opened 2021-01-24
A few minor fixes

avagin:

PR #1376, opened 2021-02-25
namespaces: properly handle errors of snprintf

Snorch:
If snprintf was truncated we should probably know about it instead of
continuing to increase off, as snprintf returns number of characters
which would have been written and not the number which was actually
written.

Normally we check snprintf only for overflow not for error, some modern
compilers print warnings if truncation was not checked.

Probably it was the case why we implemented [1], the truncation happened
and on the next iteration of for loop we've hit negative size for
snprintf and got -1.

Fixes: 90f043dea ("namespaces: handle errors of snprintf") [1]


Snorch:
Related to https://github.com/checkpoint-restore/criu/pull/1342, the problem there seems not fixed completely.

Issue #1372, opened 2021-02-19
ci: arm build failure


adrianreber:
On Debian Testing and Unstable and Ubuntu 20.10 arm7vl builds are failing (see https://build.opensuse.org/project/show/devel:tools:criu)
```
[  330s]   CC       criu/protobuf-desc.o
[  330s] criu/plugin.c: In function 'cr_plugin_init':
[  330s] criu/plugin.c:246:36: error: '%s' directive output may be truncated writing between 3 and 2147483645 bytes into a region of size 4095 [-Werror=format-truncation=]
[  330s]   246 |   snprintf(path, sizeof(path), "%s/%s", opts.libdir, de->d_name);
[  330s]       |                                    ^~
[  330s] In file included from /usr/include/stdio.h:866,
[  330s]                  from criu/plugin.c:5:
[  330s] /usr/include/arm-linux-gnueabihf/bits/stdio2.h:70:10: note: '__builtin___snprintf_chk' output 5 or more bytes (assuming 2147483647) into a destination of size 4096
[  330s]    70 |   return __builtin___snprintf_chk (__s, __n, __USE_FORTIFY_LEVEL - 1,
[  330s]       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[  330s]    71 |        __bos (__s), __fmt, __va_arg_pack ());
[  330s]       |        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[  330s]   CC       criu/protobuf.o
[  330s] cc1: all warnings being treated as errors
[  330s] make[3]: *** [/usr/src/packages/BUILD/scripts/nmk/scripts/build.mk:119: criu/plugin.o] Error 1
[  330s] make[3]: *** Waiting for unfinished jobs....
[  333s] make[2]: *** [criu/Makefile:76: criu/built-in.o] Error 2
[  333s] make[1]: *** [Makefile:250: criu] Error 2
[  333s] make[1]: Leaving directory '/usr/src/packages/BUILD'
[  333s] dh_auto_build: error: make -j6 "INSTALL=install --strip-program=true" returned exit code 2
```

PR #1377, opened 2021-02-25
plugin: check for plugin path truncation

Snorch:
New compilators print warnings if snprintf return value is not checked
for truncation. Let's make them happy.

Fixes: #1372

### DIFF
--- a/criu/plugin.c
+++ b/criu/plugin.c
@@ -243,7 +243,11 @@ int cr_plugin_init(int stage)
 		if (len < 3 || strncmp(de->d_name + len - 3, ".so", 3))
 			continue;
 
-		snprintf(path, sizeof(path), "%s/%s", opts.libdir, de->d_name);
+		if (snprintf(path, sizeof(path), "%s/%s", opts.libdir, de->d_name) >=
+		    sizeof(path)) {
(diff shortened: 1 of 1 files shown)


criu/plugin.c
```diff
@@ -243,7 +243,11 @@ int cr_plugin_init(int stage)
 		if (len < 3 || strncmp(de->d_name + len - 3, ".so", 3))
 			continue;
 
-		snprintf(path, sizeof(path), "%s/%s", opts.libdir, de->d_name);
+		if (snprintf(path, sizeof(path), "%s/%s", opts.libdir, de->d_name) >=
+		    sizeof(path)) {
```
adrianreber, line 247: Instead of `sizeof(path)` wouldn't it be simpler to just use `PATH_MAX`? In `snprintf()` and in the compare?

Snorch: I actually like sizeof() option more as a more universal way, and as less error prone. I can do better and add size variable so that we don't need to calculate sizeof twice. I thought prefered to make diff smaller. -> will add size variable if you like.

adrianreber: Not necessary from my point of view.

adrianreber (approved):

Issue #1368, opened 2021-02-17
aarch64 rawhide test runs are failing


adrianreber:
```
In file included from compel/include/uapi/asm/infect-types.h:7,
                 from compel/src/lib/infect.c:20:
/usr/include/sys/ptrace.h:86:3: error: expected identifier before numeric constant
   86 |   PTRACE_SYSEMU = 31,
      |   ^~~~~~~~~~~~~
```

I opened https://bugzilla.redhat.com/show_bug.cgi?id=1929819. Should be solved soon.

adrianreber:
This has been fixed in Fedora.

PR #1379, opened 2021-02-26
compel/criu: Restore extended registers set on parasite curing

0x7f454c46:
There's an ugly issue in compel: it saves extended registers into images but doesn't
keep a copy in memory for threads. Thread-leader gets his register set restored with
`rt_sigreturn()`, but sub-threads suffer. In case the parasite changes any ext. register
(i.e, by using fpu/sse) than it stays corrupted in the dumpee-thread.
For pre-dump or leave-running case it resulted in weird, obfuscated failures.

Previously, PIEs were conservatively built with options that restrict optimisations
and still, some hits were appearing.

Let's properly address it by restoring fpu/xsave in threads (and ext. regs set on non-x86).
That will allow gcc to optimize routines like `strcpy()` and `memcmp()` with sse opts.
Registers are preserved inside thread context, with the exception of the main thread that
gets his proper regset on rt_sigreturn jump at parasite's self-curing.

I have tested it on x86, it builds on other architectures, but if it adds any issues,
the per-arch `compel_set_task_ext_regs()` can be ifdeffed-out as a temporary measure
(then it will work just like it "was" before).

I see Alpine fails with:
```
2021-02-26T03:24:22.7875002Z b'(00.062786) page-pipe: Created tee pipe size 4194304'
2021-02-26T03:24:22.7875876Z b'(00.079281) Error (criu/tls.c:321): tls: Pull callback recv failed: Connection reset by peer'
2021-02-26T03:24:22.7876822Z b'(00.079296) Error (criu/tls.c:147): tls: Failed receiving data: Error in the pull function.'
2021-02-26T03:24:22.7877812Z b"(00.079300) Error (criu/page-xfer.c:1225): page-xfer: Can't read pagemap from socket: I/O error"
2021-02-26T03:24:22.7878745Z b'(00.079308) Error (criu/tls.c:309): tls: Push callback send failed: Broken pipe'
2021-02-26T03:24:22.7879519Z ------------------------ ERROR OVER ------------------------
2021-02-26T03:24:22.7880352Z ##### Test zdtm/transition/lazy-thp FAIL at criu page-server exited with 1 #####
```
I hope it's expected.

Fixes: #1317
Reverts: #908


adrianreber:
The alpine error is known CI failure. I do not think there is actually an open issue tracking it, but it was discussed in #1277 

I have seen that the Fedora Rawhide x86_64 tests times out also on other runs. Not sure if that is a real bug or a problem with the CI environment.

The Fedora Rawhide aarch64 error is also already tracked at: #1368

rst0git:
> I do not think there is actually an open issue tracking it

I just opened https://github.com/checkpoint-restore/criu/issues/1380.

0x7f454c46:
@adrianreber thanks for your inputs, I appreciate them!

adrianreber:
Please rebase

cyrillos (approved):
Looks ok from a glance. Thanks!

0x7f454c46:
Rebased on criu-dev with minor conflict in
```
 struct thread_ctx {
        k_rtsigset_t            sigmask;
        user_regs_struct_t      regs;
+#ifdef ARCH_HAS_PTRACE_GET_THREAD_AREA
+       tls_t                   tls;
+#endif
        user_fpregs_struct_t    ext_regs;
 };
```

adrianreber:
I cannot say anything about the code changes, but it is nice to see a fix for #1317 which has been giving us CI failures for the last months. Thanks.

Issue #1380, opened 2021-02-26
lazy-thp: fail at criu page-server


rst0git:
#### Description

The `lazy-thp` test fails when running in an Alpine container. (https://github.com/checkpoint-restore/criu/pull/1277)

#### Steps to reproduce the issue:

```sh
sudo podman run -it --rm \
    --privileged \
    --mount type=tmpfs,tmpfs-size=512M,destination=/tmp \
    --mount type=tmpfs,tmpfs-size=512M,destination=/run \
    --network=host \
    --cgroupns=host \
    -v /lib/modules:/lib/modules \
    alpine sh
```

```sh
apk update && apk add \
gcc \
bash \
build-base \
coreutils \
git \
gnutls-dev \
libaio-dev \
libcap-dev \
libnet-dev \
libnl3-dev \
nftables \
nftables-dev \
pkgconfig \
protobuf-c-dev \
protobuf-dev \
py3-pip \
py3-protobuf \
python3 \
ip6tables \
iptables \
nftables \
iproute2 \
bash \
e2fsprogs \
py-yaml \
tar

pip3 install junit_xml

git clone https://github.com/checkpoint-restore/criu
cd criu && make -j4
ln -s /usr/bin/python3 /usr/bin/python
python3 ./test/zdtm.py run -t zdtm/transition/lazy-thp --remote-lazy-pages --tls
```

#### Describe the results you received:
```
====================== Run zdtm/transition/lazy-thp in h =======================
Start test
./lazy-thp --pidfile=lazy-thp.pid --outfile=lazy-thp.out
Run criu dump
=[log]=> dump/zdtm/transition/lazy-thp/141/1/dump.log
------------------------ grep Error ------------------------
b'(00.007220) Parasite syscall_ip at 0x560878ee0000'
b'(00.007546) Set up parasite blob using memfd'
b'(00.007634) Putting parasite blob into 0x7f6766617000->0x7f38ddd28000'
b'(00.007749) Dumping general registers for 141 in native mode'
b'(00.007759) Warn  (compel/arch/x86/src/lib/infect.c:341): Will restore 141 with interrupted system call'
------------------------ ERROR OVER ------------------------
Run criu page-server
Run criu lazy-pages
Run criu restore
Send the 15 signal to  141
Wait for zdtm/transition/lazy-thp(141) to die for 0.100000
=[log]=> dump/zdtm/transition/lazy-thp/141/1/page-server.log
------------------------ grep Error ------------------------
[...]

PR #1383, opened 2021-02-28
ci: Skip lazy-thp for remote pages over tls

rst0git:
Temporarily disable the lazy-thp test for remote lazy pages with enabled tls until #1380 is resolved.

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -221,8 +221,10 @@ LAZY_OPTS="-p 2 -T $LAZY_TESTS $LAZY_EXCLUDE $ZDTM_OPTS"
 ./test/zdtm.py run $LAZY_OPTS --lazy-pages
 # shellcheck disable=SC2086
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages
+# FIXME: post-copy migration of THP over TLS (sometimes) fails with:
+#     Error (criu/tls.c:321): tls: Pull callback recv failed: Connection reset by peer
 # shellcheck disable=SC2086
-./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
+./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls -x lazy-thp
 
 bash -x ./test/jenkins/criu-fault.sh
 if [ "$UNAME_M" == "x86_64" ]; then


rst0git:
cc: @adrianreber @0x7f454c46

criu/uffd.c
```diff
@@ -1249,14 +1249,14 @@ static void lazy_pages_summary(struct lazy_pages_info *lpi)
 	lp_debug(lpi, "UFFD transferred pages: (%ld/%ld)\n",
 		 lpi->copied_pages, lpi->total_pages);
 
-#if 0
+
 	if ((lpi->copied_pages != lpi->total_pages) && (lpi->total_pages > 0)) {
```
rppt, line 1253: The same information is already printed above without the frightening text ;-)

The real problem if that `lpi->total_pages` only shows the initial count of the pages at the dump time, but it is not updated when some pages are unmapped or dropped with `madvise(MADV_DONTNEED)`.

rst0git: Thank you Mike, I have dropped this commit.

When running the lazy-thp test it fails but not always, and when it fails the output in `lazy-pages.log` is as follows:

successful: [lazy-pages.log](https://github.com/checkpoint-restore/criu/files/6058187/lazy-pages-log-1.txt)
```
UFFD transferred pages: (1033/1033)
```

failed: [lazy-pages.log](https://github.com/checkpoint-restore/criu/files/6058188/lazy-pages-log-2.txt)
```
UFFD transferred pages: (137/1033)
```

rppt (approved):
I hope @rst0git will remember to re-enable it :)

scripts/ci/run-ci-tests.sh
```diff
@@ -235,7 +235,7 @@ LAZY_OPTS="-p 2 -T $LAZY_TESTS $LAZY_EXCLUDE $ZDTM_OPTS"
 # shellcheck disable=SC2086
 ./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages
 # shellcheck disable=SC2086
-./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls
+./test/zdtm.py run $LAZY_OPTS --remote-lazy-pages --tls -x lazy-thp
```
avagin, line 238: pls add a fixme command that explains why we need to skip lazy-thp here

rst0git: Thank you Andrei, I added a "fixme" comment about the error.

Issue #1332, opened 2021-01-05
Restore failed with Error (criu/pstree.c:452): BUG at criu/pstree.c:452


zcdb:
Hello, i want to dump/restore a process in different container but I ran into some problems.
Below is my test environment：
docker version:
19.03
kernel version:
3.10.0-327.28.3.el7.x86_64
criu version:
3.15

Below is what i did：
First, i dump in one container:
criu dump -o dump.log -v4 -j -t pid --tcp-close
and result:
[dump.log](https://github.com/checkpoint-restore/criu/files/5769622/dump.log)

Then, i restore in another container:
criu restore -o restore.log -v4 -d -j --tcp-close
but failed, the pid in the log does not match the above, because I lost the error log, but the error message is the same:
[restore_error.log](https://github.com/checkpoint-restore/criu/files/5769612/restore_error.log)

However, when i try the restore again, it successed:
[restore.log](https://github.com/checkpoint-restore/criu/files/5769626/restore.log)

Can anyone help explain this situation? Why did it fail the first time？

zcdb:
I found that this problem occurs in the code of the dev branch, and 3.15 is no problem.

avagin:
@Snorch could you take a look at this?

Snorch:
From dump log we see:

```
(00.027817)     Seizing 37's 223 thread
(00.027888) seccomp: Collected tid_real 223 mode 0
```

From restore log we see (I hope restore.log corresponds to dump log):

```
(00.048039) Migrating process tree (SID 14->15)
(00.048042) Migrating process tree (GID 28->223)
(00.048043) Error (criu/pstree.c:452): BUG at criu/pstree.c:452
```

With high probability I can guess that the session id (sid) of your task which 
called criu accidentally conincided with thread id (tid) of one of dumped 
processes, and criu lookup_create_item instead of creating new item for a new
sid found some existing item (thread) and it was luckily catched by BUG 
statement.

So probably the answer here is that `-j` option should not be used when dumping
too much processes. These sid/gid replacement is racy and could've coninsided 
with some regular (non-thread) task and we would not be lucky to detect reuse
and that would leed to other problems when restoring process ids (likely the 
tree topology would be unrestorable).

You should use pidns to fix such pidreuse problems for good.

Note1:

> However, when i try the restore again, it successed

You've likely just changed sid/gid of a process calling criu and that fixed the race.

Note2:

[...]

PR #1384, opened 2021-03-01
pstree: check for pid collision before switching to new sid/gid

Snorch:
Without this check we can hit the BUG in lookup_create_item just a few
steps later (if one thread in images has same pid with new sid/gid). And
also this check saves us from different sorts of unexpected errors on
restore (if one non-thread task in images has same pid/sid/gid already).

Fixes: #1332


Snorch:
I have a failing java test, but I don't see criu error only:

https://github.com/checkpoint-restore/criu/pull/1384/checks?check_run_id=2002531261
```java.lang.AssertionError: FileRead: ERROR: Error during restore: exitCode of restore process was not zero.```

PR #1385, opened 2021-03-01
ci: fix Fedora rawhide CI failures

adrianreber:
It seems the Fedora rawhide /tmp is no longer 1777 but 755.

Change it back to 1777 to make our CI runs successful again.


adrianreber:
This fixes the timeout we are seeing in Fedora Rawhide x86_64 based tests

PR #1389, opened 2021-03-04
Fix s390x Jenkins vt test failure

adrianreber:
The new Jenkins s390x test host does not have `/dev/ttyS0` try `/dev/ttysclp0` which seems to exist on multiple hosts we checked.

Also add a checkskip for the vt test to not run it if the device used in the test does not exist.

@avagin : as discussed

### DIFF
--- a/test/zdtm/static/vt.c
+++ b/test/zdtm/static/vt.c
@@ -16,7 +16,7 @@ char *filename;
 TEST_OPTION(filename, string, "file name", 1);
 
 #ifdef __s390x__
-#define MINOR	64 /* ttyS0 */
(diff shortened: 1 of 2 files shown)


avagin:
I am ready to merge "￼test: move vt test to minor 65 on s390x" and I am not sure that we need "test: skip vt test if a device does not exist".

adrianreber:
Do you want me to drop the patch from the PR or can you drop it some other way. No problem from my side if we do not need it.

avagin:
Yes, let drop "￼test: skip vt test if device does not exist" and go with the second patch.

PR #1397, opened 2021-03-10
ci: fix recode errors seen in Jenkins

adrianreber:
Although we are running crit-recode.py also in all CI runs we never seen
following error except in Jenkins:
```
  File "/usr/lib/python3.8/base64.py", line 510, in _input_type_check
    m = memoryview(s)
TypeError: memoryview: a bytes-like object is required, not 'str'

The above exception was the direct cause of the following exception:

  File "./test/crit-recode.py", line 25, in recode_and_check
    r_img = pycriu.images.dumps(pb)
  File "/var/lib/jenkins/workspace/Q/test/pycriu/images/images.py", line 635, in dumps
    dump(img, f)
  File "/var/lib/jenkins/workspace/Q/test/pycriu/images/images.py", line 626, in dump
    handler.dump(img['entries'], f)
  File "/var/lib/jenkins/workspace/Q/test/pycriu/images/images.py", line 289, in dump
    f.write(base64.decodebytes(item['extra']))
  File "/usr/lib/python3.8/base64.py", line 545, in decodebytes
    _input_type_check(s)
  File "/usr/lib/python3.8/base64.py", line 513, in _input_type_check
    raise TypeError(msg) from err
TypeError: expected bytes-like object, not str
```
This commit fixes this by encoding the string to bytes.

### DIFF
--- a/lib/py/images/images.py
+++ b/lib/py/images/images.py
@@ -284,9 +284,9 @@ class ghost_file_handler:
                 size = len(pb_str)
                 f.write(struct.pack('i', size))
                 f.write(pb_str)
-                f.write(base64.decodebytes(item['extra']))
+                f.write(base64.decodebytes(str.encode(item['extra'])))
         else:
-            f.write(base64.decodebytes(item['extra']))
+            f.write(base64.decodebytes(str.encode(item['extra'])))
 
     def dumps(self, entries):
         f = io.BytesIO('')
--- a/lib/py/images/pb2dict.py
+++ b/lib/py/images/pb2dict.py
@@ -246,11 +246,11 @@ def encode_dev(field, value):
 
 
 def encode_base64(value):
-    return base64.encodebytes(value)
+    return base64.encodebytes(value).decode()
 
 
 def decode_base64(value):
-    return base64.decodebytes(value)
+    return base64.decodebytes(str.encode(value))
 
 
 def encode_unix(value):
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -281,7 +281,7 @@ ip net add test
 
 ./test/zdtm.py run --empty-ns -T zdtm/static/socket-tcp*-local --iter 2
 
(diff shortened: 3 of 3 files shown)


avagin:
[root@usr-LAN-dhcp-38 criu]# PYTHONPATH="$(pwd)/lib/" python3.9 ./crit/crit decode < test/dump/zdtm/static/env00/197/1/seccomp.img
  File "/root/criu/./crit/crit", line 6, in <module>
    cli.main()
  File "/root/criu/crit/pycriu/cli.py", line 410, in main
    opts["func"](opts)
  File "/root/criu/crit/pycriu/cli.py", line 32, in decode
    img = pycriu.images.load(inf(opts), opts['pretty'], opts['nopl'])
  File "/root/criu/crit/pycriu/images/images.py", line 575, in load
    m, handler = __rhandler(f)
  File "/root/criu/crit/pycriu/images/images.py", line 549, in __rhandler
    img_magic, = struct.unpack('i', f.read(4))
TypeError: a bytes-like object is required, not 'str'

adrianreber:
Is this with my patches or without?

avagin:
> ci: fix recode errors seen in Jenkins 
You change files in lib/images, so the "ci" prefix isn't good in this case.

avagin:
Should we stop supporting python2?

adrianreber:
> Should we stop supporting python2?

I would like to stop supporting Python 2, but RHEL/CentOS 7 is still available until 2024. Not sure if we want to drop support for that.

avagin:
> I would like to stop supporting Python 2, but RHEL/CentOS 7 is still available until 2024. Not sure if we want to drop support for that.

Is it possible to install python 3 there? I think the answer is yes, so it is not a problem.

adrianreber:
> > I would like to stop supporting Python 2, but RHEL/CentOS 7 is still available until 2024. Not sure if we want to drop support for that.
> 
> Is it possible to install python 3 there? I think the answer is yes, so it is not a problem.

Yes, Python 3 is available. We can do that.

PR #1399, opened 2021-03-10
lib: correctly handle stdin/stdout

adrianreber:
This changes stdin to be opened as binary if the input is not a tty.

This changes stdout to be opened as binary if encoding or if the output
is not a tty.

This should fix the error in #1397 
```
crit/crit-python3 decode < test/dump/zdtm/static/env00/56/1/seccomp.img 
  File "/home/criu/crit/crit-python3", line 6, in <module>
    cli.main()
  File "/home/criu/crit/pycriu/cli.py", line 410, in main
    opts["func"](opts)
  File "/home/criu/crit/pycriu/cli.py", line 32, in decode
    img = pycriu.images.load(inf(opts), opts['pretty'], opts['nopl'])
  File "/home/criu/crit/pycriu/images/images.py", line 575, in load
    m, handler = __rhandler(f)
  File "/home/criu/crit/pycriu/images/images.py", line 549, in __rhandler
    img_magic, = struct.unpack('i', f.read(4))
TypeError: a bytes-like object is required, not 'str'
```

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -268,6 +268,9 @@ ip net add test
 ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/transition/fork -t zdtm/static/ghost_holes00 -k always
 ./test/crit-recode.py
 
+# more crit testing
+make -C test/others/crit run
+
 # libcriu testing
 make -C test/others/libcriu run
 


lib/py/cli.py
```diff
@@ -11,14 +11,18 @@ def inf(opts):
     if opts['in']:
         return open(opts['in'], 'rb')
     else:
-        return sys.stdin
+        if sys.stdin.isatty():
```
avagin, line 14: pls add a comment that explains why we need this check

lib/py/cli.py
```diff
@@ -11,14 +11,18 @@ def inf(opts):
...
+            return sys.stdin
+        return sys.stdin.buffer
```
avagin, line 16: This will not work with python 2, will it? Should we force to use python3?

adrianreber: See #1405

lib/py/cli.py
adrianreber: Added a comment

adrianreber:
I enabled (and fixed) an existing crit test and extended the test to check for all possible input and output variants.

lib/py/cli.py
avagin: So we decide that we want to work with python2, but this code will not work with it:

$ python2 -c 'import sys; print sys.stdin.buffer'Traceback (most recent call last):
  File "<string>", line 1, in <module>
AttributeError: 'file' object has no attribute 'buffer'

should we do something like this:
```
if "buffer" in dir(sys.stdin):
  return sys.stdin.buffer # python3
else:
  return sys.stdin # python2
```

avagin: @adrianreber are you going to fix this?

adrianreber: I think it is. Although I also have trouble reading this here in GitHub, but line 14 contains: `if (sys.version_info < (3, 0)):`

Issue #1390, opened 2021-03-05
CI tests (CentOS 8) are failing because of a tainted kernel


adrianreber:
The CentOS 8 test case in a container started to fail with:
```
    Traceback (most recent call last):
      File "./test/zdtm.py", line 2685, in <module>
        opts['action'](opts)
      File "./test/zdtm.py", line 2365, in run_tests
        launcher.run_test(t, tdesc, run_flavs)
      File "./test/zdtm.py", line 1965, in run_test
        (taint, self.__taint))
    Exception: The kernel is tainted: '512\n' ('0\n')
```
This seems to be a new error and only on the newest GitHub Actions kernel (5.4.0-1040-azure). The following warning can be seen in `dmesg`:
```
[  226.964883] writing to auto_msgmni has no effect
[  226.978221] ------------[ cut here ]------------
[  226.978226] refcount_t hit zero at __nft_mt_tg_destroy+0x20/0x30 [nft_compat] in kworker/u4:4[103], uid/euid: 0/0
[  226.978232] WARNING: CPU: 1 PID: 103 at kernel/panic.c:677 refcount_error_report+0x9b/0xab
[  226.978233] Modules linked in: nft_counter xt_mark nft_compat nf_tables raw_diag udp_diag tcp_diag inet_diag netlink_diag af_packet_diag unix_diag veth xfs dm_thin_pool dm_persistent_data dm_bio_prison dm_bufio xt_MASQUERADE nf_conntrack_netlink nfnetlink xfrm_user xfrm_algo xt_addrtype iptable_filter iptable_nat nf_nat br_netfilter bridge stp llc xt_owner iptable_security xt_conntrack nf_conntrack nf_defrag_ipv6 nf_defrag_ipv4 libcrc32c bpfilter aufs overlay binfmt_misc nls_iso8859_1 dm_multipath scsi_dh_rdac scsi_dh_emc scsi_dh_alua joydev hid_generic crct10dif_pclmul crc32_pclmul ghash_clmulni_intel aesni_intel crypto_simd cryptd glue_helper hv_netvsc serio_raw pata_acpi hv_balloon hyperv_fb hyperv_keyboard cfbfillrect hid_hyperv cfbimgblt hid cfbcopyarea hv_utils sch_fq_codel drm drm_panel_orientation_quirks i2c_core ip_tables x_tables autofs4
[...]

PR #1401, opened 2021-03-11
ci: ignore taint in container based tests

adrianreber:
Unfortunately the kernel in GitHub Actions has a bug (#1390) which breaks our CentOS 8 test run.

To avoid broken CI results this PR introduces a `ZDTM_IGNORE_TAINT` variable which is set in all container based runs.

### DIFF
--- a/scripts/ci/docker.env
+++ b/scripts/ci/docker.env
@@ -2,3 +2,6 @@ SKIP_CI_PREP=1
 ZDTM_OPTS=-x zdtm/static/binfmt_misc -x zdtm/static/sched_policy00
 CC=gcc
 SKIP_EXT_DEV_TEST=1
+# The Ubuntu kernel 5.4.0-1040-azure has a bug with CentOS 8 userspace
+# Without this flag the test will fail early because of being tainted (512 - warning)
+ZDTM_IGNORE_TAINT=1
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1961,8 +1961,9 @@ class Launcher:
         with open("/proc/sys/kernel/tainted") as taintfd:
             taint = taintfd.read()
         if self.__taint != taint:
-            raise Exception("The kernel is tainted: %r (%r)" %
-                            (taint, self.__taint))
+            if not opts["ignore_taint"]:
+                raise Exception("The kernel is tainted: %r (%r)" %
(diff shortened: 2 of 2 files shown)


test/zdtm.py
```diff
@@ -1961,8 +1961,9 @@ def run_test(self, name, desc, flavor):
         with open("/proc/sys/kernel/tainted") as taintfd:
             taint = taintfd.read()
         if self.__taint != taint:
-            raise Exception("The kernel is tainted: %r (%r)" %
-                            (taint, self.__taint))
+            if not opts["ignore_taint"]:
```
avagin, line 1964: Eh, it was intentional. The idea was to detect changes.

scripts/ci/docker.env
```diff
@@ -2,3 +2,4 @@ SKIP_CI_PREP=1
 ZDTM_OPTS=-x zdtm/static/binfmt_misc -x zdtm/static/sched_policy00
 CC=gcc
 SKIP_EXT_DEV_TEST=1
+ZDTM_IGNORE_TAINT=1
```
avagin, line 5: pls add a comment on why we need this.

test/zdtm.py
adrianreber: I expected that it was intentional. Not sure how to handle the test failure in another way. As long as the Ubuntu kernel breaks with a CentOS 8 user space it will always report an error which might hide some other error.

scripts/ci/docker.env
adrianreber: Comment added and force pushed.

test/zdtm.py
```diff
@@ -2672,6 +2672,13 @@ def clean_stuff(opts):
 cp.add_argument("what", choices=['nsroot'])
 
 opts = vars(p.parse_args())
+
+try:
+    if os.environ["ZDTM_IGNORE_TAINT"] == "1":
```
avagin, line 2677: if os.getenv("ZDTM_IGNORE_TAINT") == "1":
  opts["ignore_taint"] = True

In this case, you will not need to handle the KeyError exception.

adrianreber:
We moved CentOS 8 based tests to a CentOS 8 based kernel on Cirrus CI and do not have to workaround broken kernels.

Issue #1387, opened 2021-03-03
Differentiate when parent symlink does not exists and when it is invalid


MTBorg:
When trying to open a parent image there is no distinction between 1) when a symlink does not exist and 2) when it does not point to a valid target. See below for reference

https://github.com/checkpoint-restore/criu/blob/a76d186d9dc984213cc1d2022de8ecf7232657c2/criu/pagemap.c#L655-L691

This causes problems when passing an invalid path, i.e. non-existing directory (case 2), using runc's `--parent-path` option (which i believe uses criu's `--prev-images-dir` option), because criu silently fails and performs a (full?) dump. However, the user is never notified of this and thus might be lead to believe that a predump was successfully made.

To reproduce:

```shell
# creates a predump exits with no error. 
runc checkpoint --image-path dumps/1 <container name>
```
then
```shell
# exits with no error and creates a dump with the 
# parent symlink pointing to a non-existing directory.
runc checkpoint --image-path dumps/2 --parent-path ../this_path_does_not_exist/1 <container name>
```

I'd argue either that the behavior is correct but the user should be notified, or criu should not perform a dump and fail.

PR #1403, opened 2021-03-13
criu: Throw error when parent path is provided but invalid

minhbq-99:
In image dump directory, there are 2 parent symlink error cases:
- Parent symlink does not exist
- Parent symlink exists but points to invalid target

At the moment, 2 cases are handled exactly the same (do full dump). However, while
the first case happen when parent path is not provided, the second one is likely
user's mistake when provides invalid parent path.

So we throw an error in the latter case instead of performing the full dump.

Fixes: #1387.

### DIFF
--- a/criu/image.c
+++ b/criu/image.c
@@ -149,8 +149,7 @@ InventoryEntry *get_parent_inventory(void)
 	InventoryEntry *ie;
 	int dir;
 
-	dir = openat(get_service_fd(IMG_FD_OFF), CR_PARENT_LINK, O_RDONLY);
-	if (dir == -1) {
+	if (open_parent(get_service_fd(IMG_FD_OFF), &dir)) {
 		/*
 		 * We print the warning below to be notified that we had some
 		 * unexpected problem on open. For instance we have a parent
@@ -158,10 +157,11 @@ InventoryEntry *get_parent_inventory(void)
 		 * when also having no parent directory is an expected case of
 		 * first dump iteration.
 		 */
-		if (errno != ENOENT)
-			pr_warn("Failed to open parent directory\n");
+		pr_warn("Failed to open parent directory\n");
 		return NULL;
 	}
+	if (dir < 0)
+		return NULL;
 
 	img = open_image_at(dir, CR_FD_INVENTORY, O_RSTR);
 	if (!img) {
@@ -561,6 +561,11 @@ int open_image_dir(char *dir, int mode)
 		if (img_streamer_init(dir, mode) < 0)
 			goto err;
 	} else if (opts.img_parent) {
+		if (faccessat(fd, opts.img_parent, R_OK, 0)) {
+			pr_perror("Invalid parent image directory provided");
+			goto err;
+		}
+
 		ret = symlinkat(opts.img_parent, fd, CR_PARENT_LINK);
 		if (ret < 0 && errno != EEXIST) {
 			pr_perror("Can't link parent snapshot");
@@ -584,6 +589,24 @@ void close_image_dir(void)
 	if (opts.stream)
 		img_streamer_finish();
 	close_service_fd(IMG_FD_OFF);
+}
+
+int open_parent(int dfd, int *pfd)
+{
+	*pfd = -1;
+	/* Check if the parent symlink exists */
+	if (faccessat(dfd, CR_PARENT_LINK, F_OK, AT_SYMLINK_NOFOLLOW) && errno == ENOENT) {
+		pr_debug("No parent images directory provided\n");
+		return 0;
+	}
+
+	*pfd = openat(dfd, CR_PARENT_LINK, O_RDONLY);
+	if (*pfd < 0) {
+		pr_perror("Can't open parent path");
+		return -1;
+	}
+
+	return 0;
 }
 
 static unsigned long page_ids = 1;
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -364,10 +364,8 @@ static int open_page_local_xfer(struct page_xfer *xfer, int fd_type, unsigned lo
 		return -1;
 
 	xfer->pi = open_pages_image(O_DUMP, xfer->pmi, &pages_id);
-	if (!xfer->pi) {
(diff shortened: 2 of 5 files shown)


adrianreber:
Thanks for working on this. Could you also make sure that both possibilities ('Parent symlink does not exist' and 'Parent symlink exists but points to invalid target') are covered by a test?

minhbq-99:
Hi, I have added a dummy test for case when 'Parent symlink exists but points to invalid target'. Case when 'Parent symlink does not exist' is just normal case when user does not provide '--prev-images-dir' so I think other tests may cover it already.

criu/pagemap.c
```diff
@@ -656,15 +656,23 @@ static int try_open_parent(int dfd, unsigned long id, struct page_read *pr, int
...
+	ret = fstatat(dfd, CR_PARENT_LINK, &st, AT_SYMLINK_NOFOLLOW);
+	if (ret < 0 && errno == ENOENT)
 		goto out;
 
+	pfd = openat(dfd, CR_PARENT_LINK, O_RDONLY);
+	if (pfd < 0 && errno == ENOENT) {
+		pr_perror("Can't open parent path\n");
+		goto err;
+	}
+
```
rppt, line 675: These checks are repeated several times, maybe it's worth splitting them into a helper function?

rppt (commented):
Overall looks fine, just one nit below. 
I'd also mention in changelog that the restore behavior is changed

criu/page-xfer.c
```diff
@@ -386,14 +384,18 @@ static int open_page_local_xfer(struct page_xfer *xfer, int fd_type, unsigned lo
...
-		pfd = openat(get_service_fd(IMG_FD_OFF), CR_PARENT_LINK, O_RDONLY);
-		if (pfd < 0 && errno == ENOENT)
-			goto out;
+		pfd = open_parent(get_service_fd(IMG_FD_OFF));
+		if (pfd < 0) {
+			if (pfd == -1)
+				goto out;
+			else
+				goto err_cli;
+		}
```
rppt, line 393: I think it would have been nicer to do something like
```
if (open_parent(get_service_fd(IMG_FD_OFF), &pfd))
	goto err_cli;
if (pfd < 0)
	goto out;
```
rather than use `-1` and `-2` as the return values of `open_parent()`.

minhbq-99:
I cleaned up the code as @rppt 's suggestion. Look like my rebase makes the history hard to to look at, I'm so sorry for that.

minhbq-99:
I cleaned up the code as @rppt 's suggestion. Look like my rebase makes the history hard to to look at, I'm so sorry for that.

Snorch (changes requested):
You probably don't handle the problem in get_parent_inventory() and open_irmap_cache().

Note that we use CR_PARENT_LINK in many other places.

Note that we might also want to check symlink is openabe just after a creation in open_image_dir().

minhbq-99:
Hi @Snorch, thank you for your review.
> Note that we might also want to check symlink is openabe just after a creation in open_image_dir().

After looking at the code more carefully, I think we should add a check for provided parent path before creating parent symlink
https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/image.c#L563-L564
with something like this

```
if (faccessat(fd, opts.img_parent, R_OK, 0)) {
	pr_perror("Invalid parent image directory provided");
	goto err;
}
```

So later in the code, we don't need to check this case anymore.

Snorch:
> Hi @Snorch, thank you for your review.
> 
> > Note that we might also want to check symlink is openabe just after a creation in open_image_dir().
> 
> After looking at the code more carefully, I think we should add a check for provided parent path before creating parent symlink
> 
> https://github.com/checkpoint-restore/criu/blob/c8beca65580da87b89964da3ffb454ca99aa6bc6/criu/image.c#L563-L564
> 
> with something like this
> 
> ```
> if (faccessat(fd, opts.img_parent, R_OK, 0)) {
> 	pr_perror("Invalid parent image directory provided");
> 	goto err;
> }
> ```
> 
> So later in the code, we don't need to check this case anymore.

You are right in general case checking path before creating a symlink pointing to this path would be enough. But in rare case that someone would accidentally or intentionally break the symlink later we probably still need the original aproach. (I think having both is the best =))

minhbq-99:
Hi, I added the checks in get_parent_inventory(), open_irmap_cache() and open_image_dir(). As in get_parent_inventory(), I think current check is fine but I still changed it to have the same open_parent() pattern. Another place that I can see using CR_PARENT_LINK is in cr_dedup() but the check looks good to me.

I removed the commit that creates the test because with the early check in open_image_dir(), image_dir is not created in failure case. This leads to error in zdtm when trying to grep_errors from dump.log in image_dir.

criu/image.c
```diff
@@ -561,6 +561,11 @@ int open_image_dir(char *dir, int mode)
 		if (img_streamer_init(dir, mode) < 0)
 			goto err;
 	} else if (opts.img_parent) {
+		if (faccessat(fd, opts.img_parent, R_OK, 0)) {
+			pr_perror("Invalid parent image directory provided\n");
```
rppt, line 565: Nit: there is no need for `\n` here because `pr_perror` will add it to the printed string just like `perror` does

rppt (commented):
Except the small nit about `pr_perror` looks Ok to me

rppt (approved):

criu/image.c
```diff
@@ -586,6 +591,24 @@ void close_image_dir(void)
 	close_service_fd(IMG_FD_OFF);
 }
 
+int open_parent(int dfd, int *pfd)
+{
+	*pfd = -1;
+	/* Check if the parent symlink exists */
+	if (faccessat(dfd, CR_PARENT_LINK, F_OK, AT_SYMLINK_NOFOLLOW)) {
+		pr_debug("No parent images directory provided\n");
```
avagin, line 599: pls check that errno is ENOENT here.

avagin:
Merged. Thanks a lot!

minhbq-99:
Thanks everyone for your reviews and feedbacks.

PR #1406, opened 2021-03-15
More Jenkins crit-recode fixes

adrianreber:
This fixes the next errors in the Jenkins crit-recode test runs.

### DIFF
--- a/lib/py/images/images.py
+++ b/lib/py/images/images.py
@@ -375,7 +375,7 @@ class ipc_sem_set_handler:
         s = array.array('H')
         if s.itemsize != sizeof_u16:
             raise Exception("Array size mismatch")
-        s.fromstring(f.read(size))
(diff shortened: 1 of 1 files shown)


lib/py/images/images.py
```diff
@@ -390,7 +390,8 @@ def dump(self, extra, f, pbuff):
         if len(s) != entry['nsems']:
             raise Exception("Number of semaphores mismatch")
         f.write(s.tobytes())
-        f.write('\0' * (rounded - size))
+        end = '\0' * (rounded - size)
```
avagin, line 393: b'\0' * (rounded - size) - will do the same, will it not?

adrianreber: Yes. Changed.

adrianreber:
Looks good. Jenkins finally passed the crit test again.

Issue #1408, opened 2021-03-16
CRIU_REQ_TYPE__FEATURE_CHECK RPC request always failing


ZeyadYasser:
I was trying to call the feature check over RPC and was getting errors. I started looking into [cr-service.c](https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/cr-service.c).

It turns out that the request handler [```handle_feature_check()```](https://github.com/checkpoint-restore/criu/blob/c8beca65580da87b89964da3ffb454ca99aa6bc6/criu/cr-service.c#L1052) was failing, I think this was caused by two things:
1. ```setup_opts_from_req()``` was used and ```msg->opts``` could be NULL (```kerndat_init()``` is enough for feature checking) [here](https://github.com/checkpoint-restore/criu/blob/c8beca65580da87b89964da3ffb454ca99aa6bc6/criu/cr-service.c#L1073)
2. ```resp.success``` was always set to false at the end [here](https://github.com/checkpoint-restore/criu/blob/c8beca65580da87b89964da3ffb454ca99aa6bc6/criu/cr-service.c#L1113)

Issue #1400, opened 2021-03-11
Current sid 18028 intersects with pid (255) in images


rst0git:
Commit https://github.com/checkpoint-restore/criu/commit/77968d43c3641baa189212ee3547e932fff7f634 causes all Java tests to fail with the following error

```
(00.000228) pstree pid_max=23001
(00.000236) Migrating process tree (SID 18028->18028)
(00.000238) Error (criu/pstree.c:377): Current sid 18028 intersects with pid (255) in images
```

rst0git:
cc: @Snorch

Snorch:
Sorry for breaking tests, I missed this case where old_sid==new_sid, please see a pr #1413. It should help.

PR #1409, opened 2021-03-16
fix CRIU_REQ_TYPE__FEATURE_CHECK RPC request

ZeyadYasser:
Closes: #1408

CRIU_REQ_TYPE__FEATURE_CHECK was failing, this was caused by two things in ```handle_feature_check()```:
	1. ```setup_opts_from_req()``` was used and it could be NULL (```kerndat_init()``` is enough for feature checking)
	2. ```resp.success``` was always set to false

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -1069,8 +1069,7 @@ static int handle_feature_check(int sk, CriuReq * msg)
 	}
 
 	if (pid == 0) {
-		/* kerndat_init() is called from setup_opts_from_req() */
-		if (setup_opts_from_req(sk, msg->opts))
+		if (kerndat_init())
 			exit(1);
 
 		setproctitle("feature-check --rpc");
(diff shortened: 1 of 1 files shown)


adrianreber (approved):
Besides the long lines in the commit message this looks correct.

Especially the missing `return 0` is interesting, because it means that after each message with `resp.success = true` there was always a message with `resp.success = false`.

Because most RPC clients do not wait for additional answers after the first answer nobody really cared.

Interestingly runc has following code:
```
	req := &criurpc.CriuReq{
		Type: &t,
		// Theoretically this should not be necessary but CRIU
		// segfaults if Opts is empty.
		// Fixed in CRIU  2.12
		Opts:     rpcOpts,
		Features: criuFeat,
	}
```

I had a look at the history and it was initially correct but fixes for 3.11 broke it.

adrianreber:
I also re-triggerd the Vagrant test. It failed twice because of network related problems. Then CentOS 8 test failure is tracked in #1401 and the Java test failure is tracked in #1400.

ZeyadYasser:
Thanks for the fast feedback, I adjusted the commit message.

PR #1413, opened 2021-03-17
pstree: don't change sid/gid-s if current sid/gid is the same

Snorch:
Previously we only skipped replacing sid for shell job if root_item was
session leader, but there is other case where root_item sid is the same
as current_sid we can safely skip replacing for this case. Same applies
to gid-s.

Now after we have pid collision check we not only "can" but should skip
pid collision checks for the latter case. It is quite obvious that
there are tasks in tree with sid==current_sid if current_sid==old_sid.

Fixes: #1400
Fixes: 77968d43c ("pstree: check for pid collision before switching to
new sid/gid")

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -347,6 +347,7 @@ static int prepare_pstree_for_shell_job(pid_t pid)
 	if (!opts.shell_job)
 		return 0;
 
+	/* root_item is a session leader */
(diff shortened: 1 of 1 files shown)


adrianreber (approved):
This fixes the broken Java tests. Thanks!

PR #1414, opened 2021-03-19
ci: move CentOS 8 based test to Cirrus

adrianreber:
The kernel on GitHub Actions has a bug

https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1919472

which breaks our CI. It works on Cirrus. Let's move it there.

Maybe this is a better solution than #1401. Just run the CentOS 8 tests on a CentOS 8 kernel instead in a container on Ubuntu.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,20 +1,53 @@
-environment:
-  HOME: "/root"
-  CIRRUS_WORKING_DIR: "/tmp/criu"
-
-compute_engine_instance:
-  image_project: cirrus-images
-  image: family/docker-kvm
-  platform: linux
-  cpu: 4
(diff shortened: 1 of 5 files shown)


adrianreber:
Now that the CentOS 8 run finishes again, the CentOS 7 run also does not abort any more and now we can see that my recent crit changes are not Python 2 compatible.

@avagin How do you want to handle this? First fix CentOS 8 (this PR) and then fix the Python 2 crit errors? Or all in one PR?

adrianreber:
I added a fixup commit for Python 2 test failures on CentOS 7 and now all tests are green.

Once this is merged I can rework #1399

PR #1415, opened 2021-03-22
test: fix for cgroup00 failures on cgroup2 systems

adrianreber:
Jenkins on Fedora 31 using cgroup2 on PPC64LE kept reporting 50% of the time that it failed with:
```
  Run criu dump
  =[log]=> dump/zdtm/static/cgroup00/57/1/dump.log
  b'(00.021407) cg:     `- [name=zdtmtst] -> [/subcg00/subsubcg] [0]'
  b'(00.021411) cg:     `- [name=zdtmtst.defaultroot] -> [/] [0]'
  b'(00.021444) cg: adding cgroup /proc/self/fd/15/system.slice/sshd.service'
  b'(00.021566) cg: adding cgroup /proc/self/fd/15/subcg00/subsubcg'
  b'(00.048042) Error (criu/cgroup.c:585): cg: Unable to create the cgroup (name=zdtmtst.defaultroot) file system: Device or resource busy'
  b'(00.048083) Error (criu/cgroup.c:713): cg: failed walking /proc/self/fd/-1/ for empty cgroups: No such file or directory'
  b'(00.048089) ----------------------------------------'
  b'(00.048093) Error (criu/cr-dump.c:1402): Dump core (pid: 59) failed with -1'
  b'(00.048555) Unfreezing tasks into 1'
  b'(00.048559) \tUnseizing 57 into 1'
  b'(00.048573) \tUnseizing 58 into 1'
  b'(00.048584) \tUnseizing 59 into 1'
  b'(00.048598) Error (criu/cr-dump.c:1780): Dumping FAILED.'
```
With the changes in this commit this error is no longer reproducible.

@avagin I was watching the ppc64le cgroup00 error for some time now and was also able to reproduce it on ppc64le. This fixes it but I do not really understand what the problem was. If I did the steps which are in the cleanup script (`test/zdtm_umount_cgroups`) manually it always removed everything correctly. It never worked in the existing script.

Not sure why it did not work, but with the changes from the PR I was no longer able to reproduce above error.

### DIFF
--- a/test/zdtm_mount_cgroups
+++ b/test/zdtm_mount_cgroups
@@ -4,8 +4,6 @@
 # Error (cgroup.c:768): cg: Set 3 is not subset of 2
 # so lets create all test controllers before executing tests.
 
-cat /proc/self/cgroup | grep -q zdtmtst.defaultroot && exit
-
 tdir=`mktemp -d zdtm.XXXXXX`
 for i in "zdtmtst" "zdtmtst.defaultroot"; do
 	mount -t cgroup -o none,name=$i zdtm $tdir &&
--- a/test/zdtm_umount_cgroups
+++ b/test/zdtm_umount_cgroups
@@ -2,8 +2,6 @@
 
 # Lets delete all test controllers after executing tests.
 
-cat /proc/self/cgroup | grep -q zdtmtst.defaultroot || exit 0
-
 tdir=`mktemp -d zdtm.XXXXXX`
 for i in "zdtmtst" "zdtmtst.defaultroot"; do
 	mount -t cgroup -o none,name=$i zdtm $tdir || { rmdir $tdir; exit 1; }


mihalicyn:
We already tried to fix that https://lkml.org/lkml/2020/9/23/772
But patch wasn't approved. May be I need to try for second time :)

adrianreber:
Thanks for the pointer to LKML.

avagin:
> Not sure why it did not work, but with the changes from the PR I was no longer able to reproduce above error.

@mihalicyn do you understand why this patch helps in this case?

adrianreber:
Another, maybe helpful information. The check fails every second time.

If the test succeeds I see the following about cgroups:
```
# cat /proc/self/cgroup 
50:name=zdtmtst.defaultroot:/
48:name=zdtmtst:/
0::/user.slice/user-0.slice/session-76.scope
```
Now I run the test again and it fails and I see the following:
```
[root@ibm-p9z-16-lp1 criu]# cat /proc/self/cgroup 
48:name=zdtmtst:/
0::/user.slice/user-0.slice/session-76.scope
```

After a successful test I can do `mount -t cgroup -o none,name=zdtmtst.defaultroot zdtm /mnt/; umount /mnt` and the test also works again. Although the `zdtm_umount_cgroups` does the same it does not help if it is done by the script. Only if I do it manually.

adrianreber:
The cleanup script `zdtm_umount_cgroups` is always doing correct cleanups on x86_64. Only on ppc64le is the script running and doing the same steps but without any result.

adrianreber:
I tried it with a newer kernel, but on 5.11 CRIU does not work at all on ppc64le (#1417).

mihalicyn:
@adrianreber great catch!

I've continued your investigation and found minimal reproducer.
Two scripts:
- mount.sh
```
#!/bin/sh

tdir=`mktemp -d zdtm.XXXXXX`
for i in "zdtmtst" ; do
	mount -t cgroup -o none,name=$i zdtm $tdir &&
	# a fake group prevents destroying of a controller
	mkdir -p $tdir/holder &&
	umount -l $tdir || exit 1
done
rmdir $tdir

cat /proc/self/cgroup
cat /proc/self/mountinfo | grep cgroup
```

- umount.sh
```
#!/bin/sh

tdir=`mktemp -d zdtm.XXXXXX`
for i in "zdtmtst" ; do
	mount -t cgroup -o none,name=$i zdtm $tdir || { rmdir $tdir; exit 1; }
	# remove a fake group if exists
	if [ -d "$tdir/holder" ]; then
		rmdir $tdir/holder || { umount -l $tdir && rmdir $tdir; exit 1; }
	fi

	#sleep 1
	
	umount -l $tdir || exit 1;
	#umount $tdir || exit 1;
done
rmdir $tdir

cat /proc/self/cgroup
cat /proc/self/mountinfo | grep cgroup
```

Important things:
- as you can see I've removed second cgroup zdtmtst.defaultroot.
- I've experimented with umount command (tried umount without lazy) - it makes no sense
- I've added `sleep 1` just before `umount -l ...` and this totally changes the picture.
- our ppc64le test node is 1 CPU VM (looks like that circumstance may affect)

Reproducer (scripts as it is above. without sleep):
```
[root@criu mikhalitsyn]# ./mount 
114:name=zdtmtst:/
0::/system.slice/sshd.service
28 21 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel
[root@criu mikhalitsyn]# cat /proc/self/cgroup 
114:name=zdtmtst:/
0::/system.slice/sshd.service
=============== first call =======================
[root@criu mikhalitsyn]# ./umount 
114:name=zdtmtst:/
0::/system.slice/sshd.service
28 21 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel
[root@criu mikhalitsyn]# cat /proc/self/cgroup 
114:name=zdtmtst:/
0::/system.slice/sshd.service
=============== second call =======================
[root@criu mikhalitsyn]# ./umount 
114:name=zdtmtst:/
0::/system.slice/sshd.service
28 21 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel
[root@criu mikhalitsyn]# cat /proc/self/cgroup 
0::/system.slice/sshd.service
```

It is worth to notice two things here:
1. second umount really destroys cgroups
2. in umount script (at the end) we have `cat /proc/self/cgroup` but as you can see
after second invocation `./umount` at the end of the script we can see that `zdtmtst` cgroup is shown. But just after that
`cat /proc/self/cgroup` shows nothing! It means that we deal with some latency (workqueues intensively used in cgroups destroy code).

If we change `umount` script and uncomment `sleep 1` then we get:
```
[root@criu mikhalitsyn]# ./mount 
115:name=zdtmtst:/
0::/system.slice/sshd.service
28 21 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel
[root@criu mikhalitsyn]# cat /proc/self/cgroup 
115:name=zdtmtst:/
0::/system.slice/sshd.service
=============== first call =======================
[root@criu mikhalitsyn]# ./umount 
115:name=zdtmtst:/
0::/system.slice/sshd.service
28 21 0:25 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:4 - cgroup2 cgroup2 rw,seclabel
[root@criu mikhalitsyn]# cat /proc/self/cgroup 
0::/system.slice/sshd.service
```

`sleep 1` completely solves problem. That's looks like magic :)
I've started to trace that thing and what I've found...

What was traced and how
```
echo path_mount > /sys/kernel/debug/tracing/set_graph_function
echo path_umount >> /sys/kernel/debug/tracing/set_graph_function
echo cgroup_rmdir >> /sys/kernel/debug/tracing/set_graph_function
echo css_killed_work_fn >> /sys/kernel/debug/tracing/set_graph_function
echo css_release_work_fn >> /sys/kernel/debug/tracing/set_graph_function
echo css_release >> /sys/kernel/debug/tracing/set_graph_function
echo css_free_rwork_fn >> /sys/kernel/debug/tracing/set_graph_function

echo function_graph > /sys/kernel/debug/tracing/current_tracer
echo funcgraph-tail > /sys/kernel/debug/tracing/trace_options
echo funcgraph-proc > /sys/kernel/debug/tracing/trace_options
echo 1 > /sys/kernel/debug/tracing/tracing_on
```

I'd decided to compare cases with/without sleep. And trace each `./umount` invocation separately.
In case with `sleep 1` we have only one invocation. So, we have 2 (without sleep) + 1 (with sleep) traces.

[trace_sleep.txt](https://github.com/checkpoint-restore/criu/files/6217250/trace_sleep.txt)
[trace_nosleep2.txt](https://github.com/checkpoint-restore/criu/files/6217251/trace_nosleep2.txt)
[trace_nosleep1.txt](https://github.com/checkpoint-restore/criu/files/6217253/trace_nosleep1.txt)

Interesting place for us is function `css_release` which called when percpu refcounter `cgrp->self.refcnt` is 0.
Looks like when we do `mount -> rmdir -> umount` without sleep in between some reference is not dropped.
But if we do `mount -> rmdir -> sleep -> umount` then it is dropped and cgroup is sucessfuly destroys.
Looks like I need to continue investigation here to fully understand why it may happen and why repeating `mount -> rmdir -> umount` scheme helps even without any sleeps in between.

Snorch:
Note that after some tests we get not only /holder but also /bar and /test subdirs of zdtmtst cgroup which might be related to an initial problem.

```
[root@snorch criu]# ./test/zdtm.py run -t zdtm/static/cgroup_stray --ignore-taint
The kernel is tainted: '512\n'
Checking feature cgroupns
cgroupns is supported
userns is supported
=== Run 1/1 ================ zdtm/static/cgroup_stray
timens isn't supported on 5.10.21-200.snorch.fc33.x86_64
====================== Run zdtm/static/cgroup_stray in h =======================
 DEP       sk-unix01.d
 DEP       sk-unix-unconn.d
Start test
Test is SUID
./cgroup_stray --pidfile=cgroup_stray.pid --outfile=cgroup_stray.out --dirname=cgroup_stray.test
Run criu dump
Run criu restore
Send the 15 signal to  81
Wait for zdtm/static/cgroup_stray(81) to die for 0.100000
Removing dump/zdtm/static/cgroup_stray/81
====================== Test zdtm/static/cgroup_stray PASS ======================
====================== Run zdtm/static/cgroup_stray in ns ======================
Start test
Test is SUID
./cgroup_stray --pidfile=cgroup_stray.pid --outfile=cgroup_stray.out --dirname=cgroup_stray.test
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/cgroup_stray/145/1/restore.log
b'(00.047996) Unlock network'
b'(00.048000) Running network-unlock scripts'
b'iptables-restore v1.8.5 (nf_tables):'
b'line 6: CHAIN_USER_DEL failed (Device or resource busy): chain CRIU'
b'(00.060046) Error (criu/util.c:643): exited, status=4'
b'ip6tables-restore v1.8.5 (nf_tables):'
b'line 6: CHAIN_USER_DEL failed (Device or resource busy): chain CRIU'
b'(00.072005) Error (criu/util.c:643): exited, status=4'
Send the 15 signal to  166
Wait for zdtm/static/cgroup_stray(166) to die for 0.100000
Removing dump/zdtm/static/cgroup_stray/145
====================== Test zdtm/static/cgroup_stray PASS ======================
[root@snorch criu]# cat /proc/self/cgroup
20:name=zdtmtst:/
0::/user.slice/user-1000.slice/user@1000.service/vte-spawn-3ded8e2e-a06e-4a32-8e07-47350643f52d.scope
[root@snorch criu]# mount -t cgroup -o none,name=zdtmtst zdtm /mnt
[root@snorch criu]# ls /mnt
bar  cgroup.clone_children  cgroup.procs  cgroup.sane_behavior  notify_on_release  release_agent  tasks
[root@snorch criu]# 
```

Also that is sad that simple umount is not enough and cgroup hierarchy is not removed imediately. Probably simple solution would be to retry umount several times until /proc/self/cgroup is clean.

adrianreber:
> upd: I mean, probably adding `rmdir $tdir/*` would fix the original problem.

@mihalicyn Can you check if this works on the ppc64le machine you have access to?

Snorch:
>>upd: I mean, probably adding rmdir $tdir/* would fix the original problem.
>
>@mihalicyn Can you check if this works on the ppc64le machine you have access to?

In personal talk @mihalicyn  already convinced me that at least the problem detected by his reproducers https://github.com/checkpoint-restore/criu/pull/1415#issuecomment-808887888 is not as simple as I expected, looks like even without extra holders there is some race on cgroup removal...

adrianreber:
> > > upd: I mean, probably adding rmdir $tdir/* would fix the original problem.
> > 
> > 
> > @mihalicyn Can you check if this works on the ppc64le machine you have access to?
> 
> In personal talk @mihalicyn already convinced me that at least the problem detected by his reproducers [#1415 (comment)](https://github.com/checkpoint-restore/criu/pull/1415#issuecomment-808887888) is not as simple as I expected, looks like even without extra holders there is some race on cgroup removal...

Good to know. So, we can probably just close this. Makes no sense to keep it open if it does not solve the problem.

mihalicyn:
> @mihalicyn Can you check if this works on the ppc64le machine you have access to?

No, it doesn't fix the problem.

Ok, let's close, I will continue with investigation and will write results into issue.

adrianreber:
Thanks.

Issue #1426, opened 2021-03-30
Alpine tests failing with `Error (criu/image.c:605): Can't open parent path: No such file or directory"`


adrianreber:
Since #1403 is merged a  lot of Alpine based tests are failing with:
```
2021-03-29T15:34:14.7978347Z ========================== Run zdtm/static/aio00 in h ==========================
2021-03-29T15:34:14.7978781Z Start test
2021-03-29T15:34:14.7979451Z ./aio00 --pidfile=aio00.pid --outfile=aio00.out
2021-03-29T15:34:14.7979945Z Run criu dump
2021-03-29T15:34:14.7980409Z =[log]=> dump/zdtm/static/aio00/52/1/dump.log
2021-03-29T15:34:14.7981163Z ------------------------ grep Error ------------------------
2021-03-29T15:34:14.7981919Z b'(00.000113) Running pre-dump scripts'
2021-03-29T15:34:14.7982650Z b'(00.000145) irmap: Searching irmap cache in work dir'
2021-03-29T15:34:14.7983340Z b'(00.000159) No irmap-cache image'
2021-03-29T15:34:14.7984075Z b'(00.000162) irmap: Searching irmap cache in parent'
2021-03-29T15:34:14.7985061Z b"(00.000167) Error (criu/image.c:605): Can't open parent path: No such file or directory"
2021-03-29T15:34:14.7985834Z b'(00.000171) Unlock network'
2021-03-29T15:34:14.7986445Z b'(00.000181) Unfreezing tasks into 1'
2021-03-29T15:34:14.7987082Z b'(00.000183) \tUnseizing 52 into 1'
2021-03-29T15:34:14.7987917Z b'(00.000187) Error (compel/src/lib/infect.c:346): Unable to detach from 52: No such process'
2021-03-29T15:34:14.7988804Z b'(00.000196) Error (criu/cr-dump.c:1780): Dumping FAILED.'
2021-03-29T15:34:14.7989585Z ------------------------ ERROR OVER ------------------------
2021-03-29T15:34:14.7990215Z ################### Test zdtm/static/aio00 FAIL at CRIU dump ###################
```

@minhbq-99 Can you have a look at what is wrong: https://github.com/checkpoint-restore/criu/runs/2219865399

minhbq-99:
Hmm, this is because this check in musl
https://git.musl-libc.org/cgit/musl/tree/src/unistd/faccessat.c#n33
[...]

PR #1428, opened 2021-03-30
criu: Replace faccessat with fstatat when using AT_SYMLINK_NOFOLLOW flag

minhbq-99:
Currently, alpine musl libc library returns Invalid argument error (EINVAL)
when calling faccessat with AT_SYMLINK_NOFOLLOW flag.

Fix this by using fstatat instead.

Fixes: #1426.

### DIFF
--- a/criu/image.c
+++ b/criu/image.c
@@ -593,9 +593,11 @@ void close_image_dir(void)
 
 int open_parent(int dfd, int *pfd)
 {
+	struct stat st;
+
 	*pfd = -1;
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks for the quick fix. Alpine CI still failed but because of another error concerning TLS. I restarted the Alpine run. The TLS errors usually goes away after a while.

minhbq-99:
Yes, it works now. Thank you :)

Issue #1416, opened 2021-03-24
netns-nft test fails if nft tables exist


adrianreber:
On Cirrus CI and Jenkins where tests are running on native CentOS 8 systems the `netns-nft` fails with:
```
======================= Run zdtm/static/netns-nft in uns =======================
Start test
Test is SUID
./netns-nft --pidfile=netns-nft.pid --outfile=netns-nft.out --filename=netns-nft.test
Run criu dump
Run criu lazy-pages
Run criu restore
Send the 15 signal to  104
Wait for zdtm/static/netns-nft(104) to die for 0.100000
############### Test zdtm/static/netns-nft FAIL at result check ################
Test output: ================================
Error: No such file or directory
list table inet netns-nft-zdtm-test
                ^^^^^^^^^^^^^^^^^^^
04:01:35.029:     4: FAIL: netns-nft.c:51: Can't get nft table (errno = 11 (Resource temporarily unavailable))

 <<< ================================
```

The CentOS 8 systems I have seen actually have nft tables configured, maybe that is something which the test does not expect.

For the Cirrus CI CentOS 8 test I have been doing `mv /usr/sbin/nft /usr/sbin/nft.away` to skip the test, but a real fix would be nice.

@mihalicyn any ideas how to fix that?

mihalicyn:
Please, assign me. The problem is that in recent distributions of CentOS there is no nftables statically linked library.
They lives us with no choice - we have to use dynamically linked libraries.

mihalicyn:
I've submitted an issue about CentOS 8 nftables package.

https://bugs.centos.org/view.php?id=18151

On my Fedora 33 CRIU builds fine with dynamic nftables library and `netns-nft` test passes.

adrianreber:
So test would need to check if CRIU is built with nft support or not. If nft support is not included in CRIU the test should be skipped.

Maybe a feature check that returns false if not linked against nftables?

mihalicyn:
@adrianreber yeah.

We can use something like `ldd ../criu/criu | grep nftables` in checkskip file for `netns-nft` test, but this will only work for the dynamically linked way. If the nftables library linked statically then we have no fully reliable way to detect that.
I think it will be okay just to check `pkg-config --libs libnftables`. Because in our `Makefile` we detecting the nftables library in user system using `pkg-config`.

adrianreber:
> @adrianreber yeah.
> 
[...]

PR #1437, opened 2021-04-11
zdtm: netns-nft test checkskip fix

mihalicyn:
Let's check the nftables support basing on `pkg-config`.
On CentOS 8 we meet with the problem when the nft binary is
present, but there is no pkgconfig file for the nftables library.

See #1416

### DIFF
--- a/test/zdtm/static/netns-nft.checkskip
+++ b/test/zdtm/static/netns-nft.checkskip
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-test -f /usr/sbin/nft || exit 1
+pkg-config --libs libnftables


mihalicyn:
That's RFC patch. @adrianreber proposes a different way - add a separate feature in CRIU to check that the nftables support is compiled into CRIU. Let's decide and I will prepare patches.

test/zdtm/static/netns-nft.checkskip
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-test -f /usr/sbin/nft || exit 1
+pkg-config --libs libnftables
```
rst0git, line 3: Would the following workaround become redundant after this change?
https://github.com/checkpoint-restore/criu/blob/e49ce513c8df43e8d3b30b7acfbe7bd496d89862/.cirrus.yml#L49

mihalicyn: @rst0git yes

mihalicyn:
closing due to https://github.com/checkpoint-restore/criu/pull/1438

PR #1438, opened 2021-04-11
skip iptables dump if it has nft backend and nft dump is supported

### LINKED ISSUE #1435: nft support conflicts with network locking
I was looking at an interesting problem on the latest Fedora version.

It seems Fedora also switched to the nft backend of iptables and network unlocking no longer works:
```
b'(00.055529) Unlock network'
b'(00.055533) Running network-unlock scripts'
b'iptables-restore v1.8.7 (nf_tables):'
b'line 7: CHAIN_USER_DEL failed (Device or resource busy): chain CRIU'
b'(00.065053) Error (criu/util.c:643): exited, status=4'
b'ip6tables-restore v1.8.7 (nf_tables):'
b'line 7: CHAIN_USER_DEL failed (Device or resource busy): chain CRIU'
b'(00.074787) Error (criu/util.c:643): exited, status=4'
```

The reason for this error is that the locking rules exist twice:
```
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         
CRIU       all  --  anywhere             anywhere            
CRIU       all  --  anywhere             anywhere            

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         
CRIU       all  --  anywhere             anywhere            
CRIU       all  --  anywhere             anywhere            

Chain CRIU (4 references)
target     prot opt source               destination         
ACCEPT     all  --  anywhere             anywhere             mark match 0xc114
DROP       all  --  anywhere             anywhere            
ACCEPT     all  --  anywhere             anywhere            
DROP       all  --  anywhere             anywhere 
```

From what I understand the following happens:

1. CRIU locks the network using iptables-restore (nft backend)
2. CRIU collects nft rules (which includes the network locking)
3. CRIU collects iptables rules (which includes the network locking)
4. CRIU restores the process with nft and iptables rules and now we have the network locked twice
5. CRIU unlocks the network only once (iptables-restore) and that fails

This is the content of the nft and iptables images:
```
$ cat test/dump/zdtm/static/socket-tcp6-closing/63/1/nftables-10.img
table ip filter {
	chain CRIU {
		mark 0xc114 counter packets 0 bytes 0 accept
		counter packets 0 bytes 0 drop
	}

	chain INPUT {
		type filter hook input priority filter; policy accept;
		counter packets 0 bytes 0 jump CRIU
	}

	chain OUTPUT {
		type filter hook output priority filter; policy accept;
		counter packets 0 bytes 0 jump CRIU
	}
}
table ip6 filter {
	chain CRIU {
		mark 0xc114 counter packets 0 bytes 0 accept
		counter packets 0 bytes 0 drop
	}

	chain INPUT {
		type filter hook input priority filter; policy accept;
		counter packets 0 bytes 0 jump CRIU
	}

	chain OUTPUT {
		type filter hook output priority filter; policy accept;
		counter packets 0 bytes 0 jump CRIU
	}
}
$ cat test/dump/zdtm/static/socket-tcp6-closing/63/1/iptables-10.img
# Generated by iptables-save v1.8.7 on Sat Apr 10 16:22:27 2021
*filter
:INPUT ACCEPT [0:0]
:FORWARD ACCEPT [0:0]
:OUTPUT ACCEPT [0:0]
:CRIU - [0:0]
-A INPUT -j CRIU
-A OUTPUT -j CRIU
-A CRIU -m mark --mark 0xc114 -j ACCEPT
-A CRIU -j DROP
COMMIT
# Completed on Sat Apr 10 16:22:27 2021
```

Somehow CRIU should only remember the network locking only once. Either via nft or via iptables-restore.

CC: @mihalicyn I think you were involved in the nft code, maybe you have an idea how solve this

mihalicyn:
Should fix #1435

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -1494,3 +1494,91 @@ int cut_path_ending(char *path, char *ending)
 	path[ending_pos - 1] = 0;
 	return 0;
 }
+
+static int is_iptables_nft(char *bin)
+{
+	int pfd[2] = { -1, -1 }, ret = -1;
+	char *cmd[] = { bin, "-V", NULL };
+	char buf[100];
+
+	if (pipe(pfd) < 0) {
+		pr_perror("Unable to create pipe");
+		goto err;
+	}
+
+	ret = cr_system(-1, pfd[1], -1, cmd[0], cmd, 0);
+	if (ret) {
+		pr_err("%s -V failed\n", cmd[0]);
+		goto err;
+	}
+
+	close_safe(&pfd[1]);
+
+	ret = read(pfd[0], buf, sizeof(buf) - 1);
+	if (ret < 0) {
+		pr_perror("Unable to read %s -V output", cmd[0]);
+		goto err;
+	}
+
+	buf[ret] = '\0';
+	ret = 0;
+
+	if (strstr(buf, "nf_tables")) {
+		pr_info("iptables has nft backend: %s\n", buf);
+		ret = 1;
+	}
+
+err:
+	close_safe(&pfd[1]);
+	close_safe(&pfd[0]);
+	return ret;
+}
+
+char *get_legacy_iptables_bin(bool ipv6)
+{
+	static char iptables_bin[2][32];
+	/* 0  - means we don't know yet,
+	 * -1 - not present,
+	 * 1  - present.
+	 */
+	static int iptables_present[2] = { 0, 0 };
+	char bins[2][2][32] = {
+		{
+			"iptables-save",
+			"iptables-legacy-save"
+		},
+		{
+			"ip6tables-save",
+			"ip6tables-legacy-save"
+		}
+	};
+	int ret;
+
+	if (iptables_present[ipv6] == -1)
+		return NULL;
+
+	if (iptables_present[ipv6] == 1)
+		return iptables_bin[ipv6];
+
+	memcpy(iptables_bin[ipv6], bins[ipv6][0], strlen(bins[ipv6][0]) + 1);
+	ret = is_iptables_nft(iptables_bin[ipv6]);
+
+	/*
+	 * iptables on host uses nft backend (or not installed),
+	 * let's try iptables-legacy
+	 */
+	if (ret < 0 || ret == 1) {
+		memcpy(iptables_bin[ipv6], bins[ipv6][1],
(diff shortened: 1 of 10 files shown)


adrianreber:
Thanks for providing a fix. Maybe it would be better to run the check for the iptables backend only once and store the value in the kerndat cache. This way we would not need to run `iptables-save` for each dump just to get the version. One less `system()`.

This would not work for users who replace iptables with iptables-nft without rebooting, but I think this is pretty unlikely. I expect that a change of the iptables backend usually only happens when upgrading the Linux distribution from one version to another which usually comes with a reboot.

mihalicyn:
@adrianreber thanks!

Yes, we can do that. Also I think it would be nice to try use `iptables-legacy` if it's present in the user system.
If we consider a situation when user dump LXC container (in LXC container both iptables and nftables rules may be set),
we should properly save both rulesets. But if host iptables will have nftables backend then we just skip it. I think we need to try `iptables-legacy` to dump iptables rules, and use nft to dump nftables rules.

adrianreber:
This looks correct. Can you remove the hack `mv /usr/sbin/nft /usr/sbin/nft.away` from `.cirrus.yml`. That should show that it works correctly, because it will use the feature check you added.

mihalicyn:
@adrianreber thank you! Sure, I've added commit with that.

rst0git (approved):

mihalicyn:
I've rebased PR on top of new criu-dev

avagin:
> net: skip iptables dump if it has nft backend and nft dump is supported 

pls write a commit message for this commit.

criu/util.c
```diff
@@ -1493,4 +1493,90 @@ int cut_path_ending(char *path, char *ending)
...
+		pr_perror("Unable to create pipe");
+		goto err;
+	}
+
+	ret = cr_system(-1, pfd[1], -1, cmd[0], cmd, 0);
+	if (ret) {
+		pr_err("%s -V failed\n", cmd[0]);
+		goto err;
+	}
+
```
avagin, line 1514: you need to close pfd[1] here, otherwise, the next call can block forever.

mihalicyn: thanks, fixed

PR #1443, opened 2021-04-13
tiny fix on dump_empty_fs

liusdu:
return value of is_empty_dir:
* -1 : open directory stream failed
*  0 : directory is not empty
*  1 : directory is empty

### DIFF
--- a/criu/filesystems.c
+++ b/criu/filesystems.c
@@ -656,12 +656,12 @@ static int dump_empty_fs(struct mount_info *pm)
 		return fd;
 
 	ret = is_empty_dir(fd);
-	if (ret < 0) {
+	if (ret == 0) {
 		pr_err("%s isn't empty\n", pm->fstype->name);
 		return -1;
 	}
 
-	return ret ? 0 : -1;
+	return ret == 1 ? 0 : -1;
 }
 
 /*
--- a/criu/util.c
+++ b/criu/util.c
@@ -731,7 +731,13 @@ int is_root_user(void)
 /*
  * is_empty_dir will always close the FD dirfd: either implicitly
  * via closedir or explicitly in case fdopendir had failed
+ *
+ * return values:
+ *   < 0 : open directory stream failed
+ *     0 : directory is not empty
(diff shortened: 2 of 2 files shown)


criu/filesystems.c
```diff
@@ -657,6 +657,9 @@ static int dump_empty_fs(struct mount_info *pm)
 
 	ret = is_empty_dir(fd);
 	if (ret < 0) {
+		pr_perror("open mountpount: %s failed",pm->mountpoint);
```
avagin, line 660: this has to be pr_err

liusdu: @avagin  done, please take a look

criu/util.c
```diff
@@ -740,6 +740,7 @@ int is_empty_dir(int dirfd)
 
 	fdir = fdopendir(dirfd);
 	if (!fdir) {
+		 pr_perror("open directory stream by fd %d failed", dirfd);
```
avagin, line 743: pls remove an extra space after two tabs.

liusdu: sorry for this mistake.. fixed

criu/util.c
```diff
@@ -740,6 +740,7 @@ int is_empty_dir(int dirfd)
 
 	fdir = fdopendir(dirfd);
 	if (!fdir) {
+		pr_perror("open directory stream by fd %d failed", dirfd);
 		close_safe(&dirfd);
 		return -1;
```
rst0git, line 745: Could you please add the return values description in the comment above `is_empty_dir()` please?
Something like:
```C
/*
 * is_empty_dir will always close the FD dirfd: either implicitly
 * via closedir or explicitly in case fdopendir had failed
 * 
 * return value:
 *  < 0 : open directory stream failed
 *    0 : directory is not empty
 *    1 : directory is empty
 */
int is_empty_dir(int dirfd)
```

liusdu: @rst0git Added~

avagin:
Applied. Thanks a lot.

PR #1450, opened 2021-04-23
kerndat: ensure that the CRIU process is dumpable

nviennot:
When running the CRIU binary is secure mode (because setuid, or setcap)
the kernel sets the DUMPABLE property to 0. This in turn makes
/proc/self/* root owned, preventing some of the kerndat logic to
function properly.

This change ensures that the CRIU process is dumpable.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -417,7 +417,7 @@ static bool kerndat_has_memfd_create(void)
 	else if (ret == -1 && errno == EFAULT)
 		kdat.has_memfd = true;
 	else {
-		pr_err("Unexpected error from memfd_create(NULL, 0): %m\n");
+		pr_perror("Unexpected error from memfd_create(NULL, 0)");
 		return -1;
 	}
 
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1045,7 +1045,7 @@ static bool kerndat_has_clone3_set_tid(void)
 	if (pid == -1 && errno == EINVAL) {
 		kdat.has_clone3_set_tid = true;
 	} else {
-		pr_perror("Unexpected error from clone3\n");
+		pr_perror("Unexpected error from clone3");
 		return -1;
 	}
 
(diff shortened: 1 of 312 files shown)


criu/kerndat.c
```diff
@@ -1067,6 +1075,14 @@ int kerndat_init(void)
...
 	preload_netfilter_modules();
 
+	/*
+	 * When our process is running in secure mode (e.g., with
+	 * setuid/setcap CRIU binary), our process becomes non-dumpable, which
+	 * makes our /proc/self root owned, preventing some of the kerndat
+	 * logic to function properly.
+	 */
+	pr_set_dumpable(1);
+
```
adrianreber, line 1085: I need this for my non-root PR, but calling it in `kerndat_init()` feels wrong. Ah, but according to the comment you only need it for kerndat checks which then makes sense. For non-root I need for multiple things. I can then move it somewhere else. Still, calling it unconditionally is strange, but it probably does not hurt

nviennot: Why do you think doing it unconditionally is strange?

Where should we put this call? around https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/crtools.c#L216 ?

adrianreber: > Why do you think doing it unconditionally is strange?

Until now we didn't do it all and now always. Just strange to always do it, if it is not needed.

> Where should we put this call? around https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/crtools.c#L216 ?

In my PR I do it at https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/crtools.c#L126 in the function where I check for capabilities. But in the end you can put it wherever it makes sense for your problem and I will just move it.

I guess the main question is how long do we need for the non-root PR to be merged and maybe that solves your problem.

nviennot: Let's leave it to the non-root PR. I'll close this PR.

avagin: The non-root PR is stuck and I would like to split it into smaller parts.

@adrianreber I don't understand why kerndat_init() is a bad place for this call? pr_set_dumpable is lightweight.
@nviennot if we set the "dumpable" flag to 1, it means that an user that executes criu will be able to attach to this process with ptrace. This looks like a security issue, because CRIU is running with extra capabilities and the user will be able to run arbitrary code in its context. Have I missed something?

PR #1451, opened 2021-04-23
logging: pr_perror() -> pr_msg() when execvp fails in action scripts and others

nviennot:
When invoking an action-script, all file descriptors >= 3 are closed.
If execvp() fails, we can only log the error on stderr. pr_msg() outputs
on stderr, so we use this as opposed to pr_perror().

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -626,7 +626,9 @@ int cr_system_userns(int in, int out, int err, char *cmd,
 
 		execvp(cmd, argv);
 
-		pr_perror("exec(%s, ...) failed", cmd);
+		/* We can't use pr_error() as log file fd is closed. */
+		fprintf(stderr, "Error (%s:%d): " LOG_PREFIX "execvp(\"%s\", ...) failed: %s\n",
+		       __FILE__, __LINE__, cmd, strerror(errno));
 out_chld:
 		_exit(1);
 	}


criu/util.c
```diff
@@ -626,7 +626,12 @@ int cr_system_userns(int in, int out, int err, char *cmd,
 
 		execvp(cmd, argv);
 
-		pr_perror("exec(%s, ...) failed", cmd);
+		/*
+		 * pr_error() doesn't work as the log file fd is closed.
+		 * Luckily, pr_msg() outputs on stderr, which is not closed.
+		 */
+		pr_msg("Error (%s:%d): " LOG_PREFIX "execvp(\"%s\", ...) failed: %s\n",
+		       __FILE__, __LINE__, cmd, strerror(errno));
```
avagin, line 634: pr_msg prints to stdout. I think we can use dprintk(STDERR_FILENO, ....).

rst0git: What about something like the following?
```diff
--- a/criu/include/util.h
+++ b/criu/include/util.h
@@ -383,4 +383,6 @@ static inline void print_stack_trace(pid_t pid) {}
 
 extern int mount_detached_fs(const char *fsname);
 
+extern int is_valid_fd(int fd);
+
 #endif /* __CR_UTIL_H__ */
diff --git a/criu/log.c b/criu/log.c
index 439a899f0..cb0ecb271 100644
--- a/criu/log.c
+++ b/criu/log.c
@@ -81,7 +81,10 @@ int log_get_fd(void)
 {
        int fd = get_service_fd(LOG_FD_OFF);
 
-       return fd < 0 ? DEFAULT_LOGFD : fd;
+       if (fd < 0 || !is_valid_fd(fd))
+               return DEFAULT_LOGFD;
+
+       return fd;
 }
 
 void log_get_logstart(struct timeval *s)
diff --git a/criu/util.c b/criu/util.c
index ca09f02a1..50489f297 100644
--- a/criu/util.c
+++ b/criu/util.c
@@ -1493,4 +1493,9 @@ int cut_path_ending(char *path, char *ending)
 
        path[ending_pos - 1] = 0;
        return 0;
+}
+
+int is_valid_fd(int fd)
+{
+       return fcntl(fd, F_GETFL) != -1 || errno != EBADF;
 }
```

avagin: It is not obvious behavior and I would avoid this sort of hacks.  Even if a file descriptor is valid, we can't be sure that it is the right one. But in the code, we always understand where to find the logging file descriptor.

nviennot: Using fprintf(stderr, ...). I confirm that it works.

rst0git (approved):

PR #1453, opened 2021-04-23
crit: fix bytes vs str issues with unix sock names

nviennot:
Improve the utf8 handling for unix socket

### DIFF
--- a/lib/py/images/pb2dict.py
+++ b/lib/py/images/pb2dict.py
@@ -3,6 +3,7 @@ import collections
 import os
 import quopri
 import socket
+import sys
 from ipaddress import IPv4Address, IPv6Address, ip_address
 
 from google.protobuf.descriptor import FieldDescriptor as FD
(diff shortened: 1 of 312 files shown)


adrianreber:
Is there a certain image that needs this fix. Would be nice to have it included in our CI. Especially with the parallel Python 2 and Python 3 support it would be good to have a test for this change.

At https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/ci/run-ci-tests.sh#L270 is a good place to add additional images to the recode test which triggered a lot of the other recent changes to the Python code.

lib/py/images/pb2dict.py
```diff
@@ -261,11 +261,11 @@ def decode_base64(value):
 
 
 def encode_unix(value):
-    return quopri.encodestring(value)
+    return quopri.encodestring(value).decode('utf8')
 
 
 def decode_unix(value):
-    return quopri.decodestring(value)
+    return quopri.decodestring(value.encode('utf'))
```
Snorch, line 268: Probably: s/utf/utf8/

Snorch (commented):
Hi, @nviennot 

Your commit message lacks details:

```
 crit: fix bytes vs str issues with unix sock names

Improve the utf8 handling for unix socket
```

You can put information about which exact errors you get into commit message, this would help people which would face same issues with older criu version to understand that they need your patch. Also it would help review the patch.

Also it would be nice to have a description of how you fix the problem, so that we can understand that what you want to do corresponds to what you actually do.

Please take a look on https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes our contribution guide.

nviennot:
You are right, I have to revisit this. We probably would benefit from test cases as well.

adrianreber:
@nviennot Do you want to rebase this?

PR #1454, opened 2021-04-23
pstree: when computing pid_max, disregard pgid and sid

nviennot:
computing pid_max is useful when creating new processes (they have pids
we want to control). pgid and sid are related, but not relevent. In
fact, if the sid is outside of our application tree, then we should not
care. This is the case when using --shell-job. pgid/sid is should not
influence the adjustment of /proc/sys/kernel/pid_max

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -557,11 +557,7 @@ static int read_one_pstree_item(struct cr_img *img, pid_t *pid_max)
 	if (e->pid > *pid_max)
 		*pid_max = e->pid;
 	pi->pgid = e->pgid;
-	if (e->pgid > *pid_max)
-		*pid_max = e->pgid;
 	pi->sid = e->sid;
-	if (e->sid > *pid_max)
-		*pid_max = e->sid;
 	pi->pid->state = TASK_ALIVE;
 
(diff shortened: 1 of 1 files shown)


Snorch:
@avagin Thanks for linking this to me.

Please sorry for a long response...

> if the sid is outside of our application tree, then we should not care

This is wrong. Processes with dead session leaders is quite a common thing to have in containers, one can do setsid(), fork some processes and die. CRIU when recreating this kind of process tree should recreate sids as it was before and would need to restore this session by creating the helper process with pid == needed sid, having wrong pid_max at this point would break restore.

Nacked-by: Pavel Tikhomirov <ptikhomirov@virtuozzo.com>

nviennot:
@Snorch you are right! I didn't know.
I think this is only relevant when using `--shell-job` in certain circumstances (e.g., https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/pstree.c#L365 where the sid from the checkpoint is ignored and instead inherited from the restore environment), but I'm not sure how to write the code properly.

I'll close the PR for now, it's really minor, and perhaps not worth the effort of chasing a minimal pid_max computation

PR #1455, opened 2021-04-23
tty: Only store the stdin fd when it corresponds to a tty

nviennot:
Doing otherwise can lead to problems when using inherit-fd on stdin as
it gets closed.


criu/tty.c
```diff
@@ -2342,13 +2342,14 @@ int tty_prep_fds(void)
 
 	if (!isatty(STDIN_FILENO))
 		pr_info("Standard stream is not a terminal, may fail later\n");
-	else
+	else {
```
rst0git, line 2345: Could you please add braces to both branches of a conditional statement?

rst0git (approved):

Snorch (commented):
I like the code. I think that there is no point to put an fd into fdstore if it would not be used from fdstore, as self_stdin_fdid is currently only used if stdin_isatty == true.

>Doing otherwise can lead to problems when using inherit-fd on stdin as
it gets closed.

But, can you please explain how adding the fd to fdstore can results in the original fd been closed?

nviennot:
When using `--inherit-fd fd[0]:XXX` (for example, to replace resource XXX with the file provided via stdin), the following gets invoked: https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/files.c#L1599
This sends stdin into the fdstore, and closes stdin. From that point, fd=0 is now closed, and the tty code tried to use it again.

Snorch:
>When using --inherit-fd fd[0]:XXX (for example, to replace resource XXX with the file provided via stdin), the following gets invoked: https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/files.c#L1599
This sends stdin into the fdstore, and closes stdin. From that point, fd=0 is now closed, and the tty code tried to use it again.

Probably we can instead don't allow --inherit-fd on 0-2 standard io fds?

rst0git:
> Probably we can instead don't allow --inherit-fd on 0-2 standard io fds?

I remember having a similar discussion in https://github.com/checkpoint-restore/criu/pull/681

Snorch:
> > Probably we can instead don't allow --inherit-fd on 0-2 standard io fds?
> 
> I remember having a similar discussion in #681

Nice note. Seems like the problem of reusing 0-2 fds was already there with opts.ps_socket, and we've had "keep_fd" logic to protect from it, but we've removed it, relying that opts.ps_socket just would not use 0-2.

Snorch:
@avagin I reworked it:
- removed stdin_isatty
- added patch to prohibit --inherit-fd on std fds

Documentation/criu.txt
```diff
@@ -474,6 +474,7 @@ Restores previously checkpointed processes.
     file descriptor 'N' for restoring a file identified by 'resource'.
     This option can be used to restore an external resource dumped
     with the help of *--external* *file*, *tty*, *pid* and *unix* options.
+    Important: std fds 0-2 are non allowed.
```
rst0git, line 477:
```Suggestion
    Important: std fds 0-2 are not allowed.
```

rst0git (approved):
LGTM

Issue #1402, opened 2021-03-12
criu/lsm.c:90: Invalid selinux context kernel


Snorch:
I can see all tests failing on my system (f33) like this:

```
[root@snorch criu]# ./test/zdtm.py run -t zdtm/static/env00 --ignore-taint
userns is supported
=== Run 1/1 ================ zdtm/static/env00
timens isn't supported on 5.10.19-200.snorch.fc33.x86_64
========================== Run zdtm/static/env00 in h ==========================
Start test
./env00 --pidfile=env00.pid --outfile=env00.out --envname=ENV_00_TEST
Run criu dump
=[log]=> dump/zdtm/static/env00/56/1/dump.log
------------------------ grep Error ------------------------
b'pie: 56: __fetched msg: 64 0 0'
b'pie: 56: __sent ack msg: 64 64 0'
b'pie: 56: Daemon waits for command'
b'(00.012355) Fetched ack: 64 64 0'
b'(00.012387) Error (criu/lsm.c:90): Invalid selinux context kernel'
b'(00.012391) ----------------------------------------'
b'(00.012394) Error (criu/cr-dump.c:1402): Dump core (pid: 56) failed with -1'
b'(00.012786) 56 was stopped'
b'(00.012941) Unlock network'
b'(00.012950) Unfreezing tasks into 1'
b'(00.012953) \tUnseizing 56 into 1'
b'(00.012963) Error (criu/cr-dump.c:1780): Dumping FAILED.'
------------------------ ERROR OVER ------------------------
################### Test zdtm/static/env00 FAIL at CRIU dump ###################
Send the 9 signal to  56
Wait for zdtm/static/env00(56) to die for 0.100000
##################################### FAIL #####################################
```

I have selinux "enabled" on my system in some strange way:

```
[root@snorch criu]# sestatus
SELinux status:                 disabled
[root@snorch criu]# getenforce
Disabled
```
But there is empty dir in /sys/fs/selinux:

```
[root@snorch criu]# ls /sys/fs/selinux/
[root@snorch criu]# 
```

And kerndat_lsm() states that it means that selinux is though enabled.

That's why in selinux_get_label() we try to get selinux process context but get "kernel" in responce.

E.g.:
```
[snorch@snorch criu]$ cat /proc/self/attr/current 
kernel[snorch@snorch criu]$
```

[...]

PR #1457, opened 2021-04-27
lsm: handle half initialized SELinux setups

adrianreber:
CRIU used to check for the existence of /sys/fs/selinux to see if SELinux is enabled on a system. We have seen systems with SELinux kind of enabled but reading out the labels gives does not return real labels.

To work around this, this commit adds a check during LSM detection if SELinux labels are in the right format. For CRIU this check means to see if there are at least 3 ':' in a label. If not CRIU switches to no LSM mode.

Fixes: #1402

### DIFF
--- a/criu/lsm.c
+++ b/criu/lsm.c
@@ -62,11 +62,37 @@ static int apparmor_get_label(pid_t pid, char **profile_name)
 }
 
 #ifdef CONFIG_HAS_SELINUX
+static int verify_selinux_label(char *ctx)
+{
+	char *pos;
+	int i;
+
+	/*
+	 * There are SELinux setups where SELinux seems to be enabled,
+	 * but the returned labels are not really valid. See also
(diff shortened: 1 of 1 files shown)


criu/lsm.c
```diff
@@ -62,41 +62,53 @@ static int apparmor_get_label(pid_t pid, char **profile_name)
 }
 
 #ifdef CONFIG_HAS_SELINUX
-static int selinux_get_label(pid_t pid, char **output)
+static int verify_selinux_labal(char *ctx)
```
rst0git, line 65: *label

adrianreber: Thanks. Fixed.

rst0git (approved):

rajbhar (approved):
Looks good to me. 
Tested-by: Rajneesh Bhardwaj <rajneesh.bhardwaj@amd.com>

PR #1459, opened 2021-04-27
Add annotations for code analyzers to ignore false positives

adrianreber:
I have taken the commits from #1235 which only add annotations to tell code analyzers about false positives and collected them all in this PR.

This PR only adds comments.

### DIFF
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -284,6 +284,7 @@ int uffd_open(int flags, unsigned long *features)
 	}
 
 	if (uffdio_api.api != UFFD_API) {
+		/* coverity[format_error] */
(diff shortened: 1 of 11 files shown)


criu/uffd.c
```diff
@@ -284,6 +284,7 @@ int uffd_open(int flags, unsigned long *features)
 	}
 
 	if (uffdio_api.api != UFFD_API) {
+		/* coverity[format_error] */
```
avagin, line 287: I think we need to support each annotation with a comment that explains why we think that it is a false positive case.

adrianreber:
The patches are over 6 months old. I do not remember any more. Closing.

PR #1460, opened 2021-04-27
Small static code analyzer fixes

adrianreber:
The last two remaining commits from #1235


criu/namespaces.c
```diff
@@ -527,7 +527,7 @@ static int open_ns_fd(struct file_desc *d, int *new_fd)
 	struct ns_desc *nd = NULL;
 	struct ns_id *ns;
 	int nsfd_id, fd;
-	char path[64];
+	char path[64] = "";
```
avagin, line 530: I don't see where we use path before initializing it.

adrianreber:
The patches are over 6 months old. I do not remember any more. Closing.

PR #1463, opened 2021-04-30
tools.mk: Use Python 3 by default

rst0git:
As of January 1st, 2020 Python 2 is no longer supported and many distributions no longer provide packages for Python 2 dependencies. This patch allows CRIU to use Python 3 by default when both major versions are available on a system.


adrianreber (approved):
Thanks. That was also something I wanted to change.

PR #1465, opened 2021-05-01
zdtm: add network namespace locking test

ZeyadYasser:
This test would be useful when testing that the network is locked using different ways (using iptables currently and eBPF later).

When CRIU dumps a process in a network namespace it locks the network so that no packets from peer enters the stack, otherwise RST will be sent by a kernel causing the connection to fail.

In netns_lock.c we try to enter the netns created by post-start hook so that criu locks the network namespace between dump and restore.

A TCP server is started in post-start hook inside the test netns and runs in the background detached from its parent so that it stays alive for the duration of the test.

Other hooks (pre-dump, pre-restore, post-restore) try to connect to the server.

Pre-dump and post-restore hooks should be able to connect successfully.

Pre-restore hook should not be able to connect but also it should not get connection refused as all packets are dropped in the namespace so the kernel shouldn't send an RST packet as a result. Instead we check that the connect operation causes a timeout.

### DIFF
--- a/test/zdtm/static/netns_lock.c
+++ b/test/zdtm/static/netns_lock.c
@@ -1,0 +1,71 @@
+#include "zdtmtst.h"
+
+const char *test_doc = "Check network namespace is locked between dump and restore\n";
+const char *test_author = "Zeyad Yasser <zeyady98@gmail.com>";
+
+#include <errno.h>
+#include <fcntl.h>
+#include <sched.h>
+
+#define NS_PATH "/var/run/netns/criu-net-lock-test"
+#define MAX_RETRY 3
+
+int main(int argc, char **argv)
+{
+	int i, ns_fd;
+
+	test_init(argc, argv);
+
+	/*
+	 * We try to enter the netns created by post-start hook so that
+	 * criu locks the network namespace between dump and restore.
+	 *
+	 * A TCP server is started in post-start hook inside the netns
+	 * and runs in the background detached from its parent so that
+	 * it stays alive for the duration of the test.
+	 *
+	 * Other hooks (pre-dump, pre-restore, post-restore) try to
+	 * connect to the server.
+	 *
+	 * Pre-dump and post-restore hooks should be able to connect
+	 * successfully.
+	 *
+	 * Pre-restore hook client with SOCCR_MARK should also connect
+	 * successfully.
+	 *
+	 * Pre-restore hook client without SOCCR_MARK should not be able
+	 * to connect but also should not get connection refused as all
+	 * packets are dropped in the namespace so the kernel shouldn't
+	 * send an RST packet as a result. Instead we check that the
+	 * connect operation causes a timeout.
+	 */
+
+	for (i = 0; i < MAX_RETRY; i++) {
+		ns_fd = open(NS_PATH, O_RDONLY);
+		if (ns_fd < 0) {
+			/* Netns not created yet by post-start hook */
+			sleep(1);
+			continue;
+		}
+		break;
+	}
+
+	if (ns_fd < 0) {
+		pr_perror("can't open network ns");
+		return 1;
+	}
+
+	if (setns(ns_fd, CLONE_NEWNET)) {
+		pr_perror("setns %d", ns_fd);
+		return 1;
+	}
+
+	close(ns_fd);
+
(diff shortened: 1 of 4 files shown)


test/zdtm/static/netns_lock.hook
```diff
@@ -0,0 +1,118 @@
...
+
+CLONE_NEWNET = 0x40000000
+PORT = 8880
+NETNS = "criu-net-lock-test"
+TIMEOUT = 0.1
+
+def nsenter():
+    f = open("/var/run/netns/{}".format(NETNS))
+    libc.setns(f.fileno(), CLONE_NEWNET)
+    f.close()
```
rst0git, line 20: Python has a [`with`](https://www.python.org/dev/peps/pep-0343/) statement that could handle exceptions to ensure closing the file
```py
with open("/var/run/netns/{}".format(NETNS)) as f:
    libc.setns(f.fileno(), CLONE_NEWNET)
```

rst0git (approved):

test/zdtm/static/netns_lock.desc
```diff
@@ -0,0 +1,4 @@
+{
+    'flavor': 'h',
+    'opts': '--tcp-established --join-ns net:/var/run/netns/criu-net-lock-test'
```
Snorch, line 3: We can remove --join-ns from dump ops, as we don't nedd it on dump.
```--- a/test/zdtm/static/netns_lock.desc
+++ b/test/zdtm/static/netns_lock.desc
@@ -1,4 +1,5 @@
 {
     'flavor': 'h',
-    'opts': '--tcp-established --join-ns net:/var/run/netns/criu-net-lock-test'
+    'opts': '--tcp-established',
+    'ropts': '--join-ns net:/var/run/netns/criu-net-lock-test'
 }
```

test/zdtm/static/netns_lock.c
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	if (setns(ns_fd, CLONE_NEWNET)) {
+		pr_perror("setns %d", ns_fd);
+		return 1;
+	}
+
+	close(ns_fd);
+
+	test_init(argc, argv);
```
Snorch, line 60: Probably we can move test_init() up into begining of main() so that we actually enter netns from final dumpable process and not from some intermediate helper. Note that this would also make you add `'flags': 'suid'` to test description file explicitly and not implicitly when helper process is not yet switched user ids, we need to have `suid` for nsenter.

Snorch (changes requested):

ZeyadYasser:
Thanks a lot for the feedback, Updated.

Snorch (approved):

ZeyadYasser:
I just noticed that I forget to test that packets with SOCCR_MARK are allowed to pass regardless of the netns lock. I will add a check for this shortly.

rst0git (approved):

test/zdtm/static/netns_lock.hook
```diff
@@ -0,0 +1,146 @@
...
+        cln.setsockopt(socket.SOL_SOCKET, SO_MARK, SOCCR_MARK)
+        cln.connect(("localhost", PORT))
+        resp = cln.recv(100).decode()
+        if resp != sys.argv[1]:
+            print("Expected '{}', got '{}'".format(sys.argv[1], resp))
+            sys.exit(1)
+    except socket.timeout:
+        print("Timeout when trying to connect to server")
+        sys.exit(1)
+    else:
```
ZeyadYasser, line 105: This is wrong, I shouldn't exit in case of success here. Test that the non-SOCCR_MARK connection fails first

ZeyadYasser:
This test does not check if other protocols are locked (they shouldn't) as only TCP packets should be dropped after dump.
Should this check be added to this test or a in separate test?

Snorch:
>This test does not check if other protocols are locked (they shouldn't) as only TCP packets should be dropped after dump.
Should this check be added to this test or a in separate test?

In my opinion it's enough to have a TCP-only check (at least for now, as we know that currently we just add protocol independent locking iptables rules).

If, as I can guess, you plan to rework network locking, you would be able to update the test case with specifics of new locking algorithm later.

rst0git (approved):

Issue #1425, opened 2021-03-29
Fail to dump systemcfg file on ppc


zl-wang:
on both RHEL8.3 and Ubuntu18.04 I tested, the same error is encountered on ppc64le.  At the end of dump log, something looks like the below:

...
(00.336930) vma 7fffa8030000 borrows vfi from previous 7fffa8020000
(00.337077) Dumping path for -3 fd via self 12 [/proc/powerpc/systemcfg]
(00.337085) Only file size could be stored for validation for file /proc/powerpc/systemcfg
(00.337090) Error (criu/proc_parse.c:672): Unsupported mapping found 00007fffa8040000-00007fffa8050000
(00.337097) Error (criu/cr-dump.c:1250): Collect mappings (pid: 304727) failed with -1
(00.337170) Unlock network
(00.337173) Unfreezing tasks into 1
(00.337175) 	Unseizing 304727 into 1
(00.337488) Error (criu/cr-dump.c:1768): Dumping FAILED.

avagin:
@zl-wang what app are you dumping? How does it use  /proc/powerpc/systemcfg?

zl-wang:
@avagin I am dumping JVM.  systemcfg is used to calculate time from epoch, which is a few times faster than using vdso gettimeofday etc.

avagin:
Cc: @ldu4  @0x7f454c46

ldu4:
This is a procfs file, how the others files from this fs are managed?

avagin:
Could you try ouy this patch:
```
diff --git a/criu/proc_parse.c b/criu/proc_parse.c
index 7b0b32e2e..f45c89be9 100644
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -173,7 +173,8 @@ static void parse_vma_vmflags(char *buf, struct vma_area *vma_area)
         * only exception is VVAR area that mapped by the kernel as
         * VM_IO | VM_PFNMAP | VM_DONTEXPAND | VM_DONTDUMP
         */
-       if (io_pf && !vma_area_is(vma_area, VMA_AREA_VVAR))
+       if (io_pf && !vma_area_is(vma_area, VMA_AREA_VVAR) &&
+           !vma_entry_is(vma_area->e, VMA_FILE_SHARED))
                vma_area->e->status |= VMA_UNSUPP;
 
        if (vma_area->e->madv)
```

[...]

PR #1471, opened 2021-05-07
Allow systemcfg proc file to be dumped

zl-wang:
Currently, it cannot be check-pointed, because that type of file is on UNSUPP list.

Signed-off-by: zl-wang <zlwang@ca.ibm.com>


zl-wang:
criu #1425

Issue #1429, opened 2021-03-30
Restored JVM process died silently


zl-wang:
This particular exercise was carried out on RHEL8.3 ppc64le:

I rebuilt JDK which is not using systemcfg file anymore, in order to avoid the non-checkpointable situation described here: #1425 .  Now, I can both checkpoint it and restore it successfully (according to the criu log files). However, I either cannot see the restored java process running (ps -ef|grep java  ... no show, that is), or sometimes i can see it alive but just die silently when I curl to ping it (supposed to respond the same).  No other output whatsoever to know why it died.  Tail of the dump.log:

(00.065423) Dumping pstree (pid: 1825601)
(00.065426) ----------------------------------------
(00.065428) Process: 1825601(1825601)
(00.065465) ----------------------------------------
(00.065537) cg: All tasks in criu's cgroups. Nothing to dump.
(00.065539) unix: Dumping external sockets
(00.065547) unix: 	Dumping extern: ino 5190844 peer_ino 1442227 family    1 type    1 state  1 name /var/lib/sss/pipes/nss
(00.065553) unix: 	Dumped extern: id 0x145 ino 5190844 peer 0 type 2 state 10 name 23 bytes
(00.065557) unix: Found ino 1442227 in exception unix sk list
(00.065559) unix: found exception for unix name-less external socket.
(00.065563) unix: 	Dumped: id 0xb2 ino 1442227 peer 5190844 type 1 state 1 name 0 bytes
(00.065568) Writing image inventory (version 1)
(00.065668) Running post-dump scripts
(00.065671) Unfreezing tasks into 2
(00.065673) 	Unseizing 1825601 into 2
(00.068127) Writing stats
(00.068242) Dumping finished successfully

Tail of the restore.log:

(00.040256) 1825603 was trapped
(00.040259) 1825603 (native) is going to execute the syscall 221, required is 172
(00.040283) 1825601 was trapped
(00.040286) 1825601 (native) is going to execute the syscall 91, required is 172
(00.040292) 1825605 was trapped
(00.040295) 1825605 (native) is going to execute the syscall 172, required is 172
(00.040304) 1825605 was stopped
(00.040307) 1825601 was trapped
(00.040310) 1825601 (native) is going to execute the syscall 172, required is 172
(00.040318) 1825601 was stopped
(00.040322) 1825603 was trapped
(00.040325) 1825603 (native) is going to execute the syscall 172, required is 172
(00.040333) 1825603 was stopped
[...]

PR #1473, opened 2021-05-10
SIGTRAP missed in restored process' caught sigset

zl-wang:
When checkpoint is restored, SIGTRAP is missed from the sigset
    which are in the original caught set. If the checkpoint was using
    SIGTRAP, the restored process just died silently. This fix
    adds it back into the caught set.

criu #1429

### DIFF
--- a/compel/src/lib/infect.c
+++ b/compel/src/lib/infect.c
@@ -447,6 +447,13 @@ static int parasite_run(pid_t pid, int cmd, unsigned long ip, void *stack,
 	k_rtsigset_t block;
 
 	ksigfillset(&block);
+
+	/* This meant to be a work-around for the existing bug that the restored
+	 * process' caught sigset is not equal to the checkpoint process'. It was
+	 * noticed that SIGTRAP is missing there outstandingly.
+	 */
+	block.sig[0] = block.sig[0] & ~(1<<(SIGTRAP - 1));
+
 	if (ptrace(PTRACE_SETSIGMASK, pid, sizeof(k_rtsigset_t), &block)) {
 		pr_perror("Can't block signals for %d", pid);
 		goto err_sig;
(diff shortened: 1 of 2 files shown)


compel/src/lib/infect.c
```diff
@@ -447,6 +447,7 @@ static int parasite_run(pid_t pid, int cmd, unsigned long ip, void *stack,
 	k_rtsigset_t block;
 
 	ksigfillset(&block);
+	block.sig[0] = block.sig[0] & ~(1<<(SIGTRAP - 1));
```
avagin, line 450: It is better to have a comment that explains why SIGTRAP isn't blocked.
We need to check that a process doesn't have a sigtrap signal in queues.

Snorch: Yes, this probably should be explained a little bit more... Why not-blocking SIGTRAP helps? PTRACE_SETSIGINFO seem to only change task_struct->blocked https://github.com/torvalds/linux/blob/8ac91e6c6033ebc12c5c1e4aa171b81a662bd70f/kernel/ptrace.c#L1117 But SigCgt is determined from the fact that actual signal handler was set https://github.com/torvalds/linux/blob/8ac91e6c6033ebc12c5c1e4aa171b81a662bd70f/fs/proc/array.c#L262. From the firs look they seem independent.... Sorry if I miss something obvious.

zl-wang: this appeared to myself as a workaround ... for the existing bug that the caught sigset is not restored correctly as the checkpoint process was, as shown in this investigation:  https://github.com/checkpoint-restore/criu/issues/1429#issuecomment-815802412   and  https://github.com/checkpoint-restore/criu/issues/1429#issuecomment-815807630

this issue existed on zLinux too.  Although i didn't verify this on x86 linux, i expected it is there too.  There might be a better place to fix this, rather than in infect.c.  However, I am not familiar with the whole code enough to locate where it should be.

avagin: @Snorch in the issue, I explained why we can't block SIGTRAP here and how it affects SigCgt.

@zl-wang Thank you for the patch. I understand that you are not familiar with the CRIU code, so I will fix the patch and merge it.

PR #1482, opened 2021-05-20
parasite: don't block SIGTRAP

avagin:
This is the workaround for #1429.
    
The parasite code contains instructions that trigger SIGTRAP to stop at certain points. In such cases, the kernel sends a force SIGTRAP that can't be ignored and if it is blocked, the kernel resets its signal handler to a default one and unblocks it. It means that if we want to  save the origin signal handle


zl-wang:
@avagin   thanks for taking care of this ...

0x7f454c46 (approved):
Yeah, let's have it for now.
In theory, sigtrap might be unexpectedly raised by hw watchpoint/breakpoint or `X86_EFLAGS_TF`. But it didn't work for such dumpee before this PR, so no change in that.
I have some ideas how-to rework this later.

PR #1483, opened 2021-05-21
uffd: fixed for exiting exiting program as user

nithin-jaikar:
While Dumping the process, it fails and exits as a user because of kerndat_uffd() returns -1 and due permission error 
uffd = syscall(SYS_userfaultfd, flags); this function fails inturn kerndat_uffd() fails. 
In the change it ignores the permission error and proceeds further.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -919,7 +919,10 @@ static int kerndat_uffd(void)
 	if (uffd < 0) {
 		if (uffd == -ENOSYS)
 			return 0;
-
+		if (uffd == -EPERM) {
+			pr_info("Lazy pages are disabled\n");
+			return 0;
+		}
 		pr_err("Lazy pages are not available\n");
 		return -1;
 	}


criu/kerndat.c
```diff
@@ -864,7 +864,8 @@ static void kerndat_save_cache(void)
 	int fd, ret;
 	struct statfs s;
 
-	fd = open(KERNDAT_CACHE_FILE_TMP, O_CREAT | O_EXCL | O_WRONLY, 0600);
+	fd = open(KERNDAT_CACHE_FILE_TMP, O_CREAT | O_EXCL | O_WRONLY, 0700);
+	pr_msg("open fd: %d, %s  \n",fd,strerror(errno));
```
xemul: This pr_msg() doesn't belong to this patch.

criu/kerndat.c
```diff
@@ -863,7 +863,8 @@ static void kerndat_save_cache(void)
 {
 	int fd, ret;
 	struct statfs s;
-
+	ret= access(KERNDAT_CACHE_FILE_TMP, R_OK | W_OK | X_OK | F_OK );
+	pr_msg("open acess: %d, %s  \n",ret,strerror(errno));
```
xemul: Please, drop this change. Missing permissions are found out by EACCESS errno of open.

xemul (changes requested):
@nithin-jaikar rework the PR, leave only the EPERM checking of uffd opening and write a proper patch comment.

criu/uffd.c
```diff
@@ -270,6 +270,10 @@ int uffd_open(int flags, unsigned long *features)
 	uffd = syscall(SYS_userfaultfd, flags);
 	if (uffd == -1) {
 		pr_info("Lazy pages are not available: %s\n", strerror(errno));
+		if (errno == EPERM) {
+			pr_info("Uffd loader is not loaded \n");
+			return 0;
+		}
```
Snorch, line 276: I worry that this patch introduces a bug. E.g. If we got EPERM errno from uffd_open() and return 0 from uffd_open() to setup_uffd(). Next send_uffd would send random fd at index "0" into lazy pages socket, which is probably unexpected.

xemul:
@nithin-jaikar , don't add 5th commit, please. Remove all commits and leave only one.

criu/kerndat.c
```diff
@@ -909,7 +909,6 @@ static int kerndat_uffd(void)
 
 	kdat.uffd_features = 0;
 	uffd = uffd_open(0, &kdat.uffd_features);
-
```
xemul, line 912: Remove this hunk

criu/kerndat.c
```diff
@@ -919,7 +918,10 @@ static int kerndat_uffd(void)
 	if (uffd < 0) {
 		if (uffd == -ENOSYS)
 			return 0;
-
+		if (uffd == -EPERM) {
+		pr_err("Uffd loader is not loaded \n");
```
xemul, line 922: Indentation is broken

xemul (changes requested):
The patch (not PR, but the very commit) must have a descriptive comment

xemul:
@nithin-jaikar , you've somehow pulled in the Adrian's patch, then merged yours with his and then added the fix I requested as another commit. Please, be more attentive and make a proper PR.

xemul:
Ok, @nithin-jaikar , now let's add the comment to the commit (the text that accompanies the pull-request can be just copied)

criu/kerndat.c
```diff
@@ -919,6 +919,10 @@ static int kerndat_uffd(void)
 	if (uffd < 0) {
 		if (uffd == -ENOSYS)
 			return 0;
+		if (uffd == -EPERM) {
+			pr_info("uffd loader is not loaded. \n");
```
xemul, line 923: The logging text  should be "Lazy pages are disabled\n"

xemul (changes requested):

adrianreber:
From the PR description it is not clear to me what this tries to solve. Some more details how this could happen would be interesting.

xemul:
@adrianreber , this is the part of "dump shared memory as user" task.

The "criu dump" used to work from non-root some time ago, but then this mode testing stopped and the functionality got broken. On of the found problems is that non-root is not allowed to open uffd, so criu cannot even check the kerndat stuff.

adrianreber:
Ah, thanks. I saw similar errors when doing the non-root changes in #1155 

Anyway, that information would be good to have in the commit and/or PR.

xemul:
Whee, comment commit message is mastered!
@nithin-jaikar , now please read https://criu.org/How_to_submit_patches and make the commit message look the way it must. E.g. like this

````
kerndat: Handle non-root mode when checking uffd

When criu is run as user it fails and exits because of kerndat_uffd() returning -1.
This, in turn, happens after uffd = syscall(SYS_userfaultfd, flags); which only works
for root.

In the change it ignores the permission error and proceeds further just like it's done
for e.g. pagemap checking.

Signed-off-by: Nithin Jaikar J <jaikar006@gmail.com>
````

xemul (approved):

rst0git (approved):
LGTM

Snorch (changes requested):
`#define EPERM            1`

Should not we differentiate errno == EPERM from syscall SYS_userfaultfd and later errors in uffd_open()

Snorch:
I still think that my objection is valid, so https://github.com/checkpoint-restore/criu/pull/1493

PR #1491, opened 2021-05-30
test/others: fix race condition

rst0git:
`criu dump` is not expected to fail, but when this happens the test should exit with an error.

### DIFF
--- a/test/others/crit/test.sh
+++ b/test/others/crit/test.sh
@@ -18,12 +18,10 @@ function _exit {
 function gen_imgs {
 	setsid ./loop.sh < /dev/null &> /dev/null &
 	PID=$!
-	$CRIU dump -v4 -o dump.log -D ./ -t $PID
-	# shellcheck disable=SC2181
-	if [ $? -ne 0 ]; then
+	if ! $CRIU dump -v4 -o dump.log -D ./ -t "$PID"; then
 		cat dump.log
-		kill -9 $PID
-		_exit 1
+		kill -9 "$PID"
+		exit 1
 	fi
 
 	images_list=$(ls -1 ./*.img)
--- a/test/others/crit/test.sh
+++ b/test/others/crit/test.sh
@@ -27,7 +27,7 @@ function gen_imgs {
 	images_list=$(ls -1 ./*.img)
 	if [ -z "$images_list" ]; then
 		echo "Failed to generate images"
-		_exit 1
+		exit 1
 	fi
 }
 
--- a/test/others/crit/test.sh
+++ b/test/others/crit/test.sh
@@ -21,6 +21,7 @@ function gen_imgs {
 	$CRIU dump -v4 -o dump.log -D ./ -t $PID
 	# shellcheck disable=SC2181
 	if [ $? -ne 0 ]; then
+		cat dump.log
 		kill -9 $PID
 		_exit 1
 	fi
--- a/test/others/crit/test.sh
+++ b/test/others/crit/test.sh
@@ -16,8 +16,8 @@ function _exit {
 }
 
 function gen_imgs {
-	setsid ./loop.sh < /dev/null &> /dev/null &
-	PID=$!
+	# Assign PID after the background task has started
+	PID=$( (setsid ./loop.sh < /dev/null &> /dev/null) & jobs -r -p)
 	if ! $CRIU dump -v4 -o dump.log -D ./ -t "$PID"; then
 		cat dump.log
 		kill -9 "$PID"
--- a/test/others/criu-coredump/test.sh
+++ b/test/others/criu-coredump/test.sh
@@ -10,10 +10,10 @@ function _exit {
 function gen_imgs {
 	# Assign PID after the background task has started
 	PID=$( (setsid ./loop.sh < /dev/null &> /dev/null) & jobs -r -p)
-	$CRIU dump -v4 -o dump.log -D ./ -t $PID
-	if [ $? -ne 0 ]; then
-		kill -9 $PID
-		_exit 1
+	if ! $CRIU dump -v4 -o dump.log -D ./ -t "$PID"; then
+		cat dump.log
+		kill -9 "$PID"
+		exit 1
 	fi
 
 	images_list=$(ls -1 *.img)
--- a/test/others/criu-coredump/test.sh
+++ b/test/others/criu-coredump/test.sh
@@ -8,8 +8,8 @@ function _exit {
 }
 
 function gen_imgs {
-	setsid ./loop.sh < /dev/null &> /dev/null &
-	PID=$!
+	# Assign PID after the background task has started
+	PID=$( (setsid ./loop.sh < /dev/null &> /dev/null) & jobs -r -p)
 	$CRIU dump -v4 -o dump.log -D ./ -t $PID
 	if [ $? -ne 0 ]; then
 		kill -9 $PID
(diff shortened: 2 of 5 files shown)


adrianreber:
Any ideas why it fails? Is it a race condition or timing? The process to be dumped looks rather simple.

rst0git:
Yes, I think we can add `sleep 1` before `criu dump` to fix it, but I haven't been able to replicate the error locally yet.

adrianreber:
pgrep does not work on alpine and so it still fails

rst0git:
> pgrep does not work on alpine and so it still fails

Thanks, this should be fixed now.

adrianreber (approved):

test/others/criu-coredump/test.sh
```diff
@@ -9,11 +9,12 @@ function _exit {
...
 	setsid ./loop.sh < /dev/null &> /dev/null &
-	PID=$!
-	$CRIU dump -v4 -o dump.log -D ./ -t $PID
-	if [ $? -ne 0 ]; then
-		kill -9 $PID
-		_exit 1
+	sleep .1  # wait for loop.sh to start
+	PID=$(pgrep -n loop.sh)
+	if ! $CRIU dump -v4 -o dump.log -D ./ -t "$PID"; then
+		cat dump.log
```
Snorch, line 15: Printing full log may be a bit too noizy, isn't it? You can probably grep Error or print the tail of the file instead.

Snorch (commented):
I see one more use of function _exit which checks $? just after echo... That looks wrong, may be also fix it? https://github.com/checkpoint-restore/criu/blob/d34c7cee8614f5eeb9a58546f7867a1dafb67200/test/others/crit/test.sh#L31

test/others/crit/test.sh
```diff
@@ -17,7 +17,8 @@ function _exit {
 
 function gen_imgs {
 	setsid ./loop.sh < /dev/null &> /dev/null &
-	PID=$!
+	sleep .1  # wait for loop.sh to start
```
avagin, line 20: I am not a big fan of using sleeps to fix races ;). I think we can use pipes here:
https://github.com/opencontainers/runc/blob/master/tests/integration/checkpoint.bats#L24

test/others/criu-coredump/test.sh
rst0git: We don't expect `criu dump` to fail and most likely this was the reason why the logs were not shown when it fails. However, when the test fails in CI environment it would be difficult to find out why without the logs.

> You can probably grep Error or print the tail of the file instead.

Being able to see the whole content of `dump.log` is more useful.

avagin: @Snorch we print it only when criu fails, so it should be fine. The verbosity has to be 4.

test/others/crit/test.sh
rst0git: Hi @avagin, I was able to replace `sleep` with nested command substitution.
```
2021-06-06T11:12:02.5890420Z ++ ++ setsid ./loop.sh
2021-06-06T11:12:02.5894022Z jobs -r -p
2021-06-06T11:12:02.5900376Z + PID=658293
2021-06-06T11:12:02.5901311Z + /criu/criu/criu dump -v4 -o dump.log -D ./ -t 658293
```

rst0git:
> That looks wrong, may be also fix it? 

Done.

test/others/criu-coredump/test.sh
Snorch: Ok, sounds reasonable.

Snorch (approved):

test/others/crit/test.sh
avagin: pls explain in more detail. I don't understand...

test/others/crit/test.sh
```diff
@@ -16,8 +16,7 @@ function _exit {
 }
 
 function gen_imgs {
-	setsid ./loop.sh < /dev/null &> /dev/null &
-	PID=$!
+	PID=$( (setsid ./loop.sh < /dev/null &> /dev/null) & jobs -r -p)
```
avagin, line 19: I don't understand how this fixes the race. Could you explain and add a comment to the code? Thanks.

test/others/crit/test.sh
rst0git: The problem that was causing the test to fail was that `PID=!$` was executed before `setsid ./loop.sh`. These commands were not executed in the correct order on Alpine and Fedora Rawhide, as show below.
```bash
setsid ./loop.sh </dev/null &> /dev/null &
PID=$!
$CRIU dump -v4 -o dump.log -D ./ -t $PID
```
```
2021-05-30T13:36:43.8309467Z + PID=494959
2021-05-30T13:36:43.8310637Z + /criu/criu/criu dump -v4 -o dump.log -D ./ -t 494959
2021-05-30T13:36:43.8314025Z + setsid ./loop.sh
```
To solve this problem, we could use command substitution to preserve the order of execution.
Bash performs expansion of `$( (command) )` in a subshell environment and replacing it with the standard output of the command. Hence, `PID` would be assigned after `setsid ./loop.sh` has started.
```bash
PID=$( (setsid ./loop.sh < /dev/null &> /dev/null) & jobs -r -p)
```
```
2021-06-06T11:12:02.5890420Z ++ ++ setsid ./loop.sh
2021-06-06T11:12:02.5894022Z jobs -r -p
2021-06-06T11:12:02.5900376Z + PID=658293
```

test/others/crit/test.sh
rst0git: I will add a comment

test/others/crit/test.sh
kolyshkin: These commands were not executed in the correct order on Alpine and Fedora Rawhide, as show below.
```
2021-05-30T13:36:43.8309467Z + PID=494959
2021-05-30T13:36:43.8310637Z + /criu/criu/criu dump -v4 -o dump.log -D ./ -t 494959
2021-05-30T13:36:43.8314025Z + setsid ./loop.sh
```

I don't believe this is an order of execution problem, but rather some logging race.

The actual problem I think is `setsid`. Quoting `setsid(1)` man:

> The command calls fork(2) if already a process group leader.

So it might or might not fork, depending on the environment. I did a quick test and in my terminal it does not -- but we can add `-f` to make it fork.

```console
[kir@kir-rhat criu]$ cat x
setsid ls -ld /proc/self &
echo PID=$!
[kir@kir-rhat criu]$ bash -x x 
+ echo PID=986412
PID=986412
+ setsid ls -ld /proc/self
lrwxrwxrwx. 1 root root 0 Jun  1 14:02 /proc/self -> 986412
[kir@kir-rhat criu]$ vim x # Add -f to setsid so it always forks
[kir@kir-rhat criu]$ cat x
setsid -f ls -ld /proc/self &
echo PID=$!
[kir@kir-rhat criu]$ bash -x x
+ echo PID=986436
PID=986436
+ setsid -f ls -ld /proc/self
lrwxrwxrwx. 1 root root 0 Jun  1 14:02 /proc/self -> 986438
[kir@kir-rhat criu]$ 
```

As you can see, if `setsid -f` is used, PIDs don't match. Same happens in GHA environments even without `-f` I guess.

So this is the cause.

As to the solution, I think the best way is to have `loop.sh` write its PID to a file, and wait for this file in the parent (using a polling loop, or `inotifywait(1)`, or, better, use a pipe (as @avagin suggested above) to pass the PID. The implementation is left as an exercise for the reader; runc checkpoint.bats provides a good example.

PR #1493, opened 2021-05-31
kdat: make kerndat_uffd differentiate -EPERM and -1 from uffd_open

Snorch:
Else "Failed to get uffd API" and "Incompatible uffd API ..." errors are
just ignored, which is probably not what we want.

Fixes: cfdeac4a4 ("kerndat: Handle non-root mode when checking uffd")

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -905,22 +905,24 @@ unl:
 
 static int kerndat_uffd(void)
 {
-	int uffd;
+	int uffd, err = 0;
 
 	kdat.uffd_features = 0;
-	uffd = uffd_open(0, &kdat.uffd_features);
+	uffd = uffd_open(0, &kdat.uffd_features, &err);
 
 	/*
-	 * uffd == -ENOSYS means userfaultfd is not supported on this
-	 * system and we just happily return with kdat.has_uffd = false.
-	 * Error other than -ENOSYS would mean "Houston, Houston, we
+	 * err == ENOSYS means userfaultfd is not supported on this system and
+	 * we just happily return with kdat.has_uffd = false.
+	 * err == EPERM means that userfaultfd is not allowed as we are
+	 * non-root user, so we also return with kdat.has_uffd = false.
+	 * Errors other than ENOSYS and EPERM would mean "Houston, Houston, we
 	 * have a problem!"
 	 */
 	if (uffd < 0) {
-		if (uffd == -ENOSYS)
+		if (err == ENOSYS)
 			return 0;
-		if (uffd == -EPERM) {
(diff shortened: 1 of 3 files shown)


Snorch:
Note: I'm not sure this is a best way to fix it, probable se should just add "int *err" argument to uffd_open to return errno...

criu/kerndat.c
```diff
@@ -917,9 +917,9 @@ static int kerndat_uffd(void)
 	 * have a problem!"
 	 */
 	if (uffd < 0) {
-		if (uffd == -ENOSYS)
+		if (uffd == -ENOSYS-1)
```
rppt, line 920: I'd prefer `"int *err"` argument to `uffd_open`.
Now looking at it, I'm not sure we need to fail at all if `uffd_open()` returns error. Apparently, simply stetting `kdat.has_uffd` to false if `uffd_open()` failed would be sufficient.
Probably we'd also need to make sure that `opts.lazy_pages` and `kdat.has_uffd` match, somewhere in `main()`...

xemul: Me too. @Snorch , please

xemul (changes requested):

criu/kerndat.c
Snorch: `"int *err"` part done.

Not sure if you instead mean to

>simply stetting kdat.has_uffd to false if uffd_open() failed would be sufficient

Please correct me if I did wrong one from two options =)

rppt: > Not sure if you instead mean to
> 
> > simply stetting kdat.has_uffd to false if uffd_open() failed would be sufficient
> 
> Please correct me if I did wrong one from two options =)

I was mostly thinking aloud :)
The point is that we don't really care if we can use userfaultfd unless user explicitly asked for lazy pages. So it does not matter what was the exact reason `uffd_open()` failed and we can always return 0 from `kerndat_uffd()`. The availability of userfault will be set in `kdat.has_uffd`. If the user didn't ask for lazy-pages we can continue perfectly fine even if userfault is not available and only fail restore if both `kdat.has_uffd = false` and `opts.lazy_pages = true`

Snorch: > The point is that we don't really care if we can use userfaultfd unless user explicitly asked for lazy pages.

From other point of view, we might want to know early if kernel behaves strangely and after successful SYS_userfaultfd we don't expect to see uffd_api_ioctl, to fail and see bad uffdio api. We expect kernel to report absence of uffd gracefuly...

With the way it is done now, we would detect this strange behavior even if uffd is not actually used.

rppt (approved):

PR #1500, opened 2021-06-08
Enable changing of mount context on restore

adrianreber:
This change is motivated by checkpointing and restoring container in Pods.

When restoring a container into a new Pod the SELinux label of the existing Pod needs to be used and not the SELinux label saved during checkpointing.

The option `--lsm-profile` already enables changing of process SELinux labels on restore. If there are, however, _tmpfs_ checkpointed they will be mounted during restore with the same context as during checkpointing. This can look like the following example:

 `context="system_u:object_r:container_file_t:s0:c82,c137"`

On restore we want to change this context to match the mount label of the Pod this container is restored into. Changing of the mount label is now possible with the new option `--mount-context`:

 `criu restore --mount-context "system_u:object_r:container_file_t:s0:c204,c495"`

This will lead to mount options being changed to

 `context="system_u:object_r:container_file_t:s0:c204,c495"`

Now the restored container can access all the files in the container again.

This has been tested in combination with runc and CRI-O.


rst0git:
@adrianreber Would it be possible to rename `--mount-context` to `--lsm-mount-context`?
IMHO this would clarify the meaning of "context".

Documentation/criu.txt
```diff
@@ -544,6 +544,28 @@ The 'mode' may be one of the following:
     Specify an LSM profile to be used during restore. The _type_ can be
     either *apparmor* or *selinux*.
 
+*--mount-context*::
```
rst0git, line 547: `*--mount-context* ['context']::`

adrianreber:
> @adrianreber Would it be possible to rename `--mount-context` to `--lsm-mount-context`?
> IMHO this would clarify the meaning of "context".

No problem. I will wait a bit to see if someone else has an opinion on the name, but I am okay with any name.

adrianreber:
@avagin What do you think? Should `lsm` be included in the option?

criu/mount.c
```diff
@@ -3006,9 +3007,58 @@ static int collect_mnt_from_image(struct mount_info **head, struct mount_info **
...
+			 * The content between 'context=' and ',inode64' will be replaced
+			 * with opts.mount_context.
+			 */
+			cleanup_free char *before_context = NULL;
+			char *other_options;
+			char *context_start;
+			char *context_end;
+
+			/* Skip 'context=' */
+			context_start = old_context + 8;
```
Snorch, line 3032: Please don't use magic numbers.
```
#define CONTEXT_OPT "context="
old_context + strlen(CONTEXT_OPT)
```

criu/mount.c
```diff
@@ -3006,9 +3007,58 @@ static int collect_mnt_from_image(struct mount_info **head, struct mount_info **
 		if (!pm->source)
 			goto err;
 
-		pm->options = xstrdup(me->options);
-		if (!pm->options)
-			goto err;
+		old_context = strstr(me->options, "context=");
+		if (opts.mount_context && old_context) {
+			/*
```
Snorch, line 3012: Let's put these "if"'s code into separate function, mount_update_lsm_context or something.

Snorch (commented):
Generally looks good. But do we always have single lsm context? For system containers we would probably have mounts with different lsm contexts... So I'm just wondering if it should be a explicit old_context->new_context mapping instead of *any_context*->new_context

test/zdtm/static/change_mnt_context.c
```diff
@@ -0,0 +1,60 @@
...
+		goto out;
+	}
+
+	while (fgets(line, sizeof(line), mi)) {
+		int result;
+
+		if (!strstr(line, source))
+			continue;
+
+		result = sscanf(line, "%*d %*d %*d:%*d %*s %*s %*s - %*s %*s %s", opts);
```
Snorch, line 44: Should use `strstr(line, " - ")` as before `-` we can have variable number of arguments.

test/zdtm/static/change_mnt_context.c
```diff
@@ -0,0 +1,60 @@
...
+		if (result < 0) {
```
Snorch, line 45: Probably `result != 1`

adrianreber:
> Generally looks good. But do we always have single lsm context?

For my use case (OCI containers and Kubernetes Pods) it is only one context.

> For system containers we would probably have mounts with different lsm contexts... So I'm just wondering if it should be a explicit old_context->new_context mapping instead of _any_context_->new_context

We have the same problem with SELinux process labels. Currently we are only supporting that all processes have the same label.

So this follows the existing SELinux process label support and extends it to mounts.

That solves the problem I am currently trying to solve. If we ever want to support different mount labels and different process labels it is a much bigger effort.

I would prefer to keep this patch limited to replacing the context as is it does right now. But you are right, it does not support multiple mount labels. But from what I have seen it is not currently necessary for anything using CRIU.

Thanks for the code review, I will update the PR.

adrianreber:
Addressed all review comments and force pushed. Let's see what CI says.

criu/mount.c
```diff
@@ -2942,6 +2944,68 @@ static int get_mp_mountpoint(char *mountpoint, struct mount_info *mi, char *root
...
+
+	before_context = xstrdup(mount_opts);
+	pr_debug("%p\n", before_context);
+	if (unlikely(!before_context))
+		return NULL;
+	before_context[context_start - mount_opts] = 0;
+
+	ret = asprintf(&new_options, "%s\"%s\"%s", before_context,
+			opts.lsm_mount_context, other_options ? other_options : "");
+	pr_debug("ret %d\n", ret);
```
rst0git, line 3001: The information printed with this `pr_debug` doesn't look useful.

rst0git:
The problem with the image streamer test looks unrelated to the changes in this PR and I opened this issue for it: https://github.com/checkpoint-restore/criu/issues/1504.

criu/mount.c
adrianreber: Yes, that is wrong.

adrianreber:
Right now my assumption is that the CentOS 7 failure is related to pid caching in the old version of glibc. I will try to test it later to verify.

criu/mount.c
adrianreber: removed the pr_debug

criu/mount.c
```diff
@@ -2942,6 +2944,67 @@ static int get_mp_mountpoint(char *mountpoint, struct mount_info *mi, char *root
...
+		}
+	} else {
+		context_end = context_start;
+	}
+
+	/* Find next after optionally skipping quotes. */
+	other_options = strchr(context_end, ',');
+
+	before_context = xstrdup(mount_opts);
+	pr_debug("%p\n", before_context);
```
rst0git, line 2994: This debug print statement looks unnecessary.

criu/mount.c
```diff
@@ -3006,8 +3069,9 @@ static int collect_mnt_from_image(struct mount_info **head, struct mount_info **
 		if (!pm->source)
 			goto err;
 
-		pm->options = xstrdup(me->options);
-		if (!pm->options)
+		pm->options = mount_update_lsm_context(me->options);
+		pr_debug("options %p\n", pm->options);
```
rst0git, line 3073: This debug print statement looks unnecessary.

adrianreber:
Sorry, I left too many debug statements. The should all be gone now.

rst0git (approved):

adrianreber:
@Snorch please take a look if you think this is ready to be merged.

criu/mount.c
```diff
@@ -2942,6 +2944,66 @@ static int get_mp_mountpoint(char *mountpoint, struct mount_info *mi, char *root
...
+		return NULL;
+	before_context[context_start - mount_opts] = 0;
+
+	ret = asprintf(&new_options, "%s\"%s\"%s", before_context,
+			opts.lsm_mount_context, other_options ? other_options : "");
+	if (unlikely(ret < 0))
+		return NULL;
+	pr_debug("\t\tChanged mount 'context=' to %s\n", new_options);
+
+	return xstrdup(new_options);
```
avagin, line 3004: we can return new_options here, can't we?

avagin:
LGTM

criu/mount.c
adrianreber: yes

test/zdtm/static/change_mnt_context.c
Snorch: Brief check shows that you haven't check the problem still.

I've just took your parse string and tried parsing one simple mount line in mountinfo...

```
[snorch@fedora tmp]$ cat test.c 
#include <stdio.h>
#include <string.h>

#define MNTINFO_EX "96 90 259:1 / /boot/efi rw,relatime shared:48 - vfat /dev/nvme0n1p1 rw,fmask=0077,dmask=0077,codepage=437,iocharset=ascii,shortname=winnt,errors=remount-ro"

int main()
{
	char buf[1024];
	int ret;

	ret = sscanf(MNTINFO_EX, "%*d %*d %*d:%*d %*s %*s %*s - %*s %*s %s", buf);
	if (ret != 1) {
		printf("Failed sscanf: %d\n", ret);
		return 1;
	}

	printf("Parsed :\"%s\"\n", buf);
	return 0;
}
[snorch@fedora tmp]$ gcc -o test test.c
[snorch@fedora tmp]$ ./test
Failed sscanf: 0
```

Snorch:
I still don't like mountinfo parsing, it's not robust.

test/zdtm/static/change_mnt_context.c
adrianreber: Any recommendations how to make the parsing better?

I added
```

		if (!strstr(line, " - "))
			continue;
```

a couple of days ago and I am only looking at line that contain source of the mount `change_mount_context.test`. Any check I can add to make sure it works?

avagin: > Should use strstr(line, " - ") as before - we can have variable number of arguments.

@Snorch do you mean something like this:

diff --git a/test/zdtm/static/change_mnt_context.c b/test/zdtm/static/change_mnt_context.c
index 4d2bf7ccd..6d436014b 100644
--- a/test/zdtm/static/change_mnt_context.c
+++ b/test/zdtm/static/change_mnt_context.c
@@ -37,14 +37,16 @@ int main(int argc, char **argv)
 
        while (fgets(line, sizeof(line), mi)) {
                int result;
+               char *pos;
 
                if (!strstr(line, source))
                        continue;
 
-               if (!strstr(line, " - "))
+               pos = strstr(line, " - ");
+               if (!pos)
                        continue;
 
-               result = sscanf(line, "%*d %*d %*d:%*d %*s %*s %*s - %*s %*s %s", opts);
+               result = sscanf(pos, " - %*s %*s %s", opts);
                if (result != 1) {
                        fail("Not able to sscanf line from mountinfo");
                        goto out;

adrianreber:
> I still don't like mountinfo parsing, it's not robust.
> 
> See [#1500 (comment)](https://github.com/checkpoint-restore/criu/pull/1500#discussion_r658584302)
> 
> Upd: except that Looks good.

@Snorch Does your "Upd: except that Looks good" mean that the review discussion can be marked as resolved? You could also mark it as resolved if you want.

avagin:
> > I still don't like mountinfo parsing, it's not robust.
> > See [#1500 (comment)](https://github.com/checkpoint-restore/criu/pull/1500#discussion_r658584302)

@adrianreber Have you seen my last comment there?

adrianreber:
> > > I still don't like mountinfo parsing, it's not robust.
> > > See [#1500 (comment)](https://github.com/checkpoint-restore/criu/pull/1500#discussion_r658584302)
> 
> @adrianreber Have you seen my last comment there?

No I have not. But now I saw it. Thanks. I will update the PR.

adrianreber:
Updated with @avagin test code mountinfo parsing changes.

Snorch (approved):
Looks good now!

Thanks to @avagin for comment with diff, yes I meant exactly this change!

test/zdtm/static/change_mnt_context.c
adrianreber: Thanks @avagin for the example.

adrianreber:
Thanks for everyone's help.

PR #1501, opened 2021-06-09
test/others: Refactor loop process

rst0git:
This pull request is a follow-up from the discussion in https://github.com/checkpoint-restore/criu/pull/1491. It aims to replace the `loop.sh` script with a `loop` program that is better suited for the tests.


test/others/loop.c
```diff
@@ -0,0 +1,40 @@
...
+	pid_t sid;
+
+	pid = fork();
+	if (pid < 0) {
+		perror("fork failed!");
+		return -1;
+	}
+
+	if (pid == 0) {
+		sid = setsid();
```
avagin, line 19: but the parent process doesn't wait when a child will do setsid() and other actions. Did I miss something?

test/others/loop.c
```diff
@@ -0,0 +1,40 @@
...
+#include <stdio.h>
+
+int main() {
+	pid_t pid;
+	pid_t sid;
+
+	pid = fork();
+	if (pid < 0) {
+		perror("fork failed!");
+		return -1;
```
avagin, line 15: return 1;

rst0git: Fixed

test/others/loop.c
rst0git: I've updated the commit to wait for the child process to complete the setup.

test/others/loop.c
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	if (pid == 0) {
+		close(start_pipe[0]);
+
+		sid = setsid();
+		if(sid < 0) {
+			perror("setsid failed!");
+			res = EXIT_FAILURE;
+			write(start_pipe[1], &res, 1);
```
avagin, line 32: write(..., sizeof(res))

test/others/rpc/Makefile
```diff
@@ -7,6 +7,7 @@ LDLIBS +=  -lprotobuf-c
 PYTHON ?= python
 
 run: all
+	@make -C ..
```
avagin, line 10: nit: make -C loop

test/others/loop.c
rst0git: Fixed

test/others/rpc/Makefile
rst0git: Updated

Issue #1503, opened 2021-06-12
CI failures with new github actions kernel


adrianreber:
It seems with the updated kernel (5.8.0-1033-azure) we are seeing new CI failures:

CRIU Image Streamer test fails with:
```
2021-06-12T12:37:15.8856608Z === Run 41/41 ================ zdtm/transition/fifo_loop
2021-06-12T12:37:15.8857315Z timens isn't supported on 5.8.0-1033-azure
2021-06-12T12:37:15.8857917Z ====================== Run zdtm/transition/fifo_loop in h ======================
2021-06-12T12:37:15.8858424Z Start test
2021-06-12T12:37:15.8859283Z ./fifo_loop --pidfile=fifo_loop.pid --outfile=fifo_loop.out --filename=fifo_loop.test
2021-06-12T12:37:15.8860256Z Run criu-image-streamer in capture mode
2021-06-12T12:37:15.8860790Z Run criu dump
2021-06-12T12:37:15.8861479Z Run criu-image-streamer in extract mode
2021-06-12T12:37:15.8862303Z Run criu-image-streamer in serve mode
2021-06-12T12:37:15.8862857Z Run criu restore
2021-06-12T12:37:15.8863403Z =[log]=> dump/zdtm/transition/fifo_loop/50/1/restore.log
2021-06-12T12:37:15.8864254Z ------------------------ grep Error ------------------------
2021-06-12T12:37:15.8866472Z b'(00.003388)     50: Collecting 42/27 (flags 0)'
2021-06-12T12:37:15.8867208Z b'(00.003417)     50: No pipes-data.img image'
2021-06-12T12:37:15.8867861Z b'(00.003421)     50:  `- ... done'
2021-06-12T12:37:15.8868478Z b'(00.003423)     50: Collecting 43/27 (flags 0)'
2021-06-12T12:37:15.8869249Z b'(00.004231)     50: Error (criu/image.c:694): Img trimmed 1048552/1048576'
2021-06-12T12:37:15.8869889Z b'(00.004249)     50:  `- ... done'
2021-06-12T12:37:15.8870638Z b'(00.004269) Error (criu/cr-restore.c:2514): Restoring FAILED.'
2021-06-12T12:37:15.8871478Z ------------------------ ERROR OVER ------------------------
2021-06-12T12:37:15.8872191Z ############# Test zdtm/transition/fifo_loop FAIL at CRIU restore ##############
```
@nviennot do you have any idea how to fix this?

[...]

PR #1506, opened 2021-06-13
ci: move CentOS 7 tests to Cirrus CI

adrianreber:
On CI systems using newer kernels with clone3() and old user space which does PID caching (like CentOS 7) the getpid() call returns the wrong PID as it does not know how clone3() assigns PIDs.

Avoiding glibc and doing the getpid() syscall directly fixes PID caching errors on the combination of new kernel and very old user space.

This tries to fix one of the errors described in #1503


adrianreber:
@avagin: This tries to fix the clone3() error which can be seen on CentOS7. But once the getpid() returns the right value the tests are failing with time namespace problems:
```
(00.000416) Forking task with 52 pid (flags 0x0)
userns is supported
userns is supported
(00.000421) Creating process using clone3()
(00.000532) PID: real 52 virt 52
(00.000775)     52: timens: monotonic -1 989797667
(00.000803)     52: Error (criu/timens.c:99): Unable to set a monotonic clock offset: Permission denied
(00.000846) Error (criu/cr-restore.c:2521): Restoring FAILED.
(00.000963) Error (criu/cr-restore.c:1537): 52 killed by signal 9: Killed
```
I can reproduce this error locally running a CentOS 7 container on a 5.11 kernel. Any ideas how solve this error?

adrianreber:
Thanks to @avagin's suggestion I switched all `getpid()` calls to avoid the glibc and locally the `timens` problems disappeared.

Let's see if CI also likes it.

adrianreber:
This PR is now completely changed. Instead of changing code I moved the CentOS 7 based CI from GitHub to Cirrus CI.

On Cirrus CI we can on a CentOS 7 VM and not like on GitHub on 5.8 kernel CentOS user space which did not really work and there is no need to support such an unusual combination.

adrianreber:
@checkpoint-restore/maintainers PTAL

adrianreber:
If there are no objections I would like to get this merged to remove the failing CentOS 7 test from our CI runs.

scripts/ci/run-ci-tests.sh
```diff
@@ -237,7 +218,7 @@ bash -x ./test/jenkins/criu-inhfd.sh
 if [ -z "$SKIP_EXT_DEV_TEST" ]; then
 	make -C test/others/mnt-ext-dev/ run
 fi
-#make -C test/others/exec/ run
```
Snorch, line 240: Heh it looks like `make -C test/others/exec/ run` is broken long ago...
```
+ ../../../criu/criu exec -t 186212 fake_syscall
...
Error (criu/crtools.c:162): excessive parameter for command exec
```
As criu exec is already deprecated.

Snorch (approved):
I'm not too good at CI. Though looks good to me.

scripts/ci/run-ci-tests.sh
adrianreber: Good point. Maybe we should just remove that directory.

adrianreber: Let me open a PR to remove that directory.

PR #1509, opened 2021-06-13
ci: no longer avoid overlayfs

adrianreber:
Now that the Ubuntu kernel is no longer broken with regards to overlayfs, let's switch back to overlayfs instead of devicemapper and
vfs graphdrivers.

### DIFF
--- a/scripts/ci/docker-test.sh
+++ b/scripts/ci/docker-test.sh
@@ -21,10 +21,7 @@ add-apt-repository \
 
 . /etc/lsb-release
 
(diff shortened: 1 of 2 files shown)


adrianreber:
The CI failures are documented in #1503 but the changes to docker-test and podman-test did not lead to any failures.

This means the Ubuntu kernel has really been fixed. Thanks @mihalicyn for getting the necessary fix into the Ubuntu kernel!

rst0git (approved):

PR #1514, opened 2021-06-16
Make config file parser more robust

adrianreber:
Trying to see how robust the configuration parser I was able to crash CRIU pretty quickly. This fixes a few crashes in the existing configuration file parser.

This also adds tests for not tested option parsing and configuration file handling.

### DIFF
--- a/criu/config.c
+++ b/criu/config.c
@@ -44,72 +44,208 @@ static int count_elements(char **to_count)
 /* Parse one statement in configuration file */
 int parse_statement(int i, char *line, char **configuration)
 {
+	cleanup_free char *input = NULL;
 	int offset = 0, len = 0;
-	bool was_newline = true;
-	char *tmp_string, *quoted, *quotedptr;
+	char *tmp_string;
 
-	while (1) {
-		/* Ignore white-space */
-		while ((isspace(*(line + offset)) && (*(line + offset) != '\n'))) offset++;
+	/*
+	 * A line from the configuration file can be:
+	 *  - empty
+	 *  - a boolean option (tcp-close)
+	 *  - an option with one parameter (verbosity 4)
+	 *    - a parameter can be in quotes (lsm-profile "selinux:something")
+	 *    - a parameter can contain escaped quotes
+	 *
+	 * Whenever a '#' is found we ignore everything after as a comment.
+	 *
+	 * This function adds none, one (boolean option) or two entries
+	 * in **configuration and returns i + (the number of entries).
+	 */
 
-		/* Read a single word. A word is everything
-		 * that doesn't contain white-space characters. */
-		if (sscanf(line + offset, "%m[^ \t\n]s", &configuration[i]) != 1) {
-			configuration[i] = NULL;
-			break;
-		}
+	if (strlen(line) == 0)
+		return i;
 
-		/* Ignore comments - everything between '#' and '\n' */
-		if (configuration[i][0] == '#') {
-			configuration[i] = NULL;
-			break;
-		}
+	/* Ignore leading white-space */
+	while ((isspace(*(line + offset)) && (*(line + offset) != '\n')))
+		offset++;
 
-		if ((configuration[i][0] == '\"') && (strchr(line + offset + 1, '"'))) {
-			/* Handle empty strings which strtok ignores */
-			if (!strcmp(configuration[i], "\"\"")) {
-				configuration[i] = "";
-				offset += strlen("\"\"");
-			} else if ((configuration[i] = strtok_r(line + offset, "\"", &quotedptr))) {
-				/* Handle escaping of quotes in quoted string */
-				while (configuration[i][strlen(configuration[i]) - 1] == '\\') {
-					offset++;
-					len = strlen(configuration[i]);
-					configuration[i][len - 1] = '"';
-					if (*quotedptr == '"') {
-						quotedptr++;
-						break;
-					}
-					quoted = strtok_r(NULL, "\"", &quotedptr);
-					tmp_string = xmalloc(len + strlen(quoted) + 1);
-					if (tmp_string == NULL)
-						return -1;
+	/* Ignore empty line */
+	if (line[offset] == '\n')
+		return i;
 
-					memmove(tmp_string, configuration[i], len);
-					memmove(tmp_string + len, quoted, strlen(quoted) + 1);
-					configuration[i] = tmp_string;
-				}
-				offset += 2;
-			}
-		}
+	/* Ignore line starting with a comment */
+	if (line[offset] == '#')
+		return i;
 
-		offset += strlen(configuration[i]);
+	input = xstrdup(line + offset);
+	if (unlikely(!input))
+		return -1;
 
-		if (was_newline) {
-			was_newline = false;
-			len = strlen(configuration[i]);
-			tmp_string = xrealloc(configuration[i], len + strlen("--") + 1);
-			if (tmp_string == NULL)
-				return -1;
+	offset = 0;
 
-			memmove(tmp_string + strlen("--"), tmp_string, len + 1);
-			memmove(tmp_string, "--", strlen("--"));
-			configuration[i] = tmp_string;
-		}
+	/* Remove trailing '\n' */
+	if ((tmp_string = strchr(input, '\n')))
+		tmp_string[0] = 0;
+
+	if ((tmp_string = strchr(input, ' ')) || (tmp_string = strchr(input, '\t'))) {
+		configuration[i] = xzalloc(tmp_string - input + strlen("--") + 1);
+		if (unlikely(!configuration[i]))
+			return -1;
+		memcpy(configuration[i], "--", strlen("--"));
+		memcpy(configuration[i] + strlen("--"), input, tmp_string - input);
+		configuration[i][tmp_string - input + strlen("--")] = 0;
+		/* Go to the next character */
+		offset += tmp_string - input + 1;
 		i++;
+	} else {
+		if (unlikely(asprintf(&configuration[i], "--%s", input) == -1))
+			return -1;
+		return i + 1;
 	}
 
-	return i;
+	while ((isspace(*(input + offset))))
+		offset++;
+
+	/* Check if the next token is a comment */
+	if (input[offset] == '#')
+		return i;
+
+	if (input[offset] == '"') {
+		bool found_second_quote = false;
+		char *quote_start;
+		int quote_offset;
+
+		/* Move by one to skip the leading quote. */
+		offset++;
+		quote_start = input + offset;
+		quote_offset = offset;
+
+		if (input[offset] == 0) {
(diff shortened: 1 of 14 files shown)


0x7f454c46 (approved):
Nice! LGTM

criu/config.c
```diff
@@ -46,53 +46,57 @@ int parse_statement(int i, char *line, char **configuration)
...
+					skipped++;
+					continue;
 				}
-				offset += 2;
+				tmp_string[dest++] = configuration[i][j];
 			}
+			memcpy(configuration[i], tmp_string, strlen(tmp_string));
+			configuration[i][strlen(tmp_string)] = 0;
+			offset += skipped + 1;
+			free(tmp_string);
```
avagin, line 92: xfree

avagin:
```
diff --git a/test/others/config-file/conf.test b/test/others/config-file/conf.test
index 1b9970bcf..2051137ba 100644
--- a/test/others/config-file/conf.test
+++ b/test/others/config-file/conf.test
@@ -1,6 +1,7 @@
 #
 #
 #
+a s d f g h j k l o o o o o o
 verbosity 4
 lsm-profile "" # more comments
 lsm-profile "something"# comment
[avagin@laptop criu]$  ./criu/criu --config test/others/config-file/conf.test 
double free or corruption (out)
Aborted (core dumped)
```

criu/config.c
```diff
@@ -46,53 +46,57 @@ int parse_statement(int i, char *line, char **configuration)
...
 			configuration[i] = NULL;
 			break;
 		}
 
-		/* Ignore comments - everything between '#' and '\n' */
-		if (configuration[i][0] == '#') {
-			configuration[i] = NULL;
-			break;
+		len = strlen(configuration[i]);
+		if (configuration[i][0] == '"') {
```
avagin, line 67: if an argument starts with the quotation mark, it means that it can contain spaces. configuration contains only one word without spaces. You need to find the next quotation mark and copy everything that is between these two marks. Have I missed something?

test/others/config-file/run.sh
```diff
@@ -0,0 +1,91 @@
...
+
+if [ ! -d /etc/criu ]; then
+	mkdir -p /etc/criu
+fi
+
+if [ ! -e /etc/criu/default.conf ]; then
+	touch /etc/criu/default.conf
+fi
+
+# This tries to capture any exit codes other than 0 and 1
```
avagin, line 25: We probably need to think about how to implement unit tests that will check correctness.

criu/config.c
adrianreber: done

adrianreber:
> ```
> diff --git a/test/others/config-file/conf.test b/test/others/config-file/conf.test
> index 1b9970bcf..2051137ba 100644
> --- a/test/others/config-file/conf.test
> +++ b/test/others/config-file/conf.test
> @@ -1,6 +1,7 @@
>  #
>  #
>  #
> +a s d f g h j k l o o o o o o
>  verbosity 4
>  lsm-profile "" # more comments
>  lsm-profile "something"# comment
> [avagin@laptop criu]$  ./criu/criu --config test/others/config-file/conf.test 
> double free or corruption (out)
> Aborted (core dumped)
> ```

Fixed.

criu/config.c
adrianreber: No, you have not missed something. You are right. This problem exists in the parser since the beginning and should be fixed. This PR, however, was just trying to avoid crashes. I think this should be fixed, but not in this PR.

test/others/config-file/run.sh
adrianreber: Yes, I was just looking at code coverage to see which part of the code was untested and tried to use all those options which code path was untested.

criu/config.c
avagin: Could you file an issue and add "// FIXME(issue/XXX): description" in the code

criu/config.c
```diff
@@ -139,9 +143,22 @@ static char ** parse_config(char *filepath)
...
+		 */
+		for (j = 0; j < strlen(line); j++)
+			if (line[j] == ' ')
+				spaces++;
+
 		/* Extend configuration buffer if necessary */
-		if (i >= config_size - 1) {
-			config_size *= 2;
+		if (i + spaces >= config_size - 1) {
+			config_size *= spaces * 2;
```
avagin, line 161: hmmm. What is going on here?

criu/config.c
adrianreber: I started to write the issue and then I just fixed it. Spaces should be now handled.

criu/config.c
adrianreber: Oops. That is not right and fixed now.

test/others/config-file/run.sh
adrianreber: Do you have experience with unit test C code? Any frameworks you know?

ivalery111: > Do you have experience with unit test C code? Any frameworks you know?

Hey,

Maybe the [CMocka](https://cmocka.org/) will useful?

criu/config.c
avagin: I don't see how you fix it and my experiments show that it doesn't work...

avagin: Do we describe the config format somewhere? Can we suppose that the first word is an option name and a rest of the line is an option argument?

test/others/config-file/run.sh
avagin: Let's try without any frameworks and then we will see whether we need something or not. You can create a test binary that uses our config parser. It can read configs from a specified directory, parse them and check that results are correct.

criu/config.c
adrianreber: > I don't see how you fix it and my experiments show that it doesn't work...

It worked with my inputs. What do you use as input?

avagin:
```
$ cat test/others/config-file/conf.test
dsfa "aaaaa \"bbbbbb\""
```

I really recommend to implement the test for the parser. Otherwise it will never work reliably...

test/others/config-file/run.sh
adrianreber: I tried to create a unit test. Please have a look. Our code is very interdependent so it is not easy to take out small parts for unit testing, but it worked for the configuration file parser.

criu/config.c
adrianreber: implemented a unit test

avagin:
@0x7f454c46 could you review makefile changes?

adrianreber:
> @0x7f454c46 could you review makefile changes?

Yes, please, because I am not sure I did it correctly.

criu/unittest/unit.c
```diff
@@ -0,0 +1,74 @@
...
+
+	i = parse_statement(0, "work-dir \"/tmp with spaces\"\n", configuration);
+	assert(i == 2);
+	assert(!strcmp(configuration[0], "--work-dir"));
+	assert(!strcmp(configuration[1], "/tmp with spaces"));
+
+	i = parse_statement(0, "a b c d e f g h i\n", configuration);
+	assert(i == 9);
+	assert(!strcmp(configuration[0], "--a"));
+	assert(!strcmp(configuration[1], "b"));
```
avagin, line 63: I still don't understand an use-case for this. Do we have arguments that can accept multiple values?

I mean --arg-list val1 val2 val3 val3. If we are not, we can simplify the code a lot. We need to get the first word and it will be an argument and other part of the line will be a value. We will not need to handle quotas...

adrianreber: You are right. I blindly copied your example, but we do not need to handle it. This can be simplified.

adrianreber: The current behaviour of the parser is documented at the end of the man page. There is says to use quotes if the value behind the option contains whitespaces. So we probably want to keep that functionality. Also we support comments after the value to an option `lsm-profile selinux:something #comment`. So we need to scan for whitespace to see if there is a comment at the end. Not sure if we still can simplify the code and keep the existing functionality.

0x7f454c46:
> > @0x7f454c46 could you review makefile changes?
> 
> Yes, please, because I am not sure I did it correctly.

Well, I may tell you they look good to me.
There are a couple of small things, but not sure they matter much:
- you could add `$(obj)/unittest/%: .FORCE` rule in case someone wants to do `make criu/unittest/unit.o` (not sure if anyone wants that)
- some flags like `GMONLDOPT` make little sense to me, but on the other hand, do no harm

Yeah, I think you did quite fine :-)

criu/config.c
```diff
@@ -46,53 +46,89 @@ int parse_statement(int i, char *line, char **configuration)
...
+			 * next unescaped quote needs to be looked at.
+			 */
+			while ((tmp_string = strchr(line + quote_offset + 1, '"'))) {
+				quote_offset += tmp_string - (line + quote_offset);
+				/* Check if it is escaped */
+				if (*(tmp_string - 1) == '\\')
+					continue;
+
+				/* Not escaped. That is the end of the quoted string. */
+				configuration[i] = xrealloc(configuration[i], quote_offset - offset + 1);
```
avagin, line 87: coverity will not like this. It will tell you that here can be a memory leak.

new = xrealloc(configuration[i]configuration[i], quote_offset - offset + 1);
if (new == NULL) {
  ... fail
}
configuration[i] = new;

criu/config.c
```diff
@@ -46,53 +46,89 @@ int parse_statement(int i, char *line, char **configuration)
...
+			memmove(configuration[i], configuration[i] + 1, len - 1);
+			offset++;
+			quote_start = line + offset;
+			quote_offset = offset;
+
+			/*
+			 * If it starts with a quote everything until the
+			 * next unescaped quote needs to be looked at.
+			 */
+			while ((tmp_string = strchr(line + quote_offset + 1, '"'))) {
```
avagin, line 80: this code has to return an error if it fails to find the second  unescaped quote.

criu/config.c
```diff
@@ -46,53 +46,89 @@ int parse_statement(int i, char *line, char **configuration)
...
+				quote_offset += tmp_string - (line + quote_offset);
```
avagin, line 81: quote_offset = tmp_string - line;

adrianreber:
I just pushed a completely different version.

adrianreber:
And it fails spectacularly.

adrianreber:
All review comments should be addressed (I hope) and CI is happy again.

criu/config.c
```diff
@@ -44,72 +44,208 @@ static int count_elements(char **to_count)
...
+		return -1;
+
+	offset = 0;
+
+	/* Remove trailing '\n' */
+	if ((tmp_string = strchr(input, '\n')))
+		tmp_string[0] = 0;
+
+	if ((tmp_string = strchr(input, ' ')) || (tmp_string = strchr(input, '\t'))) {
+		configuration[i] = xzalloc(tmp_string - input + strlen("--") + 1);
```
avagin, line 91: xmalloc

PR #1515, opened 2021-06-16
ci: update docker test matrix

rst0git:
Ubuntu 18.04 still has a problem with overlayfs and breaks CRIU
https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257


adrianreber:
Good idea. We know that it works with the test on 20.04. Not really necessary to test it on multiple distributions from out side.

All other failures are known and are being worked on in other PRs.

PR #1518, opened 2021-06-17
docker-test: use latest containerd release

rst0git:
This patch improves the changes from 19be9ced9. To use the newer version of containerd, we need to make sure that the containerd service has been restarted after install. Instead of hard-coding a version number, we can use github API to get the latest release. In addition, the tar file contains all binary files in a `./bin` sub-folder. Thus, it should be extracted in `/usr`.


adrianreber (approved):
Makes sense. We are also tracking latest release of Podman.

PR #1520, opened 2021-06-18
bfd: loop through read()/write() when the action is incomplete

### LINKED ISSUE #1504: image streamer test fail
The following test sometimes fails with `Error (criu/image.c:694): Img trimmed 1048551/1048576` ([restore.log](https://github.com/checkpoint-restore/criu/files/6642277/restore.log)).

```
./test/zdtm.py run --stream -p2 -t zdtm/transition/fifo_loop
```
https://github.com/checkpoint-restore/criu/runs/2809666897

nviennot:
The callers of bread() and bwrite() assume the operation reads/writes
the complete length of the passed buffer.
We must loop when invoking the read()/write() APIs.

Fixes #1504

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -1584,3 +1584,59 @@ char *get_legacy_iptables_bin(bool ipv6)
 
 	return iptables_bin[ipv6];
 }
+
+/*
+ * read_all() behaves like read() without the possibility of partial reads.
+ * Use only with blocking I/O.
+ */
+ssize_t read_all(int fd, void *buf, size_t size)
+{
+	ssize_t n = 0;
+	while (size > 0) {
+		ssize_t ret = read(fd, buf, size);
+		if (ret == -1) {
+			if (errno == EINTR)
+				continue;
+			/*
+			 * The caller should use standard read() for
+			 * non-blocking I/O.
+			 */
+			if (errno == EAGAIN || errno == EWOULDBLOCK)
+				errno = EINVAL;
+			return ret;
+		}
+		if (ret == 0)
+			break;
+		n += ret;
+		buf = (char *)buf + ret;
+		size -= ret;
+	}
+	return n;
+}
+
+/*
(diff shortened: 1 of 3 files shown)


adrianreber (approved):
Thanks. Looks good. It fixes the error and sounds reasonable.

adrianreber:
@checkpoint-restore/maintainers PTAL. This fixes one of the currently existing CI errors.

criu/util.c
```diff
@@ -1583,4 +1583,38 @@ char *get_legacy_iptables_bin(bool ipv6)
...
+	return n;
+}
+
+ssize_t write_all(int fd, const void *buf, size_t size)
+{
+	ssize_t n = 0;
+	while (n < size) {
+		ssize_t ret = write(fd, buf, size - n);
+		if (ret == -1) {
+			if (errno == EINTR)
```
rst0git, line 1612: In addition to `EINTR`, should we also check for `errno == EAGAIN`?

nviennot: So what happens when we've already made some progress, and then we get EAGAIN.

Should we return the number of bytes already written? in which case, this is not that better than the original write()

Should we return a EAGAIN? But then we've already written some content, and the caller will need to know about it. So that's bad. Can't do that.

In other words, if we can get EAGAIN, the caller is just better off using regular read()/write(). Which means that the read_all()/write_all() APIs are for blocking fds only.
In this case, if we get EAGAIN maybe it would be appropriate to return EINVAL? and add a comment on these functions that these APIs are for blocking fds only.
What do you think?

rst0git: >In this case, if we get EAGAIN maybe it would be appropriate to return EINVAL? and add a comment on these functions that these APIs are for blocking fds only.

Sounds good. read_all() and write_all() are generic functions and it would be good to keep them simple.

nviennot: I've updated the PR

PR #1521, opened 2021-06-18
add PKG_CONFIG default in a few more places

vapier:
These files use $PKG_CONFIG before they include the common files that
setup a default, so set early defaults in them too.

Signed-off-by: Mike Frysinger <vapier@chromium.org>


adrianreber:
Thanks. Looks good. That is also what I would have done.

PR #1525, opened 2021-06-23
Keep inherit-fd strings alive until task restore

ymanton:
If inherit-fd is read from a config file its buffer will be freed after the config file is parsed but before task restore, which is when we need to use the mapping. Therefore, when adding an inherit-fd mapping to the opts list, copy the key string to a new buffer and free all such buffers after task restore.

### DIFF
--- a/criu/files.c
+++ b/criu/files.c
@@ -1568,7 +1568,10 @@ int inherit_fd_add(int fd, char *key)
 	if (fd > inh_fd_max)
 		inh_fd_max = fd;
 
-	inh->inh_id = key;
+	inh->inh_id = xstrdup(key);
+	if (inh->inh_id == NULL)
+		return -1;
+
 	inh->inh_fd = fd;
 	list_add_tail(&inh->inh_list, &opts.inherit_fds);
 	return 0;


criu/files.c
```diff
@@ -1568,12 +1568,22 @@ int inherit_fd_add(int fd, char *key)
 	if (fd > inh_fd_max)
 		inh_fd_max = fd;
 
-	inh->inh_id = key;
+	inh->inh_id = xstrdup(key);
```
adrianreber, line 1571: `xstrdup()` can return NULL. Please add a check if it fails here.

ymanton: Thanks, check added.

adrianreber (approved):

adrianreber:
@checkpoint-restore/maintainers please review

rst0git (approved):

criu/files.c
```diff
@@ -1568,12 +1568,25 @@ int inherit_fd_add(int fd, char *key)
...
 	list_add_tail(&inh->inh_list, &opts.inherit_fds);
 	return 0;
 }
 
+void inherit_fd_clean(void)
+{
+	struct inherit_fd *inh;
+
+	list_for_each_entry(inh, &opts.inherit_fds, inh_list) {
+		free(inh->inh_id);
```
avagin, line 1585: xfree

avagin:
> If inherit-fd is read from a config file its buffer will be freed after the config file is parsed but before task restore

maybe we don't need to free its buffer. There are probably a few other arguments that affected by this issue.

adrianreber:
> > If inherit-fd is read from a config file its buffer will be freed after the config file is parsed but before task restore
> 
> maybe we don't need to free its buffer. There are probably a few other arguments that affected by this issue.

I think this PR is correct. Without config files it would not be necessary as the buffer was directly coming from optarg. With configuration files we need to copy the buffer because the memory used by the configuration file parser is freed after going through getopt() as we have multiple configuration sources and did not want to keep all possible configuration buffers alive.

avagin:
> I think this PR is correct. Without config files, it would not be necessary as the buffer was directly coming from optarg. With configuration files, we need to copy the buffer because the memory used by the configuration file parser is freed after going through getopt() as we have multiple configuration sources and did not want to keep all possible configuration buffers alive.

This is counterintuitive...

> int parse_options(int argc, char **argv, bool *usage_error,
>                bool *has_exec_cmd, int state)

If I will change this function, I will think that I'm working with argv that gives me some assumption on a lifetime of its elements.

If we want to free memory after going through getopt(), we need to rename argv to something else and write a warning in the comment.

And I think cgroup-root, action-script, libdir, external options are affected too.

adrianreber:
> > I think this PR is correct. Without config files, it would not be necessary as the buffer was directly coming from optarg. With configuration files, we need to copy the buffer because the memory used by the configuration file parser is freed after going through getopt() as we have multiple configuration sources and did not want to keep all possible configuration buffers alive.
> 
> This is counterintuitive...
> 
> > int parse_options(int argc, char **argv, bool *usage_error,
> > bool *has_exec_cmd, int state)
> 
> If I will change this function, I will think that I'm working with argv that gives me some assumption on a lifetime of its elements.
> 
> If we want to free memory after going through getopt(), we need to rename argv to something else and write a warning in the comment.

But that is independent of this PR, right? We can still merge this fix.

> And I think cgroup-root, action-script, libdir, external options are affected too.

Probably yes.

avagin:
> But that is independent of this PR, right? We can still merge this fix.

Why is it independent? We found an issue, we found the root cause, we found that there are more options that are affected. What should we do next? Right, we need to fix the issue for all options, need to fix the root cause, and make the next release.

criu/files.c
ymanton: Fixed.

ymanton:
I agree, holding all config files in memory will eliminate this sort of problem not only for current options but also any added in the future. I have some patches for that, shall I open another PR or just add them here? They're independent of the current patch.

adrianreber:
> > But that is independent of this PR, right? We can still merge this fix.
> 
> Why is it independent? We found an issue, we found the root cause, we found that there are more options that are affected. What should we do next? Right, we need to fix the issue for all options, need to fix the root cause, and make the next release.

Yes, it needs to be fixed. But this PR directly fixes an issue. Not sure if you do not want to merge this PR at all because it is the wrong fix and want the complete solution or if you want to merge this PR and fix the current problem and additionally have a fix for the bigger problem.

adrianreber (approved):

criu/files.c
```diff
@@ -1568,12 +1568,25 @@ int inherit_fd_add(int fd, char *key)
...
 	return 0;
 }
 
+void inherit_fd_clean(void)
+{
+	struct inherit_fd *inh;
+
+	list_for_each_entry(inh, &opts.inherit_fds, inh_list) {
+		xfree(inh->inh_id);
+		inh->inh_id = NULL;
```
avagin, line 1586: xfree(inh) and you will need to use list_for_each_entry_safe.

I think we don't need inherit_fd_clean, CRIU usually exits right after restoring processes.

criu/crtools.c
```diff
@@ -255,6 +255,7 @@ int main(int argc, char *argv[], char *envp[])
 			pr_warn("Using -t with criu restore is obsoleted\n");
 
 		ret = cr_restore_tasks();
+		inherit_fd_clean();
```
avagin, line 258: I think we don't need inherit_fd_clean, but if you want to have it, it will be better to call it from cr_restore_tasks.

criu/files.c
ymanton: > I think we don't need inherit_fd_clean, CRIU usually exits right after restoring processes.

Agree, I've removed it. What's left of the patch is more or less trivial.

criu/crtools.c
ymanton: Removed it altogether.

PR #1526, opened 2021-06-23
get rid of using glibc's getpid

liusdu:
Sometimes getpid() returns parents pid when process cloned by clone3().
Error message is as following:
```
(00.002797) No pidns-9.img image
(00.002844) Warn (criu/cr-restore.c:1329): Set CLONE_PARENT | CLONE_NEWPID but it might cause restore problem,because not all kernels support such clone flags combinations!
(00.002846) Forking task with 1 pid (flags 0x6c028000)
(00.002848) Creating process using clone3()

(00.003321) PID: real 4124639 virt 1
(00.003413) Wait until namespaces are created
(00.003412) Error (criu/cr-restore.c:1816): Pid 4124632 do not match expected 1

(00.004503) Error (criu/cr-restore.c:2514): Restoring FAILED.
```

Man page of clone3 shows specific versions of glibc has a pid-cacheing behavior, which
may be not up to date. So get rid of using glibc's getpid

### DIFF
--- a/criu/action-scripts.c
+++ b/criu/action-scripts.c
@@ -62,7 +62,7 @@ static int run_shell_scripts(const char *action)
 
 	if (!(env_set & ENV_IMGDIR)) {
 		char image_dir[PATH_MAX];
-		sprintf(image_dir, "/proc/%ld/fd/%d", (long) getpid(), get_service_fd(IMG_FD_OFF));
+		sprintf(image_dir, "/proc/%ld/fd/%d", (long) syscall(__NR_getpid), get_service_fd(IMG_FD_OFF));
 		if (setenv("CRTOOLS_IMAGE_DIR", image_dir, 1)) {
 			pr_perror("Can't set CRTOOLS_IMAGE_DIR=%s", image_dir);
 			return -1;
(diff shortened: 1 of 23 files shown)


adrianreber:
Thanks for the pull request. Which combination of glibc and kernel are you using. We have actually seen this in CI when using a new kernel with a very old (CentOS 7) userspace and decided it is not really worth supporting this strange combination.

@avagin also had a much simpler fix:
```
diff --git a/criu/cr-restore.c b/criu/cr-restore.c
index 0d85db6..46a183e 100644
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -1782,6 +1782,11 @@ static int create_children_and_session(void)
        return 0;
 }

+pid_t getpid(void)
+{
+       return syscall(__NR_getpid);
+}
+
 static int restore_task_with_children(void *_arg)
 {
        struct cr_clone_arg *ca = _arg;
```

Would that change also work for you?

liusdu:
@adrianreber yes, this works for me.

liusdu:
closed, I will fix it internal

PR #1528, opened 2021-06-24
pidfd_store: move to separate file and tidy its api

ZeyadYasser:
This is a follow up to https://github.com/checkpoint-restore/criu/pull/1341 as requested by @avagin.

I created a wiki article https://criu.org/Pidfd_store.

I added an entry for the pidfd_store_sk in the options table https://criu.org/API_compliance.

Also I polished the code base: 
- Moved pidfd store related code to a separate file.
- Tidied up its api, hiding unneeded details.
- Moved pidfd store hashtable initialization to cr-service.c since pidfd store is only usable in RPC mode.


avagin:
```
- Move pidfd store related code to a separate file.
- Tidy up its api, hiding uneeded details.
- Move pidfd store hashtable initialization to cr-service.c
  Since pidfd store is only usable in RPC mode.
```

This looks like three separate commits ;)

avagin (approved):

criu/pidfd-store.c
```diff
@@ -0,0 +1,326 @@
...
+		return 0;
+
+	/*
+	 * Drain all queued pidfd entries in pidfd_store_sk from
+	 * the last predump into pidfd_hash.
+	 */
+	cnt = 0;
+	while (1) {
+		entry = xmalloc(sizeof(struct pidfd_entry));
+		if (entry == NULL) {
```
adrianreber, line 161: remove braces

criu/pidfd-store.c
```diff
@@ -0,0 +1,326 @@
...
+		addrlen += sizeof(addr.sun_family);
+
+		addr.sun_path[0] = 0;
+
+		if (bind(pidfd_store_sk, (struct sockaddr *)&addr, addrlen)) {
+			pr_perror("Unable to bind a socket");
+			goto err;
+		}
+
+		if (connect(pidfd_store_sk, (struct sockaddr *) &addr, addrlen)) {
```
adrianreber, line 113: Unnecessary space before `&addr`.

criu/include/pidfd-store.h
```diff
@@ -0,0 +1,13 @@
...
+#include <sys/types.h>
+
+int init_pidfd_store_sk(pid_t pid, int fd);
+int init_pidfd_store_hash(void);
+void free_pidfd_store(void);
+int pidfd_store_add(pid_t pid);
+int pidfd_store_check_pid_reuse(pid_t pid);
+bool pidfd_store_ready(void);
+
+#endif /* __CR_PIDFD_STORE_H__ */
```
adrianreber, line 13: Our files usually end with a newline. That is missing here. I think this might be connected to some editor settings, but github also highlights the missing newline. Not important. Nice for consistency.

adrianreber (changes requested):

criu/pidfd-store.c
```diff
@@ -0,0 +1,326 @@
...
+	}
+
+	close_safe(&pidfd_store_sk);
+	pidfd_store_sk = syscall(SYS_pidfd_getfd, pidfd, sk, 0);
+	if (pidfd_store_sk == -1) {
+		pr_perror("Can't steal fd %d using pidfd_getfd", sk);
+		close(pidfd);
+		goto err;
+	}
+	close(pidfd);
```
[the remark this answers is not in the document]
ZeyadYasser: Couldn't the close overwrite the errno of pidfd_getfd?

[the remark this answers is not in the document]
adrianreber: You are right.

adrianreber (approved):

criu/cr-service.c
```diff
@@ -737,6 +739,8 @@ static int dump_using_req(int sk, CriuOpts *req)
 
 	success = true;
 exit:
+	free_pidfd_store();
+pidfd_store_err:
```
Snorch, line 743: If init_pidfd_store_hash() fails you should also call free_pidfd_store() anyway.

Because init_pidfd_store_hash() can fail on some non-frist iteration in while on xmalloc when it already filled the hash with some number of entries..

Actually before this patchset we were calling free_pidfd_store always even

Snorch (changes requested):
In general It looks good. Please fix hash memory leak on error paths.

criu/cr-service.c
ZeyadYasser: Thanks, I didn't think of the xmalloc case before.

To avoid a redundant call to free_pidfd_store I can think of two ways:
1. I could add a ```free_pidfd_store``` if xmalloc fails and keep ```pidfd_store_err:``` label as it is.
2. I could remove ```free_pidfd_store``` from ```init_pidfd_store_hash``` and always call it after ```init_pidfd_store_hash```.

I am inclined towards the second option.

ZeyadYasser: I went the first option because I think ```init_pidfd_store_hash()``` should clean its failure, any feedback?

Snorch (approved):
Looks good, Thanks!

Issue #1530, opened 2021-06-27
zdtm: netns_lock test unexpected CI behavior


ZeyadYasser:
I created a test for network internal locking which was recently merged: https://github.com/checkpoint-restore/criu/pull/1465

For some reason this test doesn't always pass on CentOS7:
- This is running and failing https://api.cirrus-ci.com/v1/task/6633582350303232/logs/build.log
- Here I just reran the test and it passed https://api.cirrus-ci.com/v1/task/6302001949573120/logs/build.log

Also here https://ci.kernoops.org/job/CRIU/job/CRIU-iter/job/criu-dev/431/testReport/(root)/criu/zdtm_static_netns_lock/ the test failure is interesting because it states that the --pre-dump hook is being run again after restore which expectedly causes the test to fail as it is highly dependent on the order of the hooks

I just wanted to document this weird behavior in an open issue until I figure out what is wrong.

adrianreber:
Thanks for opening the ticket to track this. I also saw this and was confused why it fails sometimes. Good that you are aware of it.

ZeyadYasser:
@adrianreber  I think I figured out the cause of random failure in CentOS7.

I believe this is caused by a race condition where the following execution flow takes place:
1. ```--post-start``` hook starts just after the test (both are run in parallel) https://github.com/checkpoint-restore/criu/blob/98186981d9eb430e39da3ee692e41ccdcfa67bff/test/zdtm.py#L426
2. ```--post-start``` hook calls ```ip netns add``` to create the test netns https://github.com/checkpoint-restore/criu/blob/98186981d9eb430e39da3ee692e41ccdcfa67bff/test/zdtm/static/netns_lock.hook#L25
3. ```ip``` creates the netns file https://github.com/shemminger/iproute2/blob/9f366536edb5158343152604e82b968be46dbf26/ip/ipnetns.c#L888
4. ```netns_lock``` test opens that file and uses it in ```setns``` https://github.com/checkpoint-restore/criu/blob/98186981d9eb430e39da3ee692e41ccdcfa67bff/test/zdtm/static/netns_lock.c#L44 https://github.com/checkpoint-restore/criu/blob/98186981d9eb430e39da3ee692e41ccdcfa67bff/test/zdtm/static/netns_lock.c#L58
5. ```ip``` mounts the netns to the file https://github.com/shemminger/iproute2/blob/9f366536edb5158343152604e82b968be46dbf26/ip/ipnetns.c#L910

[...]

PR #1531, opened 2021-06-28
test: fix netns_lock CI test failure

ZeyadYasser:
As mentioned in https://github.com/checkpoint-restore/criu/issues/1530 netns_lock test fails in jenkins iterative tests.

```netns_lock``` test is highly dependent on the order of the hooks, and iterations causes the ```--pre-dump``` hook to be called multiple times which expectedly causes the test to fail.

https://ci.kernoops.org/job/CRIU/job/CRIU-iter/job/criu-dev/431/testReport/(root)/criu/zdtm_static_netns_lock/

This only fixes the jenkins test failure and not the random CentOS7 failure mentioned in the issue. 

### DIFF
--- a/test/zdtm/static/netns_lock.c
+++ b/test/zdtm/static/netns_lock.c
@@ -8,6 +8,7 @@ const char *test_author = "Zeyad Yasser <zeyady98@gmail.com>";
 #include <sched.h>
 
 #define NS_PATH "/var/run/netns/criu-net-lock-test"
+#define SYNCFILE_PATH "net_lock.sync"
 #define MAX_RETRY 3
 
 int main(int argc, char **argv)
@@ -41,8 +42,7 @@ int main(int argc, char **argv)
 	 */
 
 	for (i = 0; i < MAX_RETRY; i++) {
-		ns_fd = open(NS_PATH, O_RDONLY);
-		if (ns_fd < 0) {
+		if (access(SYNCFILE_PATH, F_OK )) {
 			/* Netns not created yet by post-start hook */
 			sleep(1);
 			continue;
@@ -50,6 +50,7 @@ int main(int argc, char **argv)
 		break;
 	}
 
+	ns_fd = open(NS_PATH, O_RDONLY);
 	if (ns_fd < 0) {
 		pr_perror("can't open network ns");
 		return 1;
--- a/test/zdtm/static/netns_lock.hook
+++ b/test/zdtm/static/netns_lock.hook
@@ -35,30 +35,32 @@ if sys.argv[1] == "--post-start":
         srv.bind(("127.0.0.1", PORT))
         srv.listen(1)
 
-        # We should accept connections normally from
-        # pre-dump client, post-restore client and
-        # the pre-restore client using SOCCR_MARK
-        # The pre-restore client without SOCCR_MARK
-        # should fail with a timeout
+        # Loop allows zdtm multiple iterations to work (i.e. --iter 3)
+        while True:
+            # We should accept connections normally from
+            # pre-dump client, post-restore client and
+            # the pre-restore client using SOCCR_MARK
+            # The pre-restore client without SOCCR_MARK
+            # should fail with a timeout
 
-        # Accept pre-dump client
-        cln, addr = srv.accept()
-        cln.sendall(str.encode("--pre-dump"))
-        cln.close()
+            # Accept pre-dump client
+            cln, addr = srv.accept()
+            cln.sendall(str.encode("--pre-dump"))
+            cln.close()
 
-        # Accept pre-restore client with SOCCR_MARK
-        srv.setsockopt(socket.SOL_SOCKET, SO_MARK, SOCCR_MARK)
-        cln, addr = srv.accept()
-        cln.sendall(str.encode("--pre-restore"))
-        cln.close()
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: test/jenkins/criu-iter.sh [hunk 1]
```diff
@@ -5,4 +5,6 @@ set -e
 source `dirname $0`/criu-lib.sh
 prep
 mount_tmpfs_to_dump
-./test/zdtm.py run --all --keep-going --report report --parallel 4 --iter 3 -x 'maps04' || fail
+
+ITER_EXCLUDE="-x netns_lock -x maps04"
+./test/zdtm.py run --all --keep-going --report report --parallel 4 --iter 3 $ITER_EXCLUDE || fail
```
rst0git, line 10: We could use an infinite loop for the server to enable support for more than one iterations:
```diff
--- a/test/zdtm/static/netns_lock.hook
+++ b/test/zdtm/static/netns_lock.hook
@@ -35,30 +35,29 @@ if sys.argv[1] == "--post-start":
         srv.bind(("127.0.0.1", PORT))
         srv.listen(1)
 
-        # We should accept connections normally from
-        # pre-dump client, post-restore client and
-        # the pre-restore client using SOCCR_MARK
-        # The pre-restore client without SOCCR_MARK
-        # should fail with a timeout
-
-        # Accept pre-dump client
-        cln, addr = srv.accept()
-        cln.sendall(str.encode("--pre-dump"))
-        cln.close()
-
-        # Accept pre-restore client with SOCCR_MARK
-        srv.setsockopt(socket.SOL_SOCKET, SO_MARK, SOCCR_MARK)
-        cln, addr = srv.accept()
-        cln.sendall(str.encode("--pre-restore"))
-        cln.close()
-        srv.setsockopt(socket.SOL_SOCKET, SO_MARK, 0)
-
-        # Accept post-restore client
-        cln, addr = srv.accept()
-        cln.sendall(str.encode("--post-restore"))
-        cln.close()
-
-        srv.close()
+        while True:
+            # We should accept connections normally from
+            # pre-dump client, post-restore client and
+            # the pre-restore client using SOCCR_MARK
+            # The pre-restore client without SOCCR_MARK
+            # should fail with a timeout
+
+            # Accept pre-dump client
+            cln, addr = srv.accept()
+            cln.sendall(str.encode("--pre-dump"))
+            cln.close()
+
+            # Accept pre-restore client with SOCCR_MARK
+            srv.setsockopt(socket.SOL_SOCKET, SO_MARK, SOCCR_MARK)
+            cln, addr = srv.accept()
+            cln.sendall(str.encode("--pre-restore"))
+            cln.close()
+            srv.setsockopt(socket.SOL_SOCKET, SO_MARK, 0)
+
+            # Accept post-restore client
+            cln, addr = srv.accept()
+            cln.sendall(str.encode("--post-restore"))
+ 
```

test/jenkins/criu-iter.sh
```diff
@@ -5,4 +5,6 @@ set -e
 source `dirname $0`/criu-lib.sh
 prep
 mount_tmpfs_to_dump
-./test/zdtm.py run --all --keep-going --report report --parallel 4 --iter 3 -x 'maps04' || fail
+
+ITER_EXCLUDE="-x netns_lock -x maps04"
```
avagin, line 9: pls add a comment that explains why we skip netns_lock.

test/jenkins/criu-iter.sh
ZeyadYasser: Great, but is it okay if the server is not closed? It will be freed eventually I think.

ZeyadYasser:
I added a fix for random failures in other CIs that were mentioned in https://github.com/checkpoint-restore/criu/issues/1530.

ZeyadYasser:
Closes #1530

test/jenkins/criu-iter.sh
rst0git: zdtm would send [`SIGKILL`](https://github.com/checkpoint-restore/criu/blob/cbaa59976dbf9db618a7bbf3e9f175f40bba72bc/test/zdtm.py#L1362) to the server process.

test/jenkins/criu-iter.sh
ZeyadYasser: I went with @rst0git loop method

adrianreber:
See #1534 for a fix for the cross compilation errors.

test/zdtm/static/netns_lock.c
```diff
@@ -8,6 +8,7 @@ const char *test_author = "Zeyad Yasser <zeyady98@gmail.com>";
 #include <sched.h>
 
 #define NS_PATH "/var/run/netns/criu-net-lock-test"
+#define SYNCFILE_PATH "/tmp/criu-net-lock-test-sync"
```
rst0git, line 11: It would be better to use a relative path instead, for example:
```diff
-#define SYNCFILE_PATH "/tmp/criu-net-lock-test-sync"
+#define SYNCFILE_PATH "zdtm/static/net_lock.sync"
```

ZeyadYasser: Thanks, that looks more elegant.

ZeyadYasser:
> See #1534 for a fix for the cross compilation errors.

@adrianreber Thanks, I rebased and all tests pass now.

rst0git (approved):

PR #1532, opened 2021-06-29
criu: dump and restore notify_thread_id of posix timer

liuchao173:
When sigev_notify_thread_id is not set, get_pid will return a NULL
pointer and do_timer_create will return -EINVAL in kernel. So criu
will failed to create posix timer:

(09.806760) pie: 41301: Error (criu/pie/restorer.c:1998): Can't restore posix timers -22
(09.806824) pie: 41301: Error (criu/pie/restorer.c:2133): Restorer fail 41301
(09.891880) Error (criu/cr-restore.c:2596): Restoring FAILED.

Signed-off-by: Liu Chao <liuchao173@huawei.com>

### DIFF
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -1062,7 +1062,7 @@ int parse_pid_status(pid_t pid, struct seize_task_status *ss, void *data)
 	if (bfdopenr(&f))
 		return -1;
 
-	while (done < 12) {
+	while (done < 13) {
 		str = breadline(&f);
 		if (str == NULL)
 			break;
@@ -1080,6 +1080,20 @@ int parse_pid_status(pid_t pid, struct seize_task_status *ss, void *data)
 				pr_err("Unable to parse: %s\n", str);
 				goto err_parse;
 			}
+			done++;
+			continue;
+		}
+
+		if (!strncmp(str, "NSpid:", 6)) {
+			/* Get a thread ID in the thread PID namespace. */
+			char *last;
+
+			last = strrchr(str, '\t');
+			if (!last || sscanf(last, "%d", &cr->s.vpid) != 1) {
+				pr_err("Unable to parse: %s\n", str);
+				goto err_parse;
+			}
+
 			done++;
 			continue;
 		}
@@ -1165,7 +1179,7 @@ int parse_pid_status(pid_t pid, struct seize_task_status *ss, void *data)
 	}
 
 	/* seccomp is optional */
-	if (done >= 11 || (done == 10 && !parsed_seccomp))
+	if (done >= 12 || (done == 11 && !parsed_seccomp))
 		ret = 0;
 
 err_parse:
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -2342,6 +2342,7 @@ int parse_posix_timers(pid_t pid, struct proc_posix_timers_stat *args)
 
 			if ( tidpid[0] == 't') {
 				timer->spt.it_sigev_notify = SIGEV_THREAD_ID;
+				timer->spt.notify_thread_id = pid_t;
 			} else {
 				switch (sigpid[0]) {
 					case 's' :
--- a/images/timer.proto
+++ b/images/timer.proto
@@ -21,6 +21,7 @@ message posix_timer_entry {
 	required uint64		insec		= 8;
 	required uint64		vsec		= 9;
 	required uint64		vnsec		= 10;
+	optional int32		notify_thread_id= 11;
 }
 
 message task_timers_entry {
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -59,6 +59,7 @@ TST_NOFILE	:=				\
 		pthread00			\
 		pthread01			\
 		pthread02			\
+		pthread_timers			\
 		vdso00				\
 		vdso01				\
 		vdso02				\
@@ -519,6 +520,7 @@ inotify_system_nodel:	CFLAGS += -DNODEL
 pthread00:		LDLIBS += -pthread
 pthread01:		LDLIBS += -pthread
 pthread02:		LDLIBS += -pthread
+pthread_timers:		LDLIBS += -lrt -pthread
 different_creds:	LDLIBS += -pthread
 sigpending:		LDLIBS += -pthread
 sigaltstack:		LDLIBS += -pthread
--- a/compel/include/uapi/infect.h
+++ b/compel/include/uapi/infect.h
@@ -19,6 +19,7 @@ struct seize_task_status {
 	unsigned long long	sigpnd;
 	unsigned long long	shdpnd;
 	char			state;
+	int			vpid;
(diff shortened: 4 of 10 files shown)


adrianreber:
Thanks for the pull request. If you look at the CI results based on Alpine you will see compilation errors. Can you please fix these errors.

liuchao173:
The definition of struct sigevent in glibc and musl is different. I will try to fix it later.

images/timer.proto
```diff
@@ -21,6 +21,7 @@ message posix_timer_entry {
 	required uint64		insec		= 8;
 	required uint64		vsec		= 9;
 	required uint64		vnsec		= 10;
+	required int32		notify_thread_id= 11;
```
avagin, line 24: this has to be "optional". Images have to be backward-compatible.

liuchao173: I think users should dump and restore using the same version of criu.
If we don't save notify_thread_id in images, recovery won't success at all.

avagin: > I think users should dump and restore using the same version of criu.

We declare that criu restore should work for images of the same or lower version.

> If we don't save notify_thread_id in images, recovery won't success at all.

It won't succeed only in the case of SIGEV_THREAD_ID, will it? Actually, notify_thread_id means something only in the SIGEV_THREAD_ID case, and this looks like another reason why it has to be optional.

Snorch:
@liuchao173 Please add a test for this PR. In test/zdtm/static/posix_timers.c we only test `sigev_notify = SIGEV_SIGNAL`, so we need to add `SIGEV_THREAD_ID` test to zdtm too.

Also I have some doubts about this kind of direct use of pid. You should consider that create_posix_timers is called from container pidns, but parse_posix_timers gets pid relative to /proc pidns which is likely hosts one. So one should convert pid to the right pidns.

Snorch (commented):
@liuchao173 Please add a test for this PR. In test/zdtm/static/posix_timers.c we only test `sigev_notify = SIGEV_SIGNAL`, so we need to add `SIGEV_THREAD_ID` test to zdtm too.

Also I have some doubts about this kind of direct use of pid. You should consider that create_posix_timers is called from container pidns, but parse_posix_timers gets pid relative to /proc pidns which is likely hosts one. So one should convert pid to the right pidns.

avagin:
@liuchao173 I prepared a few fixes and a test. You can find my changes here: https://github.com/avagin/criu/commits/posix_thread_timers. If they look good to you, you can cherry-pick them in your branch.

The test can be executed with this command:
python test/zdtm.py run -t zdtm/static/pthread_timers  --iter 3

liuchao173:
> @liuchao173 I prepared a few fixes and a test. You can find my changes here: https://github.com/avagin/criu/commits/posix_thread_timers. If they look good to you, you can cherry-pick them in your branch.
> 
> The test can be executed with this command:
> python test/zdtm.py run -t zdtm/static/pthread_timers --iter 3

I think we shoud add a test for SIGEV_THREAD_ID timers and there's already a test file named pthread_timers.c

test/zdtm/static/Makefile
```diff
@@ -240,6 +241,7 @@ TST_NOFILE	:=				\
 		zombie_leader			\
 		sigtrap				\
 		sigtrap01			\
+		change_mnt_context		\
```
rst0git, line 244: This change looks unrelated and it was already introduced with https://github.com/checkpoint-restore/criu/commit/cbaa59976dbf9db618a7bbf3e9f175f40bba72bc.

images/timer.proto
```diff
@@ -21,7 +21,8 @@ message posix_timer_entry {
 	required uint64		insec		= 8;
 	required uint64		vsec		= 9;
 	required uint64		vnsec		= 10;
-	required int32		notify_thread_id= 11;
+
+	optional int32		notify_thread_id= 11;
```
rst0git, line 25: Should we fix this in https://github.com/checkpoint-restore/criu/pull/1532/commits/875c5434c2fee8b65e118c4b91c6ef175b9f9511 instead of creating a new patch/commit?

rst0git (approved):

criu/proc_parse.c
```diff
@@ -1084,6 +1084,20 @@ int parse_pid_status(pid_t pid, struct seize_task_status *ss, void *data)
 			continue;
 		}
 
+		if (!strncmp(str, "NSpid:", 6)) {
```
Snorch, line 1087: This is a first use of NSpid feature in CRIU. But this is quiet a "new" kernel feature and we use it without compat check. Does it mean that we don't support kernel older than 4.1?

```
git describe --contains e4bc33245124d
v4.1-rc1~115^2~41
```

In VZ criu we do something like this https://src.openvz.org/projects/OVZ/repos/criu/commits/94f4653f20#criu/kerndat.c for compatibility.

Alternatively we can probably use gettid() via parasite (like in dump_one_task->dump_task_threads ... ->dump_thread).

And anyway, should not this hunk be merged in the original patch or be a separate preparation patch instead of fixup like we have it now.

Snorch (approved):

criu/proc_parse.c
Snorch: In personal talk with @avagin we've decided that requiring NSpid feature from kernel is ok and adding ns support as a later patch is also ok. So it's merged.

Snorch:
Thanks @liuchao173! Merged.

adrianreber:
This commit breaks CenOS 7 as the kernel has no NSPid. With this PR merged we are no longer supporting CentOS 7.

@checkpoint-restore/maintainers Do we want to revert this change or drop CentOS 7 support?

Snorch:
@adrianreber Thanks for info! I will prepare a fix soon.

Snorch:
Here is a pr to fix it https://github.com/checkpoint-restore/criu/pull/1562 But let's first see if it actually helps and the tests pass.

adrianreber:
@Snorch Just saw that Jenkins on s390x also started to fail with the introduction of this commit:

https://ci.kernoops.org/job/CRIU/view/CRIU-devel/job/CRIU-snap/job/criu-dev/549/testReport/junit/(root)/criu/zdtm_static_pthread_timers/

CC: @mihalicyn

Snorch:
@adrianreber This one looks like a coincidence...

```
b'(00.156158) Sent msg to daemon 69 0 0'
b'(00.156160) Wait for ack 69 on daemon socket'
b'(00.156164) Fetched ack: 69 69 36'
b'(00.156166) Error (compel/src/lib/infect-rpc.c:71): Command 69 for daemon failed with 36'
b"(00.156169) Error (criu/cr-dump.c:1406): Can't dump posix timers (pid: 66)"
```

69 must be PARASITE_CMD_DUMP_POSIX_TIMERS, so we look into dump_posix_timers() and see that it can return positive overrun (36) of last timer if we are lucky =) which is treated by compel_rpc_sync as an error... But it's not regression.

Will also send a fixup.

PR #1534, opened 2021-07-01
ci: fix mips64el-cross test

adrianreber:
The mips64el-cross test target started to show following error:

`error: listing the stack pointer register '$29' in a clobber list is deprecated [-Werror=deprecated]`

This fixes it in three different places by removing '$29' from the clobber list. This is only compile tested as we have no mips hardware for testing.

### DIFF
--- a/compel/arch/mips/src/lib/include/uapi/asm/sigframe.h
+++ b/compel/arch/mips/src/lib/include/uapi/asm/sigframe.h
@@ -51,7 +51,7 @@ struct rt_sigframe {
 	"syscall					\n"	\
 	:							\
 	: "r"(new_sp)						\
-	: "$29","$2","memory")
+	: "$2","memory")
 
 int sigreturn_prep_fpu_frame(struct rt_sigframe *sigframe,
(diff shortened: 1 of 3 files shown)


adrianreber:
@sunny868 CRIU's mips cross compilation test is failing and this is a fix for it. Does that make sense from your point of view?

rst0git (approved):
On MIPS register $29 is a stack pointer and it should not be included in the clobber list.

>This is because the compiler requires the value of the stack pointer to be the same after an asm statement as it was on entry to the statement. However, previous versions of GCC did not enforce this rule and allowed the stack pointer to appear in the list, with unclear semantics. This behavior is deprecated and listing the stack pointer may become an error in future versions of GCC.

https://gcc.gnu.org/onlinedocs/gcc/Extended-Asm.html#Clobbers-and-Scratch-Registers

sunny868 (approved):

sunny868:
> @sunny868 CRIU's mips cross compilation test is failing and this is a fix for it. Does that make sense from your point of view?

I t`s right. Thanks for your patch.

PR #1495, opened 2021-06-02
lsm: support basic AppArmor stacking

brauner:
This is a revival of @tych0's patchset. I had to fix it up in quite a few places but with this I at least manage to restore the outer AppArmor namespace correctly. But I could use some good reviewers for this.
//Cc @tycho

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

PR #1535, opened 2021-07-02
CRIU 3.16

adrianreber:
I started to look into what a CRIU 3.16 release would look like.

This takes everything from the criu-dev branch except flog and tries to move it to the release branch.

In the last release we also skipped criu-ns, but with all the recent work on criu-ns I thought it might be ready for the release branch.

@checkpoint-restore/maintainers any thoughts about things from `criu-dev` which should not be part of the next release?


avagin:
We have to fix the issue with config files that we've found recently.

And I think it would be nice to have #1495 in this release too.

adrianreber:
> We have to fix the issue with config files that we've found recently.

The memory handling, right?

> And I think it would be nice to have #1495 in this release too.

Yes, there are a couple other PRs which would also be good to have. #1528 and #1531 for example

avagin:
> > We have to fix the issue with config files that we've found recently.
> 
> The memory handling, right?

Right. #1525 fixed the issue for the inherit-fd option, but we have more options that are affected by this issue.

adrianreber:
> > > We have to fix the issue with config files that we've found recently.
> > 
> > 
> > The memory handling, right?
> 
> Right. #1525 fixed the issue for the inherit-fd option, but we have more options that are affected by this issue.

Right, but it is not really a regression. It has been this way since we have support for configuration files. It would be nice to have a fix but releasing without a fix would give our users at least a fix for inherit-fd via configuration files without any new regressions.

avagin:
> Right, but it is not really a regression. It has been this way since we have support for configuration files. It would be nice to have a fix but releasing without a fix would give our users at least a fix for inherit-fd via configuration files without any new regressions.

I don't want to discuss this. My opinion is that this is a release blocker. This is a use-after-free issue... how can it be exploited? What do you think how many human hours we need to fix it? My estimation is about 1-2 hours.

xemul:
@checkpoint-restore/maintainers , please, share your thoughts on this

0x7f454c46:
I'm going to agree that it's not regression (that's why I originally gave Reviewed-by to PR with config fixes as before were no tests, but after some tests for config files were created).
But making a release is in my view a separate question from merging a non-regressing PR.

0x7f454c46:
So, in linux/kernel terms it can be a release candidate, but not a final release.

rst0git:
> My opinion is that this is a release blocker. This is a use-after-free issue... how can it be exploited? What do you think how many human hours we need to fix it? My estimation is about 1-2 hours.

I can open a PR in the next few days adding tests for configuration files to cover all CRIU options and fix any use-after-free issues.

adrianreber:
> > My opinion is that this is a release blocker. This is a use-after-free issue... how can it be exploited? What do you think how many human hours we need to fix it? My estimation is about 1-2 hours.
> 
> I can open a PR in the next few days adding tests for configuration files to cover all CRIU options and fix any use-after-free issues.

Thanks for your help. That was the anyway the main problem from my point of view. Somebody needs to write the patch. Thanks for stepping up.

avagin:
@rst0git https://github.com/avagin/criu/commit/07dad2ae7caa596e3378a9d5f33830452d2b83c8 is an idea of how a test can look like.

adrianreber:
@checkpoint-restore/maintainers Everything for the 3.16 release should be ready in the PR. Please take a look.

avagin (approved):

avagin:
Why don't I see all previous comments for this PR?

adrianreber:
> Why don't I see all previous comments for this PR?

Maybe because they were Draft comments?

adrianreber:
@xemul Any comments from your side?

xemul (approved):

rst0git (approved):

PR #1536, opened 2021-07-02
zdtm: make --sbs also stop on each predump/snap iteration

Snorch:
This is useful to investigate problems on predump iterations. After this
patch test output with "--pre=2 --sbs" would have new usefull stop
points:

========================== Run zdtm/static/env00 in h ==========================
Start test
./env00 --pidfile=env00.pid --outfile=env00.out --envname=ENV_00_TEST
Pause at iter. Press Enter to continue.
Run criu pre-dump
Pause at iter. Press Enter to continue.
Run criu pre-dump
Pause at pre-dump. Press Enter to continue.
Run criu dump

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1494,7 +1494,7 @@ def init_sbs():
 
 def sbs(what):
     if do_sbs:
-        input("Pause at %s. Press Enter to continue." % what)
+        input("Pause %s. Press Enter to continue." % what)
 
 
 #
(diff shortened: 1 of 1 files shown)


test/zdtm.py
```diff
@@ -1515,6 +1515,7 @@ def cr(cr_api, test, opts):
     for i in iters[0]:
         pres = iter_parm(opts['pre'], 0)
         for p in pres[0]:
+            sbs('iter')
```
avagin, line 1518: this is pre-pre-dump :)

Snorch: Yes =) Hope its ok to just call it `iter`.

adrianreber:
If you rebase on latest criu-dev the cross compilation errors should disappear.

test/zdtm.py
adrianreber: Could you also print the number of the iteration?

Snorch:
Rebased to latest criu-dev + added iter number.

adrianreber (approved):

test/zdtm.py
avagin: > Yes =) Hope its ok to just call it iter.

I think we need to use a different name for this stop point. Here is another loop that enumerates iterations.

Snorch: >> Yes =) Hope its ok to just call it iter.
>
> I think we need to use a different name for this stop point. Here is another loop that enumerates iterations.

Looks like the only way to fix naming here is to replace "pre-" with "before" in places where "pre-" is not related to criu-pre-dump. -> Done.

PR #1539, opened 2021-07-04
[GSoC] Add nftables based network locking/unlocking

ZeyadYasser:
## Overview

During checkpointing and restoring, CRIU locks the network to make sure no TCP packets are accepted by the network stack during the time the process is checkpointed (to avoid the kernel from sending a RST packet).
Currently CRIU calls out to ```iptables-restore``` to create and delete the corresponding iptables rules (external binary).

This approach uses ```nftables``` as a replacement for ```iptables-restore```.
Using ```libnftables``` eliminates the external ```iptables-restore``` binary dependency, also ```nftables``` is much more efficient than ```iptables```.

### Kerndat

*Check needed kernel features exist*
Check that nftables [concatenations](https://wiki.nftables.org/wiki-nftables/index.php/Concatenations) is supported [has_nftables_concat]

### Feature check
`criu check --feature network-lock-nftables`

Add ability to check if nftables based locking/unlocking is possible.
This would typically check the corresponding kerndat fields.

### Add --network-lock option

After nftables based locking/unlocking we would have two values available for this option (`iptables` and `nftables`)
It would be better to default to `iptables` until nftables method is tested thoroughly.
e.g.
`criu dump -t 2221 --tcp-established --network-lock iptables`
`criu dump -t 2221 --tcp-established --network-lock nftables`

Corresponding RPC and libcriu option must be added.

### Algorithm flow

I want to use the same hooks that invoke iptables locking/unlocking.
i.e. `network_lock_internal`, `network_unlock_internal` for network-wide case
`nf_lock_connection` for per-socket case (could be renamed to `lock_connection`)

Note: Per-socket rule should be loaded in `network_lock`, `lock_connection` will just add connection tuple entries to the associated nftables set.

CRIU could just decide in those hooks which method should be used based on ```--network-lock``` option.

```c
if (opts.network_lock_method == NETWORK_LOCK_IPTABLES)
	ret |= iptables_network_unlock_internal();
else if (opts.network_lock_method == NETWORK_LOCK_NFTABLES)
	ret |= nftables_network_unlock_internal();
```

This flow would centralize network locking/unlocking in those hooks and avoid forgetting using one or the other anywhere in the source code, both should have identical behavior anyway and should be interchangeable seamlessly with the ```--network-lock``` option.

#### Locking (netns-wide)

1. Create a table named CRIU
    Similar to `nft add table inet CRIU`
2. Create INPUT/OUTPUT chains with a default *"drop"* policy
    Similar to `nft add chain inet CRIU output { type filter hook output priority 0 ; policy drop; }`
3. Create a rule to *"accept"* packets with SOCCR_MARK
    Similar to `nft add rule inet CRIU output meta mark 0xC114 accept`

#### Locking (per-socket)

**Preparation**

1. Create a table named CRIU
    Similar to `nft add table inet CRIU`
2. Create INPUT/OUTPUT chains
    Similar to `nft add chain inet CRIU output { type filter hook output priority 0 }`
3. Create a rule to *"accept"* packets with SOCCR_MARK
    Similar to `nft add rule inet CRIU output meta mark 0xC114 accept`
4. Create connections set, which should contain connection identifying tuples (*src_addr, src_port, dst_addr, dst_port*)
    This is possible due to concatenations support for nftables sets which requires kernel ≥ 4.1 [https://wiki.nftables.org/wiki-nftables/index.php/Concatenations](https://wiki.nftables.org/wiki-nftables/index.php/Concatenations).
     Similar to `nft add set inet CRIU conns { type ipv4_addr . inet_service . ipv4_addr . inet_service ;}`
5. Create a rule to *"drop"* packets that match connections in the `conns` set
    Similar to `nft add rule inet CRIU output ip saddr . tcp sport . ip daddr . tcp dport @conns drop`

Note: A separate set/rule is needed for IPv6.

**Locking one socket**

CRIU just needs to add the connection identifying tuple to the `conns` set in `nftables_lock_connection`.
Similar to `nft add element inet CRIU conns { 192.168.1.96 . 46315 . 1.1.1.1 . 53 }`
Nftables sets are very efficient and avoids adding a separate rule for each connection, we only need to add connection entries to the `conns` set.

#### Unlocking

With nftables the process is much easier, we only need to delete the CRIU table and that is it.
Similar to `nft delete table inet CRIU`

## Progress
- [x] Network-wide locking test #1465 
- [x] Add ```--network-lock``` option
- [x] Feature check
- [x] Nftables based network-wide locking
- [x] Per-socket locking tests
- [x] Nftables based per-socket locking


criu/config.c
```diff
@@ -691,6 +692,7 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		{ "pre-dump-mode",		required_argument,	0, 1097},
 		{ "file-validation",		required_argument,	0, 1098	},
 		{ "lsm-mount-context",		required_argument,	0, 1099	},
+		{ "network-lock",		required_argument,	0, 1100	},
```
rst0git, line 695: We also need to add a description in `criu --help` and [`man criu`](https://github.com/checkpoint-restore/criu/blob/f29a02a395083941fea8d5f09ccf7efacdd60836/Documentation/criu.txt)

criu/config.c
```diff
@@ -1023,6 +1025,14 @@ int parse_options(int argc, char **argv, bool *usage_error,
 		case 1099:
 			SET_CHAR_OPTS(lsm_mount_context, optarg);
 			break;
+		case 1100:
+			if (!strcmp("iptables", optarg)) {
+				opts.network_lock_method = NETWORK_LOCK_IPTABLES;
+			} else {
+				pr_err("Unable to parse value of --network-lock\n");
```
rst0git, line 1032: I think the following message would be more descriptive
```
pr_err("Invalid value for --network-lock: %s\n", optarg);
```

criu/config.c
ZeyadYasser: I was thinking of adding that later when I add the ```--network-lock nftables``` option, but now that you pointed it out, I think I should add the ```nftables``` option + help/docs here because the rest depends on the option's existence anyway.
I will update it right away.

lib/c/criu.c
```diff
@@ -1793,3 +1793,21 @@ void criu_set_pidfd_store_sk(int sk)
...
+{
+	opts->rpc->has_network_lock = true;
+	if (method == CRIU_NETWORK_LOCK_IPTABLES) {
+		opts->rpc->network_lock = (CriuNetworkLockMethod)method;
+		return 0;
+	} else if (method == CRIU_NETWORK_LOCK_NFTABLES) {
+		opts->rpc->network_lock = (CriuNetworkLockMethod)method;
+		return 0;
+	}
+	return -1;
```
rst0git, line 1807: This could be simplified to something like:
```C
int criu_local_set_network_lock(criu_opts *opts, enum criu_network_lock_method method)
{
	int ret = -1;

	if (method == CRIU_NETWORK_LOCK_IPTABLES || method == CRIU_NETWORK_LOCK_NFTABLES) {
		opts->rpc->has_network_lock = true;
		opts->rpc->network_lock = (CriuNetworkLockMethod)method;
		ret = 0;
	}

	return ret;
}
```

criu/kerndat.c
```diff
@@ -1133,6 +1136,36 @@ static int kerndat_has_pidfd_getfd(void)
...
+	int ret = 0;
+
+	nft = nft_ctx_new(NFT_CTX_DEFAULT);
+	if (!nft)
+		return -1;
+
+	kdat.has_nftables_concat = true;
+	if (NFT_RUN_CMD(nft, "add table inet CRIU")) {
+		ret = -1;
+		goto nft_ctx_free_out;
```
rst0git, line 1152: When `NFT_RUN_CMD()` fails here, `kdat.has_nftables_concat` would remain set to `true`.

criu/net.c
```diff
@@ -3053,6 +3054,54 @@ static int iptables_restore(bool ipv6, char *buf, int size)
...
+	struct nft_ctx *nft;
+	int ret = 0;
+
+	nft = nft_ctx_new(NFT_CTX_DEFAULT);
+	if (!nft)
+		return -1;
+
+    if (NFT_RUN_CMD(nft, "add table inet CRIU")) {
+		ret = -1;
+        goto free;
```
rst0git, line 3069: Could you please change the indentation on these lines to tabs?

criu/net.c
```diff
@@ -3053,6 +3054,54 @@ static int iptables_restore(bool ipv6, char *buf, int size)
...
+    if (NFT_RUN_CMD(nft, "add rule inet CRIU input meta mark " __stringify(SOCCR_MARK) " accept")) {
+		ret = -1;
+        goto err;
+    }
+
+    goto free;
+
+err:
+    NFT_RUN_CMD(nft, "delete table inet CRIU");
+free:
```
rst0git, line 3096: It would be better to use a different name here because `free` is a function defined in `stdlib.h`. For example:
```diff
-free:
+out:
```

test/zdtm/static/per_socket_lock_iptables.hook
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
+
+import subprocess
+import socket
+import time
+import sys
+import os
+
+SO_MARK = 36  # Define socket option for python2 compatibility
+SOCCR_MARK = 0xC114
```
rst0git, line 10: These `SO_MARK` and `SOCCR_MARK` constants are not used.

test/zdtm/static/per_socket_lock_iptables.hook
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python
+
+import subprocess
```
rst0git, line 3: `subprocess` is not used

test/zdtm/static/Makefile
```diff
@@ -216,6 +216,7 @@ TST_NOFILE	:=				\
 		pipe03				\
 		netns_lock_iptables			\
 		netns_lock_nftables			\
+		per_socket_lock_iptables		\
```
rst0git, line 219: It would be better to add a prefix to the test name, because "lock" could mean something else (e.g. file lock).
For example: `net_lock_socket_iptables`.

test/zdtm/static/per_socket_lock_iptables.hook
```diff
@@ -0,0 +1,82 @@
...
+
+    internal_sock.close()
+
+if sys.argv[1] in ["--pre-dump", "--pre-restore", "--post-restore"]:
+    internal_conn = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
+    internal_conn.connect(INTERNAL_SERVER)
+    internal_conn.sendall(str.encode(sys.argv[1]))
+
+    status = internal_conn.recv(100).decode()
+    if status == "fail":
```
rst0git, line 70: It would be more reliable to check for a `pass` message instead.
```diff
-if status == "fail":
+if status != "pass":
```

test/zdtm/static/per_socket_lock_iptables.hook
```diff
@@ -0,0 +1,82 @@
...
+        sys.exit(1)
+
+    internal_conn.close()
+
+if sys.argv[1] == "--clean":
+    # Delete sync file
+    os.remove(SYNCFILE)
```
rst0git, line 77: It would be better to move the sync file deletion inside `wait_sync_file()`

test/zdtm/static/per_socket_lock_iptables.hook
```diff
@@ -0,0 +1,82 @@
...
+TIMEOUT = 0.1
+INTERNAL_SERVER = "\0internal_server"
+SYNCFILE = "zdtm/static/socket_lock.sync"
+
+def wait_sync_file():
+    for _ignore in range(3):
+        if os.path.exists(SYNCFILE):
+            return False
+        time.sleep(1)
+    return True
```
rst0git, line 21: `raise Exception("Sync timeout: file ({}) not found".format(SYNCFILE))`

criu/sk-tcp.c
```diff
@@ -51,7 +67,7 @@ static int tcp_repair_established(int fd, struct inet_sk_desc *sk)
 	}
 
 	if (!(root_ns_mask & CLONE_NEWNET)) {
-		ret = nf_lock_connection(sk);
+		ret = lock_connection(sk);
 		if (ret < 0)
 			goto err2;
```
rst0git, line 72: We could add an error message here. It might be helpful when debugging.
```
pr_err("Failed to lock TCP connection\n");
```

criu/net.c
```diff
@@ -3053,6 +3054,54 @@ static int iptables_restore(bool ipv6, char *buf, int size)
...
+
+	goto out;
+
+err:
+	NFT_RUN_CMD(nft, "delete table inet CRIU");
+out:
+	nft_ctx_free(nft);
+	return ret;
+#else
+	pr_err("CRIU must be compiled with libnftables support\n");
```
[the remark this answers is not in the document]
ZeyadYasser: I am not sure I understand clearly, Do you mean the message should appear when CRIU is installed from a package manager?

Documentation/criu.txt
```diff
@@ -420,6 +420,23 @@ The 'mode' may be one of the following:
...
+
+    *iptables*:::
+                Iptables rules are used to drop the packets. 'iptables-restore'
+                executable is required. This is the default if mode is
+                unspecified.
+
+    *nftables*:::
+                Nftables rules are used to drop the packets. 'libnftables' is
+                required at compile time.
+
```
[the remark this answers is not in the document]
rst0git: Screenshot

criu/net.c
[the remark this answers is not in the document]
rst0git: No, I think that we could make the following change:
```diff
-pr_err("CRIU must be compiled with libnftables support\n");
+pr_err("CRIU was built without libnftables support\n");
```
The reason is because if CRIU is installed from a package (e.g. `apt-get install criu`) this message may appear to a user.

test/jenkins/criu-iter.sh
```diff
@@ -5,4 +5,9 @@ set -e
 source `dirname $0`/criu-lib.sh
 prep
 mount_tmpfs_to_dump
-./test/zdtm.py run --all --keep-going --report report --parallel 4 --iter 3 -x 'maps04' || fail
+
+ITER_EXCLUDE="-x maps02"
+# Exclude those tests from multiple iterations because their hooks are
+# dependant on the exact order of the different stages
+ITER_EXCLUDE="$ITER_EXCLUDE -x net_lock_socket_iptables -x net_lock_socket_nftables"
+./test/zdtm.py run --all --keep-going --report report --parallel 4 --iter 3 $ITER_EXCLUDE || fail
```
rst0git, line 13: Hmm, the tests are passing when I run them with multiple iterations locally.

```console
sudo ./zdtm.py run -T 'zdtm/static/net_lock_socket_*' --parallel 4 --iter 3

Checking feature network_lock_nftables
network_lock_nftables is supported
userns is supported
=== Run 1/2 ========-------- zdtm/static/net_lock_socket_nftables
timens isn't supported on 5.13.4-200.fc34.x86_64
================ Run zdtm/static/net_lock_socket_nftables in h =================
Start test
Test is SUID
./net_lock_socket_nftables --pidfile=net_lock_socket_nftables.pid --outfile=net_lock_socket_nftables.out
Running zdtm/static/net_lock_socket_nftables.hook(--post-start)
Running zdtm/static/net_lock_socket_nftables.hook(--pre-dump)
Run criu dump
Running zdtm/static/net_lock_socket_nftables.hook(--pre-restore)
Run criu restore
Running zdtm/static/net_lock_socket_nftables.hook(--post-restore)
Running zdtm/static/net_lock_socket_nftables.hook(--pre-dump)
Run criu dump
Running zdtm/static/net_lock_socket_nftables.hook(--pre-restore)
Run criu restore
Running zdtm/static/net_lock_socket_nftables.hook(--post-restore)
Running zdtm/static/net_lock_socket_nftables.hook(--pre-dump)
Run criu dump
Running zdtm/static/net_lock_socket_nftables.hook(--pre-restore)
Run criu restore
Running zdtm/static/net_lock_socket_nftables.hook(--post-restore)
Send the 15 signal to  57
Wait for zdtm/static/net_lock_socket_nftables(57) to die for 0.100000
Running zdtm/static/net_lock_socket_nftables.hook(--clean)
Removing dump/zdtm/static/net_lock_socket_nftables/57
================ Test zdtm/static/net_lock_socket_nftables PASS ================

userns is supported
=== Run 2/2 ================ zdtm/static/net_lock_socket_iptables
timens isn't supported on 5.13.4-200.fc34.x86_64
================ Run zdtm/static/net_lock_socket_iptables in h =================
Start test
Test is SUID
./net_lock_socket_iptables --pidfile=net_lock_socket_iptables.pid --outfile=net_lock_socket_iptables.out
Running zdtm/static/net_lock_socket_iptables.hook(--post-start)
Running zdtm/static/net_lock_socket_iptables.hook(--pre-dump)
Run criu dump
Running zdtm/static/net_lock_socket_iptables.hook(--pre-restore)
Run criu restore
Running zdtm/static/net_lock_socket_iptables.hook(--post-restore)
Running zdtm/static/net_lock_socket_iptables.hook(--pre-dump)
Run criu dump
Running zdtm/static/net_lock_socket_iptables.hook(--pre-restore)
Run criu restore
Running zdtm/static/net_lock_socket_iptables.hook(--post-restore)
Running zdtm/static/net_lock_socket_iptables.hook(--pre-dump)
Run criu dump
Running zdtm/static/net_lock_socket_iptables.hook(--pre-restore)
Run criu restore
Running zdtm/static/net_lock_socket_iptables.hook(--post-restore)
Send the 15 signal to  57
Wait for zdtm/static/net_lock_socket_iptables(57) to die for 0.100000
Running zdtm/static/net_lock_socket_iptables.hook(--clean)
Removing dump/zdtm/static/net_lock_socket_iptables/57
================ Test zdtm/static/net_lock_socket_iptables PASS ================
```

ZeyadYasser: I am sorry, this is my bad. I assumed that it would fail the same as the netns tests did.

adrianreber:
This looks very good. I have, however, one question. What happens if I checkpoint with iptables and restore with nftables locking. Will that work? Will it warn me that this might not work?

ZeyadYasser:
> This looks very good. I have, however, one question. What happens if I checkpoint with iptables and restore with nftables locking. Will that work? Will it warn me that this might not work?

@adrianreber That will not work because unlocking is done by removing the iptables/nftables rules, so the network would stay locked after restore in this case. Currently it doesn't warn even about a failure in unlocking the network as the main unlocking function never returns an error.

https://github.com/ZeyadYasser/criu/blob/a26f4abd383ad1461b779a3b651b7fd254a5d905/criu/net.c#L3220-L3234
https://github.com/ZeyadYasser/criu/blob/a26f4abd383ad1461b779a3b651b7fd254a5d905/criu/netfilter.c#L159

adrianreber:
Would it make sense to track the locking method in an image file so that it would not be necessary to specify it during restore and at the same time ensure to intercept wrong locking methods?

rst0git:
> Would it make sense to track the locking method in an image file so that it would not be necessary to specify it during restore and at the same time ensure to intercept wrong locking methods?

It makes sense. This is similar to the `pre_dump_mode` value included in [images/inventory.proto](https://github.com/checkpoint-restore/criu/commit/e0ea21a#diff-25c209727062dd3ed9a2b18f3c54c2ad4f4df5b53b5389ca3d98d6626cc21044R19).

ZeyadYasser:
@adrianreber @rst0git If we store the method in the inventory image, Does it make sense to allow overwriting the ```--network-lock``` option on restore? Can't we just always use the value stored in the image during restore?

rst0git:
> Does it make sense to allow overwriting the --network-lock option on restore?

Not really. If iptables rules have been added to drop packets, it doesn't make sense to try to remove nftables rules and vice versa.

> Can't we just always use the value stored in the image during restore?

Yes

ZeyadYasser:
I just noticed that the ```rpc.proto``` is excluded from ```pycriu```. I imported it in ```inventory.proto``` to reuse ```enum criu_network_lock_method```
https://github.com/checkpoint-restore/criu/blob/f97e638d8318e477f48073a1b12da2ff8f79636c/lib/py/images/Makefile#L2-L8

Do you recommend that I redefine the enum in ```inventory.proto```? or something else completely?

rst0git:
> I just noticed that the rpc.proto is excluded from pycriu.

pycriu provides rpc.proto. It is used in the python version of [p.haul](https://github.com/checkpoint-restore/p.haul/blob/02199d875887eeff9f2c7d36da135e71e6bc4f5f/phaul/criu_req.py)
https://github.com/checkpoint-restore/criu/blob/f97e638d8318e477f48073a1b12da2ff8f79636c/lib/py/Makefile#L11

```python
>>> import pycriu.rpc_pb2
>>> help(pycriu.rpc_pb2)
```

The following line is used to exclude `rpc.proto` from `pycriu.images`.
This python package is used, for example by `crit`, to encode/decode CRIU images and `rpc.proto` is not an image, it specifies the communication protocol used with `criu service`.
```Makefile
    proto := $(filter-out images/rpc.proto, $(sort $(wildcard images/*.proto))) 
```


> Do you recommend that I redefine the enum in inventory.proto? or something else completely?

We could use `optional uint32` type to avoid unnecessary dependencies.

rst0git (approved):

adrianreber:
What happens if I dump with iptables locking and, although it is specified in the image, I add nftables as unlocking method during restore? Is there a check that locking method is not specified during restore? I have not seen that there is a check for specifying `--lock-method` only during dump.

adrianreber:
zdtm/static/net_lock_socket_iptables seems to hang on CentOS 7 during CI runs

test/zdtm/static/net_lock_socket_iptables.hook
```diff
@@ -0,0 +1,86 @@
...
+    for _ignore in range(3):
+        try:
+            with open(SYNCFILE, "r") as f:
+                addr = f.read(4)
+            os.remove(SYNCFILE)
+            if addr == "ipv4":
+                return "127.0.0.1"
+            elif addr == "ipv6":
+                return "::1"
+            else:
```
rst0git, line 23: unnecessary `elif` and `else` after return:
```python
if addr == "ipv4":
    return "127.0.0.1"

if addr == "ipv6":
    return "::1"

raise Exception("Invalid address type")
```

rst0git:
> What happens if I dump with iptables locking and, although it is specified in the image, I add nftables as unlocking method during restore? Is there a check that locking method is not specified during restore?

Thanks for point this out Adrian. We need to also consider the case when `network-lock` is in a config file. Perhaps CRIU should ignore the value of this option during restore?

criu/net.c
```diff
@@ -3053,6 +3054,54 @@ static int iptables_restore(bool ipv6, char *buf, int size)
...
+	}
+
+	if (NFT_RUN_CMD(nft, "add chain inet CRIU input { type filter hook input priority 0; policy drop; }")) {
+		ret = -1;
+		goto err;
+	}
+
+	if (NFT_RUN_CMD(nft, "add rule inet CRIU input meta mark " __stringify(SOCCR_MARK) " accept")) {
+		ret = -1;
+		goto err;
```
rst0git, line 3089: to avoid duplicating `ret = -1;`, it could be move after the `err:` label

criu/net.c
```diff
@@ -3086,13 +3135,36 @@ int network_lock_internal(void)
...
+{
+#if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
+	struct nft_ctx *nft;
+
+	nft = nft_ctx_new(NFT_CTX_DEFAULT);
+	if (!nft)
+		return -1;
+
+	if (NFT_RUN_CMD(nft, "delete table inet CRIU")) {
+		nft_ctx_free(nft);
```
ZeyadYasser, line 3157: ctx should always be freed not on failure only.

ZeyadYasser:
> What happens if I dump with iptables locking and, although it is specified in the image, I add nftables as unlocking method during restore? Is there a check that locking method is not specified during restore? I have not seen that there is a check for specifying `--lock-method` only during dump.

```--network-lock``` option is always ignored in restore and the option is overwritten using the value from the image. I removed it from the restore options documentation.
https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/image.c#L233-L235
https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/image.c#L89-L94

**Comment by ZeyadYasser** [comment]
> zdtm/static/net_lock_socket_iptables seems to hang on CentOS 7 during CI runs

I will install a CentOS 7 vm to debug.

**Comment by ZeyadYasser** [comment]
> zdtm/static/net_lock_socket_iptables seems to hang on CentOS 7 during CI runs

Thanks to @rst0git, it is now fixed.
It seems like there isn't FileNotFoundError in Python 2, so I just used IOError for both.

**Comment by adrianreber** [comment]
> > What happens if I dump with iptables locking and, although it is specified in the image, I add nftables as unlocking method during restore? Is there a check that locking method is not specified during restore? I have not seen that there is a check for specifying `--lock-method` only during dump.
> 
> `--network-lock` option is always ignored in restore and the option is overwritten using the value from the image. I removed it from the restore options documentation.

I think I would prefer that CRIU just errors out if someone specifies an option that has no effect instead of silently ignoring it.

> https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/image.c#L233-L235
> https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/image.c#L89-L94

```
	if (restore) {
		if (!he->has_network_lock_method)
			pr_warn("Network lock method not found in inventory image\n");
		else
			opts.network_lock_method = he->network_lock_method;
	}
```

Will if fall back to iptables locking in the `(!he->has_network_lock_method)` case? Because before we had the network locking method specified in the inventory CRIU would always use the iptables backend. Is that still the case?

**Comment by rst0git** [comment]
> I think I would prefer that CRIU just errors out if someone specifies an option that has no effect instead of silently ignoring it.

Hmm, this would mean if someone adds `network-lock` in a config file, it would work with `criu dump` and fail with `criu restore`. So they would have to remove it from the config file before restoring.

Currently, runc supports [checkpoint/restore annotations](https://github.com/opencontainers/runc/blob/master/docs/checkpoint-restore.md#checkpointrestore-annotations) to specify a config file per container, but the same config file is used for both checkpoint and restore. If CRIU fails with an error during restore when `network-lock` is included in a config file, this option would not be usable.

**Comment by adrianreber** [comment]
> > I think I would prefer that CRIU just errors out if someone specifies an option that has no effect instead of silently ignoring it.
> 
> Hmm, this would mean if someone adds `network-lock` in a config file, it would work with `criu dump` and fail with `criu restore`. So they would have to remove it from the config file before restoring.
> 
> Currently, runc supports [checkpoint/restore annotations](https://github.com/opencontainers/runc/blob/master/docs/checkpoint-restore.md#checkpointrestore-annotations) to specify a config file per container, but the same config file is used for both checkpoint and restore. If CRIU fails with an error during restore when `network-lock` is included in a config file, this option would not be usable.

Good point. So maybe just a debug message that network-lock has no effect during restore.

**Comment by adrianreber** [comment]
> > > What happens if I dump with iptables locking and, although it is specified in the image, I add nftables as unlocking method during restore? Is there a check that locking method is not specified during restore? I have not seen that there is a check for specifying `--lock-method` only during dump.
> > 
> > 
> > `--network-lock` option is always ignored in restore and the option is overwritten using the value from the image. I removed it from the restore options documentation.
> 
> I think I would prefer that CRIU just errors out if someone specifies an option that has no effect instead of silently ignoring it.
> 
> > https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/image.c#L233-L235
> > https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/image.c#L89-L94
> 
> ```
> 	if (restore) {
> 		if (!he->has_network_lock_method)
> 			pr_warn("Network lock method not found in inventory image\n");
> 		else
> 			opts.network_lock_method = he->network_lock_method;
> 	}
> ```
> 
> Will if fall back to iptables locking in the `(!he->has_network_lock_method)` case? Because before we had the network locking method specified in the inventory CRIU would always use the iptables backend. Is that still the case?

A warning feels a bit much here. Maybe a info message that CRIU will use iptables based locking.

**Comment by ZeyadYasser** [comment]
Do you think that all nftables commands being run should be logged? or would that clutter the log file?
I was thinking something like adding it to the ```NFT_RUN_CMD``` macro.
https://github.com/ZeyadYasser/criu/blob/nftables-net-lock/criu/include/netfilter.h#L16-L22

**Comment by rst0git** [comment]
> Do you think that all nftables commands being run should be logged? or would that clutter the log file?
> I was thinking something like adding it to the `NFT_RUN_CMD` macro.

We could add an error message similar to iptables. 

https://github.com/checkpoint-restore/criu/blob/493f8c25691f95fce8111b72631805b94665b51f/criu/net.c#L3076

It would provide enough information for debugging and avoid additional log messages when it works.

**Review by rst0git — APPROVED** [review]

**Review by avagin — APPROVED** [review]

**Review by mihalicyn — APPROVED** [review]
Great job!

I've left some small comments.

File: criu/kerndat.c [hunk 20]
```diff
@@ -1146,6 +1150,39 @@ int kerndat_has_nspid(void)
 	return ret;
 }
 
+static int kerndat_has_nftables_concat(void)
+{
+#if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
+	struct nft_ctx *nft;
+	int ret = 0;
+
+	nft = nft_ctx_new(NFT_CTX_DEFAULT);
```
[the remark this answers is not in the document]
ZeyadYasser: I can add it, but I think it wouldn't collide with anything, since no filtering rules are added (only a test set is created) which is added to separate CRIU table.
Having two instances of CRIU running at the same time would be a problem since the table name will for sure collide, I think actually that would also be the case during normal locking/unlocking like [here](https://github.com/checkpoint-restore/criu/pull/1539/files/1d7d0d252e350c0ec4f441c5b95dbb9ce6f99b02#diff-ffdccecb1d7573f5f13830e82a1b1c6e200f33b3bdb934ea9add47d68ad23406R3028). Is that a use case I should consider?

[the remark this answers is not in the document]
rst0git: It looks like the only difference between `add table inet CRIU` and `create table inet CRIU` is that the former will not return an error if the table already exists while `create` will return an error.

To avoid collisions, we could replace `add` with `create` and append the PID number specified with `-t` (`--tree`) to the table name.

avagin:
The test fails in our CI:
https://ci.kernoops.org/job/CRIU/job/CRIU-dump-only/job/criu-dev/668/consoleFull

```
=== Run 21/410 ---------------- zdtm/static/net_lock_socket_iptables
==== ALARM ====
timens isn't supported on 5.10.0-1023-oem
================ Run zdtm/static/net_lock_socket_iptables in h =================
Start test
Test is SUID
./net_lock_socket_iptables --pidfile=net_lock_socket_iptables.pid --outfile=net_lock_socket_iptables.out
    PID TTY          TIME CMD
     53 ?        00:00:00 net_lock_socket
      1 ?        S      0:00 python zdtm.py
      3 ?        S      0:00 python zdtm.py
     62 ?        R      0:00  \_ ps axf 53
     53 ?        Ss     0:00 ./net_lock_socket_iptables --pidfile=net_lock_socket_iptables.pid --outfile=net_lock_socket_iptables.out
     54 ?        Ss     0:00 python zdtm/static/net_lock_socket_iptables.hook --post-start
Running zdtm/static/net_lock_socket_iptables.hook(--post-start)
Running zdtm/static/net_lock_socket_iptables.hook(--pre-dump)
Run criu dump
Send the 15 signal to  53
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.100000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.200000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.400000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.800000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 1.600000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 3.200000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 6.400000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 12.800000
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 25.600000
 Test zdtm/static/net_lock_socket_iptables FAIL at zdtm/static/net_lock_socket_iptables die 
Send the 9 signal to  53
Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.100000
Running zdtm/static/net_lock_socket_iptables.hook(--clean)
```

ZeyadYasser:
> The test fails in our CI:
> https://ci.kernoops.org/job/CRIU/job/CRIU-dump-only/job/criu-dev/668/consoleFull
> 
> ```
> === Run 21/410 ---------------- zdtm/static/net_lock_socket_iptables
> ==== ALARM ====
> timens isn't supported on 5.10.0-1023-oem
> ================ Run zdtm/static/net_lock_socket_iptables in h =================
> Start test
> Test is SUID
> ./net_lock_socket_iptables --pidfile=net_lock_socket_iptables.pid --outfile=net_lock_socket_iptables.out
>     PID TTY          TIME CMD
>      53 ?        00:00:00 net_lock_socket
>     PID TTY      STAT   TIME COMMAND
>       1 ?        S      0:00 python zdtm.py
>       3 ?        S      0:00 python zdtm.py
>      62 ?        R      0:00  \_ ps axf 53
>      53 ?        Ss     0:00 ./net_lock_socket_iptables --pidfile=net_lock_socket_iptables.pid --outfile=net_lock_socket_iptables.out
>      54 ?        Ss     0:00 python zdtm/static/net_lock_socket_iptables.hook --post-start
> Running zdtm/static/net_lock_socket_iptables.hook(--post-start)
> Running zdtm/static/net_lock_socket_iptables.hook(--pre-dump)
> Run criu dump
> Send the 15 signal to  53
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.100000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.200000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.400000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.800000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 1.600000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 3.200000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 6.400000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 12.800000
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 25.600000
>  Test zdtm/static/net_lock_socket_iptables FAIL at zdtm/static/net_lock_socket_iptables die 
> Send the 9 signal to  53
> Wait for zdtm/static/net_lock_socket_iptables(53) to die for 0.100000
> Running zdtm/static/net_lock_socket_iptables.hook(--clean)
> ```

This is most likely due to the ```--norst``` option passed to zdtm in dump-only and crit jenkins pipelines. The network lock tests are highly dependant on dump/restore hooks. I think it hangs because it needs post-restore hook in order to function properly.
I will send a patch shortly to exclude those tests from the dump-only and crit jenkins pipelines.

avagin:
There are a few other jobs that fail too.

ZeyadYasser:
> There are a few other jobs that fail too.


I think the other tests are failing for unrelated reasons.
- Alpine and Arch tests are failing because of ```zdtm/static/uffd-events```
- Vagrant Fedora based test (no VDSO) is failing because of ```zdtm/static/socket-tcp-reuseport```

I believe jenkins/crit & jenkins/dump-only are failing for reasons I mentioned.

avagin:
> I will send a patch shortly to exclude those tests from the dump-only and crit jenkins pipelines.

You need to add the reqrst flag.

rst0git:
@ZeyadYasser 
```diff
index d3b82067f..c34398a4a 100644
--- a/test/zdtm/static/net_lock_socket_nftables6.desc
+++ b/test/zdtm/static/net_lock_socket_nftables6.desc
@@ -1,6 +1,6 @@
 {
     'flavor': 'h',
-    'flags': 'suid excl',
+    'flags': 'suid excl reqrst',
     'feature': 'network_lock_nftables',
     'dopts': '--tcp-established --network-lock nftables',
     'ropts': '--tcp-established',
```

criu/kerndat.c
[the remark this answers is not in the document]
Snorch: I believe that there is still a race window here... One kerndat_has_nftables_concat can remove the table under another kerndat_has_nftables_concat, and another would fail on no table and wrongly set kdat.has_nftables_concat = false and will save it to kerndat cache.

Actually running multiple migrations of different containers in parallel can be quite a usecase for cloud cluster load balancing.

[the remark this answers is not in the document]
avagin: kerndat is cached, so it has to be collected once. But I agree that this looks like a race and we need to generate a unique name for this test table.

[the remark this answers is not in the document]
ZeyadYasser: Thank you for pointing this out. https://github.com/checkpoint-restore/criu/pull/1588

PR #1541, opened 2021-07-05
restore: cleanup cgroup properly in error path

liusdu:
Cgroup yard is setup in crtools_prepare_shared. But it is not cleaned up
 properly in some code path, which leads cgroup mountpoint leaking.

We can see this type of errors in runc's issue:https://github.com/opencontainers/runc/issues/3042

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -253,9 +253,6 @@ static int crtools_prepare_shared(void)
 	if (tty_prep_fds())
 		return -1;
 
-	if (prepare_cgroup())
-		return -1;
-
 	return 0;
 }
 
@@ -2459,8 +2456,6 @@ skip_ns_bouncing:
 	if (restore_freezer_state())
 		pr_err("Unable to restore freezer state\n");
 
-	fini_cgroup();
-
 	/* Detaches from processes and they continue run through sigreturn. */
 	if (finalize_restore_detach())
 		goto out_kill_network_unlocked;
@@ -2507,7 +2502,6 @@ out_kill:
 	}
 
 out:
-	fini_cgroup();
 	depopulate_roots_yard(mnt_ns_fd, true);
 	stop_usernsd();
 	__restore_switch_stage(CR_STATE_FAIL);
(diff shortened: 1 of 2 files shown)


criu/cr-restore.c
```diff
@@ -2598,12 +2595,14 @@ int cr_restore_tasks(void)
...
-		goto err;
+		goto clean_cgroup;
 
 	if (prepare_lazy_pages_socket() < 0)
-		goto err;
+		goto clean_cgroup;
 
 	ret = restore_root_task(root_item);
+clean_cgroup:
+	fini_cgroup();
```
0x7f454c46, line 2605: Nit: could you also move `prepare_cgroup()` out of `crtools_prepare_shared()` as the last one is static and used only here - it would make cgroup prepare/fini pair and therefore a bit more readable.
Otherwise, the patch makes sense to me.

liusdu: Done. What a about the this  version?

avagin: I think we need to call prepare_cgroup from cr_restore_tasks to match fini_cgroup:

```
diff --git a/criu/cr-restore.c b/criu/cr-restore.c
index f7dc9d279..2fd238b0a 100644
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -253,9 +253,6 @@ static int crtools_prepare_shared(void)
        if (tty_prep_fds())
                return -1;
 
-       if (prepare_cgroup())
-               return -1;
-
        return 0;
 }
 
@@ -2592,18 +2589,20 @@ int cr_restore_tasks(void)
                goto err;
 
        if (crtools_prepare_shared() < 0)
-               goto clean_cgroup;
+               goto err;
+
+       if (prepare_cgroup())
+               goto err;
 
        if (criu_signals_setup() < 0)
-               goto clean_cgroup;
+               goto err;
 
        if (prepare_lazy_pages_socket() < 0)
-               goto clean_cgroup;
+               goto err;
 
        ret = restore_root_task(root_item);
-clean_cgroup:
-       fini_cgroup();
 err:
+       fini_cgroup();
        cr_plugin_fini(CR_PLUGIN_STAGE__RESTORE, ret);
        return ret;
 }
```

liusdu: Removing label `clean_cgroup`  leads to call `fini_cgroup`  on all error paths.. Should we add it back to let  `prepare_cgroup` and `fini_cgroup` match?

avagin (approved):

0x7f454c46 (approved):

PR #1543, opened 2021-07-05
ci: combine cross compile container definitions

adrianreber:
The cross compile container definitions for each architecture were almost the same files except for the architecture.

This moves the architecture to variables so that all cross compile setups can use the same container definition.


scripts/build/Dockerfile.aarch64-cross.hdr
```diff
@@ -0,0 +1,5 @@
+FROM docker.io/dockcross/base:latest
+
+ENV ARCH=aarch64
+ENV DEBIAN_ARCH=arm64
+ENV CROSS_TRIPLE=aarch64-linux-gnu
```
rppt, line 5: Nit:  maybe we can update `TRIPLE` to commonly used `TRIPLET` as we already change things here?
Otherwise looks good to me

rppt (approved):
Acked-by: Mike Rapoport <rppt@linux.ibm.com>

scripts/build/Dockerfile.aarch64-cross.hdr
adrianreber: done

scripts/build/Makefile
```diff
@@ -1,11 +1,8 @@
-ARCHES := x86_64 fedora-asan fedora-rawhide centos7 armv7hf centos8
+ARCHES := x86_64 fedora-asan fedora-rawhide centos7 armv7hf centos8 armv7-cross aarch64-cross ppc64-cross mips64el-cross
```
0x7f454c46, line 1: Yeah, I remember why I added it to TARGETS separately: now you add cross-builds to TARGETS_CLANG, which won't work properly as long as there isn't clang cross-compiler installed.

0x7f454c46 (commented):
I love this cleanup.

scripts/build/Makefile
adrianreber: I changed it to split non clang targets (cross compilation) away from clang targets. Please have a look.

0x7f454c46 (approved):

.gitignore
```diff
@@ -42,4 +42,3 @@ lib/.crit-setup.files
 compel/include/asm
 include/common/asm
 include/common/config.h
-build/
```
rst0git, line 45: @adrianreber why `build/` has been removed in this commit?

adrianreber: I am not sure any more but I think because new files in scripts/build were hidden from git. As I added files in scripts/build it seemed wrong.

Do you see problems with this change?

rst0git: The `build` folder is created by `make install`. This line was introduced with commit https://github.com/checkpoint-restore/criu/commit/967797a.

rst0git: I opened a PR to add `build/` in gitignore: https://github.com/checkpoint-restore/criu/pull/1609

PR #1545, opened 2021-07-05
criu: introduce uncrustify to format source code

adrianreber:
This is another attempt to introduce a tool to format CRIU's source code. This time it is based on uncrustify. I am regularly using
uncrustify on my patches if possible and the uncrustify configuration file is based on the uncrustify linux.cfg example with minimal changes for CRIU.

This runs uncrustify only on a small number of files to see if it works as expected and can later be extended to a larger number of files.

The changes look correct and if this is accepted I would extend it to more files in the CRIU code base.

The biggest changes will probably happen around declaration or initialization of structures. Sometimes they are space or tab aligned, sometimes not. Maybe a tool like uncrustify would help us to get consistent code style.

Personally which style we have is not that important to me, but having a consistent style would be nice. Happy to change things in the configuration file. I tried to go with a style which changes the existing code as little as possible.

@mihalicyn similar to your attempt to introduce C code formatting

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -5,9 +5,21 @@ on: [push, pull_request]
 jobs:
   build:
     runs-on: ubuntu-latest
+    container:
+      image: registry.fedoraproject.org/fedora:latest
     steps:
-    - uses: actions/checkout@v2
     - name: Install tools
-      run: sudo apt-get install -qqy flake8 shellcheck
+      run: sudo dnf -y install git make python3-flake8 ShellCheck uncrustify which
+    - uses: actions/checkout@v2
     - name: Run make lint
       run: make lint
+    - name: Run make indent
+      run: >
+        make indent &&
+        STATUS=$(git status --porcelain) &&
+        if [ ! -z "$STATUS" ]; then
+          echo "FAIL: some files are not correctly formatted.";
+          echo "$STATUS"
+          echo "FAIL: please run 'make indent'";
+          exit 1;
+        fi
--- a/Makefile
+++ b/Makefile
@@ -202,7 +202,7 @@ criu-deps	+= include/common/asm
 #
 # Configure variables.
 export CONFIG_HEADER := include/common/config.h
-ifeq ($(filter tags etags cscope clean mrproper,$(MAKECMDGOALS)),)
+ifeq ($(filter tags etags cscope clean lint indent help mrproper,$(MAKECMDGOALS)),)
 include Makefile.config
 else
 # To clean all files, enable make/build options here
@@ -408,6 +408,8 @@ help:
 	@echo '      test            - Run zdtm test-suite'
 	@echo '      gcov            - Make code coverage report'
 	@echo '      unittest        - Run unit tests'
+	@echo '      lint            - Run code linters'
+	@echo '      indent          - Indent C code'
 .PHONY: help
 
 lint:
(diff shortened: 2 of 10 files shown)


rppt:
On Mon, Jul 05, 2021 at 08:21:44AM -0700, Adrian Reber wrote:
> This is another attempt to introduce a tool to format CRIU's source code. This
> time it is based on uncrustify. I am regularly using
> uncrustify on my patches if possible and the uncrustify configuration file is
> based on the uncrustify linux.cfg example with minimal changes for CRIU.
> 
> This runs uncrustify only on a small number of files to see if it works as
> expected and can later be extended to a larger number of files.
> 
> The changes look correct and if this is accepted I would extend it to more
> files in the CRIU code base.
> 
> The biggest changes will probably happen around declaration or initialization
> of structures. Sometimes they are space or tab aligned, sometimes not. Maybe a
> tool like uncrustify would help us to get consistent code style.
> 
> Personally which style we have is not that important to me, but having a
> consistent style would be nice. Happy to change things in the configuration
> file. I tried to go with a style which changes the existing code as little as
> possible.
> 
> @mihalicyn similar to your attempt to introduce C code formatting

I think that before moving forward with tooling around the coding style, we
need to decide whether we want to update the existing code or we'll just
enforce the coding style only for new changes, then we should define the
coding style more precisely than "almost like Linux".

-- 
Sincerely yours,
Mike.

adrianreber:
> I think that before moving forward with tooling around the coding style, we need to decide whether we want to update the existing code or we'll just enforce the coding style only for new changes,

Although CRIU is already following the Linux coding style in many parts changing existing code will introduce a really big patch. From what I have seen the most problems are around initializing and defining structures or nicely aligned variable names. Also macros would probably change a lot.

For me the actual coding style is not important, but I really would love to see a tool which does it for us and can verify it on pull requests.

Tooling wise 'only for new changes' is difficult. It is doable to a certain degree, but not easily. Introducing something like checkpatch also sounds like too much work for the small community around CRIU.

Changing the existing code sounds more practical. Not all files at once was my approach as in this PR.

> then we should define the coding style more precisely than "almost like Linux".

Again, for me the actual coding style is not important. I would prefer to have a tool to verify the coding style whatever it is. I am not sure there is a Linux kernel coding style which defines everything so to a certain degree we have to do something almost like Linux or whatever the tool we select does.

Using Rust, Golang and Python it is just nice to have a tool which does the formatting for you and can verify it on pull requests.

Maybe using `clang-format` as seen in #1196 makes more sense as it based on the file `.clang-format` which is directly copied from the Linux kernel.

0x7f454c46:
Hmm, just wondering why so:
```
-	struct fd_id *n;
+	struct fd_id *	n;
```
vs
```
-	struct page_read * prp;
+	struct page_read *prp;
```
seems a little bit random.
I also liked align in setup_sas(), but I can live without it.

adrianreber:
> Hmm, just wondering why so:
> 
> ```
> -	struct fd_id *n;
> +	struct fd_id *	n;
> ```
> 
> vs
> 
> ```
> -	struct page_read * prp;
> +	struct page_read *prp;
> ```
> 
> seems a little bit random.

The one thing is controlled by how to align variables in a struct and the other is how align variables outside of a struct. The other PR based on clang-format (#1546) is probably better than this one.

> I also liked align in setup_sas(), but I can live without it.

I think alignments like the ones in `setup_sas()` are the biggest things that will change.

adrianreber:
Closing in favour of #1546

PR #1546, opened 2021-07-07
criu: introduce clang-format to format source code

adrianreber:
This is another attempt to introduce a tool to format CRIU's source code. This time it is based on clang-format.

The .clang-format file is taken from the linux kernel git tree.

I removed all comments from lines which state that it requires at least clang-format 4 or 5. This change is done using clang-format 12.

I also changed the maximum line length from 80 to 120 and adapted the list of ForEachMacros to the one from Alexander Mikhalitsyn clang-format pull request.

This runs clang-format only on a small number of files to see if it works as expected and can later be extended to a larger number of files.

See #1545 #1196

The result from clang-format is very similar to uncrustify, but the advantage is that it is based on the `.clang-format` file taken from the Linux kernel.


.clang-format
```diff
@@ -0,0 +1,140 @@
+# SPDX-License-Identifier: GPL-2.0
+#
+# clang-format configuration file. Intended for clang-format >= 4.
```
avagin, line 3: Let's  add in this comment:
* I think we must mention that it has been taken from the Linux kernel.
* All changes that have been done comparing with the kernel version.

adrianreber: done

rppt:
> This is another attempt to introduce a tool to format CRIU's source code. This time it is based on clang-format.
> 
> The .clang-format file is taken from the linux kernel git tree.
> 
> I removed all comments from lines which state that it requires at least clang-format 4 or 5. This change is done using clang-format 12.

Worked for me with `clang-format 11.0.0 (Fedora 11.0.0-3.fc33)` as well.

> I also changed the maximum line length from 80 to 120 and adapted the list of ForEachMacros to the one from Alexander Mikhalitsyn clang-format pull request.

I wonder if we can't do something like

```
curl https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/.clang-format | ./scripts/fixup-clang-format
```
to avoid maintaining our own copy?

The script that removes comments, changes maximal line width and adds one or two `ForEachMacros` should be quite simple.
Leaving the whole bunch of `ForEachMacros` from Linux won't be a problem, IMHO.

> This runs clang-format only on a small number of files to see if it works as expected and can later be extended to a larger number of files.

I think that once we get a consensus on a tool and its config, reformatting everything in one huge commit is better that doing this several files at a time. I did some experiments with checkpatch.pl and we are only 5k lines away from strict Linux  coding style. So that automated style update won't be that big and having it in a single commit will ease the future `git blame`

> The result from clang-format is very similar to uncrustify, but the advantage is that it is based on the `.clang-format` file taken from the Linux kernel.

adrianreber:
@rppt I like the idea of `curl | fixup-clang-format`. I will update this PR.

adrianreber:
I changed the .clang-format file to be created by `make .clang-format` by downloading the latest version from kernel tree and adapting to our needs.

rppt:
> I changed the .clang-format file to be created by `make .clang-format` by downloading the latest version from kernel tree and adapting to our needs.

Nit: I'd use a well know version from a tagged kernel, e.g.

```
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/.clang-format?h=v5.13
```

criu/cr-dedup.c
```diff
@@ -14,7 +14,7 @@ int cr_dedup(void)
 {
 	int close_ret, ret = 0;
 	unsigned long img_id;
-	DIR * dirp;
```
avagin, line 17: pls move code changes to a separate patch.

rppt: Right, it should be three commits IMHO:
* add scripts and `Makefile` target for coding style updates
* run 'make indent' and commit the change
* add CI/CD checks to verify the coding style

The `indent` target in the `Makefile` could be something like
```
indent:
	find . -name '*.[chS]' -type f -exec clang-format -i {} \;
```

BTW, this would change ~10% of the code: 

```
:~/git/criu$ wc -l $(find . -name '*.[chS]' -type f ) | grep total
 158904 total
:~/git/criu$ git diff --stat | grep "files changed"
 825 files changed, 15641 insertions(+), 18587 deletions(-)

Comment (adrianreber) [reply criu/cr-dedup.c:17]: Made three commits as suggested.

**Comment by adrianreber** [comment]
I changed this PR to actually run `make indent' on all *.c and *.h files.

I had to disabled clang-format for a few assembler macros because clang-format does correctly handle those.

I am also not running it on assembler files because clang-format breaks assembler files.

And now I see that CI is also not really happy with my changes. Will update this until CI is happy.

**Review by rppt — APPROVED** [review]
I only skimmed over the actual style changes, but as long as they are automated and consistent I think it's ok.

As a follow up it would be nice to update coding style descriptions in CONTRIBUTING.md

Acked-by: Mike Rapoport `<rppt@linux.ibm.com>`

**Comment by adrianreber** [comment]
@rppt Thanks for your help. This looks like a PR that needs to rebased whenever another PR enters the repository. So it would make it easier to merge it as soon as possible.

As this is a big change I would like to get some more approvals from @checkpoint-restore/maintainers before merging it. PTAL.

**Comment by adrianreber** [comment]
CentOS 7 tests will fail because of #1532

File: compel/arch/x86/src/lib/include/uapi/asm/cpu.h [hunk 3]
```diff
@@ -15,334 +15,334 @@
...
+#define X86_FEATURE_SYSCALL (1 * 32 + 11) /* SYSCALL/SYSRET */
+#define X86_FEATURE_MP (1 * 32 + 19) /* MP Capable */
+#define X86_FEATURE_NX (1 * 32 + 20) /* Execute Disable */
+#define X86_FEATURE_MMXEXT (1 * 32 + 22) /* AMD MMX extensions */
+#define X86_FEATURE_FXSR_OPT (1 * 32 + 25) /* FXSAVE/FXRSTOR optimizations */
+#define X86_FEATURE_GBPAGES (1 * 32 + 26) /* "pdpe1gb" GB pages */
+#define X86_FEATURE_RDTSCP (1 * 32 + 27) /* RDTSCP */
+#define X86_FEATURE_LM (1 * 32 + 29) /* Long Mode (x86-64, 64-bit support) */
+#define X86_FEATURE_3DNOWEXT (1 * 32 + 30) /* AMD 3DNow extensions */
+#define X86_FEATURE_3DNOW (1 * 32 + 31) /* 3DNow */
```
Snorch, line 88: Not sure about this one, do we really want to enforce no human readable aligned neighbor lines (for macro values or comments or something else)?

Can we use `AlignConsecutiveMacros: true`?

```
--- a/scripts/fetch-clang-format.sh
+++ b/scripts/fetch-clang-format.sh
@@ -11,4 +11,5 @@ curl -s "${URL}" | sed -e "
        s,ColumnLimit: 80,ColumnLimit: 120,g;
        s,Intended for clang-format >= 4,Intended for clang-format >= 11,g;
        s,ForEachMacros:,ForEachMacros:\n  - 'for_each_pstree_item',g;
+       s,\(AlignTrailingComments:.*\)$,\1\nAlignConsecutiveMacros: true,g;
 "  > .clang-format
```

Snorch (changes requested):
Generally I like this change. Hope it would not break something =) Looking here and there I see only valid changes in code. Though see the comment about define alignment.

scripts/fetch-clang-format.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+# This script fetches .clang-format from the Linux kernel
+# and slightly adapts it for CRIU.
+
+URL="https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/plain/.clang-format?h=v5.13"
+
+curl -s "${URL}" | sed -e "
+	s,^\(\ *\)#\([A-Z]\),\1\2,g;
```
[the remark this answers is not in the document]
adrianreber: Thanks. Will try that.

compel/arch/x86/src/lib/include/uapi/asm/cpu.h
adrianreber: I can add it. No Problem.

Makefile
```diff
@@ -439,6 +441,13 @@ codecov:
 	bash <(curl -s https://codecov.io/bash)
 .PHONY: codecov
 
+.clang-format: .FORCE
+	scripts/fetch-clang-format.sh
```
0x7f454c46, line 445: Nit:
```
	$(E) " .clang-format"
	$(Q) scripts/fetch-clang-format.sh
```

Makefile
```diff
@@ -439,6 +441,13 @@ codecov:
...
+
+indent:
```
0x7f454c46, line 447: Should it depend on `.clang-format`?
If not, maybe it should be a target instead of file-rule, like `fetch-clang-format`?
No hard feelings though.

adrianreber:
@rppt Are you still okay with this after the changes @Snorch requested?

compel/arch/x86/src/lib/handle-elf.c
```diff
@@ -5,10 +5,8 @@
 #include "piegen.h"
 #include "log.h"
 
-static const unsigned char __maybe_unused
-elf_ident_64_le[EI_NIDENT] = {
-	0x7f, 0x45, 0x4c, 0x46, 0x02, 0x01, 0x01, 0x00,
-	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+static const unsigned char __maybe_unused elf_ident_64_le[EI_NIDENT] = {
+	0x7f, 0x45, 0x4c, 0x46, 0x02, 0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
 };
```
0x7f454c46, line 10: If we go with this, can we add some comment to keep this aligned?
As on the example of code_syscall[]:
```
static const unsigned char __maybe_unused elf_ident_64_le[EI_NIDENT] = {
	0x7f, 0x45, 0x4c, 0x46, 0x02, 0x01, 0x01, 0x00, /* clang-format */
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
};
```

0x7f454c46 (commented):
I see value in forcing one code-style over all sources.
But on the other side, I also see a value (aka code is more readable) in some of the alignments.
I wish it was possible to add a file with "ignore those lines" or a file with existing warnings on codebase so that after doing `make indent` we could compare/diff if there is anything new.
That would also make commits here smaller.
Not insisting though.

Makefile
adrianreber: done

Makefile
adrianreber: changed

compel/arch/x86/src/lib/handle-elf.c
adrianreber: also changed

Snorch (approved):

rppt:
> @rppt Are you still okay with this after the changes @Snorch requested?

Yep, I'm fine

rppt (approved):

adrianreber:
I moved the 'flog/' indentation into a separate commit for an easier release as we currently do not plan to bring `flog/` to the release branch.

Currently this has two approvers. @0x7f454c46 and @avagin if you add your approval I think this could be merged.

Would be good to get this PR in before the GSoC commits to avoid a lot of rebasing.

0x7f454c46 (approved):
Thanks for doing this, it seems quite an amount of work.
I understand that it's auto-generated change, so should be safe, but probably to double-check could you verify that the criu binary and PIEs generated are the same? Like diff saying objdumps are the same or even checksum (for this CRIU_GITID should be same)?
Not that I expect any difference, but just as it's impossible to cross all code changes with eyes. 
Other than that, LGTM.

adrianreber:
> Thanks for doing this, it seems quite an amount of work.
> I understand that it's auto-generated change, so should be safe, but probably to double-check could you verify that the criu binary and PIEs generated are the same? Like diff saying objdumps are the same or even checksum (for this CRIU_GITID should be same)?
> Not that I expect any difference, but just as it's impossible to cross all code changes with eyes.
> Other than that, LGTM.

I am not able to get the same md5sum, I am using the same CRIU_GITID for both builds, but something else must be changing the binary. But the diff of a objdump is identical:
```
$ objdump -t criu/criu > unindented
$ objdump -t criu/criu > indented
$ diff -u indented unindented
<nothing>
```

Is that objdump enough or should I use some other parameters?

0x7f454c46:
> $ objdump -t criu/criu > unindented
> [ git checkout and make]
> $ objdump -t criu/criu > indented
> $ diff -u indented unindented
> <nothing>
> ```
> 
> Is that objdump enough or should I use some other parameters?

`objdump -d` sounds good enough for me.

adrianreber:
objdump -d gives me a diff with differences around movs

[objdump.diff.log](https://github.com/checkpoint-restore/criu/files/6932120/objdump.diff.log)

.clang-format
```diff
@@ -0,0 +1,563 @@
...
+PenaltyBreakBeforeFirstCallParameter: 30
+PenaltyBreakComment: 10
+PenaltyBreakFirstLessLess: 0
+PenaltyBreakString: 10
+PenaltyExcessCharacter: 100
+PenaltyReturnTypeOnItsOwnLine: 60
+
+PointerAlignment: Right
+ReflowComments: false
+SortIncludes: false
```
rst0git, line 544: Could we set `SortIncludes: true`?

adrianreber: I have no preferences, but I think if we already have so many approvals I would prefer to get it merged as soon as possible and if changes are requested we can open smaller PRs for it. If all current approvers want to have this change I can also do it right now, but I would like to see this big PR merged as soon as possible now.

rst0git (approved):

adrianreber:
Thanks everyone for getting this done.

Issue #1524, opened 2021-06-22
Unable to restore into a different cgroup


kolyshkin:
**Description**

I am working on runc, and we're experiencing occasional failures during the lazy migration test (https://github.com/opencontainers/runc/issues/2924, https://github.com/opencontainers/runc/issues/2760).

The nature of the lazy migration test requires the original container to stay until we finish restoring
the new one, meaning we have two cgroups (and systemd units, in case systemd is used).

I suspect the cause of failures we see is a race between runc+criu restoring a container into
the same cgroup as the original container was in, and systemd deciding to remove that cgroup.

I tried two ways of restoring into a different cgroup:

1. Using the swrk analog of `--cgroup-root`, described at https://criu.org/CGroups#Restoring_into_different_CGroups. Setting a different value(s) on restore apparently doesn't do anything.
2. Putting the `criu swrk` into a proper cgroup on restore, and using the swrk analog of `--manage-cgroup-mode ignore`, described at https://criu.org/CGroups#CGroups_restoring_strategy. As I understand, it should just skip dealing with cgroups (i.e. do not do anything related to cgroups).

I forgot the details but from what I remember neither method work (at least on cgroup v2).

I will do more tests and add further details tomorrow.

PR #1550, opened 2021-07-11
cgroup: cgroup_contains has to update the mask for cgroupv2

avagin:
Fixes: #1524

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -196,8 +196,14 @@ static bool cgroup_contains(char **controllers,
 	bool all_match = true;
 
 	/* Check whether this cgroup2 or not.*/
-	if (n_controllers == 1 && controllers[0][0] == 0)
-		return name[0] == 0;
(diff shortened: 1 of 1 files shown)


criu/cgroup.c
```diff
@@ -196,8 +196,14 @@ static bool cgroup_contains(char **controllers,
...
-	if (n_controllers == 1 && controllers[0][0] == 0)
-		return name[0] == 0;
+	if (n_controllers == 1 && controllers[0][0] == 0) {
+		bool match = name[0] == 0;
+
+		if (mask && match)
+			*mask &= ~(1ULL);
+
+		return match;
+	}
```
rst0git, line 206: Should we add a test for this?

avagin: I was thinking about it. It is unclear how we can test this in the zdtm test. It isn't specific for this patch. We don't have a good test for cgroupv2.

rst0git (approved):

PR #1519, opened 2021-06-18
RFC: AMD ROCm support with plugin 

rajbhar:
Hello CRIU Community and Maintainers,

Can you please review this "RFC" pull request as we are still adding more features and making some changes to our API IOCTL design. Further to our earlier announcement https://www.spinics.net/lists/amd-gfx/msg62112.html , we made some more enhancements in terms of functionalities and based on some feedback that we received on the mailing list with regards to placement of the plugin and how it should be built etc.

This PR includes another important enhancement that brings in multi-GPU support.  The checkpoint / restore time is improved now compared to what we had earlier when we made the announcement, as we added multithreading support and we are further optimizing it using the DMA engines, to copy VRAM contents to image files. This will be added as soon as its ready, probably on top of the feedback we receive here.  There are some changes made to core CRIU itself and one possible recent regression was noticed during our testing so we seek your guidance as we are not sure if we understand that well enough.

This work was tested on multi GPU AMD systems with real pytorch workloads such as BERT(Transformers) and BERT Large. We also tested some Tensorflow workloads and found some probable limitations of protobuf that its unable  to handle buffer objects more than 4GB. We are still investigating this more but would like to inquire here if any such issue was observed or reported before?

Link to the kernel changes:     https://github.com/RadeonOpenCompute/ROCK-Kernel-Driver/commits/fxkamd/criu-wip

We again thank you for your time and appreciate your feedback.

PR #1556, opened 2021-07-15
Amdgpu rfc 210715 1

rajbhar:
Hi @checkpoint-restore/maintainers ,

Thank you for the review feedback on our previous RFC PR - https://github.com/checkpoint-restore/criu/pull/1519.

As per the recommendation, we have split our work into two parts and we'll close the previous PR, #1519  
1 - This Pull Request
      *  brings in the proposed changes to CRIU
      *  new hooks and a dummy plugin
      *  some workaround/fixes for issues seen with CRIU to support amdgpu_plugin
   
2 - https://github.com/RadeonOpenCompute/criu/commits/amdgpu_rfc-210715-2
      *  changes that implement actual amdgpu_plugin
      *  several features such as multi gpu support, fast checkpoint restore using SDMA. 
      *  fixes for lint, clang, fedora32, ubuntu focal build failures.
      *  sets up build dependencies on libdrm which is required for amdgpu_plugin
   
There will be another PR with part 2.

Thanks again for your time and patience.

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2388,6 +2388,29 @@ skip_ns_bouncing:
 		pr_err("Unable to flush breakpoints\n");
 
 	finalize_restore();
+	/*
+	 * Some external devices such as GPUs might need a very late
+	 * trigger to kick-off some events, memory notifiers and for
+	 * restarting the previously restored queues during criu restore
+	 * stage. This is needed since criu pie code may shuffle VMAs
+	 * around so things such as registering MMU notifiers (for GPU
+	 * mapped memory) could be done sanely once the pie code hands
+	 * over the control to master process.
+	 */
+	for_each_pstree_item(item) {
+		pr_info("Run late stage hook from criu master for external devices\n");
+		ret = run_plugins(RESUME_DEVICES_LATE, item->pid->real);
+		/*
+		 * This may not really be an error. Only certain plugin hooks
+		 * (if available) will return success such as amdgpu_plugin that
+		 * validates the pid of the resuming tasks in the kernel mode.
+		 * Most of the times, it'll be -ENOTSUP and in few cases, it
+		 * might actually be a true error code but that would be also
+		 * captured in the plugin so no need to print the error here.
+		 */
+		if (ret < 0)
+			pr_debug("restore late stage hook for external plugin failed\n");
+	}
 
 	ret = run_scripts(ACT_PRE_RESUME);
 	if (ret)
--- a/criu/file-ids.c
+++ b/criu/file-ids.c
@@ -77,8 +77,14 @@ int fd_id_generate_special(struct fd_parms *p, u32 *id)
 
 		fi = fd_id_cache_lookup(p);
 		if (fi) {
-			*id = fi->id;
-			return 0;
+			if (p->stat.st_mode & (S_IFCHR | S_IFBLK)) {
+				/* Don't cache the id for mapped devices */
+				*id = fd_tree.subid++;
+				return 1;
+			} else {
+				*id = fi->id;
+				return 0;
+			}
 		}
 	}
 
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -2267,6 +2267,23 @@ static int open_filemap(int pid, struct vma_area *vma)
 	BUG_ON((vma->vmfd == NULL) || !vma->e->has_fdflags);
 	flags = vma->e->fdflags;
 
+	/* update the new device file page offsets and file paths set during restore */
+	if (vma->e->status & VMA_UNSUPP) {
+		uint64_t new_pgoff;
+		char new_path[PATH_MAX];
+		int ret;
+
+		struct reg_file_info *rfi = container_of(vma->vmfd, struct reg_file_info, d);
+		ret = run_plugins(UPDATE_VMA_MAP, rfi->rfe->name, new_path, vma->e->start, vma->e->pgoff, &new_pgoff);
+		if (ret == 1) {
+			pr_info("New mmap %#016" PRIx64 "->%#016" PRIx64 " path %s\n", vma->e->pgoff, new_pgoff,
+				new_path);
+			vma->e->pgoff = new_pgoff;
+			rfi->path = xstrdup(new_path);
+			pr_debug("Updated rfi->path %s\n", rfi->path);
+		}
+	}
+
 	if (ctx.flags != flags || ctx.desc != vma->vmfd) {
 		if (vma->e->status & VMA_AREA_MEMFD)
 			ret = memfd_open(vma->vmfd, &flags);
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -45,6 +45,7 @@
 #include "protobuf.h"
 #include "images/fdinfo.pb-c.h"
 #include "images/mnt.pb-c.h"
+#include "plugin.h"
 
 #include <stdlib.h>
 
@@ -101,6 +102,19 @@ static bool __is_vma_range_fmt(char *line)
 bool is_vma_range_fmt(char *line)
 {
 	return __is_vma_range_fmt(line);
+}
+
+bool handle_vma_plugin(int *fd, struct stat *stat)
+{
+	int ret;
+
+	ret = run_plugins(HANDLE_DEVICE_VMA, *fd, stat);
+	if (ret < 0) {
+		pr_perror("handle_device_vma plugin failed");
+		return false;
+	}
+
+	return true;
 }
 
 static void __parse_vmflags(char *buf, u32 *flags, u64 *madv, int *io_pf)
@@ -188,6 +202,7 @@ struct vma_file_info {
 	int dev_min;
 	unsigned long ino;
 	struct vma_area *vma;
+	bool has_device_plugin;
 };
 
 static inline int vfi_equal(struct vma_file_info *a, struct vma_file_info *b)
@@ -577,11 +592,17 @@ static int handle_vma(pid_t pid, struct vma_area *vma_area, const char *file_pat
 	} else if (*vm_file_fd >= 0) {
 		struct stat *st_buf = vma_area->vmst;
 
-		if (S_ISREG(st_buf->st_mode))
+		if (S_ISREG(st_buf->st_mode)) {
 			/* regular file mapping -- supported */;
-		else if (S_ISCHR(st_buf->st_mode) && (st_buf->st_rdev == DEVZERO))
+			pr_debug("Found regular file mapping, OK\n");
+		} else if (S_ISCHR(st_buf->st_mode) && (st_buf->st_rdev == DEVZERO)) {
 			/* devzero mapping -- also makes sense */;
-		else {
+			pr_debug("Found devzero mapping, OK\n");
+		} else if (handle_vma_plugin(vm_file_fd, st_buf)) {
+			pr_info("Found device file mapping, plugin is available\n");
+			vfi->has_device_plugin = true;
+		} else {
+			/* non-regular mapping with no supporting plugin */
 			pr_err("Can't handle non-regular mapping on %d's map %" PRIx64 "\n", pid, vma_area->e->start);
 			goto err;
 		}
@@ -646,9 +667,23 @@ static int vma_list_add(struct vma_area *vma_area, struct vm_area_list *vma_area
 			struct vma_file_info *vfi, struct vma_file_info *prev_vfi)
 {
 	if (vma_area->e->status & VMA_UNSUPP) {
-		pr_err("Unsupported mapping found %016" PRIx64 "-%016" PRIx64 "\n", vma_area->e->start,
-		       vma_area->e->end);
-		return -1;
+		if (vfi->has_device_plugin) {
+			/* Unsupported VMAs that provide special plugins for
+			 * backup can be treated as regular VMAs and criu
+			 * should only save their metadata in the dump files.
+			 * There can be several special backup plugins hooks
+			 * that might run at different stages during checkpoint
+			 * and restore.
+			 */
+			pr_debug("Device file mapping %016" PRIx64 "-%016" PRIx64 " "
+				 "must be supported via device plugins\n",
+				 vma_area->e->start, vma_area->e->end);
+
+		} else {
+			pr_err("Unsupported mapping found %016" PRIx64 "-%016" PRIx64 "\n", vma_area->e->start,
+			       vma_area->e->end);
+			return -1;
+		}
 	}
 
 	/* Add a guard page only if here is enough space for it */
--- a/plugins/amdgpu/Makefile
+++ b/plugins/amdgpu/Makefile
@@ -1,0 +1,13 @@
+all: dummy_plugin.so
+
+dummy_plugin.so: dummy_plugin.c
+	gcc -g -Werror -D _GNU_SOURCE -Wall -shared -nostartfiles dummy_plugin.c -o dummy_plugin.so -iquote ../../../criu/include -iquote ../../criu/include -fPIC
+
+clean:
+	$(Q) $(RM) dummy_plugin.so
+install:
+	$(Q) mkdir -p $(PLUGINDIR)
+	$(Q) install -m 644 dummy_plugin.so $(PLUGINDIR)
+
+uninstall:
+	$(Q) $(RM) $(PLUGINDIR)/dummy_plugin.so
--- a/plugins/amdgpu/dummy_plugin.c
+++ b/plugins/amdgpu/dummy_plugin.c
@@ -1,0 +1,36 @@
+#include <sys/stat.h>
+
+#include "criu-log.h"
+#include "criu-plugin.h"
+
+int dummy_plugin_handle_device_vma(int fd, const struct stat *stat)
+{
+	pr_info("dummy_plugin: Inside %s for fd = %d\n", __func__, fd);
+	/* let criu report failure for the unsupported mapping */
+	return -ENOTSUP;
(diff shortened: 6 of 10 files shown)


criu/cr-restore.c
```diff
@@ -2448,6 +2448,21 @@ static int restore_root_task(struct pstree_item *init)
...
+	 * trigger to kick-off some events, memory notifiers and for
+	 * restarting the previously restored queues during criu restore
+	 * stage. This is needed since criu pie code may shuffle VMAs
+	 * around so things such as registering MMU notifiers (for GPU
+	 * mapped memory) could be done sanely once the pie code hands
+	 * over the control to master process.
+	 */
+	for_each_pstree_item(item) {
+		pr_info("Run late stage hook from criu master for external devices\n");
+		ret = run_plugins(RESUME_DEVICES_LATE, item->pid->real);
```
adrianreber, line 2462: Not familiar enough with plugins myself, but is that correct to try to run the plugin always. More a general question if there isn't a flag which tells CRIU that a plugin exists. Or does `run_plugins()` handle it.

criu/file-ids.c
```diff
@@ -77,11 +77,18 @@ int fd_id_generate_special(struct fd_parms *p, u32 *id)
 {
 	if (p) {
 		struct fd_id *fi;
+		struct stat st_kfd;
 
 		fi = fd_id_cache_lookup(p);
 		if (fi) {
-			*id = fi->id;
-			return 0;
+			if (stat("/dev/kfd", &st_kfd) == -1) {
```
adrianreber, line 84: I saw `/dev/kfd` already a couple of times all over your patches. A `#define` would be a good idea for that string.

criu/files-reg.c
```diff
@@ -2314,6 +2314,23 @@ static int open_filemap(int pid, struct vma_area *vma)
...
+		int ret;
+
+		struct reg_file_info *rfi = container_of(vma->vmfd, struct reg_file_info, d);
+		ret = run_plugins(UPDATE_VMA_MAP, rfi->rfe->name, new_path, vma->e->start,
+					vma->e->pgoff, &new_pgoff);
+		if (ret == 1) {
+			pr_info("New mmap %#016"PRIx64"->%#016"PRIx64" path %s\n", vma->e->pgoff,
+					new_pgoff, new_path);
+			vma->e->pgoff = new_pgoff;
+			strcpy(rfi->path, new_path);
```
adrianreber, line 2330: This looks like a line static analysis could complain. Copying without size. Null termination not verified. Things like that.

criu/files-reg.c
```diff
@@ -2331,6 +2348,7 @@ static int open_filemap(int pid, struct vma_area *vma)
 
 	ctx.vma = vma;
 	vma->e->fd = ctx.fd;
+
```
adrianreber, line 2351: Unnecessary empty line.

criu/proc_parse.c
```diff
@@ -105,6 +106,18 @@ bool is_vma_range_fmt(char *line)
 	return __is_vma_range_fmt(line);
 }
 
+bool handle_vma_plugin(int *fd, struct stat *stat)
+{
+	int ret;
+	ret = run_plugins(HANDLE_DEVICE_VMA, *fd, stat);
```
adrianreber, line 112: I think we usually have an empty line between declaration and code.

plugins/amdgpu/Makefile
```diff
@@ -0,0 +1,14 @@
+all: dummy_plugin.so
+
+dummy_plugin.so: dummy_plugin.c
+	gcc -g -Werror -D _GNU_SOURCE -Wall -shared -nostartfiles dummy_plugin.c -o dummy_plugin.so -iquote ../../../criu/include -iquote ../../criu/include -fPIC
```
adrianreber, line 4: You should not hardcode CFLAGS but take them from CRIU's build environment. Not sure how to do that correctly, but it needs to be integrated in the existing Makefile infrastructure.

plugins/amdgpu/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+clean:
+	rm dummy_plugin.so
+
+install:
+	mkdir -p /var/lib/criu && cp dummy_plugin.so /var/lib/criu/
```
adrianreber, line 10: I am not happy with the hardcoded location. I know CRIU defaults to `/var/lib/criu`, but that sounds wrong anyway. Especially when looking at it from a Linux distribution package. Not sure what to here as the plugin infrastructure does not really exist. Also install must respect `$DESTDIR` for Linux distribution package generation.

plugins/amdgpu/dummy_plugin.c
```diff
@@ -0,0 +1,51 @@
...
+#include <sys/ioctl.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <stdint.h>
+
+#include "criu-log.h"
+#include "criu-plugin.h"
+
+#include "common/list.h"
+
```
adrianreber, line 18: Are all those includes really needed?

adrianreber (changes requested):

adrianreber:
And a test case to verify the plugin functionality is needed.

criu/cr-restore.c
rajbhar: I think run_plugin can handle if there is no plugin. This is a unique scenario when we want to give a final chance to all resuming tasks to do some special handling like restarting MMU notifications, events etc because *now* all the children of criu master restorer have their VMAs finalized and tasks are ready to run. In our case, the kernel mode driver makes sure if the pid indeed belongs to kfd and then only starts the previously setup (during criu restore stage) MMU notifies etc.  https://github.com/RadeonOpenCompute/ROCK-Kernel-Driver/commit/7ec3510634fdcdbca0c1f1673f31a4915d748ee2 and https://github.com/RadeonOpenCompute/criu/commit/a9979fabfd023a0fa2e993c350c2f0d451e330cd

criu/file-ids.c
rajbhar: OK

criu/files-reg.c
rajbhar: OK

criu/files-reg.c
rajbhar: ACK

criu/proc_parse.c
rajbhar: ACK

plugins/amdgpu/Makefile
rajbhar: Actually a future patch in the series 2 updates the Makefile quite a lot. Could you please take a look at https://github.com/RadeonOpenCompute/criu/blob/amdgpu_rfc-210715-2/plugins/amdgpu/Makefile  and advise

plugins/amdgpu/Makefile
rajbhar: I changed to $PLUGINDIR from criu env https://github.com/RadeonOpenCompute/criu/blob/amdgpu_rfc-210715-2/plugins/amdgpu/Makefile . Does that sound right?

plugins/amdgpu/dummy_plugin.c
rajbhar: I think criu-plugin.h is needed but I will recheck. Will remove list.h

rajbhar:
> And a test case to verify the plugin functionality is needed.

Right now, I think ZDTM already triggers the plugin calls. The one that is called for all resuming tasks is below. Other onces currently return ENOTSUPP so criu call try calling other hooks that ACK like amdgpu_plugin. I think if a ZDTM test case has a device file VMA, It should trigger the other plugin hooks already. I will test again with rtc example.

Below log is from an existing CI build and test run. This error is expected since there is no real implementation of this hook yet. Does this seem OK?

b'(00.050793) 301 was trapped'
b'(00.050797) 301 (native) is going to execute the syscall 11, required is 11'
b'(00.050829) 301 was stopped'
b'(00.050834) Run late stage hook from criu master for external devices'
b'(00.050836) Error (criu/cr-restore.c:2464): criu: restore late stage hook for external plugin failed'
b'(00.050838) Run late stage hook from criu master for external devices'

plugins/amdgpu/dummy_plugin.c
adrianreber: I would say most of them are unneeded.

adrianreber:
> > And a test case to verify the plugin functionality is needed.
> 
> Right now, I think ZDTM already triggers the plugin calls. The one that is called for all resuming tasks is below. Other ones currently return ENOTSUPP so criu call try calling other hooks that ACK like amdgpu_plugin. I think if a ZDTM test case has a device file VMA, It should trigger the other plugin hooks already. I will test again with rtc example.
> 
> Below log is from an existing CI build and test run. This error is expected since there is no real implementation of this hook yet. Does this seem OK?
> 
> ------------------------ grep Error ------------------------
> b'(00.050793) 301 was trapped'
> b'(00.050797) 301 (native) is going to execute the syscall 11, required is 11'
> b'(00.050829) 301 was stopped'
> b'(00.050834) Run late stage hook from criu master for external devices'
> b'(00.050836) Error (criu/cr-restore.c:2464): criu: restore late stage hook for external plugin failed'
> b'(00.050838) Run late stage hook from criu master for external devices'

Thanks for pointing to the CI logs. This is not really a good solution. We cannot introduce an error message into each restore just because a plugin does not exist. If there is no plugin there should be no error message. Only if the plugin failed to function correctly we should see an error message. I also think for your case (`ENOTSUP`) we should not see an error. Maybe a debug message if the plugin does not support something.

Concerning a test case. There should be a positive test case. Not just some 'Error' messages. As `/dev/kfd` seems to be once the main decision points to run the plugin and the only thing you do is a `stat()` I could imagine as test creating a file `/dev/kfd` (if it does not exist), write a something to it, verify in the test plugin that the normal file `/dev/kfd` has the expected value. In the late stage hook replace the content of `/dev/kfd` with something else and verify it in the test case, after the restore. That way we are running the `/dev/kfd` code paths already a bit and are doing a positive test to see if it the plugin actually does some work.

Also only install the plugin for this test case using the zdtm hooks and also remove it after running the test. Also make sure, if the test fails, that the `/dev/kfd` file is removed.

Would that make sense from your point of view?

adrianreber:
Looking at mount.c I see that plugin usage during checkpoint is tracked and only if a plugin was used during checkpointing it is required during restore. Maybe that would also make sense for your changes.

criu/proc_parse.c
```diff
@@ -173,8 +173,7 @@ static void parse_vma_vmflags(char *buf, struct vma_area *vma_area)
 	 * only exception is VVAR area that mapped by the kernel as
 	 * VM_IO | VM_PFNMAP | VM_DONTEXPAND | VM_DONTDUMP
 	 */
-	if (io_pf && !vma_area_is(vma_area, VMA_AREA_VVAR) &&
-	    !vma_entry_is(vma_area->e, VMA_FILE_SHARED))
+	if (io_pf && !vma_area_is(vma_area, VMA_AREA_VVAR))
```
avagin, line 189: > I am not sure about how to handle systemcfg file but the current solution looks like a workaround so revert it.

This doesn't work this way. Any new changes must not affect existing users. If you explain how it breaks your plugin, we will be able to think together about how to fix this issue.

rajbhar: I saw this issue a couple of months ago when this change was first introduced. Since then, our plugin and criu has advanced and on the latest code, I didn't see this issue again. I will drop this change for now but will keep an eye on it and will re-open it if I ever see this again.

rajbhar:
> 
> Thanks for pointing to the CI logs. This is not really a good solution. We cannot introduce an error message into each restore just because a plugin does not exist. If there is no plugin there should be no error message. Only if the plugin failed to function correctly we should see an error message. I also think for your case (`ENOTSUP`) we should not see an error. Maybe a debug message if the plugin does not support something.

Agree, I will remove the error message and change it to pr_debug.

> 
> Concerning a test case. There should be a positive test case. Not just some 'Error' messages. As `/dev/kfd` seems to be once the main decision points to run the plugin and the only thing you do is a `stat()` I could imagine as test creating a file `/dev/kfd` (if it does not exist), write a something to it, verify in the test plugin that the normal file `/dev/kfd` has the expected value. In the late stage hook replace the content of `/dev/kfd` with something else and verify it in the test case, after the restore. That way we are running the `/dev/kfd` code paths already a bit and are doing a positive test to see if it the plugin actually does some work.
> 
> Also only install the plugin for this test case using the zdtm hooks and also remove it after running the test. Also make sure, if the test fails, that the `/dev/kfd` file is removed.
> 
> Would that make sense from your point of view?

Its not entirely true that the plugin only does a stat on /dev/kfd. In order to verify the authenticity of the PID of the resuming task, we take help from the kfd kernel driver. Plz see this https://github.com/RadeonOpenCompute/ROCK-Kernel-Driver/blob/99d888b3b7f2592786dd2731451437a3d2fe8340/drivers/gpu/drm/amd/amdkfd/kfd_chardev.c#L2834

 Even in the dummy plugin, returning 0 for a pid that isn't really associated with kfd will not be correct. For UPDATE_VMA hook, unless we return 1, we don't really impact the VMA offsets. Besides, this plugin hook won't even be called if the target process doesn't a device VMA which I think most ZDTM test cases are. So, I implemented ENOTSUP for dummy plugin so it doesnt break anything else with criu and then actual amdgpu_plugin does all the positive testing anyway.

criu/file-ids.c
```diff
@@ -77,11 +78,18 @@ int fd_id_generate_special(struct fd_parms *p, u32 *id)
 {
 	if (p) {
 		struct fd_id *fi;
+		struct stat st_kfd;
 
 		fi = fd_id_cache_lookup(p);
 		if (fi) {
-			*id = fi->id;
-			return 0;
+			if (stat(AMDGPU_KFD_DEVICE, &st_kfd) == -1) {
```
avagin, line 85: This looks wrong. If /dev/kfd exists, criu doesn't cache file fd-s.... Do I miss something?

rajbhar: Yes, we don't want criu to cache the fds for kfd processes. I've put some more details in the commit message.

avagin: but you disable the cache for all files if /dev/kfd exits and I haven't found any explanation why we need to do this in the commit message.

Maybe you want to disable the cache only for /dev/kfd files?

rajbhar: I will update the PR with the suggested changes.

rajbhar: Hi @avagin - Yes, I wanted to just disable caching for /dev/kfd (amdgpu devices). I spent some more time debugging this but so far I haven't figured out a clean way to deal with this. I was thinking of using HANDLE_DEVICE_VMA plugin which I introduced in the first patch of the series to also detect whether we don't want the id caching during checkpoint and during the last stage of checkpoint i.e. cr_dump_finish enable it again but thats not a good solution either. Is there a cleaner way of disabling caching of ids on the fly and then enable again? Maybe a new command line param if thats acceptable? Also, I am trying to figure out if this is a generic issue with tasks that have device file fds containing shared memory / anonymous memory mappings.  Please advice.

avagin:
```
diff --git a/criu/file-ids.c b/criu/file-ids.c
index 1b9d68888..7a6e6b5b6 100644
--- a/criu/file-ids.c
+++ b/criu/file-ids.c
@@ -72,6 +72,10 @@ static struct fd_id *fd_id_cache_lookup(struct fd_parms *p)
 
 int fd_id_generate_special(struct fd_parms *p, u32 *id)
 {
+       // Don't cache mapped devices.
+       if (p->st_mode & (S_IFCHR|S_IFBLK))
+               p = NULL;
+
        if (p) {
                struct fd_id *fi;
```

I think we can skip all devices here. fd_id_cache_lookup is used for memory mappings and special files like cwd, exe, etc.

avagin: In an ideal case, we have to be able to get a file descriptor that is associated with a mapped struct file. In this case, we will be able to use kcmp and find out all mappings and file descriptors that are shared the same struct file. But this will require some kernel changes...

rajbhar: Thanks @avagin .. I have uploaded a slightly modified version otherwise I was getting a seg fault while testing.

rajbhar:
@avagin I have updated the PR with your suggested changes with slight modifications otherwise, we get seg fault.

avagin (approved):

adrianreber:
@rajbhar Please rebase and fix the errors from `make indent`.

About testing the real plugin. Any chance AMD can give access to a machine with the necessary hardware where someone could install the GitHub Actions self hosted runner. That way we could integrate AMD GPU based tests in our CI runs to make sure nothing breaks it.

rajbhar:
> @rajbhar Please rebase and fix the errors from `make indent`.
> 
> About testing the real plugin. Any chance AMD can give access to a machine with the necessary hardware where someone could install the GitHub Actions self hosted runner. That way we could integrate AMD GPU based tests in our CI runs to make sure nothing breaks it.

Hi @adrianreber - I have rebased to the criu-dev tip as of today and fixed the indent errors.   Local testing also looks good to me. I used the same fedora container as used by the CI system to verify.  If anything else is needed to get this PR merged, please let me know. Thanks you.

Just on a side note, If I use Ubuntu 20.04 and latest clang-format-12 with the style file from CRIU repo, the make indent behaviour is much different, basically I always got some or the other error on Ubuntu.

rajbhar:
> Looking at mount.c I see that plugin usage during checkpoint is tracked and only if a plugin was used during checkpointing it is required during restore. Maybe that would also make sense for your changes.

PR #1562, opened 2021-07-26
Cleanup posix notify thread id dumping

Snorch:
Handle non-pidns and non-nspid cases for compatibility. Need it to fix test errors in Centos7.


adrianreber:
The Fedora Rawhide failure is because glibc switched to `clone3()` and the seccomp docker profiles are blocking `clone3()` if I understood it correctly.

adrianreber (approved):
Can't really comment if it is correct, but the broken CentOS 7 test is gone and all other tests are green.

Fedora Rawhide is currently known to fail. So, LGTM!

Snorch:
@avagin Please also take a look!

avagin (approved):

PR #1563, opened 2021-07-26
ci: fix Fedora Rawhide

adrianreber:
Fedora Rawhide updated to a glibc using clone3(). clone3() is, however, not yet part of the seccomp filter. Unfortunately 'docker build' does not allow dropping seccomp but luckily 'podman build' does.

This switches the Fedora Rawhide test to use Podman. Podman is part of GitHub Actions and no additional packages need to be installed.

### DIFF
--- a/scripts/ci/Makefile
+++ b/scripts/ci/Makefile
@@ -14,8 +14,9 @@ endif
 TARGETS := alpine fedora-rawhide centos7 centos8 archlinux
 ZDTM_OPTIONS :=
 UNAME := $(shell uname -m)
-
 export UNAME
+CONTAINER_RUNTIME := docker
+export CONTAINER_RUNTIME
 
 alpine: ZDTM_OPTIONS=-x zdtm/static/binfmt_misc -x zdtm/static/netns-nf -x zdtm/static/sched_policy00 -x zdtm/static/seccomp_strict -x zdtm/static/sigaltstack -x zdtm/static/signalfd00 -x zdtm/static/config_inotify_irmap
 
@@ -45,8 +46,19 @@ else
 	CONTAINER_OPTS := --rm -v /lib/modules:/lib/modules --tmpfs /run
 endif
 
+ifeq ($(CONTAINER_RUNTIME),podman)
+	# Just as Docker needs to use devicemapper Podman needs vfs
+	# as graphdriver as overlayfs does not support all test cases
+	STORAGE_DRIVER := vfs
+	# Podman limits the number of processes in a container using cgroups.
+	# Disable it as it breaks the thread-bomb test
+	CONTAINER_OPTS += --pids-limit=0
+endif
+
+export STORAGE_DRIVER
+
 restart-docker:
-	if [ "$$UNAME" = "x86_64" ]; then \
+	if [ "$$UNAME" = "x86_64" ] && [ "$$CONTAINER_RUNTIME" = "docker" ]; then \
 		echo "$$DOCKER_JSON" > /etc/docker/daemon.json; \
 		cat /etc/docker/daemon.json; \
 		systemctl status docker; \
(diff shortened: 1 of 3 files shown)


adrianreber:
CentOS 7 failure is expected and fixed in #1562

scripts/ci/Makefile
```diff
@@ -45,8 +46,19 @@ else
...
+	# Podman limits the number of processes in a container using cgroups.
+	# Disable it as it breaks the thread-bomb test
+	CONTAINER_OPTS += --pids-limit=0
+endif
+
+export STORAGE_DRIVER
+
 restart-docker:
-	if [ "$$UNAME" = "x86_64" ]; then \
+	if [ "$$UNAME" = "x86_64" ] && [ -d /etc/docker ]; then \
```
Snorch, line 61: Not sure about this hunk... Doesn't it check that docker was not yet started? And we put it into "restart-docker:" target which is a prerequisite for "$(TARGETS):" target and "fedora-asan:" so it would probably be called _before_ we call "$(CONTAINER_RUNTIME)" recipe and would never restart docker and never add device-mapper graph driver...

Sorry I'm not too good at makefiles and probably missing something.

adrianreber: It is a bit ugly. You are right. But now I am also not sure about it :wink: 

The goal of restart docker is to reconfigure it to use devicemapper instead of overlayfs for our tests. If I am running on a system without docker installed (i.e. /etc/docker does not exist), we are skipping the reconfiguration of the the docker daemon.

In the case of fedora-rawhide which should now be using Podman, because docker fails with the latest glibc in Fedora rawhide because of the blocked clone3() by seccomp, we are no longer using docker, so that docker-restart should be a noop if /etc/docker does not exist. This assumes if /etc/docker does not exist, restarting of the docker daemon will fail.

So in this case I am running the test with CONTAINER_RUNTIME=podman and if /etc/docker does not exist it should not fail restarting docker and thus failing the complete test.

I think this does the right thing, but maybe it is a bit ugly. It would not be really necessary for the CI environment where we have docker and podman installed, so even if we run with CONTAINER_RUNTIME=podman we can still restart the docker daemon even if it is totally unnecessary. I added this change because on my system, without docker installed, it failed restarting the docker daemon and that's why I added the check for /etc/docker.

Maybe a cleaner solution would be to check if CONTAINER_RUNTIME is == docker and only then reconfigure the docker daemon.

adrianreber: I updated the PR to check if CONTAINER_RUNTIME is docker and only then reconfigure it.

Snorch (approved):
Now it looks good, Thanks!

avagin:
Why does the Docker Test fail?

adrianreber:
> Why does the Docker Test fail?

@rst0git and I are seeing docker test failures more and more recently. The CRIU dump is always successful, but something inside docker fails most of the time during the second or the third checkpoint. We assume it is probably some race condition or similar. Almost impossible to reproduce locally.

@rst0git do you have any further insides?

rst0git:
> Why does the Docker Test fail?

There is a race condition in containerd (or moby). It fails when restoring a container that has been started from a checkpoint, and immediately create another checkpoint.

Here is a minimal reproducer:
```console
docker run --name cr -d alpine /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'
docker checkpoint create cr checkpoint1
docker start --checkpoint checkpoint1 cr
docker checkpoint create cr checkpoint2
```
The output is
```
715ebdab600cd481d746e4acfd4b22618611802bfc8eb49c3a84151e9f673912
checkpoint1
Error response from daemon: Cannot checkpoint container cr: content sha256:12c69b7a9d25695dd5f9d37d4e858e2f7c3f9da738ccf86f8d3042f6973af1df: already exists
```

We currently use [`sleep 1`](https://github.com/checkpoint-restore/criu/blob/29d5865ec43fecffb6484fbcab3e5c9e584383d4/scripts/ci/docker-test.sh#L65) as a solution to this problem, but it doesn't work well.

rst0git (approved):

avagin:
> @rst0git and I are seeing docker test failures more and more recently. The CRIU dump is always successful, but something inside docker fails most of the time during the second or the third checkpoint. We assume it is probably some race condition or similar. Almost impossible to reproduce locally.

@kolyshkin do you know what can be a root cause of this? Could you help to investigate?

rst0git:
@avagin I opened https://github.com/checkpoint-restore/criu/pull/1567 to address the problem with the Docker test, but I think this is unrelated to the fix in this PR.

PR #1564, opened 2021-07-26
ci: use pre-installed Podman

adrianreber:
GitHub Actions already has Podman installed. No need to install it a second time.

### DIFF
--- a/scripts/ci/podman-test.sh
+++ b/scripts/ci/podman-test.sh
@@ -1,30 +1,6 @@
 #!/bin/bash
 set -x -e -o pipefail
 
-if [ ! -e /etc/lsb-release ]; then
-       # This expects to run on Ubuntu
-       exit 1
-fi
-
-#shellcheck disable=SC1091
-. /etc/lsb-release
-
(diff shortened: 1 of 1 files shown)


scripts/ci/podman-test.sh
```diff
@@ -1,30 +1,6 @@
...
-# this is a podman packaging bug (https://github.com/containers/libpod/issues/4747)
-apt-get -y purge docker-ce || :
-
-./apt-install \
-    apt-transport-https \
-    ca-certificates \
-    curl \
-    software-properties-common
-
-./apt-install podman containernetworking-plugins
```
rst0git, line 26: In the [installation script](https://github.com/actions/virtual-environments/blob/main/images/linux/scripts/installers/containers.sh) only `podman` is installed. Do we need `containernetworking-plugins` as well?

adrianreber: The test seems to run happily without it. So I would say we do not need it.

rst0git (approved):

PR #1565, opened 2021-07-27
criu-ns: add man page and install target

rst0git:
This pull request adds a man page for the [`criu-ns`](https://criu.org/CR_in_namespace) script and includes it in the install/uninstall criu makefile targets.


adrianreber (approved):
Thanks. Looks good.

adrianreber:
@avagin This would be the criu-ns documentation for the upcoming 3.16 release.

PR #1566, opened 2021-07-28
posix-timers: fix getoverrun error handling

Snorch:
If retcode of dump_posix_timers is not zero it is treated as an error in
compel_rpc_sync. And currently we can return positive overrun of last
timer (if we are lucky and it is not zero) as retcode of function
dump_posix_timers, let's fix it. Also I don't see any point in putting
negative value into .overrun on error path, fix it too.

### DIFF
--- a/criu/pie/parasite.c
+++ b/criu/pie/parasite.c
@@ -157,12 +157,13 @@ static int dump_posix_timers(struct parasite_dump_posix_timers_args *args)
 			pr_err("sys_timer_gettime failed (%d)\n", ret);
 			return ret;
 		}
-		args->timer[i].overrun = sys_timer_getoverrun(args->timer[i].it_id);
-		ret = args->timer[i].overrun;
+		ret = sys_timer_getoverrun(args->timer[i].it_id);
 		if (ret < 0) {
 			pr_err("sys_timer_getoverrun failed (%d)\n", ret);
 			return ret;
 		}
+		args->timer[i].overrun = ret;
+		ret = 0;
 	}
 
 	return ret;


adrianreber (approved):
Thanks for the quick fix. The description and the code make sense to me.

criu/pie/parasite.c
```diff
@@ -157,12 +157,13 @@ static int dump_posix_timers(struct parasite_dump_posix_timers_args *args)
 			pr_err("sys_timer_gettime failed (%d)\n", ret);
 			return ret;
 		}
-		args->timer[i].overrun = sys_timer_getoverrun(args->timer[i].it_id);
-		ret = args->timer[i].overrun;
+		ret = sys_timer_getoverrun(args->timer[i].it_id);
 		if (ret < 0) {
 			pr_err("sys_timer_getoverrun failed (%d)\n", ret);
 			return ret;
```
0x7f454c46, line 163: Hmm, that got me interested what happens when the overrun count is too big for (int):
```
BUGS
       POSIX.1  specifies  that  if  the timer overrun count is equal to or greater
       than an implementation-defined maximum, DELAYTIMER_MAX, then  timer_getover‐
       run()  should  return  DELAYTIMER_MAX.   However,  before Linux 4.19, if the
       timer overrun value exceeds the maximum representable integer,  the  counter
       cycles,   starting   once   more   from   low  values.   Since  Linux  4.19,
       timer_getoverrun() returns DELAYTIMER_MAX (defined as INT_MAX in <limits.h>)
       in this case (and the overrun value is reset to 0).
```
And indeed, here is the commit, seems to not have been backported to -stable trees:
https://github.com/torvalds/linux/commit/78c9c4dfbf8c
We still claim to support older than v4.19 kernels, but I'm not sure if we can do anything to make them work right for this rare corner-case.
Anyway, LGTM.

0x7f454c46 (approved):

PR #1567, opened 2021-07-30
docker-test: refactor test scenario

rst0git:
In the current test for docker, we create a single container that is checkpoint/restore it 50 times sequentially. It is fairly uncommon (and not very useful) to checkpoint a container immediately after restore. We have seen the following error from containerd that occurs when trying to checkpoint a container immediately after restore:

```
Error response from daemon: Cannot checkpoint container cr: content sha256:12c69b7a9d25695dd5f9d37d4e858e2f7c3f9da738ccf86f8d3042f6973af1df: already exists
```
This PR aims to address the error above by modifying the test to remove the container after restore. Thus, it would use a new container for each checkpoint/restore test iteration.

### DIFF
--- a/scripts/ci/docker-test.sh
+++ b/scripts/ci/docker-test.sh
@@ -51,21 +51,44 @@ docker info
 
 criu --version
 
-# shellcheck disable=SC2016
-docker run --tmpfs /tmp --tmpfs /run --read-only --security-opt seccomp=unconfined --name cr -d alpine /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'
+run_container () {
+	docker run \
+		--tmpfs /tmp \
+		--tmpfs /run \
+		--read-only \
+		--name cr \
+		--health-cmd='sleep 1' \
+		--health-interval=1s \
+		-d \
+		alpine \
+		/bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'
+}
 
-sleep 1
-for i in $(seq 50); do
-	# docker start returns 0 silently if a container is already started
-	# docker checkpoint doesn't wait when docker updates a container state
-	# Due to both these points, we need to sleep after docker checkpoint to
-	# avoid races with docker start.
-	docker exec cr ps axf &&
-	docker checkpoint create cr checkpoint"$i" &&
-	sleep 1 &&
-	docker ps &&
-	(docker exec cr true && exit 1 || exit 0) &&
-	docker start --checkpoint checkpoint"$i" cr 2>&1 | tee log || {
+wait_running () {
+	until [ "$(docker inspect -f '{{.State.Running}}' cr)" = "true" ]; do
+		sleep 1;
+	done;
+}
+
+wait_healthy () {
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks for the PR. I am not 100% convinced that this is the right thing to do. I think both approaches are a good test.

1. Create container and checkpoint and restore it multiple times
2. Create multiple containers and checkpoint and restore them once

You changed from the existing scenario 1 to 2. Which is not wrong. But could we do both?

Scenario 1 was the one which gave us trouble because docker is not working correctly, so switching to scenario 2 is a valid solution as we do not really want to test docker, but we want to test if docker still works with the latest version of CRIU.

Could we fix scenario 1 by doing a `docker checkpoint rm` after each run? Maybe also include a `docker checkpoint ls`?

Extend the test by looping over scenario 1 and 2 maybe only 10 times instead of 50?

Would that be possible? Is that even a useful suggestion?

rst0git:
> Could we fix scenario 1 by doing a docker checkpoint rm after each run?

It fails with the same error: https://github.com/checkpoint-restore/criu/runs/3199781735

adrianreber:
> > Could we fix scenario 1 by doing a docker checkpoint rm after each run?
> 
> It fails with the same error: https://github.com/checkpoint-restore/criu/runs/3199781735

Ah, well, no need for us to fix docker. If it does not work not much CRIU can do about it. Then let's go with your first try. That way we can ensure that checkpoint restore works with docker.

scripts/ci/docker-test.sh
```diff
@@ -51,20 +51,26 @@ docker info
...
-
-sleep 1
-for i in $(seq 50); do
-	# docker start returns 0 silently if a container is already started
-	# docker checkpoint doesn't wait when docker updates a container state
-	# Due to both these points, we need to sleep after docker checkpoint to
-	# avoid races with docker start.
-	docker exec cr ps axf &&
+run_container () {
+	docker run --tmpfs /tmp --tmpfs /run --read-only --security-opt seccomp=unconfined --name cr -d alpine /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'
```
adrianreber, line 55: The seccomp unconfined should not really be necessary anymore. I know it was needed for CentOS 7 because seccomp suspending was not backported, but besides that it should work

scripts/ci/docker-test.sh
```diff
@@ -74,6 +80,5 @@ for i in $(seq 50); do
 		exit 1
 	}
 	docker ps
-	sleep 1
-done
-
+	docker rm -f cr
+done
```
adrianreber, line 84: newline at end of file would be nice

adrianreber:
Rebasing should pick up the  fedora rawhide ci fix

avagin:
pls update the commit so that it describes the changes.

rst0git:
@adrianreber I was able to fix the test for scenario 1 by adding healthcheck and performing checkpoint only when the container is in a healthy state.

@avagin I've updated the commit message to describe the changes.

adrianreber (approved):
Very good solution!. Thanks. I restarted the failing compat test. Let's see if it finishes successfully.

PR #1568, opened 2021-07-30
Socket bufs lock support (tcp perfomance decrease after migration fix)

Snorch:
When one sets socket buffer sizes with setsockopt(SO_{SND,RCV}BUF*), kernel sets corresponding SOCK_SNDBUF_LOCK or SOCK_RCVBUF_LOCK flags on struct sock. It means that such a socket with explicitly changed buffer size can not be auto-adjusted by kernel (e.g. if there is free memory kernel can auto-increase default socket buffers to improve performance). (see tcp_fixup_rcvbuf() and tcp_sndbuf_expand())  
 
CRIU is always changing buf sizes on restore, that means that all sockets receive lock flags on struct sock and become non-auto-adjusted after migration. In some cases it can decrease performance of network connections quite a lot.  
 
So let's c/r socket buf locks (SO_BUF_LOCKS), so that sockets for which auto-adjustment is available does not lose it.  
 
Origin of the problem: 
 
We have a customer in Virtuozzo who mentioned that nginx server becomes slower after migration. Especially it is easy to mention when you wget some big file via localhost from the same container which was just migrated. 
 
By strace-ing all nginx processes I see that nginx worker process before c/r sends data to local wget with big chunks ~1.5Mb, but after c/r it only succeeds to send by small chunks ~64Kb (also to remote wget it is around ~128Kb). That can explain the decrease in download speed. 
 
Before: 
sendfile(12, 13, [7984974] => [9425600], 11479629) = 1440626 <0.000180> 
 
After: 
sendfile(8, 13, [1507275] => [1568768], 17957328) = 61493 <0.000675> 
 
As a POC (this way one can check it without a kernel patch) I just commented out all buffer setting manipulations. I managed to get big chunks after c/r and download speed became the same (fast) as before migration.

This is yet a draft as a kernel patch does not hit mainstream kernel. See https://lkml.org/lkml/2021/7/30/346

### DIFF
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -1372,6 +1372,14 @@ static int check_network_lock_nftables(void)
 	return 0;
 }
 
+static int check_sockopt_buf_lock(void)
+{
+	if (!kdat.has_sockopt_buf_lock)
+		return -1;
+
+	return 0;
+}
+
 static int (*chk_feature)(void);
 
 /*
@@ -1490,6 +1498,7 @@ int cr_check(void)
 		ret |= check_ns_pid();
 		ret |= check_apparmor_stacking();
 		ret |= check_network_lock_nftables();
+		ret |= check_sockopt_buf_lock();
 	}
 
 	/*
@@ -1602,6 +1611,7 @@ static struct feature_list feature_list[] = {
 	{ "ns_pid", check_ns_pid },
 	{ "apparmor_stacking", check_apparmor_stacking },
 	{ "network_lock_nftables", check_network_lock_nftables },
+	{ "sockopt_buf_lock", check_sockopt_buf_lock },
 	{ NULL, NULL },
 };
 
--- a/criu/include/kerndat.h
+++ b/criu/include/kerndat.h
@@ -74,6 +74,7 @@ struct kerndat_s {
 	bool has_pidfd_getfd;
 	bool has_nspid;
 	bool has_nftables_concat;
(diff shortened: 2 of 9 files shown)


Snorch:
Now when a kernel patch is in net-next https://lore.kernel.org/linux-mips/162807840601.323.12461942403559350811.git-patchwork-notify@kernel.org/ we can start reviewing it. Note that on older kernels this functionality/test would be automatically disabled by corresponding kerndat checks.

0x7f454c46:
LGTM, but `vz_so_buf_lock` is a bit questionable for mainstream criu.

Snorch:
> LGTM, but `vz_so_buf_lock` is a bit questionable for mainstream criu.

Sure, `vz_` prefix is a leftover, will fix after I'm back from vocation.

0x7f454c46:
> > LGTM, but `vz_so_buf_lock` is a bit questionable for mainstream criu.
> 
> Sure, `vz_` prefix is a leftover, will fix after I'm back from vocation.

Enjoy your quest!

Snorch:
Thanks for reminder! Rebased + removed "vz_" prefix.

Snorch:
+ fix indentation in test

Snorch:
@checkpoint-restore/maintainers Please give some review here, I believe it's important as it fixes network (tcp) perfomance decrease after migration.

Snorch:
Thanks!

PR #1571, opened 2021-08-03
[GSOC] udp: Add support for checkpoint/restore UDP socket's queues

minhbq-99:
In these patches, I implement

- A new protobuf image for storing dumped UDP packets. The image has the
following format: UdpQueueEntry at the beginning for storing queues'
information, each packet has UdpPacketEntry at the beginning for storing
address information and packet's length followed by packet's data.

- Checkpoint/restore support for UDP send queue via the newly proposed
UDP_REPAIR in the Linux kernel. Here is the UDP_REPAIR patch
https://gist.github.com/minhbq-99/256be103c96aa939c75e05a009e6fc1c

- Checkpoint/restore support for UDP receive queue. When checkpointing,
packets are peeked from the socket. When restoring, a temporary socket is
used to send these dumped packet. The source IP addresses of these are
modified to matched the IP addresses when checkpointing by using
IP_PKTINFO and IP_TRANSPARENT. However, the source port cannot be
restored in some cases and the packet will be dropped.


adrianreber:
Please have a look at the CentOS 7 test failures. Those failures look like real errors.

minhbq-99:
Hi, I am fixing the error in CentOS 7, my local version has fixed the error in CI but there are still crashes.

The crashes are when I use `udp_packet_entry__free_unpacked(upe, NULL)` and `upe == NULL`. It is weird as on Ubuntu that `*free_unpacked` functions have the NULL check but in CentOS 7, they don't. 

As I can see in other places like this in sk-tcp.c
https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/sk-tcp.c#L396
`tcp_stream_entry__free_unpacked(tse, NULL);`

`*free_unpacked` is used in case when `pb_read_one` failed. At this point, I think the `tse == NULL` (https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/protobuf.c#L63)

So do we need to add NULL check in these cases or we need to do something so that protobuf compiler always produces the NULL check in `*free_unpacked` functions?

adrianreber:
> Hi, I am fixing the error in CentOS 7, my local version has fixed the error in CI but there are still crashes.
> 
> The crashes are when I use `udp_packet_entry__free_unpacked(upe, NULL)` and `upe == NULL`. It is weird as on Ubuntu that `*free_unpacked` functions have the NULL check but in CentOS 7, they don't.
> 
> As I can see in other places like this in sk-tcp.c
> https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/sk-tcp.c#L396
> `tcp_stream_entry__free_unpacked(tse, NULL);`
> 
> `*free_unpacked` is used in case when `pb_read_one` failed. At this point, I think the `tse == NULL` (https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/protobuf.c#L63)
> 
> So do we need to add NULL check in these cases or we need to do something so that protobuf compiler always produces the NULL check in `*free_unpacked` functions?

This is strange that we never had this problem before. Anyway, we cannot change what protobuf-c creates on CentOS 7 as we have to use the version that comes with the OS, so we need to fix it on our side. Maybe with a comment that it is only needed for CentOS 7 which goes EOL in 2024.

adrianreber:
Please rebase

criu/include/magic.h
```diff
@@ -100,6 +100,7 @@
 #define BPFMAP_FILE_MAGIC    0x57506142 /* Alapayevsk */
 #define BPFMAP_DATA_MAGIC    0x64324033 /* Arkhangelsk */
 #define APPARMOR_MAGIC	     0x59423047 /* Nikolskoye */
+#define UDP_QUEUE_MAGIC	     0x55803795 /* Balasiha */
```
avagin, line 103: 0x55483756

criu/sk-udp.c
```diff
@@ -0,0 +1,655 @@
...
+	int ret;
+
+	sk->rfd = dup(fd);
+	if (sk->rfd < 0) {
+		pr_perror("Can't save socket fd for repair");
+		return -1;
+	}
+
+	if (!(root_ns_mask & CLONE_NEWNET)) {
+		ret = nf_lock_connection(sk);
```
avagin, line 267: pls add a comment that explains why we need to block a UDP connection.

avagin:
> - Checkpoint/restore support for UDP send queue via the newly proposed UDP_REPAIR in the Linux kernel.

Have you sent these patches to lkml? Could you update https://criu.org/Upstream_kernel_commits?

criu/sk-udp.c
```diff
@@ -0,0 +1,655 @@
...
+	list_for_each_entry_safe(up, n, &recv_queue_packets, packet_list)
+		free_udp_packet(up);
+	INIT_LIST_HEAD(&recv_queue_packets);
+	xfree(g_data);
+	g_data = NULL;
+restore_off:
+	setsockopt(sk->rfd, SOL_SOCKET, SO_PEEK_OFF, &peek_off, sizeof(peek_off));
+out:
+	close_image(img);
+	return ret;
```
avagin, line 393: pls use the exit_code pattern here.

criu/sk-udp.c
minhbq-99: Looking at this now I think the network lock here is redundant. As with TCP, the network lock prevents the changes in TCP state when a new packet arrives. But with UDP, the packets are quite independent so the lock is not needed. What is your opinion?

criu/sk-udp.c
minhbq-99: Can you give me an example of exit_code pattern?

criu/include/magic.h
minhbq-99: Thank you, just a bit curiousity how can you get that number? I search for Balasiha's coordinate and the result is 55.7982° N, 37.9680° E

minhbq-99:
> > * Checkpoint/restore support for UDP send queue via the newly proposed UDP_REPAIR in the Linux kernel.
> 
> Have you sent these patches to lkml? Could you update https://criu.org/Upstream_kernel_commits?

I have just sent it, I will update the wiki when I get the requested account

criu/include/magic.h
avagin: I get them from wikipedia:
https://en.wikipedia.org/wiki/Balashikha

55°49′N 37°58′E is is actually the same with 55.7982° N, 37.9680° E. But I think we use the first format for others.

criu/sk-udp.c
avagin: We may want to lock them only if we want to be sure that we will not see new packets in receive queues. Do we need this?

criu/sk-udp.c
avagin: git grep exit_code ;)

The idea is that exit_code is initialized to -1 and changes to 0 only at the end. We use it in many places.

It allows to avoid bugs like this:
```
ret = -1;
if (x)
  goto err;
ret = check_smth()
if (ret)
  goto err;
if (y)
  goto err; <--- and you are in trouble...
```

criu/sk-udp.c
```diff
@@ -0,0 +1,655 @@
...
+
+struct udp_packet {
+	UdpPacketEntry *upe;
+	char *data;
+	struct list_head packet_list;
+};
+
+static LIST_HEAD(recv_queue_packets);
+/* The send queue only contains 1 packet */
+static struct udp_packet *send_queue_packet = NULL;
```
mihalicyn, line 28: I've already wrote about that. Using global variables like that is not safe.

Let's take a look on `send_queue_packet` which is used in two functions
`dump_send_queue` and `dump_one_udp`.
`dump_one_udp` calls `dump_send_queue`.

Stack:
`dump_one_udp`
`do_dump_one_inet_fd`
`dump_one_inet_fd` (->dump callback on `struct fdtype_ops`)
`do_dump_gen_file`
`dump_socket`
`dump_one_file`
`dump_task_files_seized` / `dump_my_file` (used in unix sock scm's support when transferring fd through unixsk)
for_each_pstree_item `dump_one_task`
`cr_dump_tasks`

As for now we dump file descriptors sequentially, so, no bug here. But I think it's better to move `send_queue_packet` and `g_data` to local variables of `dump_one_udp` to be on the safe side. The same about list `recv_queue_packets`.

Now let's analyze `udp_repair_sockets`. Restore stage of files is done in parallel way, so we need to have locking here. If you decide that udp sk network locking is needed and keep `udp_repair_sockets` list, then looks like we need to protect this list by mutex.

criu/sk-udp.c
```diff
@@ -0,0 +1,655 @@
...
+
+	pr_info("Dumping a UDP socket\n");
+
+	if (dump_opt(fd, SOL_UDP, UDP_CORK, &aux))
+		return -1;
+
+	udp_queue_entry.cork = false;
+	if (aux) {
+		udp_queue_entry.cork = true;
+	}
```
mihalicyn, line 327: no `{` needed here

criu/sk-udp.c
```diff
@@ -0,0 +1,655 @@
...
+	if (!img)
+		return -1;
+
+	ret = -1;
+	len = sizeof(peek_off);
+	if (getsockopt(fd, SOL_SOCKET, SO_PEEK_OFF, &peek_off, &len)) {
+		/* CentOS 7 does not support SO_PEEK_OFF on AF_INET sockets, the
+		 * peek off is 0 every recvmsg
+		 */
+		if (errno != EOPNOTSUPP) {
```
mihalicyn, line 342: maybe emit `pr_warn` here?

criu/sk-udp.c
minhbq-99: I changed these variables into local variables of `dump_one_udp` in the latest update

criu/sk-udp.c
minhbq-99: I use normal recv to dump receive queues, if new packets arrive after that recv, they are not dumped. And these new packets does not affect any previous dumped information so I removed the network lock

PR #1572, opened 2021-08-06
clang-format: enable sort includes

rst0git:
Following on the discussion in https://github.com/checkpoint-restore/criu/pull/1546#discussion_r683217430, this pull request enables `SortIncludes` for clang-format.


adrianreber:
Looks like you uncovered some bugs of header files depending on other header files.

Besides that. Please split it in two commits. One for `flog/` and one for all the other files. The releases branch does not have `flog/` so splitting the commits would make that easier.

adrianreber:
Do not change `.clang-format` directly. There is a script which fetches `.clang-format` and modifies it. The `.clang-format` changes should also be a separate commit.

PR #1573, opened 2021-08-08
dump: suspend/resume lsm on pre-dump

avagin:
Otherwise, criu pre-dump can fail with errors like this:
lib/infect.c:650: Unable to connect a transport socket: Permission denied'

Reported-by: Mr Jenkins


avagin:
Cc: @brauner

brauner (approved):
Looks good! Thank you!

PR #1575, opened 2021-08-09
ci: fix 'crit.sh: 3: source: not found'

adrianreber:
Jenkins test runs are failing with:

 ./test/jenkins/run_ct ./test/jenkins/crit.sh
 ./test/jenkins/crit.sh: 3: source: not found

Switch to bash which has 'source'.

Fixes: https://ci.kernoops.org/job/CRIU/view/CRIU-devel/job/CRIU-crit/job/criu-dev/627/console

CC: @mihalicyn

### DIFF
--- a/test/jenkins/_run_ct
+++ b/test/jenkins/_run_ct
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 mount --make-rslave /
--- a/test/jenkins/run_ct
+++ b/test/jenkins/run_ct
@@ -1,3 +1,3 @@
-#!/bin/sh
+#!/bin/bash
 
 unshare --mount --pid --fork -- $(readlink -f `dirname $0`/_run_ct) "$@"


rst0git (approved):

mihalicyn:
```
root@criu-test:/var/lib/jenkins/workspace/CRIU_CRIU-crit_criu-dev-DBNKUFN2EIV77FCLF3JCP2VBDOBGR7GVBO3FUACODQLMNTD7X6AQ# ls -la /bin/sh
lrwxrwxrwx 1 root root 4 Jul 18  2019 /bin/sh -> dash
```

We have `/bin/dash` as a default shell for a long time. The test begins to fail between
Aug 8, 2021, 9:05 PM
and
Aug 8, 2021, 3:05 PM

VM was rebooted in between.

Looks like VM rootfs was restored from snapshot or something else.

cc: @avagin

avagin:
I think I bind-mounted /bin/bash to /bin/dash after each boot.

PR #1578, opened 2021-08-10
configs: fix used after free cases

avagin:
We have some code written with the assumption that argv is never
destroyed, but when we handle configs, we construct an argv-like array
for each config, parse it and release it.

I am still not sure that we need to release the memory of per-config argv
arrays... The current scheme is going to be a source of used-after-free
bugs. When we will add the non-privileged mode, all these bugs will be
serious security issues.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1994,10 +1994,20 @@ int new_cg_root_add(char *controller, char *newroot)
 	if (!o)
 		return -1;
 
-	o->controller = controller;
-	o->newroot = newroot;
+	o->controller = xstrdup(controller);
+	if (!o->controller)
+		goto err_ctrl;
+	o->newroot = xstrdup(newroot);
+	if (!o->newroot)
+		goto err_newroot;
 	list_add(&o->node, &opts.new_cgroup_roots);
+
 	return 0;
+err_newroot:
+	xfree(o->controller);
+err_ctrl:
+	xfree(o);
+	return -1;
 }
 
 struct ns_desc cgroup_ns_desc = NS_DESC_ENTRY(CLONE_NEWCGROUP, "cgroup");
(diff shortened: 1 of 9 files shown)


criu/cgroup.c
```diff
@@ -1994,10 +1994,20 @@ int new_cg_root_add(char *controller, char *newroot)
...
+	o->newroot = xstrdup(newroot);
+	if (!o->newroot)
+		goto err_newroot;
 	list_add(&o->node, &opts.new_cgroup_roots);
+
 	return 0;
+err_newroot:
+	xfree(o->controller);
+err_ctrl:
+	xfree(0);
```
rst0git, line 2009: What is the purpose of `xfree(0)`?

avagin: Nice catch :). Thanks.

rst0git:
@avagin perhaps we should update `parse_join_ns()` as well?

```diff
--- a/criu/config.c
+++ b/criu/config.c
@@ -549,7 +549,11 @@ static int parse_join_ns(const char *ptr)
 {
        char *aux, *ns_file, *extra_opts = NULL;
 
-       aux = strchr(ptr, ':');
+       char *ns = xstrdup(ptr);
+       if (ns == NULL)
+               return -1;
+
+       aux = strchr(ns, ':');
        if (aux == NULL)
                return -1;
        *aux = '\0';
@@ -562,7 +566,7 @@ static int parse_join_ns(const char *ptr)
        } else {
                extra_opts = NULL;
        }
-       if (join_ns_add(ptr, ns_file, extra_opts))
+       if (join_ns_add(ns, ns_file, extra_opts))
                return -1;
 
        return 0;
```

avagin:
@rst0git I have added this in a separate commit. Are you working on a test to catch this sort of issues?

rst0git:
> @rst0git I have added this in a separate commit. Are you working on a test to catch this sort of issues?

@avagin Thanks, I've opened https://github.com/checkpoint-restore/criu/pull/1581 for it.

rst0git (approved):

Issue #1523, opened 2021-06-21
Restoring TCP connection: No tcp-stream-XXXX.img image


yimin-zhao:
Hi there,

I met a restoring error when I try to restore a container pod.
```
(00.068100)      1: unix: Opening standalone (stage 0 id 0x46 ino 101122607 peer 0)
(00.068103)      1:             Create fd for 40
(00.068105)      1:             Create fd for 37
(00.068106)      1: inet:       Restore: family AF_INET    type SOCK_STREAM    proto IPPROTO_TCP      port 5701 state TCP_CLOSE        src_addr 10.85.0.97
(00.068109)      1: tcp: Restoring TCP connection
(00.068110)      1: tcp: Restoring TCP connection id 47 ino 608276c
(00.068116)      1: No tcp-stream-608276c.img image
(00.068117)      1: Error (criu/protobuf.c:74): Unexpected EOF on (empty-image)
(00.068123)      1: Error (criu/files.c:1216): Unable to open fd=38 id=0x47
(00.084407) mnt: Switching to new ns to clean ghosts
(00.094449) Error (criu/cr-restore.c:2483): Restoring FAILED.
```
I already use `tcp-close` option for both dump and restore, but I checked the this [doc](https://criu.org/CLI/opt/--tcp-close), it says  `--tcp-close option affects all the process tree TCP sockets that have any TCP state except CLOSED and LISTEN`. In this way, the `tcp-close` flag doesn't affect the closed inet socket, but I also see a comment in the [source code](https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/sk-inet.c#L205)  
```
case TCP_CLOSE:	
    /* Trivial case, we just need to create a socket on restore */		
    break;
```
Because the container is going to be migrated to a host with a different IP, so I have to use `tcp-close` to drop all TCP connections.
I am kind of confused, anyway, for this error how can I resolve it?

Besides, is it related to the `nftables-devel` package? I cannot find this package at CentOS 8, I built criu without nftables support. 

**CRIU logs and information:**

<details><summary>CRIU full dump/restore logs:</summary>
<p>

[restore.log](https://github.com/checkpoint-restore/criu/files/6685549/restore.log)

</p>
</details>

<details><summary>Output of `criu --version`:</summary>
<p>

```
Version: 3.15
GitID: v3.14-174-gf68da4a86
```

</p>
</details>

<details><summary>Output of `criu check --all`:</summary>
<p>

```
[...]

PR #1579, opened 2021-08-11
tcp: Skip restoring TCP state when dumping with --tcp-close

minhbq-99:
Since commit e42f5e0 ("tcp: allow to specify --tcp-close on dump"),
--tcp-close option can be used when checkpointing. This option skips
checkpointing TCP_CLOSE socket's state. However, when restoring, we
still try to restore TCP_CLOSE socket's state. As a result, a
non-existent protobuf image is opened.

This commit allows dumping TCP_CLOSE socket's state in --tcp-close case.
This commit also removes a redundant check for TCP_LISTEN state in
restore_one_tcp as TCP_LISTEN socket cannot reach this function

Fixes: #1523

### DIFF
--- a/criu/sk-tcp.c
+++ b/criu/sk-tcp.c
@@ -451,7 +451,7 @@ int restore_one_tcp(int fd, struct inet_sk_info *ii)
 
 	pr_info("Restoring TCP connection\n");
 
-	if (opts.tcp_close && ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {
+	if (opts.tcp_close) {
 		if (shutdown(fd, SHUT_RDWR) && errno != ENOTCONN) {
 			pr_perror("Unable to shutdown the socket id %x ino %x", ii->ie->id, ii->ie->ino);
 		}
--- a/test/zdtm/static/socket-tcp-close2.c
+++ b/test/zdtm/static/socket-tcp-close2.c
@@ -1,0 +1,67 @@
+#include <sys/types.h>
+#include <sys/socket.h>
+#include <netinet/tcp.h>
+#include <netinet/in.h>
+#include <linux/types.h>
+#include <signal.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check both dump and restore with tcp_close on TCP_CLOSE sockets";
+const char *test_author = "Bui Quang Minh <minhquangbui99@gmail.com>";
+
+static int port = 8880;
+
+int main(int argc, char **argv)
+{
+	int fd_s, fd, client;
+	char c;
+
+	test_init(argc, argv);
+	signal(SIGPIPE, SIG_IGN);
+
+	fd_s = tcp_init_server(AF_INET, &port);
+	if (fd_s < 0) {
+		pr_err("Server initializations failed\n");
+		return 1;
+	}
+
+	client = tcp_init_client(AF_INET, "localhost", port);
+	if (client < 0) {
+		pr_err("Client initializations failed\n");
+		return 1;
+	}
+
+	fd = tcp_accept_server(fd_s);
+	if (fd < 0) {
+		pr_err("Can't accept client\n");
+		return 1;
+	}
+	close(fd_s);
+
+	shutdown(client, SHUT_WR);
+	shutdown(fd, SHUT_WR);
+
+	test_daemon();
+	test_waitsig();
+
+	if (read(fd, &c, 1) != 0) {
+		fail("read server");
+		return 1;
+	}
+	if (read(client, &c, 1) != 0) {
+		fail("read client");
+		return 1;
+	}
+	if (write(client, &c, 1) != -1) {
+		fail("write client");
+		return 1;
+	}
+	if (write(fd, &c, 1) != -1) {
+		fail("write server");
+		return 1;
+	}
+
+	pass();
+	return 0;
+}
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -334,7 +334,8 @@ mount -t cgroup -o devices,freezer none devices,freezer
     Checkpoint established TCP connections.
 
 *--tcp-close*::
-    Don't dump the state of, or block, established tcp connections.
+    Don't dump the state of, or block, established tcp connections
+    (including the connection is once established but now closed).
     This is useful when tcp connections are not going to be restored.
 
 *--skip-in-flight*::
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -107,6 +107,7 @@ TST_NOFILE	:=				\
 		socket-tcp4v6-closed		\
 		socket-tcp-close0 		\
 		socket-tcp-close1 		\
+		socket-tcp-close2		\
 		socket-dump-tcp-close 		\
 		socket-tcp-unconn		\
 		socket-tcp6-unconn		\
--- a/test/zdtm/static/socket-tcp-close2.desc
+++ b/test/zdtm/static/socket-tcp-close2.desc
@@ -1,0 +1,1 @@
+{'opts': '--tcp-close', 'flags': 'reqrst '}


criu/sk-tcp.c
```diff
@@ -428,7 +430,7 @@ int restore_one_tcp(int fd, struct inet_sk_info *ii)
 
 	pr_info("Restoring TCP connection\n");
 
-	if (opts.tcp_close && ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {
+	if (dump_with_tcp_close | (opts.tcp_close && ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE)) {
 		if (shutdown(fd, SHUT_RDWR) && errno != ENOTCONN) {
```
rst0git, line 434: The problem occurs because when the `--tcp-close` option is used during checkpoint CRIU does not dump sockets in `TCP_LISTEN` and `TCP_CLOSE`.

Looking closer at the commit message in https://github.com/checkpoint-restore/criu/commit/2c37042821906f013634e305899cb25ff1a5a7b1:
```
New restore option 'tcp-close' was introduced. It restores all connected
TCP sockets in TCP_CLOSE state. Here we consider tcp sockets in
TCP_ESTABLISHED, TCP_FIN_WAIT2, TCP_FIN_WAIT1, TCP_CLOSE_WAIT,
TCP_LAST_ACK, TCP_CLOSING, TCP_SYN_SENT states as connected sockets.
This is consistent with current CRIU usage of these states. Thus this
option doesn't affect sockets with original states of TCP_LISTEN and
TCP_CLOSE.
```

and https://github.com/checkpoint-restore/criu/commit/e42f5e032e0a518374f8e38aae6af198d2b66da8:

```
In this case, states of established tcp connections will not be dumped
and they will not be blocked. This will be useful in case of snapshots,
when we don't need to restore tcp connections.
```

It looks like the intended behaviour is for sockets in `TCP_LISTEN` and `TCP_CLOSE` to be checkpointed and restored when `--tcp-closed` is used. The following change could fix this:

```diff
diff --git a/criu/sk-tcp.c b/criu/sk-tcp.c
index 8e78fca8c..2f64f0f40 100644
--- a/criu/sk-tcp.c
+++ b/criu/sk-tcp.c
@@ -240,7 +240,7 @@ int dump_one_tcp(int fd, struct inet_sk_desc *sk, SkOptsEntry *soe)
        if (sk->dst_port == 0)
                return 0;
 
-       if (opts.tcp_close) {
+       if (opts.tcp_close && sk->state != TCP_LISTEN && sk->state != TCP_CLOSE) {
                return 0;
        }
```

rst0git: @avagin what do you think?

avagin: I am not sure that you understand what you mean. We check that sk->dst_port isn't 0, it means that this is an established connection. Why do we need to check TCP_LISTEN and  TCP_CLOSE here?

criu/sk-tcp.c
```diff
@@ -428,7 +430,7 @@ int restore_one_tcp(int fd, struct inet_sk_info *ii)
 
 	pr_info("Restoring TCP connection\n");
 
-	if (opts.tcp_close && ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {
+	if (dump_with_tcp_close | (opts.tcp_close && ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE)) {
```
avagin, line 433: This looks wrong. It looks like it should be:

if ((dump_with_tcp_close || opts.tcp_close) && ii->ie->state != TCP_LISTEN && ii->ie->state != TCP_CLOSE) {

but restore_one_tcp is never called for listening sockets and I am not sure about TCP_CLOSE too...

test/zdtm/static/socket-tcp-close2.c
```diff
@@ -0,0 +1,47 @@
...
+		pr_err("Client initializations failed\n");
+		return 1;
+	}
+
+	fd = tcp_accept_server(fd_s);
+	if (fd < 0) {
+		pr_err("Can't accept client\n");
+		return 1;
+	}
+	close(fd_s);
```
avagin, line 37: why do we need to close a server socket here?  I think we need to check that it will be restored properly.

test/zdtm/static/socket-tcp-close2.c
```diff
@@ -0,0 +1,47 @@
...
+
+	shutdown(client, SHUT_WR);
+	shutdown(fd, SHUT_WR);
+
+	test_daemon();
+	test_waitsig();
+
```
avagin, line 44: Why do we check nothing here? I think we need to call send and recv for the socket.

test/zdtm/static/socket-tcp-close2.c
```diff
@@ -0,0 +1,47 @@
...
+	fd = tcp_accept_server(fd_s);
+	if (fd < 0) {
+		pr_err("Can't accept client\n");
+		return 1;
+	}
+	close(fd_s);
+
+	shutdown(client, SHUT_WR);
+	shutdown(fd, SHUT_WR);
```
avagin, line 40: why do we need to shutdown both sockets?

criu/sk-tcp.c
minhbq-99: Yes, I see that TCP_LISTEN socket cannot reach here too. However, TCP_CLOSE sockets can. Without applying the patch, my added zdtm test will fail with the same error as the referenced issue `No tcp-stream-XXXX.img image`. This is because when `--tcp-close` is used when checkpointing, we skip dumping TCP for all TCP states. However, when restoring we still try to restore TCP_CLOSE state. So in my patch, I will skip restore TCP_CLOSE if the checkpoint happens with `--tcp-close`, the idea from @rst0git is great too as his idea is keep dumping TCP state in TCP_CLOSE case.

test/zdtm/static/socket-tcp-close2.c
minhbq-99: This is for making sockets into TCP_CLOSE state but still have `dst_port` for reaching `restore_one_tcp`

criu/sk-tcp.c
rst0git: > Why do we need to check TCP_LISTEN and TCP_CLOSE here?

You're right, only check for `TCP_CLOSE` is necessary here. I've updated the patch in the comment above.

test/zdtm/static/socket-tcp-close2.c
avagin: Should we have another pair of sockets that are not closed?

minhbq-99: The other cases when TCP sockets are in connected or listening state are covered in your `socket-dump-tcp-close.c` test. Maybe we can merge these 2 into 1 test case.

criu/sk-tcp.c
Snorch: @avagin Looks like `(dump_with_tcp_close || opts.tcp_close)` should be equal to `opts.tcp_close` in restore_one_tcp due to check in check_img_inventory.

Snorch:
As opts.tcp_close is about closing "established" sockets and it should not affect closed ones, I like the solution from @rst0git.

> We check that sk->dst_port isn't 0, it means that this is an established connection.

By debug printing in dump_one_tcp I see that at least on zdtm test socket-tcp-close2 that can be situations where state=TCP_CLOSE and port is not zero.

Moreover, probably we still want to restore ports of established sockets which we close with --tcp-close?

rst0git:
@minhbq-99 would you be able to update this PR?

minhbq-99:
Look like the kernel behavior differs in CentOS 7, shutting down 2 ends of TCP connection still cannot make the client socket into `TCP_CLOSE`, it is in `TCP_FIN_WAIT1` (4)

criu/sk-tcp.c
```diff
@@ -263,7 +263,7 @@ int dump_one_tcp(int fd, struct inet_sk_desc *sk, SkOptsEntry *soe)
 	if (sk->dst_port == 0)
 		return 0;
 
-	if (opts.tcp_close) {
+	if (opts.tcp_close && sk->state != TCP_CLOSE) {
```
avagin, line 266: It is unclear without a comment what is going on here.

And this doesn't look right. We don't need to dump closed sockets if tcp_close is set. If you want to separated them from just created sockets, I think it would be better to introduce a new state and use it for all undumped tcp sockets.

minhbq-99: I don't think it's necessary to introduce a new state, the problem here is the inconsistency between dumping and restoring

- When dumping with tcp_close, we don't dump established states and TCP_CLOSE
- In restoring with tcp_close, we skip restoring established states but still try to restore TCP_CLOSE

To solve this problem, I think we need to determine whether we want to checkpoint/restore TCP_CLOSE socket when tcp_close is used. We need to dump TCP_CLOSE sockets like the current solution or we must skip restoring TCP_CLOSE sockets in restoring path.

rst0git: @minhbq-99 Thank you for working on this.

> As opts.tcp_close is about closing "established" sockets and it should not affect closed ones

I agree with @Snorch's comment above ^

> It is unclear without a comment what is going on here.

@avagin Maybe we could add a comment above the if statement?

criu/sk-tcp.c
avagin: The commit messages that you mentioned do explain why we may want to dump closed sockets. Do you remember why we decided to add --tcp-close? If we want to make a snapshot, we don't care about tcp connections, because when we will restore processes other sides of tcp connections will not be there. The second reason is that we don't want to leave any iptables or nftables rules that we use to lock connections.

criu/sk-tcp.c
avagin: > I agree with @Snorch's comment above ^

Do you mean this one:

> As opts.tcp_close is about closing "established" sockets and it should not affect closed ones, I like the solution from @rst0git.

The problem with this comment is that it doesn't explain why it should work this way. We are talking about sockets for which dst_port isn't zero. It means they are all "established" but some of them have been closed...

rst0git: One reason we need to checkpoint and restore sockets that are in TCP_CLOSE state is because there might be data in the buffered socket that we need to restore.

Here is an example of a test https://github.com/rst0git/criu/commit/e45dd76a67c0f229e1d9c457a54f211be3a0ea77 that would: 

- Fail without a patch:
```
==================== Run zdtm/static/socket-tcp-close3 in h ====================
 DEP       socket-tcp-close3.d
 CC        socket-tcp-close3.o
 LINK      socket-tcp-close3
Start test
./socket-tcp-close3 --pidfile=socket-tcp-close3.pid --outfile=socket-tcp-close3.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/socket-tcp-close3/74/1/restore.log
b'(00.000696)     74: inet: \tRestore: family AF_INET    type SOCK_STREAM    proto IPPROTO_TCP      port 44244 state TCP_CLOSE        src_addr 127.0.0.1'
b'(00.000703)     74: tcp: Restoring TCP connection'
b'(00.000706)     74: tcp: Restoring TCP connection id 8 ino 137f68'
b'(00.000711)     74: No tcp-stream-137f68.img image'
b'(00.000713)     74: Error (criu/protobuf.c:72): Unexpected EOF on (empty-image)'
b'(00.000720)     74: Error (criu/files.c:1200): Unable to open fd=5 id=0x8'
b'(00.000735) Error (criu/cr-restore.c:2470): Restoring FAILED.'
########### Test zdtm/static/socket-tcp-close3 FAIL at CRIU restore ############
```
- Fail if we don't restore the `TCP_CLOSE` sockets (i.e., using the solution proposed in https://github.com/checkpoint-restore/criu/commit/4720a914cd92b7dc121e270cde1b0459f9256a37)
```
==================== Run zdtm/static/socket-tcp-close3 in h ====================
 DEP       socket-tcp-close3.d
 CC        socket-tcp-close3.o
 LINK      socket-tcp-close3
Start test
./socket-tcp-close3 --pidfile=socket-tcp-close3.pid --outfile=socket-tcp-close3.out
Run criu dump
Run criu restore
Send the 15 signal to  74
Wait for zdtm/static/socket-tcp-close3(74) to die for 0.100000
########### Test zdtm/static/socket-tcp-close3 FAIL at result check ############
02:26:19.205:    74: FAIL: socket-tcp-close3.c:58: Did't receieve all buffered data (errno = 11 (Resource temporarily unavailable)
```
- And pass with the current patch:
```
=================== Run zdtm/static/socket-tcp-close3 in ns ====================
Construct root for zdtm/static/socket-tcp-close3
Start test
./socket-tcp-close3 --pidfile=socket-tcp-close3.pid --outfile=socket-tcp-close3.out
Run criu dump
Run criu restore
Send the 15 signal to  88
Wait for zdtm/static/socket-tcp-close3(88) to die for 0.100000
Removing dump/zdtm/static/socket-tcp-close3/62
=================== Test zdtm/static/socket-tcp-close3 PASS ====================
==================== Run zdtm/static/socket-tcp-close3 in h ====================
Start test
./socket-tcp-close3 --pidfile=socket-tcp-close3.pid --outfile=socket-tcp-close3.out
Run criu dump
Run criu restore
Send the 15 signal to  160
Wait for zdtm/static/socket-tcp-close3(160) to die for 0.100000
Removing dump/zdtm/static/socket-tcp-close3/160
=================== Test zdtm/static/socket-tcp-close3 PASS ====================
=================== Run zdtm/static/socket-tcp-close3 in uns ===================
Start test
./socket-tcp-close3 --pidfile=socket-tcp-close3.pid --outfile=socket-tcp-close3.out
Run criu dump
Run criu restore
Send the 15 signal to  257
Wait for zdtm/static/socket-tcp-close3(257) to die for 0.100000
Removing dump/zdtm/static/socket-tcp-close3/229
=================== Test zdtm/static/socket-tcp-close3 PASS ====================
```

avagin: I think you don't understand basic things about --tcp-close. If an user sets this options, they agree to drop all established sockets including closed ones.

avagin: what do you think about https://github.com/avagin/criu/commit/4181b237a1b98feddaaed441f58a836289951214?

rst0git: > If an user sets this option, they agree to drop all established sockets including closed ones.

I don't agree.

The commit message introducing this option (2c37042) explicitly states:
```
this option doesn't affect sockets with original states of TCP_LISTEN and TCP_CLOSE.
```

and the man page for `--tcp-close` doesn't say anything about closed sockets:
```
--tcp-close
 Don't dump the state of, or block, established tcp connections.
 This is useful when tcp connections are not going to be restored.
```

Why do you think closed sockets should not be restored?

rst0git: > what do you think about avagin@4181b23?

The test in https://github.com/rst0git/criu/commit/e60ee8152a8e018db763a70348a8385049942858 fails with
> 54: FAIL: socket-tcp-close3.c:58: Did't receieve all buffered data (errno = 11 (Resource temporarily unavailable))

If we want this to be the expected behaviour we should update the man page and https://criu.org/CLI/opt/--tcp-close

avagin: @rst0git Do you remember why we decided to add --tcp-close? We have users that want to dump processes and then restore them later on another host with another IP address. And they don't want to see iptables rules that we use to block TCP connections. If we apply the fix that you suggested here, we will not be able to restore processes on a host with another IP address and we will have iptables rules after criu dumped.

```
# python test/zdtm.py run -t zdtm/static/socket-tcp-close3 -f h --sbs
userns is supported
=== Run 1/1 ================ zdtm/static/socket-tcp-close3
timens isn't supported on 5.10.28-00069-g96f0f4a40fdd-dirty
==================== Run zdtm/static/socket-tcp-close3 in h ====================
Start test
./socket-tcp-close3 --pidfile=socket-tcp-close3.pid --outfile=socket-tcp-close3.out
Pause before dump. Press Enter to continue.
Run criu dump
Pause before restore. Press Enter to continue.^Z
[1]+  Stopped                 python test/zdtm.py run -t zdtm/static/socket-tcp-close3 -f h --sbs
# ps axf | grep zdtm.py
   2315 pts/0    T      0:00              \_ python test/zdtm.py run -t zdtm/static/socket-tcp-close3 -f h --sbs
   2333 pts/0    T      0:00              |   \_ ./zdtm_ct zdtm.py
   2334 pts/0    S      0:00              |       \_ python zdtm.py
   2336 pts/0    T      0:00              |           \_ python zdtm.py
   2396 pts/0    S+     0:00              \_ grep --color=auto zdtm.py     
# nsenter -n -t 2336 iptables -L
Chain INPUT (policy ACCEPT)
target     prot opt source               destination         
DROP       tcp  --  localhost            localhost            mark match ! 0xc114 tcp spt:60646 dpt:cddbp-alt
DROP       tcp  --  localhost            localhost            mark match ! 0xc114 tcp spt:cddbp-alt dpt:60646

Chain FORWARD (policy ACCEPT)
target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
target     prot opt source               destination         
DROP       tcp  --  localhost            localhost            mark match ! 0xc114 tcp spt:cddbp-alt dpt:60646
DROP       tcp  --  localhost            localhost            mark match ! 0xc114 tcp spt:60646 dpt:cddbp-alt
```

> If we want this to be the expected behaviour we should update the man page and https://criu.org/CLI/opt/--tcp-close

Yes, this is expected behavior and we need to update our documentation. We need to tell that --tcp-close affects all established TCP sockets.

rst0git: I see, thank you for the explanation!
@minhbq-99 would you be able to apply Andrei's fix (https://github.com/avagin/criu/commit/4181b237a1b98feddaaed441f58a836289951214) and update [Documentation/criu.txt](https://github.com/checkpoint-restore/criu/blob/criu-dev/Documentation/criu.txt)?

minhbq-99: Hi, I updated the pull request.

```c
if (opts.tcp_close && (ii->ie->state != TCP_CLOSE || ii->ie->dst_port != 0))
```

As I can see, `ii->ie->dst_port != 0` is guaranteed to be true when reaching `restore_one_tcp` so I simplified it to

```c
if (opts.tcp_close)
```

rst0git (approved):
LGTM

PR #1581, opened 2021-08-15
zdtm: Add option for testing config file options

rst0git:
This pull request adds the `--criu-config` option to ZDTM to allow running tests with CRIU options provided in a configuration file instead of command-line options.


test/zdtm.py
```diff
@@ -2519,179 +2514,205 @@ def clean_stuff(opts):
...
-    tst.available()
-
-opts['action'](opts)
-
-for tst in test_classes.values():
-    tst.cleanup()
+if __name__ == '__main__':
+    os.chdir(os.path.dirname(os.path.abspath(__file__)))
+    signal.signal(signal.SIGALRM, alarm)
+    opts = get_cli_args()
```
avagin, line 2729: opts isn't used, is it?

avagin:
What tests will fail if we run zdtm.py --criu-config without the last config fixes?

test/zdtm.py
rst0git: Actually `opts` has a global scope. It is used in the `main()` function, as well as `finish()`, `wait()`, `wait_all()` methods of `Launcher`, and `check(feature)`, `available()` static methods of the `criu` class, etc.

adrianreber:
Have you seen the CentOS 7 errors? Everything seems broken.

avagin:
@rst0git why do you close this pr?

rst0git:
>@rst0git why do you close this pr?

@avagin I was considering to open another PR since this one is tagged as `stale-pr`

> What tests will fail if we run zdtm.py --criu-config without the last config fixes?

No tests failed without the last config fixes

> Have you seen the CentOS 7 errors? Everything seems broken.

@adrianreber Thanks, it should be fixed now

test/zdtm.py
```diff
@@ -2519,179 +2514,205 @@ def clean_stuff(opts):
...
+    """
+    Collect this namespace's zombies
+    """
+    while True:
+        wpid, status = os.wait()
+        if wpid == pid:
+            if os.WIFEXITED(status):
+                status = os.WEXITSTATUS(status)
+            else:
+                status = 1
```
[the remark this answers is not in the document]
Snorch: Still looks unresolved.

Snorch:
I also believe that some explanation needed about new patches in PR comments, why do we need them for config file options testing? On previous review step I saw 6 patches now we have 10 here.

```
[snorch@fedora criu]$ git check-rebase 5a0136988..fefe21e82 04f8368ea..79a87ef7c
a64eaf8fd  c7f6f06dd  zdtm: refactor main
464a96581  8b554e1b6  zdtm: import builtins module
           b0a39ac02  zdtm: sort import lines
ecb65fa54  6c7f0bbd8  zdtm: use long form cli options
e94a01b07  e95c915f5  zdtm: add --criu-config option
8e45a101e  a4409d2d6  zdtm: drop redundant config_inotify_irmap test
fefe21e82  6b919cb86  ci: run criu-config tests
           669f0bcac  test/Makefile: drop unused other target
           d5f349e5f  Modularize zdtm.py
           79a87ef7c  test/autofs: fix use-after-free
```

rst0git:
@Snorch Thank you for asking.

I've dropped commit `test/Makefile: drop unused other target`.

Commit `Modularize zdtm.py` doesn't introduce functional changes. It moves parts of `zdtm.py` into separate Python modules to reduce its size and make it easier to maintain.
```
 test/zdtm.py                 | 932 ++++++-------------------------------------------------------------------------------------------------------
 test/zdtm/__init__.py        |  25 +++
 test/zdtm/cg_freezer.py      |  97 ++++++++++++
 test/zdtm/exceptions.py      |  19 +++
 test/zdtm/groups_test.py     |  58 +++++++
 test/zdtm/host_flavor.py     |  20 +++
 test/zdtm/inherit_fd_test.py | 179 +++++++++++++++++++++
 test/zdtm/ns_flavor.py       | 161 +++++++++++++++++++
 test/zdtm/userns_flavor.py   |  25 +++
 test/zdtm/utils.py           | 107 +++++++++++++
 test/zdtm/zdtm.py            | 250 +++++++++++++++++++++++++++++
```

Commit `test/autofs: fix use-after-free` is fixing GCC build errors.

> why do we need them for config file options testing?

The patches introducing the `--criu-config` option to zdtm contain changes in `zdtm.py` that make it difficult to move `Modularize zdtm.py` in a separate PR without merge conflict.

Snorch:
>>  why do we need them for config file options testing?
>The patches introducing the --criu-config option to zdtm contain changes in zdtm.py that make it difficult to move Modularize zdtm.py in a separate PR without merge conflict.

You actually could've just created another pr having needed commits from this pr in it as a "base" and after this one merged - rebase it.

test/zdtm/static/autofs.c
```diff
@@ -63,7 +63,6 @@ static char *xvstrcat(char *str, const char *fmt, va_list args)
 	if (ret == -ENOMEM) {
 		/* realloc failed. We must release former string */
 		pr_err("Failed to allocate string\n");
-		free(str);
```
Snorch, line 66: I believe that with this you break the underline logic of xvstrcat that it should free str on error, imagine that on first iteration of do-while loop realloc fails, str is still valid and we need to free it, because the caller will not know if it was fail on first iteration or on the second where str is already freed implicitly by realloc.

I believe we should instead do:
```
diff --git a/test/zdtm/static/autofs.c b/test/zdtm/static/autofs.c
index 53ece4458..ad1795842 100644
--- a/test/zdtm/static/autofs.c
+++ b/test/zdtm/static/autofs.c
@@ -47,6 +47,7 @@ static char *xvstrcat(char *str, const char *fmt, va_list args)
                ret = -ENOMEM;
                new = realloc(str, offset + delta);
                if (new) {
+                       str = new;
                        va_copy(tmp, args);
                        ret = vsnprintf(new + offset, delta, fmt, tmp);
                        va_end(tmp);
@@ -54,7 +55,6 @@ static char *xvstrcat(char *str, const char *fmt, va_list args)
                                /* NOTE: vsnprintf returns the amount of bytes
                                 *                                  * to allocate. */
                                delta = ret + 1;
-                               str = new;
                                ret = 0;
                        }
                }
```

Now str is always updated after successful realoc, not conditionally as it was before, so it always represent the currently allocated string.

rst0git: Thanks @Snorch!
I moved this patch in https://github.com/checkpoint-restore/criu/pull/1732/ and applied this change there.

test/zdtm.py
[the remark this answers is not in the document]
Snorch: >> I think we break status reporting to the caller by this patch (I don't see that status is global).
>
> It should be fixed now. Thanks.

Yes, now fixed.

> collect_zombies

Ok, let's leave it collect_zombies(), looks reasonable.

test/zdtm.py
```diff
@@ -2566,179 +2561,196 @@ def clean_stuff(opts):
...
+def collect_zombies(pid):
+    """
+    Collect this namespace's zombies
+    """
+    while True:
+        wpid, status = os.wait()
+        if wpid == pid:
+            if os.WIFEXITED(status):
+                return os.WEXITSTATUS(status)
+            return 1
```
avagin, line 2730: should it return 0 in other cases?

rst0git: We use this function to wait for the `zdtm.py` process. I believe `return 1` is used when `zdtm.py` did not terminate normally, i.e., by calling `exit()` or by returning from `main()`.

btw, thank you for point this out. I noticed the here we have same issue as in https://github.com/checkpoint-restore/criu/pull/1740#discussion_r799743246.

avagin: My comment was wrong, because here is no other cases.

>  I noticed the here we had same issue as in #1740 (comment) and updated the patch to fix it.
We don't have here an issue like #1740. Because it return 1 if a process has been killed (WIFSIGNALED).

rst0git: Thank you Andrei!

> We don't have here an issue like https://github.com/checkpoint-restore/criu/pull/1740. Because it return 1 if a process has been killed (WIFSIGNALED).

You are right!

rst0git:
I have rebased this pull request on the criu-dev branch, and as a follow-up on the discussion in https://github.com/checkpoint-restore/criu/pull/1733#discussion_r798172241, I removed the commit `zdtm: import builtins module`.

rst0git:
I have updated commit `zdtm: sort import lines` after running [autoflake](https://pypi.org/project/autoflake/) and [isort](https://pypi.org/project/isort/) and rebased on the criu-dev branch

PR #1582, opened 2021-08-16
selinux: fix no security label condition bug

time-river:
SELinux has three status: disabled, permissive, and enforcing. If the
status of the SELinux wasn't disabled, it would configure the rules
using `/etc/selinux/targeted`. However, because of the non-existing
rules in `/etc/selinux/targeted`, the security label of processes is
`kernel` instead of
`unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023` readed from
`/proc/<pid>/attr/current`. It will result the failure of the criu
dumping.

This patch fixed the above problem.


Snorch:
Should not it already be fixed by kdat check here https://github.com/checkpoint-restore/criu/pull/1457/files ?

Or possibly there can be situations when we save LSMTYPE__SELINUX to kdat.lsm (meaning criu process had something similar to "unconfined_u:unconfined_r:unconfined_t:s0-s0:c0.c1023")  and still have "kernel" for dumped process in /proc/<pid>/attr/current ?

adrianreber:
@time-river Same question as  @Snorch. I think we already fixed this. Can you give us additional details when you see the described error?

time-river:
@Snorch @adrianreber 
No. SELinux has a special status: `SELINUX=permissive` in */etc/selinux/semanage.conf*, but there is no rules (means no files in `/etc/selinux/targeted` directory, any processes' *attr/current* shows `kernel` string. criu can't checkpoint / restore the task in that SELinux status.

You can test the above by the following steps:

1. change the SELinux config to `permissive` in `/etc/selinux/semanage.conf`
2. delete `/etc/selinux/targeted` directory
3. reboot
4. `# cat /proc/self/attr/current` , then you can see `kernel` string
5. use criu checkpoint or restore any task, criu will report failure

adrianreber:
> @Snorch @adrianreber
> No. SELinux has a special status: `SELINUX=permissive` in _/etc/selinux/semanage.conf_, but there is no rules (means no files in `/etc/selinux/targeted` directory, any processes' _attr/current_ shows `kernel` string. criu can't checkpoint / restore the task in that SELinux status.
> 
> You can test the above by the following steps:
> 
>     1. change the SELinux config to `permissive` in `/etc/selinux/semanage.conf`
> 
>     2. delete `/etc/selinux/targeted` directory
> 
>     3. reboot
> 
>     4. `# cat /proc/self/attr/current` , then you can see `kernel` string
> 
>     5. use criu checkpoint or restore any task, criu will report failure

But isn't this error scenario already fixed with: https://github.com/checkpoint-restore/criu/pull/1457/files ?

During `criu dump` CRIU should already detect that `/proc/self/attr/current` returns `kernel` and switch to the NO_LSM mode.

adrianreber:
@time-river Is this PR still necessary? Do you want to rebase? Or is this solved?

PR #1583, opened 2021-08-17
zdtm: fix network lock tests when run with --norst

ZeyadYasser:
In test/jenkins/{crit.sh,criu-dump}, ZDTM is run with --norst,
Causing tests to only go through dump wihtout restoring.

The network locking tests are highly dependant on dump/restore hooks
causing them to hang when run with --norst.

We just add a reqrst flag to all network lock tests.


rst0git (approved):
LGTM

PR #1587, opened 2021-08-20
restore cwd when restore mnt namespace

liusdu:
When CONFIG_BINFMT_MISC_VIRTUALIZED is enabled, I found stats-dump is write to /, but not  work directory in dump  process. Finaly I found criu do not restore cwd when restoring mnt ns. The patch series do the following:

-  add helpers to switch/restore mnt ns
-  restore current work directory after restoring mnt ns in cr_dump_task
- use swich_mnt_ns/restore_mnt_ns helpers to simplify code in open_mountpoint

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -3749,12 +3749,14 @@ int dump_mnt_namespaces(void)
 void clean_cr_time_mounts(void)
 {
 	struct mount_info *mi;
-	int mnt_fd, ret;
+	int ns_old, ret;
 
 	for (mi = mntinfo; mi; mi = mi->next) {
+		int cwd_fd;
+
 		if (mi->mnt_id != CRTIME_MNT_ID)
 			continue;
-		ret = switch_ns(mi->nsid->ns_pid, &mnt_ns_desc, &mnt_fd);
+		ret = switch_mnt_ns(mi->nsid->ns_pid, &ns_old, &cwd_fd);
 		if (ret) {
 			pr_err("Can't switch to pid's %u mnt_ns\n", mi->nsid->ns_pid);
 			continue;
@@ -3763,7 +3765,7 @@ void clean_cr_time_mounts(void)
 		if (umount(mi->mountpoint) < 0)
 			pr_perror("Can't umount forced mount %s", mi->mountpoint);
 
-		if (restore_ns(mnt_fd, &mnt_ns_desc)) {
+		if (restore_mnt_ns(ns_old, &cwd_fd)) {
 			pr_err("cleanup_forced_mounts exiting with wrong mnt_ns\n");
 			return;
 		}
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -1358,13 +1358,8 @@ int open_mountpoint(struct mount_info *pm)
 	 * In both cases we can't do the thing from criu's mount namespace, so
 	 * we need to switch to mount's mount namespace, and later switch back.
 	 */
-	cwd_fd = open(".", O_DIRECTORY);
-	if (cwd_fd < 0) {
-		pr_perror("Unable to open cwd");
-		return -1;
-	}
 
-	if (switch_ns(pm->nsid->ns_pid, &mnt_ns_desc, &ns_old) < 0)
+	if (switch_mnt_ns(pm->nsid->ns_pid, &ns_old, &cwd_fd) < 0)
 		goto err;
 
 	if (!mnt_is_overmounted(pm)) {
@@ -1406,28 +1401,17 @@ int open_mountpoint(struct mount_info *pm)
 		}
 	}
 
-	if (restore_ns(ns_old, &mnt_ns_desc)) {
+	if (restore_mnt_ns(ns_old, &cwd_fd)) {
 		ns_old = -1;
 		goto err;
 	}
-
-	if (fchdir(cwd_fd)) {
-		pr_perror("Unable to restore cwd");
-		close(cwd_fd);
-		close(fd);
-		return -1;
-	}
-	close(cwd_fd);
 
 	return __open_mountpoint(pm, fd);
 err:
 	if (ns_old >= 0)
 		/* coverity[check_return] */
-		restore_ns(ns_old, &mnt_ns_desc);
+		restore_mnt_ns(ns_old, &cwd_fd);
 	close_safe(&fd);
-	if (fchdir(cwd_fd))
-		pr_perror("Unable to restore cwd");
-	close(cwd_fd);
 	return -1;
 }
 
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -1511,10 +1511,10 @@ err:
 static __maybe_unused int mount_cr_time_mount(struct ns_id *ns, unsigned int *s_dev, const char *source,
 					      const char *target, const char *type)
 {
-	int mnt_fd, ret, exit_code = 0;
+	int mnt_fd, cwd_fd, ret, exit_code = 0;
 	struct stat st;
 
-	ret = switch_ns(ns->ns_pid, &mnt_ns_desc, &mnt_fd);
+	ret = switch_mnt_ns(ns->ns_pid, &mnt_fd, &cwd_fd);
 	if (ret < 0) {
 		pr_err("Can't switch mnt_ns\n");
 		goto out;
@@ -1536,7 +1536,7 @@ static __maybe_unused int mount_cr_time_mount(struct ns_id *ns, unsigned int *s_
 	}
 
 restore_ns:
-	ret = restore_ns(mnt_fd, &mnt_ns_desc);
+	ret = restore_mnt_ns(mnt_fd, &cwd_fd);
 out:
 	return ret < 0 ? 0 : exit_code;
 }
--- a/criu/namespaces.c
+++ b/criu/namespaces.c
@@ -282,6 +282,51 @@ int restore_ns(int rst, struct ns_desc *nd)
 	return ret;
 }
 
+int switch_mnt_ns(int pid, int *rst, int *cwd_fd)
+{
+	int ret;
+	int fd;
+
+	if (!cwd_fd)
+		return switch_ns(pid, &mnt_ns_desc, rst);
+
+	fd = open(".", O_PATH);
+	if (fd < 0) {
+		pr_perror("unable to open current directory");
+		return fd;
+	}
+
+	ret = switch_ns(pid, &mnt_ns_desc, rst);
+	if (ret < 0) {
+		close(fd);
+		return ret;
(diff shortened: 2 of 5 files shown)


criu/namespaces.c
```diff
@@ -227,6 +227,46 @@ bool check_ns_proc(struct fd_link *link)
...
+	if (ret < 0) {
+		if (cwd_fd)
+			close(*cwd_fd);
+		return ret;
+	}
+
+	return 0;
+}
+
+int restore_mnt_ns(int rst, int *cwd_fd) {
```
rst0git, line 252:
```diff
-int restore_mnt_ns(int rst, int *cwd_fd) {
+int restore_mnt_ns(int rst, int *cwd_fd)
+{
```

criu/mount.c
```diff
@@ -1511,10 +1496,10 @@ static __maybe_unused int add_cr_time_mount(struct mount_info *root, char *fsnam
 static __maybe_unused int mount_cr_time_mount(struct ns_id *ns, unsigned int *s_dev, const char *source,
 					      const char *target, const char *type)
 {
-	int mnt_fd, ret, exit_code = 0;
+	int mnt_fd, cwd_fd, ret, exit_code = 0;
 	struct stat st;
 
-	ret = switch_ns(ns->ns_pid, &mnt_ns_desc, &mnt_fd);
+	ret = switch_mnt_ns(ns->ns_pid, &mnt_fd, & cwd_fd);
```
rst0git, line 1502:
```diff
-	ret = switch_mnt_ns(ns->ns_pid, &mnt_fd, & cwd_fd);
+	ret = switch_mnt_ns(ns->ns_pid, &mnt_fd, &cwd_fd);
```

criu/mount.c
```diff
@@ -1406,28 +1401,18 @@ int open_mountpoint(struct mount_info *pm)
...
 
-	if (fchdir(cwd_fd)) {
-		pr_perror("Unable to restore cwd");
-		close(cwd_fd);
-		close(fd);
-		return -1;
-	}
-	close(cwd_fd);
 
 	return __open_mountpoint(pm, fd);
```
rst0git, line 1410: Please remove the extra empty line:
```diff
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -1406,7 +1406,6 @@ int open_mountpoint(struct mount_info *pm)
 		goto err;
 	}
 
-
 	return __open_mountpoint(pm, fd);
```

liusdu:
Thanks @rst0git fixed and checked by make indent ~

criu/namespaces.c
```diff
@@ -227,6 +227,47 @@ bool check_ns_proc(struct fd_link *link)
...
+	if (ret < 0)
+		goto err_restore;
+
+	if (cwd_fd) {
+		ret = fchdir(*cwd_fd);
+	}
+
+err_restore:
+	if (cwd_fd) {
+		close(*cwd_fd);
```
avagin, line 266: close_safe

liusdu: fixed

Snorch (commented):
@liusdu Hi.

First, CONFIG_BINFMT_MISC_VIRTUALIZED is only applicable to Virtuozzo(/Openvz) container migration. There is no binfmt_misc virtualization in mainstream https://github.com/torvalds/linux/blob/master/fs/binfmt_misc.c#L794 The whole host has single binfmt_misc for everybody, so if you migrate your app from one host to another this code would probably do something bad with binfmt on destination host. So you probably doing something wrong.

If you actually migrate Virtuozzo container, then probably you need to use special criu version for Virtuozzo instead https://src.openvz.org/projects/OVZ/repos/criu/browse =).

Second, though I like the idea of the patch in general, you can see that binfmt_misc c/r code had changed in VZ criu https://src.openvz.org/projects/OVZ/repos/criu/commits/671d62cb30#criu/mount.c quiet a lot, and we don't have this problem anymore as we don't switch_ns() at all, because binfmt_misc is virtualized per ve (virtual environment - say virtuozzo container) not per mntns.

But anyway if it's in mainstream code base it is worth to get fixed.

Can you please fix commit messages, I would prefer adding component prefixes, something like:

```
mount: use swich_mnt_ns/restore_mnt_ns helpers to simplify code
binfmt_misc: restore current work directory after restoring mnt ns
namespaces: add helpers to switch/restore mnt ns
```

Also please see inline comments.

liusdu:
> @liusdu Hi.
> 
> First, CONFIG_BINFMT_MISC_VIRTUALIZED is only applicable to Virtuozzo(/Openvz) container migration. There is no binfmt_misc virtualization in mainstream https://github.com/torvalds/linux/blob/master/fs/binfmt_misc.c#L794 The whole host has single binfmt_misc for everybody, so if you migrate your app from one host to another this code would probably do something bad with binfmt on destination host. So you probably doing something wrong.

Thanks for you explain @Snorch , I  got  `BINFMT_MISC_VIRTUALIZED` in my .config.  But I found it is not for default. strange.. 
> 
> If you actually migrate Virtuozzo container, then probably you need to use special criu version for Virtuozzo instead https://src.openvz.org/projects/OVZ/repos/criu/browse =).
> 
My platform is CentOS, so it is my mistake to enable `BINFMT_MISC_VIRTUALIZED`...
> Second, though I like the idea of the patch in general, you can see that binfmt_misc c/r code had changed in VZ criu https://src.openvz.org/projects/OVZ/repos/criu/commits/671d62cb30#criu/mount.c quiet a lot, and we don't have this problem anymore as we don't switch_ns() at all, because binfmt_misc is virtualized per ve (virtual environment - say virtuozzo container) not per mntns.
> 

> But anyway if it's in mainstream code base it is worth to get fixed.
> 
> Can you please fix commit messages, I would prefer adding component prefixes, something like:
> 
> ```
> mount: use swich_mnt_ns/restore_mnt_ns helpers to simplify code
> binfmt_misc: restore current work directory after restoring mnt ns
> namespaces: add helpers to switch/restore mnt ns
> ```
> 
> Also please see inline comments.

criu/namespaces.c
```diff
@@ -227,6 +227,47 @@ bool check_ns_proc(struct fd_link *link)
...
+int switch_mnt_ns(int pid, int *rst, int *cwd_fd)
+{
+	int ret;
+	int fd;
+
+	if (cwd_fd) {
+		fd = open(".", O_DIRECTORY);
+		if (fd < 0)
+			return fd;
+		*cwd_fd = fd;
```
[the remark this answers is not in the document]
liusdu: fixed

criu/namespaces.c
```diff
@@ -227,6 +227,47 @@ bool check_ns_proc(struct fd_link *link)
 	return false;
 }
 
+int switch_mnt_ns(int pid, int *rst, int *cwd_fd)
```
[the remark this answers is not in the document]
liusdu: fixed

criu/namespaces.c
```diff
@@ -227,6 +227,47 @@ bool check_ns_proc(struct fd_link *link)
...
+	int fd;
+
+	if (cwd_fd) {
+		fd = open(".", O_DIRECTORY);
+		if (fd < 0)
+			return fd;
+		*cwd_fd = fd;
+	}
+
+	ret = switch_ns(pid, &mnt_ns_desc, rst);
```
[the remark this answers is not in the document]
liusdu: After looking into `switch_ns `, I find it works as you mentained~

criu/namespaces.c
```diff
@@ -227,6 +227,47 @@ bool check_ns_proc(struct fd_link *link)
...
+{
+	int ret = -1;
+
+	ret = restore_ns(rst, &mnt_ns_desc);
+	if (ret < 0)
+		goto err_restore;
+
+	if (cwd_fd) {
+		ret = fchdir(*cwd_fd);
+	}
```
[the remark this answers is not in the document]
liusdu: fixed

criu/namespaces.c
```diff
@@ -227,6 +227,47 @@ bool check_ns_proc(struct fd_link *link)
...
+
+err_restore:
+	if (cwd_fd) {
+		close_safe(cwd_fd);
+	}
+	return ret;
```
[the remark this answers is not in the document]
liusdu: fixed

liusdu:
@Snorch  thanks for your explanation and reivew, please take a look at the new patch.

Snorch (approved):
Thanks @liusdu ! Looks good.

Note, I see some other places where cwd restore is wrong... Here are some more fixes https://github.com/Snorch/criu/tree/restore-cwd-after-mntns-restore

You can incorporate them into you PR or I will just send separate PR later. Change in prep_unix_sk_cwd is quite a mind blowing one (how did it work before? =) ) so I would ask @mihalicyn to review it, may be I miss something...

liusdu:
> Thanks @liusdu ! Looks good.
> 
> Note, I see some other places where cwd restore is wrong... Here are some more fixes https://github.com/Snorch/criu/tree/restore-cwd-after-mntns-restore
> 
> You can incorporate them into you PR or I will just send separate PR later. Change in prep_unix_sk_cwd is quite a mind blowing one (how did it work before? =) ) so I would ask @mihalicyn to review it, may be I miss something...

@Snorch Thanks for you review.  I have put your commits together into this PR.

criu/namespaces.c
```diff
@@ -282,6 +282,46 @@ int restore_ns(int rst, struct ns_desc *nd)
...
+int switch_mnt_ns(int pid, int *rst, int *cwd_fd)
+{
+	int ret;
+	int fd;
+
+	if (!cwd_fd)
+		return switch_ns(pid, &mnt_ns_desc, rst);
+
+	fd = open(".", O_DIRECTORY);
+	if (fd < 0)
```
avagin, line 294: Need to print an error here in other places, otherwise it will be unclear what syscall returns an error.

criu/namespaces.c
```diff
@@ -282,6 +282,46 @@ int restore_ns(int rst, struct ns_desc *nd)
...
 
+int switch_mnt_ns(int pid, int *rst, int *cwd_fd)
+{
+	int ret;
+	int fd;
+
+	if (!cwd_fd)
+		return switch_ns(pid, &mnt_ns_desc, rst);
+
+	fd = open(".", O_DIRECTORY);
```
avagin, line 293: O_PATH looks more appropriate in this case, doesn't it?

liusdu: Yes, Done~
```
 O_PATH (since Linux 2.6.39)
              Obtain a file descriptor that can be used for two purposes: to indicate a location in the file-system tree and to perform operations that act purely  at  the  file
              descriptor  level
```

PR #1588, opened 2021-08-22
kerndat: use unique table names for has_nftables_concat check

ZeyadYasser:
There was a race window in kerndat_has_nftables_concat(void) when two or more instances of CRIU are running in parallel.

One instance would create the table normally, and another would fail when trying to create a table with the same name and wrongly set kdat.has_nftables_concat = false, and will save it to kerndat cache.

**NOTE:**
I have one concern about this approach but I think it is very rare. If two CRIU instance are in separate pid namespaces while sharing the same network namespace, they might have the same pid and hence the same table name. Should I go for randomly generating the name instead? Or just appending the pid will do?

Thanks @Snorch for pointing this out.


avagin:
>  Should I go for randomly generating the name instead?

I think you should use a random id here.

Snorch:
Generally it looks good.

I must agree with Andrei that random generated ids would probably be more robust than pids. 

Note: I don't see many uses of lrand48/rand in criu, don't forget to set random seed somewhere.

Note: it's never late to use netns isolation =)

criu/kerndat.c
```diff
@@ -1153,29 +1155,59 @@ int kerndat_has_nspid(void)
...
-	else
-		kdat.has_nftables_concat = true;
+	if (waitpid(pid, &status, 0) != pid) {
+		pr_perror("Unable to wait %d", pid);
+		return -1;
+	}
 
-	/* Clean up */
-	NFT_RUN_CMD(nft, "delete table inet CRIU-kerndat-test");
+	if (status < 0)
```
Snorch, line 1206: status would never be < 0...

Let's apply some debug patch:
```
[snorch@fedora criu]$ git diff
diff --git a/criu/kerndat.c b/criu/kerndat.c
index baae1a886..cae4d5df4 100644
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1166,6 +1166,7 @@ static int kerndat_has_nftables_concat(void)
        }
 
        if (pid == 0) {
+               exit(-1);
                /*
                 * Create a separate network namespace to avoid
                 * collisions between two CRIU instances.
@@ -1203,6 +1204,7 @@ static int kerndat_has_nftables_concat(void)
                return -1;
        }
 
+       pr_err("DEBUG %d\n", status);
        if (status < 0)
                return -1;
```
And run it:
```
[snorch@fedora criu]$ sudo criu/criu check --all
Warn  (criu/kerndat.c:846): Stale /run/criu.kdat file
Error (criu/kerndat.c:1207): DEBUG 65280
Looks good but some kernel features are missing
which, depending on your process tree, may cause
dump or restore failure.
```

Exit(-1) is equivalent of exit(255) if I recall correctly.


Note: Probably it's a good idea to reuse `call_in_child_process()`.

criu/kerndat.c
```diff
@@ -1153,29 +1155,59 @@ int kerndat_has_nspid(void)
...
+		return -1;
```
rst0git, line 1207: This condition for `status` doesn't work as expected. For example, with the following change:
```diff
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1195,6 +1195,7 @@ static int kerndat_has_nftables_concat(void)
 
        nft_ctx_free_out:
                nft_ctx_free(nft);
+               ret = -1;
                exit(ret);
        }
```

it shows:

```bash
$ sudo rm -f /run/criu.kdat && sudo criu check --feature network_lock_nftables
network_lock_nftables is supported
```

criu/kerndat.c
Snorch: E.g. like this https://github.com/checkpoint-restore/criu/pull/1591/commits/aa4b8b2b4d5269a099c1a66fd280e312285ed27d Feel free to merge to your patch if you are ok with it.

PR #1590, opened 2021-08-24
Restore cwd after mntns restore

Snorch:
Add this to check tests. Main PR here https://github.com/checkpoint-restore/criu/pull/1587


avagin:
I think all commits have been merged to the main pr.

PR #1591, opened 2021-08-24
Kerndat nftables concat fixup

Snorch:
This is just to run tests for fixup, original pr here https://github.com/checkpoint-restore/criu/pull/1588


Snorch:
All is merged in the original PR.

PR #1597, opened 2021-08-30
[GSoC 2021] Introduce io_uring dump/restore support

kkdwvd:
Currently, missing features are:
* Registered files support
* Registered eventfd support
* Registered buffers support
* Registered workqueue fd support
* ... and some other minor things depending on the above.

All of these depend on the eBPF iterator patches that are in https://github.com/kkdwivedi/linux/tree/criu-iter.

The following blockers exist currently:
* For registered files support, the io_uring_file iterator can be used to match on files present in the fdtable of task by matching directly on the struct file pointer. To make this safe against closing of files (so that such an iteration to gather data could potentially be moved to pre_dump stage and update itself dynamically), a new [file local storage map](https://lore.kernel.org/bpf/20210826133913.627361-1-memxor@gmail.com/) would be utilized.
* For eventfd support, finding the backing fd would rely on matching the eventfd context pointer with the one stashed in f_private of the fds held open by the task. This would require pairing the io_uring iterator and task_file iterator.
* Buffer registration takes reference to struct page internally, which means that the actual mapping can be destroyed after buffers are registered. This makes the task of mapping back it to a vma hard. The proposed solution to resolve is to introduce a eBPF iterator for io_uring buffers. Paired with with task_vma iterator, this should allow matching on the struct page backing the vma and the mapped buffer.
* Finding the backing workqueue for an io_uring (registered using wq_fd, another io_uring) relies on matching ctx->sq_data inside the kernel, which would again rely on eBPF iterator support.

For gathering data about io_uring itself:
* We can lift data like e.g. restrictions from the io_uring iteration stage itself, but there are stability issues with eBPF, so if e.g. member name is changed across kernel versions, it would require updating CRIU. The solution to this may be shipping stable eBPF iterators with the kernel (like modules) which maintain a stable output but can be modified internally across kernel versions. Right now the stopgap solution chosen is using fdinfo interface.

Signed-off-by: Kumar Kartikeya Dwivedi <<memxor@gmail.com>>


mihalicyn:
Hi, Kumar!

Please take a look on tests failures.
Looks like on CentOS 7 we need to install liburing-devel package and use header `liburing/io_uring.h` from liburing.
On CentOS 8 we get 362 tests failed:
```
un criu dump
=[log]=> dump/zdtm/transition/fifo_loop/54/1/dump.log
b'(00.010197) ----------------------------------------'
b'(00.010199)'
b'(00.010200) Collecting fds (pid: 54)'
b'(00.010201) ----------------------------------------'
b'(00.010208) Error (criu/cr-dump.c:214): pidfd_open system call not supported'
b'(00.010211) Error (criu/cr-dump.c:1326): Collect fds (pid: 54) failed with -1'
b'(00.014873) \tUnseizing 57 into 1'
b'(00.014877) \tUnseizing 58 into 1'
b'(00.014880) \tUnseizing 59 into 1'
b'(00.014883) \tUnseizing 60 into 1'
b'(00.014900) Error (criu/cr-dump.c:1834): Dumping FAILED.'
```

Regards,
Alex

criu/cr-dump.c
```diff
@@ -204,6 +206,59 @@ static int collect_fds(pid_t pid, struct parasite_drain_fd **dfds)
...
+
+			if (pidfd == -1) {
+				pidfd = syscall(SYS_pidfd_open, pid, 0);
+				if (pidfd < 0) {
+					pr_err("Failed to open pidfd for pid %d\n", pid);
+					return pidfd;
+				}
+			}
+
+			if (io_uring_synchronize_fd(syscall(__NR_pidfd_getfd, pidfd, atoi(de->d_name), 0))) {
```
rst0git, line 248: criu/cr-dump.c:248:40: error: '__NR_pidfd_getfd' undeclared (first use in this function)

adrianreber:
I started the CI tests which were not running.

kkdwivedi:
This should fix everything except the 'header not found' problem for <linux/io_uring.h>.

criu/mem.c
```diff
@@ -704,6 +709,8 @@ int prepare_mm_pid(struct pstree_item *i)
 			ret = collect_filemap(vma);
 		else if (vma_area_is(vma, VMA_AREA_SOCKET))
 			ret = collect_socket_map(vma);
+		else if (vma_area_is(vma, VMA_AREA_IO_URING))
+			ret = collect_io_uring_map(vma);
```
rst0git, line 713: We need to make sure that CRIU continues to work with older kernel versions that don't support io_uring. Adding a kerndat check for that might help.

criu/proc_parse.c
```diff
@@ -792,6 +817,7 @@ int parse_smaps(pid_t pid, struct vm_area_list *vma_area_list, dump_filemap_t du
 		if (handle_vma(pid, vma_area, str + path_off, map_files_dir, &vfi, &prev_vfi, &vm_file_fd))
 			goto err;
 
+		//pr_info("dump_filemap=%p IO_URING=%d?\n", dump_filemap, (int) vma_area_is(vma_area, VMA_AREA_IO_URING));
```
rst0git, line 820: This could be `pr_debug`?

rst0git:
> This should fix everything except the 'header not found' problem for <linux/io_uring.h>.

You can add `linux/io_uring.h` under `criu/include/linux/`.
For example, f68e5a6b3dd669cafab3cd3e5c556363d6ac357b adds `linux/userfaultfd.h` to solve a similar problem.

criu/proc_parse.c
```diff
@@ -1708,6 +1734,261 @@ static int parse_bpfmap(struct bfd *f, char *str, BpfmapFileEntry *bpf)
 	return 0;
 }
 
+static int parse_io_uring(struct bfd *f, char *str, struct io_uring_ctx *ctx)
+{
+	IoUringFileEntry *iofe = io_uring_get_iofe(ctx);
+	unsigned int nr;
+	pid_t pid;
+	int r;
```
rst0git, line 1742: `str`, `r`, and `f` are not very descriptive. Maybe we could use a better variable name?

kkdwivedi:
@rst0git Thanks for the review. I'll update this with your suggestions applied once I get the kernel side patches in order, which shouldn't take too long.

kkdwivedi:
A [prerequisite series](https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git/commit/?id=32a16f6bfe512a29fbe7e10ccaced118801a178f) for this and other stuff planned for later got in, now the iterators are left. After that (leaving out some specific things on CRIU side, like buffer support), this can move forward.

kkdwivedi:
The main patch for this has now been posted: https://lore.kernel.org/bpf/20211116054237.100814-1-memxor@gmail.com

kkdwivedi:
Greetings. v2 is out, now with an example (see Patch 10) tackling all the missing stuff listed in this PR (leaving out some of the details that are dependent on CRIU).

https://lore.kernel.org/bpf/20211122225352.618453-1-memxor@gmail.com

xinyangge-db:
As the io_uring interface is getting mature, is there any plan to resume the effort in supporting io_uring in CRIU?  We have a data-intensive application that is going to significantly benefit from it.

rahulk789:
Hello, I am just trying to setup and test this PR. I have a couple questions about the same:
- [(criu-iter)]( https://github.com/kkdwivedi/linux/tree/criu-iter) This kernel branch no longer seems to exist, May I get the updated branch? (if any)
-  Do we need more kernel patches to implement the left out features?
Thank you. cc: @kkdwivedi

kkdwivedi:
@rahulk789 The support in this PR is incomplete, there were a bunch of things to implement. The latest update was the kernel patch set I sent out https://lore.kernel.org/bpf/20211201042333.2035153-1-memxor@gmail.com.

I also had an example BPF sample demonstrating how the kernel support was to be used https://lore.kernel.org/bpf/20211201042333.2035153-11-memxor@gmail.com.

It might be difficult to rebase them onto the latest kernel as a lot has changed since then. If you're looking to implement this, I'd suggest taking inspiration from these changes but sketching out a path from scratch.

rahulk789:
Got, it. Thanks for the prompt reply!

rst0git:
> I unfortunately no longer have the time or bandwidth to continue pursuing this.
>
> It might be difficult to rebase them onto the latest kernel as a lot has changed since then. If you're looking to implement this, I'd suggest taking inspiration from these changes but sketching out a path from scratch.

@kkdwivedi Thank you for the pointers! To provide more context, Niranjan (@nravic) mentioned some time ago that this feature would be useful for them, and I believe Rahul is currently looking into completing the implementation to enable checkpoint/restore support with io_uring.

PR #1604, opened 2021-09-03
lib: install images/rpc.pb-c.h

rst0git:
Since commit 1c25914 compiling [crun](http://github.com/containers/crun) with libcriu also requires `/usr/include/criu/rpc.pb-c.h`

### DIFF
--- a/scripts/ci/podman-test.sh
+++ b/scripts/ci/podman-test.sh
@@ -11,6 +11,22 @@ make install
 
 criu --version
 
+# Install crun build dependencies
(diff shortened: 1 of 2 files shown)


scripts/ci/podman-test.sh
```diff
@@ -11,6 +11,22 @@ make install
 
 criu --version
 
+# Install crun build dependencies
```
adrianreber, line 14: Please add a comment that we install it from source to test libcriu

rst0git: Thanks, I've updated the comment.

Issue #1552, opened 2021-07-12
v3.15 tag is not "annotated"


kolyshkin:
I have noticed this:

```console
[kir@kir-rhat criu]$ ./criu/criu --version
Version: 3.15
GitID: v3.14-452-gf4201fdbe
```

The GitID should have v3.15 but it still shows v3.14.

This happens because `git describe` (used to generate GitID) only uses annotated tags by default, and v3.15 tag is not an annotated one:

```console
[kir@kir-rhat criu]$ git show v3.14
tag v3.14
Tagger: Pavel Emelyanov <xemul@openvz.org>
Date:   Wed Apr 29 16:35:32 2020 +0300

π

commit 95ead14874244f3c12e5970a74d1f4dd2433d652 (tag: v3.14)
Author: Pavel Emelyanov <xemul@parallels.com>
Date:   Wed Apr 29 16:31:49 2020 +0300
<...>

commit f68da4a86fee62af216e9e7520f4916aa29e797b (tag: v3.15)
Author: Andrei Vagin <avagin@gmail.com>
Date:   Thu Oct 29 18:41:48 2020 +0300

    criu: Version 3.15
    
<...>
```

Here's an excerpt from `git tag` man page:

<blockquote>

       Tag objects (created with -a, -s, or -u) are called "annotated" tags; they contain a
       creation date, the tagger name and e-mail, a tagging message, and an optional GnuPG
       signature. Whereas a "lightweight" tag is simply a name for an object (usually a commit
       object).

       Annotated tags are meant for release while lightweight tags are meant for private or
       temporary object labels. For this reason, some git commands for naming objects (like
       git describe) will ignore lightweight tags by default.

</blockquote>

So, there are two ways to fix this.

1. Replace `v3.15` tag with an annotated one.
2. Something like
```diff
diff --git a/Makefile b/Makefile
index 14769b571..6dbd5c95b 100644
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ include Makefile.versions
 
 VERSION_HEADER         := criu/include/version.h
 GITID_FILE             := .gitid
-GITID          := $(shell if [ -d ".git" ]; then git describe --always; fi)
+GITID          := $(shell if [ -d ".git" ]; then git describe --always --tags; fi)
 
 # Git repository wasn't inited in CRIU folder
 ifeq ($(GITID),)
```

I strongly prefer the first option.

kolyshkin:
[...]

PR #1607, opened 2021-09-06
CRIU 3.16

adrianreber:
This brings the last few fixups from `criu-dev` over for the 3.16 release.

@checkpoint-restore/maintainers PTAL

If possible I would like to do the 3.16 release this week.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ lib/.crit-setup.files
 compel/include/asm
 include/common/asm
 include/common/config.h
+build/
--- a/criu/autofs.c
+++ b/criu/autofs.c
@@ -391,7 +391,7 @@ static int access_autofs_mount(struct mount_info *pm)
 	const char *mnt_path = pm->mountpoint + 1;
 	dev_t dev_id = pm->s_dev;
 	int new_pid_ns = -1, old_pid_ns = -1;
-	int old_mnt_ns;
+	int old_mnt_ns, old_cwd_fd;
 	int autofs_mnt;
 	int err = -1;
 	int pid, status;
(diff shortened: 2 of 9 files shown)


Makefile.versions
```diff
@@ -1,10 +1,10 @@
 #
 # CRIU version.
 CRIU_VERSION_MAJOR	:= 3
-CRIU_VERSION_MINOR	:= 15
+CRIU_VERSION_MINOR	:= 16
 CRIU_VERSION_SUBLEVEL	:=
 CRIU_VERSION_EXTRA	:=
-CRIU_VERSION_NAME	:= Titanium Falcon
+CRIU_VERSION_NAME	:= Petrified Puffin (TODO)
```
rst0git, line 7: Is `(TODO)` in the version name intentional?

adrianreber: No. Removed.

rst0git:
image

rst0git (approved):

Snorch (approved):
Looks good.

Note: last time we had some release-tag annotation problem https://github.com/checkpoint-restore/criu/issues/1552 Probably we need to create tag manually or something like this...

adrianreber:
> Looks good.
> 
> Note: last time we had some release-tag annotation problem #1552 Probably we need to create tag manually or something like that...

Thanks for the reminder. I will tag it first locally on my system and then push the tag.

avagin:
Have you prepared a wiki page with release notes?
How is the Jenkins?

adrianreber:
> Have you prepared a wiki page with release notes?

I will do that today. Probably.

> How is the Jenkins?

Pretty good. The ppc64le tests are, however, not running for a over a month now. @mihalicyn any reason ppc64le is not running. Is it still the cgroup race condition problem? Maybe we should a change for criu-dev which runs at least all other tests on ppc64le even if we cannot fix the cgroup problem as it seems to be kernel related.

adrianreber:
@checkpoint-restore/maintainers Can I get a couple of approvals to release 3.16?

rst0git:
@adrianreber would it be possible to add https://github.com/checkpoint-restore/criu/pull/1609 as well?

adrianreber:
> @adrianreber would it be possible to add #1609 as well?

Done

avagin:
@adrianreber could you add a link to the wiki page with this release? As I already said, we need to push all commits in another PR to be sure that all CI-s are happy.

adrianreber:
https://criu.org/Download/criu/3.16

PR #1608, opened 2021-09-06
bfd: avoid memory move by using ring buffer in bfd

anatasluo:
By using ring buffer in bfd, we can avoid moving memory.
And the implementation will be more simple.

And, lockless ring buffer also supports two threads work
at the same time which is helpful to deal with the situation
when the size of the proc file is up to tens of megabytes.

Signed-off-by: anatasluo <luolongjuna@gmail.com>

### DIFF
--- a/criu/bfd.c
+++ b/criu/bfd.c
@@ -1,3 +1,7 @@
+#ifndef _GNU_SOURCE
+#define _GNU_SOURCE
+#endif
+
 #include <unistd.h>
 #include <stdbool.h>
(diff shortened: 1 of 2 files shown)


anatasluo:
Hi guys, I am trying to optimize the performance of read/write of proc file in criu. My /proc/xxx/smaps has a size of 42M, it consumes some time. I wonder if I use one thread to read and the other to write at the same time will help to solve this problem. Has anyone tried it before?

PR #1612, opened 2021-09-17
tty: fix the null pointer of get_tty_driver

time-river:
as the title


criu/tty.c
```diff
@@ -1977,7 +1977,10 @@ static int dump_one_tty(int lfd, u32 id, const struct fd_parms *p)
 	pr_info("Dumping tty %d with id %#x\n", lfd, id);
 
 	driver = get_tty_driver(p->stat.st_rdev, p->stat.st_dev);
-	if (driver->fd_get_index)
+	if (driver == NULL) {
+		pr_err("Unable to find a tty driver (rdev %#lx dev %#lx)\n", p->stat.st_rdev, p->stat.st_dev);
+		return -1;
+	} else if (driver->fd_get_index)
```
rst0git, line 1983: There is no need to add `else` here because `return -1;` would prevent the execution of subsequent instructions.

adrianreber:
Same broken Rawhide tests as everywhere right now. Besides that, this looks ready.

PR #1615, opened 2021-09-20
ci: Update openj9 container images

rst0git:
The AdoptOpenJDK docker images have been deprecated in favor of
eclipse-temurin.

https://github.com/AdoptOpenJDK/openjdk-docker/pull/604

### DIFF
--- a/scripts/build/Dockerfile.openj9-ubuntu
+++ b/scripts/build/Dockerfile.openj9-ubuntu
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk8-openj9:latest
+FROM docker.io/library/eclipse-temurin:8-focal
 ARG CC=gcc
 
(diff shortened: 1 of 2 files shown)


scripts/build/Dockerfile.openj9-ubuntu
```diff
@@ -1,4 +1,4 @@
-FROM adoptopenjdk/openjdk8-openj9:latest
+FROM eclipse-temurin:8-focal
```
adrianreber, line 1: If we change it we could move away from short-names to a container name including the registry. Can you change that?

rst0git: You're right, I appreciate your mentioning that.
I've replaced the short-name with `docker.io/library/eclipse-temurin:8-focal` to include the registry.

PR #1617, opened 2021-09-21
libcriu: add API for join-ns option

rst0git:
We use the join-ns RPC API in runc to enable checkpoint/restore of containers with shared namespaces. Shared namespaces are often used when containers run inside Kubernetes Pod. However, crun is using libcriu to interface with CRIU and libcriu currently doesn't provide an API for join-ns. This pull request adds the libcriu API necessary to enable checkpoint/restore of containers with shared namespaces with crun.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -431,6 +431,7 @@ lint:
 	shellcheck scripts/*.sh
 	shellcheck scripts/ci/*.sh scripts/ci/apt-install
 	shellcheck test/others/crit/*.sh
+	shellcheck test/others/libcriu/*.sh
 	shellcheck test/others/config-file/*.sh
 	# Do not append \n to pr_perror or fail
 	! git --no-pager grep -E '^\s*\<(pr_perror|fail)\>.*\\n"'
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -214,6 +214,18 @@ void criu_local_free_opts(criu_opts *opts)
 	}
 	opts->rpc->n_external = 0;
 
+	if (opts->rpc->join_ns) {
+		for (i = 0; i < opts->rpc->n_join_ns; i++) {
+			free(opts->rpc->join_ns[i]->ns);
+			free(opts->rpc->join_ns[i]->ns_file);
+			if (opts->rpc->join_ns[i]->extra_opt) {
+				free(opts->rpc->join_ns[i]->extra_opt);
+			}
+			free(opts->rpc->join_ns[i]);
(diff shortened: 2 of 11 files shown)


adrianreber:
I was a bit confused about the interface. Path names instead of file descriptors. Looking at the RPC and CLI interface I think your implementation is correct as RPC and CLI also use path names and not file descriptors. Which is confusing because inherit-fd uses file descriptors.

Everything looks good. Unfortunately the arm tests are currently not running. I will wait for the results from those tests and then it looks ready to be merged.

rst0git:
Thank you Adrian for the code review. I was hoping to ask for feedback for the test case of join-ns. It would currently check that checkpoint/restore has succeeded. While this is sufficient to check if the join-ns option has been passed to CRIU (example below), we don't currently have tests to verify that `--join-ns` actually works. Perhaps we could add the tests in runc/crun?
```bash
sudo make -C test/others/libcriu/ run
sudo cat test/others/libcriu/wdir/i/test_join_ns/restore.log | grep 'join namespace'
(00.000244) Added ipc:/proc/66302/ns/ipc join namespace
(00.000264) Added uts:/proc/66302/ns/uts join namespace
(00.000281) Added time:/proc/66302/ns/time join namespace
```
@avagin do you have any advice?

adrianreber:
> Thank you Adrian for the code review. I was hoping to ask for feedback for the test case of join-ns. It would currently check that checkpoint/restore has succeeded. While this is sufficient to check if the join-ns option has been passed to CRIU (example below), we don't currently have tests to verify that `--join-ns` actually works. Perhaps we could add the tests in runc/crun?
> 
> ```shell
> sudo make -C test/others/libcriu/ run
> sudo cat test/others/libcriu/wdir/i/test_join_ns/restore.log | grep 'join namespace'
> (00.000244) Added ipc:/proc/66302/ns/ipc join namespace
> (00.000264) Added uts:/proc/66302/ns/uts join namespace
> (00.000281) Added time:/proc/66302/ns/time join namespace
> ```
> 
> @avagin do you have any advice?

I would compare `/proc/<PID>/ns/<namespace>` just as we do in other tests.

rst0git:
> I would compare `/proc/<PID>/ns/<namespace>` just as we do in other tests.

Thanks, I've updated the test case.

adrianreber:
I see that you changed `run.sh`. Could you add it to the shellcheck test targets?

adrianreber (approved):

PR #1620, opened 2021-09-27
criu-ns: change python shebang to python3

rst0git:
[PEP 394](https://www.python.org/dev/peps/pep-0394/) recommends changing `python` shebangs to `python3` when Python 3.x is supported. This is similar to `crit-python3`.

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -12,7 +12,9 @@ endif
 
 FOOTER		:= footer.txt
 SRC1		+= crit.txt
+ifeq ($(PYTHON),python3)
 SRC1		+= criu-ns.txt
+endif
 SRC1		+= compel.txt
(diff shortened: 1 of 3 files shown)


adrianreber:
But this breaks python2 platforms. `make install` needs to handle the python2 and python3 interpreters correctly. Not sure what to do here. Maybe the same dance is necessary as with crit.

adrianreber:
> Since Python 2 has been deprecated for a while now, maybe we could support `criu-ns` as Python 3 only?

Works for me, but then `make install` should not install it if PYTHON is set to `python2`.

rst0git:
I've renamed the branch in fork and I had to open another PR: https://github.com/checkpoint-restore/criu/pull/1621

PR #1621, opened 2021-09-28
criu-ns: change python shebang to python3

rst0git:
PEP 394 recommends changing python shebangs to python3 when Python 3.x is supported. This is similar to crit-python3.

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -12,7 +12,9 @@ endif
 
 FOOTER		:= footer.txt
 SRC1		+= crit.txt
+ifeq ($(PYTHON),python3)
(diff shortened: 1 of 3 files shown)


rst0git:
> Looking at the CI logs I do not see criu-ns installed (or uninstalled). Is that expected?

Yes, we didn't print it out an install message for `criu-ns`. I've updated the commit to add one. It should appear as:
```bash
INSTALL  criu/criu
INSTALL  scripts/criu-ns
```

adrianreber:
Looks good now. Thanks. I will wait for the CI results and then it is ready to be merged.

Issue #1533, opened 2021-07-01
Cannot checkpoint anonymous hugepages


geminijun:
**Description**
Tried to checkpoint a program using anonymous hugepages, got error "Can't lookup mount=16 for fd=-3 path=/anon_hugepage (deleted)"

**Steps to reproduce the issue:**
1. Wrote a program using anonymous hugepage, `mmap(NULL, 134217728, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS|MAP_HUGETLB, -1, 0)` and run it on a system enabled huge page.
2. CRIU checkpoint will be failed.

**Describe the results you received:**
Tried to research this and have no idea why the mnt id of the anonymous hugepage is 16, and I cannot find this number anywhere on the system.

```
# grep "^16\>" /proc/*/mountinfo

# pmap -p `pgrep test_huge` | grep anon_hugepage
00007fd2b7200000 131072K rw--- /anon_hugepage (deleted)

# exec 100< /proc/`pgrep test_huge`/map_files/7fd2b7200000-7fd2bf200000
# cat /proc/$$/fdinfo/100 
pos:    0
flags:  0100000
mnt_id: 16
```

Use `--external` can probably get around this, but since the program can mmap a new anonymous huge page at any time, it's not feasible to get the inode number of all the huge page mappings before checkpoint.

I also feel like this is something we probably can make CRIU just support it naturally, after all, it's an anonymous huge page, we can dump the data, restore the hugepage mapping and copy the data back. Isn't it?

**Describe the results you expected:**
Can successfully checkpoint a program using anonymous hugepage.

**Additional information you deem important (e.g. issue happens only occasionally):**

**CRIU logs and information:**

<details><summary>CRIU full dump/restore logs:</summary>
<p>

```
(paste your output here)
```

</p>
</details>

<details><summary>Output of `criu --version`:</summary>
<p>

```
(paste your output here)
```

</p>
</details>

<details><summary>Output of `criu check --all`:</summary>
<p>

```
(paste your output here)
```

</p>
</details>

**Additional environment details:**

geminijun:
Hi @Snorch , do you have any idea about this?

Snorch:
Looks like MAP_HUGETLB mappings are not supported:

We don't even have a zdtm test for it, but adding this hunk to maps01:

```
diff --git a/test/zdtm/static/maps01.c b/test/zdtm/static/maps01.c
[...]

PR #1622, opened 2021-09-28
Add support for MAP_HUGETLB mappings

minhbq-99:
- Collect hugetlb devices' numbers in kerndat. These numbers are used to
  determine whether a memory mapping is backed by hugetlb. We try to
  collect the missing hugetlb devices' numbers every time we load kerndat

- Anon shmem, private memory mappings backed by hugetlb are collected
  using memfd approach as long as memfd support hugetlb.

- Collect and restore hugetlb flags for System V shared memory, memfd and
  anon shmem, private memory in case memfd does not support hugetlb.

- Add a feature check for using memfd with hugetlb.

Fixes: #1533


minhbq-99:
My fault that I didn't well test my pull request. I also realize maybe using the same method of dumping memfd in hugetlb mapping dumping is not a good solution as `MFD_HUGETLB` is not added until linux 4.14

criu/ipc_ns.c
```diff
@@ -354,6 +354,57 @@ static int dump_ipc_shm_pages(const IpcShmEntry *shm)
...
+	unsigned int major, minor;
+	unsigned long start;
+
+	addr = shmat(id, NULL, SHM_RDONLY);
+	if (addr == (void *)-1) {
+		pr_perror("Failed to attach shm");
+		return -1;
+	}
+
+	f.fd = open_proc(PROC_SELF, "maps");
```
avagin, line 372: Can we avoid parsing maps for the criu process? We already parse maps for dumped processes. Can we get all required info from there?

minhbq-99: I try to have a generic approach for dumping device's number of shm as some shm(s) may not be mapped (the maps of dumped process does not have device's number of these shm(s)). I'm looking at shmctl but can't not see any ways to get device's number. So my solution is when collecting shm identifier, we need to map that shm then get the device's number.
Your concern is right, I didn't think about it. I try to change to use stat on map_files symlink instead
https://github.com/minhbq-99/criu/commit/7f252688754d1c943b6ba0142ba8fb99595c120e

But I get CI error in userfaultfd which I still don't understand.

minhbq-99:
I force push some changes
- shm: Use map_files symlink to get device's number instead of parsing proc maps
- shmem: Use memfd for hugetlb shared mappings whenever kernel supports
Here is the main change
```
@@ -566,12 +566,14 @@ static int open_shmem(int pid, struct vma_area *vma)
-       if (kdat.has_memfd && !is_hugetlb) {
+       if (kdat.has_memfd && (!is_hugetlb || kdat.has_memfd_hugetlb)) {
```
- zdtm: Disable uns test case in maps09 (test mmap MAP_HUGETLB). When we reach open_shmem(), we first try to create memfd for that shared mapping, if we cannot, we must open map_files symlink. However, this code may be run in the restored process which is possibly in user namespace and in that case, we cannot open map_files symlink (even in newer kernel with less restriction than kernel 3.10 in CentOS 7). Currently, only CentOS 7 fails the uns maps09 test case since it does not support MFD_HUGETLB with memfd.

avagin:
> However, this code may be run in the restored process which is possibly in user namespace and in that case, we cannot open map_files symlink (even in newer kernel with less restriction than kernel 3.10 in CentOS 7). Currently, only CentOS 7 fails the uns maps09 test case since it does not support MFD_HUGETLB with memfd.

We have userns_call to handle such cases:

```
diff --git a/criu/namespaces.c b/criu/namespaces.c
index 7fa58682b..c36e631cf 100644
--- a/criu/namespaces.c
+++ b/criu/namespaces.c
@@ -1330,11 +1330,6 @@ static int usernsd(int sk)
 		unsc_msg_pid_fd(&um, &pid, &fd);
 		pr_debug("uns: daemon calls %p (%d, %d, %x)\n", call, pid, fd, flags);
 
-		if (fd < 0 && flags & UNS_FDOUT) {
-			pr_err("uns: bad flags/fd %p %d %x\n", call, fd, flags);
-			BUG();
-		}
-
 		/*
 		 * Caller has sent us bare address of the routine it
 		 * wants to call. Since the caller is fork()-ed from the
index c5ba9a61b..5f70bafd9 100644
--- a/criu/shmem.c
+++ b/criu/shmem.c
@@ -26,6 +26,7 @@
 #include "memfd.h"
 #include "protobuf.h"
 #include "images/pagemap.pb-c.h"
+#include "namespaces.h"
 
 #ifndef SEEK_DATA
 #define SEEK_DATA 3
@@ -534,6 +535,17 @@ out:
 	return ret;
 }
 
+struct open_map_file_args {
+	unsigned long addr, size;
+};
+
+static int open_map_file(void *args, int fd, pid_t pid)
+{
+	struct open_map_file_args *vma = args;
+
+	return open_proc_rw(pid, "map_files/%lx-%lx", vma->addr, vma->addr + vma->size);
+}
+
 static int open_shmem(int pid, struct vma_area *vma)
 {
 	VmaEntry *vi = vma->e;
@@ -607,7 +619,11 @@ static int open_shmem(int pid, struct vma_area *vma)
 	}
 
 	if (f == -1) {
-		f = open_proc_rw(getpid(), "map_files/%lx-%lx", (unsigned long)addr, (unsigned long)addr + si->size);
+		struct open_map_file_args vma = {
+			.addr = (unsigned long) addr,
+			.size = si->size,
+		};
+		f = userns_call(open_map_file, UNS_FDOUT, &vma, sizeof(vma), 0);
 		if (f < 0)
 			goto err;
 	}
```

criu/kerndat.c
```diff
@@ -226,6 +235,46 @@ static int kerndat_get_shmemdev(void)
...
+		size = hugetlb_info[i].size;
+		flag = hugetlb_info[i].flag;
+		map = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_HUGETLB | flag, 0, 0);
+		if (map == MAP_FAILED && errno == EINVAL) {
+			kdat.hugetlb_dev[i] = (dev_t)-1;
+			continue;
+		}
+
+		if (map == MAP_FAILED && errno == ENOMEM) {
+			pr_info("Hugetlb size %llu Mb is supported but cannot get dev's number\n", size >> 20);
```
rppt, line 259: Here `kdat.hugetlb_dev[i]`will be 0. Is this what you expect?

criu/kerndat.c
```diff
@@ -226,6 +235,46 @@ static int kerndat_get_shmemdev(void)
...
+			continue;
+		} else if (map == MAP_FAILED) {
```
rppt, line 261: `if (map == MAP_FAILED)` repeats three times here. Consider nesting `errno` checks under it.

criu/kerndat.c
```diff
@@ -1242,8 +1308,15 @@ int kerndat_init(void)
...
 	ret = kerndat_try_load_cache();
-	if (ret <= 0)
+	if (ret < 0)
 		return ret;
+
+	/* Try to collect missing hugetlb dev in kerndat cache */
+	if (ret == 0) {
+		ret = kerndat_get_hugetlb_dev();
+		return ret;
+	}
```
rppt, line 1318: Please add a comment why we have to re-detect hugetlb devices after successful load of the cache. Don't we want to update the cache here as well?

Also, you can simply this with
```
	if (ret == 0)
		return kerndat_get_hugetlb_dev();
```

criu/kerndat.c
```diff
@@ -1260,6 +1333,10 @@ int kerndat_init(void)
 		pr_err("kerndat_get_shmemdev failed when initializing kerndat.\n");
 		ret = -1;
 	}
+	if (!ret && kerndat_get_hugetlb_dev()) {
+		pr_err("kerndat_get_hugetlb_dev failed when initializing kerndat.\n");
+		ret = -1;
+	}
```
rppt, line 1339: This call seems redundant here. IMO, it would better to rework logic aroind re-detection and cache loading so that we won't need the repeated call to `kerndat_get_hugetlb_dev()`

criu/kerndat.c
minhbq-99: Yes, it is what I want. `kdat.hugetlb_dev[i]` will be -1 if we detect that hugetlb size is not supported, 0 if it is supported but we cannot mmap (this is based on my assumption that the major, minor number cannot be both 0, I didn't look carefully at this https://github.com/torvalds/linux/blob/master/fs/super.c#L1032). Each time `kerndat_get_hugetlb_dev()` gets called, it only look at `kdat.hugetlb_dev[i] == 0` to try to get the device number of the hugetlb size. I will add comment on this.

criu/kerndat.c
minhbq-99: I forget to update the cache here.
`kerndat_get_hugetlb_dev()` only tries to update the device number that at the previous time we check, we see that hugetlb size is supported but we cannot mmaped to get that device number. As the hugetlb page can be adjusted between run, the hugetlb size that fails before may be available now.
I will adjust `kerndat_get_hugetlb_dev()` so that it will return -1 if there is an error, 0 if it is successful but find nothing new, 1 if it is successful and we find some new hugetlb device number. We will only update the cache in case that function return 1

criu/kerndat.c
minhbq-99: I think this call is needed at the first time we collect kerndat or when we fail to load the cache

criu/files.c
```diff
@@ -548,7 +548,8 @@ static int dump_one_file(struct pid *pid, int fd, int lfd, struct fd_opts *opts,
 
 		p.link = &link;
 
-		if (is_memfd(p.stat.st_dev))
+		/* TODO: Dump for hugetlb fd when memfd hugetlb is not supported */
+		if (is_memfd(p.stat.st_dev) || (is_hugetlb_dev(p.stat.st_dev, NULL) && kdat.has_memfd_hugetlb))
```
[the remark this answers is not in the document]
minhbq-99: I use `is_hugetlb_dev` in some other places that are not related to memfd like in shm_sysv so I don't combine these two

criu/hugetlb.c
```diff
@@ -0,0 +1,38 @@
...
+#include "kerndat.h"
+
+struct htlb_info hugetlb_info[HUGETLB_MAX] = {
+	[HUGETLB_16KB] = { 16 * KB, MAP_HUGETLB_16KB },	   [HUGETLB_64KB] = { 64 * KB, MAP_HUGETLB_64KB },
+	[HUGETLB_512KB] = { 512 * KB, MAP_HUGETLB_512KB }, [HUGETLB_1MB] = { 1 * MB, MAP_HUGETLB_1MB },
+	[HUGETLB_2MB] = { 2 * MB, MAP_HUGETLB_2MB },	   [HUGETLB_8MB] = { 8 * MB, MAP_HUGETLB_8MB },
+	[HUGETLB_16MB] = { 16 * MB, MAP_HUGETLB_16MB },	   [HUGETLB_32MB] = { 32 * MB, MAP_HUGETLB_32MB },
+	[HUGETLB_256MB] = { 256 * MB, MAP_HUGETLB_256MB }, [HUGETLB_512MB] = { 512 * MB, MAP_HUGETLB_512MB },
+	[HUGETLB_1GB] = { 1 * GB, MAP_HUGETLB_1GB },	   [HUGETLB_2GB] = { 2 * GB, MAP_HUGETLB_2GB },
+	[HUGETLB_16GB] = { 16 * GB, MAP_HUGETLB_16GB },
```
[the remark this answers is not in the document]
minhbq-99: This format is because of the clang-format, I can't find any option to fix this.

What is your suggestion here? At first, I think of using macro to convert between size and flag. However, I will still need to keep a list of valid size and I want to iterate on that list like in `kerndat_get_hugetlb_dev` so I choose an easy solution to use an array like above.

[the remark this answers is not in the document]
rppt: > This format is because of the clang-format, I can't find any option to fix this.

Oh, my... @adrianreber can we do something about this?

> What is your suggestion here?

Maybe I'm missing somthing but it seems to me that a single array  of valid sizes and a macro to convert a size to flag would suffice.

criu/kerndat.c
rppt: What I meant is that there is a call `kerndat_get_hugetlb_dev()` if we succeeded loading the cache, and then the call here when the cache is not available. We might leave only the call `kerndat_get_hugetlb_dev()` in the beginning of the function and adjust results checks accordingly.

criu/hugetlb.c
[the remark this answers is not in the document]
adrianreber: See #1644 for possible workarounds

criu/kerndat.c
minhbq-99: I have looked at bringing this call to the beginning of the function but in my opinion it adds more complexity but not so necessary. The reason is that there is another case when loading cache is that the cache is corrupted and we have to collect all the things again. Currently, `kerndat_get_hugetlb_dev()` collect directly into kdat, but if we move this before loading cache, we may need to change `kerndat_get_hugetlb_dev()` to load to a temporary buffer instead than copy to kdat later. Moreover, I change `kerndat_get_hugetlb_dev()` a little bit, so it return 0 on successful but no new hugetlb device's number is collect, 1 if successful and we collect new device's numbers and -1 on error. With this, I need to place this function after loading cache.

criu/hugetlb.c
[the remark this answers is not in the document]
minhbq-99: > Maybe I'm missing somthing but it seems to me that a single array of valid sizes and a macro to convert a size to flag would suffice

I use the array of valid sizes with something like `arr[i]` so I think it's not an easy way to use macro here

test/zdtm/static/shm.c
```diff
@@ -133,6 +138,12 @@ int main(int argc, char **argv)
 		goto out;
 	}
 
+	dev = get_mapping_dev(mem);
+	if (dev == (dev_t)-1) {
+		fail("Can't get mapping dev");
+		return 1;
```
rst0git, line 144:
```suggestion
		return -1;
```

test/zdtm/static/shm.c
```diff
@@ -155,6 +166,11 @@ int main(int argc, char **argv)
 		return -1;
 	}
 
+	if (dev != get_mapping_dev(mem)) {
+		fail("Mapping dev mismatch");
+		return 1;
```
rst0git, line 171:
```suggestion
		return -1;
```

adrianreber:
Rebase should fix the CentOS 8 error. The PR sounds like a good thing to have. @rppt is this ready from your point of view?

criu/namespaces.c
```diff
@@ -1330,11 +1330,6 @@ static int usernsd(int sk)
 		unsc_msg_pid_fd(&um, &pid, &fd);
 		pr_debug("uns: daemon calls %p (%d, %d, %x)\n", call, pid, fd, flags);
 
-		if (fd < 0 && flags & UNS_FDOUT) {
-			pr_err("uns: bad flags/fd %p %d %x\n", call, fd, flags);
-			BUG();
-		}
```
rst0git, line 1336: @minhbq-99 This change looks unrelated to the commit. Why do we need it?

adrianreber:
Pinging @rppt once more.

rst0git (approved):
LGTM

Snorch:
Can we please separate first patch into several separate logical entities?

kerndat_has_memfd_hugetlb should be one patch
kerndat_get_hugetlb_dev should likely be another patch
and so on.

Looks like there should be at least 4 patches:
```
    - Collect hugetlb devices' numbers in kerndat. These numbers are used to
      determine whether a memory mapping is backed by hugetlb. We try to
      collect the missing hugetlb devices' numbers every time we load kerndat
    
    - Anon shmem, private memory mappings backed by hugetlb are collected
      using memfd approach as long as memfd support hugetlb.
    
    - Collect and restore hugetlb flags for System V shared memory, memfd and
      anon shmem, private memory in case memfd does not support hugetlb.
    
    - Add a feature check for using memfd with hugetlb.
```

JFYI: https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#separate-your-changes

minhbq-99:
I split the patch into smaller commits.

The criu-image-streamer hangs seem related to the patch but I haven't found the reason yet.

I currently disable lazy dumping in zdtm tests as userfaultfd for hugetlb seems not ready yet. But there is another reason, when the trace is restore_task_with_children()->prepare_mappings()->premap_priv_vmas()->premap_private_vma(), there is alignment issue when using mmap with MAP_FIXED. I think it is not hard to fix this. Later, we use mremap() to remap these mapping to the correct address but mremap() is not supported hugetlb until recently (https://github.com/torvalds/linux/commit/550a7d60bd5e35a56942dba6d8a26752beb26c9f). So maybe we need to mark these hugetlb memory mappings and don't premap them or we need to remap them by creating new mapping and copying data. What do you think?

minhbq-99:
I add 2 more commits to skip premapping and lazy mode on hugetlb page to fix the error in image-streamer and CentOS 7 CI.

However, as I see image-streamer does not run on maps* test cases
https://github.com/checkpoint-restore/criu/blob/4152a88cf588e94c9bf6edb3706cdf81fd25bce2/scripts/ci/run-ci-tests.sh#L64-L69

I try image-streamer on my maps09 test case
`./test/zdtm.py run --ignore-taint --stream -t zdtm/static/maps09`

and get this error in restore.log
```
(00.022604)      5: Opening 0x007f42edc00000-0x007f42ee000000 0000000000000000 (4041) vma
(00.022610)      5: Restoring memfd id=8
(00.022774)      5: Error (criu/protobuf.c:93): Read 0 bytes while 32746 expected from socket:[118265]: Connection reset by peer
(00.022782)      5: Error (criu/img-streamer.c:130): Failed to communicate with the image streamer: Connection reset by peer
(00.022787)      5: Error (criu/image.c:475): Unable to open pagemap-shmem-115472.img: Input/output error
(00.022790)      5: Error (criu/shmem.c:526): Can't restore shmem content
(00.022801)      5: Error (criu/mem.c:1352): `- Can't open vma
(00.023281)      1: Error (criu/cr-restore.c:1476): 5 exited, status=1
```

The img-streamer.log
```
criu-image-streamer Error: vmsplice() failed on fd 3
```
I try to print the errno and get `EPIPE`

Do you have any ideas of possible cause of this?

adrianreber:
@nviennot any ideas/suggestions about the streamer problems

nviennot:
I'm on it

nviennot:
I've looked a little bit.
First, criu-image-streamer has been updated to display a little more useful error message than it used to 😇
Second, I've enabled all the tests with streaming, and it passes: https://github.com/checkpoint-restore/criu/pull/1715

The problem seems to be the following:
* CRIU requests a file `pages-5.img`
* criu-image-streamer starts sending the content of that file to CRIU.
* CRIU doesn't care about the content because of commit `mem: Skip premapping hugetlb mapping. As we cannot use mremap() to move the hugetlb mapping around until Linux kernel version 5.16, we need to skip premapping hugetlb mapping.`. It skips, closes the file and goes to access `pagemap-shmem-286975.img`
* criu-image-streamer yells that it cannot send `pages-5.img` and exits.

Currently, all the files that are being requested are being served and consumed sequentially.

I tried to remove the comment that skips premapping hugetlb, but that ends in a deadlock where CRIU asks for file `pagemap-shmem-286975.img`, and criu-image-streamer tries to send `pages-5.img`, which CRIU doesn't read.
Note that if I remove `MAP_HUGETLB`  from maps09.c, the test passes. Is it because the UNIX pipe buffer between CRIU and criu-image-streamer is large enough that when we don't use MAP_HUGETLB, we don't see the deadlock? Probably. This is fixable.

Note that the streamer will not send a file twice. This is for memory usage optimization, so skipping premapping is a bit difficult because that means that CRIU will close to reopen the `pages-5.img` file later on, and that's not going to fly.
So if you can fix the file re-opening (because you can't do premapping), and I look into the deadlock, we have a win

minhbq-99:
@nviennot Thank you very much for a great explanation of how image-stream works.

Looks like there are still a lot of things to work on.

nviennot:
I've implemented concurrent streaming (used during restore) for the streamer here: https://github.com/checkpoint-restore/criu-image-streamer/tree/nonblocking
This allows CRIU to open multiple files at once (like `pages-%d.img` and `pagemap-shmem-286975.img`) during restore. You can use it if you see tests hanging.

criu/kerndat.c
```diff
@@ -420,6 +469,23 @@ static bool kerndat_has_memfd_create(void)
...
 
+static bool kerndat_has_memfd_hugetlb(void)
+{
+	int ret;
+
+	ret = memfd_create(NULL, MFD_HUGETLB);
+	if (ret == -1 && errno == EFAULT)
+		kdat.has_memfd_hugetlb = true;
+	else if (ret == -1 && (errno == ENOSYS || errno == EINVAL))
+		kdat.has_memfd_hugetlb = false;
```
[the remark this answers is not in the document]
Snorch: 1) I think that @rppt does not mean to drop errno checking completely, please don't, we should be able to differentiate "flag is not valid" from other possible unexpected errors (ENOMEM and stuff).

2) I think that we can use kdat.has_memfd as a hint in kerndat_has_memfd_hugetlb to save some cpu cycles, we can skip calling second memfd_create with MFD_HUGETLB if we know it is not supported at all.

3) We have successfully created a memfd, please, don't forget to close it.


, please still check errno on error case it should be either ENOSYS, .

2) Ok, now memfd_create is successful

Snorch:
In dump_shm_hugetlb_flag you put MFD_HUGETLB into IpcShmEntry shm->hugetlb_flag = hugetlb_flag | SHM_HUGETLB, but in dump_memfd_inode you put into MemfdInodeEntry only mie.hugetlb_flag = flag. And in both cases you actually use MFD_HUGETLB, except that in first case it is saved to image and in second case it is appended to flags before creating memfd on restore (memfd_open_inode_nocache). Maybe we can make those places same/symmetric? (either both to images or both on restore...) Sorry if I miss something if it was done intentionaly.

minhbq-99:
> In dump_shm_hugetlb_flag you put MFD_HUGETLB into IpcShmEntry shm->hugetlb_flag = hugetlb_flag | SHM_HUGETLB, but in dump_memfd_inode you put into MemfdInodeEntry only mie.hugetlb_flag = flag. And in both cases you actually use MFD_HUGETLB, except that in first case it is saved to image and in second case it is appended to flags before creating memfd on restore (memfd_open_inode_nocache). Maybe we can make those places same/symmetric? (either both to images or both on restore...) Sorry if I miss something if it was done intentionaly.
> 
> Please also take a look on my other comment [#1622 (comment)](https://github.com/checkpoint-restore/criu/pull/1622#discussion_r786527483)

@Snorch thanks for your feedback. I decided to store MFD_HUGETLB into the image to be consistent with SHM_HUGETLB, also updated the kerndat check for memfd_hugetlb

I added a zdtm test for parent-child relationship process which is currently failed since mremap helper for hugetlb mapping is not implemented.

criu/kerndat.c
```diff
@@ -205,16 +197,33 @@ static int kerndat_get_shmemdev(void)
...
+				return -1;
 			}
 		} else {
 			pr_perror("Can't stat self map_files %d", e);
-			goto err;
+			return -1;
 		}
 	} else
-		dev = buf.st_dev;
+		*dev = buf.st_dev;
```
rst0git, line 209: @minhbq-99 could you please add braces for the else branch? In CRIU we try to follow the Kernel coding style:

criu/kerndat.c
```diff
@@ -1260,13 +1323,44 @@ static int kerndat_has_nftables_concat(void)
...
+	int ret, has_new_data = 0;
+
+	ret = kerndat_get_hugetlb_dev();
+	if (ret < 0)
+		return ret;
+	has_new_data |= ret;
+
+	/* New information is found, we need to save to the cache */
+	if (has_new_data)
+		kerndat_save_cache();
```
rst0git, line 1349: Why do we need the `has_new_data` variable? Could we do something like the following instead?

```C
int kerndat_try_load_new(void)
{
	int ret;

	ret = kerndat_get_hugetlb_dev();
	if (ret < 0)
		return ret;

	/* New information is found, we need to save to the cache */
	if (!ret)
		kerndat_save_cache();

	return 0;
}
```

criu/files.c
rst0git, line 552: @minhbq-99 the order here is important. `is_hugetlb_dev()` is called when `kdat.has_memfd_hugetlb` is `0`.

criu/proc_parse.c
```diff
@@ -616,10 +621,29 @@ static int handle_vma(pid_t pid, struct vma_area *vma_area, const char *file_pat
 			pr_info("path: %s\n", file_path);
 			vma_area->e->status |= VMA_AREA_SYSVIPC;
 		} else {
-			if (is_anon_shmem_map(st_buf->st_dev)) {
+			/* Dump shmem dev, hugetlb dev (private and share) mappings the same way as memfd
+			 * when possible.
+			 */
+			if (is_memfd(st_buf->st_dev) || is_anon_shmem_map(st_buf->st_dev) ||
+			    (is_hugetlb_dev(st_buf->st_dev, &hugetlb_flag) && kdat.has_memfd_hugetlb)) {
```
rst0git, line 628: Should it be `kdat.has_memfd_hugetlb && is_hugetlb_dev(st_buf->st_dev, &hugetlb_flag)`?

test/zdtm.py
```diff
@@ -2738,7 +2738,17 @@ def clean_stuff(opts):
...
+with open("/proc/sys/vm/nr_hugepages", "r") as f:
+    orig_hugepages = f.read()
+
+with open("/proc/sys/vm/nr_hugepages", "w") as f:
+    f.write("20\n")
+
 opts['action'](opts)
 
+with open("/proc/sys/vm/nr_hugepages", "w") as f:
+    f.write(orig_hugepages)
```
rst0git, line 2751: Would it be possible to move these into functions, for example, `set_nr_hugepages()` and `restore_nr_hugepages()`?

test/zdtm/static/maps10.c
```diff
@@ -0,0 +1,133 @@
...
+	test_waitsig();
+
+	kill(pid, SIGTERM);
+	wait(&status);
+	if (WIFEXITED(status)) {
+		if (WEXITSTATUS(status))
+			goto err;
+	} else
+		goto err;
+
```
rst0git, line 100: Could you please add braces for the `else` branch as well?

criu/kerndat.c
minhbq-99: At first, I want to add a kerndat check for mremap with hugetlb in this function too. However, as my machine kernel does not have that feature, I need to build a kernel for testing it so maybe it will be in later pull request.

criu/files.c
minhbq-99: Same as the below comment

criu/proc_parse.c
rst0git: @minhbq-99 here we have the following condition: `( is_hugetlb_dev(...) && kdat.has_memfd_hugetlb )`

This condition will be _true_ if both `is_hugetlb_dev()` AND `kdat.has_memfd_hugetlb` are true.
It would be _false_, if `kdat.has_memfd_hugetlb` is false.
If `kdat.has_memfd_hugetlb` is false, we don't need to call `is_hugetlb_dev()`.
Swapping `kdat.has_memfd_hugetlb` and `is_hugetlb_dev(...)` is just an optimization.

minhbq-99: Thanks, I get your point.

scripts/ci/run-ci-tests.sh
```diff
@@ -63,8 +63,9 @@ ci_prep () {
 
 test_stream() {
 	# Testing CRIU streaming to criu-image-streamer
+	STREAM_TEST_EXCLUDE="-x maps09 -x maps10"
```
rst0git, line 66: @minhbq-99 Would it be possible to add a comment (e.g., FIXME or TODO) that explains why we skip these tests?

https://github.com/checkpoint-restore/criu/pull/1622#issuecomment-1002355360

rst0git (approved):
LGTM

adrianreber:
Fedora test failures are fixed by: #1750

CentOS 8 based test failures are related to error injection and tracked here: #1734

Everything else looks good.

avagin:
Ok. Let's merge it and fix all other issues on top. @minhbq-99 thanks a lot!

PR #1624, opened 2021-10-06
3.16.1 release

adrianreber:
We (@rst0git and myself) would like to see a 3.16.1 release with the following two changes:

 * use a versioned python interpreter for `criu-ns`
 * support `--join-ns` via libcriu

The versioned python interpreter is a change which is necessary for our binary packages provided via OBS, COPR, Fedora and PPA. RPMs as well as DEBs are not happy with an unversioned python interpreter in `criu-ns`. We have workarounds but having the fix in 3.16.1 would be cleaner.

The main reason for this .1 release is, however, the `--join-ns` support in libcriu. To support checkpointing container out of pods and restoring into different pods we need to be able to join existing namespaces. This option is exported via CLI and RPC but not via libcriu. To add pod support to crun this release is needed as crun uses libcriu as an interface. It is easier for upstream crun to depend on a released version instead of the latest built from the `criu-dev` branch. As there are usually 8 to 10 months between CRIU releases we would like to see the `--join-ns` via libcriu released as soon as possible to be able to use crun.

CC: @checkpoint-restore/maintainers

We can also pull additional patches from `criu-dev` if it would be useful in a 3.16.1 release.

### DIFF
--- a/Documentation/Makefile
+++ b/Documentation/Makefile
@@ -12,7 +12,9 @@ endif
 
 FOOTER		:= footer.txt
 SRC1		+= crit.txt
+ifeq ($(PYTHON),python3)
 SRC1		+= criu-ns.txt
+endif
 SRC1		+= compel.txt
 SRC8		+= criu.txt
 SRC		:= $(SRC1) $(SRC8)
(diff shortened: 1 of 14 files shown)


adrianreber:
@xemul @rppt any comments from you side regarding a 3.16.1 release?

xemul:
LGTM

PR #1627, opened 2021-10-14
criu.pc: Add libprotobuf-c as a dependency

rst0git:
CRIU has a dependency on protobuf-c-devel. We express this dependency in pkgconfig to be auto-detected when building a package.


rst0git:
I'm closing this pull request because the protobuf-c package name is different in Ubuntu/Debian and Fedora. This issue can be fixed in the package itself.

Issue #232, opened 2016-10-19
A session leader of x(x) is outside of its pid namespace; criu_ns and criu restore by hand


Tobeez:
I want to dump and restore multiple times on same machine and on different machines as well. Using criu-ns, I am able to do a first restore successfully (processes running multiple threads). However, dumping the restored process using criu-ns (as suggested in https://www.criu.org/CR_in_namespace) gives the error `A session leader of x(x) is outside of its pid namespace`. I only get through if I comment out this part of the code in the cr-dump.c. This is probably unsafe to do. Is there a work around to get criu-ns dump and restore multiple times working or can I risk ignoring the `sid==0` check?

xemul:
Can you show the process' pid, pgid and sid before the first dump and after first restore?

Tobeez:
Before dump, I have
PID=10508;   PGI=10508 ;   SID=1768 
I dump using `sudo criu-ns dump -D criu_dir  --shell-job --leave-stopped -t 10508`
I restore: `sudo criu-ns restore -d -D criu_dir  --shell-job`

New process parameters: 
PID=3285;  PGID=3275;  SID=1768 

Attempt at second dump gives: `sudo criu-ns dump -D criu_dir  --shell-job --leave-stopped -t 3285`
Replace pid 3285 with 10508
Error (criu/cr-dump.c:1317): A session leader of 10508(10508) is outside of its pid namespace
Error (criu/cr-dump.c:1683): Dumping FAILED.

xemul:
Ah, I see. So you use the --shell-job to ignore the shared session even on first dump. That's known issue of criu_ns. You can workaround it if calling setsid before spawning your tasks for the first time.

And to proper handle the --shell-job in criu-ns we just need to teach criu-ns call setsid itself on restore :) I'll try to cook the patch for this ... someday :) but if you do it faster than me I won't mind :)

Tobeez:
`setsid` works perfectly when restoring "by hand". However, the first run of `sudo criu restore -d --shell-job -D criu_dir` gives the error (PID=20; PGID=20; SID=20):
Error (criu/cr-restore.c:1370): Pid 21 do not match expected 20
Error (criu/cr-restore.c:1131): 21 exited, status=1

A re-execution of the restore command works. Guess i'll have to execute twice, but is there a reason why this is happening?

xemul:
Because on criu restore (not criu-NS restore) you restore tree into current pid namespace, thus pids conflict may occur.

[...]

PR #1628, opened 2021-10-19
criu-ns: make pidns init first do setsid

Snorch:
We see that on criu-ns dump/restore/dump of the process which initially
was not a session leader (with --shell-job option) we see sid == 0 for
it and fail with something like:

Error (criu/cr-dump.c:1333): A session leader of 41585(41585) is outside of its pid namespace

Note: We should not dump processes with sid 0 (even with --shell-job) as
on restore we can can put such processes from multiple sessions into
one, which is wrong.

Fixes: #232

### DIFF
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -102,6 +102,7 @@ def wrap_restore():
 
     criu_pid = os.fork()
     if criu_pid == 0:
+        os.setsid()
         _mount_new_proc()
         run_criu(restore_args)
 


Snorch:
Strange fail, don't believe it's related to PR:
```
=== Run 1/1 ================ zdtm/static/fpu00
timens isn't supported on 5.4.0-1025-aws
========================== Run zdtm/static/fpu00 in h ==========================
Start test
./fpu00 --pidfile=fpu00.pid --outfile=fpu00.out
Run criu dump
Forcing 134 fault
=[log]=> dump/zdtm/static/fpu00/54/1/dump.log
b'(00.002978) Set up parasite blob using memfd'
b'(00.003005) Putting parasite blob into 0x7f9e5b21f000->0x7fa1d5310000'
b'(00.003048) Dumping general registers for 54 in native mode'
b'(00.003052) Dumping GP/FPU registers for 54'
b'(00.003061) Error (compel/arch/x86/src/lib/infect.c:298): Corrupting xsave for 54, seed 3509897626'
b'(00.003081) x86: magic1:0x46505853 extended_size:2700 xstate_bv:0x2f7 xstate_size:2696'
b'(00.003083) x86: xstate_bv: 0x2f7'
b'(00.003085) x86: -----------------------'
b'(00.003091) Putting tsock into pid 54'
b'(00.003123) Error (criu/parasite-syscall.c:88): si_code=4 si_pid=54 si_status=11'
b'(00.003126) Error (criu/parasite-syscall.c:95): 54 was stopped by 11 unexpectedly'
################### Test zdtm/static/fpu00 FAIL at CRIU dump ###################
Send the 9 signal to  54
Wait for zdtm/static/fpu00(54) to die for 0.100000
```

scripts/criu-ns
```diff
@@ -102,6 +102,7 @@ def wrap_restore():
 
     criu_pid = os.fork()
     if criu_pid == 0:
+        os.setsid()
```
rst0git, line 105: Hi @Snorch, this change is results in the following error:

```console
cat > test.sh <<-EOF
#!/bin/sh
while : ; do date; sleep 1; done
EOF
chmod +x test.sh
./test.sh
```

```console
mkdir /tmp/test
sudo criu dump -t `pgrep test.sh` -D /tmp/test -j -v4
sudo criu-ns restore -D /tmp/test -j -v4
```

Restore error:
```
(00.003168)  16390: tty: open driver pts id 0x7 index 5 (master 0 sid 16105 pgrp 16390 inherit 1)
(00.003172)  16564: 	Receive fd for 2
(00.003191)  16390: tty: Migrated slave peer 0x7 -> to fd 0
(00.003222)  16390: Error (criu/tty.c:689): tty: Failed to set group 16390 on 0: Inappropriate ioctl for device
(00.003229)  16390: Error (criu/files.c:1200): Unable to open fd=0 id=0x7
(00.003272) Error (criu/cr-restore.c:2470): Restoring FAILED.
```

If I comment out this line `criu-ns restore` works as expected.

adrianreber: Do you know why CI did not catch this?

rst0git: > Do you know why CI did not catch this?

Good question, I think we don't currently have CI tests for `criu-ns`.

adrianreber: That would explain it 😁

Snorch: Thanks for catching this, I will try looking closer on it, but not sure if there is good solution.

Difference of your example (from my test with setsid() added) is that your test.sh creates root_item with sid inherited but  with own pgid. Without setsid() on restore pgid would be 0 that was probably a reason why problem was hidden...

If I do:
```
@@ -1043,8 +1045,7 @@ static int pty_open_unpaired_slave(struct file_desc *d, struct tty_info *slave)
                 */
                if (root_item->sid != vpid(root_item)) {
                        if (root_item->pgid == vpid(root_item)) {
-                               if (tty_set_prgp(fd, root_item->pgid))
-                                       goto err;
+                               ;
                        } else {
                                pr_debug("Restore inherited group %d\n", getpgid(getppid()));
                                if (tty_set_prgp(fd, getpgid(getppid())))
```
It helps at least on first dump/restore.

But second dump gives:
```
(00.001391) Error (criu/mount.c:927): mnt: Mount 1623 ./var/lib/nfs/rpc_pipefs (master_id: 566 shared_id: 0) has unreachable sharing. Try --enable-external-masters.
```
Which is probably another unrelated problem of criu-ns script, external mounts get to our criu-ns container when we do unshare(CLONE_NEWNS) =)

I think that adding setsid() is not a bug, but it reveals that we have some problem here in some cases...

PR #1641, opened 2021-10-29
tests: improve the image streamer process control

nviennot:
When exceptions are raised during testing, the image streamer process should be terminated as opposed to being left hanging.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1039,6 +1039,7 @@ class criu:
         self.__lazy_pages_p = None
         self.__page_server_p = None
         self.__dump_process = None
+        self.__img_streamer_process = None
         self.__tls = self.__tls_options() if opts['tls'] else []
         self.__criu_bin = opts['criu_bin']
         self.__crit_bin = opts['crit_bin']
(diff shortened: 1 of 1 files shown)


nviennot:
cc @rst0git

test/zdtm.py
```diff
@@ -1287,7 +1299,15 @@ def spawn_criu_image_streamer(self, action):
                 raise test_fail_exc(
                     "criu-image-streamer is not starting (exit_code=%d)" % p.wait())
 
-        return p
+        progress.close()
+
+        self.__img_streamer_process = p
+
+    def wait_for_criu_image_streamer(self):
+        ret = self.__img_streamer_process.wait()
```
rst0git, line 1307: It might be good to set a `timeout` value for `wait()`.

rst0git (approved):
LGTM

test/zdtm.py
rst0git: Thanks, I will try to come up with a more general solution for this problem. Also, I noticed a few places where `wait()` is used with stdout/stderr set to `PIPE`, and it looks like this may cause a deadlock: https://docs.python.org/3.8/library/subprocess.html#subprocess.Popen.wait

nviennot: Regarding deadlocks, I fixed some of them here: https://github.com/checkpoint-restore/criu/pull/1642 but I didn't think this PR was worth looking at, but it is. Let me re-open it.

adrianreber:
Can you rebase. This looks correct, but I see failures which should be fixed on `criu-dev`.

nviennot:
Rebased

PR #1642, opened 2021-10-29
tests: improve the deterministic behavior of the test suite

nviennot:
Various I/O objects are unclosed when the object falls out of scope.
This can lead to non-deterministic behavior.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -202,6 +202,8 @@ class ns_flavor:
 
     def __copy_libs(self, binary):
         ldd = subprocess.Popen(["ldd", binary], stdout=subprocess.PIPE)
+        stdout, _ = ldd.communicate()
+
         xl = re.compile(
             r'^(linux-gate.so|linux-vdso(64)?.so|not a dynamic|.*\s*ldd\s)')
 
@@ -216,10 +218,8 @@ class ns_flavor:
                     map(
                         lambda x: str(x).strip(),
                         filter(lambda x: str(x).startswith('\t'),
-                               ldd.stdout.read().decode(
+                               stdout.decode(
                                    'ascii').splitlines())))))
-
-        ldd.wait()
 
         for lib in libs:
             if not os.access(lib, os.F_OK):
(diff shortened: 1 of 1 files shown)


rst0git:
I think we need to update `__get_start_cmd()` as well. `wait()` may result in a [deadlock](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait) when using `stdout=PIPE`.

test/zdtm.py
```diff
@@ -202,6 +202,8 @@ def __copy_one(self, fname):
 
     def __copy_libs(self, binary):
         ldd = subprocess.Popen(["ldd", binary], stdout=subprocess.PIPE)
+        stdout, stderr = ldd.communicate()
```
rst0git, line 205: `stderr` is not used, we could use `unused` or `_` instead
https://www.python.org/dev/peps/pep-0640/

test/zdtm.py
```diff
@@ -331,8 +331,7 @@ def decode_flav(i):
 
 def tail(path):
     p = subprocess.Popen(['tail', '-n1', path], stdout=subprocess.PIPE)
-    out = p.stdout.readline()
-    p.wait()
+    out, err = p.communicate()
```
rst0git, line 334: `err` is not used, we could use `unused` or `_` instead

test/zdtm.py
nviennot: Fixed

test/zdtm.py
nviennot: Fixed

nviennot:
Regarding `__get_start_cmd()`, I also changed it to use `communicate()`. I had to fix a bit of the code to make things work (invoking `list()`).

Note that there was no risk of deadlock as there's only one output that is redirected (the program was already reading stdout correctly). There's a risk of deadlock when both stdout and stderr are redirected. Reading stdout and then stderr sequentially can create a deadlock as the child process may want to write to both stdout and stderr, and one of the pipe can fill up. communicate() drain both at the same time.

It's just good practice to use communicate. Also it waits for the process at the same time. It's nice.

adrianreber:
Rebasing should remove most of the CI errors.

nviennot:
Rebased

rst0git (approved):
LGTM

adrianreber:
Another rebase is necessary. Looks reasonable.

avagin:
LGTM

nviennot:
Rebased

PR #1644, opened 2021-10-29
RCF: Clang format improve human readability with "comment" allignment

Snorch:
There are many places where we had intentionally short alligned lines for human readability, to be able to resurect those places we can use the hack with adding comments at the end of line where we don't want the break-line to collapse.

I want to understand if this solution is ok for us or not. Didn't find a better way with clang.

v2: switched to simple clang-format disabling instead of comments.


Snorch:
Please I need some suggestions on this, alternatively we can just disable clang in such places https://github.com/Snorch/criu/commit/692b21e5ae1187b7f51a0c0d9f58c45824d52866

Snorch:
@checkpoint-restore/maintainers

adrianreber:
Both solutions (comments or disabling clang-format for the section) work for me.

rppt:
> Please I need some suggestions on this, alternatively we can just disable clang in such places [Snorch@692b21e](https://github.com/Snorch/criu/commit/692b21e5ae1187b7f51a0c0d9f58c45824d52866)

I like this one more

rst0git:
> > Please I need some suggestions on this, alternatively we can just disable clang in such places [Snorch@692b21e](https://github.com/Snorch/criu/commit/692b21e5ae1187b7f51a0c0d9f58c45824d52866)
> 
> I like this one more

Me too.

rst0git (approved):

Snorch:
> > > Please I need some suggestions on this, alternatively we can just disable clang in such places [Snorch@692b21e](https://github.com/Snorch/criu/commit/692b21e5ae1187b7f51a0c0d9f58c45824d52866)
> > 
> > 
> > I like this one more
> 
> Me too.

Note that the way with comments has an advantage of still checking everything else clang-format related except wrapping on binary operator. But as we don't plan to change style much regions with clang format off should also be fine.

Issue #1633, opened 2021-10-25
criu-image-streamer CI error


adrianreber:
Hello @nviennot,

we are seeing criu-image-streamer errors in CI:

```
2021-10-22T15:08:57.4851233Z ===================== Run zdtm/static/mntns_ghost01 in ns ======================
2021-10-22T15:08:57.4852015Z Start test
2021-10-22T15:08:57.4852667Z Test is SUID
2021-10-22T15:08:57.4854019Z ./mntns_ghost01 --pidfile=mntns_ghost01.pid --outfile=mntns_ghost01.out --dirname=mntns_ghost01.test
2021-10-22T15:08:57.4855462Z Run criu-image-streamer in capture mode
2021-10-22T15:08:57.4856338Z Run criu dump
2021-10-22T15:08:57.4857367Z Run criu-image-streamer in extract mode
2021-10-22T15:08:57.4858573Z Run criu-image-streamer in serve mode
2021-10-22T15:08:57.4859440Z Run criu restore
2021-10-22T15:08:57.4860961Z zdtm/static/mntns_ghost01: link-remap files left: ['zdtm/static/link_remap.13']
2021-10-22T15:08:57.4862142Z ########### Test zdtm/static/mntns_ghost01 FAIL at link remaps left ############
2021-10-22T15:08:57.4862996Z Send the 9 signal to  99
2021-10-22T15:08:57.4863781Z Wait for zdtm/static/mntns_ghost01(99) to die for 0.100000
2021-10-22T15:08:57.4864402Z 
2021-10-22T15:08:57.4865383Z === Run 1
```

and ideas what might be going on?

See #1631

**Description**

**Steps to reproduce the issue:**
1.
2.
3.

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**CRIU logs and information:**

<details><summary>CRIU full dump/restore logs:</summary>
<p>

```
(paste your output here)
```

</p>
</details>

<details><summary>Output of `criu --version`:</summary>
<p>

```
(paste your output here)
```

</p>
</details>

<details><summary>Output of `criu check --all`:</summary>
<p>

```
(paste your output here)
```

</p>
</details>

**Additional environment details:**

Snorch:
Remap is probably from other test, I see
```
 ===================== Run zdtm/static/unlink_fstat03 in ns =====================
Start test
[...]

PR #1645, opened 2021-11-01
Fix ghost files and improve error handling

Snorch:
Port patches from Virtuozzo criu:

files-reg: temporary remount writable the mount we do unlink on
ghost/mount: allocate remounted_rw in shmem to get info from other processes
files-reg: fix error handling of rm_parent_dirs
files-reg: fix error handling in open_path

Should fix the problem related to https://github.com/checkpoint-restore/criu/pull/1643

Note: test mntns_ghost01 detects the problem fixed by 4-th patch when error handling  (1-2 patches) is fixed.

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -2133,6 +2133,11 @@ ext:
 
 	if (rfi->remap) {
 		if (!rfi->remap->is_dir) {
+			struct mount_info *mi = lookup_mnt_id(rfi->rfe->mnt_id);
+
+			if (mi && try_remount_writable(mi, true))
+				goto err;
+
 			pr_debug("Unlink: %d:%s\n", rfi->rfe->mnt_id, rfi->path);
 			if (unlinkat(mntns_root, rfi->path, 0)) {
 				pr_perror("Failed to unlink the remap file");
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -1867,6 +1867,9 @@ out:
  * This routine properly resolves d's path handling ghost/link-remaps.
  * The open_cb is a routine that does actual open, it differs for
  * files, directories, fifos, etc.
+ *
+ * Return 0 on success, -1 on error and 1 to indicate soft error, which can be
+ * retried.
  */
 
 static int rfi_remap(struct reg_file_info *rfi, int *level)
@@ -1932,6 +1935,8 @@ out_root:
 		int errno_saved = errno;
 		rm_parent_dirs(mntns_root, path, *level);
 		errno = errno_saved;
+		if (errno == EEXIST)
+			return 1;
 		return -1;
 	}
 
@@ -2008,11 +2013,12 @@ static bool validate_file(const int fd, const struct stat *fd_status, const stru
 
 int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_file_info *, void *), void *arg)
 {
-	int tmp, mntns_root, level = 0;
+	int tmp = -1, mntns_root, level = 0;
 	struct reg_file_info *rfi;
 	char *orig_path = NULL;
 	char path[PATH_MAX];
 	int inh_fd = -1;
+	int ret;
 
 	if (inherited_fd(d, &tmp))
 		return tmp;
@@ -2049,13 +2055,8 @@ int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_fil
 			 */
 			orig_path = rfi->path;
 			rfi->path = rfi->remap->rpath;
-		} else if (rfi_remap(rfi, &level) < 0) {
+		} else if ((ret = rfi_remap(rfi, &level)) == 1) {
 			static char tmp_path[PATH_MAX];
-
-			if (errno != EEXIST) {
-				pr_perror("Can't link %s -> %s", rfi->remap->rpath, rfi->path);
-				return -1;
-			}
 
 			/*
 			 * The file whose name we're trying to create
@@ -2070,12 +2071,15 @@ int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_fil
 			orig_path = rfi->path;
 			rfi->path = tmp_path;
 			snprintf(tmp_path, sizeof(tmp_path), "%s.cr_link", orig_path);
-			pr_debug("Fake %s -> %s link\n", rfi->path, rfi->remap->rpath);
+			pr_debug("Fake %s -> %s link\n", rfi->remap->rpath, rfi->path);
 
-			if (rfi_remap(rfi, &level) < 0) {
+			if (rfi_remap(rfi, &level)) {
 				pr_perror("Can't create even fake link!");
-				return -1;
+				goto err;
 			}
+		} else if (ret < 0) {
+			pr_perror("Can't link %s -> %s", rfi->remap->rpath, rfi->path);
+			goto err;
 		}
 	}
 
@@ -2085,7 +2089,7 @@ ext:
 	if (tmp < 0) {
 		pr_perror("Can't open file %s", rfi->path);
 		close_safe(&inh_fd);
-		return -1;
+		goto err;
 	}
 	close_safe(&inh_fd);
 
@@ -2094,15 +2098,15 @@ ext:
 
 		if (fstat(tmp, &st) < 0) {
 			pr_perror("Can't fstat opened file");
-			return -1;
+			goto err;
 		}
 
 		if (!validate_file(tmp, &st, rfi))
-			return -1;
+			goto err;
 
 		if (rfi->rfe->has_mode && (st.st_mode != rfi->rfe->mode)) {
 			pr_err("File %s has bad mode 0%o (expect 0%o)\n", rfi->path, (int)st.st_mode, rfi->rfe->mode);
-			return -1;
+			goto err;
 		}
 
 		/*
@@ -2115,7 +2119,11 @@ ext:
 
 	if (rfi->remap) {
 		if (!rfi->remap->is_dir) {
-			unlinkat(mntns_root, rfi->path, 0);
+			pr_debug("Unlink: %d:%s\n", rfi->rfe->mnt_id, rfi->path);
+			if (unlinkat(mntns_root, rfi->path, 0)) {
+				pr_perror("Failed to unlink the remap file");
+				goto err;
+			}
 			rm_parent_dirs(mntns_root, rfi->path, level);
 		}
 
@@ -2124,10 +2132,17 @@ ext:
 	if (orig_path)
 		rfi->path = orig_path;
 
-	if (restore_fown(tmp, rfi->rfe->fown))
+	if (restore_fown(tmp, rfi->rfe->fown)) {
+		close(tmp);
 		return -1;
+	}
 
 	return tmp;
+err:
+	if (rfi->remap)
+		mutex_unlock(remap_open_lock);
+	close_safe(&tmp);
+	return -1;
 }
 
 int do_open_reg_noseek_flags(int ns_root_fd, struct reg_file_info *rfi, void *arg)
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -1792,30 +1792,42 @@ out:
 	return ret;
 }
 
-static void rm_parent_dirs(int mntns_root, char *path, int count)
+static int rm_parent_dirs(int mntns_root, char *path, int count)
 {
 	char *p, *prev = NULL;
+	int ret = -1;
 
-	if (!count)
-		return;
-
-	while (count > 0) {
-		count -= 1;
+	while (count-- > 0) {
 		p = strrchr(path, '/');
-		if (p)
+		if (p) {
+			/* We don't handle "//" in path */
+			BUG_ON(prev && (prev - p == 1));
 			*p = '\0';
+		} else {
+			/* Inconsistent path and count */
+			pr_perror("Can't strrchr \"/\" in \"%s\"/\"%s\"]"
+				  " left count=%d\n",
+				  path, prev ? prev + 1 : "", count + 1);
+			goto err;
+		}
+
 		if (prev)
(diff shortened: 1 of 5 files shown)


criu/files-reg.c
```diff
@@ -1792,30 +1792,49 @@ static int linkat_hard(int odir, char *opath, int ndir, char *npath, uid_t uid,
 	return ret;
 }
 
-static void rm_parent_dirs(int mntns_root, char *path, int count)
+/*
+ * FIXME these function changes path string in place, if this string is used
+ * simultaneousely in multiple processes we can have a race. Do we need strdup?
```
nviennot, line 1797: Yes, it's better to be safe

criu/files-reg.c
```diff
@@ -2115,19 +2140,36 @@ int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_fil
...
-			unlinkat(mntns_root, rfi->path, 0);
-			rm_parent_dirs(mntns_root, rfi->path, level);
+			struct mount_info *mi = lookup_mnt_id(rfi->rfe->mnt_id);
+
+			if (mi && try_remount_writable(mi, true))
+				goto err;
+
+			pr_debug("%d: Unlink: %s\n", rfi->rfe->mnt_id, rfi->path);
+			if (unlinkat(mntns_root, rfi->path, 0)) {
+				pr_perror("Failed to unlink the remap file");
```
nviennot, line 2150: `pr_perror("Failed to unlink the remap file %d:%s", mntns_root, rfi->path);`

criu/mount.c
```diff
@@ -2734,6 +2736,13 @@ struct mount_info *mnt_entry_alloc()
 
 	new = xzalloc(sizeof(struct mount_info));
 	if (new) {
+		if (rst) {
+			new->remounted_rw = shmalloc(sizeof(int));
```
nviennot, line 2740: Shouldn't we free `remounted_rw` in `mnt_entry_free()` ?

criu/files-reg.c
Snorch: - mntns_root is a ROOT_FD_OFF service fd number, I believe printing it has not much value
- the pr_debug above already prints mnt_id+path so we can skip it in pr_perror

criu/mount.c
Snorch: No we can not, with shmalloc we can only free last allocation (see shfree_last) so in general we can't free shmalloc'ed data.

criu/files-reg.c
Snorch: Something like https://github.com/Snorch/criu/commit/98f0445f884e3b2317d9efc74a77fd0a83bf48c9 can be done, but I'm not sure that we urgently need this because we've never seen such a race and all uses of path seem local not shared. But for future we should have this in mind.

nviennot: Even if it's not racy, i would do it.

It's very surprising that a function that unlinks things actually modifies an argument. It makes the code more difficult to reason about, and thus more difficult to maintain.

Correctness is more important that saving a few cpu cycles

criu/files-reg.c
nviennot: Log messages can get interleaved due to parallelism. It could br useful to have a more precise error message.

criu/mount.c
nviennot: Oof okay I see. I thought from the name that "last" meant that the user should do reference counting or something.
shfree_last() shouldn't take an argument. I think I wouldn't have been confused if it was just `shfree_last()`

criu/files-reg.c
Snorch: It's very common practice in criu that not only the error itself should be considered when debugging but also previous log messages. I get used to this practice probably too much =) So I would've just grep logs by pid to neutralize interleaved messages. But OK I can just make error message printing similar thing to pr_debug.

criu/files-reg.c
Snorch: > Correctness is more important that saving a few cpu cycles

Excess correctness checks are not always welcome =) One example that I can remember is --check-mounts patch (https://github.com/checkpoint-restore/criu/pull/660#issuecomment-548736668).

But anyway I tried to fix this comment. PR become a bit more complex. I'm OK to revert it to the original form in case @checkpoint-restore/maintainers are against this "strdup and staff" change.

Snorch:
Oops, stupid compilation errors fixed.

Snorch:
```
 b'(00.034649)      1: uffd: Lazy pages are not available: Operation not permitted'
b'(00.034651)   7809: uffd: Sending PID -7809'
b'(00.034654)      1: Error (criu/uffd.c:314): uffd: Unable to open an userfaultfd descriptor: Operation not permitted'
```
Don't think these errors are related to the patch...

Snorch:
I've reverted debugging of https://github.com/checkpoint-restore/criu/issues/1633 as I've found the root cause.

criu/files-reg.c
nviennot: > Excess correctness checks are not always welcome
I agree. I don't think using defensive programming is useful in most cases.

Here, we are calling `rm_parent_dirs(mntns_root, rfi->path, level)`. I'm willing to bet that most programmers will say that It's unexpected that this call modifies the rfi->path string for two reasons 1) semantically, there's no reason that unlink()/rm_dirs() modifies its arguments 2) we are modifying the content of a struct. There's no reason to think that this string is temporary or won't be reused.

This can lead to future bugs. I stand strong on this stance.

For implementation details, it can be easier to reason about it by putting the dup logic outside of the function:

```c
int __rm_parent_dirs(..., char *path, ...) {
    ...
}

int rm_parent_dirs(..., const char *path, ...) {
    char *dup_path = strdup(path);
    if (!dup_path)
        return -1;
    int ret = __rm_parent_dirs(..., dup_path, ...);
    free(dup_path);
    return ret;
}
```

criu/files-reg.c
nviennot: Fair, I don't feel strongly about this one, your call

Snorch: So let's revert it to my variant then.

criu/files-reg.c
Snorch: > Here, we are calling rm_parent_dirs(mntns_root, rfi->path, level). I'm willing to bet that most programmers will say that It's unexpected that this call modifies the rfi->path string for two reasons 1) semantically, there's no reason that unlink()/rm_dirs() modifies its arguments 2) we are modifying the content of a struct. There's no reason to think that this string is temporary or won't be reused.

All modifications of rm_parent_dirs are invisible to the caller, it only temporary inserts '\0' in place of '\' and reverts everything back later.

> This can lead to future bugs. I stand strong on this stance.

That's why I initially wanted to add FIXME comment, so that it would be easier to track such bugs in case we would have them in future. And now I tend to think that fixing rm_parent_dirs argument modification can be a matter of another PR. This pr goal is to fix remount writable code, and fix error handling so that the problem with remount writable code would be visible (test mntn_ghost01 starts to fail).

> For implementation details, it can be easier to reason about it by putting the dup logic outside of the function:

The Idea looks good, but as we need to intrusively change rm_parent_dirs making it not replace '\0' back to '\' as this code becomes excess, so I don't think we need __rm_parent_dirs ...

Let's be lazy an leave strdup like I did it, and wait for some more review.

nviennot: > All modifications of rm_parent_dirs are invisible to the caller, it only temporary inserts '\0' in place of '\' and reverts everything back later.

Ohh okay, well in that case, it's fine. Nevermind.

Snorch:
No comments on it for a long time...
I consider xstrdup and inplace_dirname changes too heavy to bee easily reviewed, and also not actually needed because it does not fix actual errors =)

So let's revert everything to the original state, other bonus from this is that I would have less pain porting other mount-v2 patches on top of it. Also let's remove FIXME completely as it prove itself confusing.

All current review comments are resolved, would like to have more review, to be able to finally merge this.

rst0git:
>I believe it's a bug in centos8 kernel that IPPROTO_MPTCP_KERN is visible to userspace, in centos9 they don't have this problem.

`IPPROTO_MPTCP_KERN` has been added in RHEL with this commit:
```
[net] add and use MPTCP_PROTO_KERN


Message-id: <20200515101824.19445-21-fwestpha@redhat.com>
Patchwork-id: 309886
Patchwork-instance: patchwork
O-Subject: [RHEL8.3 net 20/38] net: add and use MPTCP_PROTO_KERN
Bugzilla: 1835796
RH-Acked-by: Paolo Abeni's avatarPaolo Abeni <pabeni@redhat.com>
RH-Acked-by: Marcelo Ricardo Leitner's avatarMarcelo Leitner <mleitner@redhat.com>
RH-Acked-by: Jarod Wilson's avatarJarod Wilson <jarod@redhat.com>
RH-Acked-by: Tony Camuso's avatarTony Camuso <tcamuso@redhat.com>

Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1835796


Upstream Status: RHEL ONLY

Upstream kernel simply increased IPPROTO_MAX, which we can't
do due to KABI implications (the define is used e.g. for array sizing).

This patch adds a kernel-internal MPTCP_PROTO_KERN and
maps UAPI IPPROTO_MPTCP to it, i.e. if userspace asks for
protocol 262 (IPPROTO_MPTCP), the kernel changes that to "66" under
the hood so we maintain UAPI with upstream kernel.
```

adrianreber:
> To @adrianreber : can we just disable socket-raw test on centos 8?

Works for me.

Snorch:
> > To @adrianreber : can we just disable socket-raw test on centos 8?
> 
> Works for me.

I'm not too experienced with our ci, so advice on how to properly pass -x socket-raw would be appreciated.
https://github.com/checkpoint-restore/criu/pull/1658

nviennot (approved):
Great! Thank you @Snorch !

criu/files-reg.c
```diff
@@ -2008,11 +2013,12 @@ static bool validate_file(const int fd, const struct stat *fd_status, const stru
 
 int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_file_info *, void *), void *arg)
 {
-	int tmp, mntns_root, level = 0;
+	int tmp = -1, mntns_root, level = 0;
 	struct reg_file_info *rfi;
 	char *orig_path = NULL;
 	char path[PATH_MAX];
 	int inh_fd = -1;
+	int ret;
```
rst0git, line 2038: The usage of `tmp` and `ret` is slightly confusing here because we have `return tmp;` and `ret` is used as temporary variable.

criu/files-reg.c
```diff
@@ -1792,30 +1792,45 @@ static int linkat_hard(int odir, char *opath, int ndir, char *npath, uid_t uid,
 	return ret;
 }
 
-static void rm_parent_dirs(int mntns_root, char *path, int count)
+static int rm_parent_dirs(int mntns_root, char *path, int count)
 {
 	char *p, *prev = NULL;
+	int ret = -1;
 
 	if (!count)
```
nviennot, line 1800: This if statement is not useful (the while loop condition is sufficient), but I don't think it matters.

criu/files-reg.c
Snorch: I believe that a better name for "tmp" is "fd" or "tmp_fd", also I think "ret" is in a proper place here, though it may be a bit confusing.

So I don't want to touch it as:
- "tmp" was named "tmp" for a long time
- patch goal is a bit unrelated - it is to fix error handling so that hidden errors became finally visible, not to fix variable naming
- Using "ret" as a temporary variable name to check called function ret code is a common practice, ret is not always a return value and it's ok (more over we tend to use "exit_code" name for return values in multiple places)

rst0git: I agree with you but what do you think about renaming "tmp" in a separate patch to improve readability?

criu/files-reg.c
Snorch: Thanks! Let's remove it.

adrianreber (approved):
Sounds good. CI is happy.

rst0git (approved):

criu/files-reg.c
```diff
@@ -2133,6 +2133,11 @@ int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_fil
 
 	if (rfi->remap) {
 		if (!rfi->remap->is_dir) {
+			struct mount_info *mi = lookup_mnt_id(rfi->rfe->mnt_id);
+
+			if (mi && try_remount_writable(mi, true))
```
avagin, line 2138: pls check in the test that a test mount is read-only after restore.

criu/files-reg.c
```diff
@@ -2124,10 +2132,17 @@ int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_fil
...
 		rfi->path = orig_path;
 
-	if (restore_fown(tmp, rfi->rfe->fown))
+	if (restore_fown(tmp, rfi->rfe->fown)) {
+		close(tmp);
 		return -1;
+	}
 
 	return tmp;
+err:
```
avagin, line 2161: nit: I would rename it to err_with_unlock or something like this to make it obvious that it unlocks remap_open_lock

criu/files-reg.c
Snorch: done:
zdtm: add ro-mount check after c/r to mntns_ghost01

criu/files-reg.c
Snorch: Please don't:

1) only unlocks for remap case
2) extra 9 lines change which does not directly related to commit purpose

Snorch:
Fails for Vagrant Fedora based test (no VDSO) check same as here:
https://github.com/checkpoint-restore/criu/pull/1664

So looks unrelated to this PR.

adrianreber:
> Fails for Vagrant Fedora based test (no VDSO) check same as here: #1664
> 
> So looks unrelated to this PR.

Yes, see #1669

PR #1646, opened 2021-11-01
cr-dump: do not dump zombie process with sid 0

liusdu:
A zombie process with 0 sid has a session leader in
outer pidns and has ignored SIGHUP. So there is no
process in this pid ns to wait on it. No need to
dump it.

### DIFF
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -1129,6 +1129,13 @@ static int dump_zombies(void)
 		item->pgid = pps_buf.pgid;
 
 		BUG_ON(!list_empty(&item->children));
+
+		if (!item->sid) {
+			pr_err("A session leader of zombie process %d(%d) is outside of its pid namespace\n",
+			       item->pid->real, vpid(item));
+			goto err;
+		}
+
 		if (dump_one_zombie(item, &pps_buf) < 0)
 			goto err;
 	}


liusdu:
related issue: https://github.com/checkpoint-restore/criu/issues/1647

Snorch:
Hi @liusdu , Thanks for your patch and nice catch of the problem!

Criu can not just skip dumping zombies if they have bad ids (sid from external pidns), we should fail instead. Note I already explained why sid=0 is bad here https://github.com/checkpoint-restore/criu/issues/232#issuecomment-946478237 .

So can you please rework your patch by instead adding similar to this https://github.com/checkpoint-restore/criu/blob/b405b1b0142bea0eeec1437a867b1c0ef7a3afe1/criu/cr-dump.c#L1343 check to dump_zombies after item->sid is set here https://github.com/checkpoint-restore/criu/blob/b405b1b0142bea0eeec1437a867b1c0ef7a3afe1/criu/cr-dump.c#L1129/

Also can you please check item->pgid is not zero too, it looks reasonable while we are looking at this code.

liusdu:
> Hi @liusdu , Thanks for your patch and nice catch of the problem!
> 
> Criu can not just skip dumping zombies if they have bad ids (sid from external pidns), we should fail instead. Note I already explained why sid=0 is bad here [#232 (comment)](https://github.com/checkpoint-restore/criu/issues/232#issuecomment-946478237) . We just forgot to add same check for zombies...
> 
> So can you please rework your patch by instead adding similar to this
> 
> https://github.com/checkpoint-restore/criu/blob/b405b1b0142bea0eeec1437a867b1c0ef7a3afe1/criu/cr-dump.c#L1343
> 
> check to dump_zombies after item->sid is set here https://github.com/checkpoint-restore/criu/blob/b405b1b0142bea0eeec1437a867b1c0ef7a3afe1/criu/cr-dump.c#L1129/
> upd: there is no need to check pgid, sid check is enough.

Hi @Snorch  thanks for your relay.  I have two reasons to ignore zombies with sid=0 but not fail this process.

1. This situations are quite frequent for container migration. 
 - We enter a container's  pid ns using   `nsenter` and  run a backgroud jobs.
 - Most container are application containers, whose init can not reap zombies..
 - When background jobs exits, they become zombies.
2. This is no harm to ignore zombies with sid=0.(correct me if I am wrong..)
截屏2021-11-09 上午10 37 00

rst0git:
I also agree with @Snorch that we can not just skip dumping zombies, because we need to ensure that a container checkpoint is consistent with the state of a container. However, we could add an option (e.g., `--skip-zombies`)  that would allow to ignore zombies with sid=0. This is similar to the approach we use with `--skip-in-flight`.

@liusdu what do you think?

Snorch:
I believe you can just `nsenter -p -t <pid> setsid <your command>` to make zombies with internal sids (no 0-es) and everything would just work for you. I don't think --skip-zombies is needed.

Snorch (approved):

PR #1648, opened 2021-11-02
Replace CONFIG_ARMV* with CONFIG_CPU_V*

rst0git:
This pull request is a rebase of https://github.com/checkpoint-restore/criu/pull/1071.

We have seen this problem reported in https://bugs.debian.org/918229 as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,7 @@ ifeq ($(ARCH),arm)
                 ARMV := 7
         endif
 
-        DEFINES		:= -DCONFIG_ARMV$(ARMV) -DCONFIG_VDSO_32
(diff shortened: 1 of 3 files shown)


adrianreber:
I am not convinced about the change from `CONFIG_ARMV` to `CONFIG_CPU_V`. This might make sense for someone using ARM every day, but for CRIU it seems to be easier to read if the define contains `ARM`. I am not able to immediately make the connection from `CONFIG_CPU_V` to ARM. It seems to be just a rename of a define, right?

The second commit sounds like a good idea. Is the second commit enough to fix the debian bug report?

rst0git:
> Is the second commit enough to fix the debian bug report?

I was not able to replicate the error described in the bug report.
It looks like this error occurs when `SUBARCH` is not specified when cross-compiling for 32-bit armel/armhf on arm64.

PR #1650, opened 2021-11-03
zdtm.py: make tests with --link_remap exclusive

Snorch:
We see that tests mntns_ghost01 and unlink_fstat03 can run
simultaneousely and thus the former sees leftover link_remap.* files in
the test directory created by the latter, and the latter is still
running so it's ok to have link_remap.* at this point.

Let's implicitly make all --link-remap tests exclusive (not running in
parallel).

Fixes: #1633

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1997,7 +1997,22 @@ class Launcher:
             raise Exception("The kernel is tainted: %r (%r)" %
                             (taint, self.__taint))
 
-        if test_flag(desc, 'excl'):
+        '''
+        The option --link-remap allows criu to hardlink open files back to the
+        file-system on dump (should be removed on restore) and we have a sanity
+        check in check_visible_state that they were actually removed at least
+        from the root test directory after restore.
+
+        As zdtm runs all tests from the same cwd (e.g.: test/zdtm/static) in
+        parallel, hardlinks from one test can mess up with sanity checks of
+        another test or even one test can by mistake use hardlinks created by
+        another test which is even worse.
+
+        So let's make all tests using --link-remap option non parallel.
+        '''
+        link_remap_excl = '--link-remap' in desc.get('opts', '').split() + desc.get('dopts', '').split() + desc.get('ropts', '').split()
+
+        if test_flag(desc, 'excl') or link_remap_excl:
             self.wait_all()
 
         self.__nr += 1
@@ -2030,7 +2045,7 @@ class Launcher:
             "start": time.time()
         }
 
-        if test_flag(desc, 'excl'):
+        if test_flag(desc, 'excl') or link_remap_excl:
             self.wait()
 
     def __wait_one(self, flags):


nviennot:
While this works, are you convinced that this is the proper way to fix the problem?
For example, could we use a per-test directory for remapped files instead of the global folder `test/zdtm/static/` ?

adrianreber:
Rebasing should solve most CI errors.

Snorch:
> Thank you for looking into this

Sure, no problem

> While this works, are you convinced that this is the proper way to fix the problem? For example, could we use a per-test directory for remapped files instead of the global folder `test/zdtm/static/` ?

Fact 1: all criu tests run from the same directory (e.g. test/zdtm/static)
```
[root@fedora criu]# pstree -Sp 236792
python(236792)───zdtm_ct(236818,ipc,mnt,net)───python(236819,pid)─┬─mntns_ghost01(236880,ipc,mnt,net,pid,+
                                                                  └─python(236821)
[root@fedora criu]# ls /proc/236880/cwd/mntns_ghost01
/proc/236880/cwd/mntns_ghost01
```
This way it is just easier to organize tests. I don't believe that we would want to change it.

Fact 2: link remap files are created by criu on dump in the closest existing ancestor directory to the original file path. https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/files-reg.c#L1010 And I don't believe we would change it either.

So due to 1+2 in the worst case we get link_remap.* file created in the test directory. And it's expected. And zdtm sanity check fails on it.

Let's just fix the "unexpected" thing that link_remap.* files from first test intervene with sanity checks of the second test by disabling parallelism for them.

Alternative solution could be manually adding "excl" flag to description of each test with --link-remap instead of this semi-automatic way, but it is more error prone because we can forget to add this when adding new tests.

nviennot:
Okay, LGTM

test/zdtm.py
```diff
@@ -1997,7 +1997,8 @@ def run_test(self, name, desc, flavor):
             raise Exception("The kernel is tainted: %r (%r)" %
                             (taint, self.__taint))
 
-        if test_flag(desc, 'excl'):
+        link_remap_excl = '--link-remap' in desc.get('opts', '').split() + desc.get('dopts', '').split()
```
avagin, line 2000: pls write the comment why they are exclusive

test/zdtm.py
```diff
@@ -2030,7 +2031,7 @@ def run_test(self, name, desc, flavor):
             "start": time.time()
         }
 
-        if test_flag(desc, 'excl'):
+        if test_flag(desc, 'excl') or link_remap_excl:
```
avagin, line 2034: If such tests will create a temporary directories and create their files there, will it solve the problem?

Snorch: Yes.

test/zdtm.py
Snorch: done.

Snorch:
v2:
- add explaining comment why --link-remap tests should be exclusive.
- also look --link-remap in ropts

PR #1651, opened 2021-11-04
ci: fix userfaultfd and clang-format test failures

adrianreber:
Newer kernels (5.11) require echo 1 > /proc/sys/vm/unprivileged_userfaultfd

Without the 'echo 1' the kernel prints a message like this:

 uffd: Set unprivileged_userfaultfd sysctl knob to 1 if kernel faults must be handled without obtaining CAP_SYS_PTRACE capability

I also pinned the clang-format to Fedora 34 which comes with clang 12. Fedora 35 has clang 13 which formats the code differently.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,7 +6,7 @@ jobs:
   build:
     runs-on: ubuntu-latest
     container:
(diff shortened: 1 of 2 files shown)


rppt (approved):
Ack

PR #1657, opened 2021-11-18
seize: restore cgroup freezer to right state

liusdu:
Now if criu dump failed before freezing dumpee, criu always freeze
dumpee in cr_dump_finish.

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -131,11 +131,11 @@ static enum freezer_state get_freezer_state(int fd)
 	return get_freezer_v1_state(fd);
 }
 
-static bool freezer_thawed;
+static enum freezer_state origin_freezer_state = FREEZER_ERROR;
 
 const char *get_real_freezer_state(void)
 {
-	return freezer_thawed ? thawed : frozen;
+	return origin_freezer_state == THAWED ? thawed : frozen;
 }
 
 static int freezer_write_state(int fd, enum freezer_state new_state)
@@ -192,7 +192,7 @@ static int freezer_restore_state(void)
 	int fd;
 	int ret;
 
-	if (!opts.freeze_cgroup || freezer_thawed)
+	if (!opts.freeze_cgroup || origin_freezer_state != FROZEN)
 		return 0;
 
 	fd = freezer_open();
@@ -481,9 +481,10 @@ static int freeze_processes(void)
 		close(fd);
 		return -1;
 	}
-	if (state == THAWED) {
-		freezer_thawed = true;
 
+	origin_freezer_state = state == FREEZING ? FROZEN : state;
+
+	if (state == THAWED) {
 		if (freezer_write_state(fd, FROZEN)) {
 			close(fd);
 			return -1;
@@ -534,7 +535,7 @@ static int freeze_processes(void)
 	}
 
 err:
-	if (exit_code == 0 || freezer_thawed)
+	if (exit_code == 0 || origin_freezer_state == THAWED)
 		exit_code = freezer_write_state(fd, THAWED);
 
 	if (close(fd)) {


Snorch:
I believe that this patch introduces code change which does not change semantics. -> Nack.

```
        state = get_freezer_state(fd);
# here state can be one of FREEZER_ERROR, THAWED, FROZEN, FREEZING
        if (state == FREEZER_ERROR) {
                close(fd);
                return -1;
        }
# here state can be one of THAWED, FROZEN, FREEZING
-       if (state == THAWED) {
-               freezer_thawed = true;
# so (freezer_thawed == true) is equivalent to (state == THAWED) or ((state != FROZEN) && (state != FREEZING))
 
+       freezer_state = state == FREEZING ? FROZEN : state;
# and (freezer_state == THAWED) is equivalent to (state == THAWED)
# and (freezer_state != FROZEN) is equivalent to ((state != FROZEN) && (state != FREEZING))
+
+       if (state == THAWED) {
```

Meaning that (freezer_thawed == true) and (freezer_state == THAWED) and (freezer_state != FROZEN) all are equivalents.

Snorch (approved):
> I believe that this patch introduces code change which does not change semantics. -> Nack.
> 
> ```
>         state = get_freezer_state(fd);
> # here state can be one of FREEZER_ERROR, THAWED, FROZEN, FREEZING
>         if (state == FREEZER_ERROR) {
>                 close(fd);
>                 return -1;
>         }
> # here state can be one of THAWED, FROZEN, FREEZING
> -       if (state == THAWED) {
> -               freezer_thawed = true;
> # so (freezer_thawed == true) is equivalent to (state == THAWED) or ((state != FROZEN) && (state != FREEZING))
>  
> +       freezer_state = state == FREEZING ? FROZEN : state;
> # and (freezer_state == THAWED) is equivalent to (state == THAWED)
> # and (freezer_state != FROZEN) is equivalent to ((state != FROZEN) && (state != FREEZING))
> +
> +       if (state == THAWED) {
> ```
> 
> Meaning that (freezer_thawed == true) and (freezer_state == THAWED) and (freezer_state != FROZEN) all are equivalents.

Ah, big sorry, looks I'm wrong, the initial value is what you've actually changed. Now it looks reasonable for me.

liusdu:
@Snorch yes, The actual change of this commit is the  initial value and it is enough for the bug I met.  Some functions using this variable maybe  need to be changed, but I have no idea about the influence.. Should I dig it up？

Snorch:
> @Snorch yes, The actual change of this commit is the initial value and it is enough for the bug I met.

Please add a note about it in the commit message. It would be useful to know not only what problem you solve but also how you solve it https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes.
```
Once the problem is established, describe what you are actually doing
about it in technical detail.
```

> Some functions using this variable maybe need to be changed, but I have no idea about the influence.. Should I dig it up？

AFAICS you did everything right, no change needed, that's why I approved the PR.

avagin:
LGTM

Snorch:
@liusdu Please change description to something like
```
The new freezer_state is a complete equivalent of old freezer_thawed
except for the initial value. If old freezer_thawed was not initialized
it was 0 and in freezer_restore_state were threated as if we need to
freeze cgroup "back", thus before this patch if criu dump failed before
freezing dumpee, criu always freeze dumpee in cr_dump_finish which is
wrong. Switching to freezer_state initialized with FREEZER_ERROR fixes
the problem.
```

And also please rename freezer_state to origin_freezer_state, because having same variable name with enum name makes grepping it harder.

Snorch:
I've updated the PR myself, will merge when tests would pass.

Snorch (approved):

liusdu:
@Snorch  Thanks～

PR #1659, opened 2021-11-19
check: cleanup child processes

nviennot:
Always wait() for forked child processes. It avoid zombie processes in containers that don't have an init process reaping orphans.


rst0git (approved):
LGTM

adrianreber (approved):
Makes sense.

Snorch (approved):

PR #1660, opened 2021-11-23
clang-format/zdtm: fix clang complains about strange elseifs

Snorch:
Clang-format v13 on my Fedora 35 complains about these hunks, more over
reading the formating we had before is a pain:

```
} else /* comment */
	if (smth) {
	fail("")
	return -1;
}
```

Let's make explicit {} braces for else, this way it looks much better.

Fixes: 93dd984ca ("Run 'make indent' on all C files")

### DIFF
--- a/test/zdtm/static/mprotect00.c
+++ b/test/zdtm/static/mprotect00.c
@@ -44,10 +44,12 @@ static int check_prot(char *ptr, int prot)
 			fail("PROT_READ bypassed");
 			return -1;
 		}
-	} else /* we come here on return from SIGSEGV handler */
+	} else {
+		/* we come here on return from SIGSEGV handler */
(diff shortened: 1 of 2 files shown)


adrianreber:
I have also seen this and thought this is a regression in clang-format. That is why the CI run to verify code formatting is running on Fedora 34. When I checked it was still using an RC release of Clang 13 and had an additional wrong indentation but that seems to be fixed.

This change looks reasonable and simple :+1: 

Can you also change `.github/workflows/lint.yml` to use `latest` instead of `34`?

Snorch:
Thanks for merging!

> Can you also change .github/workflows/lint.yml to use latest instead of 34?

@adrianreber , sure, let's do it in separate pr https://github.com/checkpoint-restore/criu/pull/1661

PR #1661, opened 2021-11-24
ci: Use latest Fedora for lint ci runs again

Snorch:
Now when we fixed clang-format complains in zdtm, let's switch to lates
clang-format available. This is effectively a revert of commit 07a2f0265
("ci: use Fedora 34 for lint CI runs").


adrianreber:
Thanks

PR #1662, opened 2021-11-24
ci: switch to centos-stream-8

adrianreber:
CentOS 8 goes EOL at the end of 2021. This switches our CentOS 8 based tests to CentOS Stream 8 which should be supported until 2024.


rst0git (approved):
LGTM

PR #1664, opened 2021-11-25
clang-format: disable wrong struct pointer declaration format

Snorch:
When we declare struct and at the same time declare variable pointer of
this struct type, it looks like clang-format threats "*" as a
multiplication operator instead of indirection (pointer declaration)
operator and puts spaces on both sides, which looks wrong.

I decided to just disable clang, other ideas are welcome. Also I considered the idea to separate structure declaration from pointer declaration, but it would not work for unnamed structures, disabling clang looks more universal.

### DIFF
--- a/criu/fdstore.c
+++ b/criu/fdstore.c
@@ -13,10 +13,12 @@
 #include "rst-malloc.h"
 #include "log.h"
 
+/* clang-format off */
 static struct fdstore_desc {
 	int next_id;
 	mutex_t lock; /* to protect a peek offset */
-} * desc;
+} *desc;
+/* clang-format on */
 
 int fdstore_init(void)
 {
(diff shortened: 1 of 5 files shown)


adrianreber:
Is this a bug/short-coming in clang-format? I am okay with the approach in this PR, just curious if it can be solved differently.

Snorch:
> Is this a bug/short-coming in clang-format?

I believe yes. But probably clang-format people just don't like our style, and would say just don't do this:

```
static struct {
        int a;
        int b;
} *pointer, variable;
```

In such a construct we have static pointer but global variable, looks weird, yes?

Much easier for understanding would be:
```
struct struct_name {
        int a;
        int b;
};

static struct struct_name *pointer;
struct struct_name variable;
```

adrianreber (approved):
The Fedora test failures seems to be related to kernel 5.15 (#1669)

avagin:
LGTM

PR #1671, opened 2021-12-03
libcriu: add single pre-dump support

adrianreber:
In contrast to the CLI it is not possible to do a single pre-dump via
RPC and thus libcriu. In cr-service.c pre-dump always goes into a
pre-dump loop followed by a final dump. runc already works around this
to only do a single pre-dump by killing the CRIU process waiting for the
message for the final dump.

Trying to implement pre-dump in crun via libcriu it is not as easy to
work around CRIU's pre-dump loop expectations as with runc that directly
talks to CRIU via RPC.

We know that LXC/LXD also does single pre-dumps using the CLI and runc
also only does single pre-dumps by misusing the pre-dump loop interface.

With this commit it is possible to trigger a single pre-dump via RPC and
libcriu without misusing the interface provided via cr-service.c. So
this commit basically updates CRIU to the existing use cases.

The existing pre-dump loop still sounds like a very good idea, but so
far most tools have decided to implement the pre-dump loop themselves.

With this change we can implement pre-dump in crun to match what is
currently implemented in runc.

### DIFF
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -1512,7 +1512,7 @@ int criu_check(void)
 	return criu_local_check(global_opts);
 }
 
-int criu_local_dump(criu_opts *opts)
+static int dump(bool pre_dump, criu_opts *opts)
 {
 	int ret = -1;
 	CriuReq req = CRIU_REQ__INIT;
@@ -1520,7 +1520,7 @@ int criu_local_dump(criu_opts *opts)
 
 	saved_errno = 0;
 
-	req.type = CRIU_REQ_TYPE__DUMP;
+	req.type = pre_dump ? CRIU_REQ_TYPE__SINGLE_PRE_DUMP : CRIU_REQ_TYPE__DUMP;
 	req.opts = opts->rpc;
 
 	ret = send_req_and_recv_resp(opts, &req, &resp);
@@ -1528,7 +1528,7 @@ int criu_local_dump(criu_opts *opts)
 		goto exit;
 
 	if (resp->success) {
-		if (resp->dump->has_restored && resp->dump->restored)
+		if (!pre_dump && resp->dump->has_restored && resp->dump->restored)
 			ret = 1;
 		else
 			ret = 0;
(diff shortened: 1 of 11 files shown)


test/others/libcriu/test_pre_dump.c
```diff
@@ -0,0 +1,153 @@
...
+	}
+	if (cur_imgdir != -1)
+		close(cur_imgdir);
+	sprintf(p, "dir-%d", ++id);
+	mkdirat(wdir_fd, p, 0700);
+	cur_imgdir = openat(wdir_fd, p, O_DIRECTORY);
+	criu_set_images_dir_fd(cur_imgdir);
+}
+
+#define SUCC_ECODE 42
```
rst0git, line 38: It looks like this constant definition is duplicated in `test_notify.c`, `test_sub.c`, `test_pre_dump.c` and `test_iters.c`.

Would it be possible to move it in `lib.h`?

lib/c/criu.c
```diff
@@ -1512,23 +1512,23 @@ int criu_check(void)
 	return criu_local_check(global_opts);
 }
 
-int criu_local_dump(criu_opts *opts)
+static int dump_or_predump(bool pre_dump, criu_opts *opts)
```
rst0git, line 1515: Hmm, what about `checkpoint` instead of `dump_or_predump`?

Snorch:
Please, can we split test to separate patch?

Snorch:
General note:

We have too many loops in cr_service...
1:
```
cr_service()
{
        while (1) {
                child_pid = fork();
                if (child_pid == 0) {
                        cr_service_work(sk);
                }
        }
}
```
2:
```
cr_service_work()
{
more:
        /* parse and handle command */
        pre_dump_loop();
        if (!ret && msg->keep_open) 
                goto more;
}
```
3:
```
pre_dump_loop()
{
        do {
                pre_dump_using_req();
        } while (msg->type == CRIU_REQ_TYPE__PRE_DUMP);
}
```
Do we really need all those loops? =) Probably we can just remove pre_dump_loop and make it just a single predump and reuse loop in cr_service_work itself?

test/others/libcriu/test_pre_dump.c
adrianreber: done

adrianreber:
> Please, can we split test to separate patch?

Done

Snorch (approved):

rst0git (approved):
LGTM

Issue #1669, opened 2021-12-01
Test Error: unix: Non-stream socket 23645 in established state


adrianreber:
We are seeing multiple socket related errors in CI. My assumption is that this is because Fedora switched to 5.15. I can reproduce the error locally after updating to 5.15:
```
$ test/zdtm.py run -t zdtm/static/socket-ext
userns is supported
=== Run 1/1 ================ zdtm/static/socket-ext
======================= Run zdtm/static/socket-ext in h ========================
Start test
./socket-ext --pidfile=socket-ext.pid --outfile=socket-ext.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/socket-ext/57/1/restore.log
------------------------ grep Error ------------------------
b'(00.001473)     57: \t\t\tGoing to dup 1 into 2'
b'(00.001475)     57: \tReceive fd for 2'
b'(00.001481)     57: \t\tCreate fd for 3'
b'(00.001483)     57: unix: Opening standalone (stage 0 id 0x8 ino 22249 peer 22883)'
b'(00.001485)     57: Error (criu/sk-unix.c:1858): unix: Non-stream socket 22249 in established state'
b'(00.001487)     57: Error (criu/files.c:1200): Unable to open fd=4 id=0x8'
b'(00.001534) Error (criu/cr-restore.c:2470): Restoring FAILED.'
b'(00.001711) Error (criu/cr-restore.c:1480): 57 killed by signal 9: Killed'
------------------------ ERROR OVER ------------------------
############### Test zdtm/static/socket-ext FAIL at CRIU restore ###############
##################################### FAIL #####################################
```

minhbq-99:
I think we need to rework this check
https://github.com/checkpoint-restore/criu/blob/6754b16c2e95766493ac27e4122a3b63867133f5/criu/sk-unix.c#L1854-L1857

This Linux kernel commit added `TCP_ESTABLISHED` state for `SOCK_DGRAM` since v5.15 https://github.com/torvalds/linux/commit/83301b5367a98c17ec0d76c7bc0ccdc3c7e7ad6d

adrianreber:
Interesting. Thanks for finding that commit. Can you provide a patch to fix it?

minhbq-99:
However, I think we have to look at this more carefully as I think there are other places in this file make the same assumption.

Issue #1635, opened 2021-10-26
CI failures on newer CPUs in CI


adrianreber:
```
+ ./test/zdtm.py run -t zdtm/static/fpu00 --fault 134 -f h --norst
userns is supported
=== Run 1/1 ================ zdtm/static/fpu00
timens isn't supported on 5.4.0-1025-aws
========================== Run zdtm/static/fpu00 in h ==========================
Start test
./fpu00 --pidfile=fpu00.pid --outfile=fpu00.out
Run criu dump
Forcing 134 fault
=[log]=> dump/zdtm/static/fpu00/54/1/dump.log
------------------------ grep Error ------------------------
b'(00.002978) Set up parasite blob using memfd'
b'(00.003005) Putting parasite blob into 0x7f9e5b21f000->0x7fa1d5310000'
b'(00.003048) Dumping general registers for 54 in native mode'
b'(00.003052) Dumping GP/FPU registers for 54'
b'(00.003061) Error (compel/arch/x86/src/lib/infect.c:298): Corrupting xsave for 54, seed 3509897626'
b'(00.003081) x86: magic1:0x46505853 extended_size:2700 xstate_bv:0x2f7 xstate_size:2696'
b'(00.003083) x86: xstate_bv: 0x2f7'
b'(00.003085) x86: -----------------------'
b'(00.003091) Putting tsock into pid 54'
b'(00.003123) Error (criu/parasite-syscall.c:88): si_code=4 si_pid=54 si_status=11'
b'(00.003126) Error (criu/parasite-syscall.c:95): 54 was stopped by 11 unexpectedly'
------------------------ ERROR OVER ------------------------
################### Test zdtm/static/fpu00 FAIL at CRIU dump ###################
Send the 9 signal to  54
Wait for zdtm/static/fpu00(54) to die for 0.100000
##################################### FAIL #####################################
```

More details can be found here: https://github.com/checkpoint-restore/criu/pull/1568#issuecomment-950762932

It seems on newer CPUs in AWS EC2 we see failures with error injection.

mihalicyn:
we've talked with @cyrillos and Kirill advised to we need to add some extra debug print's into the CRIU. I will take a look closer and prepare a patch.

mihalicyn:
I think we can close that.

PR #1673, opened 2021-12-03
ci: disable broken tests until fixed

adrianreber:
This disables know broken tests and enables fixed tests. The disabled tests are tracked in the mentioned tickets.

The disabled tests are still being run on older kernels and this helps us to easily detect is something else than the already known broken tests does not work.

Broken tests are being tracked at

 * https://github.com/checkpoint-restore/criu/issues/1669
 * https://github.com/checkpoint-restore/criu/issues/1635

This also enables previously disabled BPF related tests:

 * https://github.com/checkpoint-restore/criu/issues/1354

### DIFF
--- a/scripts/ci/vagrant.sh
+++ b/scripts/ci/vagrant.sh
@@ -50,9 +50,9 @@ fedora-no-vdso() {
 	vagrant reload
 	ssh default cat /proc/cmdline
 	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; make -j 4'
-	# BPF tests are failing see: https://github.com/checkpoint-restore/criu/issues/1354
-	# Needs to be fixed, skip for now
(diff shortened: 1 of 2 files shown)


adrianreber:
@checkpoint-restore/maintainers: We currently have (2 sometimes 3) broken CI tests on most pull requests. One error is related to new CPUs (#1635) and the other is related to kernel 5.15 (#1669). This PR disables those tests on 2 of our 24 CI runs. I think it is important to have results from CI that are not ignored. If we know that certain CI runs are always failing there is the danger of not looking correctly into it and missing new errors. So I prefer to disable broken tests until they are fixed.

I also enabled two BPF based tests which used to be disabled because they are working now (#1354).

avagin:
LGTM

PR #1674, opened 2021-12-03
test: another try to correctly fix the kernel version

adrianreber:
We try to disable time namespace based testing for kernels older than 5.11. But we fail to come up with the correct if condition.

This changes (major <= 5) to (major < 5). There are no kernels with major > 5 so currently the time namespace based are never run. This should finally change it to run time namespace based tests on kernel versions newer than 5.10.


adrianreber:
Needs #1673

PR #1678, opened 2021-12-05
test: do not use --keep-going for single zdtm tests

adrianreber:
Looking at CI logs there are often messages like:

 "[WARNING] Option --keep-going is more useful when running multiple tests"

This commit removes '--keep-going' from single zdtm test runs.


avagin:
LGTM

rst0git (approved):
LGTM

PR #1680, opened 2021-12-06
sk-unix: Fix TCP_ESTABLISHED checks in unix sockets

minhbq-99:
Since commit 83301b5367a98 ("af_unix: Set TCP_ESTABLISHED for datagram sockets
too") in Linux kernel, SOCK_DGRAM unix sockets can have TCP_ESTABLISHED state
when connected. So we need to fix checks that assume SOCK_DRAM sockets cannot
have TCP_ESTABLISHED state.

I look around for TCP_ESTABLISHED usages in sk-unix and there is one place I'm not 
sure about

```
index 8037d9cf2..d3402c3ac 100644
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -460,7 +460,7 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
                        pr_warn("Shutdown mismatch %u:%d -> %u:%d\n", ue->ino, ue->shutdown, peer->sd.ino,
                                peer->shutdown);
                }
-       } else if (ue->state == TCP_ESTABLISHED) {
+       } else if (ue->state == TCP_ESTABLISHED && ue->type != SOCK_DGRAM) {
                const struct unix_sk_listen_icon *e;
 
                e = lookup_unix_listen_icons(ue->ino);

```
I don't know if the additional check is necessary. AFAIK, SOCK_DGRAM socket sets 
TCP_ESTABLISHED state for itself and the peer socket connect() is called. At that time,
the socket's peer will be set too. So I think we cannot get 
`ue->type == SOCK_DGRAM && ue->state == TCP_ESTABLISHED)`

Fixes: #1669

### DIFF
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -460,7 +460,7 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
 			pr_warn("Shutdown mismatch %u:%d -> %u:%d\n", ue->ino, ue->shutdown, peer->sd.ino,
 				peer->shutdown);
 		}
-	} else if (ue->state == TCP_ESTABLISHED) {
+	} else if (ue->state == TCP_ESTABLISHED && ue->type != SOCK_DGRAM) {
 		const struct unix_sk_listen_icon *e;
 
 		e = lookup_unix_listen_icons(ue->ino);
@@ -1851,13 +1851,9 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
 
 		close(sks[1]);
 		sk = sks[0];
-	} else if (ui->ue->state == TCP_ESTABLISHED && queuer && queuer->ue->ino == FAKE_INO) {
+	} else if ((ui->ue->state == TCP_ESTABLISHED && ui->ue->type == SOCK_STREAM) && queuer &&
+		   queuer->ue->ino == FAKE_INO) {
 		int ret, sks[2];
-
-		if (ui->ue->type != SOCK_STREAM) {
-			pr_err("Non-stream socket %u in established state\n", ui->ue->ino);
-			return -1;
-		}
 
 		if (ui->ue->shutdown != SK_SHUTDOWN__BOTH) {
 			pr_err("Wrong shutdown/peer state for %u\n", ui->ue->ino);
--- a/scripts/ci/vagrant.sh
+++ b/scripts/ci/vagrant.sh
@@ -50,9 +50,7 @@ fedora-no-vdso() {
 	vagrant reload
 	ssh default cat /proc/cmdline
 	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; make -j 4'
-	# Disabling tests which are broken on 5.15
-	# https://github.com/checkpoint-restore/criu/issues/1669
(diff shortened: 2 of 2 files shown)


adrianreber:
Thanks for the pull request but the following tests are still failing 5.15:
```
################### 9 TEST(S) FAILED (TOTAL 414/SKIPPED 36) ####################
 * zdtm/static/socket_close_data(ns)
 * zdtm/static/socket_close_data01(uns)
 * zdtm/static/fifo_upon_unix_socket01(uns)
 * zdtm/static/sk-unix-mntns(ns)
 * zdtm/static/fifo_upon_unix_socket00(h)
 * zdtm/static/sk-unix01(uns)
 * zdtm/static/socket_dgram_data(h)
 * zdtm/static/sockets_dgram(ns)
 * zdtm/static/sk-unix-dgram-ghost(uns)
```

criu/sk-unix.c
```diff
@@ -1851,14 +1851,10 @@ static int open_unixsk_standalone(struct unix_sk_info *ui, int *new_fd)
 
 		close(sks[1]);
 		sk = sks[0];
-	} else if (ui->ue->state == TCP_ESTABLISHED && queuer && queuer->ue->ino == FAKE_INO) {
+	} else if ((ui->ue->state == TCP_ESTABLISHED && ui->ue->type == SOCK_STREAM) && queuer &&
+		   queuer->ue->ino == FAKE_INO) {
```
rst0git, line 1855: @minhbq-99 I think we might need the follow change as well:

```diff
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -460,7 +460,7 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
                        pr_warn("Shutdown mismatch %u:%d -> %u:%d\n", ue->ino, ue->shutdown, peer->sd.ino,
                                peer->shutdown);
                }
-       } else if (ue->state == TCP_ESTABLISHED) {
+       } else if (ue->state == TCP_ESTABLISHED && ue->type != SOCK_DGRAM) {
                const struct unix_sk_listen_icon *e;
 
                e = lookup_unix_listen_icons(ue->ino);
```

minhbq-99:
> We just disabled the broken tests on 5.15 and if this PR fixes it you should re-enable the tests again. You have to remove all `-x` from `scripts/ci/vagrant.sh`. If that passes CI that would show #1669 is fixed.

But my branch does not rebase to that commit and in the log of vagrant fedora, I see that those disabled tests have been run.

adrianreber:
> > We just disabled the broken tests on 5.15 and if this PR fixes it you should re-enable the tests again. You have to remove all `-x` from `scripts/ci/vagrant.sh`. If that passes CI that would show #1669 is fixed.
> 
> But my branch does not rebase to that commit and in the log of vagrant fedora, I see that those disabled tests have been run.

Good point. Would you please rebase and enable those tests again? If you do not want to rebase we can do a follow up PR ourselves.

minhbq-99:
> Good point. Would you please rebase and enable those tests again? If you do not want to rebase we can do a follow up PR ourselves.

I rebased and added a new commit to re-enable these tests

adrianreber:
> > Good point. Would you please rebase and enable those tests again? If you do not want to rebase we can do a follow up PR ourselves.
> 
> I rebased and added a new commit to re-enable these tests

Thanks a lot. Changes look correct. Let's wait for CI results.

rst0git (approved):
LGTM

adrianreber (approved):
All broken tests are working again on 5.15. Thanks!

adrianreber:
@avagin any comments from your side.

avagin:
LGTM.
offtopic: I just found that we don't handle SOCK_SEQPACKET sockets. @minhbq-99 Maybe you have time to add support of them. It should not be hard.

PR #1681, opened 2021-12-06
lib: introduce feature check in libcriu

adrianreber:
This commit adds feature check support to libcriu. It already exists in the CLI and RPC and this just extends it to libcriu.

This commit provides one function to do all possible feature checks in one call. The parameter to the feature check function is a structure and the user can enable which features should be checked.

Using a structure makes the function extensible without the need to break the API/ABI in the future.

### DIFF
--- a/lib/c/criu.c
+++ b/lib/c/criu.c
@@ -1925,3 +1925,75 @@ int criu_join_ns_add(const char *ns, const char *ns_file, const char *extra_opt)
 {
 	return criu_local_join_ns_add(global_opts, ns, ns_file, extra_opt);
 }
+
+int criu_local_feature_check(criu_opts *opts, struct criu_feature_check *features, size_t size)
+{
+	CriuFeatures criu_features = CRIU_FEATURES__INIT;
+	struct criu_feature_check features_copy = { 0 };
+	CriuReq req = CRIU_REQ__INIT;
+	CriuResp *resp = NULL;
+	int ret = -1;
+
+	saved_errno = 0;
+
+	if (!features)
+		goto exit;
+
+	if (size > sizeof(struct criu_feature_check))
+		goto exit;
+
+	memcpy(&features_copy, features, size);
+
+	req.type = CRIU_REQ_TYPE__FEATURE_CHECK;
+	req.opts = opts->rpc;
+
+	if (features_copy.mem_track) {
+		criu_features.has_mem_track = true;
+		criu_features.mem_track = true;
+	}
+	if (features_copy.lazy_pages) {
+		criu_features.has_lazy_pages = true;
+		criu_features.lazy_pages = true;
+	}
+	if (features_copy.pidfd_store) {
+		criu_features.has_pidfd_store = true;
+		criu_features.pidfd_store = true;
+	}
+	req.features = &criu_features;
+
+	ret = send_req_and_recv_resp(opts, &req, &resp);
+	if (ret)
+		goto exit;
+
+	memset(&features_copy, 0, sizeof(struct criu_feature_check));
+
+	if (resp->success) {
+		if (resp->features->has_mem_track) {
+			features_copy.mem_track = resp->features->mem_track;
+		}
+		if (resp->features->has_lazy_pages) {
+			features_copy.lazy_pages = resp->features->lazy_pages;
+		}
+		if (resp->features->has_pidfd_store) {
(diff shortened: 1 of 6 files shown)


rst0git (approved):
LGTM

Snorch:
Can we please split test and libcriu changes?

adrianreber:
> Can we please split test and libcriu changes?

Done

test/others/libcriu/run.sh
```diff
@@ -62,6 +62,16 @@ if [ "$(uname -m)" = "x86_64" ]; then
 fi
 run_test test_errno
 run_test test_join_ns
+if criu check --feature mem_dirty_track > /dev/null; then
+	export CRIU_FEATURE_MEM_TRACK=1
```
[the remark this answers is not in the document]
adrianreber: Not sure I understand you correctly. If the feature is not enable, for example dirty page tracking will always be disabled on aarch64 because of missing kernel support, the environment is not set and then in `test_feature_check.c` it will not be set to true and so we will not check it. But that does mean it will not be checked at all. Hmm...

Ah, I think I understand. If it is disabled we should still check the feature and get the result that it is disabled. So I always have to check all features and compare the result with the result from `criu check`. Makes sense.

lib/c/criu.c
```diff
@@ -1925,3 +1925,75 @@ int criu_join_ns_add(const char *ns, const char *ns_file, const char *extra_opt)
...
 	return criu_local_join_ns_add(global_opts, ns, ns_file, extra_opt);
 }
+
+int criu_local_feature_check(criu_opts *opts, struct criu_feature_check *features, size_t size)
+{
+	int ret = -1;
+	CriuReq req = CRIU_REQ__INIT;
+	CriuResp *resp = NULL;
+	CriuFeatures criu_features = CRIU_FEATURES__INIT;
+	struct criu_feature_check features_copy = { 0 };
```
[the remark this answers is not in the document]
adrianreber: Reverse Christmas tree :smile:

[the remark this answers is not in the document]
adrianreber: done

adrianreber:
@Snorch PTAL

test/others/libcriu/test_feature_check.c
```diff
@@ -0,0 +1,65 @@
...
+		return -1;
+
+	ret = criu_feature_check(&features, sizeof(features));
+	if (ret < 0) {
+		what_err_ret_mean(ret);
+		return ret;
+	}
+
+	printf("   `- mem_track  : %d - expected : %d\n", features.mem_track, mem_track);
+	if (features.mem_track && mem_track != 1)
```
Snorch, line 55: Check is strange... Why not `features.mem_track != mem_track`? I believe your check does not detect the case where mem_track is true and features.mem_track is false.

test/others/libcriu/test_feature_check.c
```diff
@@ -0,0 +1,65 @@
...
+		.pidfd_store = true,
+	};
+
+	printf("--- Start feature check ---\n");
+	criu_init_opts();
+	criu_set_service_binary(argv[1]);
+
+	env = getenv("CRIU_FEATURE_MEM_TRACK");
+	if (env) {
+		mem_track = atoi(env);
```
Snorch, line 32: Do we really need atoi here? Why not mem_track = true?

test/others/libcriu/test_feature_check.c
adrianreber: fixed

test/others/libcriu/test_feature_check.c
adrianreber: Yes, not necessary. Removed.

test/others/libcriu/test_feature_check.c
```diff
@@ -0,0 +1,65 @@
+#include "criu.h"
```
rst0git, line 1: Would it be better to move `#include "criu.h"` after the standard headers?

adrianreber: Better? Don't know. It should work in whichever way we include it. I guess this is just copied from the other test cases in this directory.

adrianreber:
@Snorch anything missing?

Snorch (approved):
@adrianreber, sorry for slow response, all looks good.

PR #1682, opened 2021-12-07
util: make page-server IPv6 safe

adrianreber:
The function run_tcp_server() was the last place CRIU was still using the IPv4 only function inet_ntoa(). It was only used during a print, so that it did not really break anything, but with this commit the output is now no longer:

` Accepted connection from 0.0.0.0:58396`

but correctly displaying the IPv6 address

` Accepted connection from ::1:58398`

if connecting via IPv6.

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -1107,7 +1107,7 @@ out:
 int run_tcp_server(bool daemon_mode, int *ask, int cfd, int sk)
 {
 	int ret;
-	struct sockaddr_in caddr;
+	struct sockaddr_storage caddr;
 	socklen_t clen = sizeof(caddr);
 
 	if (daemon_mode) {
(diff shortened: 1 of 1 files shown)


rst0git (approved):
LGTM

criu/util.c
```diff
@@ -1139,9 +1139,17 @@ int run_tcp_server(bool daemon_mode, int *ask, int cfd, int sk)
...
 			goto err;
-		} else
-			pr_info("Accepted connection from %s:%u\n", inet_ntoa(caddr.sin_addr),
-				(int)ntohs(caddr.sin_port));
+		} else {
+			char port[6];
+			char address[INET6_ADDRSTRLEN];
+			ret = getnameinfo((struct sockaddr *)&caddr, clen, address, sizeof(address), port, sizeof(port),
+					  NI_NUMERICHOST | NI_NUMERICSERV);
+			if (unlikely(ret)) {
```
avagin, line 1147: nit: you don't need this unlikely here

criu/util.c
```diff
@@ -1139,9 +1139,17 @@ int run_tcp_server(bool daemon_mode, int *ask, int cfd, int sk)
 		if (*ask < 0) {
 			pr_perror("Can't accept connection to server");
 			goto err;
-		} else
-			pr_info("Accepted connection from %s:%u\n", inet_ntoa(caddr.sin_addr),
-				(int)ntohs(caddr.sin_port));
+		} else {
```
avagin, line 1142: else isn't needed here.

adrianreber:
@avagin thanks for the review. I fixed both review comments and force pushed.

PR #1684, opened 2021-12-10
Export current criu mode to opts

Snorch:
We have multiple options which are valid only on restore or only on dump or in any other specific criu mode, so it would be useful to have info about current mode in opts so that we can validate other options against current mode.

Plan is to use it for mount-v2 (switch from current restore mount engine to a new one) option as it is only valid on restore, and this would make handling of different types of mountpoints much easier.

- When criu mode is set from main() we just parse mode from argv[optind] just after parse_options() found optind of the command. Note that opts.mode is available before check_options().

- For rpc service we reset opts.mode to CR_SWRK each time we restart cr_service_work(), in the original service process we still have CR_SERVICE to differentiate between them, and each request handling function which does setup_opts_from_req sets opts.mode in accordance with the processed request type. And it is also available before check_options().
    
Now in check_options we can add filters on one mode only options.

Note: I chose to call this "mode" as in rpc we have a request type and on cli we have command, so "mode" should unify these two similar terms in one.

### DIFF
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -222,6 +222,9 @@ int main(int argc, char *argv[], char *envp[])
 		if (opts.mode != CR_CPUINFO && has_sub_command) {
 			pr_err("excessive parameter%s for command %s\n", (argc - optind) > 2 ? "s" : "", argv[optind]);
 			goto usage;
+		} else if (opts.mode == CR_CPUINFO && !has_sub_command) {
+			pr_err("cpuinfo requires an action: dump or check\n");
+			goto usage;
 		}
 	}
 
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -124,7 +124,7 @@ int main(int argc, char *argv[], char *envp[])
 		return 1;
 	}
 
-	if (optind < argc && !strcmp(argv[optind], "swrk")) {
+	if (!strcmp(argv[optind], "swrk")) {
 		if (argc != optind + 2) {
 			fprintf(stderr, "Usage: criu swrk <fd>\n");
 			return 1;
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -56,14 +56,23 @@ void flush_early_log_to_stderr(void)
 
 static int image_dir_mode(char *argv[], int optind)
 {
-	if (!strcmp(argv[optind], "dump") || !strcmp(argv[optind], "pre-dump") ||
-	    (!strcmp(argv[optind], "cpuinfo") && !strcmp(argv[optind + 1], "dump")))
+	switch (opts.mode) {
+	case CR_DUMP:
+		/* fallthrough */
+	case CR_PRE_DUMP:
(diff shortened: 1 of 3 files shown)


Snorch:
@checkpoint-restore/maintainers Please take a look, hope this would simplify everything. I thought about a way to differentiate between dump/restore already for a long time, in many places we already pass restore/dump info in function parameters, maybe switching to this approach would be better.

adrianreber (approved):
Sounds like a good idea and all CI tests are happy. :+1:

mihalicyn (approved):

criu/include/cr_options.h
```diff
@@ -100,6 +100,22 @@ struct irmap_path_opt {
...
+	CR_LAZY_PAGES,
+	CR_CHECK,
+	CR_PAGE_SERVER,
+	CR_SERVICE,
+	CR_SWRK,
+	CR_DEDUP,
+	CR_CPUINFO,
+	CR_EXEC_DEPRECATED,
+	CR_SHOW_DEPRECATED,
+};
```
rst0git, line 117: Would it be possible to update `image_dir_mode()` in `crtools.c` with these enum values?

Snorch: Thanks for review! Sure, I forgot about it.

criu/crtools.c
```diff
@@ -56,13 +56,20 @@ void flush_early_log_to_stderr(void)
...
 		return O_RSTR;
+	case CR_CPUINFO:
+		if (!strcmp(argv[optind + 1], "dump"))
+			return O_DUMP;
+		/* fallthrough */
+	default:
+		return -1;
+	}
 
 	return -1;
```
rst0git, line 74: This `return -1;` is not used.

criu/crtools.c
```diff
@@ -185,7 +222,7 @@ int main(int argc, char *argv[], char *envp[])
 	}
 
 	/* We must not open imgs dir, if service is called */
-	if (strcmp(argv[optind], "service")) {
+	if (opts.mode != CR_SERVICE) {
```
rst0git, line 232: Could you please add something like the following to fix a segmentation fault for `criu cpuinfo`?
```C
if (opts.mode == CR_CPUINFO && argc < 3) {
        pr_err("cpuinfo requires an action: dump or check\n");
        goto usage;
}
```

criu/crtools.c
Snorch: Sure I know, it is a combination of two best practices: 1) switch-case should have explicit default 2) int function should have proper return at the end.

I can add a comment /* Should never get here */ if needed.

criu/crtools.c
Snorch: Ok, will do.

adrianreber:
Yes, @rst0git also mentioned it. Very strange, not sure what it is.

@rst0git any insights yet?

It is a bit worrying that so many tests are failing with a segfault.

PR #1689, opened 2021-12-14
usernsd: UNS_FDOUT should not require an input descriptor

avagin:
UNS_FDOUT means only that a userns call will return a file descriptor.


rst0git (approved):
LGTM

PR #1690, opened 2021-12-14
page-server: don't set TCP_CORK for tls sockets

rst0git:
This pull request aims to address the errors seen in https://github.com/checkpoint-restore/criu/issues/1380. It looks like this error occurs sometimes after a test the restore phase of a test, when zdtm attempts to kill the page-server and lazy-pages processes.

In particular, when TCP_CORK is set on the socket it tells TCP to wait to remove the cork before sending any packets. As a results, we block the underlying channel, but still send data to gnutls which then produces multiple encryption packets.

This pull request avoids setting TCP_CORK when TLS is used, and fixes a few other minor issues.

### DIFF
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -1389,6 +1389,11 @@ int cr_page_server(bool daemon_mode, bool lazy_dump, int cfd)
 	if (init_stats(DUMP_STATS))
 		return -1;
 
+	/*
+	 * We might need a lot of pipes to fetch huge number of pages.
+	 */
+	rlimit_unlimit_nofile();
+
 	if (!opts.lazy_pages)
 		up_page_ids_base();
 	else if (!lazy_dump)
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -1178,7 +1178,8 @@ static int page_server_serve(int sk)
 		pr_debug("Created xfer pipe size %u\n", cxfer.pipe_size);
 	} else {
 		pipe_read_dest_init(&pipe_read_dest);
-		tcp_cork(sk, true);
+		if (!opts.tls)
+			tcp_cork(sk, true);
 	}
 
 	while (1) {
@@ -1259,7 +1260,7 @@ static int page_server_serve(int sk)
 		ret = -1;
 	}
 
-	if (ret == 0 && opts.ps_socket == -1) {
+	if (ret == 0 && opts.ps_socket == -1 && !opts.tls) {
 		char c;
 
 		/*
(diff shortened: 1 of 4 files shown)


adrianreber:
Hard to tell if it actually fixes #1380 because it does not happen always, but I am :+1:  merging this to see if #1380 is gone.

criu/page-xfer.c
```diff
@@ -1389,6 +1389,11 @@ int cr_page_server(bool daemon_mode, bool lazy_dump, int cfd)
 	if (init_stats(DUMP_STATS))
 		return -1;
 
+	/*
+	 * We might need a lot of pipes to fetch huge number of pages.
+	 */
+	rlimit_unlimit_nofile();
```
avagin, line 1396: Does the page server splice memory into pipes?

criu/tls.c
```diff
@@ -70,6 +70,9 @@ ssize_t tls_send(const void *buf, size_t len, int flags)
 		case GNUTLS_E_UNEXPECTED_PACKET_LENGTH:
 			errno = ENOMSG;
 			break;
+		case GNUTLS_E_PREMATURE_TERMINATION:
+			errno = ECONNRESET;
```
avagin, line 74: where do we terminate the tcp connection?

test/zdtm.py
```diff
@@ -1886,8 +1886,8 @@ def do_run_test(tname, tdesc, flavs, opts):
                 check_visible_state(t, s, opts)
                 if opts['join_ns']:
                     check_joinns_state(t)
-                t.stop()
```
avagin, line 1889: This is here intentionally. Processes have to be fully functional after restore.

avagin:
> In particular, when TCP_CORK is set on the socket it tells TCP to wait to remove the cork before sending any packets. As a result, we block the underlying channel, but still send data to gnutls which then produces multiple encryption packets.

I don't understand this. Could you elaborate when we send data to a tcp sock and when we send data to gnutls?

avagin:
TCP_CORK is a kernel feature. For tls, it should look like one packet. From the commit message, I don't understand why it is a problem. Are you able to reproduce the bug? Are you sure that this patch series fix it?

rst0git:
> TCP_CORK is a kernel feature.

> Could you elaborate when we send data to a tcp sock and when we send data to gnutls?

TLS encryption/decryption happens in userspace. Instead of directly sending data to the socket with `send()`/`recv()`, we use `gnutls_record_send()` and `gnutls_record_recv()`.

There is `gnutls_record_cork()` and `gnutls_record_uncork()` that could be used as a replacement, but this is unrelated to the failures we have seen in CI.

> From the commit message, I don't understand why it is a problem.

TCP_CORK is not the problem.

> Are you able to reproduce the bug? Are you sure that this patch series fix it?

I'm able to reproduce the error locally and as described in https://github.com/checkpoint-restore/criu/issues/1380. I have updated the GitHub issue with all logs files. I have also been using the CI in my fork for testing.

I moved the fix in a separate pull request with more detailed description: https://github.com/checkpoint-restore/criu/pull/1693

PR #1691, opened 2021-12-15
sk-unix: Add support for SOCK_SEQPACKET unix sockets

minhbq-99:
I adjusted some SOCK_STREAM cases to handle SOCK_SEQPACKET too.
I also added SOCK_SEQPACKET variants on unix stream socket tests.


adrianreber:
I guess this is a follow up of #1680 where @avagin suggested this change. Thanks!

mihalicyn:
@minhbq-99 great work! ;)

You've also deleting `test/zdtm/static/mntns-deleted-dst` file (which is empty). Do you really want to remove it in this series?

It's also good to play with and extend `test/zdtm/static/scm03.c` test with `SOCK_SEQPACKET` and check that SCM's works fine.

minhbq-99:
> You've also deleting `test/zdtm/static/mntns-deleted-dst` file (which is empty). Do you really want to remove it in this series?

I remove the change with mntns-deleted-dst in the patch.

> It's also good to play with and extend `test/zdtm/static/scm03.c` test with `SOCK_SEQPACKET` and check that SCM's works fine.

I just added a `SOCK_SEQPACKET` variant for scm03.c by changing the sock_type from `SOCK_DRAM` to `SOCK_SEQPACKET`

mihalicyn:
LGTM

Thanks!

mihalicyn (approved):

avagin:
LGTM

PR #1692, opened 2021-12-16
mount/btrfs: make __open_mountpoint fallback to get_sdev_from_fd

Snorch:
We face that btrfs returns anonymous device in stat instead of real
superblock dev for volumes, thus __open_mountpoint can't open btrfs
mountpoint due to dev missmatch between stat and mountinfo. We can use
special helper get_sdev_from_fd instead of stat to get real dev of fd.

New get_sdev_from_fd helper first gets mnt_id from fd using fdinfo and
then converts mnt_id to sdev using mountinfo.

### DIFF
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -1527,6 +1527,59 @@ out:
 	return exit_code;
 }
 
+static int get_mountinfo_sdev_from_mntid(int mnt_id, unsigned int *sdev)
+{
+	int exit_code = -1;
+	FILE *f;
+
+	f = fopen_proc(PROC_SELF, "mountinfo");
+	if (!f)
+		return -1;
+
+	while (fgets(buf, BUF_SIZE, f)) {
+		unsigned int kmaj, kmin;
+		int id;
+
+		if (sscanf(buf, "%i %*i %u:%u", &id, &kmaj, &kmin) != 3) {
+			pr_err("Failed to parse mountinfo line %s\n", buf);
+			goto err;
+		}
+
+		if (id == mnt_id) {
+			*sdev = MKKDEV(kmaj, kmin);
+			exit_code = 0;
+			break;
+		}
+	}
+err:
+	fclose(f);
+	return exit_code;
+}
+
+/* This works even on btrfs where stat does not show right sdev */
+int get_sdev_from_fd(int fd, unsigned int *sdev, bool parse_mountinfo)
+{
+	struct mount_info *mi;
+	int ret, mnt_id;
+
+	ret = get_fd_mntid(fd, &mnt_id);
+	if (ret < 0)
+		return -1;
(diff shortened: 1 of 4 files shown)


Snorch:
Segfault on rawhide fedora looks unrelated as I see same on my different PR.

criu/proc_parse.c
```diff
@@ -1527,6 +1527,56 @@ int parse_timens_offsets(struct timespec *boff, struct timespec *moff)
 	return exit_code;
 }
 
+static int get_mountinfo_sdev_from_mntid(int mnt_id, unsigned int *sdev)
+{
+	int exit_code = -1;
+	FILE *f;
+
+	f = fopen_proc(PROC_SELF, "mountinfo");
```
avagin, line 1535: can we read criu's mountiunfo once?

Snorch: Thanks for this question!

Disclamer: It looks like I had the right idea but implemented it wrong.

We have one code path where __open_mountpoint() can be called with mnt_fd != -1 in open_mountpoint(). In all other code paths we just open mnt_fd ourselfs as pm->ns_mountpoint at mntns_root, so fdinfo mount id would be from same mntns as pm, thus always
present in mountinfo.

But in open_mountpoint we can open fd via: 
1) get_clean_fd which would give fd with mnt_id of newly created clean bindmount, which we don't have in mountinfo
(patch logic with reading mountinfo should not be helping in this case as before __open_mountpoint we do restore_mnt_ns =( )
2) ns_open_mountpoint which would give fd with mnt_id from newly cloned helper-mntns, which we also don't have in mountinfo

So if we dump overmounted btrfs volume mount (which we don't do in tests, that's why I didn't initially found it) __open_mountp
oint would still fail to validate mnt_fd.

I will think some more on how to fix it, if have any ideas - please share them.

Snorch:
I've prepared an updated aproach which should handle all cases most effectively. @avagin PTAL

note: Currently open_mountpoint would not be called on btrfs, so we would never get to actually parsing mountinfo for now. But it's good to be ready for it, because we would use open_mountpoint for overmounted files dumping on any types of mounts.

We will be able to add a test for it when porting overmounted files handling on top of mount-v2.

Snorch:
Some more description of "new" approach:
1) Refactor __open_mountpoint:
  - split check_mountpoint_fd to separate function
  - remove mnt_fd argument from __open_mountpoint and use check_mountpoint_fd where we don't need to open but only check fd
2) Add argument to get_sdev_from_fd to allow parsing current mountinfo, will be used only with temporary mounts just created by criu in current mntns.
3) Move check_mountpoint_fd in open_mountpoint to places where we can parse mountinfo for temporary mounts.

mihalicyn (approved):

PR #1698, opened 2021-12-17
zdtm: remove mntns-deleted-dst test leftover from git

Snorch:
Looks like in commit [1] we've non-intentionally added this tmp file to
git, let's remove it.

Fixes: 01ee29702 ("s390:zdtm: Enable zdtm for s390") [1]


rst0git (approved):
LGTM

Issue #1699, opened 2021-12-18
No rule to make target 'umount2' when make test


Yangfisher1:
Hi, I'm trying to make a pull request, so I followed the instruction [How_to_submit-patches](https://criu.org/How_to_submit_patches). Entering `make` seemed all right, but when I entered `make test`, it showed out that `make[3]: *** No rule to make target 'umount2'.  Stop.`. And I checked out the `Makefile` in `test` directory. It actually doesn't have the target `umount2`, and the branch is `criu-dev`. So is there any solution for this problem? Also I want to know that is there anything I should pay attention to if I want to make a pull request? Thanks for your help!

mihalicyn:
@Yangfisher1 hi,

please tell us, do you have the same error when trying the `sudo ./test/zdtm.py run -a`?

The problem probably is that when you run `make test` all environment variables from "root" make passes through zdtm.py to nested make here:
```
    def available():
        if not os.access("umount2", os.X_OK):
            subprocess.check_call(["make", "umount2"])
...
```

I will send a fix for this problem. But I'm not sure that `make test` works perfectly in all cases.

mihalicyn:
Please feel free to post your PR if `./test/zdtm.py run -a` passes.

Thanks,
Alex

Yangfisher1:
Thanks a lot! It really worked out, but the result showed that the kernel is tainted. I'll try to use another machine cause the machine is shared with many people. Anyway, I'll post a PR as long as I solve the problem.

adrianreber:
> Thanks a lot! It really worked out, but the result showed that the kernel is tainted. I'll try to use another machine cause the machine is shared with many people. Anyway, I'll post a PR as long as I solve the problem.

There is the option --ignore-taint which should ignore it for your local test.

Yangfisher1:
Wow, I'll try this one too, thanks.

Yangfisher1:
May I ask a question more? I ran `sudo ./zdtm.py run -a`, when testing `zdtm/static/net_lock_socket_iptables`, it showed out that
```
Running zdtm/static/net_lock_socket_iptables.hook(--post-start)
./net_lock_socket_iptables --pidfile=net_lock_socket_iptables.pid --outfile=net_lock_socket_iptables.out
Traceback (most recent call last):
  File "zdtm/static/net_lock_socket_iptables.hook", line 44, in <module>
    test_sock = socket.create_connection((server_addr, PORT), TIMEOUT)
[...]

PR #1700, opened 2021-12-18
Fix `make test` compile failure

mihalicyn:
This PR addresses the issue from https://github.com/checkpoint-restore/criu/issues/1699

We use implicit make rules to produce `test/umount2` and `test/zdtm_ct` binaries, but when `make umount2`
is called from zdtm.py which is called from `make test` (under the CRIU tree root) this nested make inherits `MAKEFLAGS`
variable with the value `rR --no-print-directory` and that's the problem because `-r` means "no builtin rules".
Let's just drop this variable value for `make umount2` and `make zdtm_ct` this is safe.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -607,9 +607,11 @@ class zdtm_test:
     @staticmethod
     def available():
         if not os.access("umount2", os.X_OK):
-            subprocess.check_call(["make", "umount2"])
+            subprocess.check_call(
+                ["make", "umount2"], env=dict(os.environ, MAKEFLAGS=""))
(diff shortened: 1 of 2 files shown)


rst0git (approved):
LGTM

adrianreber (approved):

Snorch (approved):
I checked with `rm test/umount2` and it works. Looks good.

adrianreber:
@mihalicyn If you rebase CI should turn green.

mihalicyn:
@adrianreber thanks. All green now

Issue #1653, opened 2021-11-14
Fails to build on ARMv7 with gcc-11: cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU


carnil:
Hi

Triggered in Debian after switch to building with gcc-11 `criu` fails to build on armhf:

```
make[1]: git: No such file or directory
/usr/bin/which: this version of `which' is deprecated; use `command -v' in scripts instead.
  GEN      .gitid
  GEN      include/common/asm
  GEN      compel/include/asm
touch .config
  GEN      compel/include/version.h
  GEN      criu/include/version.h
  GEN      include/common/config.h
  GEN      compel/plugins/include/uapi/std/asm/syscall-types.h
  DEP      compel/arch/arm/plugins/std/parasite-head.d
  GEN      compel/arch/arm/plugins/std/syscalls/syscalls.S
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
  GEN      compel/plugins/include/uapi/std/syscall-codes.h
  DEP      soccr/soccr.d
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
  GEN      compel/plugins/include/uapi/std/syscall.h
  CC       soccr/soccr.o
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
make[2]: *** [/<<PKGBUILDDIR>>/scripts/nmk/scripts/build.mk:119: soccr/soccr.o] Error 1
make[1]: *** [Makefile:232: soccr/built-in.o] Error 2
make[1]: *** Waiting for unfinished jobs....
  DEP      compel/arch/arm/plugins/std/syscalls/syscalls.d
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
  DEP      compel/plugins/std/infect.d
  DEP      compel/plugins/std/string.d
  DEP      compel/plugins/std/log.d
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
  DEP      compel/plugins/std/fds.d
  DEP      compel/plugins/std/std.d
  DEP      compel/plugins/shmem/shmem.d
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
cc1: error: '-mfloat-abi=hard': selected architecture lacks an FPU
[...]

PR #1701, opened 2021-12-18
ci: fix cross-compile tests

rst0git:
We added cross-compile tests with testing debian release to be able to
replicate the error reported in #1653. However, installing build
dependencies in this release currently fails with the following error:

    libc6-dev:armhf : Breaks: libc6-dev-armhf-cross (< 2.33~) but 2.32-1cross4 is to be installed

This is not something we can fix, therefore using the debian unstable
release (instead of testing) could be more reliable option for our CI.
This would still allow us to replicate the problem reported in #1653.


adrianreber:
Changing from `testing` to `unstable` sounds more unstable :wink:

I was hoping that the repository problem would have been fixed faster, but if this runs the tests without errors it probably makes sense to switch to `unstable`.

PR #1703, opened 2021-12-19
tls: handle ECONNRESET in recv callback

rst0git:
The --remote-lazy-pages option of zdtm allows to run a test with simulated post-copy (also know as "lazy") migration. When running tests with this option zdtm starts two additional processes (1) page-server and (2) lazy-pages.

For example, a zdtm test would look like:

    1. criu dump ...
    2. criu page-server --port 12345 --lazy-pages ...
    3. criu lazy-pages --page-server --port 12345 --address 127.0.0.1 ...
    4. criu restore --lazy-pages --restore-detached ...

The `--restore-detached` would result in criu to detach itself once restore is complete, but not all memory pages have been migrated. Then zdtm would send a `SIGTERM` to the restored process and wait for the lazy-pages process and the page-server process to exit.

As a result, the TCP socket between page-server and lazy-pages could close unexpectedly and when TLS encryption is enabled we sometimes see error messages like:

    Error (criu/tls.c:319): tls: Pull callback recv failed: Connection reset by peer
    Error (criu/tls.c:146): tls: Failed receiving data: Error in the pull function.
    Error (criu/page-xfer.c:1193): page-xfer: Can't read pagemap from socket: I/O error
    Error (criu/tls.c:307): tls: Push callback send failed: Broken pipe

This patch updates the recv callback used with gnutls to handle `ECONNRESET` when the connection was not properly terminated (`man 3 gnutls_transport_set_pull_function`).

### DIFF
--- a/criu/tls.c
+++ b/criu/tls.c
@@ -142,6 +142,8 @@ ssize_t tls_recv(void *buf, size_t len, int flags)
 		case GNUTLS_E_INTERRUPTED:
 			errno = EINTR;
 			break;
+		case GNUTLS_E_PREMATURE_TERMINATION:
+			return 0;
 		default:
 			tls_perror("Failed receiving data", ret);
 			errno = EIO;
@@ -314,6 +316,9 @@ static ssize_t _tls_pull_cb(void *p, void *data, size_t sz)
 {
 	int fd = *(int *)(p);
 	int ret = recv(fd, data, sz, tls_sk_flags);
+	/* The callback should return 0 on connection termination */
+	if (ret < 0 && errno == ECONNRESET)
+		return 0;
 	if (ret < 0 && errno != EAGAIN) {
 		int _errno = errno;
 		pr_perror("Pull callback recv failed");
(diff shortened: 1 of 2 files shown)


adrianreber:
The `lazy-thp` test still seems to fail:
```
Run criu page-server
Run criu lazy-pages
Run criu restore
Send the 15 signal to  126
Wait for zdtm/transition/lazy-thp(126) to die for 0.100000
=[log]=> dump/zdtm/transition/lazy-thp/100/1/page-server.log
b'(00.000865) Starting page server on port 12345'
b'(00.000886) Running status-ready scripts'
b'(00.031113) Accepted connection from 127.0.0.1:54618'
b'(00.031153) page-pipe: Created tee pipe size 4194304'
b"(00.043590) Error (criu/page-xfer.c:1193): page-xfer: Can't read pagemap from socket: Connection reset by peer"
##### Test zdtm/transition/lazy-thp FAIL at criu page-server exited with 1 #####
```

avagin:
@rst0git could you file an issue? It is hard to follow a discussion in closed pr-s. If you know how to reproduce this issue, it would be good to have all logs.

avagin:
@rst0git handle_exit is the place where uffd handles a case when a process is exited.

rst0git:
@avagin I've uploaded the logs and added the steps for replicating the issue in https://github.com/checkpoint-restore/criu/issues/1380

Issue #1696, opened 2021-12-16
Implement rseq support, as required by glibc 2.35


fweimer-rh:
glibc 2.35 enables restartable sequences (rseq) during process startup. This designates a memory area, associated with each thread, to which the kernel can write as it sees fit.

This has two consequences for CRIU: A restored process can crash once the memory for the registered rseq area is unmapped (this is the rseq area for the *new* process). And the restored process, when running, will appear to present a successful rseq registration, but the kernel will not write to the rseq areas of any restored threads.

Linux 5.13 added `PTRACE_GET_RSEQ_CONFIGURATION` in commit torvalds/linux@90f093fa8ea48e5d991332cee160b761423d55c1 to support checkpoint and restore.

Downstream bug report: https://bugzilla.redhat.com/show_bug.cgi?id=2033397

adrianreber:
Seeing https://github.com/torvalds/linux/commit/90f093fa8ea48e5d991332cee160b761423d55c1 was developed by Google to support checkpoint/restore I am wondering if Google has already a patch for this for their internal CRIU version.

@mclapinski you seemed to have worked on CRIU for Google. Do you know if anything regarding rseq exists in Google's version of CRIU?

mihalicyn:
@fweimer-rh thanks a lot for looking into this!

If nobody wants to implement that and our Google friends have no support of that feature yet I'm ready to take this issue on myself.

adrianreber:
Please revert #1697 again if this is fixed.

mihalicyn:
I've played with `rseq` locally and written dump part implementation. But before diving into full and prod-ready implementation let's wait 1-2 days for some response from our friends.

figiel:
Hi, this was used only in a prototype. I don't have a production ready patch for CRIU which would use this yet (and not planning to focus on that in upcoming weeks).

adrianreber:
@figiel Thanks for your feedback.

fweimer-rh:
Is there anything I can to do to help with moving this forward? Thanks.

adrianreber:
This is the announcement of CentOS 9 also wanting to switch to rseq: https://lists.centos.org/pipermail/centos-devel/2022-January/120153.html

melsamathew:
I am facing one issue with glibc upgrade, so found this thread while searching for the error.. Is this issue related to the rseq support in glibc? Please see the problem below and give support for this

[...]

PR #1706, opened 2021-12-21
rseq() c/r support

mihalicyn:
This is patchset provides the `rseq()` C/R support

There are four patches which provide desired support:
cr-dump: handle rseq/rseq_cs flags properly
    
    Userspace may configure rseq cs abort policy by
    setting RSEQ_CS_FLAG_NO_RESTART_ON_* flags.
    
    In ("cr-dump: fixup thread IP when inside rseq cs") we have supported
    the case when process was caught by CRIU during rseq cs execution by
    fixing up IP to abort_ip. Thats a common case, but there is special flag
    called RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL, in this case we have to leave
    process IP as it was before CRIU seized it. Unfortunately, that's not
    all that we need here. We also must preserve (struct rseq)->rseq_cs field.
    
    You may ask like "why we need to preserve it by hands? CRIU is dumping
    all process memory and restores it". That's true. But not so easy. The problem
    here is that the kernel performs this field cleanup when it realized that
    the process gets out of rseq cs. But during dump/restore procedures we are
    executing parasite/restorer from the process context. It means that process
    will get out of rseq cs in any case and (struct rseq)->rseq_cs will be cleared
    by the kernel. So we need to restore this field by hands at the *last* stage
    of restore just before releasing processes.

cr-dump: fixup thread IP when inside rseq cs
    
    If we caught the process when it's inside rseq
    critical section we have to handle it properly.
    
    From the kernel side of view, if the process
    is executing inside the rseq cs and gets a signal,
    rseq critical section execution will be interrupted
    and after signal handler execution, we will proceed
    to rseq cs abort handler instead of continuing normal
    rseq cs execution (if RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL
    isn't set).
    
    When CRIU seizes processes that's the same thing as
    getting signal from the rseq point of view. So we need
    to fixup instruction pointer to rseq cs abort handler
    address.

rseq: fail dump if rseq is used but host doesn't support get_rseq_conf feature
    
    A lot of kernel versions lacks support for ptrace(PTRACE_GET_RSEQ_CONFIGURATION).
    But the userspace may be fresh (for instance containers with fresh Fedora runs
    on CentOS 7 host). Consider two scenarious:
    
    - kernel has no ptrace(PTRACE_GET_RSEQ_CONFIGURATION) support
    
    1. there is a process which use rseq => fail dump
    2. there is no process which use rseq => we can dump without any problems
    
    But how to determine if process use rseq or not without get_rseq_conf feature?
    Let's just try to do rseq registration from the parasite. If rseq is already
    registered then we'll got EBUSY error. If not we'll success in registration.

rseq: initial support
    
    Support basic rseq C/R scenario. Assume that:
    - there are no processes with IP inside the rseq critical section (CS)
    - kernel has ptrace(PTRACE_GET_RSEQ_CONFIGURATION) support
    
    On dump:
    1. use ptrace(PTRACE_GET_RSEQ_CONFIGURATION) to get
    struct rseq pointer, rseq size and signature from the kernel.
    2. save to the image
    
    On restore:
    1. get rseq ptr, size, signature from the image
    2. register it back using rseq() from the restorer parasite

Have done:
- [x] basic case support (just dump/restore `task_struct` rseq-related fields)
- [x] basic zdtm static test for amd64
- [x] support kernels where `rseq()` syscall is present but `ptrace(PTRACE_GET_RSEQ_CONFIGURATION)` is not.
- [x] fix cross-compile issues
- [x] prepare CI for fedora rawhide env + kernel >= 5.13
- [x] fix an issue when CRIU was compiled against new glibc (with rseq support)
- [x] fix COMPAT segfault
- [x] support "transitional" states
- [x] support for `RSEQ_CS_*` flags
- [x] write transition test

Fixes https://github.com/checkpoint-restore/criu/issues/1696

https://criu.org/Rseq


adrianreber:
You can run tests in CI with a kernel >= 5.13 if you use our KVM based setup which we use for vdso=0 testing. If you start a rawhide container on the Fedora Vagrant VM you should be able to get latest glibc with a new enough kernel.

mihalicyn:
Hehe, great. We have CentOS 8 working. That's good news because it's the corner case when we have `rseq()` syscall but lacks `ptrace(PTRACE_GET_RSEQ_CONFIGURATION)`. Let's look on fedora novdso.

mihalicyn:
> You can run tests in CI with a kernel >= 5.13 if you use our KVM based setup which we use for vdso=0 testing. If you start a rawhide container on the Fedora Vagrant VM you should be able to get latest glibc with a new enough kernel.

Yes, thanks for pointing it out. I've taken a look at `scripts/ci/vagrant.sh` it looks like for our needs can just adjust `FEDORA_VERSION` / `FEDORA_BOX_VERSION` variables. But unfortunately, on https://app.vagrantup.com/fedora repository there is no image for fedora-rawhide. :(

adrianreber:
> > You can run tests in CI with a kernel >= 5.13 if you use our KVM based setup which we use for vdso=0 testing. If you start a rawhide container on the Fedora Vagrant VM you should be able to get latest glibc with a new enough kernel.
> 
> Yes, thanks for pointing it out. I've taken a look at `scripts/ci/vagrant.sh` it looks like for our needs can just adjust `FEDORA_VERSION` / `FEDORA_BOX_VERSION` variables. But unfortunately, on https://app.vagrantup.com/fedora repository there is no image for fedora-rawhide. :(

Just run a rawhide container via Podman on the vagrant VM. You can basically re-use our rawhide on GitHub Actions setup in the VM and you should have everything you need.

mihalicyn:
> > > You can run tests in CI with a kernel >= 5.13 if you use our KVM based setup which we use for vdso=0 testing. If you start a rawhide container on the Fedora Vagrant VM you should be able to get latest glibc with a new enough kernel.
> > 
> > 
> > Yes, thanks for pointing it out. I've taken a look at `scripts/ci/vagrant.sh` it looks like for our needs can just adjust `FEDORA_VERSION` / `FEDORA_BOX_VERSION` variables. But unfortunately, on https://app.vagrantup.com/fedora repository there is no image for fedora-rawhide. :(
> 
> Just run a rawhide container via Podman on the vagrant VM. You can basically re-use our rawhide on GitHub Actions setup in the VM and you should have everything you need.

thanks, Adrian ;) Will try!

mihalicyn:
Great news! We have Fedora Rawhide CRIU fully functional. :)

figiel:
Hi, I don't see here support for case when the dumped thread is exactly in an RSEQ critical section.
Maybe it's not needed on current CRIU/vanilla kernel but the behavior I was observing when working on this was that the RSEQ abort was triggered only after the checkpointed thread was scheduled and actually executed some instructions. I.e. just stopping a thread with ptrace or sigstop doesn't cause the thread to jump to RSEQ abort if it's running a critical section.
It means that CRIU captured the instruction pointer inside critical section and restored it in such way. Problem was that this way the RSEQ invariants weren't preserved for this particular execution just after the thread was restored. 
The restored thread continued running through cs without kernel protection. This I suppose was because by the time the execution was resumed the rseq_cs pointer was long gone (set to null by kernel e.g. upon return from rseq()).
This can be worked around by interpreting contents of rseq_cs (before rseq() on restore side), checking that rip is within the cs, and fixing up rip to the address of abort handler.

criu/util.c
```diff
@@ -655,6 +655,63 @@ int cr_system_userns(int in, int out, int err, char *cmd, char *const argv[], un
...
+	if (socketpair(PF_LOCAL, SOCK_SEQPACKET, 0, sk_pair)) {
+		pr_perror("socketpair");
+		return -1;
+	}
+
+	pid = fork();
+	if (pid < 0) {
+		pr_perror("fork");
+		return -1;
+	} else if (pid == 0) {
```
rst0git, line 673: `else` is not necessary here. `return -1` would prevent the execution of subsequent lines.

criu/pie/parasite.c
```diff
@@ -313,6 +322,101 @@ static int dump_creds(struct parasite_dump_creds *args)
...
+		if (ret) {
+			pr_err("\tfailed to unregister sys_rseq(%lx, %lx, %x, %x) = %d\n", rseq_abi_pointer,
+			       rseq_abi_size, RSEQ_FLAG_UNREGISTER, rseq_signature, ret);
+
+			ret = -1;
+			goto out;
+		}
+
+		pr_info("\tsys_rseq succeed, let's unregister it back... ok Error\n");
+		pr_info("\trseq is non-initialized in the victim Error\n");
```
rst0git, line 410: I'm not sure how useful these messages would be in a log file.

mihalicyn: These is temporary logs. In the final version, I will remove that. ;)

mihalicyn:
@figiel thanks for looking at this! Yep, I have planned to support the transition scenario (it's in my TODO list at the first message of this topic ;)). I'm working on that right now.

rst0git:
> Great news! We have Fedora Rawhide CRIU fully functional. :)

As a temporary solution, I have applied the patches from this PR in the Fedora 36 / 37 packages.
https://github.com/containers/podman/issues/12949
https://github.com/containers/podman/issues/13283

compel/include/uapi/ptrace.h
```diff
@@ -66,14 +66,14 @@ typedef struct {
...
-	__u32 flags;
-	__u32 pad;
+#define PTRACE_GET_RSEQ_CONFIGURATION 0x420f
+
+struct __ptrace_rseq_configuration {
+	uint64_t rseq_abi_pointer;
+	uint32_t rseq_abi_size;
+	uint32_t signature;
+	uint32_t flags;
+	uint32_t pad;
```
rst0git, line 76: Would it be possible to move this change in the previous commit "cr-check: Add ptrace rseq conf dump feature", i.e., the commit introducing `ptrace_rseq_configuration`?

criu/cr-dump.c
```diff
@@ -45,6 +45,7 @@
 #include "proc_parse.h"
 #include "parasite.h"
 #include "parasite-syscall.h"
+#include <compel/ptrace.h>
```
rst0git, line 48:
```suggestion
#include "compel/ptrace.h"
```

criu/cr-dump.c
```diff
@@ -1003,6 +1034,68 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+
+	if (rseq.flags != 0) {
+		pr_err("something wrong with ptrace(PTRACE_GET_RSEQ_CONFIGURATION, %d) flags = 0x%x\n", tid,
+		       rseq.flags);
+		return -1;
+	}
+
+	pr_err("Dump rseq of %d: ptr = 0x%lx sign = 0x%x\n", tid, (unsigned long)rseq.rseq_abi_pointer, rseq.signature);
+
+	rseqe = xmalloc(sizeof(*rseqe));
```
rst0git, line 1074: In case of error, should we call `xfree` somewhere?

criu/pie/restorer.c
```diff
@@ -425,6 +425,27 @@ static int restore_signals(siginfo_t *ptr, int nr, bool group)
 	return 0;
 }
 
+static int restore_rseq(struct rst_rseq_param *rseq)
+{
+	int ret;
+
+	if (!rseq->rseq_abi_pointer) {
+		pr_debug("rseq: nothing to restore\n");
```
rst0git, line 433: Should we handle the case when rseq is not supported as well?

test/zdtm/static/Makefile
```diff
@@ -61,6 +61,7 @@ TST_NOFILE	:=				\
 		pthread02			\
 		pthread_timers			\
 		pthread_timers_h		\
+		rseq				\
```
rst0git, line 64:
```suggestion
		rseq00				\
```

criu/clone-noasan.c
```diff
@@ -34,16 +40,52 @@
...
+{
+	struct call_fn_args *a = arg;
+	unregister_glibc_rseq();
+	return a->fn(a->arg);
+}
+
 int clone_noasan(int (*fn)(void *), int flags, void *arg)
 {
 	void *stack_ptr = (void *)round_down((unsigned long)&stack_ptr - 1024, 16);
+	struct call_fn_args a = {
```
rst0git, line 78: Would it be possible to use a more descriptive variable name?

scripts/ci/run-ci-tests.sh
```diff
@@ -238,12 +238,7 @@ if [ -n "$TRAVIS" ] || [ -n "$CIRCLECI" ]; then
        # Error (criu/tty.c:1014): tty: Don't have tty to inherit session from, aborting
        make -C test/others/shell-job/ run
 fi
-
-# FIXME: rpc tests fail even with set glibc tunable
-# https://github.com/checkpoint-restore/criu/issues/1696
-if [ "$GLIBC_TUNABLES" != "glibc.pthread.rseq=0" ]; then
-	make -C test/others/rpc/ run
-fi
+make -C test/others/rpc/ run
```
rst0git, line 246: We also need to revert https://github.com/checkpoint-restore/criu/commit/f008f740411156b7309219c7e052cb8bc24d5aae

test/zdtm/static/rseq.c
```diff
@@ -0,0 +1,190 @@
...
+#ifndef __NR_rseq
+#define __NR_rseq 334
+#endif
+/* EOF */
+
+static __thread volatile struct rseq __rseq_abi;
+
+#define RSEQ_SIG 0x53053053
+#define TESTRSEQ 1
+#ifdef TESTRSEQ
```
rst0git, line 84: This looks odd. Isn't `TESTRSEQ` always defined in this case?

test/zdtm/static/rseq.c
```diff
@@ -0,0 +1,190 @@
...
+	else
+		fail();
+	//fail();
+	//unregister_thread();
+	//(void)unregister_thread;
+
+	return 0;
+}
+
+#else
```
rst0git, line 224: It is not very clear what this `#else` is for. Would it be possible to add a comment? For example, something like:
```C
#else  /* #if defined(__x86_64__) */
```

test/zdtm/static/rseq.c
```diff
@@ -0,0 +1,190 @@
...
+
+int main(int argc, char *argv[])
+{
+	test_init(argc, argv);
+	skip("Unsupported arch");
+	return 0;
+}
+
+#endif
```
rst0git, line 233:
```suggestion
#endif  /* #if defined(__x86_64__) */
```

mihalicyn:
- Arch Linux Test / build (pull_request)
- Fedora Rawhide Test / build (pull_request)
these checks fails with
```
2022-02-22T12:29:48.0017695Z b'(00.013160) Error (criu/cr-dump.c:203): 58 has rseq but kernel lacks get_rseq_conf feature'
```

Unfortunately, we can't overcome that. Too new Glibc and too old kernel = problems with the rseq. :(

adrianreber:
> * Arch Linux Test / build (pull_request)
> 
>     * Fedora Rawhide Test / build (pull_request)
>       these checks fails with
> 
> 
> ```
> 2022-02-22T12:29:48.0017695Z b'(00.013160) Error (criu/cr-dump.c:203): 58 has rseq but kernel lacks get_rseq_conf feature'
> ```
> 
> Unfortunately, we can't overcome that. Too new Glibc plus too old kernel = problems with the rseq. :(

I do not see this as a problem. We have your rseq feature tested on Fedora Vagrant with Rawhide as user space. On older kernels there is not much we can do. Currently GitHub Actions seems to run on 5.11.x. There should be a new Ubuntu LTS release (22.04) in a couple of months and probably (hopefully) also available in GitHub Actions. Then we can test it, but as long as there a kernel older than 5.13 there is nothing we can really do.

mihalicyn:
>We have your rseq feature tested on Fedora Vagrant with Rawhide as user space

Wow. Sounds cool. Is it stable? No crashes/failures?

adrianreber:
> >We have your rseq feature tested on Fedora Vagrant with Rawhide as user space
> 
> Wow. Sounds cool. Is it stable? No crashes/failures?

Ah, sorry. I meant your pull request. This pull request has tests with a new kernel and latest glibc thanks to your patches extending the Fedora Vagrant based tests. I just meant that we have your rseq changes covered by your tests.

scripts/ci/run-ci-tests.sh
mihalicyn: thanks!

adrianreber:
> I think all other activities like writing more tests for other architectures and so on can be addressed in another PRs. What to you think?

Sounds good to me.

mihalicyn:
Huh... now we have full support for all rseq features. But some code clean-up and review are still required here.
I'm sure that I've some memory leaks and not fully safe error paths in my code.

adrianreber:
From my point of view you should not revert the two commits that add "ENV GLIBC_TUNABLES=glibc.pthread.rseq=0" to the Dockerfiles. As long as we have to use older kernels the tests should still succeed.

criu/cr-restore.c
```diff
@@ -778,6 +779,7 @@ static int open_cores(int pid, CoreEntry *leader_core)
 {
 	int i, tpid;
 	CoreEntry **cores = NULL;
+	//RseqEntry *rseqs;
```
Snorch, line 782: Leftover?

Snorch:
```
[snorch@fedora criu]$ git diff 63a9750fa897316e62b69ed987cf34b58ff210a2 dbbddeca87b6332a25b1cf2d83c3a52d95bc1a5b | grep start_ip
+	return addr >= rseq_cs->start_ip && addr < rseq_cs->start_ip + rseq_cs->post_commit_offset;
+	if (!rseq_cs->start_ip)
+	pr_info("fixup_thread_rseq for %d: rseq_cs start_ip = %llx abort_ip = %llx post_commit_offset = %llx flags = %x version = %x; IP = %lx\n",
+		tid, rseq_cs->start_ip, rseq_cs->abort_ip, rseq_cs->post_commit_offset, rseq_cs->flags,
+	__u64 start_ip;
+	/* Offset from start_ip. */
+		/* start_ip, post_commit_ip, abort_ip */
```
I don't see rseq_cs->start_ip set before it is used in "cr-dump: fixup thread IP when inside rseq cs" is it intentional? Sorry if I miss something.

scripts/ci/run-ci-tests.sh
```diff
@@ -199,6 +199,11 @@ if [ "${STREAM_TEST}" = "1" ]; then
 	exit 0
 fi
 
+# print some useful debug info
+cat /proc/self/status
+ls -la /proc/self/ns
+cat /proc/self/cgroup
+
```
adrianreber, line 206: If we really want to keep this debug information it should be moved to the `print_env` section above.

scripts/ci/vagrant.sh
```diff
@@ -56,4 +56,25 @@ fedora-no-vdso() {
 	ssh default 'cd /vagrant/criu/test; sudo ./zdtm.py run -t zdtm/transition/pidfd_store_sk --rpc --pre 2'
 }
 
+fedora-rawhide() {
+	#ssh default sudo grubby --update-kernel ALL --args="selinux=0 systemd.unified_cgroup_hierarchy=0"
+	ssh default sudo grubby --update-kernel ALL
```
adrianreber, line 61: Probably not necessary.

scripts/ci/vagrant.sh
```diff
@@ -56,4 +56,25 @@ fedora-no-vdso() {
...
+	#
+	# Workaround the problem:
+	# error running container: error from /usr/bin/crun creating container for [...]: sd-bus call: Transport endpoint is not connected
+	# Let's just use runc instead of crun
+	# see also https://github.com/kata-containers/tests/issues/4283
+	#
+	ssh default 'sudo dnf remove -y crun || true'
+	ssh default sudo dnf install -y podman runc
+	vagrant reload
```
adrianreber, line 70: No need to do `vagrant reload` if we do not need to reboot.

scripts/ci/vagrant.sh
```diff
@@ -56,4 +56,25 @@ fedora-no-vdso() {
...
+	#ssh default sudo setenforce 0
+	ssh default cat /proc/cmdline
+	ssh default ls -la /proc/self/ns
+	ssh default sudo cat /proc/self/status
+	ssh default sudo cat /proc/self/cgroup
+	#ssh default sudo capsh --print
```
adrianreber, line 76: Please remove commented out lines and debug help if not necessary.

adrianreber:
Just saw that GitHub Actions seems to use a 5.13 kernel which means the rawhide and archlinux container should now pass CI with your patches applied.

adrianreber:
@mihalicyn Any news on this PR? Fedora 36, Archlinux and Ubuntu 22.04 will have this enabled in the upcoming releases and it would be good to have it merged as soon as possible. There are still a couple of review comments and CI is failing all over the place.

Can you have a look at getting this ready for a 3.17 release?

criu/cr-restore.c
```diff
@@ -867,8 +882,15 @@ static int restore_one_alive_task(int pid, CoreEntry *core)
 {
 	unsigned args_len;
 	struct task_restore_args *ta;
+	RseqEntry *rseqs;
 	pr_info("Restoring resources\n");
 
+	rseqs = shmalloc(sizeof(*rseqs) * current->nr_threads);
+	if (!rseqs)
+		return -1;
```
rst0git, line 890: Would it be possible to add error message here?

criu/cr-restore.c
```diff
@@ -1962,6 +1984,44 @@ static int attach_to_tasks(bool root_seized)
...
+		int i;
+
+		if (!task_alive(item))
+			continue;
+
+		if (item->nr_threads == 1) {
+			item->threads[0].real = item->pid->real;
+		} else {
+			if (parse_threads(item->pid->real, &item->threads, &item->nr_threads))
+				return -1;
```
rst0git, line 2001: Would it be possible to add error message here?

criu/cr-restore.c
```diff
@@ -811,6 +813,19 @@ static int open_cores(int pid, CoreEntry *leader_core)
 		}
 	}
 
+
+	pr_err("item %lx\n", (uint64_t)current);
```
rst0git, line 817: Why do we print error message here?

criu/cr-dump.c
```diff
@@ -1098,34 +1098,43 @@ static int dump_thread_rseq(struct pstree_item *item, int i)
...
+	rseqe->rseq_abi_size = rseqc.rseq_abi_size;
+	rseqe->signature = rseqc.signature;
 
-	if (read_rseq_cs(tid, &rseq, rseq_cs))
+	if (read_rseq_cs(tid, &rseqc, rseq_cs, &rseq))
 		goto err;
 
+	rseqe->has_rseq_cs_pointer = true;
+	rseqe->rseq_cs_pointer = rseq.rseq_cs.ptr64;
+	pr_err("cs pointer %lx\n", rseqe->rseq_cs_pointer);
```
rst0git, line 1131: Is this error intentional?

rst0git:
@mihalicyn Commit 77dc404b5 (`cr-dump: handle rseq flags field`) breaks the CI tests
```
sudo ./zdtm.py run -t zdtm/static/epoll
Checking feature kcmp_epoll
kcmp_epoll is supported
userns is supported
=== Run 1/1 ================ zdtm/static/epoll
========================= Run zdtm/static/epoll in ns ==========================
 DEP       epoll.d
 CC        epoll.o
 LINK      epoll
Construct root for zdtm/static/epoll
Start test
./epoll --pidfile=epoll.pid --outfile=epoll.out
Run criu dump
Run criu restore
################# Test zdtm/static/epoll FAIL at CRIU restore ##################
02:14:15.373:     4: epoll 4 add 6 native
02:14:15.374:     4: epoll 4 add 999 dup'ed from 6
02:14:15.374:     4: epoll source 999 closed
02:14:15.374:     4: epoll 4 add 701 native
02:14:15.374:     4: epoll 4 add 999 dup'ed from 701
02:14:15.374:     4: epoll source 999 closed
...

```

mihalicyn:
> @mihalicyn Commit [77dc404](https://github.com/checkpoint-restore/criu/commit/77dc404b513a4d57ebf8f883f769607b244b93da) (`cr-dump: handle rseq flags field`) breaks the CI tests
> 
> ```
> sudo ./zdtm.py run -t zdtm/static/epoll
> Checking feature kcmp_epoll
> kcmp_epoll is supported
> userns is supported
> === Run 1/1 ================ zdtm/static/epoll
> ========================= Run zdtm/static/epoll in ns ==========================
>  DEP       epoll.d
>  CC        epoll.o
>  LINK      epoll
> Construct root for zdtm/static/epoll
> Start test
> ./epoll --pidfile=epoll.pid --outfile=epoll.out
> Run criu dump
> Run criu restore
> ################# Test zdtm/static/epoll FAIL at CRIU restore ##################
> Test output: ================================
> 02:14:15.373:     4: epoll 4 add 6 native
> 02:14:15.374:     4: epoll 4 add 999 dup'ed from 6
> 02:14:15.374:     4: epoll source 999 closed
> 02:14:15.374:     4: epoll 4 add 701 native
> 02:14:15.374:     4: epoll 4 add 999 dup'ed from 701
> 02:14:15.374:     4: epoll source 999 closed
> ...
> 
>  <<< ================================
> ##################################### FAIL #####################################
> ```

Thanks. Fixed

criu/util.c
mihalicyn: I'm sorry, couldn't you clarify this? `else` seems required here because we need to execute one part of the code in the parent process and the second one in the child process.

criu/pie/restorer.c
mihalicyn: Sure. Fixed!

criu/cr-dump.c
```diff
@@ -1003,6 +1034,228 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+		return -1;
+	}
+
+	if (!rseq->rseq_cs.ptr64)
+		return 0;
+
+	ret = ptrace_peek_area(tid, rseq_cs, decode_pointer(rseq->rseq_cs.ptr64), sizeof(struct rseq_cs));
+	if (ret) {
+		pr_err("ptrace_peek_area(%d, %lx, %lx, %lx): fail to read rseq_cs struct\n", tid,
+		       (unsigned long)rseq_cs, (unsigned long)rseq->rseq_cs.ptr64, sizeof(struct rseq_cs));
```
rst0git, line 1071: > error: format '%lx' expects argument of type 'long unsigned int', but argument 8 has type 'unsigned int' [-Werror=format=]

criu/cr-dump.c
```diff
@@ -1003,6 +1034,228 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+	 * We have two ways to access thread memory: from the parasite and using ptrace().
+	 * But it this case we can't use parasite, because if victim process returns to the
+	 * execution, on the kernel side __rseq_handle_notify_resume hook will be called,
+	 * then rseq_ip_fixup() -> clear_rseq_cs() and user space memory with struct rseq
+	 * will be cleared. So, let's use ptrace(PTRACE_PEEKDATA).
+	 */
+	ret = ptrace_peek_area(tid, rseq, decode_pointer(rseqc->rseq_abi_pointer), sizeof(struct rseq));
+	if (ret) {
+		pr_err("ptrace_peek_area(%d, %lx, %lx, %lx): fail to read rseq struct\n", tid, (unsigned long)rseq,
+		       (unsigned long)(rseqc->rseq_abi_pointer), sizeof(uint64_t));
```
rst0git, line 1061: > error: format '%lx' expects argument of type 'long unsigned int', but argument 8 has type 'unsigned int' [-Werror=format=]

criu/cr-restore.c
```diff
@@ -1962,6 +1979,47 @@ static int attach_to_tasks(bool root_seized)
...
+			struct rst_rseq *rseqe = rsti(item)->rseqe;
+
+			if (!rseqe)
+				return -1;
+
+			if (!rseqe[i].rseq_cs_pointer || !rseqe[i].rseq_abi_pointer)
+				continue;
+
+			if (ptrace_poke_area(pid, &rseqe[i].rseq_cs_pointer,
+					     (void *)(rseqe[i].rseq_abi_pointer + offsetof(struct rseq, rseq_cs)),
```
rst0git, line 2012: > error: cast to pointer from integer of different size [-Werror=int-to-pointer-cast]

criu/cr-dump.c
mihalicyn: Yep, I will fix it now :)

criu/clone-noasan.c
```diff
@@ -78,7 +114,9 @@ int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags, int exit_sig
 	c_args.set_tid = ptr_to_u64(&pid);
 	c_args.set_tid_size = 1;
 	pid = syscall(__NR_clone3, &c_args, sizeof(c_args));
-	if (pid == 0)
+	if (pid == 0) {
+		unregister_glibc_rseq();
```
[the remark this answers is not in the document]
mihalicyn: Thanks! We need to unregister the rseq once to prevent its inheritance.

criu/cr-dump.c
```diff
@@ -1003,6 +1034,229 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+	 * will be cleared. So, let's use ptrace(PTRACE_PEEKDATA).
+	 */
+	ret = ptrace_peek_area(tid, rseq, decode_pointer(rseqc->rseq_abi_pointer), sizeof(struct rseq));
+	if (ret) {
+		pr_err("ptrace_peek_area(%d, %lx, %lx, %lx): fail to read rseq struct\n", tid, (unsigned long)rseq,
+		       (unsigned long)(rseqc->rseq_abi_pointer), (unsigned long)sizeof(uint64_t));
+		return -1;
+	}
+
+	if (!rseq->rseq_cs.ptr64)
```
rst0git, line 1065:
```
criu/cr-dump.c:1065:27: error: request for member 'ptr64' in something not a structure or union
 1065 |         if (!rseq->rseq_cs.ptr64)
      |                           ^
```

criu/cr-dump.c
```diff
@@ -1003,6 +1034,229 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+		return 0;
+
+	ret = ptrace_peek_area(tid, rseq_cs, decode_pointer(rseq->rseq_cs.ptr64), sizeof(struct rseq_cs));
```
rst0git, line 1068: ditto

criu/cr-dump.c
mihalicyn: On which Linux distribution is this reproducible? It looks like a problem with `rseq.h` system header.

fweimer-rh: The `rseq_cs` member is no longer a union as of commit bfdf4e6208051ed7165b2e92035b4bf11f43eb63. It's a source-incompatible UAPI change.

mihalicyn: @fweimer-rh ah, thanks a lot!

mihalicyn: I've pushed a new version with fixes for that problem.

mihalicyn:
Oops. My "fixes" of UAPI breakage broken everything else. :D 
Okay...

mihalicyn:
Periodic non-stable failures:
1. https://github.com/checkpoint-restore/criu/pull/1706/checks?check_run_id=5839914050

criu/cr-dump.c
rst0git: Thanks for fixing it. I can confirm that it works now.

criu/util.c
rst0git: Sorry, I meant something like the following:
```C
pid = fork();
if (pid < 0) {
	pr_perror("fork");
	return -1;
}

if (pid == 0) {
...
}
```
This is just a code-style change.

mihalicyn:
I've reproduced the problem with Alpine Linux without CRIU (`--nocr` flag is set):
```
/criu # ./test/zdtm.py run -t zdtm/transition/rseq01 --sbs --nocr --ignore-taint
The kernel is tainted: '12800\n'
Checking feature get_rseq_conf
get_rseq_conf is supported
userns is supported
=== Run 1/1 ================ zdtm/transition/rseq01
======================= Run zdtm/transition/rseq01 in h ========================
Start test
./rseq01 --pidfile=rseq01.pid --outfile=rseq01.out
Send the 15 signal to  48
  File "/criu/test/zdtm.py", line 2777, in <module>
    fork_zdtm()
  File "/criu/test/zdtm.py", line 2768, in fork_zdtm
  File "/criu/test/zdtm.py", line 1909, in do_run_test
    t.stop()
  File "/criu/test/zdtm.py", line 538, in stop
    self.kill(signal.SIGTERM)
  File "/criu/test/zdtm.py", line 507, in kill
    os.kill(int(self.__pid), sig)
ProcessLookupError: [Errno 3] No such process
```

```
[ 8531.674103] rseq01[190757]: segfault at 7fff0191d000 ip 00007f81db371953 sp 00007fff01919940 error 6 in ld-musl-x86_64.so.1[7f81db33b000+48000]
[ 8531.674111] Code: 7c 24 4e 4d 89 cf 66 8b 44 24 4e 80 cc 0c 66 89 44 24 4c d9 c0 d9 6c 24 4c df 7c 24 30 d9 6c 24 4e 49 83 c7 04 48 8b 44 24 30 <41> 89 47 fc 89 c0 48 89 44 24 30 df 6c 24 30 de e9 d8 0d 3a 4f 04
```

criu/util.c
mihalicyn: @rst0git thanks. Fixed!

mihalicyn:
Podman/Docker failures are not related to this patchset.

let's vote guys :)

rst0git (approved):
LGTM

scripts/ci/Makefile
```diff
@@ -18,7 +18,7 @@ export UNAME
 CONTAINER_RUNTIME := docker
 export CONTAINER_RUNTIME
 
-alpine: ZDTM_OPTS=-x zdtm/static/binfmt_misc -x zdtm/static/sched_policy00
+alpine: ZDTM_OPTS=-x zdtm/static/binfmt_misc -x zdtm/static/sched_policy00 -x zdtm/transition/rseq01
```
rst0git, line 21: @mihalicyn Could you add `FIXME` comment as well?

mihalicyn: great idea. Will do.

mihalicyn:
Okay,

https://api.cirrus-ci.com/v1/task/5808745342042112/logs/build.log
```
+ ./test/zdtm.py run -t zdtm/static/inotify_irmap --fault 128 --pre 2 -f uns
mem_dirty_track is supported
userns is supported
=== Run 1/1 ================ zdtm/static/inotify_irmap
===================== Run zdtm/static/inotify_irmap in uns =====================
Construct root for zdtm/static/inotify_irmap
Skipping /dev/rtc at root
Start test
Test is SUID
Running zdtm/static/inotify_irmap.hook(--post-start)
./inotify_irmap --pidfile=inotify_irmap.pid --outfile=inotify_irmap.out
Run criu pre-dump
Forcing 128 fault
=[log]=> dump/zdtm/static/inotify_irmap/63/1/pre-dump.log
*************
b"(00.012295) Error (criu/irmap.c:104): irmap: Can't stat /no-such-path: No such file or directory"
b"(00.012297) Error (criu/fsnotify.c:284): fsnotify: \tCan't dump that handle"
b'(00.012303) Error (criu/irmap.c:374): irmap: Failed to resolve 1e:56866' <======== Important
b'(00.012322) Writing image inventory (version 1)'
b'(00.012395) Error (criu/cr-dump.c:1851): Pre-dumping FAILED.'
############# Test zdtm/static/inotify_irmap FAIL at CRIU pre-dump #############
```

```
int irmap_predump_run(void)
{
...
	for (ip = predump_queue; ip; ip = ip->next) {
		pr_debug("\tchecking %x:%lx\n", ip->dev, ip->ino);
		ret = check_open_handle(ip->dev, ip->ino, &ip->fh);
		if (ret) {
			pr_err("Failed to resolve %x:%lx\n", ip->dev, ip->ino); <== this one
			break;
		}
```

```
############### Mounted file systems ###############
695 664 0:30 /root/var/lib/containers/storage/vfs/dir/b5c7b171dee24eec67e4344e0de4ed813b233ffef4765360a4f04d7f77b1d50f / rw,relatime master:1 - btrfs /dev/vda5 rw,seclabel,compress=zstd:1,space_cache,subvolid=256,subvol=/root
696 695 0:56 / /sys rw,nosuid,nodev,noexec,relatime - sysfs sysfs rw,seclabel
698 695 0:57 / /run rw,nosuid,nodev,relatime - tmpfs tmpfs rw,context="system_u:object_r:container_file_t:s0:c1022,c1023",inode64
697 695 0:58 / /proc rw,nosuid,nodev,noexec,relatime - proc proc rw
699 695 0:59 / /dev rw,nosuid - tmpfs tmpfs rw,context="system_u:object_r:container_file_t:s0:c1022,c1023",size=65536k,mode=755,inode64
700 698 0:26 /containers/storage/vfs-containers/e78818a6301661b60cf0fafef61a9a272be793a4c5770afa53dfb463c46870be/userdata/run/secrets /run/secrets rw,nosuid,nodev - tmpfs tmpfs rw,seclabel,size=801644k,nr_inodes=819200,mode=755,inode64
701 699 0:55 / /dev/mqueue rw,nosuid,nodev,noexec,relatime - mqueue mqueue rw,seclabel
702 699 0:60 / /dev/pts rw,nosuid,noexec,relatime - devpts devpts rw,context="system_u:object_r:container_file_t:s0:c1022,c1023",gid=5,mode=620,ptmxmode=666
703 695 0:30 /root/usr/lib/modules /usr/lib/modules rw,relatime - btrfs /dev/vda5 rw,seclabel,compress=zstd:1,space_cache,subvolid=256,subvol=/root
704 695 0:26 /containers/storage/vfs-containers/e78818a6301661b60cf0fafef61a9a272be793a4c5770afa53dfb463c46870be/userdata/resolv.conf /etc/resolv.conf rw,nosuid,nodev - tmpfs tmpfs rw,seclabel,size=801644k,nr_inodes=819200,mode=755,inode64
705 695 0:26 /containers/storage/vfs-containers/e78818a6301661b60cf0fafef61a9a272be793a4c5770afa53dfb463c46870be/userdata/hosts /etc/hosts rw,nosuid,nodev - tmpfs tmpfs rw,seclabel,size=801644k,nr_inodes=819200,mode=755,inode64
706 699 0:54 / /dev/shm rw,nosuid,nodev,noexec,relatime - tmpfs shm rw,context="system_u:object_r:container_file_t:s0:c1022,c1023",size=64000k,inode64
707 695 0:26 /containers/storage/vfs-containers/e78818a6301661b60cf0fafef61a9a272be793a4c5770afa53dfb463c46870be/userdata/hostname /etc/hostname rw,nosuid,nodev - tmpfs tmpfs rw,seclabel,size=801644k,nr_inodes=819200,mode=755,inode64
708 698 0:26 /containers/storage/vfs-containers/e78818a6301661b60cf0fafef61a9a272be793a4c5770afa53dfb463c46870be/userdata/.containerenv /run/.containerenv rw,nosuid,nodev - tmpfs tmpfs rw,seclabel,size=801644k,nr_inodes=819200,mode=755,inode64
709 696 0:27 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime - cgroup2 cgroup rw,seclabel
665 699 0:60 /0 /dev/console rw,nosuid,noexec,relatime - devpts devpts rw,context="system_u:object_r:container_file_t:s0:c1022,c1023",gid=5,mode=620,ptmxmode=666
```

`sdev = 30` is `btrfs`

mihalicyn:
Non-related, but just to notice:
```
=== Run 142/435 =====----------- zdtm/static/sched_policy00
Skipping zdtm/static/vfork00 (manual run only)
userns is supported
timens isn't supported on 5.4.0-1021-gcp
===================== Run zdtm/static/pthread_timers in ns =====================
Start test
./pthread_timers --pidfile=pthread_timers.pid --outfile=pthread_timers.out
Run criu dump
=[log]=> dump/zdtm/static/pthread_timers/66/1/dump.log
b'(00.001260) seccomp: Collected tid_real 70 mode 0'
b"(00.001285) \tSeizing 70's 71 thread"
b'(00.001343) seccomp: Collected tid_real 71 mode 0'
b"(00.001348) \tSeizing 70's 74 thread"
b'(00.001358) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 74 (Operation not permitted)'
b"(00.001380) \tSeizing 70's 74 thread"
b'(00.001385) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 74 (Operation not permitted)'
b"(00.001398) \tSeizing 70's 74 thread"
b'(00.001403) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 74 (Operation not permitted)'
b"(00.001415) \tSeizing 70's 74 thread"
b'(00.001420) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 74 (Operation not permitted)'
b"(00.001432) \tSeizing 70's 74 thread"
b'(00.001436) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 74 (Operation not permitted)'
b"(00.001448) \tSeizing 70's 74 thread"
b'(00.001452) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 74 (Operation not permitted)'
b'(00.001471) Unlock network'
b'(00.001476) Unfreezing tasks into 1'
b'(00.001479) \tUnseizing 66 into 1'
b'(00.001498) \tUnseizing 70 into 1'
b'(00.001532) Error (criu/cr-dump.c:2051): Dumping FAILED.'
############## Test zdtm/static/pthread_timers FAIL at CRIU dump ###############
```

mihalicyn:
@figiel Piotr, сheck it out when time permits, please

criu/arch/aarch64/include/asm/thread_pointer.h
```diff
@@ -0,0 +1,27 @@
...
+   You should have received a copy of the GNU Lesser General Public
+   License along with the GNU C Library.  If not, see
+   <https://www.gnu.org/licenses/>.  */
+
+#ifndef _SYS_THREAD_POINTER_H
+#define _SYS_THREAD_POINTER_H
+
+static inline void *__criu_thread_pointer(void)
+{
+	return __builtin_thread_pointer();
```
avagin, line 24: could you write the description for this patch? You need to explain why we need __criu_thread_pointer and why it is safe to copy this code from glibc. I mean why it will work with other libc libraries like musl.

mihalicyn:
I've played with Alpine Linux and `transition/rseq01` test. It crashes in `fmp_fp` function from `src/stdio/vfprintf.c:274`.
At the first glance, it looks like a bug in the implementation of `%f` formatter. I've managed to dump the floating-point number on which the issue occurred. However, later it turned out that nothing depends on the particular number...

From gdb session:
```
Thread 4.1 "rseq01" hit Breakpoint 1, test_msg (format=0x55555555a0d0 "exit %f\n") at msg.c:39
39	{
(gdb) bt -full
#0  test_msg (format=0x55555555a0d0 "exit %f\n") at msg.c:39
        arg = {{gp_offset = 8, fp_offset = 48, overflow_arg_area = 0x7fffffffeb80, reg_save_area = 0x7fffffffeac0}}
        off = 30
        buf = "18:55:49.984:  7801: exit inf\n\000", '\060' <repeats 12 times>, ".000000\n", '\000' <repeats 541 times>...
        _errno = 2
        tv = {tv_sec = 1649530549, tv_usec = 984638}
        tm = 0x7ffff7ffd4c0 <tm>
        __func__ = "test_msg"
#1  0x00005555555568f9 in rseq_addv (v=0x55555555de08, count=2, cpu=3) at rseq01.c:181
        a = 10000000000000000
        b = 10000000000000000
        rseq_cs1 = 93824992268544
        rseq_cs2 = 93824992268544
        buf = "\000\200\340\067y\303AC"
#2  0x0000555555556a5a in main (argc=3, argv=0x7fffffffec68) at rseq01.c:257
        cpu = 3
        ret = 0
        cpu_data = 0x55555555ddf0
        nr_cpus = 12
(gdb) c
Continuing.

Thread 4.1 "rseq01" received signal SIGSEGV, Segmentation fault.
0x00007ffff7fb2953 in fmt_fp (f=0x7fffffffe168, y=inf, w=0, p=6, fl=0, t=102) at src/stdio/vfprintf.c:274
274	src/stdio/vfprintf.c: No such file or directory.
(gdb) bt -full
#0  0x00007ffff7fb2953 in fmt_fp (f=0x7fffffffe168, y=inf, w=0, p=6, fl=0, t=102) at src/stdio/vfprintf.c:274
        big = {0 <repeats 1253 times>, 4160423727, 32767, 11, 0, 4294957112, 32767, 0, 0, 11, 0, 11, 0, 4160436391, 32767, 0, 0, 21, 11, 4294956720, 32767, 4294956744, 32767, 11, 0, 4294956864, 32767, 0, 
          2147483637, 4160720122, 32767, 0, 0, 0, 0, 1431675204, 21845, 0, 0, 1431675405, 21845, 0 <repeats 16 times>, 4160445164, 32767, 0, 0, 4294957112, 32767, 21, 0, 4294957094, 32767, 4294956720, 32767, 
          4160436995, 32767, 0, 0, 1431675200, 21845, 0, 32767, 4294956744, 32767, 40, 48, 4294957584, 32767, 4294957392, 32767, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1902474098, 1865298224, 1764652149, 1869770862, 
          1936028263, 115, 0 <repeats 56 times>, 1431660847, 21845, 4294961744, 32767, 4294962280, 32767, 4294962312, 32767, 1431685952, 21845, 0, 0, 4160445402, 32767, 0, 0, 4294957637, 32767, 4074, 
          0 <repeats 19 times>, 4160445102, 32767, 0, 0, 4294957094, 32767, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 4294967295, 4294967295, 4294967295, 0, 4294957096, 32767, 0 <repeats 20 times>, 4160429082, 32767, 0, 
          0, 24, 48, 4294957584, 32767, 4294957392, 32767, 0, 0, 0, 0, 0, 0, 4160460589, 32767, 4160744764, 32767, 0, 0, 4294957616, 32767, 1217...}
        a = 0x7fffffffdd8c
        d = <optimized out>
        r = 0x7fffffffdd8c
        z = 0x7ffffffff000
        e2 = 25
        e = <optimized out>
        i = <optimized out>
        j = <optimized out>
        l = <optimized out>
        buf = "000000000", '\000' <repeats 15 times>
        s = <optimized out>
        prefix = 0x7ffff7ff8cdc "0X+0X 0X-0x+0x 0x"
        pl = 0
        ebuf0 = '\000' <repeats 11 times>
        ebuf = 0x7fffffffc1cb "000000000"
        estr = <optimized out>
#1  0x0000000000000000 in ?? ()
No symbol table info available.
```

It looks strange that gdb can't unwind stack properly when we are in the crash point. It looks like stack smashing. Right now I have no idea how it can be related to `ld-musl` and why we can't see the same with `Glibc`.

criu/cr-dump.c
```diff
@@ -1003,6 +1004,68 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+		return -1;
+	}
+
+	if (rseq.flags != 0) {
+		pr_err("something wrong with ptrace(PTRACE_GET_RSEQ_CONFIGURATION, %d) flags = 0x%x\n", tid,
+		       rseq.flags);
+		return -1;
+	}
+
+	pr_err("Dump rseq of %d: ptr = 0x%lx sign = 0x%x\n", tid, (unsigned long)rseq.rseq_abi_pointer, rseq.signature);
```
avagin: pr_info or pr_debug

mihalicyn: fixed

criu/arch/aarch64/include/asm/thread_pointer.h
mihalicyn: Have done.

criu/include/clone-noasan.h
```diff
@@ -1,6 +1,25 @@
...
+#if defined(RSEQ_SIG)
+static inline void unregister_glibc_rseq(void)
+{
+	struct criu_rseq *rseq = __criu_thread_pointer() + __rseq_offset;
+
+	/* let's mark rseq as failed to register */
+	rseq->cpu_id = RSEQ_CPU_ID_REGISTRATION_FAILED;
+
+	/* unregister rseq */
+	syscall(__NR_rseq, (void *)rseq, __rseq_size, 1, RSEQ_SIG);
```
rst0git, line 15: Would it be possible to replace `1` with `RSEQ_FLAG_UNREGISTER`?
```
$ grep RSEQ_FLAG_UNREGISTER /usr/include/linux/rseq.h 
	RSEQ_FLAG_UNREGISTER = (1 << 0),
```

test/zdtm/static/rseq00.c
```diff
@@ -0,0 +1,212 @@
...
+static inline void unregister_glibc_rseq(void)
+{
+	/* unregister rseq */
+	syscall(__NR_rseq, (void *)((char *)__criu_thread_pointer() + __rseq_offset), __rseq_size, 1, RSEQ_SIG);
+}
+#else
+static inline void unregister_glibc_rseq(void)
+{
+}
+#endif
```
rst0git, line 55: There are quite a few `#if` / `#endif` directives here. It would be helpful to add comments:
```C
#if defined(RSEQ_SIG)
...
#ifdef __x86_64__
...
#else
...
#endif   /* __x86_64__ */
...
#endif  /* defined(RSEQ_SIG) */
```

test/zdtm/transition/rseq01.c
```diff
@@ -154,9 +172,22 @@ static int rseq_addv(intptr_t *v, intptr_t count, int cpu)
 	);
 	/* clang-format on */
 	rseq_after_asm_goto();
+	test_msg("exit %lx %lx %f %f\n", rseq_cs1, rseq_cs2, a, b);
+	if (rseq_cs1 != rseq_cs2) {
+		/*
+		 * It means that we finished critical section
+		 * *normally* (haven't jumped to abort) but the kernel had cleaned up
+		 * rseq_ptr->rseq_cs before we left critical section
+		 * and CRIU wasn't restored it correctly.
```
rst0git, line 181: *and CRIU didn't restore it correctly

test/zdtm/transition/rseq01.c
```diff
@@ -178,21 +209,45 @@ int main(int argc, char *argv[])
...
 	}
+
 	register_thread();
 
+	/*
+	 * We want to test that RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL
+	 * is handled properly by CRIU, but that flag can be used
+	 * only with all another flags set.
+	 * Please, refer to
+	 * https://github.com/torvalds/linux/blob/master/kernel/rseq.c#L192
```
rst0git, line 220: GitHub supports "permalinks": https://github.com/torvalds/linux/blob/ce522ba9/kernel/rseq.c#L192

criu/cr-restore.c
```diff
@@ -2423,6 +2478,9 @@ static int restore_root_task(struct pstree_item *init)
 	if (restore_freezer_state())
 		pr_err("Unable to restore freezer state\n");
 
+	/* just before releasing threads we have to restore rseq_cs */
+	restore_rseq_cs();
```
rst0git, line 2482: Should we check the return value of `restore_rseq_cs()` here?

criu/cr-restore.c
```diff
@@ -1966,6 +1979,48 @@ static int attach_to_tasks(bool root_seized)
...
+				return -1;
+			}
+		}
+
+		for (i = 0; i < item->nr_threads; i++) {
+			pid_t pid = item->threads[i].real;
+			struct rst_rseq *rseqe = rsti(item)->rseqe;
+
+			if (!rseqe)
+				return -1;
```
rst0git, line 2006: Do we need to print an error message in this case?

mihalicyn:
```
   CC       criu/cgroup-props.o
criu/bpfmap.c: In function ‘bpfmap_open’:
criu/bpfmap.c:289:9: error: ‘bpf_create_map_xattr’ is deprecated: libbpf v0.7+: use bpf_map_create() instead [-Werror=deprecated-declarations]
  289 |         bpfmap_fd = bpf_create_map_xattr(&xattr);
      |         ^~~~~~~~~
In file included from criu/bpfmap.c:3:
/usr/include/bpf/bpf.h:83:16: note: declared here
   83 | LIBBPF_API int bpf_create_map_xattr(const struct bpf_create_map_attr *create_attr);
      |                ^~~~~~~~~~~~~~~~~~~~
make[2]: *** [/criu/scripts/nmk/scripts/build.mk:119: criu/bpfmap.o] Error 1
```

Will fix in another PR.

rst0git:
> Arch Linux Test fails with
> 
> ```
>    CC       criu/cgroup-props.o
> criu/bpfmap.c: In function ‘bpfmap_open’:
> criu/bpfmap.c:289:9: error: ‘bpf_create_map_xattr’ is deprecated: libbpf v0.7+: use bpf_map_create() instead [-Werror=deprecated-declarations]
>   289 |         bpfmap_fd = bpf_create_map_xattr(&xattr);
>       |         ^~~~~~~~~
> In file included from criu/bpfmap.c:3:
> /usr/include/bpf/bpf.h:83:16: note: declared here
>    83 | LIBBPF_API int bpf_create_map_xattr(const struct bpf_create_map_attr *create_attr);
>       |                ^~~~~~~~~~~~~~~~~~~~
> cc1: all warnings being treated as errors
> make[2]: *** [/criu/scripts/nmk/scripts/build.mk:119: criu/bpfmap.o] Error 1
> ```
> 
> Will fix in another PR.

I have a fix in https://github.com/checkpoint-restore/criu/pull/1738

mihalicyn:
> 

great! Let's reopen that PR?

figiel:
@mihalicyn OK, I'll try to review this tomorrow.

figiel (changes requested):

test/zdtm/static/rseq00.c
```diff
@@ -0,0 +1,212 @@
...
+		"cmpl %[cpu_id], %[current_cpu_id]\n\t"
+		"jnz 4f\n\t"
+		"addq %[count], %[v]\n\t"	/* final store */
+		"2:\n\t"
+		".pushsection __rseq_failure, \"ax\"\n\t"
+		/* Disassembler-friendly signature: nopl <sig>(%rip). */
+		".byte 0x0f, 0x1f, 0x05\n\t"
+		".long 0x53053053\n\t"	/* RSEQ_FLAGS */
+		"4:\n\t"
+		"jmp abort\n\t"
```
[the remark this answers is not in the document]
mihalicyn: good caught! Thanks

test/zdtm/static/rseq00.c
```diff
@@ -0,0 +1,212 @@
...
+
+	register_thread();
+
+	test_daemon();
+	test_waitsig();
+
+	check_thread();
+
+	cpu = RSEQ_ACCESS_ONCE(rseq_ptr->cpu_id_start);
+	ret = rseq_addv(&cpu_data[cpu], 2, cpu);
```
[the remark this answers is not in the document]
mihalicyn: yep. Agreed totally, we can just check that the rseq is registered after C/R here.

test/zdtm/static/rseq00.c
```diff
@@ -0,0 +1,212 @@
...
+
+	return 0;
+}
+
+#else /* #if defined(__x86_64__) */
+
+int main(int argc, char *argv[])
+{
+	test_init(argc, argv);
+	skip("Unsupported arch");
```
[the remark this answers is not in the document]
mihalicyn: No, we don't fail checkpointing in such cases. The rseq support code is not strongly dependent on the architecture. So, it looks safe not to fail C/R in this case.
I have a plan to write tests for all supported architectures later.

criu/cr-dump.c
```diff
@@ -187,6 +188,25 @@ static int dump_sched_info(int pid, ThreadCoreEntry *tc)
 	return 0;
 }
 
+static int check_thread_rseq(pid_t tid, const struct parasite_check_rseq *ti_rseq, bool has_tc_rseq_entry)
```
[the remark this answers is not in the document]
mihalicyn: Sure. Thanks!

criu/cr-dump.c
```diff
@@ -1003,6 +1034,228 @@ static int dump_task_signals(pid_t pid, struct pstree_item *item)
...
+		 * But unfortunately, it doesn't work properly, because the kernel does
+		 * clean up of rseq_cs field in the struct rseq (modifies userspace memory).
+		 * So, we need to preserve original value of (struct rseq)->rseq_cs field in the
+		 * image and restore it's value before releasing threads (see restore_rseq_cs()).
+		 *
+		 * It's worth to mention that we need to fixup IP in CoreEntry
+		 * (used when full dump/restore is performed) and also in
+		 * the parasite regs storage (used if --leave-running option is used,
+		 * or if dump error occurred and process execution is resumed).
+		 */
```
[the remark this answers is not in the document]
mihalicyn: >By doing the IP fixup we mimic kernel behavior on preemption (or actually on signal delivery).

Yes, we did IP fixup "by hands" because otherwise, CRIU will preserve an original IP and after the restoration, the process will continue execution from the same place where it was inside the rseq critical section.

>Why do we want to restore rseq_cs? Kernel would set it to NULL, I think we should too. Otherwise there may be a subtle change in behavior introduced by c/r.

The reason is that the kernel wouldn't set `(struct rseq)->rseq_cs` field to `NULL` if the rseq CS has `RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL` flag set in the normal scenario.

The problem here is that the kernel sets this pointer to `NULL` even when `RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL` is set when the process leaves CS. In our case that happens during the dump process once CRIU injects and runs the `parasite` from the victim process context. That's why we need to checkpoint and restore this pointer explicitly.

But as I can see I've forgotten to add the check about `RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL`:
```
	if (read_rseq_cs(tid, &rseqc, rseq_cs, &rseq))
		goto err;

	rseqe->has_rseq_cs_pointer = true;
	rseqe->rseq_cs_pointer = rseq.rseq_cs;

	/* we won't save rseq_cs to the image (only pointer),
	 * so let's combine flags from both struct rseq and struct rseq_cs
	 * (kernel does the same when interpreting RSEQ_CS_FLAG_*)
	 */
	rseq_cs->flags |= rseq.flags;
```

It means that I'm dumping `rseq_cs` pointer in all cases (not only if `RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL` is set). That does not cause any problem just because, anyway kernel will `NULL`ify this pointer if `RSEQ_CS_FLAG_NO_RESTART_ON_SIGNAL` isn't set. But anyway that's not tidy. Thanks a lot for pointing this out! I've fixed that.

test/zdtm/transition/rseq01.c
```diff
@@ -0,0 +1,267 @@
...
+			  RSEQ_CS_FLAG_NO_RESTART_ON_MIGRATE;
+#endif
+
+	test_daemon();
+
+	while (test_go()) {
+		cpu = RSEQ_ACCESS_ONCE(rseq_ptr->cpu_id_start);
+		ret = rseq_addv(&cpu_data[cpu], 2, cpu);
+#ifndef NOABORT
+		/* just ignore abort */
```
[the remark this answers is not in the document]
mihalicyn: yep, we do:
```
#ifndef NOABORT
		/* just ignore abort */
		ret = 0;
#else
		if (ret) {
			fail("Failed to increment per-cpu counter");
			break;
		}
#endif
```

but agree, the condition with two "no" should be changed to more readable :)

test/zdtm/transition/rseq01.c
```diff
@@ -0,0 +1,267 @@
...
+		".pushsection __rseq_table, \"aw\"\n\t"
+		".balign 32\n\t"
+		"cs_obj:\n\t"
+		/* version, flags */
+		".long 0, 0\n\t"
+		/* start_ip, post_commit_offset, abort_ip */
+		".quad 1f, (2f-1f), 4f\n\t"
+		".popsection\n\t"
+		"1:\n\t"
+		"leaq cs_obj(%%rip), %%rax\n\t"
```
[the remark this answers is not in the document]
mihalicyn: Fixed ;)

test/zdtm/static/rseq00.c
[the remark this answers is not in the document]
rst0git: > I have a plan to write tests for all supported architectures later.

Perhaps we could add a TODO comment somewhere?

mihalicyn:
Dear friends,
`zdtm/transition/rseq01` on Alpine Linux was fixed. Thanks to Mathieu @compudj for looking into this.
I've shot myself in the foot by corrupting the stack by an unpaired `fldl` instruction. That's really strange that we've got crashes only on musl-libc.

Andrei suggests using the simpler and better approach to make the rseq cs section execution time larger by using the `pause` instruction. I will do that in the later version but let's keep the current version with FPU instructions in memory of the hours spent in searching for mistake. :D

test/zdtm/static/rseq00.c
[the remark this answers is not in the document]
mihalicyn: I thought that we will use the "compiler magic" to make tests for all architectures sit in one test file (but plus one header file per architecture).

We can get some inspiration from here:
https://github.com/torvalds/linux/tree/master/tools/testing/selftests/rseq

test/zdtm/transition/rseq01.c
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	return 0;
+abort:
+	rseq_after_asm_goto();
+	test_msg("abort %lx %lx %f %f\n", rseq_cs1, rseq_cs2, a, b);
+	return -1;
+}
+
+int main(int argc, char *argv[])
```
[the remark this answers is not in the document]
mihalicyn: I've made some trivial checks like:
`for i in {1..100}; do sudo ./test/zdtm.py run -t zdtm/transition/rseq01 --ignore-taint; done > rseq01.log`
`cat rseq01.log | grep FAIL`
with positive result.

>Are those tests normally ran multiple times a day?

That's a good idea. We can organize that on our CRIU Jenkins. Usually, we test criu-dev and master branch on it.

criu/cr-restore.c
```diff
@@ -3012,6 +3013,25 @@ static int prep_rseq(struct rst_rseq_param *rseq, ThreadCoreEntry *tc)
...
 
+#if defined(RSEQ_SIG)
+static void prep_libc_rseq_info(struct rst_rseq_param *rseq)
+{
+	if (!kdat.has_rseq) {
+		rseq->rseq_abi_pointer = 0;
+		return;
+	}
+
+	rseq->rseq_abi_pointer = encode_pointer(__criu_thread_pointer() + __rseq_offset);
```
avagin, line 3089: musl can have a different offset, can't it?

mihalicyn: This code is under `defined(RSEQ_SIG)` so it is used only for Glibc.

mihalicyn: upd: I've changed the condition to `#if defined(__GLIBC__) && defined(RSEQ_SIG)` just to be more explicit.

mihalicyn: upd2: added a comment with `TODO` about other libc libraries.

mihalicyn:
JFYI: I'm working on the article https://criu.org/Rseq

criu/cr-restore.c
compudj: It's unclear to me how the various scenarios are handled. You can have:

- glibc with rseq support, running on a Linux kernel without RSEQ support (either because the kernel is < 4.18, or because CONFIG_RSEQ=n),
- glibc with rseq support runnng on a Linux kernel with RSEQ support,
- glibc without rseq support running on kernel with RSEQ support,
- glibc without rseq support running on kernel without RSEQ support.

And you can build your application with recent kernel headers that have RSEQ support, but end up running on a Linux kernel without rseq support (either older kernel or CONFIG_RSEQ=n).

In librseq, I use "dlsym" to lookup whether glibc has the __rseq_offset and __rseq_size symbols to figure out at runtime if the glibc has rseq suport. I can use the __rseq_size value to figure out if glibc could successfully register rseq with the kernel. Ref. https://github.com/compudj/librseq/blob/master/src/rseq.c#L109

mihalicyn: Hi Mathieu,

thanks a lot for your attention to this topic.

Let me try to cover some of your points:
1. glibc with rseq support, running on the Linux kernel without RSEQ support. It's covered by check at `prep_libc_rseq_info` function. We just don't need to unregister anything and we can't restore processes with rseq dumped on the host, where rseq is not supported.
2. glibc with rseq support running on a Linux kernel with RSEQ support. covered by the same function. It finds `struct rseq` location and then it will be unregistered in the `restorer` blob.
3. glibc without rseq support running on the kernel with RSEQ support. It's okay. We just don't need to unregister rseq and can restore it without any problems.
4. glibc without rseq support running on kernel without RSEQ support. It's also okay. We will not try to unregister rseq and can't restore processes dumped on the system where rseq was supported.

>And you can build your application with recent kernel headers that have RSEQ support, but end up running on a Linux kernel without rseq support (either older kernel or CONFIG_RSEQ=n).

Good caught. But this particular case is not a problem because the check about rseq support in the kernel is dynamic (`kdat.has_rseq`). But I can analytically continue your idea and put a new question `what if criu was build on the system with old glibc (without rseq) and then run on the system with the newer glibc (with rseq)`. In this case we will fail because our check about glibc rseq support is compile-time:
```
#if defined(__GLIBC__) && defined(RSEQ_SIG)
static void prep_libc_rseq_info(struct rst_rseq_param *rseq)
{
	if (!kdat.has_rseq) {
		rseq->rseq_abi_pointer = 0;
		return;
	}

	rseq->rseq_abi_pointer = encode_pointer(__criu_thread_pointer() + __rseq_offset);
	rseq->rseq_abi_size = __rseq_size;
	rseq->signature = RSEQ_SIG;
}
#else
static void prep_libc_rseq_info(struct rst_rseq_param *rseq)
{
	/*
	 * TODO: handle built-in rseq on other libc'ies like musl
	 * We can do that using get_rseq_conf kernel feature.
	 *
	 * For now we just assume that other libc libraries are
	 * not registering rseq by default.
	 */
	rseq->rseq_abi_pointer = 0;
}
#endif
```

Fortunately, in most cases criu is built on the same userspace where it runs.

>In librseq, I use "dlsym" to lookup whether glibc has the __rseq_offset and __rseq_size symbols to figure out at runtime if the glibc has rseq suport. I can use the __rseq_size value to figure out if glibc could successfully register rseq with the kernel.

That's an interesting approach. I can try to take this way of rseq detection and unregistration.

cc @avagin

mihalicyn: We can continue discussion here https://github.com/checkpoint-restore/criu/issues/1814

PR #1711, opened 2021-12-24
criu: Add `on-demand` option for task restore

Yangfisher1:
These patches include:

- An extension to support restoring processes with `sys_mmap` instead of `sys_preadv` to optimize cases when the restored process does not need all the checkpointed pages. Note that the restoration of anonymous memory areas is left unchanged. The resulting performance improvement can be 30.65%(1.808s vs. 1.254s) for the test `zdtm/static/mntns_ghost01`. Also, it accelerates other tests, e.g., `zdtm/static/file_locks00`(1.807s vs. 1.432s) and `zdtm/static/unlink_regular`(1.936s vs. 1.321s) on my machine.
- This extension can be optionally enabled (i.e., by default disabled) with `criu restore -D dir --on-demand`.
- Proper helper messages are added so that usage of the extension can be viewed via `criu -h`.
- The extension has passed all the tests, i.e., `./zdtm.py run -a`.

Signed-off-by: Yuhan Yang [Yangfisher1@sjtu.edu.cn](mailto:Yangfisher1@sjtu.edu.cn)


Yangfisher1:
Post an issue[#1712](https://github.com/checkpoint-restore/criu/issues/1712)

adrianreber:
Not sure why you opened #1712. We can have all discussions here.

For the missing syscalls: You just have to add them to the syscall tables.

Overall: Have a look at all the CI failures and try to fix them. It still fails at many different places.

Yangfisher1:
Most of them failed because when compiling, the file `criu/pie/restore-blob.h` would have a line ` #define restorer_sym__export_restore_task.cold 0x475b`, I don't know why this happened cause the file was auto-generated. What I've done is changing some codes in `criu/pie/restorer.c`, function `__export_restore_task `.

adrianreber:
It is a strange error. Never seen the that before. gcc 5.4.0 sounds really old (although our CentOS 7 test uses something even older). Maybe try it with a newer gcc and try to reproduce the error locally. I am not familiar with that part of the code, so that I cannot really help you. Hopefully someone else can help you figure out why the same symbol is generated with and without `.cold`.

The only idea that comes to my mind, is maybe it is related to something with the stack, because you add a couple of new variables. But that is just a wild uneducated guess.

Yangfisher1:
> The only idea that comes to my mind, is maybe it is related to something with the stack, because you add a couple of new variables. But that is just a wild uneducated guess.

Yes. Compiling with the original version won't produce the problem even with gcc9, so it might be something with the stack I guess.

Yangfisher1:
By the way, how can I add a `sys_pread` syscall in arm? I know that the syscall table is in `compel/arch/arm/plugins/std/syscalls/syscall.def`, and it has a syscall whose name is `sys_pread64`, but I don't know what's the meaning of `code64` and `code32`. Is it enough to just add a new line `pread` with same `code32` and `code64`?

Yangfisher1:
Adding a flag `-fno-reorder-blocks-and-partition` in `criu/Makefile` seems to be able to solve the cold symbol problem.

Yangfisher1:
`ci/circleci: test-local-gcc` failed at test `zdtm/static/pthread_timers in ns`. However, I passed the test on my own machine, so I checked out the log and found out that
```
=== Run 136/423 =====----------- zdtm/static/pthread_timers
userns is supported
timens isn't supported on 5.4.0-1021-gcp
===================== Run zdtm/static/pthread_timers in ns =====================
Start test
./pthread_timers --pidfile=pthread_timers.pid --outfile=pthread_timers.out
Run criu dump
=[log]=> dump/zdtm/static/pthread_timers/232/1/dump.log
b'(00.001254) seccomp: Collected tid_real 236 mode 0'
b"(00.001285) \tSeizing 236's 237 thread"
b'(00.001381) seccomp: Collected tid_real 237 mode 0'
b"(00.001389) \tSeizing 236's 238 thread"
b'(00.001398) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 238 (Operation not permitted)'
b"(00.001424) \tSeizing 236's 238 thread"
b'(00.001432) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 238 (Operation not permitted)'
b"(00.001448) \tSeizing 236's 238 thread"
b'(00.001454) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 238 (Operation not permitted)'
b"(00.001470) \tSeizing 236's 238 thread"
b'(00.001476) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 238 (Operation not permitted)'
b"(00.001490) \tSeizing 236's 238 thread"
b'(00.001496) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 238 (Operation not permitted)'
b"(00.001514) \tSeizing 236's 238 thread"
b'(00.001521) Warn  (compel/src/lib/infect.c:126): Unable to interrupt task: 238 (Operation not permitted)'
b'(00.001542) Unlock network'
b'(00.001548) Unfreezing tasks into 1'
b'(00.001552) \tUnseizing 232 into 1'
b'(00.001575) \tUnseizing 236 into 1'
b'(00.001603) Error (criu/cr-dump.c:1788): Dumping FAILED.'
############## Test zdtm/static/pthread_timers FAIL at CRIU dump ###############
Send the 9 signal to  232
Wait for zdtm/static/pthread_timers(232) to die for 0.100000
```
Considering that I use a Ubuntu 20.04 with a kernel 5.11.0-43-generic, I think it might be `timens isn't supported on 5.4.0-1021-gcp` that caused the test failed?

By the way, I also wonder why `ci/circleci: test-local-clang` failed. The error info is
```
make -r -R -f /home/circleci/criu/scripts/nmk/scripts/main.mk makefile=Makefile obj=criu/arch/x86 all
criu/Makefile.packages:36: Can not find some of the required libraries
criu/Makefile.packages:37: Make sure the following packages are installed
criu/Makefile.packages:38: RPM based distros: protobuf protobuf-c protobuf-c-devel protobuf-compiler protobuf-devel protobuf-python libnl3-devel libcap-devel python3-future
criu/Makefile.packages:39: DEB based distros: libprotobuf-dev libprotobuf-c-dev protobuf-c-compiler protobuf-compiler python3-protobuf python3-future libnl-3-dev libcap-dev
criu/Makefile.packages:40: To run tests the following packages are needed
criu/Makefile.packages:41: RPM based distros: libaio-devel python3-PyYAML
criu/Makefile.packages:42: DEB based distros: python3-yaml libaio-dev libaio-dev
criu/Makefile.packages:43: *** Compilation aborted.  Stop.
```

adrianreber:
Please do not include merge commits in your pull request. Use rebase to apply your patches on the latest `criu-dev` branch in your pull request. Also, please no fix up commits like the format fix. Just do the fix in the actual commit changing the code.

The `zdtm/static/pthread_timers` test failure is a flake and happens sometimes. We have not figured out why it happens. Rerunning CI will fix this CI failure.

The other error `criu/Makefile.packages:36: Can not find some of the required libraries` is not about packages, but is related to one of your changes. Clang is not happy with your changes: `error: unknown argument: '-fno-reorder-blocks-and-partition'
`

Overall please make sure the Author of each commit matches the `Signed-off-by` exactly. You are using sometimes multiple `Signed-off-by`, sometimes they are different. Sometimes the author uses a gmail.com address, sometimes not. Try to be consitent.

Yangfisher1:
> Please do not include merge commits in your pull request. Use rebase to apply your patches on the latest `criu-dev` branch in your pull request. Also, please no fix up commits like the format fix. Just do the fix in the actual commit changing the code.
> 
> The `zdtm/static/pthread_timers` test failure is a flake and happens sometimes. We have not figured out why it happens. Rerunning CI will fix this CI failure.
> 
> The other error `criu/Makefile.packages:36: Can not find some of the required libraries` is not about packages, but is related to one of your changes. Clang is not happy with your changes: `error: unknown argument: '-fno-reorder-blocks-and-partition' `
> 
> Overall please make sure the Author of each commit matches the `Signed-off-by` exactly. You are using sometimes multiple `Signed-off-by`, sometimes they are different. Sometimes the author uses a gmail.com address, sometimes not. Try to be consitent.

I'll try to fix these after I fix the `.cold` symbol problem, using flag `-fno-reorder-blocks-and-partition` works well for gcc, however clang doesn't have this flag. That's a pity, maybe I need more time to think about it. Thank you for your reply!

Yangfisher1:
Hmm, it's quite strange. Using clang to compile seemed fine at  `Compat Tests/ build(CLANG)`, however at `X86_64 CLANG Test / build` it showed that using the flag that clang doesn’t have again. But I did add a `ifeq` in `Makefile`
```
ifeq ($(HOSTCC), gcc)
	CFLAGS += -fno-reorder-blocks-and-partition
endif
```
So, is there anything special for `X86_64 CLANG Test / build`?

Yangfisher1:
> Hmm, it's quite strange. Using clang to compile seemed fine at `Compat Tests/ build(CLANG)`, however at `X86_64 CLANG Test / build` it showed that using the flag that clang doesn’t have again. But I did add a `ifeq` in `Makefile`
> 
> ```
> ifeq ($(HOSTCC), gcc)
> 	CFLAGS += -fno-reorder-blocks-and-partition
> endif
> ```
> 
> So, is there anything special for `X86_64 CLANG Test / build`?

I guess the reason might be when building the container, in `Dockerfile.x86_64`, The ARG `CC` is `gcc`. So the flag is just added into the `CFLAGS`. However, at some point, the `CC` is replaced by `CLANG`, but I'm not very sure about that.

Yangfisher1:
> > Hmm, it's quite strange. Using clang to compile seemed fine at `Compat Tests/ build(CLANG)`, however at `X86_64 CLANG Test / build` it showed that using the flag that clang doesn’t have again. But I did add a `ifeq` in `Makefile`
> > ```
> > ifeq ($(HOSTCC), gcc)
> > 	CFLAGS += -fno-reorder-blocks-and-partition
> > endif
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So, is there anything special for `X86_64 CLANG Test / build`?
> 
> I guess the reason might be when building the container, in `Dockerfile.x86_64`, The ARG `CC` is `gcc`. So the flag is just added into the `CFLAGS`. However, at some point, the `CC` is replaced by `CLANG`, but I'm not very sure about that.

The reason is that I used `HOSTCC` to determine instead of `CC`.

criu/config.c
```diff
@@ -1082,7 +1082,7 @@ int check_options(void)
 	}
 
 	if (opts.restore_on_demand)
-		pr_info("Restore using sys_mmap");
+		pr_info("Restore using sys_mmap\n");
```
avagin, line 1085: this should be in another patch

criu/Makefile
```diff
@@ -22,6 +22,7 @@ CONFIG-DEFINES		+= -DUSER_CONFIG_DIR='".criu/"'
 #
 # General flags.
 CFLAGS			+= -fno-strict-aliasing
+CFLAGS			+= -fno-reorder-blocks-and-partition
```
avagin: in the commit message you need to explain a reason why we need to this: why it was not a problem before your changes, when a problem was introduced and etc.

criu/pie/restorer.c
```diff
@@ -1528,11 +1549,43 @@ long __export_restore_task(struct task_restore_args *args)
...
+						* which is not VMA_AREA_REGULAR, and stack area, whose flag has MAP_GROWSDOWN still need
+						* to use `sys_pread`.
+						*/
+					if (!vma_entry_is(tmp_entry, VMA_AREA_REGULAR) || (tmp_entry->flags & MAP_GROWSDOWN) || (tmp_entry->flags & MAP_ANONYMOUS)) {
+						local_r = sys_pread(args->vma_ios_fd, p, io_size, local_offset);
+						pr_debug("Can't use mmap, still choose pread, %lx, %ld, %d\n", (unsigned long)p, io_size, args->vma_ios_fd);
+
+					} else {
+						sys_munmap(p, io_size);
+						sys_mmap(p, io_size, tmp_entry->prot | PROT_WRITE, (tmp_entry->flags & ~MAP_ANONYMOUS) | MAP_FIXED | MAP_PRIVATE, args->vma_ios_fd, local_offset);
```
avagin: you have to check a return value

criu/pie/restorer.c
```diff
@@ -1528,11 +1549,43 @@ long __export_restore_task(struct task_restore_args *args)
...
+						* Now check whether it can be mapped, notice that some memory areas such as `vsyscall`,
+						* which is not VMA_AREA_REGULAR, and stack area, whose flag has MAP_GROWSDOWN still need
+						* to use `sys_pread`.
+						*/
+					if (!vma_entry_is(tmp_entry, VMA_AREA_REGULAR) || (tmp_entry->flags & MAP_GROWSDOWN) || (tmp_entry->flags & MAP_ANONYMOUS)) {
+						local_r = sys_pread(args->vma_ios_fd, p, io_size, local_offset);
+						pr_debug("Can't use mmap, still choose pread, %lx, %ld, %d\n", (unsigned long)p, io_size, args->vma_ios_fd);
+
+					} else {
+						sys_munmap(p, io_size);
```
avagin: sys_munmap isn't needed.

criu/pie/restorer.c
```diff
@@ -1528,11 +1549,43 @@ long __export_restore_task(struct task_restore_args *args)
...
-			r = sys_preadv(args->vma_ios_fd, iovs, nr, rio->off);
+			/* Using mmap to load content as an on-demand way rather than egarly read all the content */
+			local_offset = rio->off;
+			for (j = 0; j < nr; j++) {
+				void * p = (void *)iovs[j].iov_base;
+				unsigned long end = (unsigned long)p + iovs[j].iov_len;
+				pr_debug("Change preadv to map of iov buffer %d, iov base %lx\n", j, (unsigned long)iovs[j].iov_base);
+				/* Considering the iov buffer may cross multiple regions */
+				while ((unsigned long)p < end) {
+					/* Find the vma_entry */
```
avagin: too many indents usually means that you need to create a helper function

criu/pie/restorer.c
```diff
@@ -318,6 +318,25 @@ static int restore_creds(struct thread_creds_args *args, int procfd, int lsm_typ
...
+ * vma entries inside the `args`. After getting vma_entry, we
+ * can judge whether this entry can be loaded on demand or not.
+ */
+static inline VmaEntry * lookup(void * pos, struct task_restore_args *args) 
+{
+	VmaEntry * vma_entry;
+	int i;
+	for (i = 0; i < args->vmas_n; i++) {
+		vma_entry = args->vmas + i;
+		if ((unsigned long)pos >= vma_entry->start && (unsigned long)pos < vma_entry->end) {
```
avagin, line 333: why do you need to do type conversion here?

criu/config.c
Yangfisher1: Sorry, I don’t know what "another patch" mean. Should I merge it into another branch?

criu/Makefile
Yangfisher1: > Most of them failed because when compiling, the file "criu/pie/restore-blob.h" would have a line `#define restorer_sym__export_restore_task.cold 0x475b`. I don't know why this happened cause the file was auto-generated. What I've done is changing some codes in `criu/pie/restorer.c`, function `__export_restore_task` . When I test the codes, using gcc 5.4.0 wouldn't produce this problem.

I'm not quite sure how "criu/pie/restore-blob.h" and the .cold symbol is generated. Also, using clang to compile wouldn’t produce this problem. It might be because I added some variables considering that using gcc 9.3.0 with original branch wouldn’t produce this problem. Using the flag `-fno-reorder-blocks-and-partition` prevent gcc from reordering basic blocks and partition into hot and cold sections so that the `.cold` symbol would disappear. Is there any way to generating `.cold` symbol? Or do you know why the `.cold` symbol generate?

criu/pie/restorer.c
Yangfisher1: Yes, I should check whether the return value equals to p or not. I need to check the return value of `sys_munmap` too.

criu/pie/restorer.c
Yangfisher1: I'll remove it later.

criu/pie/restorer.c
Yangfisher1: I’ll deal with it later.

criu/pie/restorer.c
Yangfisher1: `pos` is a pointer of type `void` and comparison between pointer and integer will throw an error.

criu/pie/restorer.c
avagin: let me know when this pr will be ready for review.

Yangfisher1: Now it's ready for review.

criu/include/cr_options.h
```diff
@@ -134,6 +134,7 @@ struct cr_options {
 	int link_remap_ok;
 	int log_file_per_pid;
 	int pre_dump_mode;
+	int restore_on_demand;
```
rst0git, line 137: `bool restore_on_demand;`

criu/cr-restore.c
```diff
@@ -3777,6 +3777,15 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
...
 	task_args->lsm_type = kdat.lsm;
 
+	/*
+	 * Decide restore with `sys_preadv` or `sys_mmap`,
+	 * sometimes we don't want to load all memory at once.
+	 */
+	if (opts.restore_on_demand)
+		task_args->on_demand_restore = true;
+	else
+		task_args->on_demand_restore = false;
```
rst0git, line 3787: `task_args->on_demand_restore = opts.restore_on_demand;`

criu/crtools.c
```diff
@@ -405,6 +405,9 @@ int main(int argc, char *argv[], char *envp[])
 	       "                        this requires running a second instance of criu\n"
 	       "                        in lazy-pages mode: 'criu lazy-pages -D DIR'\n"
 	       "                        --lazy-pages and lazy-pages mode require userfaultfd\n"
+		   "  --on-demand           restore pages on demand\n"
+		   "                        unlike lazy-pages using userfaultfd, on-demand use\n"
+	       "                        sys_mmap directly, so the kernel handle the page fault\n"
```
rst0git, line 410: @Yangfisher1 Could you merge this commit with the commit introducing `--on-demand` and add document this option in `Documentation/criu.txt` as well?

rst0git (commented):
@Yangfisher1 you can use `git rebase` to edit your patches and combine multiple patches into a single one. When you have multiple patches, the automated tests must pass after each patch, not just after the last one. This allows us to use tools such as `git bisect`.

Yangfisher1:
> @Yangfisher1 you can use `git rebase` to edit your patches and combine multiple patches into a single one. When you have multiple patches, the automated tests must pass after each patch, not just after the last one. This allows us to use tools such as `git bisect`.

@rst0git Does this mean that I have to merge all the commits into one, or does it mean that the commits corresponding to each patch are merged into one？

criu/crtools.c
Yangfisher1: Ok, I'll add document info.

criu/include/cr_options.h
Yangfisher1: Using a type of `bool` is more appropriate.

criu/cr-restore.c
Yangfisher1: This statement is redundant, and should be simplified.

criu/include/cr_options.h
rst0git: You are right, thanks for pointing it out.

rst0git:
> Does this mean that I have to merge all the commits into one, or does it mean that the commits corresponding to each patch are merged into one? So should I just combine all commits into one?

Not necessarily because large commits are difficult to review. You should have each logical change into a separate commit. However, when dividing your changes into a series of commits, you need to ensure that CRIU builds and runs properly after each commit. This helps when other developers are using `git bisect` to track down a problem.

You can use as an example some of the previous pull request / commits that have introduced new options to CRIU:

- `--network-lock`: https://github.com/checkpoint-restore/criu/pull/1539
- `--stream`: https://github.com/checkpoint-restore/criu/commit/7d79a58f4daa9ddda015fc0aaca2a60d156d12cc
- `--pre-dump-mode`: https://github.com/checkpoint-restore/criu/commit/20d4920a8bf74d1eceebc076bcc00889ba40e9f7

Yangfisher1:
Thanks! I'll refine these commits with these examples later.

Yangfisher1:
@rst0git Done.

criu/pie/restorer.c
```diff
@@ -340,6 +359,66 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+			 */
+			if (!vma_entry_is(tmp_entry, VMA_AREA_REGULAR) ||
+				(tmp_entry->flags & MAP_GROWSDOWN) ||
+				(tmp_entry->flags & MAP_ANONYMOUS)) {
+				local_r = sys_pread(args->vma_ios_fd, p, iov_size, local_offset);
+				pr_debug("Can't use mmap, still choose pread, %lx, %ld, %d\n",
+						(unsigned long)p, iov_size, args->vma_ios_fd);
+
+			} else {
+				tmp = sys_mmap(p, iov_size, tmp_entry->prot | PROT_WRITE,
```
avagin, line 402: why do we need PROT_WRITE here?

criu/pie/restorer.c
```diff
@@ -340,6 +359,66 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+static int map_iovec(struct iovec *iovs, loff_t off, int nr, struct task_restore_args *args)
+{
+	/* Using mmap to load content as an on-demand way rather than egarly read all the content */
+	int j;
+	VmaEntry *tmp_entry;
+	unsigned long iov_size, tmp;
+	loff_t local_offset = off;
+	ssize_t local_r, r = 0;
+	for (j = 0; j < nr; j++) {
+		void *p = (void *)iovs[j].iov_base;
```
avagin, line 377: I don't like all these type conversions in this code. I think you can add the offset variable instead of p and it will work better:
```
unsigned long offset = 0;
unsigned long end = iovs[j].iov_len;
...
while (offset < end) {
....
    tmp = sys_mmap(iovs[j].iov_base + offset, iov_size, tmp_entry->prot | PROT_WRITE
    ...
```

criu/pie/restorer.c
```diff
@@ -340,6 +359,66 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+ * using `sys_mmap` instead of `sys_preadv`
+ * can accelerate restore process when
+ * there are some pages we won't touch later.
+ */
+static int map_iovec(struct iovec *iovs, loff_t off, int nr, struct task_restore_args *args)
+{
+	/* Using mmap to load content as an on-demand way rather than egarly read all the content */
+	int j;
+	VmaEntry *tmp_entry;
+	unsigned long iov_size, tmp;
```
avagin, line 373: empty line after declarations

criu/pie/restorer.c
```diff
@@ -340,6 +359,66 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+			/* The length that we exactly want to map */
+			iov_size -= (long)p;
+
+			/*
+			 * Now check whether it can be mapped, notice that some memory areas such as `vsyscall`,
+			 * which is not VMA_AREA_REGULAR, and stack area, whose flag has MAP_GROWSDOWN still need
+			 * to use `sys_pread`.
+			 */
+			if (!vma_entry_is(tmp_entry, VMA_AREA_REGULAR) ||
+				(tmp_entry->flags & MAP_GROWSDOWN) ||
```
avagin, line 395: Need to fix alignments in multiline if expressions. Look at other place how it should be.

criu/pie/restorer.c
```diff
@@ -340,6 +359,66 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+		/* Considering the iov buffer may cross multiple regions */
+		while ((unsigned long)p < end) {
+			/* Find the vma_entry */
+			tmp_entry = lookup(p, args);
+			iov_size = tmp_entry->end < end ? tmp_entry->end : end;
+			/* The length that we exactly want to map */
+			iov_size -= (long)p;
+
+			/*
+			 * Now check whether it can be mapped, notice that some memory areas such as `vsyscall`,
```
avagin, line 390: vsyscall is provided by kernel and it isn't possible to change it.

Yangfisher1: Yes, I mean vsyscall and stack can not be mapped. They still need to use `sys_pread` to load.

> notice that some memory areas such as `vsyscall`, which is not VMA_AREA_REGULAR, and stack area, whose flag has MAP_GROWSDOWN still need to use `sys_pread`.

criu/pie/restorer.c
Yangfisher1: Ok, but it was the test `code linter` that made me change the alignments to this.

criu/pie/restorer.c
Yangfisher1: Ok, I'll change the type conversion.

criu/pie/restorer.c
Yangfisher1: My bad, I'll fix it.

criu/pie/restorer.c
Yangfisher1: My bad. PROT_WRITE is unnecessary. It might already be in `tmp_entry->prot`.

Yangfisher1:
@avagin I've already fixed the type conversion problem. It still need some type conversion to compare the address.

Yangfisher1:
Hmm, it's quite strange because I tried to use `make indent` and it showed me format like this. However, it still failed the code linter. I don't know why.

adrianreber:
> Hmm, it's quite strange because I tried to use `make indent` and it showed me format like this. However, it still failed the code linter. I don't know why.

Make sure to use the same version of clang-format (13) as our CI.

Yangfisher1:
> > Hmm, it's quite strange because I tried to use `make indent` and it showed me format like this. However, it still failed the code linter. I don't know why.
> 
> Make sure to use the same version of clang-format (13) as our CI.

My bad. I'm using a clang whose version is 11.0.0.

avagin:
How do you test your changes? Could you add the on-demand option in zdtm.py and execute tests in CI?

Yangfisher1:
> How do you test your changes? Could you add the on-demand option in zdtm.py and execute tests in CI?

I just simply replace line 1616, criu/pie/restorer.c
```
			if (args->on_demand_restore) {
```
into
```
			if (!args->on_demand_restore) {
```
and then test these changes. I don't know how to add this `on-demand` option in zdtm.py. Could you please tell me how to do?

rst0git:
> I don't know how to add this on-demand option in zdtm.py. Could you please tell me how to do?

@Yangfisher1 perhaps looking at the git history could help. This commit adds the `--pre-dump-mode` option to zdtm.py
https://github.com/checkpoint-restore/criu/commit/20d4920a8bf74d1eceebc076bcc00889ba40e9f7#diff-c9838f32b66301675f1aaba9ee24003ac88febf1f77cd993feba0db4b75d7ea6

Yangfisher1:
@rst0git Thank you! I'll try it later.

Yangfisher1:
@avagin Now I add this option into zdtm test. Could you please tell me whether it's enough for CI because I followed the commit 20d4920 ? Do I need to add some extra test for this new feature? I'm not sure how to add new tests.

rst0git:
@Yangfisher1 you might need to update [scripts/ci/run-ci-tests.sh](https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/ci/run-ci-tests.sh) to run the CI tests with `--on-demand`.

Yangfisher1:
@rst0git Get it. I'll update this file.

Yangfisher1:
It seems like the test `aarch64 Fedora Rawhide` would fail when compiling, quite strange.

adrianreber:
> It seems like the test `aarch64 Fedora Rawhide` would fail when compiling, quite strange.

Please rebase. There are already fixes on criu-dev which should make the CI errors go away.

Yangfisher1:
@adrianreber Thanks!

Yangfisher1:
@adrianreber I'm not sure how to pull new changes of `criu-dev` into my repo, sorry. I think maybe I should add a new remote repo with `git remote add` and pull the changes into my repo. Is that correct?

Yangfisher1:
@adrianreber Hi. I've already fetched the upstream but it still failed test `aarch64 Fedora Rawhide` as I mentioned before.

adrianreber:
This seems to be a GCC 12 related error unrelated to your pull request: https://github.com/checkpoint-restore/criu/issues/1731

There is, however, an error that you need to fix:
```
shellcheck scripts/*.sh
shellcheck scripts/ci/*.sh scripts/ci/apt-install

In scripts/ci/run-ci-tests.sh line 199:
./test/zdtm.py run -a --on-demand -p 2 --keep-going $ZDTM_OPTS
                                                    ^--------^ SC2086: Double quote to prevent globbing and word splitting.

Did you mean: 
./test/zdtm.py run -a --on-demand -p 2 --keep-going "$ZDTM_OPTS"

For more information:
  https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
```

Yangfisher1:
> This seems to be a GCC 12 related error unrelated to your pull request: #1731
> 
> There is, however, an error that you need to fix:
> 
> ```
> shellcheck scripts/*.sh
> shellcheck scripts/ci/*.sh scripts/ci/apt-install
> 
> In scripts/ci/run-ci-tests.sh line 199:
> ./test/zdtm.py run -a --on-demand -p 2 --keep-going $ZDTM_OPTS
>                                                     ^--------^ SC2086: Double quote to prevent globbing and word splitting.
> 
> Did you mean: 
> ./test/zdtm.py run -a --on-demand -p 2 --keep-going "$ZDTM_OPTS"
> 
> For more information:
>   https://www.shellcheck.net/wiki/SC2086 -- Double quote to prevent globbing ...
> ```

Thanks! I'm not familiar with shell scripts and forgot to check the code style. Maybe I still need to check these failed test and modify codes.

Yangfisher1:
@adrianreber Could you please tell me the difference between
`Run zdtm/static/maps06 in ns` and `Run zdtm/static/maps06 in h`.
It seems like these patches would fail at zdtm/static/maps06 in ns on CentOS 7 based test. However it worked well on my machine and ci/circleci.

rst0git:
> Could you please tell me the difference between
Run zdtm/static/maps06 in ns and Run zdtm/static/maps06 in h.

`ns` and `h` are flavors: https://criu.org/ZDTM_test_suite#Flavors

>It seems like these patches would fail at zdtm/static/maps06 in ns on CentOS 7 based test. However it worked well on my machine and ci/circleci.

The CentOS 7 test is using Python 2. You can try to replicate the error with CentOS 7 container.

Yangfisher1:
Thanks! I will try to reproduce this error in container.

compel/arch/arm/plugins/std/syscalls/syscall.def
```diff
@@ -22,6 +22,7 @@ rt_sigprocmask	sigprocmask	135	175	(int how, k_rtsigset_t *set, k_rtsigset_t *ol
 rt_sigreturn			139	173	(void)
 ioctl				29	54	(unsigned int fd, unsigned int cmd, unsigned long arg)
 pread64				67	180	(unsigned int fd, char *buf, size_t count, loff_t pos)
+pread				67	180	(unsigned int fd, char *buf, size_t count, loff_t pos)
```
0x7f454c46, line 25: There's no `pread()` syscall on aarch64/arm32.
Aliasing one to another shouldn't be done with a hack in syscall table file, rather in a C header.
Besides, what's the reason why you just use `pread64()` instead of `pread()`?

Yangfisher1: Sure, I know there's no `pread()` on aarch64. However, these syscall functions are autogenerated when compiling and related to system architecture(for example, in my ubuntu 20.04 virtual machine, there is no sys_pread64() syscall). So I think the only solution is to add this into syscall table.

0x7f454c46: Ok, I see what you mean, the other platforms are already "renaming" it in their syscall tables:
```
compel/arch/mips/plugins/std/syscalls/syscall_64.tbl:__NR_pread64                       5016            sys_pread
compel/arch/ppc64/plugins/std/syscalls/syscall-ppc64.tbl:__NR_pread64           179             sys_pread
compel/arch/s390/plugins/std/syscalls/syscall-s390.tbl:__NR_pread64             180             sys_pread
compel/arch/x86/plugins/std/syscalls/syscall_32.tbl:__NR_pread64                180             sys_pread64
compel/arch/x86/plugins/std/syscalls/syscall_64.tbl:__NR_pread64                        17              sys_pread
```
Could you than just "rename" it too `pread64 = > pread`, rather than adding a new alias?

criu/Makefile
```diff
@@ -21,6 +21,9 @@ CONFIG-DEFINES		+= -DUSER_CONFIG_DIR='".criu/"'
 
 #
 # General flags.
+ifeq (gcc, $(findstring gcc, $(CC)))
+    CFLAGS += -fno-reorder-blocks-and-partition
+endif
```
0x7f454c46, line 26: Could you try this: https://github.com/checkpoint-restore/criu/pull/1749 ?
Disabling optimization in the whole project because of a minor issue in PIEs would be overkill.
At worst, you could have disabled it for a specific restorer's file.

Yangfisher1: OK, I'll try this.

> At worst, you could have disabled it for a specific restorer's file.

I agree, but I just don't know how to modify the Makefile to implement this. I'm not that familiar with Makefile.

compel/arch/arm/plugins/std/syscalls/syscall.def
Yangfisher1: OK, considering that there's no `sys_pread64()` used in this project.

criu/Makefile
Yangfisher1: > Could you try this: https://github.com/checkpoint-restore/criu/pull/1749 ?

Quite strange because the .cold symbol is still generated. Maybe I should modify the Makefile and disable it for `criu/pie/restorer.c` but I have no idea how to do this.

Yangfisher1:
It's quite strange. The test `zdtm/static/maps06` failed on CentOS 7 while it passed on Ubuntu 20.04. However, the test `zdtm/static/maps09` passed on CentOS 7 while it failed on Ubuntu 20.04. I'm a little confused. On Ubuntu 20.04, the test `zdtm/static/maps09` failed when using `sys_mmap`. It returned a very high address `0xffffffffffffffea` rather than what's expected even with a flag `MAP_FIXED`(I also tried to use `sys_munmap` and it also failed). But It worked fine on CentOS 7.

Yangfisher1:
I've fixed the CentOS 7 failure. It's due to the file permission. The permissions of these links in `/proc/pid/map_files` are decided by the access mode of their file descriptors. E.g. if you open a file with `O_RDWR` and mmap a memory region with the fd, the permission of the link in `/proc/pid/map_files` would be `lrw-------`, whose `st_mode` would be `0o120600`. So using `sys_mmap` with the flag `MAP_FIXED` would discard the existing mapping and access mode of `page.img` is `O_RDONLY`, so the permission bit would change. I chose to use more `sys_pread` to avoid the problem, and finally it passed on my CentOS 7 virtual machine.

However, I start to think, is it necessary to check these permission bits? I mean that without the flag `MAP_SHARED` or `MAP_ANONYMOUS`, any changes in a mmaped memory region with flag `MAP_PRIVATE` couldn't affect the underlying file. And you can still write something into these so-called "read only" memory region with a permission of `lr--------` in `/proc/pid/map_files`.

criu/pie/restorer.c
```diff
@@ -340,6 +359,70 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+			iov_size -= (unsigned long)iovs[j].iov_base + iov_offset;
+
+			/*
+			 * Now check whether it can be mapped, notice that some memory areas such as `vsyscall`,
+			 * which is not VMA_AREA_REGULAR, and stack area, whose flag has MAP_GROWSDOWN still need
+			 * to use `sys_pread`.
+			 */
+			if (!vma_entry_is(tmp_entry, VMA_AREA_REGULAR) || (tmp_entry->flags & MAP_GROWSDOWN) ||
+			    (tmp_entry->flags & MAP_ANONYMOUS) || (tmp_entry->flags & MAP_HUGETLB) ||
+			    (tmp_entry->flags & MAP_SHARED) || (tmp_entry->prot & PROT_WRITE)) {
```
avagin, line 399: Did you test your changes? Here I see that we don't use mmap for anon and shared mappings. What other mappings do we expect here? Did I miss something?

avagin: Why do we need to skip writable mappings?

Yangfisher1: > Did you test your changes?


Sure. I've added them into CI test in my commits.

> Here I see that we don't use mmap for anon and shared mappings.

Sure, anonymous mappings and shared mappings are commonly used for IPC, and any updates are carried through to the underlying file. Remapping them into the `page.img` is not a good idea, I think.

> What other mappings do we expect here?

However, the flag `MAP_PRIVATE` is still ok. It won't affect the underlying file.

> Why do we need to skip writable mappings?

I'm doing this just to pass the CI test with `on-demand` mode. I explained the reason above.

avagin: >> What other mappings do we expect here?

> However, the flag MAP_PRIVATE is still ok. It won't affect the underlying file.

Sorry. I didn't notice that MAP_PRIVATE isn't here. So do you want to handle only file private mappings this way? Do you have any real workload that benefits from this? Have you tested it?

>> Why do we need to skip writable mappings?

> I'm doing this just to pass the CI test with on-demand mode. I explained the reason above.

This is a bad explanation. CI passes because the introduced code path isn't used in tests. Read-only  private file mappings usually don't have any pages in images.

Yangfisher1: >  So do you want to handle only file private mappings this way?

I don't want it, but it seems like remapping these shared and anonymous regions with `page.img` is a bad idea. 

> Do you have any real workload that benefits from this? Have you tested it?

Sure. For example, if you import some packages, like `import requests`, python will load all contents into memory while using a little part of it. I use CRIU to optimize my project about serverless, so removing these unused parts by using `sys_mmap` benefits.

> CI passes because the introduced code path isn't used in tests.

Nope. I've introduced it into the CI test in commit [d920aee](https://github.com/Yangfisher1/criu/commit/d920aee2e75c865255a3ce5bc550bfc38199dcdb). At first, It failed at `maps06` on CentOS 7. I fixed the failure by still choosing to use `sys_pread` for these `PROT_WRITE` regions, but I don't think it's necessary to check these permission bits as discussed [above](https://github.com/checkpoint-restore/criu/pull/1711#issuecomment-1047377556) .

lijunqiang123:
Hello，I want use the option`on-demand` to speed up the time of `criu restore`. Can I use C API to run the option? And I don't find the related funcation in `lib/c/criu.h`.

Yangfisher1:
@lijunqiang123 You can clone the repo and compile it. Use command `criu restore --on-demand`.

criu/pie/restorer.c
avagin: > I don't want it, but it seems like remapping these shared and anonymous regions with page.img is a bad idea.

Could you explain why it is bad?

> Sure. For example, if you import some packages, like import requests, python will load all contents into memory

Could you be more detailed here? How does it load all contents into memory? Does it create a read-only file mapping?  If the answer is yes,  how many pages does criu dump?

> Nope. I've introduced it into the CI test in commit d920aee. 

It is unclear for me whether tests triggers the new code path...

Could you show some numbers how it speeds up your workload? In ideal case, I would like to have an example that I can play with on my host.

criu/pie/restorer.c
```diff
@@ -340,6 +359,70 @@ static inline int restore_pdeath_sig(struct thread_restore_args *ta)
...
+				local_r = sys_pread(args->vma_ios_fd, iovs[j].iov_base + iov_offset, iov_size,
+						    local_offset);
+				pr_debug("Can't use mmap, still choose pread, %p, %ld, %d\n",
+					 iovs[j].iov_base + iov_offset, iov_size, args->vma_ios_fd);
+
+			} else {
+				tmp = sys_mmap(iovs[j].iov_base + iov_offset, iov_size, tmp_entry->prot,
+					       tmp_entry->flags | MAP_FIXED | MAP_PRIVATE, args->vma_ios_fd,
+					       local_offset);
```
avagin, line 408: Will it work for anon rw mappins with this changes:

```
				tmp = sys_mmap(iovs[j].iov_base + iov_offset, iov_size,
￼					       tmp_entry->prot,
￼					       (tmp_entry->flags & ~MAP_ANONYMOUS) |
￼					       MAP_FILE| MAP_FIXED | MAP_PRIVATE,
￼					       args->vma_ios_fd,
￼					       local_offset);
```

Yangfisher1: In fact, It’s my first version of the implementation. However, it led to CI failure because the CI test checks some process status before dump and after restore and /proc/$pid/proc/%s/map_files is one of the targets. Changing these no-file-backed anonymous memory mapping into file-based memory mapping makes this check fail.

criu/pie/restorer.c
Yangfisher1: > Could you explain why it is bad?

For anonymous memory regions, it failed at the CI test, I explained the reason below.

> Could you be more detailed here? How does it load all contents into memory?

I'm not very familiar with this part but I'll search for some information. 

> Does it create a read-only file mapping? If the answer is yes, how many pages does criu dump?

Sure, mostly they're shared libraries like `.local/lib/python3.5/site-packages/Pillow.libs/libXdmcp-e15573e7.so.6.0.0` and `.local/lib/python3.5/site-packages/Pillow.libs/libopenjp2-b3d7668a.so.2.3.1`. The number of pages depends on applications and the packages that you import.

> It is unclear for me whether tests triggers the new code path...

file [scripts/ci/run-ci-tests.sh](https://github.com/checkpoint-restore/criu/commit/d920aee2e75c865255a3ce5bc550bfc38199dcdb#diff-00f0e392c1d1032a08a4204b60f0f3406cf270ed83bb0b8975e856bc01d5ed85), line 206.
```
./test/zdtm.py run -a -p 2 --on-demand --keep-going $ZDTM_OPTS
```
The option `--on-demand` makes it use on-demand mode.

> Could you show some numbers how it speeds up your workload? In ideal case, I would like to have an example that I can play with on my host.

Sure, I put two of my test cases on pastebin. Just call them TCA(Test Case A) and TCB(Test Case B).
TCA: https://pastebin.com/ujZjmrE5
TCB: https://pastebin.com/GXyr24Xf

We use a lock file to control the behavior of the program before dumping and after restoring. In our test, for TCA, the runtime is 10ms vs 5ms. For TCB, the runtime is 80ms vs 50ms.

Yangfisher1:
@adrianreber Hi, Could you please approve the next workflows so that I can test the correctness?

PR #1713, opened 2021-12-26
DRAFT: RISCV64

nirousseau:
Work in progress and efforts for bringing riscv64 arch to criu


mihalicyn:
Let's close this as no activity there.

Yixue (@felicitia) is working on RISC-V port right now with help from @rushi47 and me.

PR #1718, opened 2022-01-06
pagemap: tiny fix on truncating memory image

liusdu:
When requested iovs are huge, criu needs to invoke more then one
preadv()s. In this situation criu truncates memory image with
offset of first preadv() and length of last one, which leads
to leakage of memory image. This patch fixs truncating with right
offset and length.


avagin:
lgtm

liusdu:
ping @Snorch

Snorch (approved):
Looks good. Thanks for the fix!

PR #1721, opened 2022-01-14
Mount v2 porting

Snorch:
After we've merged MOVE_MOUNT_SET_GROUP feature to mainstream linux v5.15 https://github.com/torvalds/linux/commit/9ffb14ef61bab now we can use it to restore sharing groups of mounts without the need to care about inheriting those groups when create mounts, we can just set sharing groups at later stage and before that construct mount trees with private mounts.

This is a *big* PR, containing multiple fixes/reworks/refactoring to current mount engine to make it suitable to integrate with new mount engine ("mount-v2").

Main changes:
- Bindmounts detection fixed and is now done as early as possible just after reading mounts from images (same to overmounted detection and merging namespaces to root yard)
- Don't adding "helper" mounts to mntinfo list
- Mark device-external mounts ->external similar to mountpoint-external ones
- Rework how mnt_is_external works and do many other useful helpers
- Detect mountpoint-external mounts only in root mntns of dumpee
- Allow nested mount namespaces with different root mount 
- Create "helper" binfmtmisc mount in root yard
- Do not override master_id to -1 for root binds

Other important part is switching most of places where we used mount_info->mountpoint to ->ns_mountpoint (e.g. in cases where we get relative paths between mounts, or just do prints or on dump where these two are equal)

All places which actually used ->mountpoint to access <root_yard>/<ns_dir>/<ns_mountpoint> switched to use service_mountpoint() helper which would allow to switch these paths easily in one place for mount-v2 as paths in root yard for mounts in mount-v2 are different ("plain mountpoints" - <root_yard>/<mnt_dir>).

Also PR exports some useful functions/defines/global-vars from mount.c so that mount-v2 can reuse it.

"mount: add new mounts-v2 engine" is a main thing (full algorithm description in commit message), it integrates in current mount engine, it resolves sharing groups and then restores namespaces restores mounts (plain) in them, assembles mount trees for each mntns and when all mounts are ready restores sharing groups on mounts.

Many patches are just refactoring and it's hard to prepare tests for them, but there are some tests added to cover actual changes, I've added FIXME to improve two of them as in Virtuozzo criu they just relied on --check-mounts to check that mounts had restored right, but we don't have it in mainstream so I will improve tests in near future.


avagin:
@Snorch pls describe a new way of restoring mounts on the wiki

Snorch:
> @Snorch pls describe a new way of restoring mounts on the wiki

Ok, I should probably update https://criu.org/Mounts-v2 with a new details.

Snorch:
Cherry-pick https://github.com/checkpoint-restore/criu/pull/1706/commits/c2cf6cdf6b0b30b69c957513c831cb91faacdd69 to make tests go on newer kernel.

Snorch:
After enabling running tests on new kernel in ci, I have a testfail in inotify_irmap. This fail reproduces on my F35:

```
./test/zdtm.py run -t zdtm/static/inotify_irmap --fault 128 --pre 2 -f uns
```

And it is not related to PR, but it is related to running on btrfs, again dev returned by stat does not match dev returned by inotify wd in fdinfo...

brauner:
> After enabling running tests on new kernel in ci, I have a testfail in inotify_irmap. This fail reproduces on my F35:
> 
> ```
> ./test/zdtm.py run -t zdtm/static/inotify_irmap --fault 128 --pre 2 -f uns
> ```
> 
> And it is not related to PR, but it is related to running on btrfs, again dev returned by stat does not match dev returned by inotify wd in fdinfo...

IIrct, that's a longstanding "bug" in btrfs and a discussion how to fix this upstream has been going on for quite a while. I _think_ it's https://lore.kernel.org/linux-nfs/202108091545.FBDf88Zt-lkp@intel.com

criu/util.c
```diff
@@ -1804,3 +1806,173 @@ int run_command(char *buf, size_t buf_size, int (*child_fn)(void *), void *args)
...
+ * 	///a///b///c is the same as /a/b/c
+ * 3) It always returns relative path, with no leading slash:
+ * 	get_relative_path("/a/b/c", "/") would be "a/b/c"
+ * 	get_relative_path("/a/b/c", "/a/b") would be "c"
+ * 	get_relative_path("/", "/") would be ""
+ * 4) It can handle paths with single dots:
+ * 	get_relative_path("./a/b", "a/") would be "b"
+ * 5) Note ".." in paths are not supported and handled as normal directory name
+ */
+char *get_relative_path(char *path, char *sub_path)
```
adrianreber, line 1824: These functions look like they would be good candidates for unit testing.

criu/mount.c
```diff
@@ -929,6 +929,28 @@ static void search_bindmounts(void)
...
+	BUG_ON(!mi);
+
+	if (pick(mi, mi))
+		return mi;
+
+	/*
+	 * Shouldn't use mnt_bind list before it was setup in
+	 * __search_bindmounts
+	 */
+	BUG_ON(list_empty(&mi->mnt_bind) && !mi->mnt_no_bind);
```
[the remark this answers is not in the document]
Snorch: The mi->mnt_no_bind==true is an indicator that empty mi->mnt_bind list reflects that mount has no bindmounts other than itself. If both mnt_no_bind is false and list is empty than bindmount search haven’t been done yet and one can’t yet use the list

criu/util.c
```diff
@@ -1804,3 +1804,75 @@ int run_command(char *buf, size_t buf_size, int (*child_fn)(void *), void *args)
...
```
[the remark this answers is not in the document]
Snorch: Function get_relative_path has two main features:
 
1) It works with almost any complex paths (".." is exception, ".." are handled as simple directory names because resolving them as actual "parent directory" link may require fs accesses);
2) It does NO allocation and NO memory writes, returned relative path pointer is a shifted path pointer.

In contrast path_simplify need to allocate memory and need to move chars. In criu file/mount code we construct a lot of paths, to be robust we would need to allocate and write again when simplifying each constructed path.

So we already have many path construction code and patching all of it is a big deal I believe. Integrating path_simplify into get_relative_path looks like a bad solution because the latter is(/would be) called in many places and I like that it does not add memory complexity to the overall algorithm.

criu/util.c
Snorch: Nice idea, I will try to add unit test.

Snorch: Please see cfde751fd ("unittest: add some tests for get_relative_path helper")

criu/util.c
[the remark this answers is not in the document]
brauner: Yeah, sounds reasonable. I do generally prefer to always simplify and clean paths before processing them further in the code so I'm biased.

criu/mount.c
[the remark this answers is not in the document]
brauner: That's a lot of information stuffed into that little line of code. :) Maybe a tiny static inline helper with a comment explaining that would help. But just a thought.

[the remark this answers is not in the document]
Snorch: I've just simplified the code a bit. Now we have mnt_bind_is_populated instead of mnt_no_bind.

The mnt_no_bind was set only for mounts with empty mnt_bind list, it was a bit more effective in terms of memory writes, but mnt_bind_is_populated is more simple, mnt_bind_is_populated now is set for all mounts just after we've collected bindmounts for them, so proper bug_on is simple as `BUG_ON(!m->mnt_bind_is_populated);`

Snorch:
>Cherry-pick c2cf6cd to make tests go on newer kernel.

Thanks to @adrianreber for pointing out that "Vagrant Fedora based test (no VDSO)" already tests mount-v2 and MOVE_MOUNT_SET_GROUP is supported there, so don't need this cherry-pick.

Snorch:
Here is a new wiki https://criu.org/Mount-v2

Snorch:
Slight improvement to move_mount_to_tree, make it use more robust sys_move_mount. (symlinks and automounts are not a problem for it now.)

Snorch:
Added test/jenkins/criu-mount-v2.sh script to run mount-v2 tests in jenkins:
https://ci.kernoops.org/job/CRIU/job/CRIU-mount-v2/job/mount-v2-porting/

adrianreber:
Not sure if you have seen the aarch64 clang failure:
```
make[3]: Nothing to be done for 'criu/string.c'.

  DEP      criu/unittest/unit.d

  DEP      criu/unittest/mock.d

  CC       criu/unittest/mock.o

  CC       criu/unittest/unit.o

  LINK     criu/unittest/built-in.o

  LINK     criu/unittest/unittest

criu/util.o: in function `rlimit_unlimit_nofile':

/drone/src/criu/util.c:1366:(.text+0x2c28): relocation truncated to fit: R_AARCH64_LDST32_ABS_LO12_NC against symbol `kdat' defined in .bss section in criu/unittest/built-in.o

/usr/bin/ld: /drone/src/criu/util.c:1366: warning: one possible cause of this error is that the symbol is being referenced in the indicated code as if it had a larger alignment than was declared where it was defined

/drone/src/criu/util.c:1375:(.text+0x2c88): relocation truncated to fit: R_AARCH64_LDST32_ABS_LO12_NC against symbol `kdat' defined in .bss section in criu/unittest/built-in.o

/usr/bin/ld: /drone/src/criu/util.c:1375: warning: one possible cause of this error is that the symbol is being referenced in the indicated code as if it had a larger alignment than was declared where it was defined

clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The CentOS 8 based test errors still seems to be #1734.

Nice to see Jenkins test results in PRs. Independent of this PR it would be good to have s390x builds also available for every PR because that is something that is not tested during PRs. Neither cross-compile (does not work because endianess) nor native.

Snorch:
> Not sure if you have seen the aarch64 clang failure:
> 
> ```
> make[3]: Nothing to be done for 'criu/string.c'.
> 
>   DEP      criu/unittest/unit.d
> 
>   DEP      criu/unittest/mock.d
> 
>   CC       criu/unittest/mock.o
> 
>   CC       criu/unittest/unit.o
> 
>   LINK     criu/unittest/built-in.o
> 
>   LINK     criu/unittest/unittest
> 
> criu/util.o: in function `rlimit_unlimit_nofile':
> 
> /drone/src/criu/util.c:1366:(.text+0x2c28): relocation truncated to fit: R_AARCH64_LDST32_ABS_LO12_NC against symbol `kdat' defined in .bss section in criu/unittest/built-in.o
> 
> /usr/bin/ld: /drone/src/criu/util.c:1366: warning: one possible cause of this error is that the symbol is being referenced in the indicated code as if it had a larger alignment than was declared where it was defined
> 
> /drone/src/criu/util.c:1375:(.text+0x2c88): relocation truncated to fit: R_AARCH64_LDST32_ABS_LO12_NC against symbol `kdat' defined in .bss section in criu/unittest/built-in.o
> 
> /usr/bin/ld: /drone/src/criu/util.c:1375: warning: one possible cause of this error is that the symbol is being referenced in the indicated code as if it had a larger alignment than was declared where it was defined
> 
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```
> 
> The CentOS 8 based test errors still seems to be #1734.
> 
> Nice to see Jenkins test results in PRs. Independent of this PR it would be good to have s390x builds also available for every PR because that is something that is not tested during PRs. Neither cross-compile (does not work because endianess) nor native.

Seems that defining empty `struct kerndat_s {};` is not enough to mock linker then linking on aarch64 clang for unittests... Need to think on how to fix it as I don't want to include kerndat.h (it will bring more and more things to mock) and also don't want to make full declaration as kdat content may change in future... Same problem may also happen with `struct VmaEntry {};` and `struct ns_desc {};`.

Oh, will think on how to fix it... Any advice would be greatly appreciated.

adrianreber:
I agree that unit testing is difficult because  what it might all pull in.

The only idea I had at the time is moving code that is unit tested to a separate file which we know is used in unit testing but has not other function which depend on other CRIU parts. This means that we partly have to design our code to fit unit testing. Which is not really nice. I have no real idea than moving it to a separate files without dependencies.

Snorch:
I've found https://github.com/checkpoint-restore/criu/pull/1721/commits/b82f268c091a2620ebae1063bca970af87a13bf8 this way, though I don't know if it would help or not...

Snorch:
As --gc-sections link flag approach does not help, I've added this hunk to mock the problem:

```
--- a/criu/unittest/mock.c
+++ b/criu/unittest/mock.c
@@ -129,6 +129,7 @@ int clone_noasan(int (*fn)(void *), int flags, void *arg)
 }
 
 struct kerndat_s {
+       unsigned int sysctl_nr_open;
 };
 struct kerndat_s kdat = {};
```

And according to aarch64 CLANG tests it seem to work fine. -> will merge this to original unittest patch.

Snorch:
Re-based with minor changes to fix conflicts.

Snorch:
FIXME improvement/fixing (--check-moutns port to zdtm) is here https://github.com/checkpoint-restore/criu/pull/1755 I believe it's better to send it separately, as it's over 1000 more lines.

criu/mount-v2.c
```diff
@@ -0,0 +1,1159 @@
...
+
+	if (remount_ro)
+		sflags &= ~MS_RDONLY;
+
+	if (do_mount(mi, src, mnt_fsname(mi), sflags) < 0) {
+		pr_perror("Can't mount at %s", mi->plain_mountpoint);
+		return -1;
+	}
+
+	if (mount(NULL, mi->plain_mountpoint, NULL, MS_PRIVATE, NULL)) {
```
avagin, line 235: Why do we need this extra mount call?  Could you add a comment that explain this?

criu/config.c
```diff
@@ -697,6 +698,7 @@ int parse_options(int argc, char **argv, bool *usage_error, bool *has_exec_cmd,
 		{ "file-validation", required_argument, 0, 1098 },
 		{ "lsm-mount-context", required_argument, 0, 1099 },
 		{ "network-lock", required_argument, 0, 1100 },
+		BOOL_OPT("mounts-v2", &opts.mounts_v2),
```
avagin, line 701: I think mounts_v2 should be used by default if it is supported. We can add --force_mount_v1 to disable mounts_v2.

avagin:
LGTM

criu/mount-v2.c
Snorch: This is a place where we make sure that the mount created by do_new_mount_v2() is 100% private and we would never get any unexpected propagation with it on mount tree construction stage. We do MS_PRIVATE mount call in all places just after mounting something including do_bind_mount_v2(), do_mount_root_v2() and make_yard().

I understand that this call can be considered "excess" as if we know that parent mount is 100% private and we create a fresh new (not bindmount) children mount for it child would be private too, and also even for bindmount of private source new mount would also be private. But I prefer do one extra call for mount syscall but make code simpler so that it is just obvious from the nearby code that newly created mount would be private.

Not sure if I need to add this explanation everywhere or just some note on top of mount-v2.c would be enough?

criu/config.c
Snorch: That's a good point, Thanks!

We have "--mounts-compat" option in vzcriu, and we also have a fallback if mount-v2 is not supported to switch to compat mode. Does this name sounds ok for you?

rst0git: `--mounts-compat` sounds good

Snorch: I already prepared a rework with `--force-mounts-compat`, I believe using the word `force` enlightens the fact that we also have an implicit fallback to compat mode if needed kernel features are absent.

Snorch:
Currently I prepared PR update to use mounts-v2 engine by default as @avagin proposed here https://github.com/checkpoint-restore/criu/pull/1721#discussion_r820166993.

To-do is:
- to add more comments to mount-v2.c as discussed with @avagin personally
- fix jenkins job, which would obviousely fail after rename of .sh file

Snorch:
I see strange centos 8 job error:
```======================= Run zdtm/static/cmdlinenv00 in h =======================
Start test
Test is SUID
./cmdlinenv00 --pidfile=cmdlinenv00.pid --outfile=cmdlinenv00.out --arg1=arg1 --arg2=arg2 --arg3=arg3
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/cmdlinenv00/52/1/restore.log
b'(00.000025) Version: 3.16.1 (gitid e97b9aa03)'
b'(00.000038) Running on cirrus-task-5615469339082752 Linux 4.18.0-365.el8.x86_64 #1 SMP Thu Feb 10 16:11:23 UTC 2022 x86_64'
b'(00.000048) Loaded kdat cache from /run/criu.kdat'
b"(00.000061) Hugetlb size 1024 Mb is supported but cannot get dev's number"
b'(00.000065) Warn  (criu/mount-v2.c:36): mnt-v2: Mounts-v2 requires MOVE_MOUNT_SET_GROUP support'
b'(00.000066) Warn  (criu/config.c:1113): Mount engine fallback to --force-compat-mounts mode'
Send the 15 signal to  52
Wait for zdtm/static/cmdlinenv00(52) to die for 0.100000
############## Test zdtm/static/cmdlinenv00 FAIL at result check ###############
18:15:22.882:    52: old cmdline: ./cmdlinenv00
18:15:22.882:    52: old environ: CIRRUS_CHANGE_TITLE=Mount v2 porting
18:15:22.902:    52: new cmdline: ./cmdlinenv00
18:15:22.902:    52: new environ: CIRRUS_CHANGE_TITLE=Mount v2 porting
18:15:22.902:    52: FAIL: cmdlinenv00.c:116: auxv corrupted on restore (errno = 11 (Resource temporarily unavailable))

```
It does not reproduce on my F35, has anyone seen anything similar?

adrianreber:
> It does not reproduce on my F35, has anyone seen anything similar?

I have never seen that one. Must be something new. It seems the kernel is newer than it used to be. It  worked on 4.18.0-358.el8.x86_64 but does not work on 4.18.0-365.el8.x86_64.

I am just booting a VM to verify.

adrianreber:
I cannot verify it locally with exactly the same kernel. Strange.

adrianreber:
It could be related to the addition of AT_MINSIGSTKSZ, but I am not sure. Just guessing.

adrianreber:
@Snorch The error is gone in my PR.

Snorch:
Changes:
- rebase on newest criu-dev
- rename option `%s/force\(.\)compat\(.\)mounts/mntns\1compat\2mode/g` according to @avagin idea
- add one more patch in the end to disable mount-v2 for pty-console test as it does not survive two c/r iterations (see more info in patch comment, planing to fix it later, idea is in patch FIXME comment).

criu/config.c
Snorch: Updated to --mntns-compat-mode as discussed with @avagin

Snorch:
> To-do is:
> 
>     * to add more comments to mount-v2.c as discussed with @avagin personally
> 
>     * fix jenkins job, which would obviousely fail after rename of .sh file
> 
>     * review coverity scan

All seem to be done except jenkins jobs (we have a temporary node problems due to migration of our jenkins test nodes to a new hosting), hope it's enough for merging.

adrianreber (approved):
I cannot really comment on the content as I do not really understand it, but CI seems happy. Andrei has given his LGTM. For what it is worth I am also :+1: 

As Andrei has done most of the review I let him decide if this can be merged.

avagin:
BTW: pls don't add such lines in commit messages:
```
Cherry-picked from Virtuozzo criu:
https://src.openvz.org/projects/OVZ/repos/criu/commits/2be0ff276
```

PR #1723, opened 2022-01-16
zdtm: fix zdtm/static/maps00 case in arm64

time-river:
zdtm/static/maps00 case sometimes will cause SIGILL signal in arm64 platform.
<<ARM Coretex-A series Programmer's Guide for ARMv8-A>> notes:
> The ARM architecture does not require the hardware to ensure coherency
> between instruction caches and memory, even for locations of shared
> memory.

Therefore, we need flush dcache and icache for self-modifying code.

### DIFF
--- a/test/zdtm/static/maps00.c
+++ b/test/zdtm/static/maps00.c
@@ -158,7 +158,13 @@ static int check_map(struct map *map)
 
 	if (!sigsetjmp(segv_ret, 1)) {
 		if (map->prot & PROT_WRITE) {
-			memcpy(map->ptr, test_func, getpagesize());
+			memcpy(map->ptr, test_func, ONE_MAP_SIZE);
+			/* The ARM ARM architecture does not require the
+			 * hardware to ensure coherency between instruction
+			 * caches and memory, flushing dcache and icache is
+			 * necessory to prevent SIGILL signal.
+			 */
+			__builtin___clear_cache(map->ptr, map->ptr + ONE_MAP_SIZE);
 		} else {
 			if (!(map->flag & MAP_ANONYMOUS)) {
 				uint8_t funlen = (uint8_t *)check_map - (uint8_t *)test_func;


adrianreber:
> The python2 is deprecated, switch python3, add src encodings.

Thanks for the pull request, but we want to support CentOS/RHEL 7 until the end of its lifetime which is 2024. CentOS/RHEL 7 has a working and supported Python 2. Until then we want to keep Python 2 support (if possible).

Using an unversioned Python binary like in `zdtm.py` is not really a sign of Python 2. Looking at https://www.python.org/dev/peps/pep-0394/ the version the `python` binary can either point to Python 2 or Python 3.

time-river:
> > The python2 is deprecated, switch python3, add src encodings.
> 
> Thanks for the pull request, but we want to support CentOS/RHEL 7 until the end of its lifetime which is 2024. CentOS/RHEL 7 has a working and supported Python 2. Until then we want to keep Python 2 support (if possible).
> 
> Using an unversioned Python binary like in `zdtm.py` is not really a sign of Python 2. Looking at https://www.python.org/dev/peps/pep-0394/ the version the `python` binary can either point to Python 2 or Python 3.

All right. I drop this commit.

adrianreber:
Please rebase to get the CI errors fixed.

test/zdtm/static/maps00.c
```diff
@@ -158,7 +158,8 @@ static int check_map(struct map *map)
 
 	if (!sigsetjmp(segv_ret, 1)) {
 		if (map->prot & PROT_WRITE) {
-			memcpy(map->ptr, test_func, getpagesize());
+			memcpy(map->ptr, test_func, ONE_MAP_SIZE);
```
adrianreber, line 161: Why the change from `getpagesize()` to `ONE_MAP_SIZE`?

time-river: I find `mmap(ONE_MAP_SIZE)` instead of `mmap(getpagesize())`. `getpagesize() == ONE_MAP_SIZE` isn't always correct, changing it to prevent the potential overflow.

adrianreber: Looking at the code of the test I agree. It is strange that `getpagesize()` is used here.

It does not really fit the commit description, but splitting in two one-line commits seems unnecessary. I am okay with this.

adrianreber (approved):
CI is happy. The description sounds reasonable. I have never seen this error but I am not running CRIU very often on aarch64.

adrianreber:
@Snorch this PR looks good from my point of view. Can you merge it if it also makes sense to you?

test/zdtm/static/maps00.c
```diff
@@ -158,7 +158,8 @@ static int check_map(struct map *map)
...
+			__builtin___clear_cache(map->ptr, map->ptr + ONE_MAP_SIZE);
```
avagin, line 162: pls add a comment why we need to cleare cache here

time-river: done

time-river:
> CI is happy. The description sounds reasonable. I have never seen this error but I am not running CRIU very often on aarch64.

This error looks like raise easily in some environment. I ran the case in three environment, only one raised this problem frequently.

Issue #1725, opened 2022-01-17
GCC 12 breaks CRIU


adrianreber:
Fedora rawhide switched to GCC 12 and CRIU build fails at least on aarch64, ppc64le and x86_64:
```
gcc -c -O2 -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-D_FORTIFY_SOURCE=2 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -m64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -O2 -g -Wall -Wformat-security -Wdeclaration-after-statement -Wstrict-prototypes -DCONFIG_X86_64 -D_FILE_OFFSET_BITS=64 -D_GNU_SOURCE -DCONFIG_HAS_LIBBSD -DCONFIG_HAS_SELINUX -DCONFIG_GNUTLS -DCONFIG_HAS_NFTABLES_LIB_API_1 -DCONFIG_COMPAT -iquote include/ -DCONFIG_HAS_LIBBSD -DCONFIG_HAS_SELINUX -DCONFIG_GNUTLS -DCONFIG_HAS_NFTABLES_LIB_API_1 -DCONFIG_COMPAT -I ./compel/include/uapi -fno-strict-aliasing -iquote criu/include -iquote include -iquote images -iquote criu/arch/x86/include -iquote . -I/usr/include/libnl3 -DSYSCONFDIR='"/etc"' -DGLOBAL_CONFIG_DIR='"/etc/criu/"' -DDEFAULT_CONFIG_FILENAME='"default.conf"' -DUSER_CONFIG_DIR='".criu/"' -DCR_NOGLIBC -Wstrict-prototypes -fno-stack-protector -nostdlib -fomit-frame-pointer -fpie -I ./compel/include/uapi -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0 -Wp,-U_FORTIFY_SOURCE -Wp,-D_FORTIFY_SOURCE=0 criu/pie/restorer.c -o criu/pie/restorer.o
<command-line>: warning: "_FORTIFY_SOURCE" redefined
<command-line>: note: this is the location of the previous definition
[...]

PR #1726, opened 2022-01-18
fix GCC 12 failures

adrianreber:
This is a confusing change as it seems the original code was just wrong.
GCC 12 complains with:
```
In function ‘__conv_val’,
    inlined from ‘std_strtoul’ at compel/plugins/std/string.c:202:7:
compel/plugins/std/string.c:154:24: error: array subscript 97 is above array bounds of ‘const char[37]’ [-Werror=array-bounds]
  154 |                 return &conv_tab[__tolower(c)] - conv_tab;
      |                        ^~~~~~~~~~~~~~~~~~~~~~~
compel/plugins/std/string.c: In function ‘std_strtoul’:
compel/plugins/std/string.c:10:19: note: while referencing ‘conv_tab’
   10 | static const char conv_tab[] = "0123456789abcdefghijklmnopqrstuvwxyz";
      |                   ^~~~~~~~
```
Which sounds correct. The array conv_tab has just 37 elements.

If I understand the code correctly we are trying to convert anything
that is character between a-z and A-Z to a number for cases where
the base is larger than 10. For a base 11 conversion b|B should return 11.
For a base 35 conversion z|Z should return 35. This is all for a strtoul()
implementation.

The original code was:

  static const char conv_tab[] = "0123456789abcdefghijklmnopqrstuvwxyz";

  return &conv_tab[__tolower(c)] - conv_tab;

and that seems wrong. If conv_tab would have been some kind of hash it could
have worked, but '__tolower()' will always return something larger than
97 ('a') which will always overflow the array.

But maybe I just don't get that part of the code.

I replaced it with

  return __tolower(c) - 'a' + 10;

which does the right thing: 'A' = 10, 'B' = 11 ... 'Z' = 35

### DIFF
--- a/compel/plugins/std/string.c
+++ b/compel/plugins/std/string.c
@@ -151,7 +151,12 @@ static unsigned int __conv_val(unsigned char c)
 	if (__isdigit(c))
 		return c - '0';
 	else if (__isalpha(c))
-		return &conv_tab[__tolower(c)] - conv_tab;
+		/**
+		 * If we want the value of something which __isalpha() == true
+		 * it has to be base > 10. 'A' = 10, 'B' = 11 ... 'Z' = 35
+		 */
+		return __tolower(c) - 'a' + 10;
+
 	return -1u;
 }
 
--- a/compel/src/main.c
+++ b/compel/src/main.c
@@ -19,6 +19,7 @@
 
 #define CFLAGS_DEFAULT_SET     \
 	"-Wstrict-prototypes " \
+	"-ffreestanding "      \
 	"-fno-stack-protector -nostdlib -fomit-frame-pointer "
 
 #define COMPEL_CFLAGS_PIE   CFLAGS_DEFAULT_SET "-fpie"
--- a/criu/config.c
+++ b/criu/config.c
@@ -229,7 +229,7 @@ out:
 		tmp_string[0] = 0;
 
 	/* Check for unsupported configuration file entries */
-	if (configuration[i] + offset + 1 != 0 && strchr(configuration[i] + offset, ' ')) {
+	if (strchr(configuration[i] + offset, ' ')) {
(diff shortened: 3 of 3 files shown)


adrianreber:
This is the first part of a fix for #1725

adrianreber:
It seems there was the expectation that '-nostdlib' should result in not linking against standard C functions which was wrong. We actually need '-ffreestanding'.

adrianreber:
I mentioned following on fedora-devel to the GCC developers:

> > Thanks for mentioning freestanding compilations. That seems to have been
> > the problem.
> >   
> > On aarch64 I actually got a slightly different error message:
> >
>  >    ld: criu/pie/restorer.o: in function `lsm_set_label':                                  
>  >    /drone/src/criu/pie/restorer.c:174: undefined reference to `strlen'                                    
>                      
>  >  Line 174 is: "for (len = 0; label[len]; len++)"                                    
>   
> > Although there is no direct use of strlen(), it seems GCC 12 uses                                            
> > strlen() for that line which did not happen with GCC 11. Using                                                       
> > '-ffreestanding' makes the compilation work with GCC 12 and CI still                                                      
> > looks happy. Thanks.

On which I got the following answer:

> https://gcc.gnu.org/r12-4283-g6f966f06146be76
> Note, gcc has been doing something similar for years for memcpy and memset.
> -fno-tree-loop-distribute-patterns
> will work too.

https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/KIDESTPUXJVVYRERVF7W26E75ZSZU337/

rst0git (approved):
LGTM

@adrianreber Thank you for fixing it.

mihalicyn (commented):
LGTM

0x7f454c46:
LGTM

PR #1728, opened 2022-01-26
Fix formatting in criu documentation

ashu-mehra:
Signed-off-by: Ashutosh Mehra <asmehra@redhat.com>

### DIFF
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -419,7 +419,7 @@ By default the option is set to *fpu* and *ins*.
     Set the method to be used to validate open files. Validation is done
     to ensure that the version of the file being restored is the same
     version when it was dumped.
-
++
 The 'mode' may be one of the following:
 
     *filesize*:::
@@ -532,7 +532,7 @@ usually need to be escaped from shell.
     Restore cgroups configuration associated with a task from the image.
     Controllers are always restored in an optimistic way -- if already present
     in system, *criu* reuses it, otherwise it will be created.
-
++
 The 'mode' may be one of the following:
 
     *none*:::   Do not restore cgroup properties but require cgroup to
@@ -656,7 +656,7 @@ are not adequate, but this can be suppressed by using *--cpu-cap=none*.
     Set the method to be used to validate open files. Validation is done
     to ensure that the version of the file being restored is the same
     version when it was dumped.
-
++
 The 'mode' may be one of the following:
 
     *filesize*:::


adrianreber:
Thanks for your pull request. Can you be a bit more specific what this changes in your commit message? I all for merging this but I am not sure what `+` exactly does.

adrianreber:
CI is currently unhappy, which is not related to your PR.

ashu-mehra:
When viewing the man page for criu, the indentation for some options is off, making it slightly awkward to follow. Eg:

```
       --manage-cgroups [mode]
           Restore  cgroups  configuration  associated  with a task from the image. Controllers are always restored in an optimistic way — if already present in system, criu reuses it, otherwise it will be
           created.

       The mode may be one of the following:

       none
           Do not restore cgroup properties but require cgroup to pre-exist at the moment of restore procedure.

       props
           Restore cgroup properties and require cgroup to pre-exist.

       soft
           Restore cgroup properties if only cgroup has been created by criu, otherwise do not restore properties. This is the default if mode is unspecified.

       full
           Always restore all cgroups and their properties.

       strict
           Restore all cgroups and their properties from the scratch, requiring them to not present in the system.

       ignore
           Don’t deal with cgroups and pretend that they don’t exist.

           --cgroup-yard path
               Instead of trying to mount cgroups in CRIU, provide a path to a directory with already created cgroup yard. For more information look in the dump section.

           --cgroup-root [controller:]/newroot
               Change the root cgroup the controller will be installed into. No controller means that root is the default for all controllers not specified.

```

I added '+' (that stands for 'list continuation') following the similar use of it in other sections of the document, and that seems to fix the indentation:

```
       --manage-cgroups [mode]
           Restore  cgroups  configuration  associated  with a task from the image. Controllers are always restored in an optimistic way — if already present in system, criu reuses it, otherwise it will be
           created.

           The mode may be one of the following:

           none
               Do not restore cgroup properties but require cgroup to pre-exist at the moment of restore procedure.

           props
               Restore cgroup properties and require cgroup to pre-exist.

           soft
               Restore cgroup properties if only cgroup has been created by criu, otherwise do not restore properties. This is the default if mode is unspecified.

           full
               Always restore all cgroups and their properties.

           strict
               Restore all cgroups and their properties from the scratch, requiring them to not present in the system.

           ignore
               Don’t deal with cgroups and pretend that they don’t exist.

       --cgroup-yard path
           Instead of trying to mount cgroups in CRIU, provide a path to a directory with already created cgroup yard. For more information look in the dump section.

       --cgroup-root [controller:]/newroot
           Change the root cgroup the controller will be installed into. No controller means that root is the default for all controllers not specified.

       --tcp-established
           Restore previously dumped established TCP connections. This implies that the network has been locked between dump and restore phases so other side of a connection simply notice a kind of lag.

       --tcp-close
           Restore connected TCP sockets in closed state.

       --veth-pair IN=OUT
           Correspondence between outside and inside names of veth devices.
```

The `modes` for `manage-cgroups` options are now correctly indented, and so are the options that follow.

rst0git (approved):
LGTM

Thanks for fixing this. It was also mentioned in this discussion: https://github.com/checkpoint-restore/criu/pull/1539#discussion_r673498537

adrianreber:
Thanks.

PR #1730, opened 2022-01-27
pstree: when updating sid for shell job also update matching pgid

Snorch:
If we replace old_sid with current_sid we should also do same
replacement for matching pgid (=old_sid).

 Reported in CRIU gitter by Younes Manton (@ymanton)

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -382,6 +382,9 @@ static int prepare_pstree_for_shell_job(pid_t pid)
 		for_each_pstree_item(pi) {
 			if (pi->sid == old_sid)
 				pi->sid = current_sid;
+
+			if (pi->pgid == old_sid)
+				pi->pgid = current_sid;
 		}
 
 		if (lookup_create_item(current_sid) == NULL)


criu/pstree.c
```diff
@@ -384,6 +384,11 @@ static int prepare_pstree_for_shell_job(pid_t pid)
 				pi->sid = current_sid;
 		}
 
+		for_each_pstree_item(pi) {
+			if (pi->pgid == old_sid)
+				pi->pgid = current_sid;
+		}
+
```
rst0git, line 391: Would it be possible to combine both loops into a single one?

```C
for_each_pstree_item(pi) {
  if (pi->sid == old_sid)
    pi->sid = current_sid;

  if (pi->pgid == old_sid)
    pi->pgid = current_sid;
}
```

Snorch: Sure, done.

rst0git (approved):
LGTM

ymanton (approved):
Works for me. Thanks!

PR #1733, opened 2022-01-27
Modularize zdtm

rst0git:
The zdtm.py file has grown over time to more than 2500 lines. It contains a number of classes, functions and variables all combined into a single monolithic python module.

This patch moves test classes, helper functions, flavors and exceptions into separate files that are imported in zdtm.py. This makes the code easier to extend and maintain in the future.

This pull request is rebased on top of https://github.com/checkpoint-restore/criu/pull/1581.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -7,7 +7,6 @@ from __future__ import (
 )
 
 import argparse
-import atexit
 import datetime
 import errno
 import fcntl
@@ -21,25 +20,33 @@ import shutil
 import signal
 import socket
 import stat
-import string
-import struct
 import subprocess
 import sys
-import tempfile
 import time
-import uuid
-from builtins import input, int, open, range, str, zip
+from builtins import input, int, open, range, str
 
 import yaml
 
 import pycriu as crpc
-from zdtm.criu_config import criu_config
+from zdtm import (
+    TestFailException,
+    TestFailExpectedException,
+    criu_config,
+    decode_flav,
+    encode_flav,
+    flavors,
+    get_freezer,
+    get_test_desc,
+    print_sep,
+    test_classes,
+    test_flag,
+    try_run_hook
+)
 
 # File to store content of streamed images
 STREAMED_IMG_FILE_NAME = "img.criu"
 
 prev_line = None
-uuid = uuid.uuid4()
 
 
 def alarm(*args):
@@ -60,28 +67,6 @@ def traceit(f, e, a):
                 print("+%4d: %s" % (lineno, line.rstrip()))
 
     return traceit
-
-
-# Root dir for ns and uns flavors. All tests
-# sit in the same dir
-tests_root = None
-
-
-def clean_tests_root():
-    global tests_root
-    if tests_root and tests_root[0] == os.getpid():
-        os.rmdir(os.path.join(tests_root[1], "root"))
-        os.rmdir(tests_root[1])
-
-
-def make_tests_root():
-    global tests_root
-    if not tests_root:
-        tests_root = (os.getpid(), tempfile.mkdtemp("", "criu-root-", "/tmp"))
-        atexit.register(clean_tests_root)
-        os.mkdir(os.path.join(tests_root[1], "root"))
-    os.chmod(tests_root[1], 0o777)
-    return os.path.join(tests_root[1], "root")
 
 
 # Report generation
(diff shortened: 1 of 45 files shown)


test/zdtm.py
```diff
@@ -2566,179 +2561,196 @@ def clean_stuff(opts):
...
+                    action='append')
+
+    cp = sp.add_parser("clean", help="Clean something")
+    cp.set_defaults(action=clean_stuff)
+    cp.add_argument("what", choices=['nsroot'])
+
+    return vars(p.parse_args())
+
+
+def collect_zombies(pid):
```
avagin, line 1883: This is not a good name for this function. The main idea of it is to wait for the target process.

test/zdtm.py
```diff
@@ -112,7 +112,7 @@ def add_to_output(path):
         return
 
     output_path = os.path.join(report_dir, "output")
-    with open(path, "r") as fdi, open(output_path, "a") as fdo:
+    with builtins.open(path, "r") as fdi, builtins.open(output_path, "a") as fdo:
```
avagin, line 102: Do we still need this module? I remember we added it to make zdtm.py compatible with py2 and py3. Do we still want to be compatible with py2?

I don't like the idea of this change. At some point, we will need to stop using this module and other py2-compatibility hacks.

test/zdtm.py
rst0git: @Snorch also mentioned that this may not a good name (in https://github.com/checkpoint-restore/criu/pull/1581#discussion_r731994285). I am happy to change it.

@avagin Do you have any suggestions?

test/zdtm.py
rst0git: > Do we still want to be compatible with py2?

The CentOS 7 CI tests are still running with Python 2.

> I don't like the idea of this change. At some point, we will need to stop using this module and other py2-compatibility hacks.

Should I just revert this change or drop py2-compatibility hacks?

@adrianreber What do you think about the Python 2/3 support for zdtm?

adrianreber: > > Do we still want to be compatible with py2?
> 
> The CentOS 7 CI tests are still running with Python 2.
> 
> > I don't like the idea of this change. At some point, we will need to stop using this module and other py2-compatibility hacks.
> 
> Should I just revert this change or drop py2-compatibility hacks?
> 
> @adrianreber What do you think about the Python 2/3 support for zdtm?

From my point of view CentOS/RHEL7 is so old that we do not have to support Python 2 any more. I proposed it at some point but @Snorch said for their use of CRIU they would prefer if Python 2 is still supported to be able to run tests.

test/zdtm.py
Snorch: Note for @avagin: untill we've waited for a target process (which can run for a long time without exiting) we are also collecting zombies from this function, and we only stop collecting them when target process stops, that's why I changed my mind and said that I'm now ok with the name of the function. Also note that previously this code block which collected zombies in this "partial" maner had a comment proving that it meant to be zombies collection there:
```
-    # Fork here, since we're new pidns init and are supposed to
-    # collect this namespace's zombies
```

More over I believe we should review only last patch in this draft PR, other patches are all from the original https://github.com/checkpoint-restore/criu/pull/1581 and should be reviewed there.

> This pull request is rebased on top of #1581.

test/zdtm.py
Snorch: From the first look I failed to run zdtm on VZ7 via python3:

```
make -j5 PYTHON=python3
[root@silo criu]# grep unicode /root/build/criu/test/pycriu/rpc_pb2.py
      has_default_value=False, default_value=unicode("", "utf-8"),
[root@silo criu]# python3 ./test/zdtm.py run -t zdtm/static/env00 --keep-going --ignore-taint
  File "./test/zdtm.py", line 26, in <module>
    import pycriu as crpc
  File "/root/build/criu/test/pycriu/__init__.py", line 1, in <module>
    from . import rpc_pb2 as rpc
  File "/root/build/criu/test/pycriu/rpc_pb2.py", line 184, in <module>
    has_default_value=False, default_value=unicode("", "utf-8"),
NameError: name 'unicode' is not defined
```
even after `pip3 install protobuf` it still somehow generates python2 code in pycriu... I'm not too good in makefiles so probably I'm doing something wrong.

When run make criu on fedora I see `b"".decode('utf-8')` instead of `unicode` there, which helps it to run on python3 but I don't know why.

adrianreber: I think zdtm restarts zdtm. So you have to change the shebang to python3 to make sure it does not use python2

Snorch: I mean that If even I remove python symlink and python2 binary completely, protobuf (when make `PYTHON=python3 -j5`) still generates code in lib/py/rpc_pb2.py which does not run on python3.

rst0git: @Snorch We still have many `#!/usr/bin/python` shebang lines in the tests. The following line makes the tests work in CI with python3:

https://github.com/checkpoint-restore/criu/blob/8b0fabc92dfd5bce25e9e2d1f978100ece5ef254/scripts/ci/run-ci-tests.sh#L61

rst0git: > still generates code in lib/py/rpc_pb2.py which does not run on python3.

This comment explains why:
https://github.com/checkpoint-restore/criu/blob/8b0fabc92dfd5bce25e9e2d1f978100ece5ef254/lib/py/images/Makefile#L6

rst0git: > NameError: name 'unicode' is not defined

I cannot replicate this error on Fedora 35. The file generated with `make PYTHON=python3` doesn't contain `unicode`.
This file is generated with `protoc -I=images/ --python_out=lib/py images/rpc.proto`

test/zdtm.py
avagin: Don't try to convince me that this is a good name. No, it isn't. The name has to describe what function is for.

> Do you have any suggestions?

waitpid_and_rip_zombies

rst0git: I've renamed the function in https://github.com/checkpoint-restore/criu/pull/1581

test/zdtm.py
Snorch: ``` ln -sf /usr/bin/python3 /usr/bin/python ```
does not help either, on Centos7 based environment I still get unicode in there...

rst0git:
Rebased on criu-dev branch and updated PR.

avagin:
@rst0git CI jobs failed.

avagin:
```
===================== Run zdtm/transition/thread-bomb in h =====================
Start test
  File "zdtm.py", line 1939, in <module>
    fork_zdtm()
  File "zdtm.py", line 1930, in fork_zdtm
  File "zdtm.py", line 1077, in do_run_test
    t.start()
  File "/tmp/criu/test/zdtm/zdtm.py", line 118, in start
    self.__make_action('pid', env, self.__flavor.root)
  File "/tmp/criu/test/zdtm/zdtm.py", line 60, in __make_action
    preexec_fn=self.__freezer and self.__freezer.attach or None)
  File "/usr/lib64/python2.7/subprocess.py", line 711, in __init__
    errread, errwrite)
  File "/usr/lib64/python2.7/subprocess.py", line 1327, in _execute_child
    raise child_exception
TypeError: must be unicode, not str

```

avagin:
Tests on CentOS7 stil fail with "TypeError: must be unicode, not str".

rst0git:
> Tests on CentOS7 stil fail with "TypeError: must be unicode, not str".

@avagin Thanks for pointing it out. I fixed this error by adding the following import to all python modules.
```python
from __future__ import unicode_literals
```

Snorch:
```
-        # This Mayakovsky-style code gets list of libraries a binary
-        # needs minus vdso and gate .so-s
-        libs = map(
-            lambda x: x[1] == '=>' and x[2] or x[0],
-            map(
-                lambda x: str(x).split(),
-                filter(
-                    lambda x: not xl.match(x),
-                    map(
-                        lambda x: str(x).strip(),
-                        filter(lambda x: str(x).startswith('\t'),
-                               stdout.decode(
-                                   'ascii').splitlines())))))
-
-        for lib in libs:
-            if not os.access(lib, os.F_OK):
-                raise test_fail_exc("Can't find lib %s required by %s" %
-                                    (lib, binary))
-            self.__copy_one(lib)
+        # This Mayakovsky-style code gets list of libraries a binary
+        # needs minus vdso and gate .so-s
+        libs = map(
+            lambda x: x[1] == '=>' and x[2] or x[0],
+            map(
+                lambda x: builtins.str(x).split(),
+                filter(
+                    lambda x: not xl.match(x),
+                    map(
+                        lambda x: builtins.str(x).strip(),
+                        filter(lambda x: builtins.str(x).startswith('\t'),
+                               ldd.stdout.read().decode(
+                                   'ascii').splitlines())))))
+
+        ldd.wait()
+
+        for lib in libs:
+            if not os.access(lib, os.F_OK):
+                raise TestFailException("Can't find lib %s required by %s" %
+                                        (lib, binary))
+            self.__copy_one(lib)
```
Can we move all additional changes to separate patches with explanations? Why do you add ldd.wait() here? it was not present in the original code. Also I see some changes like `range -> builtins.range`

rst0git:
@Snorch Thank you for the comment. I removed `ldd.wait()` and changed `import builtins` to `from builtins import ...`

Snorch:
Same with "stdout->decode(" -> "ldd.stdout.read().decode(" in the same codeblock.

PR #1736, opened 2022-01-27
ci: set continue-on-error for cross-compile

rst0git:
Running cross compile tests with Debian unstable sometimes fails due to missing or outdated packages.

In this patch we set [continue-on-error](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontinue-on-error) to prevent the GitHub Action workflow run from failing when one or more cross-compile tests with Debian unstable fails.

In other words, we can still see test errors with Debian _unstable_, but the workflow would fail only when the Debian _stable_ tests fail.

img1
img2

### DIFF
--- a/.github/workflows/cross-compile.yml
+++ b/.github/workflows/cross-compile.yml
@@ -6,18 +6,26 @@ jobs:
   build:
 
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


adrianreber:
Sounds good. Thanks.

PR #1737, opened 2022-01-28
Fix building with annobin

adrianreber:
Annobin (used at least in Fedora and RHEL) injects annotation into the compiled objects which break the parasite and restorer.

This removes the annobin flags as used in Fedora and RHEL and makes CRIU work on Fedora and RHEL with annobin enabled.

This change is unfortunately where distribution specific and I do not know if we want to carry it. If there are any other suggestions how to remove this compiler flag I am happy to use other approaches.

I can also carry this patch only downstream in the CRIU packages for Fedora and RHEL.

CC: @rst0git


rst0git:
> I can also carry this patch only downstream in the CRIU packages for Fedora and RHEL.

As this is distribution specific, we could use an env variable and set this variable when building the CRIU packages.

adrianreber:
I am closing this as it is very distribution specific.

Issue #1739, opened 2022-01-31
criu-ns returns non-0 exit code after successful dump


PavloMykhailyshyn:
I noticed one strange error
when I run the dump with criu and it succeeds, criu returns 0
but when I run criu-ns and it succeeds too - return code is now 5
I checked the logs for both criu and criu-ns everything is the same (dump successful)

PavloMykhailyshyn:
@rst0git you did a mistake here
```
def wrap_dump():
    (pid, pid_idx) = get_varg(('-t', '--tree'))
    if pid is None:
        raise OSError(errno.EINVAL, 'No --tree option given')

    set_pidns(pid, pid_idx)
    set_mntns(pid)

    criu_pid = os.fork()
    if criu_pid == 0:
        run_criu(sys.argv[1:])
    return _wait_for_process_status(pid)
```
Instead of passing **criu_pid** to the function `_wait_for_process_status(PID)` you passed PID of the dumping process
that is why criu-ns always returns 5 (-251)

rst0git:
@PavloMykhailyshyn Thank you for pointing this out. I've opened a pull request with a fix: https://github.com/checkpoint-restore/criu/pull/1740

PR #1740, opened 2022-02-01
criu-ns: fix exit code for criu dump

rst0git:
Fixes: #1739

### DIFF
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -68,7 +68,7 @@ def _wait_for_process_status(criu_pid):
         try:
             (pid, status) = os.wait()
             if pid == criu_pid:
-                return os.WEXITSTATUS(status)
+                return os.waitstatus_to_exitcode(status)
         except OSError:
             return -251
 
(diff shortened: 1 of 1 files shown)


Snorch (approved):

scripts/criu-ns
```diff
@@ -203,7 +203,7 @@ def wrap_dump():
     criu_pid = os.fork()
     if criu_pid == 0:
         run_criu(sys.argv[1:])
-    return _wait_for_process_status(pid)
+    return _wait_for_process_status(criu_pid)
```
avagin, line 206: _wait_for_process_status returns 0 if a target process has been killed, doesn't it?

rst0git: When I test it I see the following result:
```
criu-ns dump -t `pidof test` -D /tmp/test -j -v4
echo $?
5
```

Exit code `5` appears because we have `try`/`catch` around `os.wait()`, and `os.wait()` fails as follows:

```
  File "/usr/sbin/criu-ns", line 228, in <module>
    res = wrap_dump()
  File "/usr/sbin/criu-ns", line 203, in wrap_dump
    return _wait_for_process_status(pid)
  File "/usr/sbin/criu-ns", line 68, in _wait_for_process_status
    (pid, status) = os.wait()
ChildProcessError: [Errno 10] No child processes
```

rst0git: > I think we have one more issue in _wait_for_process_status

Thanks for spotting this! You are correct.

PR #1747, opened 2022-02-13
compel: fix how PTRACE_GET_THREAD_AREA errors are handled

rppt:
When PTRACE_GET_THREAD_AREA errors on kernels with
!CONFIG_IA32_EMULATION beacuse of missing support (-EIO), compel should
ignore uch errors in native mode.

However the check for error type uses return value of ptrace rather than
errno, which will always result in error propagation.

Use errno to detect type of error to fix this.

### DIFF
--- a/compel/arch/x86/src/lib/thread_area.c
+++ b/compel/arch/x86/src/lib/thread_area.c
@@ -53,15 +53,16 @@ int __compel_arch_fetch_thread_area(int tid, struct thread_ctx *th)
 		user_desc_t *d = &ptls->desc[i];
 
 		err = ptrace(PTRACE_GET_THREAD_AREA, tid, GDT_ENTRY_TLS_MIN + i, d);
-		/*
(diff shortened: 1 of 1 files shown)


compel/arch/x86/src/lib/thread_area.c
```diff
@@ -59,7 +59,7 @@ int __compel_arch_fetch_thread_area(int tid, struct thread_ctx *th)
 		 * XXX: Add compile CONFIG_X86_IGNORE_64BIT_TLS
 		 * (for x86_64 systems with CONFIG_IA32_EMULATION)
 		 */
-		if (err == -EIO && native_mode)
+		if (errno == EIO && native_mode)
```
avagin, line 62: err == -1 && errno == EIO ...

rppt: makes sense :)

0x7f454c46 (approved):
LGTM

rst0git (approved):
LGTM

PR #1749, opened 2022-02-14
pie: Mark __export_*() functions as externally_visible

0x7f454c46:
GCC's lto source:
> To avoid this problem the compiler must assume that it sees the
> whole program when doing link-time optimization.  Strictly
> speaking, the whole program is rarely visible even at link-time.
> Standard system libraries are usually linked dynamically or not
> provided with the link-time information.  In GCC, the whole
> program option (@option{-fwhole-program}) asserts that every
> function and variable defined in the current compilation
> unit is static, except for function @code{main} (note: at
> link time, the current unit is the union of all objects compiled
> with LTO).  Since some functions and variables need to
> be referenced externally, for example by another DSO or from an
> assembler file, GCC also provides the function and variable
> attribute @code{externally_visible} which can be used to disable
> the effect of @option{-fwhole-program} on a specific symbol.

As far as I read gcc's source, ipa_comdats() will avoid placing symbols
that are either already in a user-defined section or have
externally_visible attribute into new optimized gcc sections.

Signed-off-by: Dmitry Safonov <dima@arista.com>

/* Grunt on clang-indent that doesn't let aligning `__attribute__` with tab to the same indent as attributes above */


adrianreber:
The old compiler on CentOS 7 does not like your change:
```
include/common/compiler.h:34:20: error: missing binary operator before token "("
 #if __has_attribute(__externally_visible__)
```

Issue #1734, opened 2022-01-27
CI test failure: Error (compel/arch/x86/src/lib/infect.c:315): Can't set xstate for 52: Invalid argument"


adrianreber:
There are new error injection errors in CI:
```
+ ./test/zdtm.py run -t zdtm/static/fpu03 --fault 134 -f h --norst
userns is supported
timens isn't supported on 4.18.0-358.el8.x86_64
========================== Run zdtm/static/fpu03 in h ==========================
Start test
./fpu03 --pidfile=fpu03.pid --outfile=fpu03.out
Run criu dump
Forcing 134 fault
=[log]=> dump/zdtm/static/fpu03/52/1/dump.log
------------------------ grep Error ------------------------
b'(00.002679) Set up parasite blob using memfd'
b'(00.002705) Putting parasite blob into 0x7fe3600c6000->0x7f6d3bf38000'
b'(00.002741) Dumping general registers for 52 in native mode'
b'(00.002743) Dumping GP/FPU registers for 52'
b'(00.002753) Error (compel/arch/x86/src/lib/infect.c:299): Corrupting xsave for 52, seed 1643303562'
b"(00.002763) Error (compel/arch/x86/src/lib/infect.c:315): Can't set xstate for 52: Invalid argument"
b"(00.002766) Error (compel/src/lib/infect.c:687): Can't obtain regs for thread 52"
b"(00.002802) Error (criu/cr-dump.c:1303): Can't infect (pid: 52) with parasite"
b'(00.002842) Unlock network'
b'(00.002846) Unfreezing tasks into 1'
b'(00.002847) \tUnseizing 52 into 1'
b'(00.002859) Error (criu/cr-dump.c:1788): Dumping FAILED.'
------------------------ ERROR OVER ------------------------
################### Test zdtm/static/fpu03 FAIL at CRIU dump ###################
Test output: ================================

 <<< ================================
Send the 9 signal to  52
Wait for zdtm/static/fpu03(52) to die for 0.100000
=== Run 1/1 ================ zdtm/static/fpu03
```

On following CPU:
```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
CPU(s):              4
On-line CPU(s) list: 0-3
Thread(s) per core:  2
Core(s) per socket:  2
Socket(s):           1
NUMA node(s):        1
Vendor ID:           GenuineIntel
BIOS Vendor ID:      Google
CPU family:          6
[...]

PR #1750, opened 2022-02-15
Fix Fedora based CI on 5.16 and on Archlinux with glibc 2.35

adrianreber:
5.16 has changes which break CRIU when it comes to BPF and file maps.

This handles a new BPF related field in fdinfo and removes a flock test which is no longer supported.

### DIFF
--- a/images/bpfmap-file.proto
+++ b/images/bpfmap-file.proto
@@ -21,4 +21,5 @@ message bpfmap_file_entry {
 	required string		map_name	= 13;
 	required uint32		ifindex		= 14 [default = 0];
 	optional sint32		mnt_id		= 15 [default = -1];
+	optional uint64		map_extra	= 16;
 }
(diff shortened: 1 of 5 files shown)


rst0git (approved):
LGTM

adrianreber:
The failing CentOS 8 tests are tracked here: #1734

images/bpfmap-file.proto
```diff
@@ -21,4 +21,5 @@ message bpfmap_file_entry {
 	required string		map_name	= 13;
 	required uint32		ifindex		= 14 [default = 0];
 	optional sint32		mnt_id		= 15 [default = -1];
+	optional uint64		map_extra	= 16;
```
avagin, line 24: Is it zero by default? Should we return a warning if it isn't zero? We have to be sure that we restore mappings properly.

adrianreber: Added a warning if it is non-zero.

mihalicyn (approved):
LGTM

PR #1752, opened 2022-02-18
zdtm: fix missplacement of err=True

Snorch:
There is no 'err' argument for print(), it should be in grep_errors() in
line below.

Fixes: bed670f62 ("zdtm: print tails of all logs if a test has failed")


rst0git (approved):
LGTM

PR #1759, opened 2022-03-09
compel: set mxcsr during error injection to zero

adrianreber:
During error injection tests there are random values loaded in some of
the registers. The kernel, however, has the following check:

    if (mxcsr[0] & ~mxcsr_feature_mask)
        return -EINVAL;

So depending on the random values loaded mxcsr might have values that
the kernel rejects with EINVAL. Setting mxcsr to zero during the tests
lets the error injection test pass.

Fixes: #1734

### DIFF
--- a/compel/arch/x86/src/lib/infect.c
+++ b/compel/arch/x86/src/lib/infect.c
@@ -272,6 +272,17 @@ static void validate_random_xstate(struct xsave_struct *xsave)
 
 	/* No reserved bits may be set */
 	memset(&hdr->reserved, 0, sizeof(hdr->reserved));
+
+	/*
+	 * While using PTRACE_SETREGSET the kernel checks that
+	 * "Reserved bits in MXCSR must be zero."
+	 * if (mxcsr[0] & ~mxcsr_feature_mask)
+	 *	return -EINVAL;
+	 *
+	 * As the mxcsr_feature_mask depends on the CPU the easiest solution for
+	 * this error injection test is to set mxcsr just to zero.
+	 */
+	xsave->i387.mxcsr = 0;
 }
 
 /*


compel/arch/x86/src/lib/infect.c
```diff
@@ -306,6 +306,17 @@ static int corrupt_extregs(pid_t pid)
...
+		 * "Reserved bits in MXCSR must be zero."
+		 * if (mxcsr[0] & ~mxcsr_feature_mask)
+		 *	return -EINVAL;
+		 *
+		 * As the mxcsr_feature_mask depends on the CPU the easiest solution for
+		 * this error injection test is to set mxcsr just to zero.
+		 */
+		ext_regs.i387.mxcsr = 0;
+
 		validate_random_xstate((void *)&ext_regs);
```
0x7f454c46, line 320: Could you move it into `validate_random_xstate()`? AFAIR, it's the helper to force random xstate to be valid.

0x7f454c46 (commented):
Seems fine to me, thanks!
It seems a recent change in the kernel that changes "soft-enforcing" by applying the mask over input to "hard-enforcing" by rejecting the input: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=947f4947cf00

compel/arch/x86/src/lib/infect.c
adrianreber: I see you just updated your comment. I was thinking similar. As this is only for error injection tests we could ignore the mask.

0x7f454c46: > Sure. Do you see a problem with my change that I am ignoring the mxcsr_feature_mask and setting the complete mxcsr to zero?

I don't. I think it makes this force-injection test a bit less precise - we won't check that mxcsr flags that may have been corrupted by the parasite dumping are restored on the cure of the task. But we still corrupt everything else on xstate, so I don't see how criu could revert any possible corruption, without reverting it also for mxcsr register.
So, I think the test is still fine.

> I see you just updated your comment. I was thinking similar. As this is only for error injection tests we could ignore the mask.

Yeah, pressed "return", forgot to finish part of the comment :)

adrianreber: Moved everything to `validate_random_xstate()`.

0x7f454c46 (approved):

0x7f454c46:
Thanks for fixing this!

PR #1761, opened 2022-03-09
tun: fix tun_link leak in dump_tun_link

Snorch:
coverity CID 389205:
```
452int dump_tun_link(NetDeviceEntry *nde, struct cr_imgset *fds, struct nlattr **info)
453{
...
458        struct tun_link *tl;
...
   2. alloc_fn: Storage is returned from allocation function get_tun_link_fd. [show details]
   3. var_assign: Assigning: tl = storage returned from get_tun_link_fd(nde->name, nde->peer_nsid, tle.flags).
475        tl = get_tun_link_fd(nde->name, nde->peer_nsid, tle.flags);
   4. Condition !tl, taking false branch.
476        if (!tl)
477                return ret;
478
479        tle.vnethdr = tl->dmp.vnethdr;
480        tle.sndbuf = tl->dmp.sndbuf;
481
482        nde->tun = &tle;
   CID 389205 (#1 of 1): Resource leak (RESOURCE_LEAK)5. leaked_storage: Variable tl going out of scope leaks the storage it points to.
483        return write_netdev_img(nde, fds, info);
484}
```
Function get_tun_link_fd() can both return tun_link entry from tun_links
list and a newly allocated one. So we should not free entry if it is
from list and should free it when it is a new one to fix leak.

### DIFF
--- a/criu/tun.c
+++ b/criu/tun.c
@@ -155,6 +155,7 @@ static struct tun_link *__dump_tun_link_fd(int fd, char *name, unsigned ns_id, u
 		goto err;
 	strlcpy(tl->name, name, sizeof(tl->name));
 	tl->ns_id = ns_id;
+	INIT_LIST_HEAD(&tl->l);
 
 	if (ioctl(fd, TUNGETVNETHDRSZ, &tl->dmp.vnethdr) < 0) {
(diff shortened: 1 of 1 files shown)


adrianreber (approved):
LGTM. The CentOS 8 CI failure should be fixed by #1759

PR #1762, opened 2022-03-09
mount: fix e_str leak in ext_mount_add

Snorch:
coverity CID 389202:
```
54int ext_mount_add(char *key, char *val)
 55{
 56        char *e_str;
 57
   1. alloc_fn: Storage is returned from allocation function malloc.
   2. var_assign: Assigning: ___p = storage returned from malloc(strlen(key) + strlen(val) + 8UL).
   3. Condition !___p, taking false branch.
   4. leaked_storage: Variable ___p going out of scope leaks the storage it points to.
   5. var_assign: Assigning: e_str = ({...; ___p;}).
 58        e_str = xmalloc(strlen(key) + strlen(val) + 8);
   6. Condition !e_str, taking false branch.
 59        if (!e_str)
 60                return -1;
...
   7. noescape: Resource e_str is not freed or pointed-to in sprintf.
 73        sprintf(e_str, "mnt[%s]:%s", key, val);
   8. noescape: Resource e_str is not freed or pointed-to in add_external. [show details]
   CID 389202 (#1 of 1): Resource leak (RESOURCE_LEAK)9. leaked_storage: Variable e_str going out of scope leaks the storage it points to.
 74        return add_external(e_str);
 75}
```
We need to free e_str after add_external used it.

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -55,7 +55,7 @@ static LIST_HEAD(delayed_unbindable);
 
 int ext_mount_add(char *key, char *val)
 {
-	char *e_str;
+	cleanup_free char *e_str = NULL;
 
 	e_str = xmalloc(strlen(key) + strlen(val) + 8);
 	if (!e_str)


adrianreber (approved):
You could also use `cleanup_free` which should automatically do the free like `cleanup_free char *e_str`.

Snorch:
> You could also use `cleanup_free` which should automatically do the free like `cleanup_free char *e_str`.

Ok, let's go cleanup_free way here.

PR #1764, opened 2022-03-10
sk-unix: fix uint32_t id variable printf format specifier

Snorch:
coverity CID 389193:
```
CID 389193 (#1 of 1): Printf format string issue (PW.BAD_PRINTF_FORMAT_STRING)
1. bad_printf_format_string: invalid format string conversion
598 pr_warn("Can't stat socket %#x(%s), skipping: %m (err %d)\n", id, rpath, errno);
```
Specifier "%#x" is wrong for id as it is of type uint32_t, let's change
it to "%#" PRIx32 "" to fix the problem.


rst0git (approved):
LGTM

PR #1768, opened 2022-03-11
uffd: fix __u64 print format specifier

Snorch:
coverity CID 389197:

CID 389197 (#1 of 1): Invalid printf format string (PRINTF_ARGS)
format_error: Length modifier L not applicable to conversion specifier in %Lu. [show details]
284 pr_err("Incompatible uffd API: expected %Lu, got %Lu\n", UFFD_API, uffdio_api.api);

Looking on C11 standard it seems that "%Lu" is undefined, we better not
use this, see:

"L Specifies that a following a, A, e, E, f, F, g, or G conversion
specifier applies to a long double argument."
http://port70.net/~nsz/c/c11/n1570.html#7.21.6.1p7

### DIFF
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -281,7 +281,7 @@ int uffd_open(int flags, unsigned long *features, int *err)
 	}
 
 	if (uffdio_api.api != UFFD_API) {
-		pr_err("Incompatible uffd API: expected %Lu, got %Lu\n", UFFD_API, uffdio_api.api);
+		pr_err("Incompatible uffd API: expected %llu, got %llu\n", UFFD_API, uffdio_api.api);
 		goto close;
 	}
 


criu/uffd.c
```diff
@@ -281,7 +281,7 @@ int uffd_open(int flags, unsigned long *features, int *err)
 	}
 
 	if (uffdio_api.api != UFFD_API) {
-		pr_err("Incompatible uffd API: expected %Lu, got %Lu\n", UFFD_API, uffdio_api.api);
+		pr_err("Incompatible uffd API: expected %llu, got %llu\n", UFFD_API, uffdio_api.api);
```
avagin, line 284: should we use PRI64 in this case?

Snorch: Ok, I see:
```
typedef u64			uint64_t;
```
https://github.com/torvalds/linux/blob/master/include/linux/types.h#L107
```
typedef __u64 u64;
```
https://github.com/torvalds/linux/blob/master/include/asm-generic/int-ll64.h#L23

Probably it's enough proof.

Snorch: Heh, but tests don't agree =)
```
In file included from criu/include/criu-log.h:23,
                 from criu/uffd.c:22:
criu/uffd.c: In function ‘uffd_open’:
criu/include/log.h:49:52: error: format ‘%lu’ expects argument of type ‘long unsigned int’, but argument 5 has type ‘long long unsigned int’ [-Werror=format=]
   49 | #define pr_err(fmt, ...) print_on_level(LOG_ERROR, "Error (%s:%d): " LOG_PREFIX fmt, __FILE__, __LINE__, ##__VA_ARGS__)
      |                                                    ^~~~~~~~~~~~~~~~~
criu/uffd.c:284:3: note: in expansion of macro ‘pr_err’
  284 |   pr_err("Incompatible uffd API: expected %" PRIu64 ", got %" PRIu64 "\n", UFFD_API, uffdio_api.api);
      |   ^~~~~~
```
So I'd rather revert it to `%llu` variant.

Issue #1774, opened 2022-03-14
cgroup related tests seem to be failing more often in CI


adrianreber:
```
2022-03-11T14:49:33.7079636Z b'(00.022093) cg: adding cgroup /proc/self/fd/15/docker/41af214ff35599b069c697d142ceea1813ac16aceb1cc6b389cdc72d72120cce'
2022-03-11T14:49:33.7080078Z b'(00.022301) cg: adding cgroup /proc/self/fd/15/'
2022-03-11T14:49:33.7080558Z b'(00.022429) cg: adding cgroup /proc/self/fd/15/docker/41af214ff35599b069c697d142ceea1813ac16aceb1cc6b389cdc72d72120cce'
2022-03-11T14:49:33.7081009Z b'(00.022557) cg: adding cgroup /proc/self/fd/15/subcg00/subsubcg'
2022-03-11T14:49:33.7081501Z b'(00.040339) Error (criu/cgroup.c:587): cg: Unable to create the cgroup (name=zdtmtst.defaultroot) file system: Resource busy'
2022-03-11T14:49:33.7082051Z b'(00.040374) Error (criu/cgroup.c:715): cg: failed walking /proc/self/fd/-1/ for empty cgroups: No such file or directory'
```

Recently I have seen more test failures concerning cgroup in CI. CI seems to have changed to a 5.13 although I am not sure that is related.

The error is easy to reproduce and I can see it locally about 10% of the time.

It seems to not happen if I disabled the new mount API support.

I see that `ret = sys_fsconfig(fsfd, FSCONFIG_CMD_CREATE, NULL, NULL, 0);` sometimes returns `EBUSY`. I tried to wrap it in a loop to retry if it fails with `EBUSY`, but the second time it then fails with `EINVAL`. For some reason the kernel behaves differently if I repeat the following steps: `fsopen(), fsconfig(), fsconfig(), close()`.

avagin:
@mihalicyn could you take a look at this?

minhbq-99:
I try to investigate this issue and find out that in the error case

https://github.com/checkpoint-restore/criu/blob/cc4df23fa53429d351a79ba0217a4a3f3944759f/test/zdtm_mount_cgroups#L7

[...]

PR #1780, opened 2022-03-21
scripts/ci: create test cgroup controllers once

avagin:
zdtm.py mounts two named controllers for tests. In CI, we run zdtm.py a few
times, so we can mount (create) these controllers once from the ci script to
avoid any unwanted effects.

Fixes: #1774

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -139,6 +139,13 @@ time make unittest
 
 ulimit -c unlimited
 
+cgid=$$
+cleanup_cgroup() {
+	./test/zdtm_umount_cgroups $cgid
+}
+trap cleanup_cgroup EXIT
+./test/zdtm_mount_cgroups $cgid
+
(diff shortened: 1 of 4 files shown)


adrianreber (approved):
I always had a suspicion that it might be something like this, but I was able to reproduce it locally without any other zdtm process running in parallel.

From what I have seen it is less about two zdtm running in parallel and more about a cgroup being lazily removed from the kernel and while zdtm re-creates the cgroup the kernel still removes the previous cgroup.

This sounds like something that could help and so far CI is happy.

LGTM as it seems to provide a fix for the problem we are seeing.

avagin:
> From what I have seen it is less about two zdtm running in parallel and more about a cgroup being lazily removed from the kernel and while zdtm re-creates the cgroup the kernel still removes the previous cgroup.

Yes, it looks like this and it is why the second patch is here. The case when we create and remove cgroup controllers back and forth isn't common in real life. Usually, controllers are created once and are never destroyed.

PR #1786, opened 2022-03-23
Style: delete some redundant code

jwcesign:
### Delete some redundant code
* I am new here to learn criu, here is some codereview fix.


rst0git (approved):
LGTM

Issue #1775, opened 2022-03-14
Dump fails with the error `All tasks in criu's cgroups. Nothing to dump.`


PavloMykhailyshyn:
I ran into a problem during process dump.
I am starting a process in a separate session using setsid
I use criu-ns to perform dump and restore.
The dump did not work as expected after several times of successful dumping and restoring the same process.
The dump fails with such log `All tasks in criu's cgroups. Nothing to dump`.
Everything works well before this failed dump.
What could be the reason for such a problem?

PavloMykhailyshyn:
@Snorch and @rst0git

PavloMykhailyshyn:
here are the logs for this particular failed dump
[logs.txt](https://github.com/checkpoint-restore/criu/files/8251858/logs.txt)

PavloMykhailyshyn:
and I do not understand this comment
	/*
	 * Check whether root task lives in its own set as compared
	 * to criu. If yes, we should not dump anything. Note that
	 * list_is_singular() is slightly wrong here: if the criu cgset has
	 * empty cgroups, those will not be restored on the target host, since
	 * we're not dumping anything here.
	 */
Is everything ok in regards to this?

avagin:
"All tasks in criu's cgroups. Nothing to dump." is not an error message... I don't see any errors in the log, so we need to find a code path where we forgot to print an error.

avagin:
```
(00.281640) Writing image inventory (version 1)
(00.281729) Running post-dump scripts
(00.281732) Unfreezing tasks into 2
(00.281735) 	Unseizing 352481 into 2
```

"Running post-dump scripts" means that we dumped processes successfully.

PavloMykhailyshyn:
@avagin it didn't crash, but exited with a status code of 1 (probably, but the point is that it's not 0)

avagin:
@PavloMykhailyshyn could you add more debug in the criu code and find out why it exits with a non-zero code?

PavloMykhailyshyn:
What do you mean?
Add more logs and try to debug?

avagin:
I mean something like this:
https://github.com/avagin/criu/commit/749ce82f16cf2be9ec3e07e8030a5c978d5f00e0

PavloMykhailyshyn:
Thanks
Will try

pr_err - does this write info to the file with the provided argument —log-file?

avagin:
You will see this messages in the log file.

rst0git:
[...]

PR #1787, opened 2022-03-23
criu-ns: add a helper to hold a pid namespace

avagin:
The init process can exit if it doesn't have any child processes and its
pidns is destroyed in this case. CRIU dump is running in the target pid
namespace and it kills dumped processes at the end. We need to create a
holder process to be sure that the pid namespace will not be destroy
before criu exits.

Fixes: #1775


rst0git (approved):
LGTM

PR #1790, opened 2022-03-26
mount: fix -Wunused-but-set-variable for Clang 15

MaskRay:
Since https://reviews.llvm.org/D122271, Clang -Wset-but-unused-variable
gets smarter to warn about unused post-increments.


adrianreber:
git commit author and signed-off-by are different can you fix that please

adrianreber:
Looks good and I understand the change. @Snorch and @avagin you are more familiar with `mount.c` any other ideas how to better make the compiler happy or does this change work for you?

Snorch (approved):
Looks good.

PR #1792, opened 2022-03-28
sk-unix: fix bind_on_deleted() return code

azhadchenko:
Commit d0308e5ecc1c ("sk-unix: make criu respect existing files while restoring ghost unix socket fd") broke return codes for bind_on_deleted(). Not only function will always return 0, but setting ret is pointless since it will be overriden by unlink() and rename()
~~Fix this by adding new err variable to store first-encountered error.~~
Simpilfy return codes and errno juggling. No one is relying on it. Also all encountered errors already print some info.

### DIFF
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -1472,7 +1472,7 @@ static int bind_on_deleted(int sk, struct unix_sk_info *ui)
 	char path[PATH_MAX], path_parked[PATH_MAX], *pos;
 	struct sockaddr_un addr;
 	bool renamed = false;
-	int ret;
+	int ret, exit_code = -1;
 
 	if (ui->ue->name.len >= UNIX_PATH_MAX) {
 		pr_err("ghost: Too long name for socket id %#x ino %u name %s\n", ui->ue->id, ui->ue->ino, ui->name);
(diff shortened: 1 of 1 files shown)


criu/sk-unix.c
```diff
@@ -1472,7 +1472,7 @@ static int bind_on_deleted(int sk, struct unix_sk_info *ui)
 	char path[PATH_MAX], path_parked[PATH_MAX], *pos;
 	struct sockaddr_un addr;
 	bool renamed = false;
-	int ret;
+	int ret, err = 0;
```
avagin, line 1475: In CRIU, we usually call it exit_code:
$ git grep exit_code | wc -l
174

azhadchenko:
> you should not use ret here. exit_code has to be set in error paths.
It is set in error paths.
However even when we have encounter error, we still need to rename socket back (if it was renamed) and maybe delete folders. So far my with my commit function do not override exit_code with errors encountered while renaming back or deleting folders. Do you want me to overwrite exit code even in this cases?

azhadchenko:
Reworked p/r. Use exit_code for storing errors. Simpilfy return codes and errno juggling. No one is relying on it. Also all encountered errors already print some info.

PR #1796, opened 2022-03-31
Add codespell CI job, fix or ignore existing warnings

kolyshkin:
_This is something I always wanted to do but never got around to as it's too big and daunting. Found some courage today, so here you are._

This set adds a codespell job to CI (as well as to `lint` make target), so that if any new PRs will add typos, they will be found.

Existing issues reported by codespell are dealt with using one of the following ways:

1. Where it makes sense, some variables are renamed to make codespell happy. This is done for variables whose names do not make much sense or look like a typos. For example, `sie` (siginfo entry), which looks like a misspelled `size`, is renamed to `si`.

2. Some words are added to codespell ignore list. Those are: `creat,fpr,fle,ue,bord,parms,nd,te,testng`. For every word added, the motivation is different -- either it's too much to change everything (for example, `parms` -> `params`), or the word is a legitimate acronym (FPR), or the word is a (badly chosen) project name used by our code (`testng`).
⚠️ One particular issue is with the word `useable` (which is a misspell of `usable`). This is a structure field, and I am not sure if we can change its name without breaking anything (such as backward compatibility). I am leaving that decision to reviewers.

3. The remaining issues are fixed using whatever codespell thinks is best.


criu/cr-restore.c
```diff
@@ -3081,11 +3081,11 @@ static int prepare_rlimits(int pid, struct task_restore_args *ta, CoreEntry *cor
 	return 0;
 }
 
-static int signal_to_mem(SiginfoEntry *sie)
+static int signal_to_mem(SiginfoEntry *se)
```
adrianreber, line 3084: This change looks unnecessary, but not a blocker from my side.

adrianreber: Ah, now I see the commit message. Works for me.

adrianreber:
I like it. Automating things is always good. Some of the variable renaming seems unnecessary, but I understand it now and it works for me.

Could you make sure the changes in `flog/` are a separate commit. We do not yet merge changes from `flog/` to the release branch and if it is a separate commit it makes releases much easier if we do not have to split commits.

The `usable` change should be no problem. It touches protobuf definitions, but if the ID does not change (which it doesn't) we are free the rename it however we like.

images/core-x86.proto
```diff
@@ -99,7 +99,7 @@ message user_desc_t {
 	required bool			read_exec_only	=  7 [default = true];
 	required bool			limit_in_pages	=  8;
 	required bool			seg_not_present	=  9 [default = true];
-	required bool			useable		= 10;
+	required bool			usable		= 10;
```
rst0git, line 102: Would this change break backwards compatibility?

rst0git (approved):
LGTM

kolyshkin:
> Could you make sure the changes in `flog/` are a separate commit. We do not yet merge changes from `flog/` to the release branch and if it is a separate commit it makes releases much easier if we do not have to split commits.

* Separated flog changes into separate commits (two of them, marked with flog: prefix in the subject)
* Rebased to current tip of criu-dev (no issues of new codespell warnings)

PTAL @adrianreber

kolyshkin:
Failure in Vagrant Fedora cirrus ci job (https://cirrus-ci.com/task/4505414497730560), probably unrelated.
```
================== Run zdtm/static/socket-tcp-reuseport in ns ==================
Start test
./socket-tcp-reuseport --pidfile=socket-tcp-reuseport.pid --outfile=socket-tcp-reuseport.out
Run criu dump
Run criu restore
Send the 15 signal to  251
########## Test zdtm/static/socket-tcp-reuseport FAIL at result check ##########
22:10:37.277:     4: ERR: tcp.c:95: accept() failed (errno = 11 (Resource temporarily unavailable))
22:10:37.373:     4: ERR: tcp.c:95: accept() failed (errno = 11 (Resource temporarily unavailable))
22:10:37.373:     4: ERR: tcp.c:95: accept() failed (errno = 11 (Resource temporarily unavailable))
22:10:37.373:     4: ERR: tcp.c:95: accept() failed (errno = 22 (Invalid argument))
22:10:37.373:     4: ERR: socket-tcp-reuseport.c:160: can't accept client connection 1
userns is supported
```

avagin:
LGTM

adrianreber:
`zdtm/static/socket-tcp-reuseport` is known to fail sometimes. LGTM

Issue #1798, opened 2022-03-31
Error when dumping process with same PID in different containers simultaneously with host network namespace


new-sashok724:
Hello! When i try to dump processes with same PID in different containers simultaneously, CRIU sometimes fails to bind abstract unix socket because it is bound by some other CRIU instance in different container, but same host network namespace. This bug happens only with `--network=host`.

I already reported this bug with CRIU restore (#1735), and it is fixed now (Thanks!), but same problem still happens with dump. Maybe more unique name should be generated for dump socket too :)

```
(00.022193) Putting tsock into pid 17
(00.022197) Error (compel/src/lib/infect.c:399): Can't bind socket: Address already in use
(00.022239) Error (criu/cr-dump.c:1303): Can't infect (pid: 17) with parasite
(00.022318) Unlock network
(00.022323) Unfreezing tasks into 1
(00.022326)     Unseizing 17 into 1
(00.022475) Error (criu/cr-dump.c:1786): Dumping FAILED.
```

Version: 3.16.1
GitID: v3.16.1-255-g68477be19

PR #1799, opened 2022-04-01
criu: generate unique socket names

avagin:
CRIU has a few places where it creates unix sockets and their names have to be
unique for each criu run.

Fixes: #1798


adrianreber (approved):

rst0git (approved):
LGTM

PR #1800, opened 2022-04-03
cgroup: fix --manage-cgroups=ignore

adrianreber:
Using `--manage-cgroups=ignore` was still trying to restore cgroup properties at certain places.

Restoring a OCI container in a different cgroup (than the one it was checkpointed from) resulted in CRIU trying to put the container in the original cgroup but the container runtime (runc, crun) was putting it another cgroup. A second checkpoint of the container with two different cgroup settings did no longer work.

With this change CRIU really completely ignores the cgroup and the container can be checkpointed and restored multiple times even if the cgroup changes.

This PR includes a test to verify  `--manage-cgroups=ignore'` works as expected.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1217,6 +1217,9 @@ int prepare_task_cgroup(struct pstree_item *me)
 	CgSetEntry *se;
 	u32 current_cgset;
 
+	if (opts.manage_cgroups == CG_MODE_IGNORE)
+		return 0;
+
 	if (!rsti(me)->cg_set)
 		return 0;
 
@@ -1298,6 +1301,9 @@ static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *pat
 {
 	int cg, fd, ret = -1;
 	CgroupPerms *perms = cg_prop_entry_p->perms;
+
+	if (opts.manage_cgroups == CG_MODE_IGNORE)
+		return 0;
 
 	if (!cg_prop_entry_p->value) {
 		pr_err("cg_prop_entry->value was empty when should have had a value\n");
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1035,7 +1035,7 @@ static int ctrl_dir_and_opt(CgControllerEntry *ctl, char *dir, int ds, char *opt
 		}
 
 		if (n[0] == 0)
-			doff += snprintf(dir + doff, ds - doff, "unified");
+			doff += snprintf(dir + doff, ds - doff, "unified,");
 		else
 			doff += snprintf(dir + doff, ds - doff, "%s,", n);
 		if (opt)
--- a/test/zdtm/static/cgroup_ignore.c
+++ b/test/zdtm/static/cgroup_ignore.c
@@ -1,0 +1,161 @@
+#include <unistd.h>
+#include <stdio.h>
+#include <string.h>
+#include <fcntl.h>
+#include <stdlib.h>
+#include <sys/stat.h>
+#include <sys/mount.h>
+#include "zdtmtst.h"
+
+const char *test_doc = "Check that cgroups are correctly ignored";
+const char *test_author = "Adrian Reber <areber@redhat.com>";
+
+char *dirname;
+TEST_OPTION(dirname, string, "cgroup directory name", 1);
+static const char *cgname = "zdtmtst";
+
+static size_t read_all(int fd, char *buf, size_t size)
+{
+	ssize_t r = 0, ret;
+
+	while (r < size) {
+		ret = read(fd, buf + r, size - r);
+		if (ret < 0) {
+			pr_perror("Read failed");
+			return -1;
+		} else if (ret == 0) {
+			return 0;
+		}
+		r += ret;
+	}
+
+	return 0;
+}
+
+int main(int argc, char **argv)
+{
+	cleanup_free char *cgroup_procs = NULL;
+	cleanup_close int cgroup_procs_fd = -1;
+	cleanup_free char *destination = NULL;
+	cleanup_free char *buffer_old = NULL;
+	cleanup_free char *buffer_new = NULL;
+	cleanup_close int fd = -1;
+	int ret = 1;
+
+	test_init(argc, argv);
+
+	buffer_old = malloc(PAGE_SIZE);
+	if (!buffer_old) {
+		pr_err("Could not allocate memory\n");
+		return 1;
+	}
+	memset(buffer_old, 0, PAGE_SIZE);
+	buffer_new = malloc(PAGE_SIZE);
+	if (!buffer_new) {
+		pr_err("Could not allocate memory\n");
+		return 1;
+	}
+	memset(buffer_new, 0, PAGE_SIZE);
+
+	// Read /proc/self/cgroup to later compare against it
+	fd = open("/proc/self/cgroup", O_RDONLY);
+	if (fd < 0) {
+		pr_err("Could not open /proc/self/cgroup\n");
+		return 1;
+	}
+
+	if (read_all(fd, buffer_old, PAGE_SIZE)) {
+		pr_err("Could not read data from /proc/self/cgroup\n");
+		return 1;
+	}
+
+	// Create the cgroup root directory
+	if (mkdir(dirname, 0700) < 0 && errno != EEXIST) {
+		pr_err("Cannot make directory %s\n", dirname);
+		return 1;
+	}
+
+	// Mount cgroup2, skip if cgroup2 is not available
+	if (mount("none", dirname, "cgroup2", 0, 0)) {
+		if (errno == ENODEV) {
+			skip("Test relies on cgroup2 semantics which this system does not support. Skipping");
+			test_daemon();
+			test_waitsig();
+			pass();
+			return 0;
+		} else {
+			pr_err("Could not mount cgroup2 at %s\n", dirname);
+		}
+		return 1;
+	}
+
+	// Create the cgroup cgname (if it does not already exist)
+	if (asprintf(&destination, "%s/%s", dirname, cgname) == -1) {
+		pr_err("Could not allocate memory\n");
+		goto err;
+	}
+	if (mkdir(destination, 0700) < 0 && errno != EEXIST) {
+		pr_err("Failed to create temporary cgroup directory %s\n", destination);
+		goto err;
+	}
+
+	// Move this process to the newly created cgroup
+	if (asprintf(&cgroup_procs, "%s/cgroup.procs", destination) == -1) {
+		pr_err("Could not allocate memory\n");
+		goto err;
+	}
+	cgroup_procs_fd = open(cgroup_procs, O_RDWR);
+	if (cgroup_procs_fd < 0) {
+		pr_err("Could not open %s\n", cgroup_procs);
+		goto err;
+	}
+	if (write(cgroup_procs_fd, "0", 1) != 1) {
+		pr_err("Writing to %s failed\n", cgroup_procs);
+		goto err;
+	}
+
+	// Read /proc/self/cgroup (should have changed)
+	lseek(fd, 0, SEEK_SET);
+	if (read_all(fd, buffer_new, PAGE_SIZE)) {
+		pr_err("Could not read data from /proc/self/cgroup\n");
+		goto err;
+	}
+
+	// Test if /proc/self/cgroup has changed
+	if (!memcmp(buffer_new, buffer_old, PAGE_SIZE)) {
+		fail("/proc/self/cgroup should differ after move to another cgroup");
+		pr_err("original /proc/self/cgroup content %s\n", buffer_old);
+		pr_err("new /proc/self/cgroup content %s\n", buffer_new);
+		goto err;
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	// Read /proc/self/cgroup. It should not be the same as after
+	// moving this process to another cgroup because of restore
+	// with '--manage-cgroups=ignore'. The process should be
+	// now in cgroup of the current session.
+	lseek(fd, 0, SEEK_SET);
+	memset(buffer_old, 0, PAGE_SIZE);
+	if (read_all(fd, buffer_old, PAGE_SIZE)) {
+		pr_err("Could not read data from /proc/self/cgroup\n");
+		goto err;
+	}
+
+	// Test if /proc/self/cgroup has changed again
+	if (!memcmp(buffer_new, buffer_old, PAGE_SIZE)) {
+		fail("/proc/self/cgroup should differ after restore");
+		pr_err("original /proc/self/cgroup content %s\n", buffer_new);
+		pr_err("new /proc/self/cgroup content %s\n", buffer_old);
+		goto err;
+	}
+
+	ret = 0;
+	pass();
+err:
+	rmdir(destination);
+	umount(dirname);
+
+	return ret;
+}
--- a/test/jenkins/crit.sh
+++ b/test/jenkins/crit.sh
@@ -2,6 +2,6 @@
 set -e
 source `dirname $0`/criu-lib.sh
 prep
-./test/zdtm.py run --all -f best -x maps04 -x cgroup02 --norst --keep-img always || fail
+./test/zdtm.py run --all -f best -x maps04 -x cgroup02 -x cgroup_ignore --norst --keep-img always || fail
 PYTHONPATH="$(pwd)/lib/" ./test/crit-recode.py || fail
 exit 0
--- a/test/jenkins/criu-dump.sh
+++ b/test/jenkins/criu-dump.sh
@@ -5,4 +5,4 @@ set -e
 source `dirname $0`/criu-lib.sh
 prep
 mount_tmpfs_to_dump
-./test/zdtm.py run --all --keep-going --report report --parallel 4 --norst -x 'maps04' -x 'cgroup02' || fail
+./test/zdtm.py run --all --keep-going --report report --parallel 4 --norst -x 'maps04' -x 'cgroup02' -x 'cgroup_ignore' || fail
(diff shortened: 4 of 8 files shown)


criu/cgroup.c
```diff
@@ -1217,6 +1217,9 @@ int prepare_task_cgroup(struct pstree_item *me)
 	CgSetEntry *se;
 	u32 current_cgset;
 
+	if (!opts.manage_cgroups)
```
avagin, line 1220: == CG_MODE_IGNORE

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,157 @@
...
+
+int main(int argc, char **argv)
+{
+	cleanup_close int fd = -1;
+	cleanup_close int cgroup_procs_fd = -1;
+	cleanup_free char *buffer_old = NULL;
+	cleanup_free char *buffer_new = NULL;
+	cleanup_free char *destination = NULL;
+	cleanup_free char *cgroup_procs = NULL;
+	cleanup_free char *own_pid = NULL;
```
avagin, line 43: imho: cleanup_free doesn't look good. In this case, you probably can use alloca(), or declare fixed size arrays:
char buffer_new[PAGE_SIZE];

avagin:
You need to add a commit message for the test. It has to describe what test does.

rst0git (approved):
LGTM

I can confirm that the patch in this PR fixes https://github.com/containers/podman/issues/13672

criu/cgroup.c
adrianreber: done

adrianreber:
> > Sorry, I don't understand this.
> 
> It was my type. I mean it can have a local hierarchy. Do you mount cgroups into a container? Can processes in a container create new cgroups?

Probably yes, but so far no one seems to be using it in OCI containers, because until now I have not seen any reports of people complaining about non working cgroups inside of the container.

During `run` in Podman the cgroup limit is specified. During restore this limit is again set in the cgroup and the container is restored into that newly created cgroup and the resulting container has the same limits as the original containers.

cgroup limits which are set during runtime of the container using `exec` are not preserved this way, but I am not sure anybody really uses it in OCI containers.

test/others/rpc/Makefile
```diff
@@ -22,8 +22,13 @@ run: all
 		$(PYTHON) read.py build/status"
 	rm -f build/status
 	chmod a+rw build/pidfile
-	sudo -g '#1000' -u '#1000' ./run.sh
+	sudo -g '#1000' -u '#1000' ./run.sh || true
```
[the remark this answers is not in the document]
adrianreber: yes, will remove it

adrianreber:
Anyway, the fix for CRIU is to correctly work if `ignore` is specified. Currently CRIU does not correctly handle `ignore`.

What the container engine does is not really important here.

avagin:
> Anyway, the fix for CRIU is to correctly work if ignore is specified. Currently CRIU does not correctly handle ignore.

It works how it was designed by its author. It doesn't restore a cgroup tree, but it moves tasks in their cgroups. This behaviour was introduced in commit f95b05eb75baba45c8507008c70283aca90e6087:
```
Date:   Wed Aug 6 22:23:00 2014 +0400

    opts: add --manage-cgroups option

    criu managed cgroups is now an opt-in thing, so by default criu does not manage
    (i.e. dump or restore) cgroups. This allows users to use the previous behavior.
```

adrianreber:
> > Anyway, the fix for CRIU is to correctly work if ignore is specified. Currently CRIU does not correctly handle ignore.
> 
> It works how it was designed by its author. It doesn't restore a cgroup tree, but it moves tasks in their cgroups. This behaviour was introduced in commit [f95b05e](https://github.com/checkpoint-restore/criu/commit/f95b05eb75baba45c8507008c70283aca90e6087):
> 
> ```
> Date:   Wed Aug 6 22:23:00 2014 +0400
> 
>     opts: add --manage-cgroups option
> 
>     criu managed cgroups is now an opt-in thing, so by default criu does not manage
>     (i.e. dump or restore) cgroups. This allows users to use the previous behavior.
> ```

No, I do not agree. According to the man page `ignore` should `Don’t deal with cgroups and pretend that they don’t exist.` and if I restore a process with `ignore` that does not happen:
```
# sleep infinity &
# cat /proc/`pidof sleep`/cgroup
0::/user.slice/user-0.slice/session-1.scope
# mkdir /sys/fs/cgroup/test1
# echo `pidof sleep` > /sys/fs/cgroup/test1/cgroup.procs 
# cat /proc/`pidof sleep`/cgroup
# criu dump -t `pidof sleep` -j -D /tmp/2 --manage-cgroups=ignore
# criu restore -j -D /tmp/2 -v4 -d --manage-cgroup=ignore
(00.000005) Version: 3.16.1 (gitid 0)
(00.000037) Running on fedora02 Linux 5.16.18-200.fc35.x86_64 #1 SMP PREEMPT Mon Mar 28 14:10:07 UTC 2022 x86_64
(00.000130) Loaded kdat cache from /run/criu/criu.kdat
(00.000149) rlimit: RLIMIT_NOFILE unlimited for self
(00.000509) cpu: x86_family 25 x86_vendor_id AuthenticAMD x86_model_id AMD Ryzen 5 5600G with Radeon Graphics         
(00.000561) cpu: fpu: xfeatures_mask 0x205 xsave_size 2440 xsave_size_max 2440 xsaves_size 840
(00.000590) cpu: fpu: x87 floating point registers     xstate_offsets      0 / 0      xstate_sizes    160 / 160   
(00.000598) cpu: fpu: AVX registers                    xstate_offsets    576 / 576    xstate_sizes    256 / 256   
(00.000611) cpu: fpu: Protection Keys User registers   xstate_offsets   2432 / 832    xstate_sizes      8 / 8     
(00.000616) cpu: fpu:1 fxsr:1 xsave:1 xsaveopt:1 xsavec:1 xgetbv1:1 xsaves:1
(00.000708) kernel pid_max=4194304
(00.000741) Reading image tree
(00.000906) Add mnt ns 6 pid 769
(00.000921) Add net ns 2 pid 769
(00.000924) Add pid ns 1 pid 769
(00.000929) pstree pid_max=769
(00.000938) Will restore in 0 namespaces
(00.000949) NS mask to use 0
(00.001031) Collecting 51/56 (flags 3)
(00.001121) No memfd.img image
(00.001146)  `- ... done
(00.001174) Collecting 40/54 (flags 2)
(00.001252) Collected [usr/bin/sleep] ID 0x1
(00.001280) Collected [usr/lib/locale/locale-archive] ID 0x2
(00.001304) Collected [usr/lib64/libc.so.6] ID 0x3
(00.001326) Collected [usr/lib64/ld-linux-x86-64.so.2] ID 0x4
(00.001363) Collected [dev/pts/0] ID 0x6
(00.001439) Collected [tmp] ID 0x7
(00.001496) Collected [.] ID 0x8
(00.001518)  `- ... done
(00.001529) Collecting 46/68 (flags 0)
(00.001536) No remap-fpath.img image
(00.001581)  `- ... done
(00.001771) No apparmor.img image
(00.001897) Running pre-restore scripts
(00.002084) No pidns-1.img image
(00.002246) Forking task with 769 pid (flags 0x0)
(00.002295) Creating process using clone3()
(00.002717) PID: real 769 virt 769
(00.003161)    769: timens: monotonic -3 172033336
(00.003195)    769: timens: boottime -3 171999592
(00.003281)    769: cg: Move into 2
(00.003299)    769: cg:   `-> unifie//test1/cgroup.procs
(00.003324)    769: Error (criu/cgroup.c:1092): cg: Can't move 769 into unifie//test1/cgroup.procs (-1/-1): Bad file descriptor
(00.003337)    769: Error (criu/cgroup.c:1206): cg: Can't move into unifie//test1/cgroup.procs (-1/-1): Bad file descriptor
(00.003390) Error (criu/cr-restore.c:2447): Restoring FAILED.
(00.003806) Error (criu/cr-restore.c:1480): 769 killed by signal 9: Killed
```
With my patch that works. After restore the process is back in the current cgroup:
```
# cat /proc/`pidof sleep`/cgroup
0::/user.slice/user-0.slice/session-1.scope
```
The ignore functionality is partially implemented currently but it is missing at two places. That is what my patch fixes.

avagin:
> No, I do not agree. According to the man page ignore should Don’t deal with cgroups and pretend that they don’t exist. 

You can be disagree and it is what it is. The source of truth is the commit message and the code but not the man page that was written much later than the code.

And I lean to accept this patch rather than introduce a new mode.

adrianreber:
> And I lean to accept this patch rather than introduce a new mode.

When did we talk about a new mode? This is just about fixing the existing `ignore` mode. Nothing else.

kolyshkin:
Ahh! That must explain why my earlier use of `--manage-cgroup-mode=ignore` did not fix the issues I was seeing with lazy migration (https://github.com/opencontainers/runc/issues/2924 and https://github.com/opencontainers/runc/issues/2760)!

Snorch:
I agree with @avagin, "ignore" should only mean that the container runtime is in charge of creating the whole container cgroup directory tree, but criu should still put processes it creates into same topology it had seen them on dump.

In Virtuozzo criu when we migrate container with change of container ID we do dump it with something like `--cgroup-root /machine.slice/$old_ct_ID` and restore with `--cgroup-root /machine.slice/$new_ct_ID`. This way if process was dumped in `/sys/fs/cgroup/memory/machine.slice/$old_ct_ID/sub/directory` it would restore in `/sys/fs/cgroup/memory/machine.slice/$new_ct_ID/sub/directory`.

Also note that in "soft" mode we don't restore cgroup options on root container cgroups if they were pre-created by container environment, so container environment is in charge to put the same values there as was on dump. Using "full" mode solves it but there is a fixes needed, see e.g.: https://src.openvz.org/projects/OVZ/repos/criu/commits/6cab5493c59d699fdd231fd1ad6d808fe3bc2278

adrianreber:
> I agree with @avagin, "ignore" should only mean that the container runtime is in charge of creating the whole container cgroup directory tree, but criu should still put processes it creates into same topology it had seen them on dump.

Independent of what the container runtime should do, `ignore` currently does not work. See my example in https://github.com/checkpoint-restore/criu/pull/1800#issuecomment-1088938831

Snorch:
> > I agree with @avagin, "ignore" should only mean that the container runtime is in charge of creating the whole container cgroup directory tree, but criu should still put processes it creates into same topology it had seen them on dump.
> 
> Independent of what the container runtime should do, `ignore` currently does not work. See my example in [#1800 (comment)](https://github.com/checkpoint-restore/criu/pull/1800#issuecomment-1088938831)

Yeh probably you are right, proving hunks from the original patch:
```
c7d646afb ("cgroups: Introduce cgroup management modes")
...
+#define CG_MODE_IGNORE         (0u << 0)       /* Zero is important here */
...
@@ -1122,10 +1130,14 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
        int off, i, ret;
        char paux[PATH_MAX];
 
-       pr_info("Preparing cgroups yard\n");
+       pr_info("Preparing cgroups yard (cgroups restore mode %#x)\n",
+               opts.manage_cgroups);
+       if (!opts.manage_cgroups)
+               return 0;
```
Cgroup yard was never initialized for "ignore", so it was never possible to put tasks in right cgroups with "ignore".

Please also apply this fix to you pr, so that "unified" does not convert to "unifie":

```
diff --git a/criu/cgroup.c b/criu/cgroup.c
index 82d9b16a2..7490cebf2 100644
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1035,7 +1035,7 @@ static int ctrl_dir_and_opt(CgControllerEntry *ctl, char *dir, int ds, char *opt
                }
 
                if (n[0] == 0)
-                       doff += snprintf(dir + doff, ds - doff, "unified");
+                       doff += snprintf(dir + doff, ds - doff, "unified,");
                else
                        doff += snprintf(dir + doff, ds - doff, "%s,", n);
                if (opt)
```

After this fix it is obvious that with "ignore" we always have -1 CGROUP_YARD service fd in userns_move.

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,158 @@
...
+
+	// Create the cgroup cgname (if it does not already exist)
+	asprintf(&destination, "%s/%s", dirname, cgname);
+	if (mkdir(destination, 0700) < 0 && errno != EEXIST) {
+		fail("Failed to create temporary cgroup directory %s", destination);
+		goto err;
+	}
+
+	// Move this process to the newly created cgroup
+	asprintf(&cgroup_procs, "%s/cgroup.procs", destination);
```
Snorch, line 102: Please handle asprintf errors (ENOMEM and others).

Snorch (approved):
Looks good.

avagin:
pls remove the passage about oci containers from the commit message. It is not correlate directly with the patch.

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,168 @@
...
+	cleanup_free char *cgroup_procs = NULL;
+	cleanup_free char *own_pid = NULL;
+	int length;
+	int ret = -1;
+
+	test_init(argc, argv);
+
+	buffer_old = malloc(PAGE_SIZE);
+	if (!buffer_old) {
+		fail("Could not allocate memory");
```
avagin, line 51: pr_err. It tests nothing here.

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,168 @@
...
+		fail("Could not open %s", cgroup_procs);
+		goto err;
+	}
+
+	length = asprintf(&own_pid, "%d", getpid());
+	if (length == -1) {
+		fail("Could not allocate memory");
+		goto err;
+	}
+	if (write(cgroup_procs_fd, own_pid, length) < 0) {
```
avagin, line 120: != length

avagin: and you can write 0 to move the current process into cgroup

rst0git:
> pls remove the passage about oci containers from the commit message. It is not correlate directly with the patch.

Actually, the commit message describes the problem we are trying to fix: https://github.com/containers/podman/issues/13672

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,160 @@
...
+	lseek(fd, 0, SEEK_SET);
+	if (read_all(fd, buffer_old, PAGE_SIZE)) {
+		pr_err("Could not read data from /proc/self/cgroup\n");
+		goto err;
+	}
+
+	// Test if /proc/self/cgroup has changed again
+	if (!memcmp(buffer_new, buffer_old, PAGE_SIZE)) {
+		fail("/proc/self/cgroup should differ after restore");
+		pr_err("original /proc/self/cgroup content %s\n", buffer_old);
```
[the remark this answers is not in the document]
adrianreber: changed

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,160 @@
...
+
+	test_daemon();
+	test_waitsig();
+
+	// Read /proc/self/cgroup. It should not be the same as after
+	// moving this process to another cgroup because of restore
+	// with '--manage-cgroups=ignore'. The process should be
+	// now in cgroup of the current session.
+	lseek(fd, 0, SEEK_SET);
+	if (read_all(fd, buffer_old, PAGE_SIZE)) {
```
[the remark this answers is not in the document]
avagin: yes, it is.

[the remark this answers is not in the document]
avagin: you need to exclude it for norst runs:

$ git grep norst test/ scripts/ | grep all
test/jenkins/crit.sh:./test/zdtm.py run --all -f best -x maps04 -x cgroup02 --norst --keep-img always || fail
test/jenkins/criu-dump.sh:./test/zdtm.py run --all --keep-going --report report --parallel 4 --norst -x 'maps04' -x 'cgroup02' || fail

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,161 @@
...
+	// Create the cgroup root directory
+	if (mkdir(dirname, 0700) < 0 && errno != EEXIST) {
+		pr_err("Cannot make directory %s\n", dirname);
+		return -1;
+	}
+
+	// Mount cgroup2, skip if cgroup2 is not available
+	if (mount("none", dirname, "cgroup2", 0, 0)) {
+		if (errno == ENODEV) {
+			skip("Test relies on cgroup2 semantics which this system does not support. Skipping");
```
avagin, line 81: nit: you can do this from .checkskip script. Look at test/zdtm/static/selinux00.checkskip, for example.

avagin:
BTW @Snorch explained me that what I think the ignore mode means is actually covered by the none mode.

test/zdtm/static/cgroup_ignore.c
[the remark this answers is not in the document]
adrianreber: done

test/zdtm/static/cgroup_ignore.c
adrianreber: added checkskip

test/zdtm/static/cgroup_ignore.c
```diff
@@ -0,0 +1,161 @@
...
+
+int main(int argc, char **argv)
+{
+	cleanup_free char *cgroup_procs = NULL;
+	cleanup_close int cgroup_procs_fd = -1;
+	cleanup_free char *destination = NULL;
+	cleanup_free char *buffer_old = NULL;
+	cleanup_free char *buffer_new = NULL;
+	cleanup_close int fd = -1;
+	int ret = -1;
```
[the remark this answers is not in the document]
adrianreber: change to +1

PR #1801, opened 2022-04-05
ci: Ubuntu broke overlayfs again

adrianreber:
Switch to non overlaysfs tests for Podman and Docker.

### DIFF
--- a/scripts/ci/docker-test.sh
+++ b/scripts/ci/docker-test.sh
@@ -21,7 +21,10 @@ add-apt-repository \
 
 . /etc/lsb-release
 
-echo '{ "experimental": true }' > /etc/docker/daemon.json
+# overlayfs behaves differently on Ubuntu (18.04) and breaks CRIU
+# https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
+# Switch to devicemapper
+echo '{ "experimental": true, "storage-driver": "devicemapper" }' > /etc/docker/daemon.json
 
 CRIU_LOG='/criu.log'
 mkdir -p /etc/criu
--- a/scripts/ci/podman-test.sh
+++ b/scripts/ci/podman-test.sh
@@ -25,7 +25,10 @@ make install
 popd
 rm -rf "${tmp_dir}"
 
-podman info
+# overlaysfs behaves differently on Ubuntu and breaks CRIU
+# https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
+export STORAGE_DRIVER=vfs
+podman --storage-driver vfs info
 
 # shellcheck disable=SC2016
 podman run --name cr -d docker.io/library/alpine /bin/sh -c 'i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done'


scripts/ci/podman-test.sh
```diff
@@ -25,7 +25,10 @@ make install
 popd
 rm -rf "${tmp_dir}"
 
-podman info
+# overlaysfs behaves differently on Ubuntu and breaks CRIU
+# https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
```
rst0git, line 29: Should we open a new bug? This one has "Fix Released" status.

rst0git (approved):
LGTM

scripts/ci/podman-test.sh
adrianreber: If you want you can do it. As it is reintroduced with every kernel upgrade I have no high hopes that this will be fixed correctly at some point.

rst0git: Sure, let's try one more time.

adrianreber: Sounds good

scripts/ci/docker-test.sh
```diff
@@ -21,7 +21,10 @@ add-apt-repository \
 
 . /etc/lsb-release
 
-echo '{ "experimental": true }' > /etc/docker/daemon.json
+# overlayfs behaves differently on Ubuntu (18.04) and breaks CRIU
+# https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
```
avagin, line 25: This is the old bug, isn't it? I think we need to file a new one.

avagin:
Sorry, I have to revert this change. First, it links the old ubuntu issue. Second, it is unclear what kernel is broken and what is broken. Feel free to create a new PR.

adrianreber:
> Sorry, I have to revert this change. First, it links the old ubuntu issue. Second, it is unclear what kernel is broken and what is broken. Feel free to create a new PR.

It is the same bug. They always reintroduce the broken patch.

adrianreber:
@avagin can you disable the broken tests until they are fixed. Not really helpful to have known broken CI tests running.

avagin:
@adrianreber How many times have you seen that I disabled broken tests due to bugs in the redhat kernel? It doesn't work this way, sorry. When you are working on CRIU, you are the CRIU engineer and your goal is that CRIU works on all major platforms. I don't ask you to fix bugs in Ubuntu kernels, but I think we have to do initial investigation and describe the issue in a bug for a proper project.

adrianreber:
> @adrianreber How many times have you seen that I disabled broken tests due to bugs in the redhat kernel? It doesn't work this way, sorry. When you are working on CRIU, you are the CRIU engineer and your goal is that CRIU works on all major platforms. I don't ask you to fix bugs in Ubuntu kernels, but I think we have to do initial investigation and describe the issue in a bug for a proper project.

Not sure what this has to do with the Red Hat kernel, but I have tested it on the mentioned Ubuntu kernel and it is the same error message. This is again #860. We know how to work around it. If you do not want the workaround, then disabling seems the better solution than having CI errors. This error exists since 2019 and it keeps coming back.

avagin:
@adrianreber I am only asking you to report the issue to Ubuntu engineers. In the comment, you link the bug that have been closed. If we don't reopen it, it will never be fixed.

avagin:
5.13.0-1017-azure has this issue. I filed a new ubuntu issue:

PR #1802, opened 2022-04-06
ci: Switch to non overlaysfs tests

rst0git:
Switch to non overlaysfs tests for Podman and Docker.


rst0git:
cc: @avagin @adrianreber

PR #1804, opened 2022-04-06
criu: fail migration if data was sent to an in-flight socket

mclapinski:
Before this change, CRIU would just lose that data upon migration. So
it's better to fail migration in this case.

To reproduce the bug one can:
1. Create an AF_UNIX socket and call listen on it.
2. Create a second AF_UNIX socket and call connect to the first one.
3. Send the data to the second socket.
4. Migrate.
5. Call accept on the first socket and then read. There would be no data
   available.

It should be even possible to close the second socket before migration.
This would cause accept to hang because CRIU totally misses a closed
in-flight socket.

### DIFF
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -497,9 +497,34 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
 			goto err;
 		}
 
+		if (sk->wqlen != 0) {
+			/*
+			 * There's no known way to get data out of the write
+			 * queue of an icon socket. The only good solution for
+			 * now is to fail the migration.
+			 */
+			pr_err("Non-empty write queue on an in-flight socket %#x\n", ue->ino);
+			goto err;
+		}
+
 		ue->peer = e->sk_desc->sd.ino;
 
 		pr_debug("\t\tFixed inflight socket %u peer %u)\n", ue->ino, ue->peer);
+	} else if (ue->state == TCP_LISTEN) {
+		int i;
+
+		for (i = 0; i < sk->nr_icons; i++)
+			if (sk->icons[i] == 0) {
+				/*
+				 * Inode of an icon socket equal to 0 means
+				 * it's already been closed. That means we have
+				 * no simple way to check if it sent any data.
+				 * The only good solution for now is to fail
+				 * the migration.
+				 */
+				pr_err("Found a closed in-flight socket to %#x\n", ue->ino);
+				goto err;
+			}
 	}
 dump:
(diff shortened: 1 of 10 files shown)


criu/sk-unix.c
```diff
@@ -497,9 +497,29 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
 			goto err;
 		}
 
+		if (sk->wqlen != 0) {
```
avagin, line 500: could you write comments for this and the next case that explain what exactly we don't support right now

criu/sk-unix.c
```diff
@@ -497,9 +497,29 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
...
+			       "%#x\n",
+			       ue->ino);
+			goto err;
+		}
+
 		ue->peer = e->sk_desc->sd.ino;
 
 		pr_debug("\t\tFixed inflight socket %u peer %u)\n", ue->ino, ue->peer);
+	} else if (ue->state == TCP_LISTEN) {
+		int i;
```
avagin, line 513: nit: pls add an empty line after declaring variables

mihalicyn:
LGTM

nit: maybe add TODO to commit description about writing `crfail` tests for that?

mihalicyn (approved):

avagin:
@mclapinski could you cherry-pick tests from https://github.com/avagin/criu/commits/mclapinski/unix_sock.

mclapinski:
@avagin Done. Thank you for writing them.

criu/sk-unix.c
```diff
@@ -497,9 +497,34 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
 			goto err;
 		}
 
+		if (sk->wqlen != 0) {
+			// There's no known way to get data out of the write
+			// queue of an icon socket. The only good solution for
+			// now is to fail the migration.
+			pr_err("Non-empty write queue on an in-flight socket "
+			       "%#x\n",
+			       ue->ino);
```
Snorch, line 506: Please don't split lines too much we have 120 char line length.
```
[ criu]$ grep ColumnLimit .clang-format 
ColumnLimit: 120
```

Snorch:
Fixed codding style in "criu: fail migration if data was sent to an in-flight socket" and rebased, should be ready to merge.

mclapinski:
@avagin Is there anything else I can do to get this merged?

avagin:
@mclapinski thank you for pinging me.

PR #1808, opened 2022-04-13
Shellcheck followup

kolyshkin:
1. Update shellcheck used in CI to the latest version (v0.7.2 -> v0.8.0).
2. Fix shellcheck warnings that were previously disabled (except for one).

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,18 @@ jobs:
       image: registry.fedoraproject.org/fedora:latest
     steps:
     - name: Install tools
-      run: sudo dnf -y install git make python3-flake8 ShellCheck clang-tools-extra which findutils codespell git-clang-format
+      run: sudo dnf -y install git make python3-flake8 xz clang-tools-extra which codespell git-clang-format
+
+    # TODO: remove this and use ShellCheck from repo once F37 with ShellCheck 0.8.0 is out.
+    - name: install shellcheck
+      env:
+        VERSION: v0.8.0
+        BASEURL: https://github.com/koalaman/shellcheck/releases/download
+        SHA256: f4bce23c11c3919c1b20bcb0f206f6b44c44e26f2bc95f8aa708716095fa0651
+      run: |
+        curl -sSfL --retry 5 $BASEURL/$VERSION/shellcheck-$VERSION.linux.x86_64.tar.xz |
+          tar xfJ - -C /usr/local/bin --strip 1 shellcheck-$VERSION/shellcheck
+        sha256sum --strict --check - <<<"$SHA256 /usr/local/bin/shellcheck"
 
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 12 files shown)


.github/workflows/lint.yml
```diff
@@ -8,9 +8,23 @@ jobs:
     container:
       image: registry.fedoraproject.org/fedora:latest
     steps:
+    - uses: actions/checkout@v2
     - name: Install tools
       run: sudo dnf -y install git make python3-flake8 ShellCheck clang-tools-extra which findutils
```
rst0git, line 13: perhaps we could remove `ShellCheck`?

adrianreber: I also thought about it. I see that Fedora rawhide has 0.8.0 already (https://src.fedoraproject.org/rpms/ShellCheck) but who knows if it will make its way to `fedora:latest`. I was using the Fedora container for `lint` to have up to date versions of the different tools so that I do not have to download it from "random" (not really random) places off the internet. But that is probably me just being old. I would prefer to use packages from the distribution, but if 0.8.0 does not appear really soon in `fedora:latest` than downloading it makes sense and could be removed here.

kolyshkin:
So, apparently adding an entry to `$GITHUB_PATH` breaks GHA in-container exec:
```
echo ~/bin >> $GITHUB_PATH
# ... Next step
make lint
...
OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "sh": executable file not found in $PATH: unknown
```

Rewrote without using $GITHUB_PATH

rst0git (approved):
LGTM

Thank you for working on this!

adrianreber:
Requires a rebase to fix the Arch Linux CI errors.

kolyshkin:
Rebased; no longer a stale.

.github/workflows/lint.yml
kolyshkin: We'll have to wait until F37. Added the TODO to switch back to distro packaged shellcheck to the yaml.

Snorch:
Please fix error "Unable to locate executable file: tar." of "Run code linter" check. Probably you just need to install tar there.

Snorch:
@kolyshkin Please fix error "Unable to locate executable file: tar." of "Run code linter" check. Probably you just need to install tar there.

adrianreber:
There is still the error @Snorch mentioned where `tar` cannot be found.

kolyshkin:
OK, this has nothing to do with tar. Apparenly, appending `$GITHUB_PATH` file to add more entries to `$PATH` (as recommended by GHA docs) is not working when the job is run in containers. Looking for an alternative solution.

kolyshkin:
Fedora Rawhide CI failure is obv unrelated

adrianreber (approved):
Looks good. CI errors are unrelated.

rst0git (approved):
LGTM

PR #1810, opened 2022-04-15
zdtm: flush ipt rules after program exits

time-river:
`bash ./test/jenkins/criu-join-ns.sh` will report errors:
```
  ################### 6 TEST(S) FAILED (TOTAL 425/SKIPPED 157) ###################
   * zdtm/static/net_lock_socket_iptables6(unknown)
   * zdtm/static/cgroup_yard(unknown)
   * zdtm/static/netns_lock_iptables(unknown)
   * zdtm/static/socket_udplite(h)
   * zdtm/static/net_lock_socket_iptables(unknown)
   * zdtm/static/socket_udp_shutdown(h)
```

It's caused by the case iptables rules created by
`zdtm/static/socket-tcp-nfconntrack`.


avagin:
zdtm/static/socket-tcp-nfconntrack adds new rules only in a separate network namespace:
https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm/static/socket-tcp.c#L71

time-river:
> zdtm/static/socket-tcp-nfconntrack adds new rules only in a separate network namespace: https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm/static/socket-tcp.c#L71

I pay attention to it, but it really has influence on the failure testcases. I run the command `ip netns exec zdtm_netns iptables -nvL`, and can see thoses rules. I looks like impossible...

avagin:
This is the join-ns test suite. In this case, zdtm creates /run/netns/zdtm_netns and restores all tests in this one network namespace. I am not sure that this is a good idea. I would prefer to create a per-test network namespace. Could you try out this patch:
index 0a52e1b96..d2ca25a19 100755
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1843,6 +1843,9 @@ def do_run_test(tname, tdesc, flavs, opts):
 
     fcg = get_freezer(opts['freezecg'])
 
+    if opts['join_ns']:
+        subprocess.Popen(["unshare", "-n", "sh", "-xc", "ip link set up dev lo && mount --bind /proc/self/ns/net %s" % join_ns_file]).wait()
+
     for f in flavs:
         print_sep("Run %s in %s" % (tname, f))
         if opts['dry_run']:

Issue #1811, opened 2022-04-17
help: the breakpoint problem in arm64 platform


time-river:
I try to implement the breakpoint in arm64 platform. [Here](https://github.com/time-river/criu/commit/be7a2438675d4ee209a029db56c841b91b959d35) is my code. But I met one problem which makes the feature abnormal:

In x86_64 platform, it seems that the Debug Exception event happens after the `rt_sigreturn` syscall has been executed, the change event that the content stored in the breakpoint address triggers the Debug Exception.

However, I tried to do the same procedure in arm64 platform, nothing happened, and I try to use execution breakpoint for `rt_sigreturn` syscall (code is [here](https://github.com/time-river/criu/commit/460633515ce8fa8dd7d2151460d82e75b882e1e9)), the program will triger segment fault after the restoration, the reason seems that the Debug Exception happens before the `svc` instruction fetching.

How should I implement this feature? @avagin

avagin:
I am not sure that I understand the question. You can look at the gdb code how it sets breakpoints.

time-river:
> I am not sure that I understand the question. You can look at the gdb code how it sets breakpoints.

I find the problem. The breakpoint event triggers when the instruction is fetch, and the instruction don't execute yet. Here is the phenomenon:

```
# crit show core-51754.img
                    "sp": "0xffffc21d8af0",
                    "pc": "0xffff8c88ea30",

# the good restoration (use `ptrace(SYSCALL)`)
(00.011463) 51754 (native) is going to execute the syscall 139, required is 139
(00.011464) >>> **sp: 0x28080 pc: ffff8c9a08e4**
(00.011474) 51754 was stopped
(00.011483) >>> **get regs ip ffff8c88ea30 stack ffffc21d8af0**
(00.011485) >>> parasite regs ip 11d20 stack (nil)
(00.011494) 51754 was trapped
(00.011496) 51754 (native) is going to execute the syscall 215, required is 215
(00.011498) >>> sp: 0xffffc21d8af0 pc: 15894

# the bad restoration (use hw breakpoint)
(00.054686) 51754 (native) is going to execute the syscall 139, required is 139
(00.054688) >>> **sp: 0x28080 pc: ffff8c9a08e0**
(00.054696) 51754 was stopped
(00.054703) >> unset hbrkt
(00.054713) >>> **get regs ip ffff8c9a08e0 stack 28080**
(00.054715) >>> parasite regs ip 11d20 stack (nil)
[...]

PR #1815, opened 2022-04-20
breakpoint: implement hw breakpoint for arm64 platform

time-river:
The same as commit [0b1b815](https://github.com/checkpoint-restore/criu/commit/0b1b815), I implement the hardware breakpoint for arm64 platform :-)

I notice the following commit, therefore the feature can't be enabled:
```c
        /*
         * Temporary workaround for Xen guests. Breakpoints degrade
         * performance linearly, so until we find out the reason,
         * let's disable them.
         */
        if (f == FI_NO_BREAKPOINTS)
                return true;
```
I don't know the reason here, and I don't have Xen environment. Besides the arm64 breakpoint, only x86 has this feature, therefore I only disable x86 by macro `CONFIG_X86_64`.

Note: the problem I met before is here #1811 

**NOTICE**:
DOING -- bug: Can't restore `SIGTRAP` signal (testcase `zdtm/static/sigtrap01` failed)

### DIFF
--- a/compel/arch/aarch64/src/lib/infect.c
+++ b/compel/arch/aarch64/src/lib/infect.c
@@ -2,6 +2,7 @@
 #include <sys/ptrace.h>
 #include <sys/types.h>
 #include <sys/uio.h>
+#include <sys/wait.h>
 #include <asm/ptrace.h>
 #include <linux/elf.h>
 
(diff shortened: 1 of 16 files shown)


avagin:
@time-river what is the status of this PR? Is it ready for review?

time-river:
> @time-river could you rebase this pr?

time-river:
> @time-river could you rebase this pr?

no problem.

time-river:
> @time-river what is the status of this PR? Is it ready for review?

YES, I run the test suit in arm64 platform, it works well.

Snorch:
```
commit 5211a57f4a73217c737b843389c9be80c50063aa
Author: fu.lin <fulin10@huawei.com>
Date:   Fri Apr 15 13:00:04 2022 +0800

    compel: switch breakpoint functions to non-inline at arm64 platform
    
    init
```
^ leftover comment in first patch commit message - please remove.

criu/pie/restorer.c
```diff
@@ -1418,7 +1418,7 @@ long __export_restore_task(struct task_restore_args *args)
 	n_helpers = args->helpers_n;
 	zombies = args->zombies;
 	n_zombies = args->zombies_n;
-	*args->breakpoint = rst_sigreturn;
+	*args->breakpoint = ARCH_SIGRETURN_ADDR(rst_sigreturn);
```
avagin, line 1421: ARCH_SIGRETURN_ADDR isn't required. Breakpoints are used to stop a process as close as possible to the target system call. Then we call PTRACE_SYSCALL to stop a process on the syscall-exit point.

compel/arch/aarch64/src/lib/infect.c
```diff
@@ -297,3 +298,43 @@ int ptrace_flush_breakpoints(pid_t pid, bool restore)
...
+	if (WIFSTOPPED(status))
+		pr_err("Task stopped with %d: %s\n", WSTOPSIG(status), strsignal(WSTOPSIG(status)));
+	if (WIFCONTINUED(status))
+		pr_err("Task continued\n");
+
+	return false;
+}
+
+/* wait single step debug trap event */
+int arch_wait_tasks_trap(int tasks)
```
avagin, line 321: arch_wait_tasks_trap has to be removed. You have it only because breakpoints don't work properly right now.

compel/arch/aarch64/src/lib/infect.c
```diff
@@ -177,12 +180,120 @@ unsigned long compel_task_size(void)
...
+	 * It's okay to return directly during checkpointing procedure, the
+	 * program will stop at the `svc` instruction fetching point.
+	 */
+	if (!restore)
+		return 0;
+
+	/*
+	 * The restorer stops at the breakpoint, but segment fault will trigger
+	 * if the function returns here. The reason is that the breakpoint event
+	 * happens before the instruction(`svc`) is fetched, the segment fault
```
avagin, line 287: this PTRACE_SINGLESTEP is here because compel_stop_on_syscall doesn't work as expected. It has to stop a process on the exit point from the target syscall. The problem is that we can't step over a breakpoint on arm64 and so when we call ptrace(PTRACE_SYSCALL), it stops on the breakpoint again and doesn't move forward. I think we need to flush a breakpoint as soon as we reach it.

PR #1860, opened 2022-04-27
cr-dump: do not report success to logs if post-dump script failed

Snorch:
It can be confusing to see error from post-dump action script and non
zero return from criu though at the same time see "Dumping finished
successfully" in log. I believe it is logical to consider post-dump
action script as a part of "dump" process so fail in it means that the
whole dump failed.


adrianreber (approved):
Makes sense.

PR #1862, opened 2022-05-01
criu: Version 3.17

adrianreber:
Amongst a huge number of fixes all over the place this release introduces:

* mount-v2 engine
* support for MAP_HUGETLB mappings
* support for Linux Restartable Sequences
* support for SOCK_SEQPACKET unix sockets
* CRIU AMD GPU plugin
* setsockopt(SO_BUF_LOCK) support for tcp sockets

@checkpoint-restore/maintainers this are the `Makefile.versions` changes for release 3.17. PTAL.

PR #1863, opened 2022-05-02
test: install criu in /usr

adrianreber:
GitHub Actions comes with pre-installed criu in /usr. configure scripts looking for CRIU will pickup the pre-installed version in /usr if we do not install CI criu also in /usr.

@rst0git: PTAL (without this the Podman test in https://github.com/checkpoint-restore/criu/pull/1862 fails once we increase the version to 3.17 because crun will pickup the wrong CRIU version)

### DIFF
--- a/scripts/ci/podman-test.sh
+++ b/scripts/ci/podman-test.sh
@@ -7,7 +7,7 @@ export SKIP_CI_TEST=1
 
 cd ../../
 
-make install
+make install PREFIX=/usr
 
 criu --version
 


scripts/ci/podman-test.sh
```diff
@@ -7,7 +7,7 @@ export SKIP_CI_TEST=1
 
 cd ../../
 
-make install
+make install PREFIX=/usr
```
avagin, line 10: Where does it install by default?

rst0git: `/usr/local/bin/`

rst0git, line 10: @adrianreber I think we need to add `LIBDIR` as well as to install libcriu correctly

adrianreber: I tried it in a container and on Ubuntu libcriu is installed in the correct location without explicitly setting LIBDIR.

rst0git: On Fedora 35, when I install criu with `make install PREFIX=/usr` libcriu is installed in:
```
ls -al /usr/lib/x86_64-linux-gnu/libcriu*
-rwxr-xr-x. 1 root root 292K May  2 10:08 /usr/lib/x86_64-linux-gnu/libcriu.a
lrwxrwxrwx. 1 root root   14 May  2 10:08 /usr/lib/x86_64-linux-gnu/libcriu.so -> libcriu.so.2.0
lrwxrwxrwx. 1 root root   14 May  2 10:08 /usr/lib/x86_64-linux-gnu/libcriu.so.2 -> libcriu.so.2.0
-rwxr-xr-x. 1 root root 194K May  2 10:08 /usr/lib/x86_64-linux-gnu/libcriu.so.2.0
```
However, the criu package for fedora installs these files in a different location:
```
rpm -ql criu-libs | grep libcriu
/usr/lib64/libcriu.so.2
/usr/lib64/libcriu.so.2.0
```
In this case crun would use `/usr/lib64/libcriu.so.2` instead of `/usr/lib/x86_64-linux-gnu/libcriu.so.2.0`.
Installing criu with the following command solved this problem:
```
make install PREFIX=/usr LIBDIR=/usr/lib64
```

adrianreber: Yes, but the podman test is running on Ubuntu and on Ubuntu the default location is correct.

rst0git: @adrianreber Yes, you are right

rst0git (approved):
LGTM

PR #1871, opened 2022-05-05
hugetlb: properly handle case when not supported

mihalicyn:
cc @minhbq-99

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -502,7 +502,7 @@ static bool kerndat_has_memfd_hugetlb(void)
 	if (ret >= 0) {
 		kdat.has_memfd_hugetlb = true;
 		close(ret);
-	} else if (ret == -1 && errno == EINVAL) {
+	} else if (ret == -1 && (errno == EINVAL || errno == ENOENT)) {
 		kdat.has_memfd_hugetlb = false;
 	} else {
 		pr_perror("Unexpected error from memfd_create(\"\", MFD_HUGETLB)");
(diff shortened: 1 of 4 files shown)


criu/kerndat.c
```diff
@@ -502,7 +502,7 @@ static bool kerndat_has_memfd_hugetlb(void)
 	if (ret >= 0) {
 		kdat.has_memfd_hugetlb = true;
 		close(ret);
-	} else if (ret == -1 && errno == EINVAL) {
+	} else if (ret == -1 && (errno == EINVAL || errno == ENOENT)) {
```
minhbq-99, line 505: Just curious in what case will we get ENOENT? I see that in case `MFD_HUGETLB` flag is not supported, we will get EINVAL right in the beginning of syscall handler function.

mihalicyn: I've put reference link in commit message:
https://github.com/torvalds/linux/blob/06fb4ecfeac/fs/hugetlbfs/inode.c#L1465

Issue is reproducible on real PPC64LE node with `5.12.15-200.fc33.ppc64le #1 SMP` kernel.

minhbq-99: I can't find the online link. Here is from a CentOS 7 source (linux-3.10.0-1160.el7) I downloaded in my local machine

```c
#define MFD_ALL_FLAGS (MFD_CLOEXEC | MFD_ALLOW_SEALING)

SYSCALL_DEFINE2(memfd_create,
		const char __user *, uname,
		unsigned int, flags)
{
	struct shmem_inode_info *info;
	struct file *file;
	int fd, error;
	char *name;
	long len;

	if (flags & ~(unsigned int)MFD_ALL_FLAGS)
		return -EINVAL;
```

minhbq-99: This snippet is from `mm/shmem.c`

PR #1874, opened 2022-05-06
cgroup: ignore property restore failure on EOPNOTSUPP

mihalicyn:
Let's allow EOPNOTSUPP to happen during cgroup properties
restore without failing entire restore. Why?
Yes, we can not to dump properties which can not be restored,
but it's better to handle this on restore stage to be compatible
with CRIU images created on older kernels.

Example:
58056f77502 ("memcg, kmem: further deprecate kmem.limit_in_bytes")

Reported-by: Mr. Jenkins


Snorch:
Why can't we just use `--cgroup-props-file`? With config file contents:
```
memory:
 - strategy: replace
 - properties: [ 
memory.limit_in_bytes, 
memory.memsw.limit_in_bytes, 
memory.swappiness, 
memory.soft_limit_in_bytes, 
memory.move_charge_at_immigrate, 
memory.oom_control, 
memory.use_hierarchy, 
]
```
This way we can skip dumping some cgroup options, so that we don't have to restore them later.

I don't like the idea of silently skipping dumped options restore fail.

Issue #1868, opened 2022-05-05
hugetlb: Freeze of maps09/maps10 tests on pre-dump


mihalicyn:
**Description**

pre-dump is broken with new maps09 and maps10 tests (hugetlb related).

**Steps to reproduce the issue:**
1. `python ./test/zdtm.py run -t zdtm/static/maps09 --pre 1 --ignore-taint -f h`

**Describe the results you received:**
Infinite loop on restore stage:
```
(00.005102) pie: 56: Remap 0x7f6a6b2b9000->0x7f514ad9d000 len 0x2000
(00.005109) pie: 56: Remap 0x7f6a6b2bb000->0x7f514ad9f000 len 0x2000
(00.005114) pie: 56: Remap 0x7f6a6b306000->0x7fff6674f000 len 0x2000
(00.005121) pie: 56: Remap 0x7f6a6b302000->0x7fff6674b000 len 0x4000
(00.005126) pie: 56: Remap 0x7f6a6b2bd000->0x7fff66622000 len 0x45000
(00.005132) pie: 56:    mmap(0x7f514a200000 -> 0x7f514a600000, 0x3 0x54040012 7)
(00.005140) pie: 56:    mmap(0x7f514a600000 -> 0x7f514aa00000, 0x3 0x54040011 8)
(00.005145) pie: 56:    mmap(0x7f514ad9c000 -> 0x7f514ad9d000, 0x3 0x11 9)
(00.005150) pie: 56: Preadv 0x7f514a200000:4194304... (1 iovs)
(00.005157) pie: 56: `- returned 0
(00.005162) pie: 56: Preadv 0x7f514a200000:4194304... (1 iovs)
(00.005166) pie: 56: `- returned 0
(00.005168) pie: 56: Preadv 0x7f514a200000:4194304... (1 iovs)
(00.005171) pie: 56: `- returned 0
(00.005173) pie: 56: Preadv 0x7f514a200000:4194304... (1 iovs)
(00.005176) pie: 56: `- returned 0
(00.005179) pie: 56: Preadv 0x7f514a200000:4194304... (1 iovs)
(00.005182) pie: 56: `- returned 0
(00.005184) pie: 56: Preadv 0x7f514a200000:4194304... (1 iovs)
(00.005187) pie: 56: `- returned 0
```

Kernels checked:
`5.4.0-1038-kvm #39-Ubuntu SMP`, `5.16.18-200.fc35.x86_64 #1 SMP PREEMPT`.

[...]

PR #1875, opened 2022-05-08
mem: Skip pre-dumping on hugetlb mappings

minhbq-99:
As private hugetlb mappings are not pre-mapped, the content of them is restored
in the the restorer which cannot use page_read->read_pages. As a result, we
cannot recursively read the content of pre-dumped image in the parent directory
and use preadv to read the content from the last dumped image only. Therefore,
it may freeze while restoring when the content of mapping is in pre-dumped image
in parent directory.

We need to skip pre-dumping on hugetlb mappings to resolve the issue.

Fixes: #1868

### DIFF
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -246,6 +246,12 @@ prep_dump_pages_args(struct parasite_ctl *ctl, struct vm_area_list *vma_area_lis
 		 */
 		if (vma_entry_is(vma->e, VMA_AREA_AIORING) && skip_non_trackable)
 			continue;
+		/*
+		 * We totally ignore MAP_HUGETLB on pre-dump.
+		 * See also generate_vma_iovs() comment.
+		 */
+		if ((vma->e->flags & MAP_HUGETLB) && skip_non_trackable)
+			continue;
 		if (vma->e->prot & PROT_READ)
 			continue;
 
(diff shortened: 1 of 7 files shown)


mihalicyn:
Great work, @minhbq-99 

I think it's worth adding this hunks too:
```
index 136439518..35859e5d6 100644
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -246,6 +246,12 @@ prep_dump_pages_args(struct parasite_ctl *ctl, struct vm_area_list *vma_area_lis
                 */
                if (vma_entry_is(vma->e, VMA_AREA_AIORING) && skip_non_trackable)
                        continue;
+               /*
+                * We totally ignore MAP_HUGETLB on pre-dump.
+                * See also generate_vma_iovs() comment.
+                */
+               if ((vma->e->flags & MAP_HUGETLB) && skip_non_trackable)
+                       continue;
                if (vma->e->prot & PROT_READ)
                        continue;
 
@@ -402,6 +408,13 @@ static int generate_vma_iovs(struct pstree_item *item, struct vma_area *vma, str
                        has_parent = false;
        }
 
+       /*
+        * We want to completely ignore these VMA types on the pre-dump:
+        * 1. VMA_AREA_AIORING because it non soft-dirty trackable (kernel writes)
+        * 2. MAP_HUGETLB mappings because they are not premapped, so we restore it
+        * content from the parasite context and can use only full memory image
+        * and can't use parent images from pre-dump stages.
+        */
        if (vma_entry_is(vma->e, VMA_AREA_AIORING)) {
                if (pre_dump)
                        return 0;
 ```

**Comment by minhbq-99** [comment]
I added those hunks, the CI failures look unrelated.

File: criu/mem.c [hunk 1]
```diff
@@ -402,7 +408,14 @@ static int generate_vma_iovs(struct pstree_item *item, struct vma_area *vma, str
 			has_parent = false;
 	}
 
-	if (vma_entry_is(vma->e, VMA_AREA_AIORING)) {
+	/*
+	 * We want to completely ignore these VMA types on the pre-dump:
+	 * 1. VMA_AREA_AIORING because it non soft-dirty trackable (kernel writes)
```
rst0git, line 413: *because it is not soft-dirty trackable

criu/mem.c
```diff
@@ -402,7 +408,14 @@ static int generate_vma_iovs(struct pstree_item *item, struct vma_area *vma, str
...
+	 * 2. MAP_HUGETLB mappings because they are not premapped, so we restore it
+	 * content from the parasite context and can use only full memory image
```
rst0git, line 415: *mappings because they are not pre-mapped and we can't use parent images from pre-dump stages. Instead, the content is restored from the parasite context using full memory image.

PR #1876, opened 2022-05-08
amdgpu: Set PLUGINDIR to /usr/lib/criu

rst0git:
This patch fixes the following error which appears when building criu packages for Ubuntu/Debian.

	mkdir: cannot create directory '/var/lib/criu': Permission denied

### DIFF
--- a/Makefile.install
+++ b/Makefile.install
@@ -7,7 +7,7 @@ MANDIR		?= $(PREFIX)/share/man
 INCLUDEDIR	?= $(PREFIX)/include
 LIBEXECDIR	?= $(PREFIX)/libexec
 RUNDIR		?= /run
-PLUGINDIR	?= /var/lib/criu
+PLUGINDIR	?= $(PREFIX)/lib/criu
 
 #
 # For recent Debian/Ubuntu with multiarch support.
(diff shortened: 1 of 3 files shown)


rst0git:
cc: @rajbhar

jpalus:
Would be good if this install target respected DESTDIR as well.

adrianreber:
I am not really happy with the plugin location in `/var`. I think it should be somewhere in `/usr`.

jpalus:
To add even more concerns regarding amdgpu plugin:
- `PLUGINDIR` from Makefile should be used as `#define CR_PLUGIN_DEFAULT` value via build flag `-DCR_PLUGIN_DEFAULT=\"$(PLUGINDIR)\"`
- not sure where `drm.h` header comes from in `#include <drm/drm.h>`
- `pkgconfig --cflags libdrm` flags are not used

rst0git:
@jpalus and @adrianreber Thank you for your comments. I've updated the pull request.

adrianreber:
This makes more sense to me than the previous location. :+1: 

Please rebase to pick up your CI linter run fix.

rst0git:
Thanks, I've rebased this pull request.

adrianreber (approved):

PR #1887, opened 2022-05-12
hugetlb: don't dump anonymous private hugetlb mapping using memfd approach

minhbq-99:
Currently, the content of anonymous private hugetlb mapping is dumped in 2
different images: memfd approach and normal private mapping dumping. In memfd
approach, we dump the content of the backing pseudo file (/anon_hugepage). This
is incorrect and redundant since the mapping is private, the content of backing
file may differ from the content of the mapping. With this commit, we remove the
redundant memfd approach dump and only do the normal private mapping dump on
anonymous hugetlb mapping.

Run zdtm.py run -f h --keep-img always -t zdtm/static/maps09, du -h in the
dumped image directory

Before this commit
	13M     test/dump/zdtm/static/maps09/55/1
After this commit
	8.5M    test/dump/zdtm/static/maps09/55/1

The reduction in size is approximately 4MB which is the size of anonymous
private hugetlb mapping in the test.

### DIFF
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -620,17 +620,16 @@ static int handle_vma(pid_t pid, struct vma_area *vma_area, const char *file_pat
 			pr_info("path: %s\n", file_path);
 			vma_area->e->status |= VMA_AREA_SYSVIPC;
 		} else {
-			/* Dump shmem dev, hugetlb dev (private and share) mappings the same way as memfd
-			 * when possible.
+			/* We dump memfd backed mapping, both normal and hugepage anonymous share
+			 * mapping using memfd approach when possible.
 			 */
 			if (is_memfd(st_buf->st_dev) || is_anon_shmem_map(st_buf->st_dev) ||
-			    (kdat.has_memfd_hugetlb && is_hugetlb_dev(st_buf->st_dev, &hugetlb_flag))) {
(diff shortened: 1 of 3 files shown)


criu/proc_parse.c
```diff
@@ -620,17 +620,18 @@ static int handle_vma(pid_t pid, struct vma_area *vma_area, const char *file_pat
...
-			/* Dump shmem dev, hugetlb dev (private and share) mappings the same way as memfd
-			 * when possible.
+			/* We dump memfd backed mapping, both normal and hugepage anonymous share
+			 * mapping using memfd approach when possible.
 			 */
 			if (is_memfd(st_buf->st_dev) || is_anon_shmem_map(st_buf->st_dev) ||
-			    (kdat.has_memfd_hugetlb && is_hugetlb_dev(st_buf->st_dev, &hugetlb_flag))) {
+			    (kdat.has_memfd_hugetlb && is_hugetlb_dev(st_buf->st_dev, &hugetlb_flag) &&
+			     (strncmp(file_path, ANON_HUGEPAGE_PREFIX, ANON_HUGEPAGE_PREFIX_LEN) ||
+			      vma_area->e->flags & MAP_SHARED))) {
```
avagin, line 629: Sorry for the delay. We need to add a comment to explain this expression or maybe even consider creating a separate helper for it.

minhbq-99: I split the expression into a new helper and add some comments to make it more readable.

Issue #1877, opened 2022-05-08
amdgpu plugin: Make possible to install in other location (during build)


carnil:
Hi

I'm currently trying to package 3.17 for Debian and noticed some problems with the newly introduced amdgpu plugin: The PLUGINDIR is almost fixed to `/var/lib/criu` which is not a problem, but during build the makefile makes it hard to have the plugins first installed below DESTDIR:

Something like

```
diff --git a/plugins/amdgpu/Makefile b/plugins/amdgpu/Makefile
index 84b9f871473d..42152c68e1cf 100644
--- a/plugins/amdgpu/Makefile
+++ b/plugins/amdgpu/Makefile
@@ -50,16 +50,16 @@ clean: amdgpu_plugin_clean amdgpu_plugin_test_clean
 mrproper: clean
 
 install:
-       $(Q) mkdir -p $(PLUGINDIR)
+       $(Q) mkdir -p $(DESTDIR)$(PLUGINDIR)
 ifeq ($(CONFIG_AMDGPU),y)
        $(E) "  INSTALL " $(PLUGIN_NAME)
-       $(Q) install -m 644 $(PLUGIN_SOBJ) $(PLUGINDIR)
+       $(Q) install -m 644 $(PLUGIN_SOBJ) $(DESTDIR)$(PLUGINDIR)
 endif
 .PHONY: install
 
 uninstall:
 ifeq ($(CONFIG_AMDGPU),y)
        $(E) " UNINSTALL" $(PLUGIN_NAME)
-       $(Q) $(RM) $(PLUGINDIR)/$(PLUGIN_SOBJ)
+       $(Q) $(RM) $(DESTDIR)$(PLUGINDIR)/$(PLUGIN_SOBJ)
 endif
 .PHONY: uninstall
```

I did not submit a full patch for it, as I'm not yet sure this is how you would like to handle it and I have not yet done much contributions, sorry about that :-(

A secondary problem is the build will fail afterwards ith the above changes applied, have not further investigated yet, yere the excerpt of the log:

```
gcc amdgpu_plugin_topology.c tests/test_topology_remap.c -o test_topology_remap -DCOMPILE_TESTS -iquote../../../criu/i
nclude -iquote../../criu/include -iquote../../criu/arch/x86/include/ -iquote../../ -I .
gcc -g -Wall -Werror -D _GNU_SOURCE -shared -nostartfiles -fPIC -I ../../compel/include/uapi amdgpu_plugin.c amdgpu_pl
[...]

PR #1890, opened 2022-05-15
amdgpu/Makefile: Fix include path

rst0git:
When building packages for CRIU the source directory might have a name different than `criu`.

Fixes: #1877

Reported-by: @siris


fxkamd:
This pull request looks good to me. Thanks!

rajbhar:
I was about the send a PR with similar change, this looks good to me, thanks @rst0git

PR #1905, opened 2022-05-26
seize: check process status before seizing it.

liusdu:
Ptrace() hangs forever when seizing stopped & frozen process.
So when using cgroup freezer to freeze process, we should check
whether process is stopped or not before seizing it.

references:
 Discuss in kernel community: https://lore.kernel.org/lkml/20151117193419.GA9993@redhat.com/t/

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -254,6 +254,12 @@ static int seize_cgroup_tree(char *root_path, enum freezer_state state)
 			return -1;
 		}
 
+		if (pid_is_stopped(pid)) {
+			pr_err("unexpected state of %d: stopped\n", pid);
+			fclose(f);
+			return -1;
+		}
+
 		if (!compel_interrupt_task(pid)) {
 			pr_debug("SEIZE %d: success\n", pid);
 			processes_to_wait++;
@@ -535,8 +541,10 @@ static int freeze_processes(void)
 	}
 
 err:
-	if (exit_code == 0 || origin_freezer_state == THAWED)
-		exit_code = freezer_write_state(fd, THAWED);
+	if (exit_code == 0 || origin_freezer_state == THAWED) {
+		if (freezer_write_state(fd, THAWED))
+			exit_code = -1;
+	}
 
 	if (close(fd)) {
 		pr_perror("Unable to thaw tasks");
--- a/criu/include/proc_parse.h
+++ b/criu/include/proc_parse.h
@@ -90,6 +90,7 @@ extern unsigned int parse_pid_loginuid(pid_t pid, int *err, bool ignore_noent);
 extern int parse_pid_oom_score_adj(pid_t pid, int *err);
 extern int prepare_loginuid(unsigned int value);
 extern int parse_pid_status(pid_t pid, struct seize_task_status *, void *data);
+extern int pid_is_stopped(pid_t pid);
 extern int parse_file_locks(void);
 extern int get_fd_mntid(int fd, int *mnt_id);
 
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -1012,6 +1012,44 @@ static int cap_parse(char *str, unsigned int *res)
 	return 0;
 }
 
+// check whether pid is stopped
+// return value:
+//   < 0 function executed failed
+//   = 0 pid is not stopped
+//   > 0 pid is stopped
+int pid_is_stopped(pid_t pid)
+{
+	struct bfd f;
+	char *str;
+	int ret = -1;
+
+	f.fd = open_proc(pid, "status");
+	if (f.fd < 0)
+		return -1;
+	if (bfdopenr(&f))
+		return -1;
+
+	while ((str = breadline(&f)) != NULL) {
+		if (IS_ERR(str)) {
+			pr_err("Error parsing proc status file\n");
+			goto out;
+		}
+
+		if (strncmp(str, "State:", 6)) {
+			continue;
+		}
+
+		if (!strstr(str, "T (stopped)"))
+			ret = 0;
+		else
+			ret = 1;
+		break;
+	}
+out:
+	bclose(&f);
+	return ret;
+}
+
 int parse_pid_status(pid_t pid, struct seize_task_status *ss, void *data)
 {
 	struct proc_status_creds *cr = container_of(ss, struct proc_status_creds, s);


criu/seize.c
```diff
@@ -535,8 +541,10 @@ static int freeze_processes(void)
 	}
 
 err:
-	if (exit_code == 0 || origin_freezer_state == THAWED)
-		exit_code = freezer_write_state(fd, THAWED);
+	if (exit_code == 0 || origin_freezer_state == THAWED) {
+		ret = freezer_write_state(fd, THAWED);
+		exit_code = exit_code ? exit_code : ret;
```
Snorch, line 546: I believe it's enough to return -1 on error and 0 on succes from freeze_processes(). This way code would probably look simpler with something like this:
```
if (freezer_write_state(fd, THAWED))
        exit_code = -1;
```

criu/seize.c
```diff
@@ -254,6 +254,12 @@ static int seize_cgroup_tree(char *root_path, enum freezer_state state)
 			return -1;
 		}
 
+		if (pid_is_stopped(pid)) {
+			pr_err("unexpected state of %d: stopped\n", pid);
+			fclose(f);
+			return -1;
```
Snorch, line 260: Note that the THAWED writing can have non-immediate effect and some processes may still be stopped. Please see how we can retry here https://src.openvz.org/projects/OVZ/repos/criu/browse/criu/seize.c#510 or just wait for it to actually change like this https://src.openvz.org/projects/OVZ/repos/criu/browse/criu/seize.c#310 .

liusdu: Hi @Snorch Thanks for your reply.
We care about process in state ‘T’ or not. I found freezing process did not change 'T'  State.  

 ```
[root@my tmp]# mkdir  -p /sys/fs/cgroup/freezer/test
[root@my tmp]# sleep 1000 & 
[2] 12828
[root@my tmp]# kill -s SIGSTOP 12828

[2]+  已停止               sleep 1000
[root@my tmp]# cat /proc/12828/status  |grep State
State:  T (stopped)
[root@my tmp]# echo 12828 > /sys/fs/cgroup/freezer/test/tasks 
[root@my tmp]# echo FROZEN > /sys/fs/cgroup/freezer/test/freezer.state 
[root@my tmp]# cat /proc/12828/status  |grep State
State:  T (stopped)
```

**Comment by liusdu** [comment]
> @uravas could you review this CL? Does it break #1813?

Hi @udzura  My test failed with your latest patch. 

Steps to reproduce.

1. create docker image as following:
```
# cat run.sh 
#!/bin/bash

sleep 1000 &
pid=$!
kill -s SIGSTOP $pid
wait $pid 
```
```
# cat Dockerfile 
FROM  ubuntu
ADD run.sh /
# build image.
# docker build  -t  test .
Sending build context to Docker daemon  7.168kB
Step 1/2 : FROM  ubuntu
 ---> 2b4cba85892a
Step 2/2 : ADD run.sh /
 ---> f549ca7287ae
Successfully built f549ca7287ae
Successfully tagged test:latest
```
2. run image
```
# docker run  -d test /run.sh
61f27d8100eb26592c3bad57c310907ad55620d302156f62278c641231892830
```
3. checkpoint container
```
# runc --root=/run/docker/runtime-runc/moby checkpoint --image-path=/opt/checkpoint 61f27d8100eb26592c3bad57c310907ad55620d302156f62278c641231892830
```
4. criu and dumpee in D state
```
# ps aux |grep criu
root     21172  0.0  0.0  49436  5728 pts/4    D    11:43   0:00 criu: dump --rpc -t 20942 -D /opt/checkpoint
# ps aux |grep 20942
root     20942  0.0  0.0  12776  2580 ?        Ds   11:42   0:00 /bin/bash /a.sh
```

**Comment by udzura** [comment]
Did you intend to mention to @uravas ?

**Comment by liusdu** [comment]
> Did you intend to mention to @uravas ?

Sorry for this mistake~

**Comment by uravas** [comment]
> @uravas could you review this CL? Does it break #1813?

As far as I can see, it doesn't break #1813. And it really improves the situation with stopped and frozen processes: CRIU report an error instead of a hang.

**Comment by Snorch** [comment]
After kernel fixes:
https://github.com/torvalds/linux/commit/7c3b00e06d731
https://github.com/torvalds/linux/commit/570ac9337b5c1
the original problem should be fixed and probably we don't need pid_is_stopped() hunks.

But I believe we still need the hunk in freeze_processes() as it fixes negative exit code override to 0.

@liusdu If it's ok for you, can you please update PR leaving freeze_processes() hunk.

**Comment by liusdu** [comment]
> After kernel fixes: [torvalds/linux@7c3b00e06d731](https://github.com/torvalds/linux/commit/7c3b00e06d731) [torvalds/linux@570ac9337b5c1](https://github.com/torvalds/linux/commit/570ac9337b5c1) the original problem should be fixed and probably we don't need pid_is_stopped() hunks.
> 
> But I believe we still need the hunk in freeze_processes() as it fixes negative exit code override to 0.
> 
> @liusdu If it's ok for you, can you please update PR leaving freeze_processes() hunk.

@Snorch ok~

**Comment by liusdu** [comment]
> After kernel fixes: [torvalds/linux@7c3b00e06d731](https://github.com/torvalds/linux/commit/7c3b00e06d731) [torvalds/linux@570ac9337b5c1](https://github.com/torvalds/linux/commit/570ac9337b5c1) the original problem should be fixed and probably we don't need pid_is_stopped() hunks.
> 
> But I believe we still need the hunk in freeze_processes() as it fixes negative exit code override to 0.
> 
> @liusdu If it's ok for you, can you please update PR leaving freeze_processes() hunk.

@Snorch  after a quick look at kernel source code I tested on, I found  patches you mentioned have been merged.
They do not solve this problem.  [torvalds/linux@7c3b00e06d731](https://github.com/torvalds/linux/commit/7c3b00e06d731)   
```
ptrace_attach() can hang waiting for STOPPED -> TRACED transition if the
tracee gets frozen in between, change wait_on_bit() to use TASK_KILLABLE.

This doesn't really solve the problem(s) and we probably need to fix the
freezer.  In particular, note that this means that pm freezer will fail if
it races attach-to-stopped-task.
```

**Comment by Snorch** [comment]
> freezer will fail if it races attach-to-stopped-task.

That seems completely ok for me from CRIU point of view, if freezer can't freeze processes for any reason, CRIU need to detect it and report error to user, which we already do here https://github.com/checkpoint-restore/criu/blob/c8f9880adab038481f7806173b698fc6e17ba76a/criu/seize.c#L510.

I understand that fixing freezer is a good thing too but it's out of scope of this PR.

**Comment by liusdu** [comment]
> > freezer will fail if it races attach-to-stopped-task.
> 
> That seems completely ok for me from CRIU point of view, if freezer can't freeze processes for any reason, CRIU need to detect it and report error to user, which we already do here
> 
> https://github.com/checkpoint-restore/criu/blob/c8f9880adab038481f7806173b698fc6e17ba76a/criu/seize.c#L510
> 
> .
> I understand that fixing freezer is a good thing too but it's out of scope of this PR.

ok..., fix in freezer is much better..

**Comment by avagin** [comment]
@liusdu we have a test that uses freezer to stop processes and it works fine, doesn't it? Could you describe the problem with more details?

**Comment by avagin** [comment]
zdtm tests fail with this patch:

```
$ python test/zdtm.py run -t zdtm/static/stopped --ignore-taint --freezecg zdtm:f
userns is supported
=== Run 1/1 ================ zdtm/static/stopped
======================== Run zdtm/static/stopped in uns ========================
 DEP       stopped.d
 CC        stopped.o
 LINK      stopped
Construct root for zdtm/static/stopped
Start test
./stopped --pidfile=stopped.pid --outfile=stopped.out
Run criu dump
=[log]=> dump/zdtm/static/stopped/83/1/dump.log
b'(00.000464) freezing processes: 100000 attempts with 100 ms steps'
b'(00.000478) cgroup.freeze=1'
b'(00.000529) SEIZE 83: success'
b'(00.000553) SEIZE 86: success'
b'(00.000570) Error (criu/seize.c:258): unexpected state of 87: stopped'
b'(00.000582) Unlock network'
b'(00.000593) Unfreezing tasks into 1'
b'(00.000595) \tUnseizing 83 into 1'
b'(00.000601) Error (criu/cr-dump.c:2053): Dumping FAILED.'
################## Test zdtm/static/stopped FAIL at CRIU dump ##################

Send the 9 signal to  83
Wait for zdtm/static/stopped(83) to die for 0.100000
$ git revert HEAD
[detached HEAD e401b0f8b] Revert "seize: check process status before seizing it."
 3 files changed, 2 insertions(+), 49 deletions(-)
$ make -j 4 > /dev/null
$ python test/zdtm.py run -t zdtm/static/stopped --ignore-taint --freezecg zdtm:f -f h
The kernel is tainted: '12352\n'
userns is supported
=== Run 1/1 ================ zdtm/static/stopped
========================= Run zdtm/static/stopped in h =========================
Start test
./stopped --pidfile=stopped.pid --outfile=stopped.out
Run criu dump
Run criu restore
Send the 15 signal to  52
Wait for zdtm/static/stopped(52) to die for 0.100000
Removing dump/zdtm/static/stopped/52
======================== Test zdtm/static/stopped PASS =========================
```

**Comment by liusdu** [comment]
> python test/zdtm.py run -t zdtm/static/stopped --ignore-taint --freezecg zdtm:f

Hi  @avagin This tests hangs forever in my environment.

```
[root@my criu]# python test/zdtm.py run -t zdtm/static/stopped --ignore-taint --freezecg zdtm:f                                                                                                                     
userns is supported                                                                                                                                                                                                 
=== Run 1/1 ================ zdtm/static/stopped                                                                                                                                                                    
timens isn't supported on 5.10.79                                                                                                                                                                        
========================= Run zdtm/static/stopped in h =========================                          
Start test                                                                                                                                                                                                          
./stopped --pidfile=stopped.pid --outfile=stopped.out                                                                                                                                                               
Run criu dump                                                                                                                                                                                                       
==== ALARM ====                                                                                           
    2 ?        S      0:00 [kthreadd]                                                                                                                                                                               
    3 ?        I<     0:00  \_ [rcu_gp]                                                                                                                                                                             
    4 ?        I<     0:00  \_ [rcu_par_gp]                                                                                                                                                                         
    6 ?        I<     0:00  \_ [kworker/0:0H-kb]                                                                                                                                                                    
    8 ?        I<     0:00  \_ [mm_percpu_wq]                                                                                                                                                                       
    9 ?        S      0:00  \_ [rcu_tasks_rude_]                                                                                                                                                                    
   10 ?        S      0:00  \_ [rcu_tasks_trace]                                                                                                                                                                    
   11 ?        S      0:00  \_ [ksoftirqd/0]
   12 ?        I      4:39  \_ [rcu_sched]
   13 ?        S      0:00  \_ [migration/0]
   15 ?        S      0:00  \_ [cpuhp/0]
   16 ?        S      0:00  \_ [cpuhp/1]
   17 ?        S      0:00  \_ [migration/1]
   18 ?        S      0:00  \_ [ksoftirqd/1]
   20 ?        I<     0:00  \_ [kworker/1:0H-ev]
   21 ?        S      0:00  \_ [cpuhp/2]
   22 ?        S      0:00  \_ [migration/2]
   23 ?        S      0:00  \_ [ksoftirqd/2]
   25 ?        I<     0:00  \_ [kworker/2:0H-kb]
```
```
# ps aux |grep criu
root      2585  0.0  0.0  49420  5680 pts/10   D+   12:00   0:00 ../criu/criu dump --no-default-config --log-file dump.log --images-dir dump/zdtm/static/stopped/56/1 --verbosity=4 --tree 56 --pidfile /data/home/l
iusdu/work/source/criu/test/zdtm/static/stopped.pid --freeze-cgroup /sys/fs/cgroup/freezer/zdtm --manage-cgroups --timeout 10 --pre-dump-mode splice
18943     2583  0.0  0.0   7508   120 ?        Ds   12:00   0:00 ./stopped --pidfile=stopped.pid --outfile=stopped.out

# cat /proc/2585/stack 
[<0>] ptrace_attach+0x1da/0x2a0
[<0>] __x64_sys_ptrace+0xa8/0x130
[<0>] do_syscall_64+0x60/0x110
[<0>] entry_SYSCALL_64_after_hwframe+0x44/0xa9
```

And this behavior seems has been discussed in https://lore.kernel.org/lkml/20151117193419.GA9993@redhat.com/t/

liusdu:
> After kernel fixes: [torvalds/linux@7c3b00e06d731](https://github.com/torvalds/linux/commit/7c3b00e06d731) [torvalds/linux@570ac9337b5c1](https://github.com/torvalds/linux/commit/570ac9337b5c1) the original problem should be fixed and probably we don't need pid_is_stopped() hunks.
> 
> But I believe we still need the hunk in freeze_processes() as it fixes negative exit code override to 0.
> 
> @liusdu If it's ok for you, can you please update PR leaving freeze_processes() hunk.

https://github.com/checkpoint-restore/criu/pull/1971

PR #1906, opened 2022-05-30
Fix the check for mnt namespace in criu-ns

ashu-mehra:
criu-ns script incorrectly compares the pidns fd with mntns fd.
Also reversed the condition in is_my_namespace function to align it
with the function name.

Signed-off-by: Ashutosh Mehra <asmehra@redhat.com>

### DIFF
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -153,9 +153,9 @@ def _set_namespace(fd):
         raise OSError(_errno, errno.errorcode[_errno])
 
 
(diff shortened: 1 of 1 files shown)


ashu-mehra:
I also reversed the condition in `is_my_namespace` to align it with the function name.

ashu-mehra:
Not sure why the linter is complaining for files that are not touched.

Snorch (approved):

PR #1907, opened 2022-05-31
RFC: github: use git-clang-format instead of make indent

Snorch:
This allows us to only detect bad formating in PR changes but not all
the CRIU codebase.

I checked it in my repo here https://github.com/Snorch/criu/runs/6670451552

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -9,7 +9,7 @@ jobs:
       image: registry.fedoraproject.org/fedora:latest
     steps:
     - name: Install tools
-      run: sudo dnf -y install git make python3-flake8 ShellCheck clang-tools-extra which findutils codespell
+      run: sudo dnf -y install git make python3-flake8 ShellCheck clang-tools-extra which findutils codespell git-clang-format
 
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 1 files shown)


Snorch:
It's not the first time where clang formatting changes between versions and we need to fix criu code to fix make indent. See 781676f10 clang-format/zdtm: fix clang complains about strange elseifs

This constant code change first can introduce bugs, second makes rebasing on new criu versions harder. Also we already saw places where clang-format is not good and we need our own formating:
```
git grep "clang-format off" | wc -l
35
```

It would be much easier for us, if we leave current code base untouched and only fix new patches and commits according to new clang-format.

@adrianreber Thanks for you opinion, hope to hear some more on it.

rst0git:
We can always run `make indent` locally.
For CI, it would make sense to detect formatting errors only in modified files.

mihalicyn:
IMHO, this change looks reasonable. As for me during PR checks it's better to check only code which was changed in particular PR.

adrianreber:
@Snorch seems like more people prefer your approach. So let's go with it for now. Please merge it.

Snorch:
OK, let's merge it to fix CI for now.

PR #1908, opened 2022-06-01
Add opt to ignore file r/w/x change on restore

### LINKED ISSUE #1891: Filesystem root dir permissions consistency check is problematic
**Description**

We're using CRIU to take checkpoints from within containers and we've run into the following issue.

CRIU saves the filesystem's root directory (found via the `/proc/<pid>/root` link) permissions in checkpoint images[^1]. At restore time it checks that the permissions previously saved in the image match the current root directory's permissions and aborts the restore otherwise.[^2] I cannot tell if the check is intentional or if it's simply a consequence of using the existing code for saving/restoring file references.

[^1]: https://github.com/checkpoint-restore/criu/blob/45641ab26d7bb78706a6215fdef8f9133abf8d10/criu/cr-dump.c#L333-L344
[^2]: https://github.com/checkpoint-restore/criu/blob/45641ab26d7bb78706a6215fdef8f9133abf8d10/criu/files-reg.c#L2191-L2213

In practice this is probably not a big deal for most users, however I've noticed that Ubuntu and RHEL have different permissions on the FS root dir (`rwxr-xr-x` and `r-xr-xr-x`, respectively). **Surprisingly this difference also exists in the containers started on each, even with the same image.** I wouldn't expect CRIU checkpoints to be portable across distros for many obvious reasons, but I would hope checkpoints in containers would not be affected by such a difference in the container hosts. I've not checked, but I suspect this means that container checkpoints are also not going to be portable, for the same reason.

**I'd like to suggest that this check be relaxed, made optional, or removed altogether for the FS root dir at least.** Comments and opinions welcome.

**Steps to reproduce the issue:**

***Generic steps***

1. Checkpoint a process.
2. Change the filesystem root directory permissions.
3. Try to restore the process.

***Container-specific steps***

1. Start a container on host X with FS root dir perms A
2. Take a checkpoint in the container
3. Commit the container to a new image
4. Copy the image to host Y with FS root dir perms B
5. Start another container
6. Try to restore the process

**Describe the results you received:**

Restore fails with error stating that file `.` (which refers to FS root dir) has bad mode.

**Describe the results you expected:**

A successfully restored process.

**Additional information you deem important (e.g. issue happens only occasionally):**

Happens in containers as well, as container FS root dir permissions seem to mirror host FS root dir.

**CRIU logs and information:**

```
...
   307: Error (criu/files-reg.c:2104): File . has bad mode 040555 (expect 040755)
   307: Error (criu/files.c:1348): Can't open root
Error (criu/cr-restore.c:1476): 307 exited, status=1
Error (criu/cr-restore.c:2447): Restoring FAILED.
```

ymanton:
This is a straight-forward attempt to deal with the problem described in #1891 by allowing users to override the described behaviour if they want. I'm happy to accept alternatives.

Commit message follows:

A file's r/w/x changing between checkpoint and restore does
not necessarily imply that something is wrong. For example,
if a process opens a file having perms rw- for reading and
we change the perms to r--, the process can be restored and
will function as expected.

Therefore, this patch adds an option

--ignore-file-rwx-changes-on-restore

to disable this check on restore. File validation is unaffected
and should still function as expected with respect to the content
of files.

Signed-off-by: Younes Manton <ymanton@ca.ibm.com>

Fixes #1891


Snorch:
1) Please change the subject line according to contribution guide, should be something like "config/files-reg: Add opt to ignore file r/w/x change on restore"

2) I believe best supported case for CRIU is a container migration, due to external resources problems and stuff. In this case CRIU assumes that filesystem remains untouched between dump and restore, so checking file-permissions on files is intentional to check that filesystem is untouched.

Probably in your case it's not so bad to skip this check as open_cb already opened the file with needed read/write flags. But do we really want to skip it for all files/mappings/inotify watches/scm fd in unixsk just because you have one non-matching file?

Another options are:
- fix mode in image for a desired file via crit encode/decode before restoring (probably simple)
- add an option similar to what we have with external files to skip dumping mode when file is matched by the option (probably hard)

ymanton:
Fixed whitespace for linter, added the option to the RPC API, added to man-page, and changed commit subject.

ymanton:
> I believe best supported case for CRIU is a container migration, due to external resources problems and stuff. In this case CRIU assumes that filesystem remains untouched between dump and restore, so checking file-permissions on files is intentional to check that filesystem is untouched.

That makes sense; I agree it's a reasonable default that will probably catch a lot of problems for end-users in the cases CRIU is most used for so I don't want to change that for a corner-case.

> Probably in your case it's not so bad to skip this check as open_cb already opened the file with needed read/write flags. But do we really want to skip it for all files/mappings/inotify watches/scm fd in unixsk just because you have one non-matching file?
> 
> Another options are:
> 
> * fix mode in image for a desired file via crit encode/decode before restoring (probably simple)
> * add an option similar to what we have with external files to skip dumping mode when file is matched by the option (probably hard)

Yeah, it's broad currently. Happy to limit it to a subset of file kinds. On the other hand, it seems like all of the potential corner cases associated with checkpointing/restoring a process are not known and probably not knowable since new things will always appear, so a non-default option like this might be useful as an escape hatch until an acceptable default behaviour is established.

Skipping the dumping of mode for specific files is also a workable idea; the only case we're (currently) concerned with is `/proc/<pid>/root`. I tried that, but my first attempt ended up being a bigger change than this so I didn't want to push that; rather I wanted to start with the smallest/simplest change and allow others a chance to provide alternatives.

We've considered fixing it during encode/decode and will probably do that if nothing else works out, but I thought it was worth bringing it here for the benefit of other users.

criu/crtools.c
```diff
@@ -504,6 +504,8 @@ int main(int argc, char *argv[], char *envp[])
 	       "  --file-validation METHOD\n"
 	       "			pass the validation method to be used; argument\n"
 	       "			can be 'filesize' or 'buildid' (default).\n"
+	       "  --ignore-file-rwx-changes-on-restore\n"
```
rst0git, line 507: It would be good if we can use shorter name for this option. For example:
```
--skip-file-mode-check
```

criu/files-reg.c
```diff
@@ -2199,9 +2199,20 @@ int open_path(struct file_desc *d, int (*open_cb)(int mntns_root, struct reg_fil
...
+			mode_t curr_mode = st.st_mode;
+			mode_t saved_mode = rfi->rfe->mode;
+
+			if (opts.ignore_rwx_change_on_restore) {
+				curr_mode &= ~(S_IRWXU | S_IRWXG | S_IRWXO);
+				saved_mode &= ~(S_IRWXU | S_IRWXG | S_IRWXO);
+			}
+
+			if (curr_mode != saved_mode) {
+				pr_err("File %s has bad mode 0%o (expect 0%o)\n", rfi->path, (int)curr_mode,
```
rst0git, line 2212: This error message could be updated to something like:

```
File %s has bad mode 0%o (expect 0%o)
File mode checks can be ignored with the --skip-file-mode-check option
```

rst0git:
> We've considered fixing it during encode/decode and will probably do that if nothing else works out, but I thought it was worth bringing it here for the benefit of other users.

Something similar was recently added in criu-image-streamer that allows to replace TCP listen socket ports:
https://github.com/checkpoint-restore/criu-image-streamer/commit/fd83e8521af57af6f171a16205516c6ded63db22

criu/crtools.c
ymanton: Thanks. I went with `--skip-file-rwx-check`  since the other mode bits are still going to be checked.

criu/files-reg.c
ymanton: Done. The failures without the option look as follows:

```
(00.003742) 344968: File root/criu-dev/criu/test/others/skip-file-rwx-check/testfile could only be validated with file size
(00.003747) 344968: Error (criu/files-reg.c:2212): File root/criu-dev/criu/test/others/skip-file-rwx-check/testfile has bad mode 0100444 (expect 0100644)
File r/w/x checks can be skipped with the --skip-file-rwx-check option
(00.003756) 344968: Error (criu/files.c:1213): Unable to open fd=3 id=0x8
(00.003944) Error (criu/cr-restore.c:2536): Restoring FAILED.
```

ymanton:
> @ymanton Would you be able to add ZDTM test for this option?

Done. It's a very simple shell script; tests both having and not having the option. A passing run looks as follows:

```
successful test run:

root@instant-on1:~/criu-dev/criu# make -C test/others/skip-file-rwx-check
make: Entering directory '/root/criu-dev/criu/test/others/skip-file-rwx-check'
./run.sh
+ source ../env.sh
++++ dirname ../env.sh
+++ readlink -f ../../../criu/criu
++ CRIU=/root/criu-dev/criu/criu/criu
++ criu=/root/criu-dev/criu/criu/criu
+++ which python3
++ '[' /usr/bin/python3 ']'
++ PYTHON=python3
++++ dirname ../env.sh
+++ readlink -f ../../../crit/crit-python3
++ CRIT=/root/criu-dev/criu/crit/crit-python3
++ crit=/root/criu-dev/criu/crit/crit-python3
++++ dirname ../env.sh
+++ readlink -f ../../../coredump/coredump-python3
++ CRIU_COREDUMP=/root/criu-dev/criu/coredump/coredump-python3
++ criu_coredump=/root/criu-dev/criu/coredump/coredump-python3
+ make clean
make[1]: Entering directory '/root/criu-dev/criu/test/others/skip-file-rwx-check'
rm -rf testfile *.img dump.log restore-expected-fail.log restore.log stats-dump stats-restore
make[1]: Leaving directory '/root/criu-dev/criu/test/others/skip-file-rwx-check'
+ touch testfile
+ chmod +w testfile
+ tailpid=344968
+ tail --follow testfile
+ /root/criu-dev/criu/criu/criu dump --tree=344968 --shell-job --verbosity=4 --log-file=dump.log
./run.sh: line 21: 344968 Killed                  tail --follow testfile
+ chmod -w testfile
+ /root/criu-dev/criu/criu/criu restore --restore-detached --shell-job --verbosity=4 --log-file=restore-expected-fail.log
+ /root/criu-dev/criu/criu/criu restore --skip-file-rwx-check --restore-detached --shell-job --verbosity=4 --log-file=restore.log
+ kill 344968
+ echo PASS
PASS
make: Leaving directory '/root/criu-dev/criu/test/others/skip-file-rwx-check'
```

test/Makefile
```diff
@@ -12,7 +12,7 @@ all:
 	$(MAKE) zdtm-freezer
 .PHONY: all
 
-TESTS = unix-callback mem-snap rpc libcriu mounts/ext security pipes crit socketpairs overlayfs mnt-ext-dev shell-job
+TESTS = unix-callback mem-snap rpc libcriu mounts/ext security pipes crit socketpairs overlayfs mnt-ext-dev shell-job skip-file-rwx-check
```
rst0git, line 15: It would be good to update [`run-ci-tests.sh`](https://github.com/checkpoint-restore/criu/blob/2b3763fe6a39033d4e3d3d80e28911d1737ac23b/scripts/ci/run-ci-tests.sh#L304) to run the test in CI as well.

ymanton: Missed that, thanks. Done.

images/rpc.proto
```diff
@@ -138,6 +138,7 @@ message criu_opts {
 	optional string			lsm_mount_context	= 63;
 	optional criu_network_lock_method	network_lock		= 64 [default = IPTABLES];
 	optional bool			mntns_compat_mode	= 65;
+	optional bool			skip_file_rwx_check = 66;
```
Snorch, line 141: Please fix alignment:
```
        optional bool                   mntns_compat_mode       = 65;
        optional bool                   skip_file_rwx_check = 66;
                                                            ^^^^
```

test/others/skip-file-rwx-check/run.sh
```diff
@@ -0,0 +1,32 @@
...
+fi
+chmod -w testfile
+if "$criu" restore --restore-detached --shell-job --verbosity=4 --log-file=restore-expected-fail.log
+then
+    kill $tailpid
+    echo "Unexpectedly restored process with reference to a file who's r/w/x perms changed when --skip-file-rwx-check option was not used"
+    echo FAIL
+    exit 1
+fi
+"$criu" restore --skip-file-rwx-check --restore-detached --shell-job --verbosity=4 --log-file=restore.log
```
Snorch, line 30: Printing some error message if the above fails would be helpful if ever we see this test do not pass.

Snorch (commented):
Code looks good except for small nits, please fix them.

Also please add commit subject prefix to your second commit, similar to what you've done with first, "test/others: " would be nice.

criu/crtools.c
```diff
@@ -504,6 +504,8 @@ int main(int argc, char *argv[], char *envp[])
 	       "  --file-validation METHOD\n"
 	       "			pass the validation method to be used; argument\n"
 	       "			can be 'filesize' or 'buildid' (default).\n"
+	       "  --skip-file-rwx-check\n"
+	       "            Skip checking file permissions (r/w/x for u/g/o) on restore.\n"
```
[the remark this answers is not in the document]
ymanton: Fixed and split over 2 lines to stay within the 80 char limit; now shows as:

```
  --file-validation METHOD
                        pass the validation method to be used; argument
                        can be 'filesize' or 'buildid' (default).
  --skip-file-rwx-check
                        Skip checking file permissions
                        (r/w/x for u/g/o) on restore.

Check options:
```

images/rpc.proto
ymanton: Fixed, thanks.

test/others/skip-file-rwx-check/run.sh
ymanton: Fixed; output is

```
Failed to restore process with reference to a file who's r/w/x perms changed when --skip-file-rwx-check option was used
FAIL
```

Snorch (approved):

PR #1918, opened 2022-06-17
util/mount-v2: fix resolve_mountpoint() to always return freeable pointer

Snorch:
Else we have a Segmentation fault in __move_mount_set_group() on
xfree(source_mp) if resolve_mountpoint() returned statically allocated
path.


adrianreber:
Looks good to me.

Issue #1893, opened 2022-05-17
Criu-ns restore fails with error: Failed to set group 1 on 0: Inappropriate ioctl for device


Niteesh26:
**Description**

criu-ns dump works properly as .img files are created in the CRIU folder.
criu-ns restore fails to restore the process with the error:
**Failed to set group 1 on 0: Inappropriate ioctl for device** 

**Steps to reproduce the issue:**
1. Run a simple counter program.
2. Dump the said program using criu-ns script: 

> ./criu-ns dump -D /data/CRIU/ -v3 --shell-job -t 189026

3. Try to restore the program using criu-ns script: 

> ./criu-ns restore -D /data/CRIU/ --shell-job -v3

**Describe the results you received:**
Failed to set group 1 on 0: Inappropriate ioctl for device

**Describe the results you expected:**
Restore completed successfully

**Additional information you deem important (e.g. issue happens only occasionally):**

**CRIU logs and information:**

<details><summary>CRIU full dump/restore logs:</summary>
<p>

```
(00.000415) Reading image tree
(00.000478) Add mnt ns 5 pid 189026
(00.000492) Add net ns 2 pid 189026
(00.000513) Migrating process tree (GID 189026->1 SID 188880->1)
(00.000527) Will restore in 0 namespaces
(00.000564) Collecting 37/54 (flags 2)
(00.000627) Collected [data/local/AMF/niteesh/criu-checkpoint/a.out] ID 0x1
(00.000645) Collected [usr/lib64/libc-2.17.so] ID 0x2
(00.000656) Collected [usr/lib64/libgcc_s-4.8.5-20150702.so.1] ID 0x3
(00.000664) Collected [usr/lib64/libm-2.17.so] ID 0x4
(00.000674) Collected [usr/lib64/libstdc++.so.6.0.19] ID 0x5
(00.000683) Collected [usr/lib64/ld-2.17.so] ID 0x6
(00.000696) Collected [dev/pts/6] ID 0x8
(00.000720) Collected [data/local/AMF/niteesh/criu-checkpoint] ID 0x9
(00.000749) Collected [.] ID 0xa
(00.000773) Collecting 43/59 (flags 0)
(00.000787) No remap-fpath.img image
(00.001041) cg: Preparing cgroups yard (cgroups restore mode 0x4)
(00.001380) cg: Determined cgroup dir blkio/user.slice already exist
(00.001397) cg: Skip restoring properties on cgroup dir blkio/user.slice
[...]

PR #1922, opened 2022-06-22
Criu ns tty fix

Snorch:
These are (probable) fixes for two errors when using criu-ns:

1. When restoring first time via criu-ns (from criu-ns dump):

```
Error (criu/tty.c:689): tty: Failed to set group 40816 on 0: Inappropriate ioctl for device
```

2. When dumping second time via criu-ns:

```
(00.005678) Error (criu/files-reg.c:1710): Can't lookup mount=29 for fd=0 path=/dev/pts/20
```

Fixes: #1893

### DIFF
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -4,6 +4,8 @@ import ctypes.util
 import errno
 import sys
 import os
+import fcntl
+import termios
 
 # <sched.h> constants for unshare
 CLONE_NEWNS = 0x00020000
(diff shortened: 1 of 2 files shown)


Snorch:
I'm not too sure about second patch, maybe we should skip inherited fds somewhere else, I'm open to any advise.

avagin:
lgtm

scripts/criu-ns
```diff
@@ -124,6 +126,7 @@ def wrap_restore():
     criu_pid = os.fork()
     if criu_pid == 0:
         os.setsid()
+        fcntl.ioctl(0, termios.TIOCSCTTY, 1)
```
avagin, line 129: pls write a comment that explains what we are doing here and why do we need to do this

Snorch: Done, also I updated the patch logic a bit to skip this step when not needed (not shell job) and fail on notatty.

scripts/criu-ns
```diff
@@ -124,6 +126,18 @@ def wrap_restore():
     criu_pid = os.fork()
     if criu_pid == 0:
         os.setsid()
+        '''
+        Set stdin tty to be a controlling tty of our new session, this is
+        required by --shell-job option, as for it CRIU would try to set a
+        process group of restored root task to be a foreground group on the
+        terminal.
+        '''
```
rst0git, line 134: The [convention](https://peps.python.org/pep-0257/) in Python is to always use `"""` (triple double quotes) for docstrings.
For [block comments](https://peps.python.org/pep-0008/#block-comments), each line of starts with a `#` and a single space.

Snorch: Thanks, will fix!

rst0git (approved):
LGTM

PR #1930, opened 2022-06-28
Allow CRIU to be used as non-root (Take 2)

ymanton:
We at the OpenJ9 and Open Liberty projects have been experimenting with CRIU to improve JVM and Java web application start-up times, primarily in container deployments.[^1] To that end we've been testing #1155 for a while now and having success and we want to help get it across the finish line.

---

With @adrianreber's agreement I've rebased the patches in #1155 and added 4 additional patches as follows:

Patches to add the `--unprivileged` option to `libcriu` and the RPC interface, and a fix to do the required `check_caps()` on the service worker path, equivalent to what's done on the tool path:

- 3807cc9cd non-root: Call check_caps() in service worker mode as well
- 8df76f09b non-root: Allow unprivileged flag to be set via libcriu

Patches to move some new code (that I believe requires root) recently landed in criu-dev to the root-only init path introduced by in #1155:

- b32f887ad non-root: Do kerndat_has_move_mount_set_group() in root-only init
- f42e25fe7 non-root: Do kerndat_has_nftables_concat() in root-only init

---

As suggested by @adrianreber I'm currently working on where to store `criu.kdat`, if/how to use `XDG_RUNTIME_PATH` for that purpose, and what to do when it's not set (e.g. when running via `sudo`).

I'm also looking over #1155 to see if there are any other unresolved questions.

In the meantime I wanted to open this PR now in case anyone has any fresh thoughts on the subject, comments on how to proceed, and so on.

[^1]: This use-case is a bit different from using CRIU to facilitate process migration and has some unique challenges, including the fact that we're using CRIU "under the covers" in a scenario that to end-users shouldn't appear too different from simply starting a process from scratch. Allowing non-root users to restore processes, particularly inside unprivileged containers, makes this use-case much more accessible to end-users because it reduces the privileges required to "start" a process in this way, relative to starting the process from scratch.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -734,6 +734,9 @@ int dump_task_cgroup(struct pstree_item *item, u32 *cg_id, struct parasite_dump_
 	unsigned int n_ctls = 0;
 	struct cg_set *cs;
 
+	if (opts.unprivileged)
+		return 0;
+
 	if (item)
 		pid = item->pid->real;
 	else
@@ -988,6 +991,9 @@ int dump_cgroups(void)
 {
 	CgroupEntry cg = CGROUP_ENTRY__INIT;
 	int ret = -1;
+
+	if (opts.unprivileged)
+		return 0;
 
 	BUG_ON(!criu_cgset || !root_cgset);
 
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -29,7 +29,7 @@
 #include "sockets.h"
 #include "crtools.h"
 #include "log.h"
-#include "util-pie.h"
+#include "util-caps.h"
 #include "prctl.h"
 #include "files.h"
 #include "sk-inet.h"
@@ -1654,4 +1654,48 @@ static char *feature_name(int (*func)(void))
 			return fl->name;
 	}
 	return NULL;
+}
+
+static int pr_set_dumpable(int value)
+{
+	int ret = prctl(PR_SET_DUMPABLE, value, 0, 0, 0);
+	if (ret < 0)
+		pr_perror("Unable to set PR_SET_DUMPABLE");
+	return ret;
+}
+
+int check_caps(void)
+{
+	struct proc_status_creds creds;
+	int exit_code = -1;
+
+	if (parse_pid_status(PROC_SELF, &creds.s, NULL))
+		goto out;
+
+	memcpy(&opts.cap_eff, &creds.cap_eff, sizeof(u32) * PROC_CAP_SIZE);
+
+	if (!has_cap_checkpoint_restore(opts.cap_eff))
+		goto out;
+
+	/* For some things we need to know if we are running as root. */
+	opts.uid = geteuid();
+
+	if (opts.uid) {
+		/*
+		 * At his point we know we are running as non-root with the necessary
+		 * capabilities available. Now we have to make the process dumpable
+		 * so that /proc/self is not owned by root.
+		 */
+		if (pr_set_dumpable(1))
+			return -1;
+	}
+
+	exit_code = 0;
+out:
+	if (exit_code) {
+		pr_msg("CRIU needs to have the CAP_SYS_ADMIN or the CAP_CHECKPOINT_RESTORE capability: \n");
+		pr_msg("setcap cap_checkpoint_restore+eip %s\n", opts.argv_0);
+	}
+
+	return exit_code;
 }
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -21,6 +21,7 @@
 #include <sys/prctl.h>
 #include <sched.h>
 #include <sys/mount.h>
+#include <sys/utsname.h>
 
 #include "../soccr/soccr.h"
 
@@ -514,6 +515,14 @@ err:
 static int check_ipc(void)
 {
 	int ret;
+
+	/*
+	 * Since kernel 5.16 sem_next_id can be accessed via CAP_CHECKPOINT_RESTORE, however
+	 * for non-root users access() runs with an empty set of caps and will therefore always
+	 * fail.
+	 */
+	if (opts.uid)
+		return 0;
 
 	ret = access("/proc/sys/kernel/sem_next_id", R_OK | W_OK);
 	if (!ret)
@@ -1039,10 +1048,14 @@ static int check_tcp(void)
 	}
 
 	val = 1;
-	ret = setsockopt(sk, SOL_TCP, TCP_REPAIR, &val, sizeof(val));
-	if (ret < 0) {
-		pr_perror("Can't turn TCP repair mode ON");
-		goto out;
+	if (!opts.unprivileged || has_cap_net_admin(opts.cap_eff)) {
+		ret = setsockopt(sk, SOL_TCP, TCP_REPAIR, &val, sizeof(val));
+		if (ret < 0) {
+			pr_perror("Can't turn TCP repair mode ON");
+			goto out;
+		}
+	} else {
+		pr_info("Not checking for TCP repair mode. Please set CAP_NET_ADMIN\n");
 	}
 
 	optlen = sizeof(val);
@@ -1394,9 +1407,6 @@ int cr_check(void)
 	struct ns_id *ns;
 	int ret = 0;
 
-	if (!is_root_user())
-		return -1;
-
 	root_item = alloc_pstree_item();
 	if (root_item == NULL)
 		return -1;
@@ -1666,36 +1676,43 @@ static int pr_set_dumpable(int value)
 
 int check_caps(void)
 {
-	struct proc_status_creds creds;
-	int exit_code = -1;
-
-	if (parse_pid_status(PROC_SELF, &creds.s, NULL))
+	/* Read out effective capabilities and store in opts.cap_eff. */
+	if (set_opts_cap_eff())
 		goto out;
 
-	memcpy(&opts.cap_eff, &creds.cap_eff, sizeof(u32) * PROC_CAP_SIZE);
-
+	/*
+	 * No matter if running as root or not. CRIU always needs
+	 * at least these capabilities.
+	 */
 	if (!has_cap_checkpoint_restore(opts.cap_eff))
 		goto out;
 
 	/* For some things we need to know if we are running as root. */
 	opts.uid = geteuid();
 
-	if (opts.uid) {
-		/*
-		 * At his point we know we are running as non-root with the necessary
-		 * capabilities available. Now we have to make the process dumpable
-		 * so that /proc/self is not owned by root.
-		 */
-		if (pr_set_dumpable(1))
-			return -1;
+	if (!opts.uid) {
+		/* CRIU is running as root. No further checks are necessary. */
+		return 0;
 	}
 
-	exit_code = 0;
+	if (!opts.unprivileged) {
+		pr_msg("Running as non-root requires '--unprivileged'\n");
+		pr_msg("Please consult the documentation for limitations when running as non-root\n");
+		return -1;
+	}
+
+	/*
+	 * At his point we know we are running as non-root with the necessary
+	 * capabilities available. Now we have to make the process dumpable
+	 * so that /proc/self is not owned by root.
+	 */
+	if (pr_set_dumpable(1))
+		return -1;
+
+	return 0;
 out:
-	if (exit_code) {
-		pr_msg("CRIU needs to have the CAP_SYS_ADMIN or the CAP_CHECKPOINT_RESTORE capability: \n");
-		pr_msg("setcap cap_checkpoint_restore+eip %s\n", opts.argv_0);
-	}
+	pr_msg("CRIU needs to have the CAP_SYS_ADMIN or the CAP_CHECKPOINT_RESTORE capability: \n");
+	pr_msg("setcap cap_checkpoint_restore+eip %s\n", opts.argv_0);
 
-	return exit_code;
+	return -1;
 }
--- a/criu/fdstore.c
+++ b/criu/fdstore.c
@@ -13,6 +13,8 @@
 #include "rst-malloc.h"
 #include "log.h"
 #include "util.h"
+#include "cr_options.h"
+#include "util-caps.h"
 
 /* clang-format off */
 static struct fdstore_desc {
@@ -27,6 +29,8 @@ int fdstore_init(void)
 	uint32_t buf[2] = { INT_MAX / 2, INT_MAX / 2 };
 	struct sockaddr_un addr;
 	unsigned int addrlen;
+	int rcv_opt_name;
+	int snd_opt_name;
 	struct stat st;
 	int sk, ret;
 
@@ -49,8 +53,16 @@ int fdstore_init(void)
 		return -1;
 	}
 
-	if (setsockopt(sk, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0], sizeof(buf[0])) < 0 ||
-	    setsockopt(sk, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1], sizeof(buf[1])) < 0) {
+	if (!opts.unprivileged || has_cap_net_admin(opts.cap_eff)) {
+		rcv_opt_name = SO_RCVBUFFORCE;
+		snd_opt_name = SO_SNDBUFFORCE;
+	} else {
+		rcv_opt_name = SO_RCVBUF;
+		snd_opt_name = SO_SNDBUF;
+	}
+
+	if (setsockopt(sk, SOL_SOCKET, snd_opt_name, &buf[0], sizeof(buf[0])) < 0 ||
+	    setsockopt(sk, SOL_SOCKET, rcv_opt_name, &buf[1], sizeof(buf[1])) < 0) {
 		pr_perror("Unable to set SO_SNDBUFFORCE/SO_RCVBUFFORCE");
 		close(sk);
 		return -1;
--- a/criu/files.c
+++ b/criu/files.c
@@ -21,7 +21,7 @@
 #include "image.h"
 #include "common/list.h"
 #include "rst-malloc.h"
-#include "util-pie.h"
+#include "util-caps.h"
 #include "common/lock.h"
 #include "sockets.h"
 #include "pstree.h"
@@ -1346,10 +1346,35 @@ static int fchroot(int fd)
 	return chroot(".");
 }
 
+static int need_chroot(int saved_root)
+{
+	struct stat saved_root_stat, cur_root_stat;
+	int psd;
+
+	if (fstat(saved_root, &saved_root_stat) == -1) {
+		pr_perror("Failed to stat saved root dir");
+		return -1;
+	}
+
+	psd = open_pid_proc(PROC_SELF);
+	if (psd < 0) {
+		pr_perror("Failed to open PROC_SELF");
+		return -1;
+	}
+
+	if (fstatat(psd, "root", &cur_root_stat, 0) == -1) {
+		pr_perror("Failed to stat current root dir");
+		return -1;
+	}
+
+	return saved_root_stat.st_ino != cur_root_stat.st_ino || saved_root_stat.st_dev != cur_root_stat.st_dev;
+}
+
 int restore_fs(struct pstree_item *me)
 {
 	int dd_root = -1, dd_cwd = -1, ret, err = -1;
 	struct rst_info *ri = rsti(me);
+	bool do_chroot = true;
 
 	/*
 	 * First -- open both descriptors. We will not
@@ -1369,14 +1394,23 @@ int restore_fs(struct pstree_item *me)
 	}
 
 	/*
+	 * In unprivileged mode chroot() may fail if we don't have
+	 * sufficient privileges, therefore only do it if the process
+	 * is actually chrooted.
+	 */
+	if (opts.unprivileged)
+		do_chroot = need_chroot(dd_root);
+
+	/*
 	 * Now do chroot/chdir. Chroot goes first as it calls chdir into
 	 * dd_root so we'd need to fix chdir after it anyway.
 	 */
-
-	ret = fchroot(dd_root);
-	if (ret < 0) {
-		pr_perror("Can't change root");
-		goto out;
+	if (do_chroot) {
+		ret = fchroot(dd_root);
+		if (ret < 0) {
+			pr_perror("Can't change root");
+			goto out;
+		}
 	}
 
 	ret = fchdir(dd_cwd);
--- a/criu/image.c
+++ b/criu/image.c
@@ -226,7 +226,8 @@ int prepare_inventory(InventoryEntry *he)
 	if (get_task_ids(&crt.i))
 		return -1;
 
-	he->has_root_cg_set = true;
+	if (!opts.unprivileged)
+		he->has_root_cg_set = true;
 	if (dump_task_cgroup(NULL, &he->root_cg_set, NULL))
 		return -1;
 
--- a/criu/include/cr_options.h
+++ b/criu/include/cr_options.h
@@ -1,10 +1,11 @@
 #ifndef __CR_OPTIONS_H__
 #define __CR_OPTIONS_H__
 
-#include <sys/types.h>
 #include <stdbool.h>
 #include "common/config.h"
 #include "common/list.h"
+#include "int.h"
+#include "image.h"
 
 /* Configuration and CLI parsing order defines */
 #define PARSING_GLOBAL_CONF  1
@@ -210,6 +211,20 @@ struct cr_options {
 	enum criu_mode mode;
 
 	int mntns_compat_mode;
+
+	/* Remember the program name passed to main() so we can use it in
+	 * error messages elsewhere.
+	 */
+	char *argv_0;
+	/*
+	 * This contains the eUID of the current CRIU user. It
+	 * will only be set to a non-zero value if CRIU has
+	 * the necessary capabilities to run as non root.
+	 * CAP_CHECKPOINT_RESTORE or CAP_SYS_ADMIN
+	 */
+	uid_t uid;
+	/* This contains the value from /proc/pid/status: CapEff */
+	u32 cap_eff[CR_CAP_SIZE];
 };
 
 extern struct cr_options opts;
--- a/criu/include/cr_options.h
+++ b/criu/include/cr_options.h
@@ -2,6 +2,7 @@
 #define __CR_OPTIONS_H__
 
 #include <stdbool.h>
+#include <sys/capability.h>
 #include "common/config.h"
 #include "common/list.h"
 #include "int.h"
@@ -223,8 +224,14 @@ struct cr_options {
 	 * CAP_CHECKPOINT_RESTORE or CAP_SYS_ADMIN
 	 */
 	uid_t uid;
-	/* This contains the value from /proc/pid/status: CapEff */
-	u32 cap_eff[CR_CAP_SIZE];
+	/* This contains the value from capget()->effective */
+	u32 cap_eff[_LINUX_CAPABILITY_U32S_3];
+	/*
+	 * If CRIU should be running as non-root with the help of
+	 * CAP_CHECKPOINT_RESTORE or CAP_SYS_ADMIN the user should
+	 * explicitly request it as it comes with many limitations.
+	 */
+	int unprivileged;
 };
 
 extern struct cr_options opts;
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -21,6 +21,7 @@
 #if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
 #include <nftables/libnftables.h>
 #endif
+#include <sys/utsname.h>
 
 #include "common/config.h"
 #include "int.h"
@@ -51,6 +52,7 @@
 #include "sched.h"
 #include "memfd.h"
 #include "mount-v2.h"
+#include "util-caps.h"
 
 struct kerndat_s kdat = {};
 
@@ -1075,19 +1077,66 @@ static int kerndat_has_openat2(void)
 	return 0;
 }
 
-#define KERNDAT_CACHE_FILE     KDAT_RUNDIR "/criu.kdat"
-#define KERNDAT_CACHE_FILE_TMP KDAT_RUNDIR "/.criu.kdat"
+#define KERNDAT_CACHE_NAME "criu.kdat"
+#define KERNDAT_CACHE_FILE KDAT_RUNDIR "/" KERNDAT_CACHE_NAME
 
+/*
+ * Returns:
+ * -1 if kdat_file was not written due to error
+ * 0 if kdat_file was written
+ * 1 if kdat_file was not written because cache directory undefined in env (non-root mode)
+ */
+static int get_kerndat_filename(char **kdat_file)
+{
+	int ret;
+
+	/*
+	 * Running as non-root, even with CAP_CHECKPOINT_RESTORE, does not
+	 * allow to write to KDAT_RUNDIR which usually is only writable by root.
+	 * Let's write criu.kdat file to XDG_RUNTIME_DIR for non-root cases.
+	 * Note that XDG_RUNTIME_DIR is not always defined (e.g. when executing
+	 * via su/sudo).
+	 */
+	if (opts.unprivileged) {
+		const char *cache_dir = getenv("XDG_RUNTIME_DIR");
+		if (!cache_dir) {
+			pr_warn("$XDG_RUNTIME_DIR not set. Cannot find location for kerndat file\n");
+			return 1;
+		}
+		ret = asprintf(kdat_file, "%s/%s", cache_dir, KERNDAT_CACHE_NAME);
+	} else {
+		ret = asprintf(kdat_file, "%s", KERNDAT_CACHE_FILE);
+	}
+
+	if (unlikely(ret < 0)) {
+		pr_warn("Cannot allocate memory for kerndat file name\n");
+		return -1;
+	}
+
+	return 0;
+}
+
+/*
+ * Returns:
+ * -1 if error
+ * 0 if cache was loaded
+ * 1 if cache does not exist or is stale or cache directory undefined in env (non-root mode)
+ */
 static int kerndat_try_load_cache(void)
 {
+	cleanup_free char *kdat_file = NULL;
 	int fd, ret;
 
-	fd = open(KERNDAT_CACHE_FILE, O_RDONLY);
+	ret = get_kerndat_filename(&kdat_file);
+	if (ret)
+		return ret;
+
+	fd = open(kdat_file, O_RDONLY);
 	if (fd < 0) {
 		if (ENOENT == errno)
-			pr_debug("File %s does not exist\n", KERNDAT_CACHE_FILE);
+			pr_debug("File %s does not exist\n", kdat_file);
 		else
-			pr_warn("Can't load %s\n", KERNDAT_CACHE_FILE);
+			pr_warn("Can't load %s\n", kdat_file);
 		return 1;
 	}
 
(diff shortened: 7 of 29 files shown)


criu/include/kerndat.h
```diff
@@ -82,6 +85,15 @@ struct kerndat_s {
...
 	bool has_ptrace_get_rseq_conf;
+	/*
+	 * For non-root CRIU we use uname(2) to remember the nodename and
+	 * release. If the kerndat file has a different nodename or release
+	 * we assume that there has been a reboot and we need to recreate it.
+	 * nodename in addition to release in case the kerndat file is on a
+	 * NFS shared file system.
+	 */
+	char release[RELEASE_SIZE];
+	char nodename[NODENAME_SIZE];
```
adrianreber, line 96: This can probably be removed for the new, XDG_RUNTIME_DIR based kdat approach.

ymanton: Done. In non-root mode if `XDG_RUNTIME_DIR` is defined we'll use it to form the cache filename and try to load/store to that file; if undefined we'll init the kdat from scratch and discard on exit.

scripts/ci/vagrant.sh
```diff
@@ -68,4 +68,17 @@ fedora-rawhide() {
 	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; sudo -E make -C scripts/ci fedora-rawhide CONTAINER_RUNTIME=podman BUILD_OPTIONS="--security-opt seccomp=unconfined"'
 }
 
+fedora-non-root() {
+	# Need a reboot to activate the latest Fedora kernel with CAP_CHECKPOINT_RESTORE
```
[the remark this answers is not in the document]
ymanton: Removed, thanks.

criu/cr-check.c
```diff
@@ -516,6 +516,13 @@ static int check_ipc(void)
 {
 	int ret;
 
+	/*
+	 * There seems to be no capability for accessing sem_next_id.
+	 * Let's check if root or not.
+	 */
+	if (opts.uid)
```
avagin, line 523: sem_next_id is protected by CAP_CHECKPOINT_RESTORE, isn't it?

criu/cr-check.c
```diff
@@ -1665,36 +1673,43 @@ static int pr_set_dumpable(int value)
...
+	 * No matter if running as root or not. CRIU always needs
+	 * at least these capabilities.
+	 */
 	if (!has_cap_checkpoint_restore(opts.cap_eff))
 		goto out;
 
 	/* For some things we need to know if we are running as root. */
 	opts.uid = geteuid();
 
-	if (opts.uid) {
```
avagin: I think we have to check opts.unprivileged instead of opts.uid in all places.  The user can run criu under non-root by mistake and opts.unprivileged guarantees that it is intentional.

criu/fdstore.c
```diff
@@ -49,8 +53,16 @@ int fdstore_init(void)
 		return -1;
 	}
 
-	if (setsockopt(sk, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0], sizeof(buf[0])) < 0 ||
```
avagin, line 52: we can try to run SO_SNDBUFFORCE and than fall back to SO_RCVBUF if opts.unprivileged is set.

criu/kerndat.c
```diff
@@ -1478,6 +1565,37 @@ int kerndat_try_load_new(void)
...
 }
 
+static int root_only_init(void)
+{
+	int ret = 0;
+
+	if (opts.uid)
+		return 0;
+
+	if (!ret && kerndat_loginuid()) {
```
avagin, line 1570: all kerndat changes have to be moved into a separate patch.

criu/include/cr_options.h
```diff
@@ -209,6 +210,17 @@ struct cr_options {
 	enum criu_mode mode;
 
 	int mntns_compat_mode;
+
+	char *argv_0;
```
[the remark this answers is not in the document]
ymanton: It's saved so we can show error messages containing the program name. Added a comment to that effect here.

criu/cgroup.c
```diff
@@ -988,6 +991,9 @@ int dump_cgroups(void)
 	CgroupEntry cg = CGROUP_ENTRY__INIT;
 	int ret = -1;
 
+	if (opts.uid)
```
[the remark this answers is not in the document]
ymanton: Done, thanks.

criu/cr-check.c
ymanton: Done, thanks.

I've left one in place in `check_caps()` so we can check that `opts.unprivileged` is being used in non-root mode.

criu/cr-check.c
ymanton: Yes, since kernel 5.16 (rc1) that is the case. Since this bit is part of `criu check` should we remove it explicitly or only for kernels 5.16 and later?

ymanton:
- Went back and fixed the kdat file XDG_RUNTIME_DIR patch to allow the tmpfs check be done in non-root mode as well (was previously skipped since the kdat file was being stored in HOME).
- Fixed some linking issues affecting unittest (it now uses some of the new caps util funcs @adrianreber added but the function defs were not visible at link time).
- Addressed some review comments.

criu/fdstore.c
ymanton: That would be simpler and nicer I agree. On the other hand it would require 4 syscalls total for non-root non-CAP_NET_ADMIN processes, v.s. 2 currently. Since this code is on the restore path can we keep the "free" bit of extra performance we have with the current patch?

(Also I assume you meant SO_SNDBUFFORCE fall back to SO_SNDBUF and SO_RCVBUFFORCE fall back to SO_RCVBUF.)

criu/fdstore.c
```diff
@@ -49,8 +53,16 @@ int fdstore_init(void)
...
-	    setsockopt(sk, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1], sizeof(buf[1])) < 0) {
+	if (has_cap_net_admin(opts.cap_eff)) {
```
avagin, line 56: nit: if (!opts.unprivileged ||  has_cap_net_admin(opts.cap_eff))

criu/files.c
```diff
@@ -1332,6 +1332,8 @@ int prepare_fds(struct pstree_item *me)
 
 static int fchroot(int fd)
 {
+	if (!has_cap_sys_chroot(opts.cap_eff))
```
avagin, line 1335: We need to add a comment that explain this. Right now, it looks very suspicious...

criu/image.c
```diff
@@ -226,7 +226,8 @@ int prepare_inventory(InventoryEntry *he)
 	if (get_task_ids(&crt.i))
 		return -1;
 
-	he->has_root_cg_set = true;
+	if (!opts.uid)
```
avagin: As I already suggested, we need to check opts.unprivileged instead of opts.uid. First, it is guaranteed that a user knows that the unprivileged mode is inuse. Second, it will simplify testing of this feature.

criu/kerndat.c
```diff
@@ -1060,19 +1062,51 @@ static int kerndat_has_openat2(void)
...
+{
+	int ret;
+
+	/*
+	 * Running as non-root, even with CAP_CHECKPOINT_RESTORE, does not
+	 * allow to write to KDAT_RUNDIR which usually is only writable by root.
+	 * Let's write .criu.kdat file to the home directory for non-root cases.
+	 */
+	if (opts.uid) {
+		if (!getenv("HOME")) {
```
avagin: should we store the kdat file in XDG_RUNTIME_DIR?
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

criu/kerndat.c
```diff
@@ -1060,19 +1062,51 @@ static int kerndat_has_openat2(void)
...
+static int get_kerndat_filename(char **kdat_file)
+{
+	int ret;
+
+	/*
+	 * Running as non-root, even with CAP_CHECKPOINT_RESTORE, does not
+	 * allow to write to KDAT_RUNDIR which usually is only writable by root.
+	 * Let's write .criu.kdat file to the home directory for non-root cases.
+	 */
+	if (opts.uid) {
```
avagin: nit: opts.unprivileged

criu/include/kerndat.h
```diff
@@ -8,6 +8,9 @@
 #include "util-vdso.h"
 #include "hugetlb.h"
 
+#define RELEASE_SIZE 66
+#define NODENAME_SIZE 66
+
```
adrianreber, line 13: Is this still needed?

criu/image.c
ymanton: Sorry, I realized you're probably looking at the changes patch by patch. All the `opts.uid` issues were fixed in ef9d3a0b6a5a31e3f63a7a1d4d68fb8796c1e26d.

criu/kerndat.c
ymanton: Yes, done in a71f7dc750081a841d6cfe73199019580037a62f.

criu/kerndat.c
ymanton: Fixed in ef9d3a0b6a5a31e3f63a7a1d4d68fb8796c1e26d as suggested, thanks.

criu/fdstore.c
ymanton: I don't have a strong opinion about this, but since root has all caps just checking caps is simpler than checking two conditions where one implies the other. Or are you suggesting that we should only ever look at the caps data in unprivileged mode? The caps data is initialized on all paths currently.

avagin: I mean that we should only ever look at the caps data in unprivileged mode. Otherwise, we expect a complete set of caps, and we have to fail if that isn't true.

criu/fdstore.c
avagin: ack-ed.

avagin:
Thanks for working on this. I want to ask one more thing. We are trying to support the commit history in the main tree as clean as possible. The main idea is very similar to the Linux kernel process. In this PR, I see that a few patches fix problems introduced by other patches. I think we need to merge all fixes in proper changes. I understand that the origin patches are authored by @adrianreber. I think we can use tags like Co-authored-by, Co-developed-by, Originally-by to mention all developers involved in the process.

criu/include/kerndat.h
ymanton: Nope, missed removing it. Done now, thanks.

criu/fdstore.c
ymanton: Done, all checks that look at caps now only do so if unprivileged. Thanks.

criu/kerndat.c
ymanton: All kerndat changes are now in their own patch. Thanks.

ymanton:
> Thanks for working on this. I want to ask one more thing. We are trying to support the commit history in the main tree as clean as possible. The main idea is very similar to the Linux kernel process. In this PR, I see that a few patches fix problems introduced by other patches. I think we need to merge all fixes in proper changes. I understand that the origin patches are authored by @adrianreber. I think we can use tags like Co-authored-by, Co-developed-by, Originally-by to mention all developers involved in the process.

No problem, thanks for letting me know. I've merged all my changes together with @adrianreber's and added Co-authored-bys to the patches that were significantly altered. Going forward I'll maintain them this way as things progress.

criu/files.c
ymanton: Not sure about it myself. @adrianreber it looks to me like in unprivileged mode we skip restoring the chroot if we don't have the cap for it without signaling any errors. Is this because we are assuming that the process cannot be chrooted under these conditions and this step will be a no-op? Should we be paranoid and check that we're not restoring a process outside of a chroot jail?

criu/cr-check.c
ymanton: I've changed this to check the kernel version in unprivileged mode. On kernel >= 5.16 we'll check that `sem_next_id` is accessible. On older kernels we'll skip it, since there's nothing the user can do besides run as root.

adrianreber: I would prefer if you could avoid using version checks. For kernels using backporting the version check will not work.

ymanton: True. I've reverted this back to skipping the check for non-root because `access()` runs with an empty set of caps and will always fail for non-root users. I didn't realize that before.

Checking `opts.unprivileged` is pointless because 1) non-root users must have it set, otherwise we don't reach this code and 2) if a root uses `--unprivileged` for whatever reason we'll skip the check but we don't skip actually trying to restore `sem_next_id` if it was saved, so we'd just lose an opportunity to catch a problem.

adrianreber (approved):
Looks good. Thanks for picking it up and thanks for solving the open questions.

criu/files.c
```diff
@@ -1332,6 +1332,8 @@ int prepare_fds(struct pstree_item *me)
 
 static int fchroot(int fd)
 {
+	if (opts.unprivileged && !has_cap_sys_chroot(opts.cap_eff))
+		return 0;
```
avagin, line 1336: This isn't acceptable. We need to find out where fchroot is used and how to handle it there properly.

test/zdtm.py
```diff
@@ -618,11 +622,15 @@ def available():
                 ["make", "zdtm_ct"], env=dict(os.environ, MAKEFLAGS=""))
         if not os.access("zdtm/lib/libzdtmtst.a", os.F_OK):
             subprocess.check_call(["make", "-C", "zdtm/"])
+        if opts['rootless'] or os.getuid() != 0:
```
avagin, line 625: we don't need to check uid. rootless means that we want to test the unprivileged mode.

test/zdtm.py
```diff
@@ -1156,7 +1168,10 @@ def __criu_act(self, action, opts=[], log=None, nowait=False):
         if action == "restore":
             preexec = None
         else:
-            preexec = self.__user and self.set_user_id or None
+            if os.getuid():
+                preexec = None
+            else:
+                preexec = self.__user and self.set_user_id or None
```
avagin, line 1174: hmm, does the --user option do what we need?

test/zdtm.py
```diff
@@ -1329,6 +1344,9 @@ def dump(self, action, opts=[]):
         os.mkdir(self.__ddir())
         os.chmod(self.__ddir(), 0o777)
 
+        if self.__rootless:
+            os.setgid(NON_ROOT_UID)
```
avagin, line 1348: I don't think that we need to change uid in zdtm.py. I think it should be better if a caller executes zdtm.py from a non-root user.

test/zdtm/lib/test.c
```diff
@@ -239,34 +242,36 @@ void test_init(int argc, char **argv)
...
-	if (val) {
-		char *tok = NULL;
-		unsigned int size = 0, groups[NGROUPS_MAX];
-
-		tok = strtok(val, " ");
-		while (tok) {
-			size++;
-			groups[size - 1] = atoi(tok);
-			tok = strtok(NULL, " ");
+	if (!uid) {
```
avagin, line 245: imho: It is better to pass ZDTM_ROOTLESS from zdtm.py.

criu/files.c
ymanton: So `fchroot()` is only called from one place on the restore path.

I've changed the code to check if the current fs root is the same as the process root in the dump. If they are the same we can skip `fchroot()` for both root and non-root since it is a no-op. If roots are not the same we will attempt the chroot operation and fail if CRIU doesn't have the required privileges.

test/zdtm.py
ymanton: Fixed in multiple places, thanks.

test/zdtm.py
ymanton: Removed in 2 placed, thanks.

test/zdtm/lib/test.c
ymanton: Fixed as suggested.

test/zdtm.py
ymanton: I took a quick look. The code that currently exists around this option is pretty minimal and was introduced in dde2adbc9d5a988a3624ab1056d43cef6cfd58c0.

It removes `suid` and `nouser` tests and `ns` and `uns` flavours from the launcher config and runs only `criu dump` as non-root, but that's not sufficient. Dumps fail when trying to read `/proc/$process/map_files`.

`--rootless` drops to user before invoking `zdtm_ct`.

It seems like `--user` hasn't been exercised in a while and is broken, unless I'm missing something.

adrianreber: I think we had the same discussion in my original PR with the same result that `--user` is a bit different. It is used in one of the Jenkins script, although I am not sure if that script is actually used by Jenkins.

anna-singleton:
Hi, apologies if this is the wrong way to address this (i am fairly new to using open-source gh stuff, please redirect me to the correct place if this is the wrong one). I've been doing some testing of this branch to see if it might be useful for my use-case when merged, and in order to dump a process, i need to give criu the `cap_sys_ptrace` capability in addition to `cap_checkpoint_restore` or `cap_sys_admin` or else it fails with this error message:
```
Warn  (compel/src/lib/infect.c:133): Unable to interrupt task: 3891500 (Operation not permitted)
Error (compel/src/lib/infect.c:418): Unable to detach from 3891500: No such process
Error (criu/cr-dump.c:2059): Dumping FAILED.
```
but when given `cap_sys_ptrace` it works as expected for processes not requiring the `--shell-job` / `-j` option.

Furthermore, processes that require the `-j` option are dumpable with `cap_checkpoint_restore` and `cap_sys_ptrace` but must be restored with `cap_sys_admin` due to failing on an `ioctl(fd , TIOCSLCKTRMIOS)` call which requires `cap_sys_admin` according to the manpage. If there is anything I can do about this, or if I should be able to do this, but am doing something wrong please let me know.

I hope this information is at least somewhat useful to you.

anna-singleton:
alright thanks for the confirmation. Would it be helpful to add the information about `cap_sys_ptrace` in the NON-ROOT section of the manpage? I am happy to do this (i presume i can PR the branch), since it would help others trying to use `--unprivileged` since I could not find this documented previously.

ymanton:
Yeah I ran into these kinds of issues as well and talked about them a bit, but forgot about the manpage patch in this PR.

Adding `CAP_SYS_PTRACE` and/or `ptrace_scope` and other such limitations is a good idea. @anna-singleton feel free to suggest changes via PR or the GitHub suggestions mechanism or whatever else works. I'll update the manpage text as well with any relevant info before this gets merged.

criu/files.c
```diff
@@ -1372,11 +1390,15 @@ int restore_fs(struct pstree_item *me)
 	 * Now do chroot/chdir. Chroot goes first as it calls chdir into
 	 * dd_root so we'd need to fix chdir after it anyway.
 	 */
-
-	ret = fchroot(dd_root);
-	if (ret < 0) {
-		pr_perror("Can't change root");
-		goto out;
+	if (need_chroot(dd_root)) {
```
avagin, line 1393: We don't need to do any specific actions if opts.unprivileged isn't set. The code can look like this:
```
if (opts.unprivileged)
  check_root(dd_root);
else {
		ret = fchroot(dd_root);
		if (ret < 0) {
			pr_perror("Can't change root");
			goto out;
		}
}
```

criu/files.c
```diff
@@ -1346,6 +1346,24 @@ static int fchroot(int fd)
...
+static int need_chroot(int saved_root)
+{
+	struct stat saved_root_stat, cur_root_stat;
+
+	if (fstat(saved_root, &saved_root_stat) == -1) {
+		pr_perror("Failed to stat saved root dir");
+		return -1;
+	}
+
+	if (stat("/proc/self/root", &cur_root_stat) == -1) {
```
avagin, line 1358: pls use open_pid_proc, open_proc, open_proc_rw to access proc files.
```
psd = open_pid_proc(PROC_SELF);
if (fstatat(psd, "root", &cur_root_stat) == -1)  {
...
```

criu/cr-check.c
```diff
@@ -1666,36 +1676,43 @@ static int pr_set_dumpable(int value)
...
+		return -1;
 	}
 
-	return exit_code;
+	/*
+	 * At his point we know we are running as non-root with the necessary
+	 * capabilities available. Now we have to make the process dumpable
+	 * so that /proc/self is not owned by root.
+	 */
+	if (pr_set_dumpable(1))
```
avagin, line 1709: hmm, can it have any security implications? Will a user be able to trace a criu process?

criu/files.c
ymanton: Fixed, the check will only be done in unprivileged mode.

criu/files.c
ymanton: Fixed, thanks.

criu/cr-check.c
ymanton: I looked into this a bit. It's here because CRIU will look at its own `/proc/self/pagemap` when initializing `kdat.pmap` and `kdat.zero_page_pfn`. It can possibly be changed to look at target process' pagemap (or parent's pagemap for the criu-check case), but its a little tricky because we need to find a present page to check if PFNs are available or only flags. The current code uses an easily available stack address to find a present page.

As far as security implications go I don't think there is anything sensitive in the CRIU process itself that will be exposed by tracing or core dumps, and the target process in unprivileged mode will be itself be unprivileged (meaning user can already trace/dump it), so we are not exposing anything new there.

avagin:
@ymanton I think I am ready to merge it to the criu-dev branch, but you need to fix CI failures.

avagin:
Why CI checks have not been executed?..

adrianreber:
The cross-compile test results fail sometimes because of broken repositories like in this case and can be ignored.

The CentOS 7 errors are because of a package installation problem. @rst0git I think the epel package installation fix is from you. Can you maybe add a `||:` after the epel activation line. The problem is that on CentOS 7 epel might already be installed, if you test it in RHEL the package is not installed. CentOS 7 is anyway not relevant for the non-root CRIU operation as the kernel is too old.

The Vagrant errors seem to be something new for shared memory already tracked somewhere else: #1982

The non-root Vagrant test setup is running successfully.

I think from the CI side this is ready.

ymanton:
@adrianreber Thanks for double checking, I was unsure about `Vagrant Fedora based test (no VDSO)` and started looking at it.

avagin:
@ymanton thank you for moving this pr to the finish line. @adrianreber thank you for implementing the kernel part and the initial version of the userspace changes.

criu/fdstore.c
imachug: FYI, the kernel uses `sockopt_capable(CAP_NET_ADMIN)` rather than `sockopt_ns_capable(CAP_NET_ADMIN)` (e.g. see code `sk_setsockopt` on HEAD) to check permissions. This means that for a namespaced process with CAP_NET_ADMIN inside the namespace but not outside, this code would attempt to use *FORCE and the kernel would reject it.

criu/cr-check.c
imachug: I would argue that this logic is somewhat faulty. `sem_next_id` is not always restored, yes, but I might be fine with that and rather see a warning than an error that prevents other checks from running. It's not intuitive either: it feels weird that a check passes for a non-root user and fails in a more privileged context, rather than the other way round. It took me a while to figure out why that happened, I don't think I'd realise the reason without consulting the source code.

imachug:
Perhaps this should better be asked in a separate issue, but I'm not sure if what I'm asking about is reasonable, so here goes.

One thing this PR has not handled at all is `map_files`. Problem is, proc(5) says that

> Capabilities are required to read the contents of the symbolic links in this directory: before Linux 5.9, the reading process requires CAP_SYS_ADMIN in the initial user namespace; since Linux 5.9, the reading process must have either CAP_SYS_ADMIN or CAP_CHECKPOINT_RESTORE in the user namespace where it resides.

...but that does not match what the kernel does, e.g. on HEAD we have

https://github.com/torvalds/linux/blob/27bc50fc90647bbf7b734c3fc306a5e61350da53/fs/proc/base.c#L2249

...which checks whether CAP_CHECKPOINT_RESTORE is present in the *init* user namespace. This means that reading `map_files` *always* fails in containers. Now one would think this is seldom needed, but the kernel uses a deleted `/dev/zero` device for every shared anonymous mapping:

https://github.com/torvalds/linux/blob/aae703b02f92bde9264366c545e87cec451de471/mm/shmem.c#L4243-L4264

I am not sure if deleted files can be gracefully handled in the general case, but perhaps it would be reasonable to at least hard-code the expected behavior for `/dev/zero (deleted)`?

imachug:
You certainly know what you are doing more than me, but perhaps it might be of interest to you to know what I needed to patch to get CRIU of httpd in podman working:

<details>
<summary>Patch</summary>

```diff
diff --git a/criu/proc_parse.c b/criu/proc_parse.c
index 946b0fc40..a24783da7 100644
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -313,25 +313,8 @@ static int vma_get_mapfile_user(const char *fname, struct vma_area *vma, struct
 
 	vfi_dev = makedev(vfi->dev_maj, vfi->dev_min);
 
-	if (is_memfd(vfi_dev)) {
-		char tmp[PATH_MAX];
-		strlcpy(tmp, fname, PATH_MAX);
-		strip_deleted(tmp, strlen(tmp));
-
-		/*
-		 * The error EPERM will be shown in the following pr_perror().
-		 * It comes from the previous open() call.
-		 */
-		pr_perror("Can't open mapped [%s]", tmp);
-
-		/*
-		 * TODO Perhaps we could do better than failing and dump the
-		 * memory like what is being done in shmem.c
-		 */
-		return -1;
-	}
-
 	if (is_hugetlb_dev(vfi_dev, &hugetlb_flag) || is_anon_shmem_map(vfi_dev)) {
+		vma->e->status |= VMA_AREA_REGULAR;
 		if (!(vma->e->flags & MAP_SHARED))
 			vma->e->status |= VMA_ANON_PRIVATE;
 		else
diff --git a/criu/shmem.c b/criu/shmem.c
index 81e701586..2de2ea9af 100644
--- a/criu/shmem.c
+++ b/criu/shmem.c
@@ -724,7 +724,7 @@ static int next_data_segment(int fd, unsigned long pfn, unsigned long *next_data
 	return 0;
 }
 
-static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si)
+static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si, bool seek_data_supported)
 {
 	struct page_pipe *pp;
 	struct page_xfer xfer;
@@ -750,7 +750,7 @@ static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si)
 		unsigned long pgaddr;
 		int st = -1;
 
-		if (pfn >= next_hole_pfn && next_data_segment(fd, pfn, &next_data_pnf, &next_hole_pfn))
+		if (seek_data_supported && pfn >= next_hole_pfn && next_data_segment(fd, pfn, &next_data_pnf, &next_hole_pfn))
 			goto err_xfer;
 
 		if (si->pstate_map && is_shmem_tracking_en()) {
@@ -808,20 +808,50 @@ static int dump_one_shmem(struct shmem_info *si)
 {
 	int fd, ret = -1;
 	void *addr;
+	bool seek_data_supported;
 
 	pr_info("Dumping shared memory %ld\n", si->shmid);
 
-	fd = open_proc(si->pid, "map_files/%lx-%lx", si->start, si->end);
-	if (fd < 0)
-		goto err;
 
-	addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
-	if (addr == MAP_FAILED) {
-		pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
-		goto errc;
+	fd = __open_proc(si->pid, EPERM, O_RDONLY, "map_files/%lx-%lx", si->start, si->end);
+	if (fd >= 0) {
+		addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
+		if (addr == MAP_FAILED) {
+			pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			goto errc;
+		}
+
+		seek_data_supported = true;
+	} else {
+		if(errno != EPERM) {
+			goto err;
+		}
+
+		fd = open_proc(si->pid, "mem");
+		if(fd < 0) {
+			goto err;
+		}
+
+		addr = mmap(NULL, si->size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+		if (addr == MAP_FAILED) {
+			pr_err("Can't map empty space for shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			goto errc;
+		}
+
+		if(lseek(fd, si->start, SEEK_SET) < 0) {
+			pr_perror("Can't seek virtual memory");
+			return -1;
+		}
+
+		if(read(fd, addr, si->size) < si->size) {
+			pr_perror("Can't read virtual memory");
+			return -1;
+		}
+
+		seek_data_supported = false;
 	}
 
-	ret = do_dump_one_shmem(fd, addr, si);
+	ret = do_dump_one_shmem(fd, addr, si, seek_data_supported);
 
 	munmap(addr, si->size);
 errc:
@@ -849,7 +879,7 @@ int dump_one_memfd_shmem(int fd, unsigned long shmid, unsigned long size)
 		goto err;
 	}
 
-	ret = do_dump_one_shmem(fd, addr, &si);
+	ret = do_dump_one_shmem(fd, addr, &si, true);
 
 	munmap(addr, size);
 err:
@@ -875,7 +905,7 @@ int dump_one_sysv_shmem(void *addr, unsigned long size, unsigned long shmid)
 	if (fd < 0)
 		return -1;
 
-	ret = do_dump_one_shmem(fd, addr, si);
+	ret = do_dump_one_shmem(fd, addr, si, true);
 	close(fd);
 	return ret;
 }
diff --git a/criu/sockets.c b/criu/sockets.c
index db772707b..d50866c65 100644
--- a/criu/sockets.c
+++ b/criu/sockets.c
@@ -12,6 +12,7 @@
 
 #include "int.h"
 #include "bitops.h"
+#include "cr_options.h"
 #include "libnetlink.h"
 #include "sockets.h"
 #include "unix_diag.h"
@@ -457,6 +458,27 @@ int sk_collect_one(unsigned ino, int family, struct socket_desc *d, struct ns_id
 
 int do_restore_opt(int sk, int level, int name, void *val, int len)
 {
+	void *buf;
+	socklen_t cur_len;
+	bool match;
+
+	if(opts.unprivileged) {
+		buf = xmalloc(len);
+		if (!buf) {
+			return -1;
+		}
+		cur_len = len;
+		if (getsockopt(sk, level, name, buf, &cur_len) < 0) {
+			pr_perror("Can't get %d:%d (len %d)", level, name, len);
+			return -1;
+		}
+		match = cur_len == len && memcmp(buf, val, len) == 0;
+		xfree(buf);
+		if(match) {
+			return 0;
+		}
+	}
+
 	if (setsockopt(sk, level, name, val, len) < 0) {
 		pr_perror("Can't set %d:%d (len %d)", level, name, len);
 		return -1;
@@ -469,9 +491,20 @@ static int sk_setbufs(void *arg, int fd, pid_t pid)
 {
 	u32 *buf = (u32 *)arg;
 
-	if (restore_opt(fd, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0]))
+	int rcv_opt_name;
+	int snd_opt_name;
+
+	if (!opts.unprivileged) {
+		rcv_opt_name = SO_RCVBUFFORCE;
+		snd_opt_name = SO_SNDBUFFORCE;
+	} else {
+		rcv_opt_name = SO_RCVBUF;
+		snd_opt_name = SO_SNDBUF;
+	}
+
+	if (restore_opt(fd, SOL_SOCKET, rcv_opt_name, &buf[0]))
 		return -1;
-	if (restore_opt(fd, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1]))
+	if (restore_opt(fd, SOL_SOCKET, snd_opt_name, &buf[1]))
 		return -1;
 
 	return 0;
```

</details>

The omission of `vma->e->status |= VMA_AREA_REGULAR;` seems to be a bug that should probably have been noticed earlier.  Using lseek + read + vmsplice is inefficient, I guess `process_vm_readv` should be used instead, but I just wanted to get it working first. I also updated `do_restore_opt` to not set options if they are already correct because we might not have enough permissions for that, and `SO_RCVBUF`/`SO_SNDBUF` are used instead of their *FORCE equivalents in unprivileged mode. Maybe some of this is technically wrong, please take a better look.

I tested this inside

```
podman run -it --cap-add=cap_checkpoint_restore --cap-add=cap_sys_ptrace --cap-add=cap_sys_resource --security-opt seccomp=unconfined --security-opt unmask=/proc/sys php:8.1-apache bash
```

CAP_CHECKPOINT_RESTORE is needed for obvious reasons, CAP_SYS_PTRACE is for `process_vm_readv` IIRC, CAP_SYS_RESOURCE is for `rlimit`, seccomp is so that `vmsplice` does not yield EPERM (I don't quite understand the reason for this one tbh), `/proc/sys` is unmasked for `ns_next_pid`. CAP_NET_ADMIN is *not* used due to a bug described in one of my review comments on this PR.

Hope this helps someone.

ymanton:
@imachug Thanks for testing, I'll get back to you shortly on your comments.

criu/cr-check.c
ymanton: Just so I'm clear, you're not concerned about the new behaviour that this PR introduced (which is _`check_ipc()` always returns success for non-root users_), right? Rather, you are suggesting that under root this check should be non-fatal so the later checks can be performed?

If this is the case can you open a new issue referencing this comment and make your suggestion there? This check was added in https://github.com/checkpoint-restore/criu/commit/03791bc2c3059d2412f4c19ec4b9b71540792a62 when all checks were fatal. Non-fatal checks were added in https://github.com/checkpoint-restore/criu/commit/70cb4326b61fde36edf669e536fddc3ee4c90295.

I'm not exactly clear on whether you want just `check_ipc()` made non-fatal or you want all checks to be non-fatal.

imachug: What I was *actually* wondering about is the right way to handle `sem_next_id` being unavailable in IPC namespaces, but it turns out I was just using an old kernel (yeah, I'm still using Ubuntu). https://github.com/torvalds/linux/commit/1f5c135ee509e89e0cc274333a65f73c62cb16e5 fixes this in 5.19+. I'll upgrade the kernel soon-ish and check if this really fixes the problem in my use case, but if it does, I think no changes are necessary here. Sorry for confusion.

ymanton: I see, thanks.

I think having all checks be non-fatal so `criu check` will report everything at once is more useful than the current behavior, implemented here: https://github.com/checkpoint-restore/criu/blob/18c6426eaeebc5fe7d0f9ca0acb592a3ec828b0c/criu/cr-check.c#L1434-L1511

ymanton:
> Perhaps this should better be asked in a separate issue, but I'm not sure if what I'm asking about is reasonable, so here goes.
> 
> One thing this PR has not handled at all is `map_files`. Problem is, proc(5) says that
> 
> > Capabilities are required to read the contents of the symbolic links in this directory: before Linux 5.9, the reading process requires CAP_SYS_ADMIN in the initial user namespace; since Linux 5.9, the reading process must have either CAP_SYS_ADMIN or CAP_CHECKPOINT_RESTORE in the user namespace where it resides.
> 
> ...but that does not match what the kernel does, e.g. on HEAD we have
> 
> https://github.com/torvalds/linux/blob/27bc50fc90647bbf7b734c3fc306a5e61350da53/fs/proc/base.c#L2249
> 
> ...which checks whether CAP_CHECKPOINT_RESTORE is present in the _init_ user namespace. This means that reading `map_files` _always_ fails in containers.

You're right about the man-page, it appears to be incorrect. I've sent an email to the maintainers and list and CC'd you. Thanks for catching that. As for what to do about it or how to improve it, I'm not exactly sure. Wouldn't the kernel behaviour have to change or is there something you think CRIU can do better if we're not in the init user ns but need access to `map_files` contents?

imachug:
Can we avoid using `map_files` in this case and read the memory of the process directly?

criu/fdstore.c
ymanton: You're right. @avagin's original suggestion of how this should be done would have handled that more gracefully. I'll open a PR.

imachug:
Ah, I guess I see what the problem is: we want to dump the whole file, and `mmap` may only map the file partially. Moreover, for deleted files, we might not even know if two mappings are of the same file or not. Given that `mremap` does not really work for shared memory, I guess the best that can be done is a) attempting to modify one mapping and checking how it affects others, to figure out which mappings are parts of the same file, and then b) dumping different parts of each file from memory of different processes, if no single process contains the whole file. This leaves the case when a part of a file is not mapped anywhere, but then we can just fill it with zeroes -- that should be indistinguishable from the "real" file if `map_files` is unavailable.

Issue #1935, opened 2022-07-07
glibc-2.34-38.fc35 breaks criu: SEGV in restore


edsantiago:
Latest Fedora 35:
```console
# bin/podman run -d quay.io/libpod/testimage:20220615 top
685009ff5e33a06ffa1209ece8f5c229ce10537f2cdfc6b48ce522e4dc209904
# bin/podman container checkpoint 685
685009ff5e33a06ffa1209ece8f5c229ce10537f2cdfc6b48ce522e4dc209904
# bin/podman container restore 685
Error: OCI runtime error: crun: CRIU restoring failed -52.  Please check CRIU logfile /var/lib/containers/storage/overlay-containers/685009ff5e33a06ffa1209ece8f5c229ce10537f2cdfc6b48ce522e4dc209904/userdata/restore.log
```
[criu-restore.log](https://github.com/checkpoint-restore/criu/files/9066788/criu-restore.log). Last few lines:
```
(00.030324)      1: Thread    0 stack  0x1b080 rt_sigframe  0x23080
(00.030354)      1: Going to chroot into /proc/self/fd/8
(00.030360)      1: Restoring umask to 22
(00.030379)      1: task_args: 0x25000
task_args->pid: 1
task_args->nr_threads: 1
task_args->clone_restore_fn: 0x11e70
task_args->thread_args: 0x25580(00.030385) pie: 1: Switched to the restorer 1
(00.030595) pie: 1: Mapping native vDSO at 0x2d000
(00.030622) pie: 1: vdso: Using gettimeofday() on vdso at 0x31790
(00.030634) pie: 1:     mmap(0x55bf5b61f000 -> 0x55bf5b62b000, 0x3 0x12 3)
(00.030641) pie: 1:     mmap(0x55bf5b62b000 -> 0x55bf5b6c6000, 0x7 0x12 3)
(00.030645) pie: 1:     mmap(0x55bf5b6c6000 -> 0x55bf5b6e8000, 0x3 0x12 3)
(00.030648) pie: 1:     mmap(0x55bf5b6e8000 -> 0x55bf5b6ec000, 0x3 0x12 3)
(00.030652) pie: 1:     mmap(0x55bf5b6ec000 -> 0x55bf5b6ed000, 0x3 0x12 3)
(00.030688) Error (criu/cr-restore.c:1510): 33055 stopped by signal 11: Segmentation fault
(00.030823) mnt: Switching to new ns to clean ghosts
[...]

PR #1937, opened 2022-07-20
cr-restore: rseq: dynamically handle *libc with rseq

mihalicyn:
Before this patch we assumed that CRIU is compiled against
the same GLibc as it runs with. But as we see from real
world examples like #1935 it's not always true.

The idea of this patch is to detect rseq configuration
for the main CRIU process and use it to unregister
rseq for all further child processes. It's correct,
because we restore pstree using clone*() syscalls,
don't use exec*() (!) syscalls, so rseq gets inherited
in the kernel and rseq configuration remains the same
for all children processes.

This will prevent issues like this:
https://github.com/checkpoint-restore/criu/issues/1935

Suggested-by: Florian Weimer <fweimer@redhat.com>

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -3087,7 +3087,6 @@ static int prep_rseq(struct rst_rseq_param *rseq, ThreadCoreEntry *tc)
 	return 0;
 }
 
-#if defined(__GLIBC__) && defined(RSEQ_SIG)
 static void prep_libc_rseq_info(struct rst_rseq_param *rseq)
 {
 	if (!kdat.has_rseq) {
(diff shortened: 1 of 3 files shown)


fweimer-rh:
Would it make sense to remove the glibc-specific code, so that the generic code receives wider test coverage? If you fork & ptrace anyway to get other things, the additional overhead will be really small, I assume.

mihalicyn:
> Would it make sense to remove the glibc-specific code, so that the generic code receives wider test coverage? If you fork & ptrace anyway to get other things, the additional overhead will be really small, I assume.

agreed, Florian. Have done ;) I've left "old approach" only as a fallback

minhbq-99 (approved):

minhbq-99:
LGTM

Issue #1916, opened 2022-06-15
Error: Compatible vdso's size is bigger than reserved buf


goosesima:
**CRIU logs and information:**

<details><summary>Output of `sudo criu check --all`:</summary>
<p>

```
Error: Compatible vdso's size is bigger than reserved buf
Error (criu/vdso.c:441): vdso: Compat vdso helper exited with 255
Error (criu/vdso.c:498): vdso: Failed to mmap compatible vdso with helper process
Error (criu/vdso.c:620): vdso: Failed to fill compat vdso symtable
Error (criu/kerndat.c:1594): kerndat_vdso_fill_symtable failed when initializing kerndat.
Error (criu/crtools.c:260): Could not initialize kernel features detection.
```

</p>
</details>

<details><summary>Output of `criu --version`:</summary>
<p>

```
Version: 3.17
GitID: v3.17
```

</p>
</details>

<details><summary>Output of `criu check --all`:</summary>
<p>

```
Error (criu/util.c:641): exited, status=4
Error (criu/util.c:641): exited, status=4
Warn  (criu/proc_parse.c:930): Write 4294967295 to /proc/self/loginuid failed: Operation not permitted
Warn  (criu/net.c:3427): Unable to get tun network namespace
Warn  (criu/sk-unix.c:224): unix: Unable to open a socket file: Operation not permitted
Error (criu/net.c:3711): Unable create a network namespace: Operation not permitted
Warn  (criu/net.c:3767): NSID isn't reported for network links
Error: Compatible vdso's size is bigger than reserved buf
Error (criu/vdso.c:441): vdso: Compat vdso helper exited with 255
Error (criu/vdso.c:498): vdso: Failed to mmap compatible vdso with helper process
Error (criu/vdso.c:620): vdso: Failed to fill compat vdso symtable
Error (criu/kerndat.c:1594): kerndat_vdso_fill_symtable failed when initializing kerndat.
Error (criu/crtools.c:260): Could not initialize kernel features detection.
```

</p>
</details>

<details><summary>Output of `zcat /proc/config.gz | grep CHECKPOINT`:</summary>
<p>

```
CONFIG_CHECKPOINT_RESTORE=y
```

</p>
</details>

**Additional environment details:**
OS: Arch Linux x86_64
Kernel: 5.18.3-zen1-1-zen

avagin:
@0x7f454c46 could you take a look at this issue if you have time?

minhbq-99:
[...]

PR #1938, opened 2022-07-21
vdso-compat: Increase the reserved buffer for compat vdso

minhbq-99:
On Arch Linux with 5.18.3-zen1-1-zen kernel, the vdso's size is 3 pages which
exceeds the current 2-page reserved buffer. This commit simply increases the
reserved buffer size to 4 pages.

Fixes: https://github.com/checkpoint-restore/criu/issues/1916

Signed-off-by: Bui Quang Minh <minhquangbui99@gmail.com>

### DIFF
--- a/criu/vdso.c
+++ b/criu/vdso.c
@@ -479,7 +479,7 @@ out_close:
 	return ret;
 }
 
-#define COMPAT_VDSO_BUF_SZ (PAGE_SIZE * 2)
+#define COMPAT_VDSO_BUF_SZ (PAGE_SIZE * 4)
 static int vdso_fill_compat_symtable(struct vdso_maps *native, struct vdso_maps *compat)
 {
 	void *vdso_mmap;


mihalicyn:
Looks reasonable. Did you manage to find zen-kernel specific patch which changed compat vdso size?
I can't see any specific vdso patches in they kernel (https://github.com/zen-kernel/zen-kernel).

minhbq-99:
I can't see either, I guess it's because the compiler they use to build the kernel or maybe they set some configs.

mihalicyn:
cc @0x7f454c46

0x7f454c46:
Seems reasonable. We could probably mmap() a bigger buffer as likely most systems have overcommit enabled and it won't consume any memory before the vdso pages are copied out to the buffer.
Anyway, LGTM

0x7f454c46 (approved):

tuananh:
im getting this `Error: Compatible vdso's size is bigger than reserved buf` on kernel 6.11.1-zen1-1-zen still. not sure if that's related to this?

avagin:
@tuananh could you file a new issue?

PR #1939, opened 2022-07-21
zdtm: Remove permission part check for skipping vsyscall vma

minhbq-99:
Normally, vsyscall vma has VM_READ, VM_EXEC permission. However, when
CONFIG_LEGACY_VSYSCALL_XONLY=y, that vma only has VM_EXEC. This commit removes
the permission part when checking to skip vsyscall vma in x32 tests.

Signed-off-by: Bui Quang Minh <minhquangbui99@gmail.com>

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1651,6 +1651,15 @@ def get_visible_state(test):
     return files, maps, mounts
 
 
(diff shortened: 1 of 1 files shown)


mihalicyn:
LGTM

Issue #1942, opened 2022-07-23
Fedora Rawhide Test  fails


avagin:
```
2022-07-21T16:45:39.2503402Z STEP 4/9: RUN /bin/prepare-for-fedora-rawhide.sh
2022-07-21T16:45:39.8858886Z error running container: error from /usr/bin/crun creating container for [/bin/sh -c /bin/prepare-for-fedora-rawhide.sh]: sd-bus call: Connection reset by peer
2022-07-21T16:45:39.8860140Z : exit status 1
2022-07-21T16:45:40.8531781Z Error: error building at STEP "RUN /bin/prepare-for-fedora-rawhide.sh": error while running runtime: exit status 1
```
https://github.com/checkpoint-restore/criu/runs/7453998261?check_suite_focus=true

rst0git:
It looks like this problem has been reported in https://github.com/containers/podman/issues/14920 and https://github.com/actions/runner/issues/1994

PR #1943, opened 2022-07-23
ci: Fix Fedora Rawhide build

rst0git:
When `sudo -E` is used to run Podman it results in the `XDG_RUNTIME_DIR` environment variable being set incorrectly that prevents Podman from running.

This PR also resolves include conflicts with glibc 2.36:
https://sourceware.org/glibc/wiki/Release/2.36#Usage_of_.3Clinux.2Fmount.h.3E_and_.3Csys.2Fmount.h.3E

- `criu/include/linux/mount.h` has been updated to match `/usr/include/sys/mount.h`
- inline definitions `sys_*()` functions have been moved from "linux/mount.h" to "syscall.h" to avoid conflicts with `uapi/compel/plugins/std/syscall.h` and `<unistd.h>`.
- include of `<linux/aio_abi.h>` has been replaced with local include to avoid conflicts with `<sys/mount.h>`.

Fixes: #1942
Fixes: #1949


.github/workflows/fedora-rawhide-test.yml
```diff
@@ -9,4 +9,6 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Run Fedora Rawhide Test
-      run: sudo -E make -C scripts/ci fedora-rawhide CONTAINER_RUNTIME=podman BUILD_OPTIONS="--security-opt seccomp=unconfined"
+      # Make sure that XDG_RUNTIME_DIR environment variable is not set.
+      # https://github.com/checkpoint-restore/criu/issues/1942
```
avagin, line 13: pls don't add issue in the code if it isn't todo/fixme. The comment has to contain all required info to understand the code.

.github/workflows/fedora-rawhide-test.yml
```diff
@@ -9,4 +9,6 @@ jobs:
...
+      run: sudo -E XDG_RUNTIME_DIR= make -C scripts/ci fedora-rawhide CONTAINER_RUNTIME=podman BUILD_OPTIONS="--security-opt seccomp=unconfined"
```
avagin, line 14: why do we set -E in this case? Do we pass any env var-s?

adrianreber (approved):
I have seen this also mentioned as a possible solution. Looks good to me.

.github/workflows/fedora-rawhide-test.yml
adrianreber: We need environment variables from the CI environment to distinguish between CI environments.

.github/workflows/fedora-rawhide-test.yml
rst0git: Thanks Andrei, I have updated the comment with FIXME for the GitHub issue in Podman.

mihalicyn:
LGTM

mihalicyn (approved):

PR #1955, opened 2022-08-05
cr-check: fix check for apparmor stacking

rst0git:
The feature check for AppArmor stacking was introduced in commit 8723e3f998d1ec5f125e6600436a96f7ff9c1631. However, on systems that don't support AppArmour, this check always fails. As a result, `criu check --all` shows the following message:

> Looks good but some kernel features are missing which, depending on your process tree, may cause dump or restore failure.

### DIFF
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -104,7 +104,7 @@ out:
 
 static int check_apparmor_stacking(void)
 {
-	if (!check_aa_ns_dumping())
+	if (!kdat.apparmor_ns_dumping_enabled)
 		return -1;
 
 	return 0;
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -1478,13 +1478,15 @@ int cr_check(void)
 		ret |= check_newifindex();
 		ret |= check_pidfd_store();
 		ret |= check_ns_pid();
-		ret |= check_apparmor_stacking();
 		ret |= check_network_lock_nftables();
 		ret |= check_sockopt_buf_lock();
 		ret |= check_memfd_hugetlb();
 		ret |= check_move_mount_set_group();
 		ret |= check_openat2();
 		ret |= check_ptrace_get_rseq_conf();
+
+		if (kdat.lsm == LSMTYPE__APPARMOR)
+			ret |= check_apparmor_stacking();
 	}
 
 	/*


criu/cr-check.c
```diff
@@ -1478,13 +1478,15 @@ int cr_check(void)
...
 		ret |= check_ns_pid();
-		ret |= check_apparmor_stacking();
 		ret |= check_network_lock_nftables();
 		ret |= check_sockopt_buf_lock();
 		ret |= check_memfd_hugetlb();
 		ret |= check_move_mount_set_group();
 		ret |= check_openat2();
 		ret |= check_ptrace_get_rseq_conf();
+
+		if (kdat.lsm == LSMTYPE__APPARMOR)
```
Snorch, line 1488: Can we put check inside check_apparmor_stacking? And also while on it, why do we put result of check_aa_ns_dumping in kdat itself, instead of doing the same check on each invocation?

rst0git: > Can we put check inside check_apparmor_stacking?

The `check_apparmor_stacking()` function is used for both `criu check --all` and `criu check --feature apparmor_stacking`. Thus, if we add this check inside `check_apparmor_stacking()` and return 0, `criu check --feature apparmor_stacking` would print `apparmor_stacking is supported` on systems where AppArmor is not enabled.

To avoid this ambiguity, I've added the check in `cr_check()`. This affects only the `criu check --all` command. When AppArmor is not enabled it would test this feature, but `criu check --feature apparmor_stacking` would still exit with 1.

> why don't we put result of check_aa_ns_dumping in kdat itself, instead of doing the same check on each invocation?

It looks like we already have this:

https://github.com/checkpoint-restore/criu/blob/5a0136988a328cd0db8c944459238ab4e509a90b/criu/lsm.c#L226

Snorch: > To avoid this ambiguity, I've added the check in cr_check().

Reasonable enough.

> It looks like we already have this

Can you please also do the below:

```
diff --git a/criu/cr-check.c b/criu/cr-check.c
index f589a91da..9c2e84a35 100644
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -104,7 +104,7 @@ static int check_tty(void)
 
 static int check_apparmor_stacking(void)
 {
-       if (!check_aa_ns_dumping())
+       if (!kdat.apparmor_ns_dumping_enabled)
                return -1;
 
        return 0;
```

Snorch (approved):

PR #1956, opened 2022-08-05
ci: fix docker test

rst0git:
In commits https://github.com/checkpoint-restore/criu/commit/046cad8 and https://github.com/checkpoint-restore/criu/commit/81a68ad the version of containerd installed by default in the GitHub CI virtual environment was replaced with the latest release from GitHub as a workaround to a bug in containerd.

This bug has been fixed sometime ago and the current default version of containerd (1.6.6) does not require this workaround. However, with the latest release, the containerd binaries uploaded on GitHub have been built for Ubuntu 22.04 (https://github.com/containerd/containerd/commit/6b2dc9a37). Our tests are still running on Ubuntu 20.04 and this results in the following error:
    
    /usr/bin/containerd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by /usr/bin/containerd)
    /usr/bin/containerd: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by /usr/bin/containerd)

### DIFF
--- a/scripts/ci/docker-test.sh
+++ b/scripts/ci/docker-test.sh
@@ -21,10 +21,7 @@ add-apt-repository \
 
 . /etc/lsb-release
 
-# overlayfs with current Ubuntu kernel breaks CRIU
-# https://bugs.launchpad.net/ubuntu/+source/linux-azure/+bug/1967924
-# Use devicemapper storage drive as a work-around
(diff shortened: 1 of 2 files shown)


adrianreber (approved):
Nice to see that we can remove the workarounds again.

PR #1957, opened 2022-08-08
breakpoint: implement hw breakpoint for arm64 platform 

avagin:
This is slightly reworked #1815. @time-river, could you review it?


time-river:
Great! @avagin 
Thanks your rework.

PR #1959, opened 2022-08-11
docker-test: handle race condition error

rst0git:
There is a race condition in docker/containerd that causes docker to occasionally fail when starting a container from a checkpoint immediately after the checkpoint has been created. This problem is unrelated to criu and has been reported in https://github.com/moby/moby/issues/42900

### DIFF
--- a/scripts/ci/docker-test.sh
+++ b/scripts/ci/docker-test.sh
@@ -75,17 +75,37 @@ checkpoint_container () {
 	docker wait cr
 }
 
-restore_container () {
-	CHECKPOINT_NAME=$1
-
-	docker start --checkpoint "$CHECKPOINT_NAME" cr 2>&1 | tee log || {
+print_logs () {
 	cat "$(grep log 'log file:' | sed 's/log file:\s*//')" || true
 		docker logs cr || true
 		cat $CRIU_LOG || true
 		dmesg
 		docker ps
 		exit 1
-	}
+}
+
+declare -i max_restore_container_tries=3
+current_iteration=
+
+restore_container () {
+	CHECKPOINT_NAME=$1
+
+	docker start --checkpoint "$CHECKPOINT_NAME" cr 2>&1 | tee log || {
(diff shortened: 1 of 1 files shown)


scripts/ci/docker-test.sh
```diff
@@ -75,17 +75,37 @@ checkpoint_container () {
...
 		docker logs cr || true
 		cat $CRIU_LOG || true
 		dmesg
 		docker ps
 		exit 1
-	}
+}
+
+declare -i max_restore_container_tries=3
+current_iteration=
```
avagin, line 88: why is it declared outside of restore_container?

scripts/ci/docker-test.sh
```diff
@@ -75,17 +75,37 @@ checkpoint_container () {
...
+
+restore_container () {
+	CHECKPOINT_NAME=$1
+
+	docker start --checkpoint "$CHECKPOINT_NAME" cr 2>&1 | tee log || {
+		# FIXME: There is a race condition in docker/containerd that causes
+		# docker to occasionally fail when starting a container from a
+		# checkpoint immediately after the checkpoint has been created.
+		# https://github.com/moby/moby/issues/42900
+		if [ "$current_iteration" -gt "$max_restore_container_tries" ]; then
```
avagin, line 98: https://github.com/checkpoint-restore/criu/actions/runs/6948976416/job/18929387032?pr=2299
```
+ '[' '' -gt 3 ']'
1800
./docker-test.sh: line 100: [: : integer expression expected
```

rst0git: Thanks for pointing this out. I opened a pull request to fix this: https://github.com/checkpoint-restore/criu/pull/2302

rst0git:
> There is a race condition in docker/containerd that causes docker to occasionally fail when starting a container from a checkpoint immediately after the checkpoint has been created.

It looks like this problem has now been fixed upstream (https://github.com/moby/moby/pull/47456).

PR #1964, opened 2022-08-26
Clean up Java testing, test with both HotSpot and OpenJ9 based VMs

ymanton:
The patches that added Java testing originally used AdoptOpenJDK's OpenJ9 Docker images. Later we switched to Eclipse Temurin (which uses HotSpot) when Adopt stopped building OpenJ9, but most of the OpenJ9 references in filenames and such remained. Since HotSpot and OpenJ9 are completely different implementations there's some value in testing with both. This patch set renames files and variables as necessary and brings back testing with OpenJ9 using IBM Semeru Docker images.


scripts/ci/java-test.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+
+cd ../.. || exit 1
+
+failures=""
+
+docker build -t criu-openj9-ubuntu-test:latest -f scripts/build/Dockerfile.openj9-ubuntu .
+if ! docker run --rm --privileged criu-openj9-ubuntu-test:latest; then
+	failures="$failures ubuntu"
```
rst0git, line 9: It might be good to use `openj9-ubuntu` and `hotspot-ubuntu` to be able to distinguish between both errors.

ymanton: Agree, fixed.

rst0git (approved):
LGTM

PR #1968, opened 2022-09-10
cgroup-v2: Checkpoint/restore global properties and threaded controllers

minhbq-99:
#### 1. Checkpoint/restore some global properties in cgroup-v2 controller
These global properties of cgroup-v2 controller are checkpointed/restored
	cgroup.subtree_control
	cgroup.max.descendants
	cgroup.max.depth
	cgroup.freeze
	cgroup.type
#### 2. Checkpoint/restore threaded controllers
Threaded controllers allow threads in process to belong to different controllers which break the current assumption in CRIU. This PR reworks the logic to dump cgroup controllers for each thread, add a cgroupd service the fixup the cgroup controllers of threads when they are restored.

Currently, the zdtm tests run only on CI local test (Compat test, CentOS 9) not on CI tests run inside container because the container runtime seems does not support hybrid (both v1 and v2) cgroup controller

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -8,6 +8,7 @@
 #include <ftw.h>
 #include <libgen.h>
 #include <sched.h>
+#include <sys/wait.h>
 
 #include "common/list.h"
 #include "xmalloc.h"
@@ -55,6 +56,7 @@ static u32 cg_set_ids = 1;
 
 static LIST_HEAD(cgroups);
 static unsigned int n_cgroups;
+static pid_t cgroupd_pid;
 
 static CgSetEntry *find_rst_set_by_id(u32 id)
 {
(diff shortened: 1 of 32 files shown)


criu/cgroup.c
```diff
@@ -1821,6 +1919,162 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
...
+		kill(cgroupd_pid, SIGTERM);
+		waitpid(cgroupd_pid, NULL, 0);
+
+		sigprocmask(SIG_SETMASK, &oldmask, NULL);
+	}
+
+	return 0;
+}
+
+static int prepare_cgroup_thread_sfd(void)
```
mihalicyn, line 2033: IMHO, it's better to make generalized helper like `start_service_daemon(...)`. Because this helper code is really close to the `start_usernsd`.

test/zdtm/static/cgroupv2_00.c
```diff
@@ -0,0 +1,131 @@
...
+#include "zdtmtst.h"
+
+const char *test_doc = "Check that some cgroup-v2 properties in kernel controllers are preserved";
+const char *test_author = "Bui Quang Minh <minhquangbui99@gmail.com>";
+
+char *dirname;
+TEST_OPTION(dirname, string, "cgroup-v2 directory name", 1);
+const char *cgname = "subcg00";
+
+int write_value(const char *path, const char *value)
```
mihalicyn, line 15: Let's move `write_value`/`read_value` helpers to `zdtm/lib`. As far as you adding two usage cases for it + as far as I can see we already have analogical helpers in `cgroup04` test.

mihalicyn (commented):
Great work!

I've left some comments, but this needs further testing and careful reviewing.

criu/pie/restorer.c
```diff
@@ -582,6 +583,104 @@ static void noinline rst_sigreturn(unsigned long new_sp, struct rt_sigframe *sig
...
+	ucred->pid = sys_gettid();
+	ucred->uid = 0;
+	ucred->gid = 0;
+
+	ret = sys_sendmsg(sk, &h, 0);
+	if (ret < 0) {
+		pr_err("Unable to send packet to cgroupd %d\n", ret);
+		return -1;
+	}
+	pr_info("restorer: sendmsg ret %d\n", ret);
```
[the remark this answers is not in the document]
minhbq-99: I forget to remove it after debugging, it seems necessary so I remove it.

minhbq-99:
Thanks for your review @mihalicyn. I just edit the PR message to clarify the CI test right now as the new zdtm tests run only in some local tests (CentOS9, Compat test) not those tests run inside container.

minhbq-99:
Got failed test in test/others/rpc on test_c on CentOS 8, CentOS 9. I add this patch to get some logs
```
diff --git a/criu/cgroup.c b/criu/cgroup.c
index 2cdb63609..f4f50fd38 100644
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -382,14 +382,19 @@ static int dump_cg_props_array(const char *fpath, struct cgroup_dir *ncd, const
        struct cgroup_prop *prop;

        for (j = 0; cgp && j < cgp->nr_props; j++) {
+               int ret;
+
                if (snprintf(buf, PATH_MAX, "%s/%s", fpath, cgp->props[j]) >= PATH_MAX) {
                        pr_err("snprintf output was truncated\n");
                        return -1;
                }

-               if (access(buf, F_OK) < 0 && errno == ENOENT) {
+               ret = access(buf, F_OK);
+               if (ret < 0 && errno == ENOENT) {
                        pr_info("Couldn't open %s. This cgroup property may not exist on this kernel\n", buf);
                        continue;
+               } else if (ret < 0) {
+                       pr_perror("cgroup: Path: %s", buf);
                }

                prop = create_cgroup_prop(cgp->props[j]);
```

Error log
```
(00.071032) Error (criu/cgroup.c:397): cg: cgroup: Path: /proc/self/fd/16/bar/cgroup.clone_children: Permission denied
(00.071045) cg: Dumping value 0 from /proc/self/fd/16/bar/cgroup.clone_children
(00.071049) Error (criu/cgroup.c:397): cg: cgroup: Path: /proc/self/fd/16/bar/notify_on_release: Permission denied
(00.071056) cg: Dumping value 0 from /proc/self/fd/16/bar/notify_on_release
(00.071060) Error (criu/cgroup.c:397): cg: cgroup: Path: /proc/self/fd/16/bar/cgroup.procs: Permission denied
(00.071066) cg: Dumping value  from /proc/self/fd/16/bar/cgroup.procs
(00.071069) Error (criu/cgroup.c:397): cg: cgroup: Path: /proc/self/fd/16/bar/tasks: Permission denied
(00.071075) cg: Dumping value  from /proc/self/fd/16/bar/tasks
(00.071079) Error (criu/cgroup.c:397): cg: cgroup: Path: /proc/self/fd/16/bar/cgroup.subtree_control: Permission denied
(00.071084) Error (criu/cgroup.c:292): cg: Failed opening /proc/self/fd/16/bar/cgroup.subtree_control: No such file or directory
(00.071085) Error (criu/cgroup.c:462): cg: dumping global properties failed
(00.071088) Error (criu/cgroup.c:732): cg: failed walking /proc/self/fd/16/ for empty cgroups: No such file or directory
```

I don't understand the root cause here as the files do not exist because this is a cgroup v1 controller but the error codes when `access` and `open` are different. I can't reproduce this test so hope someone can give more insights. I implement a workaround here to dump cgroup v2 global properties only when the controller is cgroup v2 and the CI test succeeds.

avagin:
LGTM

@minhbq-99 could you rebase this pr?

avagin (approved):

minhbq-99:
I just rebased the PR

PR #1971, opened 2022-09-14
seize: do not overwrite exit code from failpath

liusdu:


liusdu:
According to @Snorch 's advise. This patch is part of https://github.com/checkpoint-restore/criu/pull/1905.

PR #1979, opened 2022-10-01
Fix CentOS 7 CI

rst0git:
python2-future, python2-junit_xml, python-flake8 and libbsd-devel are now provided from the EPEL package repository.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -111,6 +111,8 @@ task:
     memory: 8G
 
   setup_script: |
+    # EPEL is needed for python2-future, python2-junit_xml, python-flake8 and libbsd-devel.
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -111,6 +111,8 @@ task:
     memory: 8G
 
   setup_script: |
+    # EPEL is needed for python2-future, python2-junit_xml, python-flake8 and libbsd-devel.
+    rpm -Uvh https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
adrianreber, line 115: Doing `yum -y install <URL>` is probably better than directly calling `rpm`. Feels more robust.

rst0git: Thanks Adrian! I have updated the pull request.

Issue #1982, opened 2022-10-16
ci: Vagrant Fedora no vdso gets permission denied when writing to ipc sysctl


minhbq-99:
**Description**

The Vagrant Fedora no vdso CI gets some failed tests in user namespace mode when writing to ipc sysctl (/proc/sys/kernel/{sem_next_id,msg_next_id,shm_next_id})

```
################### 5 TEST(S) FAILED (TOTAL 451/SKIPPED 35) ####################
 * zdtm/static/msgque(uns)
 * zdtm/static/shm(uns)
 * zdtm/static/shm-mp(uns)
 * zdtm/static/sem(uns)
 * zdtm/static/shm-unaligned(uns)
```

```
========================== Run zdtm/static/sem in uns ==========================
Start test
./sem --pidfile=sem.pid --outfile=sem.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/sem/166/1/restore.log
------------------------ grep Error ------------------------
b'(00.005610)      1: Restoring IPC semaphores sets'
b'(00.005623)      1: id: 0          key: 0x1d2015a6 uid: 18943      gid: 58467      cuid: 18943      cgid: 58467      mode: 777        (00.005627)      1: nsems: 10'
b'(00.005634)      1: uns: calling __userns_sysctl_op (13, 0)'
b'(00.005654) uns: daemon calls 0x4aa8b0 (192, 10, 0)'
b"(00.005908) Error (criu/sysctl.c:100): Can't write kernel/sem_next_id: Operation not permitted"
b'(00.006563) Error (criu/sysctl.c:313): worker failed: 256'
b'(00.006596)      1: Error (criu/ipc_ns.c:579): Failed to set desired IPC sem ID'
b'(00.006602)      1: Error (criu/ipc_ns.c:644): Failed to prepare semaphores set'
b'(00.007358) Error (criu/cr-restore.c:1503): 192 exited, status=1'
b'(00.007387) Warn  (criu/cr-restore.c:2524): Unable to wait 192: No child processes'
b'(00.007402) uns: calling exit_usernsd (-1, 1)'
b'(00.007422) uns: daemon calls 0x478fe0 (189, -1, 1)'
b'(00.007427) uns: `- daemon exits w/ 0'
b'(00.007977) uns: daemon stopped'
b'(00.007988) Error (criu/cr-restore.c:2537): Restoring FAILED.'
------------------------ ERROR OVER ------------------------
```

[...]

PR #1984, opened 2022-10-23
ipc_sysctl: Prioritize restoring IPC variables using non usernsd approach

minhbq-99:
Since commit https://github.com/torvalds/linux/commit/5563cabdde, user with enough capability can open IPC sysctl files and write to them. Therefore, we don't need to use usernsd process in the outside user namespace to help with that anymore. Furthermore, some later commits:
https://github.com/torvalds/linux/commit/1f5c135ee5, https://github.com/torvalds/linux/commit/0889f44e28 bind the IPC namespace to the opened file descriptor of IPC sysctl at the open() time, the changed value does not depend on the IPC namespace of write() time anymore. This breaks the current usernsd approach.

So, we prioritize opening/writing IPC sysctl files in the context of restored process directly without usernsd help. This approach succeeds in the newer kernel since the restored process has enough capabilities at this restore stage. With older kernel, the open() fails and we fallback to the usernsd approach.

Fixes: #1982

### DIFF
--- a/criu/include/sysctl.h
+++ b/criu/include/sysctl.h
@@ -34,8 +34,9 @@ enum {
 /*
  * Some entries might be missing mark them as optional.
  */
-#define CTL_FLAGS_OPTIONAL	1
-#define CTL_FLAGS_HAS		2
-#define CTL_FLAGS_READ_EIO_SKIP 4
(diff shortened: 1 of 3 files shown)


adrianreber (approved):
Sounds reasonable and it fixes the CI error we have been seeing.

PR #1985, opened 2022-10-25
Switch aarch64 builds to Cirrus CI

adrianreber:
It seems like drone.io no longer provides free aarch64/armhf CI runs.

This switches the aarch64 CI runs to Cirrus CI. armhf CI runs have been dropped for now as they are not directly supported.


adrianreber:
The remaining CI errors are already understood and being worked on (as far as I know).

PR #1990, opened 2022-11-07
ci: move cgroup unmounting to run-ci-tests.sh

adrianreber:
A previous commit added a cgroup cpuset unmounting to scripts/ci/Makefile. We are sometimes running in a container without the necessary privileges to unmount certain cgroups.

This commit moves the cgroup unmounting to a place in run-ci-tests.sh which already requires privileged access and does not break unprivileged build-only CI runs.

CC: @minhbq-99


minhbq-99 (approved):
Thanks for the fix, @adrianreber 
LGTM

rst0git (approved):

Issue #1986, opened 2022-10-26
Cross Compilation for ARM


sakgoyal:
I am trying to compile CRIU for ARM (Raspberry Pi 3b+) on my x86 windows laptop through WSL2. I am looking at the cross compilation page on the website and it seems like it is out of date. the url for the toolchain binary does not exist anymore so those instructions no longer work. 

How can I cross compile CRIU for my ARMv8 Pi 3B+ on my x86 computer through WSL2

adrianreber:
Please take a look how we cross compile during CI runs: https://github.com/checkpoint-restore/criu/blob/criu-dev/.github/workflows/cross-compile.yml

thesaadmemon:
@sakgoyal did you manage to get it working?

sakgoyal:
sorry ive been a little busy so I havent gotten around to it. but also, I'm not familiar with how github actions work or how CI/CD works so any help with that (directly or through other resources) would be greatly appreciated.

thesaadmemon:
I did compile the kernel with the pre-required kernel features, however, it looks like something is still missing

OS: Raspbian 32-bit; Kernel-Version: 5.15.76-v7l; CRIU Version: v3.17.1

Output: 

root@raspberrypi:~# criu check
Error (criu/kerndat.c:508): Unexpected error from memfd_create("", MFD_HUGETLB): Function not implemented
Error (criu/kerndat.c:1577): kerndat_has_memfd_hugetlb failed when initializing kerndat.
Error (criu/crtools.c:260): Could not initialize kernel features detection.

adrianreber:
Maybe you forgot to enable support for MEMFD? `CONFIG_MEMFD_CREATE=y`

adrianreber:
Not sure. The error message seems to point to that `memfd_create()` is not available.

minhbq-99:
The error seems because `memfd_create` is available but `MFD_HUGETLB` flag is not but the returned error from `memfd_create` is not expected. Currently, we expect to get `EINVAL` and `ENOENT` in those cases.
https://github.com/checkpoint-restore/criu/blob/d3ed3e90876964334526000f275c6f13cbf2e856/criu/kerndat.c#L490-L511
For quick fix, you can add `ENOSYS` to the errno list here
https://github.com/checkpoint-restore/criu/blob/d3ed3e90876964334526000f275c6f13cbf2e856/criu/kerndat.c#L503
[...]

PR #1993, opened 2022-11-08
kerndat: Mark memfd_create(MFD_HUGETLB) unavailable when ENOSYS is returned

minhbq-99:
Some users on Raspberry Pi report that the kerndat checking for memfd_create(MFD_HUGETLB) support returns ENOSYS even when memfd_create syscall is available. We currently treat this error as unexpected and return error. This commit marks the memfd_create(MFD_HUGETLB) as unavailable when ENOSYS is returned.

Fixes: #1986

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -500,7 +500,7 @@ static bool kerndat_has_memfd_hugetlb(void)
 	if (ret >= 0) {
 		kdat.has_memfd_hugetlb = true;
 		close(ret);
-	} else if (ret == -1 && (errno == EINVAL || errno == ENOENT)) {
+	} else if (ret == -1 && (errno == EINVAL || errno == ENOENT || errno == ENOSYS)) {
 		kdat.has_memfd_hugetlb = false;
 	} else {
 		pr_perror("Unexpected error from memfd_create(\"\", MFD_HUGETLB)");


adrianreber:
The change makes kind of sense. I had a look at the kernel code but I have not found the place where `memfd_create()` returns `ENOSYS` if using `MFD_HUGETLB`. It is a bit strange why this happens.

brauner:
> The change makes kind of sense. I had a look at the kernel code but I have not found the place where `memfd_create()` returns `ENOSYS` if using `MFD_HUGETLB`. It is a bit strange why this happens.

seccomp filter?

rst0git:
>It is a bit strange why this happens.

It looks like when `memfd_create()` is used with `MFD_HUGETLB`, it calls [`hugetlb_file_setup()`](https://github.com/raspberrypi/linux/blob/cad0c67b64bb9de6bc970c0a7afbccabec0fcf9d/mm/memfd.c#L315L321) and that could fail with [`-ENOSYS`](https://github.com/raspberrypi/linux/blob/6b7afd0fbd9777d8d13d10c330062edf5e8baaba/include/linux/hugetlb.h#L496L501).  I also noticed that this behavior has been documented in the [man page for Freebsd](https://www.freebsd.org/cgi/man.cgi?query=memfd_create) as follows:
```
[ENOSYS]		In memfd_create, MFD_HUGETLB was specified in flags,
			and this system	does not support forced	hugetlb	map-
			pings.
```

rst0git (approved):

adrianreber:
Thanks. Now it makes sense. I haven't looked at the header file.

PR #1996, opened 2022-11-15
non-root: Fix socket restore issues

ymanton:
Two fixes here related to running in unprivileged mode.

The first avoilds restoring `SO_MARK` on the service socket since we know the semantics of this socket and we know it doesn't need it.

The second changes how we use the `SO_*BUF*` options. As noted by @imachug[^1], checking if we have `CAP_NET_ADMIN` unfortunately doesn't tell us if we can really use these options. The easiest approach as suggested by @avagin is to simply try the `SO_*BUFFORCE` versions first and, if in unprivileged mode, fall back to the `SO_*BUF` ones on failure.

These two patches together allows users to avoid needing `CAP_NET_ADMIN` when doing unprivileged self-dump and restore, since the service socket is included in the dump and needs to be restored.

[^1]: https://github.com/checkpoint-restore/criu/pull/1930#discussion_r1007439001

### DIFF
--- a/criu/fdstore.c
+++ b/criu/fdstore.c
@@ -15,6 +15,7 @@
 #include "util.h"
 #include "cr_options.h"
 #include "util-caps.h"
+#include "sockets.h"
 
 /* clang-format off */
 static struct fdstore_desc {
@@ -29,8 +30,6 @@ int fdstore_init(void)
 	uint32_t buf[2] = { INT_MAX / 2, INT_MAX / 2 };
 	struct sockaddr_un addr;
 	unsigned int addrlen;
-	int rcv_opt_name;
-	int snd_opt_name;
 	struct stat st;
 	int sk, ret;
 
@@ -53,17 +52,7 @@ int fdstore_init(void)
 		return -1;
 	}
 
-	if (!opts.unprivileged || has_cap_net_admin(opts.cap_eff)) {
-		rcv_opt_name = SO_RCVBUFFORCE;
-		snd_opt_name = SO_SNDBUFFORCE;
-	} else {
-		rcv_opt_name = SO_RCVBUF;
-		snd_opt_name = SO_SNDBUF;
-	}
-
-	if (setsockopt(sk, SOL_SOCKET, snd_opt_name, &buf[0], sizeof(buf[0])) < 0 ||
-	    setsockopt(sk, SOL_SOCKET, rcv_opt_name, &buf[1], sizeof(buf[1])) < 0) {
-		pr_perror("Unable to set SO_SNDBUFFORCE/SO_RCVBUFFORCE");
+	if (sk_setbufs(sk, buf)) {
 		close(sk);
 		return -1;
 	}
--- a/criu/pidfd-store.c
+++ b/criu/pidfd-store.c
@@ -13,6 +13,7 @@
 #include "log.h"
 #include "util.h"
 #include "pidfd-store.h"
+#include "sockets.h"
 
 struct pidfd_entry {
 	pid_t pid;
@@ -94,9 +95,7 @@ int init_pidfd_store_sk(pid_t pid, int sk)
 	 * This is similar to how fdstore_init() works.
 	 */
 	if (addrlen == sizeof(sa_family_t)) {
-		if (setsockopt(pidfd_store_sk, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0], sizeof(buf[0])) < 0 ||
-		    setsockopt(pidfd_store_sk, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1], sizeof(buf[1])) < 0) {
-			pr_perror("Unable to set SO_SNDBUFFORCE/SO_RCVBUFFORCE");
+		if (sk_setbufs(pidfd_store_sk, buf)) {
 			goto err;
 		}
 
--- a/criu/sockets.c
+++ b/criu/sockets.c
@@ -29,6 +29,7 @@
 #include "pstree.h"
 #include "util.h"
 #include "fdstore.h"
+#include "cr_options.h"
 
 #undef LOG_PREFIX
 #define LOG_PREFIX "sockets: "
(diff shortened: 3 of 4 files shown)


ymanton:
I think the `SO_MARK` fix might be done differently: at restore time check the current mark value and if it differs from the saved one only then would we try to restore it. AFAICT in the kernel setting the mark value doesn't have any side-effects so restoring the same mark value appears to be a no-op and therefore safe to skip.

criu/sk-unix.c
```diff
@@ -530,6 +530,12 @@ static int dump_one_unix_fd(int lfd, uint32_t id, const struct fd_parms *p)
 	if (dump_socket_opts(lfd, skopts))
 		goto err;
 
+	/* We need CAP_NET_ADMIN to restore SO_MARK, but we know the
+	 * service socket doesn't use it so avoid restoring it so that
+	 * the user doesn't have to grant the cap just for this. */
+	if (opts.unprivileged && (ue->uflags & USK_SERVICE))
+		skopts->has_so_mark = false;
```
avagin, line 537: should we do something like this:
```
index db772707b..43246bbf8 100644
--- a/criu/sockets.c
+++ b/criu/sockets.c
@@ -631,8 +631,9 @@ int dump_socket_opts(int sk, SkOptsEntry *soe)
        ret |= dump_opt(sk, SOL_SOCKET, SO_PRIORITY, &soe->so_priority);
        soe->has_so_rcvlowat = true;
        ret |= dump_opt(sk, SOL_SOCKET, SO_RCVLOWAT, &soe->so_rcvlowat);
-       soe->has_so_mark = true;
        ret |= dump_opt(sk, SOL_SOCKET, SO_MARK, &soe->so_mark);
+       if (soe->so_mark != 0)
+               soe->has_so_mark = true;
 
        ret |= dump_opt(sk, SOL_SOCKET, SO_SNDTIMEO, &tv);
        soe->so_snd_tmo_sec = tv.tv_sec;
```

criu/fdstore.c
```diff
@@ -53,17 +52,8 @@ int fdstore_init(void)
 		return -1;
 	}
 
-	if (!opts.unprivileged || has_cap_net_admin(opts.cap_eff)) {
```
avagin, line 56: pls, add a few lines in the commit message to explain what problem you are solving

criu/sk-unix.c
ymanton: Thanks, I've changed the patch to do basically this but added `!unprivileged` so that existing behaviour for privileged mode is unchanged.

criu/fdstore.c
ymanton: Thanks, added the following:

>SO_SNDBUFFORCE/SO_RCVBUFFORCE require root or CAP_NET_ADMIN.
We can use SO_SNDBUF/SO_RCVBUF in some cases and avoid
needing elevated privileges.

criu/sockets.c
```diff
@@ -647,8 +647,13 @@ int dump_socket_opts(int sk, SkOptsEntry *soe)
 	ret |= dump_opt(sk, SOL_SOCKET, SO_PRIORITY, &soe->so_priority);
 	soe->has_so_rcvlowat = true;
 	ret |= dump_opt(sk, SOL_SOCKET, SO_RCVLOWAT, &soe->so_rcvlowat);
-	soe->has_so_mark = true;
+	/*
+	 * Restoring SO_MARK requires root or CAP_NET_ADMIN. Avoid saving it
+	 * in unprivileged mode if still has its default value.
+	 */
 	ret |= dump_opt(sk, SOL_SOCKET, SO_MARK, &soe->so_mark);
+	if (!opts.unprivileged || soe->so_mark != 0)
```
avagin, line 655: We don't need to check unprivileged here. This patch doesn't change existing behavior.

criu/fdstore.c
```diff
@@ -53,17 +52,8 @@ int fdstore_init(void)
...
-	} else {
-		rcv_opt_name = SO_RCVBUF;
-		snd_opt_name = SO_SNDBUF;
-	}
-
-	if (setsockopt(sk, SOL_SOCKET, snd_opt_name, &buf[0], sizeof(buf[0])) < 0 ||
-	    setsockopt(sk, SOL_SOCKET, rcv_opt_name, &buf[1], sizeof(buf[1])) < 0) {
-		pr_perror("Unable to set SO_SNDBUFFORCE/SO_RCVBUFFORCE");
+	if (sk_setbufs(sk, buf)) {
+		pr_perror("Unable to set socket snd/rcv buffer size");
```
Snorch, line 56: sk_setbufs->restore_opt->do_restore_opt already prints error messages, so we can probably remove pr_error here.

criu/sockets.c
```diff
@@ -465,18 +466,33 @@ int do_restore_opt(int sk, int level, int name, void *val, int len)
...
 {
-	u32 *buf = (u32 *)arg;
+	uint32_t sndbuf = bufs[0], rcvbuf = bufs[1];
 
-	if (restore_opt(fd, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0]))
-		return -1;
-	if (restore_opt(fd, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1]))
-		return -1;
+	if (restore_opt(sk, SOL_SOCKET, SO_SNDBUFFORCE, &sndbuf) ||
+	    restore_opt(sk, SOL_SOCKET, SO_RCVBUFFORCE, &rcvbuf)) {
```
Snorch, line 474: sk_setbufs->restore_opt->do_restore_opt has an error printing. In case of fallback printing errors may be confusing. Maybe it's better to use setsockopt here directly to have full control over error printing?

criu/fdstore.c
ymanton: Done, thanks.

criu/sockets.c
ymanton: Agreed, fixed as you suggested.

criu/sockets.c
```diff
@@ -465,18 +466,35 @@ int do_restore_opt(int sk, int level, int name, void *val, int len)
...
+			return -1;
+		}
+	}
 
 	return 0;
 }
 
+static int sk_setbufs_ns(void *arg, int fd, pid_t pid)
+{
+	u32 *buf = (u32 *)arg;
```
Snorch, line 493: Can you please use uint32_t* for bufs here like in sk_setbufs. Also probably we don't need variable at all and can do something like `return sk_setbufs(fd, (uint32_t *)arg);` to shorten the code?

criu/fdstore.c
Snorch: pr_error is still there after sk_setbufs error, probably you lost this hunk.

criu/pidfd-store.c
```diff
@@ -94,9 +95,8 @@ int init_pidfd_store_sk(pid_t pid, int sk)
 	 * This is similar to how fdstore_init() works.
 	 */
 	if (addrlen == sizeof(sa_family_t)) {
-		if (setsockopt(pidfd_store_sk, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0], sizeof(buf[0])) < 0 ||
-		    setsockopt(pidfd_store_sk, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1], sizeof(buf[1])) < 0) {
-			pr_perror("Unable to set SO_SNDBUFFORCE/SO_RCVBUFFORCE");
+		if (sk_setbufs(pidfd_store_sk, buf)) {
+			pr_perror("Unable to set socket snd/rcv buffer size");
```
Snorch, line 99: Here also we don't need pr_perror.

criu/sockets.c
```diff
@@ -631,8 +647,13 @@ int dump_socket_opts(int sk, SkOptsEntry *soe)
...
 	soe->has_so_rcvlowat = true;
 	ret |= dump_opt(sk, SOL_SOCKET, SO_RCVLOWAT, &soe->so_rcvlowat);
-	soe->has_so_mark = true;
+	/*
+	 * Restoring SO_MARK requires root or CAP_NET_ADMIN. Avoid saving it
+	 * in unprivileged mode if still has its default value.
+	 */
 	ret |= dump_opt(sk, SOL_SOCKET, SO_MARK, &soe->so_mark);
+	if (soe->so_mark != 0)
+		soe->has_so_mark = true;
```
Snorch, line 656: nit: this if does not look nice to me =)

1) Normally we don't write `if (var != 0)` in criu, instead we use shorter `if (var)`.
2) But in this exact case the whole construction can be simplified, the form:
```
if (!var)
        has_var = true;
```
can be replaced with
```
has_var = !!var;
```
As far as I see all has_* variables are of type protobuf_c_boolean which is int basically, so it is better to use "!!" to convert non-zero value of var to 1.

Snorch (approved):
Except for small nit about has_*, LGTM.

criu/sockets.c
avagin: Both ways work for me. @Snorch thanks for review.

PR #1998, opened 2022-11-17
Remove execute bit from source file

gibmat:
Follow up from https://github.com/checkpoint-restore/go-criu/pull/104


rst0git (approved):
LGTM

PR #1999, opened 2022-11-21
Fix warnings from -Wstrict-prototypes in clang 16.0.0

ajwock:
While building on a machine that has a HOL clang compiler, I ran into warnings regarding the changed lines.  It appears this warning is on by default because of anticipated changes to the C standard.


ajwock:
Their rationale is explained here: https://discourse.llvm.org/t/rfc-enabling-wstrict-prototypes-by-default-in-c/60521/9

PR #2001, opened 2022-11-23
ci/lint: install ShellCheck with dnf

rst0git:
The way ShellCheck is installed was changed in commit c056f99 to use v0.8.0 and remove some of the `"shellcheck disable=..."` annotations. Since then, Fedora 37 has been released and the ShellCheck package has been updated to v0.8.0.


adrianreber (approved):

adrianreber:
The Alpine error is fixed with #2002. Thanks.

PR #2003, opened 2022-11-24
ci: fix make indent

rst0git:
This patch fixes applies the changes required by clang-format v15.0.5 for `make indent`.


adrianreber (approved):
I like the look of the code with this version better than before.

Issue #1079, opened 2020-05-22
Intermittent travis alpine failures with socket-tcp6-local in user namespace


adrianreber:
I have seen the following test failure a couple of times during the last day in the alpine Travis runs. It goes away if the test is re-run, so it only happens sometimes.

```
=================== Run zdtm/static/socket-tcp6-local in uns ===================
Start test
./socket-tcp6-local --pidfile=socket-tcp6-local.pid --outfile=socket-tcp6-local.out
Run criu dump
Run criu restore
Send the 15 signal to  115
Wait for zdtm/static/socket-tcp6-local(115) to die for 0.100000
########### Test zdtm/static/socket-tcp6-local FAIL at result check ############
Test output: ================================
13:08:04.522:     5: ERR: tcp.c:128: can't connect to server (errno = 106 (Socket is connected))
13:08:04.522:     4: ERR: socket-tcp6-local.c:201: read less then have to (errno = 11 (Resource temporarily unavailable))
 <<< ================================
```

Snorch:
JFYI: I also see it, both on criu and vzcriu running on VZ7 (but in host flavour(both)):

```
==================== Run zdtm/static/socket-tcp6-local in h ====================
Start test
./socket-tcp6-local --pidfile=socket-tcp6-local.pid --outfile=socket-tcp6-local.out
Run criu dump
Run criu restore
Send the 15 signal to  52
Wait for zdtm/static/socket-tcp6-local(52) to die for 0.100000
########### Test zdtm/static/socket-tcp6-local FAIL at result check ############
Test output: ================================
10:08:14.380:    53: ERR: tcp.c:135: can't connect to server (errno = 106 (Transport endpoint is already connected))
10:08:14.380:    52: ERR: socket-tcp6-local.c:201: read less then have to (errno = 104 (Connection reset by peer))

 <<< ================================
```

https://ci.openvz.org/job/CRIU/job/CRIU-virtuozzo/job/criu-dev/6892/consoleFull

```
=================== Run zdtm/static/socket-tcp4v6-local in h ===================
Start test
./socket-tcp4v6-local --pidfile=socket-tcp4v6-local.pid --outfile=socket-tcp4v6-local.out
Run criu dump
=[log]=> dump/zdtm/static/socket-tcp4v6-local/36/1/dump.log
------------------------ grep Error ------------------------
(00.000000) cpu: Parsing /proc/vz/cpuid_override
[...]

PR #2006, opened 2022-11-30
 sk-inet: support of SOL_IP, IP_PKTINFO option

Snorch:
We see systemd-resolved relying on this option, and after migration this
option is lost and systemd-resolved stops serving dns requests.

The IP_PKTINFO socket option makes kernel add cmsg with destination
address to packets, see more how systemd-resolved uses it:

https://github.com/systemd/systemd/blob/00a60eaf5fcb3a0e415349aa649f2699550d26b0/src/resolve/resolved-manager.c#L826

### DIFF
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -14,6 +14,8 @@
 #include <linux/icmp.h>
 #include <linux/icmpv6.h>
 #include <poll.h>
+#include <linux/in.h>
+#include <linux/in6.h>
 
 #include "../soccr/soccr.h"
 
@@ -388,6 +390,10 @@ static int dump_ip_raw_opts(int sk, int family, int proto, IpOptsRawEntry *r)
 	return ret;
 }
 
+#ifndef IPV6_FREEBIND
+#define IPV6_FREEBIND 78
+#endif
+
 static int dump_ip_opts(int sk, int family, int type, int proto, IpOptsEntry *ioe)
 {
 	int ret = 0;
@@ -398,11 +404,18 @@ static int dump_ip_opts(int sk, int family, int type, int proto, IpOptsEntry *io
 		 * and fetch additional options.
 		 */
 		ret |= dump_ip_raw_opts(sk, family, proto, ioe->raw);
-	} else {
-		/* Due to kernel code we can use SOL_IP instead of SOL_IPV6 */
-		ret |= dump_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
-		ioe->has_freebind = ioe->freebind;
 	}
+
+	if (family == AF_INET6) {
+		if (kdat.has_ipv6_freebind)
+			ret |= dump_opt(sk, SOL_IPV6, IPV6_FREEBIND, &ioe->freebind);
+		else if (type != SOCK_RAW)
+			/* Due to kernel code we can use SOL_IP instead of SOL_IPV6 */
+			ret |= dump_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+	} else {
+		ret |= dump_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+	}
+	ioe->has_freebind = ioe->freebind;
 
 	return ret;
 }
@@ -787,8 +800,13 @@ int restore_ip_opts(int sk, int family, int proto, IpOptsEntry *ioe)
 {
 	int ret = 0;
 
-	if (ioe->has_freebind)
-		ret |= restore_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+	if (family == AF_INET6) {
+		if (ioe->has_freebind)
+			ret |= restore_opt(sk, SOL_IPV6, IPV6_FREEBIND, &ioe->freebind);
+	} else {
+		if (ioe->has_freebind)
+			ret |= restore_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+	}
 
 	if (ioe->raw)
 		ret |= restore_ip_raw_opts(sk, family, proto, ioe->raw);
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -412,10 +412,13 @@ static int dump_ip_opts(int sk, int family, int type, int proto, IpOptsEntry *io
 		else if (type != SOCK_RAW)
 			/* Due to kernel code we can use SOL_IP instead of SOL_IPV6 */
 			ret |= dump_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+		ret |= dump_opt(sk, SOL_IPV6, IPV6_RECVPKTINFO, &ioe->pktinfo);
 	} else {
 		ret |= dump_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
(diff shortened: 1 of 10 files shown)


Snorch:
@checkpoint-restore/maintainers Note: there are many other options unhandled https://github.com/torvalds/linux/blob/4bbf3422df78029f03161640dcb1e9d1ed64d1ea/include/uapi/linux/in.h#L97 This can be a good task for GSOC to check these options and dump those really needed.

rst0git:
> Note: there are many other options unhandled 

@Snorch Should we create a GitHub issue that lists the options that are not currently handled?

criu/sk-inet.c
```diff
@@ -402,6 +402,8 @@ static int dump_ip_opts(int sk, int family, int type, int proto, IpOptsEntry *io
 		/* Due to kernel code we can use SOL_IP instead of SOL_IPV6 */
 		ret |= dump_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
 		ioe->has_freebind = ioe->freebind;
+		ret |= dump_opt(sk, SOL_IP, IP_PKTINFO, &ioe->pktinfo);
```
rst0git, line 405: Would this also work with `IPV6_PKTINFO`?

Snorch:
> > Note: there are many other options unhandled
> 
> @Snorch Should we create a GitHub issue that lists the options that are not currently handled?

Will do.

criu/sk-inet.c
Snorch: Kernel code for IP(V6)_PKTINFO seem completely different unlike IP_FREEBIND:
```
case IP_PKTINFO:
                if (val)
                        inet->cmsg_flags |= IP_CMSG_PKTINFO;
                else
                        inet->cmsg_flags &= ~IP_CMSG_PKTINFO;
```
```
case IPV6_PKTINFO:
        {       
                struct in6_pktinfo pkt;
                                
                if (optlen == 0)
                        goto e_inval;
                else if (optlen < sizeof(struct in6_pktinfo) ||
                         sockptr_is_null(optval))
                        goto e_inval;

                if (copy_from_sockptr(&pkt, optval, sizeof(pkt))) {
                        retv = -EFAULT;
                        break;
                }
                if (!sk_dev_equal_l3scope(sk, pkt.ipi6_ifindex))
                        goto e_inval;

                np->sticky_pktinfo.ipi6_ifindex = pkt.ipi6_ifindex;
                np->sticky_pktinfo.ipi6_addr = pkt.ipi6_addr;
                retv = 0;
                break;
        }
```
So it seems I should have different paths for ipv6 and ipv4 here -> will try to do.

Snorch: It seems that IPV6_PKTINFO is a completely different thing, systemd uses IPV6_RECVPKTINFO for ipv6 similar to IP_PKTINFO.

Snorch: done.

criu/sk-inet.c
```diff
@@ -787,8 +790,13 @@ int restore_ip_opts(int sk, int family, int proto, IpOptsEntry *ioe)
...
 
-	if (ioe->has_freebind)
-		ret |= restore_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+	if (family == AF_INET6) {
+		if (ioe->has_freebind)
+			ret |= restore_opt(sk, SOL_IPV6, IPV6_FREEBIND, &ioe->freebind);
+	} else {
+		if (ioe->has_freebind)
+			ret |= restore_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
+	}
```
rst0git, line 806: Perhaps this condition could be simplified to:
```C
	if (ioe->has_freebind) {
		if (family == AF_INET6)
			ret |= restore_opt(sk, SOL_IPV6, IPV6_FREEBIND, &ioe->freebind);
		else
			ret |= restore_opt(sk, SOL_IP, IP_FREEBIND, &ioe->freebind);
	}
```

Snorch:
Ci found:
> criu/sk-inet.c:404:33: error: ‘IPV6_FREEBIND’ undeclared (first use in this function)

Adding proper includes for IP_FREEBIND/IPV6_FREEBIND to fix it - linux/in.h / linux/in6.h.

avagin:
test-local-gcc fails with this error:

```
=================== Run zdtm/static/socket-tcp6-local in uns ===================
Start test
./socket-tcp6-local --pidfile=socket-tcp6-local.pid --outfile=socket-tcp6-local.out
Run criu dump
Wait for zdtm/static/socket-tcp6-local(175) to die for 0.100000
Run criu restore
Send the 15 signal to  203
Wait for zdtm/static/socket-tcp6-local(203) to die for 0.100000
########### Test zdtm/static/socket-tcp6-local FAIL at result check ############
09:08:12.353:     6: ERR: tcp.c:133: can't connect to server (errno = 106 (Transport endpoint is already connected))
09:08:12.354:     5: ERR: socket-tcp6-local.c:201: read less then have to (errno = 11 (Resource temporarily unavailable))

```

Snorch:
> test-local-gcc fails with this error:
> 
> ```
> =================== Run zdtm/static/socket-tcp6-local in uns ===================
> Start test
> ./socket-tcp6-local --pidfile=socket-tcp6-local.pid --outfile=socket-tcp6-local.out
> Run criu dump
> Wait for zdtm/static/socket-tcp6-local(175) to die for 0.100000
> Run criu restore
> Send the 15 signal to  203
> Wait for zdtm/static/socket-tcp6-local(203) to die for 0.100000
> ########### Test zdtm/static/socket-tcp6-local FAIL at result check ############
> Test output: ================================
> 09:08:12.353:     6: ERR: tcp.c:133: can't connect to server (errno = 106 (Transport endpoint is already connected))
> 09:08:12.354:     5: ERR: socket-tcp6-local.c:201: read less then have to (errno = 11 (Resource temporarily unavailable))
> 
>  <<< ================================
> ```

1) I think the test is racy because on my previous push (before lint fix) all tests passed.
2) I'm not sure I understand how uns flavour test can get "someone else has connected to the server" error, I see that "Run zdtm/static/socket-tcp-syn-sent in h" is just after ours and it uses same port 8880. Not sure at all how setsockopt changes can afftect this test.
3) I believe it's an old known problem https://github.com/checkpoint-restore/criu/issues/1079

rst0git (approved):
LGTM

Snorch:
Can we merge it?

test/zdtm/static/sock_ip_opts00.c
```diff
@@ -0,0 +1,104 @@
...
+		n_opts = sk_confs[i].domain == AF_INET ? ARRAY_SIZE(sk_opts_v4) : ARRAY_SIZE(sk_opts_v6);
+
+		for (j = 0; j < n_opts; j++) {
+			len = sizeof(int);
+			if (getsockopt(sk_confs[i].sk, opts[j].level, opts[j].opt, &val, &len) == -1) {
+				pr_perror("getsockopt(%d, %d) failed", opts[j].level, opts[j].opt);
+				goto close;
+			}
+
+			if (val != 1) {
```
rst0git, line 90: @Snorch Do we need to check if `val = 0` is also restored correctly?

Snorch: Didn't thought about it, thanks, I will add a second variant of this test for 0.

rst0git (approved):
LGTM

PR #2009, opened 2022-12-04
crit: fix compatibility with Python 3.12

rst0git:
Python 3.12 includes a few breaking changes, such as the removal of the [`distutils` module](https://www.python.org/downloads/release/python-3120a2/) and the deprecation of [`setup.py install`](https://github.com/pypa/setuptools/pull/2824) in favour of `pip install`. This pull request updates the installation script for `crit` to reflect these changes. This update replaces the use of `setup.py install` with  `pip install` and replaces `distutils` with `setuptools`.


adrianreber:
Can you try to build RPMs and DEBs with this change. As this changes the installation of CRIU it would be good to know that package 📦 builds are still working.

adrianreber:
Did you verify that you can still RPMs and DEBs with this change. Especially for the older distributions that CRIU supports.

Maybe a rebase to remove a couple of fixed CI errors.

rst0git:
@adrianreber The RPM and DEB build successfully with this change.

It is worth noting that we are switching from the [egg packaging format  to wheel](https://packaging.python.org/en/latest/discussions/wheel-vs-egg/). Thus, we would need to update the RPM spec file to use `%{python3_sitelib}/*dist-info` instead of `%{python3_sitelib}/*egg-info`. In addition, we need to add `python3-pip` as a build dependency.

adrianreber (approved):
Looks good. Thanks for verifying that it still work.

PR #2012, opened 2022-12-05
Try to fix zdtm socket test 8880 port conflict

Snorch:
This may make https://github.com/checkpoint-restore/criu/issues/1079 race less noisy.


adrianreber (approved):
Sounds like a good idea to have different ports for the tests as they can run in parallel.

rst0git (approved):

avagin:
Don't we run tests in separate network namespaces?

Snorch:
> Don't we run tests in separate network namespaces?

1) not all of them
```
test/zdtm/static/net_lock_socket_iptables.desc:    'flavor': 'h',
test/zdtm/static/netns_lock_iptables.desc:    'flavor': 'h',
test/zdtm/static/socket-tcpbuf.desc:{'flavor': 'h', 'opts': '--tcp-established', 'flags': 'nouser samens'}
test/zdtm/static/socket-tcp.desc:{'flavor': 'h', 'opts': '--tcp-established', 'flags': 'nouser samens'}
test/zdtm/static/socket-tcp-reuseport.desc:{'flavor': 'h ns uns', 'opts': '--tcp-established', 'flags': 'nouser samens'}
```
2) some test has network logic before test_init() like socket-tcp6.c.

avagin:
@Snorch We execute each test in our zdtm "container":
https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm_ct.c#L108

It creates a separate net namespace. Did I miss something?

Snorch:
> @Snorch We execute each test in our zdtm "container":
> https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm_ct.c#L108
> 
> It creates a separate net namespace. Did I miss something?

You are right, thanks for pointing that out, I had wrong impression that we run several tests in one zdtm_ct. Then we can close this pr as it would not help for sure.

PR #2013, opened 2022-12-05
test: Export ARCH variable

VermaSh:
Export `ARCH` variable so that it can be accessed by sub-make. For
example, it is needed to turn on platform specific tests in
`./test/zdtm/static/Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ endif
 ifneq ($(filter-out x86 arm aarch64 ppc64 s390 mips,$(ARCH)),)
         $(error "The architecture $(ARCH) isn't supported")
 endif
+export ARCH
 
 # The PowerPC 64 bits architecture could be big or little endian.
 # They are handled in the same way.


VermaSh:
@rst0git Can I please get a review?

Makefile
```diff
@@ -22,6 +22,7 @@ endif
 ifneq ($(filter-out x86 arm aarch64 ppc64 s390 mips,$(ARCH)),)
         $(error "The architecture $(ARCH) isn't supported")
 endif
+export ARCH
```
rst0git, line 25: `ARCH` is usually explicitly set when used for cross compilation:
- https://github.com/checkpoint-restore/criu/blob/criu-dev/Documentation/HOWTO.cross-compile
- https://criu.org/ARM_crosscompile
- https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/build/Dockerfile.armv7-stable-cross.hdr

I'm not sure if adding this `export` is necessary.

VermaSh: I added `export` because I was seeing for `make: *** No rule to make target` for `s390x_regs_check`, `s390x_gs_threads`, and `s390x_runtime_instr` while running on zLinux. `ARCH` in `./test/zdtm/static/Makefile` seemed to be empty. But weirdly enough, the test is running now without this change. So let me do couple more test iterations before we close this PR.

VermaSh:
Closing this as the change isn't needed

PR #2014, opened 2022-12-08
dump and restore cpu affinity of each thread.

hdzhoujie:
Add one entry of thread_cpuallow_entry into thread_core_entry to save cpu affinity info. Restore it after threads restored but before running. Add option --with-cpu-affinity to enable this function at restore.
Signed-off-by: hdzhoujie <zhoujie133@huawei.com>
Signed-off-by: He jingxian <hejingxian@huawei.com>
Signed-off-by: Sang Yan <sangyan@huawei.com>


criu/pstree.c
```diff
@@ -543,6 +554,14 @@ static int read_one_pstree_item(struct cr_img *img, pid_t *pid_max)
 		goto err;
 	BUG_ON(pi->pid->state != TASK_UNDEF);
 
+	if (opts.with_cpu_affinity) {
+		nr_cpus = sysconf(_SC_NPROCESSORS_CONF);
+		if (e->nr_cpus > nr_cpus) {
+			pr_err("cpu num err\n");
```
0x7f454c46, line 560: The error message could be more elaborative for the end user.

rst0git:
@hdzhoujie Would it be possible to add tests to verify that the CPU affinity is restored correctly?

PR #2017, opened 2022-12-09
ci: switch from lgtm to codeql

adrianreber:
This is a replacement for #2015 which does not work correctly. This includes a couple of small changes to make it work.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,45 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "criu-dev", "master" ]
+  pull_request:
+    branches: [ "criu-dev" ]
+  schedule:
+    - cron: "11 6 * * 3"
+
+jobs:
+  analyze:
+    name: Analyze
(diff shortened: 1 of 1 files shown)


rst0git (approved):

rst0git:
> GitHub Code Scanning / CodeQL
305 new alerts including 19 critical severity security vulnerabilities
https://github.com/checkpoint-restore/criu/pull/2017/checks?check_run_id=9995693716

@adrianreber Should we address these alerts before merging the pull request?

adrianreber:
> > GitHub Code Scanning / CodeQL
> 305 new alerts including 19 critical severity security vulnerabilities
> https://github.com/checkpoint-restore/criu/pull/2017/checks?check_run_id=9995693716
> 
> @adrianreber Should we address these alerts before merging the pull request? 

I would say that it is not necessary. Non of the possible problems are new. Some are false positives from what I looked at it and fixing them all here would make this PR huge.

I am not entirely sure, but I expect that only newly introduced problems will result in a CI error. So it would help us avoid new problems and the existing reports can be fixed over time or marked as false positives.

PR #2020, opened 2022-12-12
test/javaTests: update org.testng:testng (Maven)

avagin:
TestNG is vulnerable to Path Traversal

Fixes https://github.com/checkpoint-restore/criu/security/dependabot/1.

### DIFF
--- a/scripts/build/Dockerfile.openj9-ubuntu
+++ b/scripts/build/Dockerfile.openj9-ubuntu
@@ -1,4 +1,4 @@
-FROM docker.io/library/ibm-semeru-runtimes:open-8-jdk-focal
+FROM docker.io/library/ibm-semeru-runtimes:open-11-jdk-focal
 ARG CC=gcc
 
 COPY scripts/ci/apt-install /bin/apt-install
--- a/test/javaTests/pom.xml
+++ b/test/javaTests/pom.xml
@@ -38,7 +38,7 @@
 		<dependency>
 			<groupId>org.testng</groupId>
 			<artifactId>testng</artifactId>
-			<version>6.3.1</version>
+			<version>7.7.0</version>
 		</dependency>
 	</dependencies>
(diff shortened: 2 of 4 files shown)


rst0git:
> Fixes https://github.com/checkpoint-restore/criu/security/dependabot/1

btw, this link is not publicly accessible.

scripts/build/Dockerfile.openj9-ubuntu
```diff
@@ -1,4 +1,4 @@
-FROM docker.io/library/ibm-semeru-runtimes:open-8-jdk-focal
+FROM docker.io/library/ibm-semeru-runtimes:open-11-jdk-focal
```
rst0git, line 1: We should update [Dockerfile.hotspot-alpine](https://github.com/checkpoint-restore/criu/blob/6e11e7f081828edb1bad7e83ea566cf197c2e514/scripts/build/Dockerfile.hotspot-alpine#L1) and [Dockerfile.hotspot-ubuntu](https://github.com/checkpoint-restore/criu/blob/6e11e7f081828edb1bad7e83ea566cf197c2e514/scripts/build/Dockerfile.hotspot-ubuntu#L1) as well.

test/javaTests/pom.xml
```diff
@@ -38,7 +38,7 @@
 		<dependency>
 			<groupId>org.testng</groupId>
 			<artifactId>testng</artifactId>
-			<version>6.3.1</version>
+			<version>7.6.1</version>
```
rst0git, line 41: It looks like version [7.6.1](https://mvnrepository.com/artifact/org.testng/testng/7.6.1) also has some vulnerabilities. Should we update to version [7.7.0](https://mvnrepository.com/artifact/org.testng/testng/7.7.0) instead?

rst0git (approved):

PR #2022, opened 2022-12-12
Fix VMA handling for shmem maps

### LINKED ISSUE #2019: Changes that are required to C/R httpd in podman
You certainly know what you are doing more than me, but perhaps it might be of interest to you to know what I needed to patch to get CRIU of httpd in podman working:

<details>
<summary>Patch</summary>

```diff
diff --git a/criu/proc_parse.c b/criu/proc_parse.c
index 946b0fc40..a24783da7 100644
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -313,25 +313,8 @@ static int vma_get_mapfile_user(const char *fname, struct vma_area *vma, struct
 
 	vfi_dev = makedev(vfi->dev_maj, vfi->dev_min);
 
-	if (is_memfd(vfi_dev)) {
-		char tmp[PATH_MAX];
-		strlcpy(tmp, fname, PATH_MAX);
-		strip_deleted(tmp, strlen(tmp));
-
-		/*
-		 * The error EPERM will be shown in the following pr_perror().
-		 * It comes from the previous open() call.
-		 */
-		pr_perror("Can't open mapped [%s]", tmp);
-
-		/*
-		 * TODO Perhaps we could do better than failing and dump the
-		 * memory like what is being done in shmem.c
-		 */
-		return -1;
-	}
-
 	if (is_hugetlb_dev(vfi_dev, &hugetlb_flag) || is_anon_shmem_map(vfi_dev)) {
+		vma->e->status |= VMA_AREA_REGULAR;
 		if (!(vma->e->flags & MAP_SHARED))
 			vma->e->status |= VMA_ANON_PRIVATE;
 		else
diff --git a/criu/shmem.c b/criu/shmem.c
index 81e701586..2de2ea9af 100644
--- a/criu/shmem.c
+++ b/criu/shmem.c
@@ -724,7 +724,7 @@ static int next_data_segment(int fd, unsigned long pfn, unsigned long *next_data
 	return 0;
 }
 
-static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si)
+static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si, bool seek_data_supported)
 {
 	struct page_pipe *pp;
 	struct page_xfer xfer;
@@ -750,7 +750,7 @@ static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si)
 		unsigned long pgaddr;
 		int st = -1;
 
-		if (pfn >= next_hole_pfn && next_data_segment(fd, pfn, &next_data_pnf, &next_hole_pfn))
+		if (seek_data_supported && pfn >= next_hole_pfn && next_data_segment(fd, pfn, &next_data_pnf, &next_hole_pfn))
 			goto err_xfer;
 
 		if (si->pstate_map && is_shmem_tracking_en()) {
@@ -808,20 +808,50 @@ static int dump_one_shmem(struct shmem_info *si)
 {
 	int fd, ret = -1;
 	void *addr;
+	bool seek_data_supported;
 
 	pr_info("Dumping shared memory %ld\n", si->shmid);
 
-	fd = open_proc(si->pid, "map_files/%lx-%lx", si->start, si->end);
-	if (fd < 0)
-		goto err;
 
-	addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
-	if (addr == MAP_FAILED) {
-		pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
-		goto errc;
+	fd = __open_proc(si->pid, EPERM, O_RDONLY, "map_files/%lx-%lx", si->start, si->end);
+	if (fd >= 0) {
+		addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
+		if (addr == MAP_FAILED) {
+			pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			goto errc;
+		}
+
+		seek_data_supported = true;
+	} else {
+		if(errno != EPERM) {
+			goto err;
+		}
+
+		fd = open_proc(si->pid, "mem");
+		if(fd < 0) {
+			goto err;
+		}
+
+		addr = mmap(NULL, si->size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+		if (addr == MAP_FAILED) {
+			pr_err("Can't map empty space for shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			goto errc;
+		}
+
+		if(lseek(fd, si->start, SEEK_SET) < 0) {
+			pr_perror("Can't seek virtual memory");
+			return -1;
+		}
+
+		if(read(fd, addr, si->size) < si->size) {
+			pr_perror("Can't read virtual memory");
+			return -1;
+		}
+
+		seek_data_supported = false;
 	}
 
-	ret = do_dump_one_shmem(fd, addr, si);
+	ret = do_dump_one_shmem(fd, addr, si, seek_data_supported);
 
 	munmap(addr, si->size);
 errc:
@@ -849,7 +879,7 @@ int dump_one_memfd_shmem(int fd, unsigned long shmid, unsigned long size)
 		goto err;
 	}
 
-	ret = do_dump_one_shmem(fd, addr, &si);
+	ret = do_dump_one_shmem(fd, addr, &si, true);
 
 	munmap(addr, size);
 err:
@@ -875,7 +905,7 @@ int dump_one_sysv_shmem(void *addr, unsigned long size, unsigned long shmid)
 	if (fd < 0)
 		return -1;
 
-	ret = do_dump_one_shmem(fd, addr, si);
+	ret = do_dump_one_shmem(fd, addr, si, true);
 	close(fd);
 	return ret;
 }
diff --git a/criu/sockets.c b/criu/sockets.c
index db772707b..d50866c65 100644
--- a/criu/sockets.c
+++ b/criu/sockets.c
@@ -12,6 +12,7 @@
 
 #include "int.h"
 #include "bitops.h"
+#include "cr_options.h"
 #include "libnetlink.h"
 #include "sockets.h"
 #include "unix_diag.h"
@@ -457,6 +458,27 @@ int sk_collect_one(unsigned ino, int family, struct socket_desc *d, struct ns_id
 
 int do_restore_opt(int sk, int level, int name, void *val, int len)
 {
+	void *buf;
+	socklen_t cur_len;
+	bool match;
+
+	if(opts.unprivileged) {
+		buf = xmalloc(len);
+		if (!buf) {
+			return -1;
+		}
+		cur_len = len;
+		if (getsockopt(sk, level, name, buf, &cur_len) < 0) {
+			pr_perror("Can't get %d:%d (len %d)", level, name, len);
+			return -1;
+		}
+		match = cur_len == len && memcmp(buf, val, len) == 0;
+		xfree(buf);
+		if(match) {
+			return 0;
+		}
+	}
+
 	if (setsockopt(sk, level, name, val, len) < 0) {
 		pr_perror("Can't set %d:%d (len %d)", level, name, len);
 		return -1;
@@ -469,9 +491,20 @@ static int sk_setbufs(void *arg, int fd, pid_t pid)
 {
 	u32 *buf = (u32 *)arg;
 
-	if (restore_opt(fd, SOL_SOCKET, SO_SNDBUFFORCE, &buf[0]))
+	int snd_opt_name;
+	int rcv_opt_name;
+
+	if (!opts.unprivileged) {
+		snd_opt_name = SO_SNDBUFFORCE;
+		rcv_opt_name = SO_RCVBUFFORCE;
+	} else {
+		snd_opt_name = SO_SNDBUF;
+		rcv_opt_name = SO_RCVBUF;
+	}
+
+	if (restore_opt(fd, SOL_SOCKET, snd_opt_name, &buf[0]))
 		return -1;
-	if (restore_opt(fd, SOL_SOCKET, SO_RCVBUFFORCE, &buf[1]))
+	if (restore_opt(fd, SOL_SOCKET, rcv_opt_name, &buf[1]))
 		return -1;
 
 	return 0;
```

</details>

The omission of `vma->e->status |= VMA_AREA_REGULAR;` seems to be a somewhat unrelated bug that should probably have been noticed earlier. Using lseek + read + vmsplice is inefficient, I guess `process_vm_readv` should be used instead, but I just wanted to get it working first. I also updated `do_restore_opt` to not set options if they are already correct because we might not have enough permissions for that, and `SO_RCVBUF`/`SO_SNDBUF` are used instead of their *FORCE counterparts in unprivileged mode. Maybe some of this is technically wrong, please take a better look (e.g. perhaps inaccessible `map_files` should be handled similarly in `dump_one_sysv_shmem` too).

I tested this inside

```
podman run -it --cap-add=cap_checkpoint_restore --cap-add=cap_sys_ptrace --cap-add=cap_sys_resource --security-opt seccomp=unconfined --security-opt unmask=/proc/sys php:8.1-apache bash
```

CAP_CHECKPOINT_RESTORE is needed for obvious reasons, CAP_SYS_PTRACE is for `process_vm_readv` IIRC, CAP_SYS_RESOURCE is for `rlimit`, seccomp is so that `vmsplice` does not yield EPERM (I don't quite understand the reason for this one tbh), `/proc/sys` is unmasked for `ns_next_pid`. CAP_NET_ADMIN is *not* used due to a bug described in one of my review comments on this PR.

Hope this helps someone.

_Originally posted by @imachug in https://github.com/checkpoint-restore/criu/issues/1930#issuecomment-1294356320_

ymanton:
Patch originally written by @imachug to get httpd restored inside a container in unprivileged mode.

Fixes #2019.

### DIFF
--- a/criu/shmem.c
+++ b/criu/shmem.c
@@ -724,7 +724,7 @@ static int next_data_segment(int fd, unsigned long pfn, unsigned long *next_data
 	return 0;
 }
 
-static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si, bool seek_data_supported)
+static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si)
 {
 	struct page_pipe *pp;
 	struct page_xfer xfer;
@@ -750,8 +750,7 @@ static int do_dump_one_shmem(int fd, void *addr, struct shmem_info *si, bool see
 		unsigned long pgaddr;
 		int st = -1;
 
-		if (seek_data_supported && pfn >= next_hole_pfn &&
-		    next_data_segment(fd, pfn, &next_data_pnf, &next_hole_pfn))
+		if (fd >= 0 && pfn >= next_hole_pfn && next_data_segment(fd, pfn, &next_data_pnf, &next_hole_pfn))
 			goto err_xfer;
 
 		if (si->pstate_map && is_shmem_tracking_en()) {
@@ -810,7 +809,6 @@ static int dump_one_shmem(struct shmem_info *si)
 	int fd, ret = -1;
 	void *addr;
 	unsigned long cur, remaining;
-	bool seek_data_supported;
 
 	pr_info("Dumping shared memory %ld\n", si->shmid);
 
@@ -821,8 +819,6 @@ static int dump_one_shmem(struct shmem_info *si)
 			pr_perror("Can't map shmem 0x%lx (0x%lx-0x%lx)", si->shmid, si->start, si->end);
 			goto errc;
 		}
-
-		seek_data_supported = true;
 	} else {
 		if (errno != EPERM || !opts.unprivileged) {
 			goto err;
@@ -858,14 +854,16 @@ static int dump_one_shmem(struct shmem_info *si)
 			cur += ret;
 		} while (remaining > 0);
 
-		seek_data_supported = false;
+		close(fd);
+		fd = -1;
 	}
 
-	ret = do_dump_one_shmem(fd, addr, si, seek_data_supported);
+	ret = do_dump_one_shmem(fd, addr, si);
 
 	munmap(addr, si->size);
 errc:
-	close(fd);
+	if (fd >= 0)
+		close(fd);
 err:
 	return ret;
 }
@@ -889,7 +887,7 @@ int dump_one_memfd_shmem(int fd, unsigned long shmid, unsigned long size)
 		goto err;
 	}
 
-	ret = do_dump_one_shmem(fd, addr, &si, true);
+	ret = do_dump_one_shmem(fd, addr, &si);
 
 	munmap(addr, size);
 err:
@@ -915,7 +913,7 @@ int dump_one_sysv_shmem(void *addr, unsigned long size, unsigned long shmid)
 	if (fd < 0)
 		return -1;
 
-	ret = do_dump_one_shmem(fd, addr, si, true);
+	ret = do_dump_one_shmem(fd, addr, si);
 	close(fd);
 	return ret;
 }
--- a/criu/shmem.c
+++ b/criu/shmem.c
@@ -818,7 +818,7 @@ static int dump_one_shmem(struct shmem_info *si)
 	if (fd >= 0) {
 		addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
 		if (addr == MAP_FAILED) {
-			pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			pr_perror("Can't map shmem 0x%lx (0x%lx-0x%lx)", si->shmid, si->start, si->end);
 			goto errc;
 		}
 
@@ -837,7 +837,7 @@ static int dump_one_shmem(struct shmem_info *si)
 
 		addr = mmap(NULL, si->size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
 		if (addr == MAP_FAILED) {
-			pr_err("Can't map empty space for shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			pr_perror("Can't map empty space for shmem 0x%lx (0x%lx-0x%lx)", si->shmid, si->start, si->end);
 			goto errc;
 		}
 
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -289,6 +289,21 @@ ip net add test
 ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/transition/fork -t zdtm/static/ghost_holes00 -t zdtm/static/socket-tcp -t zdtm/static/msgque -k always
 ./test/crit-recode.py
 
+# Rootless tests
+# Check if cap_checkpoint_restore is supported and also if unshare -c is supported.
+if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true; then
+	make -C test/zdtm/ cleanout
+	rm -rf test/dump
+	setcap cap_checkpoint_restore,cap_sys_ptrace+eip criu/criu
+	# Run it as non-root in a user namespace. Since CAP_CHECKPOINT_RESTORE behaves differently in non-user namespaces (e.g. no access to map_files) this tests that we can dump and restore
+	# under those conditions. Note that the "... && true" part is necessary; we need at least one statement after the tests so that bash can reap zombies in the user namespace,
+	# otherwise it will exec the last statement and get replaced and nobody will be left to reap our zombies.
+	sudo --user=#65534 --group=#65534 unshare -Ucfpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true"
+	setcap -r criu/criu
+else
+	echo "Skipping unprivileged mode tests"
+fi
+
 # more crit testing
 make -C test/others/crit run
 
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -338,22 +338,49 @@ static int vma_get_mapfile_user(const char *fname, struct vma_area *vma, struct 
 	fd = open(fname, O_RDONLY);
 	if (fd < 0) {
 		pr_perror("Can't open mapped [%s]", fname);
-		return -1;
+		goto returnerr;
 	}
 
 	if (vma_stat(vma, fd)) {
-		close(fd);
-		return -1;
+		goto closefd;
 	}
 
-	if (vma->vmst->st_dev != vfi_dev || vma->vmst->st_ino != vfi->ino) {
-		pr_err("Failed to resolve mapping %lx filename\n", (unsigned long)vma->e->start);
-		close(fd);
-		return -1;
+	if (vma->vmst->st_ino != vfi->ino) {
+		goto errmsg;
+	}
+
+	/*
+	 * If devices don't match it could be because file is on a btrfs subvolume,
+	 * which means that device number returned by stat will not match what is
+	 * seen in smaps and other places. To deal with that we need a more involved
+	 * check.
+	 */
+	if (vma->vmst->st_dev != vfi_dev) {
+		int mnt_id;
+		struct ns_id *ns;
+
+		if (get_fd_mntid(fd, &mnt_id))
+			goto errmsg;
+
+		ns = lookup_nsid_by_mnt_id(mnt_id);
+		if (!ns)
+			goto errmsg;
+
+		if (!phys_stat_dev_match(vma->vmst->st_dev, vfi_dev, ns, fname))
+			goto errmsg;
+
+		vma->mnt_id = mnt_id;
 	}
 
 	*vm_file_fd = fd;
 	return 0;
+
+errmsg:
+	pr_err("Failed to resolve mapping %lx filename\n", (unsigned long)vma->e->start);
+closefd:
+	close(fd);
+returnerr:
+	return -1;
 }
 
 static int vma_get_mapfile(const char *fname, struct vma_area *vma, DIR *mfd, struct vma_file_info *vfi,
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -21,7 +21,9 @@ RUN apk update && apk add \
 	py3-pip \
 	py3-protobuf \
 	python3 \
-	sudo
+	sudo \
+	libcap-utils \
+	util-linux
 
 COPY . /criu
 WORKDIR /criu
--- a/scripts/ci/vagrant.sh
+++ b/scripts/ci/vagrant.sh
@@ -78,6 +78,11 @@ fedora-non-root() {
 	ssh default 'cd /vagrant/criu; criu/criu check --unprivileged; ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/static/pthread00 -f h --rootless'
 	# Run it as root with '--rootless'
 	ssh default 'cd /vagrant/criu; sudo ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/static/pthread00 -f h; sudo chmod 777 test/dump/zdtm/static/{env00,pthread00}; sudo ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/static/pthread00 -f h --rootless'
+	# Run it as non-root in a user namespace. Since CAP_CHECKPOINT_RESTORE behaves differently in non-user namespaces (e.g. no access to map_files) this tests that we can dump and restore
+	# under those conditions. Note that the "... && true" part is necessary; we need at least one statement after the tests so that bash can reap zombies in the user namespace,
+	# otherwise it will exec the last statement and get replaced and nobody will be left to reap our zombies.
+	# Note: selinux in Enforcing mode prevents us from calling clone3() or writing to ns_last_pid on restore; hence set to Permissive for the test and then set back.
+	ssh default 'cd /vagrant/criu; selinuxmode=`getenforce` && sudo setenforce Permissive && unshare -Ucfpm --mount-proc bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true" && sudo setenforce $selinuxmode'
 }
 
 $1


ymanton:
@imachug, can you take a look at these patches and confirm they work as expected on top of current criu-dev? Are you OK with the co-authored-by tag as it is?

Also, can you suggest how test cases might be written?

imachug:
Hmm, I don't think this handles the case I described [here](https://github.com/checkpoint-restore/criu/issues/2019#issuecomment-1346130299)? I'm a bit unsure if it's really a problem because I was thinking about the problem from a theoretical point of view, but do my concerns about this seem reasonable to you?

ymanton:
> Hmm, I don't think this handles the case I described [here](https://github.com/checkpoint-restore/criu/issues/2019#issuecomment-1346130299)? I'm a bit unsure if it's really a problem because I was thinking about the problem from a theoretical point of view, but do my concerns about this seem reasonable to you?

Sorry, we may be misunderstanding each other. I meant does current criu-dev plus the patches here allow you to restore a containerized httpd via Podman as you mentioned in #1930? The problem of how to handle deleted files with only partial mappings across one or more processes can be discussed in #2019, but since it's currently a theoretical problem and there are no patches I don't think it should hold up the next CRIU release.

ymanton:
> Yes, it works. I'd rather you added a note that `--unprivileged` behavior wrt. shared memory is known to be broken, and if something seems to work, it might only be by chance or involve UB.

I've added the following text to the man-page in the unprivileged mode section:

```
Note that for some operations, having a capability in a namespace other than
the init namespace (i.e. the default/root namespace) is not sufficient. For
example, in order to read symlinks in proc/[pid]/map_files CRIU requires
CAP_CHECKPOINT_RESTORE in the init namespace; having CAP_CHECKPOINT_RESTORE
while running in another user namespace (e.g. in a container) does not allow
CRIU to read symlinks in /proc/[pid]/map_files.

Without access to /proc/[pid]/map_files checkpointing/restoring processes
that have mapped deleted files may not be possible.
```

imachug:
Then maybe let's add a small check to fail the dump if the mmap'ed offset in map_files is not 0?

criu/shmem.c
```diff
@@ -808,20 +809,49 @@ static int dump_one_shmem(struct shmem_info *si)
...
+			pr_err("Can't map empty space for shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
+			goto errc;
+		}
+
+		if (lseek(fd, si->start, SEEK_SET) < 0) {
+			pr_perror("Can't seek virtual memory");
+			return -1;
+		}
+
+		if (read(fd, addr, si->size) < si->size) {
```
avagin, line 846: read can transfer at most MAX_RW_COUNT bytes

criu/shmem.c
```diff
@@ -808,20 +809,49 @@ static int dump_one_shmem(struct shmem_info *si)
...
+			goto errc;
+		}
 
-	addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
-	if (addr == MAP_FAILED) {
-		pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
-		goto errc;
+		seek_data_supported = true;
+	} else {
+		if (errno != EPERM) {
```
avagin, line 826: it looks like the patch description isn't complete. It has to describe of all aspects and reasons.

avagin:
I think our CI has to run zdtm.py from a sub-user-namespace. You can try to do something like this:

```
$ unshare -Ucfp python  test/zdtm.py run --rootless -f h -t zdtm/static/maps00
```

criu/shmem.c
ymanton: Wrapped in a loop now.

criu/shmem.c
ymanton: Thanks, I wasn't thinking clearly about what the patch was trying to do. Changed to the following:

```
shmem: Fall back to /proc/$pid/mem if no map_files 

If trying to open /proc/$pid/map_files/x-x for a given VMA fails with
EPERM (can happen in unprivileged mode when running in a non-init user
ns), fall back to reading the content from /proc/$pid/mem.
```

ymanton:
> I think our CI has to run zdtm.py from a sub-user-namespace. You can try to do something like this:
> 
> ```
> $ unshare -Ucfp python  test/zdtm.py run --rootless -f h -t zdtm/static/maps00
> ```

Thanks. Needs `--mount-proc` so that ZDTM can see the right PIDs, but I had to run an interactive shell via `unshare ... bash` and then run the test. Running it directly as above or even `unshare ... bash -c "..."` leaves the test as a zombie after it's been dumped. Output looks like:

```
...
+ 368:     return flag in tdesc.get('flags', '').split()
+2399:         for t in torun:
+2483:         launcher.finish()
+2194:         self.__wait_all()
+2177:         while self.__subs:
+2178:             self.__wait_one(0)
+2111:         pid = -1
+2112:         status = -1
+2113:         signal.alarm(10)
+2114:         while True:
+2115:             try:
+2116:                 pid, status = os.waitpid(0, flags)
Start test
Run criu dump
Wait for zdtm/static/maps00(56) to die for 0.100000
Wait for zdtm/static/maps00(56) to die for 0.200000
Wait for zdtm/static/maps00(56) to die for 0.400000
Wait for zdtm/static/maps00(56) to die for 0.800000
Wait for zdtm/static/maps00(56) to die for 1.600000
Wait for zdtm/static/maps00(56) to die for 3.200000
Wait for zdtm/static/maps00(56) to die for 6.400000
+  48:     print("==== ALARM ====")
==== ALARM ====
Wait for zdtm/static/maps00(56) to die for 12.800000
Wait for zdtm/static/maps00(56) to die for 25.600000
    PID TTY          TIME CMD
     56 ?        00:00:00 maps00 <defunct>
      1 pts/3    S+     0:00 python test/zdtm.py --debug run --ignore-taint --rootless -f h -t zdtm/static/maps00
      4 pts/3    S+     0:00 ./zdtm_ct zdtm.py
      5 pts/3    S+     0:00  \_ python zdtm.py
      6 pts/3    S+     0:00      \_ python zdtm.py
     60 pts/3    R+     0:00          \_ ps axf 56
     56 ?        Zs     0:00 [maps00] <defunct>
############ Test zdtm/static/maps00 FAIL at zdtm/static/maps00 die ############
11:00:29.515:    56: map: ptr 0x7f2dcca5f000 flag        2 prot        0
11:00:29.515:    56: map: ptr 0x7f2dcca5d000 flag        1 prot        0
11:00:29.515:    56: map: ptr 0x7f2dcca5b000 flag       22 prot        0
11:00:29.516:    56: map: ptr 0x7f2dcca59000 flag       21 prot        0

...

  File "/home/ymanton/criu/test/zdtm.py", line 1932, in do_run_test
    cr(cr_api, t, opts)
  File "/home/ymanton/criu/test/zdtm.py", line 1603, in cr
    test.gone()
  File "/home/ymanton/criu/test/zdtm.py", line 597, in gone
    self.__wait_task_die()
  File "/home/ymanton/criu/test/zdtm.py", line 442, in __wait_task_die
    wait_pid_die(int(self.__pid), self.__name, self.__timeout)
  File "/home/ymanton/criu/test/zdtm.py", line 364, in wait_pid_die
    raise test_fail_exc("%s die" % who)
__main__.test_fail_exc: zdtm/static/maps00 die
...
```

Not clear to me why the dumped process is being cleaned up under an interactive shell but not otherwise... but at least it effectively tests the patches here. Fails as expected on criu-dev, passes with these patches.

criu/shmem.c
ymanton: Done, it will only fall back to `/proc/$pid/mem` in unprivileged mode and a debug message will be printed:

```
Could not access map_files/ link, falling back to /proc/$pid/mem
```

ymanton:
> Then maybe let's add a small check to fail the dump if the mmap'ed offset of a shared memory region is not 0? This would allow us to say "If it works, it works" instead of "It might or might not work and you'll never know which one it is"

@imachug, I'm not clear on this. A non-zero offset for anonymous shared mappings is not allowed; do you mean we should check this for shared mappings backed by deleted files? If so, the current code will not fall back to `/proc/$pid/mem`, it will fail as follows:

```
b'(00.008055) Failed to open map_files/7fe73250b000-7fe73250d000, try to go via [/home/ymanton/criu/test/zdtm/static/maps00.test-07 (deleted)] path'
b"(00.008080) Error (criu/proc_parse.c:340): Can't open mapped [/home/ymanton/criu/test/zdtm/static/maps00.test-07 (deleted)]: No such file or directory"
b"(00.008087) Error (criu/proc_parse.c:667): Can't open 642's mapfile link 7fe73250b000: No such file or directory"
b'(00.008097) Error (criu/cr-dump.c:1564): Collect mappings (pid: 642) failed with -1'
```

imachug:
Yes, that was what I deemed the problem. I think you're right. Sorry for confusion. Deleted mappings should perhaps be addressed in a separate PR, if possible. I'm a bit iffy on trying to open a path ending with ` (deleted)`, as that might lead to a bug if such file does exist, but that's a totally unrelated problem, and it might have been addressed already.

avagin:
I have merged this pr, so let's do fixes on top.

criu/shmem.c
```diff
@@ -808,20 +809,59 @@ static int dump_one_shmem(struct shmem_info *si)
...
+		pr_debug("Could not access map_files/ link, falling back to /proc/$pid/mem\n");
+
+		fd = open_proc(si->pid, "mem");
+		if (fd < 0) {
+			goto err;
+		}
+
+		addr = mmap(NULL, si->size, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, -1, 0);
+		if (addr == MAP_FAILED) {
+			pr_err("Can't map empty space for shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
```
avagin, line 840: nit: pr_perror

criu/shmem.c
```diff
@@ -808,20 +809,59 @@ static int dump_one_shmem(struct shmem_info *si)
...
 	pr_info("Dumping shared memory %ld\n", si->shmid);
 
-	fd = open_proc(si->pid, "map_files/%lx-%lx", si->start, si->end);
-	if (fd < 0)
-		goto err;
+	fd = __open_proc(si->pid, EPERM, O_RDONLY, "map_files/%lx-%lx", si->start, si->end);
+	if (fd >= 0) {
+		addr = mmap(NULL, si->size, PROT_READ, MAP_SHARED, fd, 0);
+		if (addr == MAP_FAILED) {
+			pr_err("Can't map shmem 0x%lx (0x%lx-0x%lx)\n", si->shmid, si->start, si->end);
```
avagin, line 821: nit: pr_perror

criu/shmem.c
```diff
@@ -808,20 +809,59 @@ static int dump_one_shmem(struct shmem_info *si)
...
+			ret = read(fd, addr + cur, remaining);
+			if (ret <= 0) {
+				pr_perror("Can't read virtual memory");
+				goto errc;
+			}
+			remaining -= ret;
+			cur += ret;
+		} while (remaining > 0);
+
+		seek_data_supported = false;
```
avagin, line 861: We can close fd here. I don't think it is a good idea to pass it to do_dump_one_shmem. I think we don't use it there, do we? And we set fd to -1 here, we can conside to eliminate the seek_data_supported argument.

avagin:
> Thanks. Needs --mount-proc so that ZDTM can see the right PIDs, but I had to run an interactive shell via unshare ... bash and then run the test. Running it directly as above or even unshare ... bash -c "..." leaves the test as a zombie after it's been dumped. 

@ymanton  I think the next command has to work:

```
unshare -Ucfpm --mount-proc  bash -c "python  test/zdtm.py run -t zdtm/static/env00 --ignore-taint --rootless -f h && echo PASSED"
```

The echo after zdtm doesn't allow bach to exec the last command and so it continues collecting zombies like init.

avagin:
```
+ ssh default 'cd /vagrant/criu; unshare -Ucfpm --mount-proc bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true"'
userns is supported
Start test
Run criu dump
=[log]=> dump/zdtm/static/maps00/70/1/dump.log
b'(00.003926)'
b'(00.003928) Collecting mappings (pid: 70)'
b'(00.003930) ----------------------------------------'
b'(00.003972) Failed to open map_files/400000-402000, try to go via [/vagrant/criu/test/zdtm/static/maps00] path'
b'(00.003980) Error (criu/proc_parse.c:350): Failed to resolve mapping 400000 filename'
b"(00.003987) Error (criu/proc_parse.c:667): Can't open 70's mapfile link 400000: Operation not permitted"
b'(00.003991) Error (criu/cr-dump.c:1564): Collect mappings (pid: 70) failed with -1'
b'(00.004031) Unlock network'
b'(00.004037) Unfreezing tasks into 1'
b'(00.004039) \tUnseizing 70 into 1'
b'(00.004059) Error (criu/cr-dump.c:2099): Dumping FAILED.'
################## Test zdtm/static/maps00 FAIL at CRIU dump ###################
```

avagin:
@ymanton could you try out this patch:

```
diff --git a/scripts/build/Dockerfile.alpine b/scripts/build/Dockerfile.alpine
index eced46c22..19b08315f 100644
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -21,7 +21,9 @@ RUN apk update && apk add \
        py3-pip \
        py3-protobuf \
        python3 \
-       sudo
+       sudo \
+       libcap-utils \
+       util-linux
 
 COPY . /criu
 WORKDIR /criu
diff --git a/scripts/ci/run-ci-tests.sh b/scripts/ci/run-ci-tests.sh
index 7b64c6b06..b00a30a72 100755
@@ -289,6 +291,13 @@ ip net add test
 ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/transition/fork -t zdtm/static/ghost_holes00 -t zdtm/static/socket-tcp -t zdtm/static/msgque -k always
 ./test/crit-recode.py
 
+# Rootless tests
+make -C test/zdtm/ cleanout
+rm -rf test/dump
+setcap cap_checkpoint_restore,cap_sys_ptrace+eip criu/criu
+sudo --user=#65534 --group=#65534 unshare -Ucfpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true"
+setcap -r criu/criu
+
 # more crit testing
 make -C test/others/crit run
 
```

ymanton:
Thanks @avagin, it works in some cases, but it fails the same way as my previous attempt on Fedora. It fails on Ubuntu 20.04 because libcap doesn't know about `cap_checkpoint_restore`; but I guess we can add a `capsh --supports=...` check to deal with that sort of thing.

I don't have easy access to Fedora, but I can get on a RHEL machine. I'll try to reproduce it there and see what's going on.

avagin:
>  It fails on Ubuntu 20.04 because libcap doesn't know about cap_checkpoint_restore;

https://github.com/checkpoint-restore/criu/pull/2061

ymanton:
Works on RHEL 9 as well. I got on a Fedora 37 VM and finally it fails like above. The error is from https://github.com/checkpoint-restore/criu/blob/0965eab83fb256121f19df47b980da5956ac077a/criu/proc_parse.c#L349-L353

When it fails it seems to be because the device number we get from `smaps` doesn't match what we get from `stat`.

For failing cases:

```
# stat
  File: /home/ymanton/criu/test/zdtm/static/maps00
  Size: 120856          Blocks: 240        IO Block: 4096   regular file
Device: 0,36    Inode: 5107        Links: 1

# smaps
00400000-00402000 r--p 00000000 00:1c 5107                               /home/ymanton/criu/test/zdtm/static/maps00
```

In working environments it looks like:

```
# stat
  File: /home/ymanton/criu/test/zdtm/static/maps00
  Size: 116224          Blocks: 232        IO Block: 4096   regular file
Device: fd00h/64768d    Inode: 135918877   Links: 1

# smaps
559a2d8d5000-559a2d8d7000 r--p 00000000 fd:00 135918877                  /home/ymanton/criu/test/zdtm/static/maps00
```

I do not know why this is, but comments near the top of `vma_get_mapfile_user` seem to say that we are hoping smaps will have the correct info but maybe not.

If anyone thinks they know what is going on I would appreciate some info, otherwise I'll see if I can find an explanation.

avagin:
@ymanton could you show /proc/self/mountinfo from this host? It can be a btrfs issue.

ymanton:
@avagin you're right, file is on btrfs.

```
[ymanton@localhost criu]$ cat /proc/self/mountinfo
...
58 1 0:28 /root / rw,relatime shared:1 - btrfs /dev/vda5 rw,seclabel,compress=zstd:1,space_cache=v2,subvolid=256,subvol=/root
...
89 58 0:28 /home /home rw,relatime shared:39 - btrfs /dev/vda5 rw,seclabel,compress=zstd:1,space_cache=v2,subvolid=257,subvol=/home
...
```

Looks like this is a known issue: https://criu.org/Filesystems_pecularities

Seems we should use something like `phys_stat_dev_match()` in the code above to compare the devices.

ymanton:
I think I've fixed the btrfs issue, and it looks like the test passes elsewhere but Fedora and Cent Stream still fail on restore; `clone3 -> EPERM`. I disabled that path and tried again but writing to `ns_last_pid` also returns `EPERM`.

criu/shmem.c
ymanton: Fixed.

criu/shmem.c
ymanton: Fixed.

criu/shmem.c
ymanton: Fixed, thanks.

ymanton:
> I think I've fixed the btrfs issue, and it looks like the test passes elsewhere but Fedora and Cent Stream still fail on restore; clone3 -> EPERM. I disabled that path and tried again but writing to ns_last_pid also returns EPERM.

@adrianreber, any idea about the above? It works as expected when I run in unprivileged mode without using unshare, but for some reason cap_checkpoint_restore inside a namespace doesn't work as expected, and only on Fedora/Cent Stream.

adrianreber:
Is it selinux related? Can you switch to permissive to test it?

ymanton:
Yeah that was it, works in Permissive mode, thanks.

adrianreber:
> Yeah that was it, works in Permissive mode, thanks.

You could report it here: https://github.com/fedora-selinux/selinux-policy

Ask to have the same access to clone3 and ns_last_pid in a user namespace.

ymanton:
Finally, all green.

scripts/ci/run-ci-tests.sh
```diff
@@ -289,6 +289,20 @@ ip net add test
 ./test/zdtm.py run -t zdtm/static/env00 -t zdtm/transition/fork -t zdtm/static/ghost_holes00 -t zdtm/static/socket-tcp -t zdtm/static/msgque -k always
 ./test/crit-recode.py
 
+# Rootless tests
+if capsh --supports=cap_checkpoint_restore; then
+	make -C test/zdtm/ cleanout
+	rm -rf test/dump
+	setcap cap_checkpoint_restore,cap_sys_ptrace+eip criu/criu
+	# Run it as non-root in a user namespace. Since CAP_CHECKPOINT_RESTORE behaves differently in non-user namespaces (e.g. no access to map_files) this tests that we can dump and restore
```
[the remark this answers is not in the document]
avagin: It is out of scope of this pr. And I am very skeptical about this especially after reading discussions around PR_HIDE_SELF_EXE. But it can be worth to send a patch and check the ground.

avagin:
Merged. Thanks a lot!

PR #2025, opened 2022-12-14
Support migration membarrier() PRIVATE_EXPEDITED registrations

osctobe:
Probe registration of membarrier() for MEMBARRIER_CMD_PRIVATE_EXPEDITED* calls on dump and restore the registrations.

This is a best-effort implementation that ignores unknown modes. The proper migration of membarrier() needs kernel support for querying the state (currently e.g. registration for MEMBARRIER_CMD_GLOBAL_EXPEDITED is not possible to detect).


adrianreber:
Please open the PR against the criu-dev branch.

mclapinski:
I think this could use a test.

Snorch (commented):
Thanks for a good PR! Please see some comments. And yes you need a zdtm test for it.

osctobe:
Rebased and fixed. A test is attached.

test/zdtm/static/membarrier.c
```diff
@@ -0,0 +1,105 @@
+#include <linux/membarrier.h>
+#include <sys/syscall.h>
+#include "zdtmtst.h"
+
+const char *const test_doc = "Test membarrier() migration";
```
rst0git, line 5: Could you also add this test in [test/zdtm/static/Makefile](https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm/static/Makefile)?

osctobe: Done.

criu/pie/parasite.c
```diff
@@ -211,6 +211,34 @@ static int dump_thread_common(struct parasite_dump_thread *ti)
...
 
+/*
+ * Returns a membarrier() registration command (it is a bitmask) if the process
+ * was registered for specified (as a bit index) membarrier()-issuing command;
+ * returns zero otherwise.
+ */
+static int get_membarrier_registration_mask(int cmd_bit)
+{
+	unsigned cmd = 1 << cmd_bit;
+	/* Issuing a barrier will be successful only if the process was registered
```
Snorch, line 222: `%s#^\(\s*\)/\* #\1/*\r\1 * #`
Please do everywhere.

criu/pie/parasite.c
```diff
@@ -211,6 +211,34 @@ static int dump_thread_common(struct parasite_dump_thread *ti)
...
+	 * for this type of membarrier. All errors are a sign that the type issued
+	 * was not registered (EPERM) or not supported by kernel (EINVAL or ENOSYS).
+	 */
+	int ret = sys_membarrier(cmd, 0, 0);
```
Snorch, line 226: nit, please separate declarations with a newline.
```
int ret;

/*
 * ....
 */
ret = sys_membarrier(cmd, 0, 0);
```

criu/pie/parasite.c
```diff
@@ -211,6 +211,34 @@ static int dump_thread_common(struct parasite_dump_thread *ti)
...
+	pr_debug("membarrier(1 << %d) returned %d\n", cmd_bit, ret);
+	/* For supported registrations, MEMBARRIER_CMD_REGISTER_xxx = MEMBARRIER_CMD_xxx << 1.
+	 * See: enum membarrier_cmd in include/uapi/linux/membarrier.h in kernel sources.
+	 */
+	return ret ? 0 : cmd << 1;
```
Snorch, line 231: Probably it's a good idea to check errno here, and fail on anything unexpected.

test/zdtm/static/membarrier.c
```diff
@@ -0,0 +1,105 @@
...
+		return 1;
+	}
+
+	barriers_registered = ret = 0;
+	for (int i = 0; i < n_membarrier_cmds; ++i) {
+		if (~barriers_supported & membarrier_cmds[i].register_cmd)
+			continue;
+		if (membarrier(membarrier_cmds[i].register_cmd, 0, 0) < 0) {
+			pr_perror("membarrier(REGISTER_PRIVATE_EXPEDITED%s)", membarrier_cmds[i].name_suffix);
+			ret = 1;
```
Snorch, line 74: Why can't we just fail immediately? I  mean switch `pr_perror` to `fail` and `ret = 1` to `return 1`.

osctobe: We can, but this is the same difference C compilers do (don't stop on first error) vs Pascal did (bail out on first failure). I prefer the former way as it minimizes the round-trips when debugging. Is this a hard requirement to follow the fail-early pattern?

Snorch: If you do some, let's say "trick", it must be meaningful - this way we improve code readability. If you don't strictly need something or don't know if you need something it is always better to avoid doing it, so that later reading the code we will not be confused that this "trick" is a must.

If something works in both variants we prefer the simple variant here in CRIU, at least me =) Because code is created to be easy to read.

If in future there would be a reason to debug this test in a manner that we need to try another membarrier if first membarrier call failed, e.g. if it would give us some new information for debugging, we can always improve the test, for now I don't see a clear reason.

criu/pie/parasite.c
```diff
@@ -211,6 +211,42 @@ static int dump_thread_common(struct parasite_dump_thread *ti)
...
+
+	/*
+	 * Issuing a barrier will be successful only if the process was registered
+	 * for this type of membarrier. All errors are a sign that the type issued
+	 * was not registered (EPERM) or not supported by kernel (EINVAL or ENOSYS).
+	 */
+	ret = sys_membarrier(cmd, 0, 0);
+	if (ret && ret != EPERM && ret != -EINVAL && ret != -ENOSYS) {
+		pr_err("membarrier(1 << %d) returned %d\n", cmd_bit, ret);
+		return ret;
```
Snorch, line 232: Maybe `return -1` ? I don't see any reason to return `-E<smth>`. It is normally a bad sign if we start returning something different from -1 on error paths except when we need it.

Also consider printing strerror(-ret) instead of just ret in the above pr_err.

criu/pie/parasite.c
```diff
@@ -225,6 +261,20 @@ static int dump_misc(struct parasite_dump_misc *args)
 	args->dumpable = sys_prctl(PR_GET_DUMPABLE, 0, 0, 0, 0);
 	args->thp_disabled = sys_prctl(PR_GET_THP_DISABLE, 0, 0, 0, 0);
 
+	args->membarrier_registration_mask = 0;
+	ret = get_membarrier_registration_mask(MEMBARRIER_CMDBIT_PRIVATE_EXPEDITED);
+	if (ret < 0)
+		return ret;
```
Snorch, line 267: `return -1;`

criu/pie/parasite.c
osctobe: The dump_misc() function already returns the -errno from a syscall (prctl). This is the parasite code - I see no strerror() uses here.

criu/pie/parasite.c
osctobe: Isn't this redundant vs the comment above?

test/zdtm/static/membarrier.c
osctobe: I've reworked the test a bit to make it more readable. Since this is a few lines change to support showing all registration/check compared to failing on first one, I think it's worth to have the extra info from the start. I see this a bit like insurance: you pay a bit more now, to reduce the amount of work in an emergency (here: a bug somewhere, not necessarily CRIU) in the future.

criu/pie/parasite.c
avagin: ret != -EPERM ?

avagin: > The dump_misc() function already returns the -errno from a syscall (prctl). 

This is the bad argument. It doesn't matter if any other code do the same thing. We should have reasons to do any specific actions. In error cases, we prefer to return -1, because it is very deterministic and hard to handle it incorrectly.

test/zdtm/static/membarrier.c
```diff
@@ -0,0 +1,116 @@
...
+	int register_cmd;
+	int execute_cmd;
+} membarrier_cmds[] = {
+	{ "",           MEMBARRIER_CMD_REGISTER_PRIVATE_EXPEDITED,           MEMBARRIER_CMD_PRIVATE_EXPEDITED },
+	{ "_SYNC_CORE", MEMBARRIER_CMD_REGISTER_PRIVATE_EXPEDITED_SYNC_CORE, MEMBARRIER_CMD_PRIVATE_EXPEDITED_SYNC_CORE },
+	{ "_RSEQ",      MEMBARRIER_CMD_REGISTER_PRIVATE_EXPEDITED_RSEQ,      MEMBARRIER_CMD_PRIVATE_EXPEDITED_RSEQ },
+};
+static const int n_membarrier_cmds = sizeof(membarrier_cmds) / sizeof(*membarrier_cmds);
+
+static int register_membarriers()
```
avagin, line 36: membarrier.c:36:12: error: function declaration isn't a prototype [-Werror=strict-prototypes]
   36 | static int register_membarriers()

avagin:
> This is a best-effort implementation that ignores unknown modes. The proper migration of membarrier() needs kernel support for querying the state (currently e.g. registration for MEMBARRIER_CMD_GLOBAL_EXPEDITED is not possible to detect).

Now we have MEMBARRIER_CMD_GET_REGISTRATIONS
https://github.com/torvalds/linux/commit/544a4f2ecd45f9d6ed78d207583f39130ad40349

Should we start using it to solve this problem?

mclapinski:
MEMBARRIER_CMD_GET_REGISTRATIONS exists only on fairly new kernels. IMO it's better to merge this and then write a new PR to use MEMBARRIER_CMD_GET_REGISTRATIONS with this implementation as fallback.

criu/pie/restorer.c
```diff
@@ -1521,6 +1521,31 @@ int cleanup_current_inotify_events(struct task_restore_args *task_args)
...
+	int i, failed_mask = 0;
+
+	if (!mask)
+		return 0;
+
+	pr_info("Restoring membarrier() registrations %x\n", mask);
+
+	for_each_bit(i, bitmap)
+		if (sys_membarrier(1 << i, 0, 0))
+			failed_mask |= 1 << i;
```
avagin, line 1539: we need to print a return code in the error message, otherwise it will be impossible to determine a reason of an error.

criu/pie/restorer.c
```diff
@@ -1537,6 +1537,30 @@ int cleanup_current_inotify_events(struct task_restore_args *task_args)
...
+	if (!mask)
+		return 0;
+
+	pr_info("Restoring membarrier() registrations %x\n", mask);
+
+	for_each_bit(i, bitmap) {
+		err = sys_membarrier(1 << i, 0, 0);
+		if (!err)
+			continue;
+		pr_err("Can't restore membarrier(1 << %d) registration: %d\n", err);
```
rst0git, line 1557:
```
criu/include/log.h:49:52: error: format ‘%d’ expects a matching ‘int’ argument [-Werror=format=]
   49 | #define pr_err(fmt, ...) print_on_level(LOG_ERROR, "Error (%s:%d): " LOG_PREFIX fmt, __FILE__, __LINE__, ##__VA_ARGS__)
      |                                                    ^~~~~~~~~~~~~~~~~
criu/pie/restorer.c:1557:17: note: in expansion of macro ‘pr_err’
 1557 |                 pr_err("Can't restore membarrier(1 << %d) registration: %d\n", err);
      |                 ^~~~~~
```

osctobe:
@avagin ping? BTW, I think even with the `no-auto-close` tag this PR seems to be falling through the cracks.

mclapinski:
Still fails CI. I don't know about the Docker test but the other 2 are real failures.

avagin:
Merged with a few minor fixes.

@osctobe I don't review CL-s that have CI failures without explanations why they are unrelated to new changes.

PR #2027, opened 2022-12-15
link-remap: Add --keep-link-remaps option

ajwock:
When specified, this option disables the automatic deletion of link-remaps on restore.  This allows checkpoints dumped with --link-remap to be restored multiple times (provided that other conditions for reuse are met).


ajwock:
As a heads up, I also have another patch coming after this one which adds extra uniqueness to the link-remap file that --link-remap creates.  It turns out that when dumping similar processes multiple times, the same name tends to be generated and that causes collisions and failed dumps.

ajwock:
Closing this to change branch

PR #2028, opened 2022-12-15
link-remap: Add --keep-link-remaps option

ajwock:
When specified, this option disables the automatic deletion of link-remaps on restore.  This allows checkpoints dumped with --link-remap to be restored multiple times (provided that other conditions for reuse are met).


rst0git:
@ajwock What is the use case for this option? Would you be able to add a test?

criu/files-reg.c
```diff
@@ -895,6 +895,9 @@ int try_clean_remaps(bool only_ghosts)
 	struct remap_info *ri;
 	int ret = 0;
 
+	if (opts.keep_link_remaps)
+		return ret;
```
Snorch, line 899: Can we be more explicit here and do `return 0;`

Snorch:
One possible downside of --keep-link-remaps is that if you leave "remap/path/file.cr_link" on the filesystem you would likely fail dumping the same process again as link creation would fail with EEXIST. (probably after first dump you will have only "remap/path/file" leftover on the filesystem and after second you would get "remap/path/file.cr_link" and third dump would fail, but anyway, see code in open_path around rfi_remap)

ajwock:
> One possible downside of --keep-link-remaps is that if you leave "remap/path/file.cr_link" on the filesystem you would likely fail dumping the same process again as link creation would fail with EEXIST. (probably after first dump you will have only "remap/path/file" leftover on the filesystem and after second you would get "remap/path/file.cr_link" and third dump would fail, but anyway, see code in open_path around rfi_remap)

It's interesting you mention this:  I have another patch I was planning on submitting after this one that finds a unique name for the remap in the case of EEXIST.  Perhaps I should include that as part of this patch?

PR #2030, opened 2022-12-26
Fix timerfd restore problem with double parsing

hejingxian123:
We restore timerfd with the state of it_value and it_interval. 
However, when it_value is zero, the timer will be restore without running. 
Because it_value is changing with the timer running, we can restore the timerfd with double parsing the it_value.
If the timer is running, the it_value can be non-zero at least once.

Signed-off-by: Jingxian He <hejingxian@huawei.com>


criu/timerfd.c
```diff
@@ -75,6 +75,12 @@ static int dump_one_timerfd(int lfd, u32 id, const struct fd_parms *p)
 	tfe.id = id;
 	tfe.flags = p->flags;
 	tfe.fown = (FownEntry *)&p->fown;
+
+	/* when it_value is zero, we need parse it again */
```
avagin, line 79: Could you write a more details comment here?  Right now, it is unclear why we need to parse it again.

avagin:
Can we introduce a test for this case?

PR #2031, opened 2023-01-02
cgroups: ignore EOPNOTSUPP on setting memory.kmem.limit_in_byte

avagin:
memory.kmem.limit_in_bytes has been deprecated. Look at e7c4184164f7 ("memcg, kmem: further deprecate kmem.limit_in_bytes") for more details.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1402,7 +1402,7 @@ static int restore_cgroup_subtree_control(const CgroupPropEntry *cg_prop_entry_p
 static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *path, int off, bool split_lines,
 			       bool skip_fails)
 {
-	int cg, fd, ret = -1, flag;
+	int cg, fd, exit_code = -1, flag;
 	CgroupPerms *perms = cg_prop_entry_p->perms;
 	int is_subtree_control = !strcmp(cg_prop_entry_p->name, "cgroup.subtree_control");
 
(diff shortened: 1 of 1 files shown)


mihalicyn:
LGTM

just for reference:
https://github.com/checkpoint-restore/criu/pull/1874
https://github.com/mihalicyn/criu/commit/00fbd541700d990f87850959f7b27ea8c5a4f134

cc @Snorch

Snorch:
What I think we can improve:

- we can check kmem limit availability once in kdat, so that we don't need to write to it for each cgroup directory but only check it once and skip later
note: see how I did it here https://src.openvz.org/projects/OVZ/repos/criu/commits/6aa0d6cfe062a34edcedea06acd334b65eae776c#criu/cgroup.c (if you like I can port it)

- we can remove kmem limit from here https://github.com/checkpoint-restore/criu/blob/008c2b9c7f86a5e27fc721a28e3669134ef98969/criu/cgroup-props.c#L430 as it is deprecated anyways

avagin:
> What I think we can improve:
> 
> * we can check kmem limit availability once in kdat, so that we don't need to write to it for each cgroup directory but only check it once and skip later
>   note: see how I did it here https://src.openvz.org/projects/OVZ/repos/criu/commits/6aa0d6cfe062a34edcedea06acd334b65eae776c#criu/cgroup.c (if you like I can port it; note that I use /sys/fs/cgroup/memory hardcoded path there, it can be improved)

It is up to you.

> * we can remove kmem limit from here https://github.com/checkpoint-restore/criu/blob/008c2b9c7f86a5e27fc721a28e3669134ef98969/criu/cgroup-props.c#L430
>    as it is deprecated anyways

We need to C/R it where it is supported.

PR #2035, opened 2023-01-04
cgroup: fix make indent error in kmem limit check

Snorch:
Fixes: a98722320 ("cgroups: ignore EOPNOTSUPP on setting memory.kmem.limit_in_byte")

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1477,8 +1477,7 @@ static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *pat
 		/* memory.kmem.limit_in_bytes has been deprecated. Look at
 		 * 58056f77502f3 ("memcg, kmem: further deprecate
 		 * kmem.limit_in_bytes") for more details. */
-		if (ret == -1 && errno == EOPNOTSUPP &&
-		    !strcmp(cg_prop_entry_p->name, "memory.kmem.limit_in_bytes"))
+		if (ret == -1 && errno == EOPNOTSUPP && !strcmp(cg_prop_entry_p->name, "memory.kmem.limit_in_bytes"))
 			ret = len;
 		if (ret != len) {
 			pr_perror("Failed writing %s to %s", cg_prop_entry_p->value, path);


criu/cgroup.c
```diff
@@ -1477,8 +1477,7 @@ static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *pat
 		/* memory.kmem.limit_in_bytes has been deprecated. Look at
 		 * 58056f77502f3 ("memcg, kmem: further deprecate
 		 * kmem.limit_in_bytes") for more details. */
-		if (ret == -1 && errno == EOPNOTSUPP &&
-		    !strcmp(cg_prop_entry_p->name, "memory.kmem.limit_in_bytes"))
+		if (ret == -1 && errno == EOPNOTSUPP && !strcmp(cg_prop_entry_p->name, "memory.kmem.limit_in_bytes"))
```
avagin, line 1480: This is annoying. Can we teach clang-format to ignore cases when a developer breaks a line as they want?

Snorch: It's our codding stile to have long(120) lines, isn't it? =)

Alternatives:
- use  `/* clang-format off/on */` everywhere where we need custom format
- don't use clang format as we do it in make indent, but always only check patches from PR with git-clang-format, this way some old code with bad formating would not be checked by clang, and we would be thus allowed to leave some custom format here and there.

e.g. to check your branch which is based on top of criu-dev:
```git clang-format --style file --extensions c,h --quiet origin/criu-dev```

Comment (cyrillos) [reply criu/cgroup.c:1480]: Since I've grown by linux kernel code I think using clang format in criu code is a big mistake. Initially code been written the way to keep it readable even if you sit more than 12 hours a day in front of display. For example we had

```
#define FD_PARMS_INIT                  \
(struct fd_parms) {                    \
       .fd     = FD_DESC_INVALID,      \
       .fown   = FOWN_ENTRY__INIT,     \
       .link   = NULL,                 \
       .mnt_id = -1,                   \
}
```

which become

```
#define FD_PARMS_INIT                                                                        \
       (struct fd_parms)                                                                    \
       {                                                                                    \
               .fd = FD_DESC_INVALID, .fown = FOWN_ENTRY__INIT, .link = NULL, .mnt_id = -1, \
       }
```

do you really think the last one is more readable?!

`clang-format` might be useful when you're drowning by permanent patches flow from persons which didn't care to read kernel coding style guide and you finally gave up and require them to pass autoformatter before applying a patch. This was never the case for CRIU, at least initially. So I don't understand why you guys merged all this into. It is not about personal preferences but rather about wasting other's men-hours for nothing. The complexity of CRIU code implies that contributor is **a sane person** and can write code similarly to other code style inside the project.

Currently `clang-format` already ruined a bunch of code inside (just like example above), or say
```
       struct list_head        mnt_bind;       /* circular list of derivatives of one real mount */
       struct list_head        mnt_share;      /* circular list of shared mounts */
       struct list_head        mnt_slave_list; /* list of slave mounts */
       struct list_head        mnt_slave;      /* slave list entry */
       struct mount_info       *mnt_master;    /* slave is on master->mnt_slave_list */
       struct list_head        mnt_propagate;  /* circular list of mounts which propagate from each other */
       struct list_head        mnt_notprop;    /* temporary list used in can_mount_now */
       struct list_head        mnt_unbindable; /* list of mounts with delayed unbindable */
```

which after formatter become

```
       struct list_head mnt_bind; /* circular list of derivatives of one real mount */
       struct list_head mnt_share; /* circular list of shared mounts */
       struct list_head mnt_slave_list; /* list of slave mounts */
       struct list_head mnt_slave; /* slave list entry */
       struct mount_info *mnt_master; /* slave is on master->mnt_slave_list */
       struct list_head mnt_propagate; /* circular list of mounts which propagate from each other */
       struct list_head mnt_notprop; /* temporary list used in can_mount_now */
       struct list_head mnt_unbindable; /* list of mounts with delayed unbindable */
```

see? We've been using words separation to make text less "heavy" for eyes parsing which simply trashed after formatter.

My personal opinion is that -- `clang-format` should be a great tool when you use it from the scratch and if you code requires not much exceptions from some "standard" style chosen by formatter authors, otherwise it causes more problem. So have fun with it :-)

Snorch: Please look on https://github.com/checkpoint-restore/criu/pull/2044 maybe it's a solution.

Snorch:
Close in favor of https://github.com/checkpoint-restore/criu/pull/2044

PR #2039, opened 2023-01-05
zdtm: fix 'zdtm.py list' command

rst0git:
After commit https://github.com/checkpoint-restore/criu/commit/0add1b6cdbedaaac93d48ea323beceb7328b69b2, the command `./zdtm.py list` fails with the following error.

    if opts['rootless']:
       ~~~~^^^^^^^^^^^^
    KeyError: 'rootless'

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -626,14 +626,14 @@ class zdtm_test:
                 ["make", "zdtm_ct"], env=dict(os.environ, MAKEFLAGS=""))
         if not os.access("zdtm/lib/libzdtmtst.a", os.F_OK):
             subprocess.check_call(["make", "-C", "zdtm/"])
-        if opts['rootless']:
+        if 'rootless' in opts and opts['rootless']:
             return
         subprocess.check_call(
             ["flock", "zdtm_mount_cgroups.lock", "./zdtm_mount_cgroups", str(uuid)])
 
     @staticmethod
     def cleanup():
-        if opts['rootless']:
+        if 'rootless' in opts and opts['rootless']:
             return
         subprocess.check_call(
             ["flock", "zdtm_mount_cgroups.lock", "./zdtm_umount_cgroups", str(uuid)])


adrianreber:
I think it makes sense, but it looks strange. Do you understand what is going on? Can you explain it a bit more in detail why `rootless` is not always a key of `opts`? Is it sometimes not set at all? I think I do not understand the code from a quick look at it.

rst0git:
> Can you explain it a bit more in detail why rootless is not always a key of opts? 

The `--rootless` option has been added to the `run` command.
```
./zdtm.py run --help | grep rootless
                           [--freezecg FREEZECG] [--user] [--rootless] [--rpc]
  --rootless            Run CRIU rootless (uid!=0) (needs

```
All other commands (`list`, `group`, `clean`) do not have this option.

test/zdtm.py
```diff
@@ -626,14 +626,14 @@ def available():
                 ["make", "zdtm_ct"], env=dict(os.environ, MAKEFLAGS=""))
         if not os.access("zdtm/lib/libzdtmtst.a", os.F_OK):
             subprocess.check_call(["make", "-C", "zdtm/"])
-        if opts['rootless']:
+        if 'rootless' in opts and opts['rootless']:
```
avagin, line 629: I think we should not run these actions if we are not going to execute tests:
index a311610c3..99b59f551 100755
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -618,6 +618,8 @@ class zdtm_test:
 
     @staticmethod
     def available():
+        if opts['action'] != "run":
+            return
         if not os.access("umount2", os.X_OK):
             subprocess.check_call(
                 ["make", "umount2"], env=dict(os.environ, MAKEFLAGS=""))
@@ -633,6 +635,8 @@ class zdtm_test:
 
     @staticmethod
     def cleanup():
+        if opts['action'] != "run":
+            return
         if opts['rootless']:
             return
         subprocess.check_call(

rst0git: It is slightly more complicated. `available()` invokes `make -C zdtm/` that then call `make` in the `lib`, `static`, and `transition` sub-directories to build the tests. Then, the `list` command counts and prints the list of tests that have been built. Thus, with this change `./zdtm.py list` would not work.

avagin: > available() invokes make -C zdtm/ that then calls make in the lib, static, and transition sub-directories to build the tests.

This part never works as expected, because it builds tests only if zdtm/lib/libzdtmtst.a doesn't exist, but it doesn't mean that all tests are compiled and binaries are up to date. I don't think that zdtm.py has to compile anything.

PR #2040, opened 2023-01-06
log: Write more details in write_pidfile

Snorch:
This is a backport of:
https://src.openvz.org/projects/OVZ/repos/criu/commits/8e6bc1aeb

### DIFF
--- a/criu/log.c
+++ b/criu/log.c
@@ -397,15 +397,28 @@ void print_on_level(unsigned int loglevel, const char *format, ...)
 
 int write_pidfile(int pid)
 {
-	int fd;
+	int fd, ret, exit_code = -1;
 
 	fd = open(opts.pidfile, O_WRONLY | O_EXCL | O_CREAT, 0600);
 	if (fd == -1) {
-		pr_perror("Can't open %s", opts.pidfile);
+		pr_perror("pidfile: Can't open %s", opts.pidfile);
 		return -1;
 	}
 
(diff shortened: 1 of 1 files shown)


criu/log.c
```diff
@@ -397,15 +397,22 @@ void print_on_level(unsigned int loglevel, const char *format, ...)
...
 
-	dprintf(fd, "%d", pid);
+	ret = dprintf(fd, "%d", pid);
+	if (ret < 0) {
+		pr_perror("pidfile: Can't write pid %d to %s", pid, opts.pidfile);
+		close(fd);
+		return -1;
+	}
+
+	pr_debug("pidfile: Wrote pid %d to %s (%d bytes)\n", pid, opts.pidfile, ret);
```
adrianreber, line 415: Looking at the return at the end of the function which returns -1 in the case of ret being 0 it is probably a bit strange to have this `pr_debug()` line if 0 bytes were written.

avagin:
pls remove a link to external resources, it doesn't add any value here, you can add it in a comment to this pr.

criu/log.c
Snorch: Sure, making all cases separate has more logic here as we need to print differently in different cases.

Snorch:
> pls remove a link to external resources, it doesn't add any value here, you can add it in a comment to this pr.

Ok. I thought about it not as a valuable info, but more of giving a credit to the original source.

PR #2044, opened 2023-01-11
clang-format: rework make indent to check specific commits

Snorch:
Previousely "make indent" checked all files in criu source directory for codding style flaws. We have several problems with it:

- clang-format default format sometimes changes in new versions of the package and we need to reformat all our code base each time it happens
- on different systems we may have different versions of clang-format and on latest criu-dev "make indent" may be still unhappy on your system
- when we want to update clang-format rules ourselves we need to update all our code base each time
- sometimes clang-format rules are not fitting all our cases, (e.g.: an option IndentGotoLabels works nice for simple C code, but is a no go for assembler and C macros) and putting "clang-format off" everywhere is a mess
- sometimes we intentionally want to break clang-format rules (e.g.: we want to put function arguments on a new line separating them "logically" not "mechanically" following 120-char rule like clang-format does).

This adds a BASE option for "make indent" where all commits in range BASE..HEAD would be checked with git-clang-format for codding style flaws. For instance when developing on top of criu-dev, one can use "make BASE=origin/criu-dev indent" to check all their commits for compliance with the clang-format rules. Default base is HEAD~1 to make last commit checked when "make indent" is called. The closest thing to the old behaviour would then be "make indent BASE=init", note that only commited files would be checked.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -461,8 +461,10 @@ fetch-clang-format: .FORCE
 	$(E) ".clang-format"
 	$(Q) scripts/fetch-clang-format.sh
 
+BASE ?= "HEAD~1"
+OPTS ?= "--quiet"
 indent:
-	find . -name '*.[ch]' -type f -print0 | xargs --null --max-args 128 --max-procs 4 clang-format -i
+	git clang-format --style file --extensions c,h $(OPTS) $(BASE)
 .PHONY: indent
 
 include Makefile.install
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -23,16 +23,12 @@ jobs:
     - name: Run make indent
       run: >
         if [ -z "${{github.base_ref}}" ]; then
-          make indent
+          if ! make indent OPTS=--diff; then
+            exit 1
+          fi
         else
           git fetch origin ${{github.base_ref}} &&
-          git clang-format --style file --extensions c,h --quiet origin/${{github.base_ref}}
-        fi &&
-        STATUS=$(git status --porcelain) &&
-        if [ ! -z "$STATUS" ]; then
-          echo "FAIL: some files are not correctly formatted.";
-          echo "$STATUS"
-          git diff
-          echo "FAIL: please run 'make indent'";
-          exit 1;
+          if ! make indent OPTS=--diff BASE=origin/${{github.base_ref}}; then
+            exit 1
+          fi
         fi


Snorch:
@checkpoint-restore/maintainers Please take a look.

Makefile
```diff
@@ -461,8 +461,9 @@ fetch-clang-format: .FORCE
 	$(E) ".clang-format"
 	$(Q) scripts/fetch-clang-format.sh
 
+BASE ?= "HEAD~1"
 indent:
-	find . -name '*.[ch]' -type f -print0 | xargs --null --max-args 128 --max-procs 4 clang-format -i
+	git clang-format --style file --extensions c,h --quiet $(BASE) || true
```
avagin, line 466: why do we need true here?

Snorch: The git-clang-format exits with non-zero exit code if it finds mistakes in code, if we remove `|| true` we have:

```
[snorch@turmoil criu]$ git diff
diff --git a/Makefile b/Makefile
index e055adcac..af3888e79 100644
--- a/Makefile
+++ b/Makefile
@@ -463,7 +463,7 @@ fetch-clang-format: .FORCE
 
 BASE ?= "HEAD~1"
 indent:
-       git clang-format --style file --extensions c,h --quiet $(BASE) || true
+       git clang-format --style file --extensions c,h --quiet $(BASE)
 .PHONY: indent
 
 include Makefile.install
[snorch@turmoil criu]$ make indent BASE=008c2b9c7f86a5e27fc721a28e3669134ef98969
git clang-format --style file --extensions c,h --quiet 008c2b9c7f86a5e27fc721a28e3669134ef98969
make: *** [Makefile:466: indent] Error 1
```
I decided to hide this error but didn't find any flag of git-clang-format itself to do it.

Reasons for hiding:
- One of usecases of `make indent` I believe is to check or fix format in newly created patch. In this case it is pretty much expected that git-clang-format would fix something and printing error each time may be missleading.
- We used to check git-diff after make indent anyways (e.g. in lint.yml) printing our custom nicely formated error anyways.

avagin: pls add a comment that explains this

avagin: Oops, I read the comment again and I think we don't need to hide the error if it finds mistakes. I expects it exits with a non-zero code in this case.  And I think we don't need the --quiet option here.

> One of usecases of make indent I believe is to check or fix format in newly created patch. In this case it is pretty much expected that git-clang-format would fix something and printing error each time may be missleading.

It depends how to look at this, but I understand your concerns. I would suggest to add the --diff option instead of hiding the error code.

Snorch: > I would suggest to add the --diff option instead of hiding the error code.

```
--diff                print a diff instead of applying the changes
```
If we use diff option we would need to apply changes manually. Is it what we want?

Snorch: See updated variant, hope it would be ok.

adrianreber:
I have no strong opinion, but I prefer to have a tool that does it all automatically. I prefer to not spend time on code formatting. Let the tool do the job. I prefer the old setup, but whatever works best for you.

Snorch:
> I have no strong opinion, but I prefer to have a tool that does it all automatically.

We don't remove the tool, we just restrict it from formatting "everything else". And thus you need to commit your changes before formatting them to separate them from "everything else", this should not be a big deal to do.

> I prefer to not spend time on code formatting. Let the tool do the job.

You still should check your auto-formated code after auto-formating it for frankensteins(https://github.com/checkpoint-restore/criu/pull/2046), and checking code carefully requires, I believe, much more attentiveness than writing code with right codding style initially "from the tips of you fingers". =) 

> I prefer the old setup, but whatever works best for you.

Old setup enforces "codding style" just for "codding style". Initial idea behind codding style is a code readability/searchability and we should probably allow to break clang-format-"codding style" a little to make code more readable in some cases.

I honestly believe this change should not change patch workflow much for everyone.

Makefile
```diff
@@ -461,8 +461,10 @@ fetch-clang-format: .FORCE
 	$(E) ".clang-format"
 	$(Q) scripts/fetch-clang-format.sh
 
+BASE ?= "HEAD~1"
```
avagin, line 464: Should it be origin/criu-dev?

avagin, line 464: do you need to fix our CI to set the right base?

Issue #2036, opened 2023-01-04
Missing rseq headers on centos7 in CI


Snorch:
```
In file included from criu/include/parasite.h:13:0,
                 from criu/cr-dump.c:46:
criu/include/linux/rseq.h:7:22: fatal error: sys/rseq.h: No such file or directory
 #include <sys/rseq.h>
                      ^
compilation terminated.
```

https://github.com/checkpoint-restore/criu/pull/2030/checks?check_run_id=10303595406
https://github.com/checkpoint-restore/criu/pull/2022/checks?check_run_id=10145226897

mihalicyn:
I'll take a look

rst0git:
I was able to replicate this error locally using CentOS 7 container. The error occurs only when the `libbsd-devel` package is installed.

adrianreber:
Thanks. So we either have to not use libbsd or a newer compiler. We are using strlcopy, strlcat and setproctitle from libbsd but there are workarounds if libbsd is not available. Sounds like possible workaround.

PR #2054, opened 2023-01-17
ci: fix build on centos7 with libbsd-devel

rst0git:
The libbsd-devel package is an optional dependency that enables support for setproctitle(). However, building CRIU on CentOS 7 with this package fails with the following error (https://github.com/checkpoint-restore/criu/issues/2036):

```
criu/include/linux/rseq.h:7:22: fatal error: sys/rseq.h: No such file or directory
 #include <sys/rseq.h>
```

### DIFF
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -37,13 +37,17 @@
 #include "crtools.h"
 #include "cr_options.h"
 #include "servicefd.h"
-#include "string.h"
 #include "ptrace-compat.h"
 #include "util.h"
 #include "namespaces.h"
 #include "image.h"
 #include "proc_parse.h"
 #include "parasite.h"
+/* string.h is included after parasite.h because the libbsd headers define
+ * __has_include as the constant 1.
+ * https://github.com/checkpoint-restore/criu/issues/2036
+ */
+#include "string.h"
 #include "parasite-syscall.h"
 #include "compel/ptrace.h"
 #include "files.h"


adrianreber (approved):
Sounds like a simple change to fix CI

adrianreber:
I opened a ticket also here: https://bugzilla.redhat.com/show_bug.cgi?id=2161633

Snorch:
Looks good. Though we can also have alternative approach https://github.com/checkpoint-restore/criu/pull/2057 to just check if __has_include works, this way likely criu would be able to compile on centos7 even with libbsd installed.

avagin:
@rst0git @Snorch can we create a wrapper for setproctitle_init and put it in a separate C file? In this case, libbsd headers will be included only in that file and so they will not interfere with other headers.

rst0git:
@avagin This is a great idea. Actually, `setproctitle.h` is the only place where we include the libbsd headers. It would be sufficient to simply change the order of include statements to fix this problem. I've updated this pull request.

@Snorch @adrianreber PTAL

criu/cr-dump.c
```diff
@@ -37,13 +37,17 @@
...
 #include "util.h"
 #include "namespaces.h"
 #include "image.h"
 #include "proc_parse.h"
 #include "parasite.h"
+/* string.h is included after parasite.h because the libbsd headers define
+ * __has_include as the constant 1.
+ * https://github.com/checkpoint-restore/criu/issues/2036
+ */
+#include "string.h"
```
Snorch, line 50: This solution seems too fragile for me. At some point we might need to include "linux/rseq.h" somewhere else, not only inside parasite.h.

Snorch:
> @rst0git @Snorch can we create a wrapper for setproctitle_init and put it in a separate C file? In this case, libbsd headers will be included only in that file and so they will not interfere with other headers.

please see what I managed to do developing this idea: https://github.com/checkpoint-restore/criu/pull/2058

rst0git:
Closing in favour of https://github.com/checkpoint-restore/criu/pull/2058

PR #2057, opened 2023-01-17
rseq: hack to check if __has_include always returns true

Snorch:
If __has_include("defunct/defunct.h") on non-existant header returns
true it means __has_include dos not work properly and can't be used.

https://github.com/checkpoint-restore/criu/issues/2036

### DIFF
--- a/criu/include/linux/rseq.h
+++ b/criu/include/linux/rseq.h
@@ -3,9 +3,11 @@
 #define _UAPI_LINUX_RSEQ_H
 
 #ifdef __has_include
+#if !__has_include("defunct/defunct.h")
 #if __has_include("sys/rseq.h")
 #include <sys/rseq.h>
 #include "asm/thread_pointer.h"
+#endif
 #endif
 #endif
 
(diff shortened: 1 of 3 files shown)


criu/include/linux/rseq.h
```diff
@@ -3,11 +3,13 @@
 #define _UAPI_LINUX_RSEQ_H
 
 #ifdef __has_include
+#if !__has_include("defunct/defunct.h")
```
avagin, line 6: you need to add a comment that explains what it is for.

test/zdtm/static/rseq00.c
```diff
@@ -20,10 +20,12 @@
 #include "zdtmtst.h"
 
 #ifdef __has_include
+#if !__has_include("defunct/defunct.h")
```
avagin, line 23: same here and in other places

Snorch:
Close in favor of https://github.com/checkpoint-restore/criu/pull/2058

PR #2058, opened 2023-01-18
Hide bsd headers

Snorch:
string: define wrapers __setproctitle and __setproctitle_init to hide bsd headers

We see that libbsd redefines __has_include to be always true, which
breaks such checks for rseq. The idea behind this patch is to put all
uses of libbsd functions to separate c files and only export wrapper
functions for them.

Using __setproctitle and __setproctitle_init everywhere in existing
code:

git grep --files-with-matches "setproctitle" | xargs sed -i 's/setproctitle/__setproctitle/g'
git grep --files-with-matches "setproctitle_init" | xargs sed -i 's/setproctitle_init/__setproctitle_init/g'

Fixes: #2036
Suggested-by: Andrei Vagin <avagin@google.com>

string: define wrapers __strlcpy and __strlcat to hide bsd headers

We see that libbsd redefines __has_include to be always true, which
breaks such checks for rseq. The idea behind this patch is to put all
uses of libbsd functions to separate c files and only export wrapper
functions for them.

Using __strlcat and __strlcpy everywhere in existing code:
git grep --files-with-matches "strlcat" | xargs sed -i 's/strlcat/__strlcat/g'
git grep --files-with-matches "strlcpy" | xargs sed -i 's/strlcpy/__strlcpy/g'

Fixes: #2036
Suggested-by: Andrei Vagin <avagin@google.com>

### DIFF
--- a/criu/string.c
+++ b/criu/string.c
@@ -6,7 +6,6 @@
 
 #include "string.h"
 
-#ifndef CONFIG_HAS_STRLCPY
 /**
  * strlcpy - Copy a %NUL terminated string into a sized buffer
  * @dest: Where to copy the string to
@@ -18,7 +17,7 @@
  * of course, the buffer size is zero). It does not pad
  * out the result like strncpy() does.
  */
-size_t strlcpy(char *dest, const char *src, size_t size)
+size_t __strlcpy(char *dest, const char *src, size_t size)
 {
 	size_t ret = strlen(src);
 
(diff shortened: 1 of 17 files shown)


mihalicyn:
Issue on mips64el-unstable-cross looks not related to this.
```
  GEN      criu/pie/restorer-blob.h
Error (compel/src/lib/handle-elf-host.c:337): Unexpected undefined symbol: `strlen'. External symbol in PIE?
make[2]: *** [criu/pie/Makefile:58: criu/pie/restorer-blob.h] Error 255
```

LGTM

mihalicyn (approved):

rst0git (approved):
LGTM

adrianreber:
The libbsd dependency could also be dropped completely. strlcat and strlcpy are implemented in CRIU already, which means only setproctitle would be lost. I think I have not seen the result of setproctitle although I often interact with CRIU through the RPC service.

If setproctitle should be kept: LGTM

criu/string.c
```diff
@@ -18,7 +22,7 @@
  * of course, the buffer size is zero). It does not pad
  * out the result like strncpy() does.
  */
-size_t strlcpy(char *dest, const char *src, size_t size)
+static size_t strlcpy(char *dest, const char *src, size_t size)
```
avagin, line 25: can we always use our strlcpy?

Snorch: Will do, ok.

Snorch:
Sorry for delay, I replaced libbsd strlcat and strlcpy with our __strlcat and __strlcpy completely. Please take a look.

adrianreber (approved):
Looks good. The only CI failure it the mips-unstable which seems to fail all the time.

PR #2062, opened 2023-01-24
log: make pr_* work nicely with %m

Snorch:
As vprint_on_level calls different functions in different cases it is possible that errno can become overwritten by some operation. Let's resset errno to the original value just before calling vsnprintf.

Note: in compel %m is not allowed so we don't need to fix errno there.

### DIFF
--- a/criu/log.c
+++ b/criu/log.c
@@ -316,6 +316,7 @@ static void early_vprint(const char *format, unsigned int loglevel, va_list para
 {
 	unsigned int log_size = 0;
 	struct early_log_hdr *hdr;
+	int _errno = errno;
 
 	if ((early_log_buf_off + sizeof(hdr)) >= EARLY_LOG_BUF_LEN)
 		return;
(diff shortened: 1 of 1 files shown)


adrianreber (approved):
Makes sense. The CI error is already being fixed in another PR. LGTM

Snorch:
Talking with Andrey in person we came to a problem:
```
       m      (Glibc extension; supported by uClibc and musl.)  Print
              output of strerror(errno).  No argument is required.
```
https://man7.org/linux/man-pages/man3/printf.3.html

Do we want to support running on other libcs? https://en.wikipedia.org/wiki/C_standard_library

So alternatively to this approach we can remove all uses of %m and add rule to lint that %m should not be used, only strerror would be allowed.

adrianreber:
> Do we want to support running on other libcs?

We kind of already do by testing everything on Alpine. In that case removing %m sounds right.

Snorch:
Closing in favor of https://github.com/checkpoint-restore/criu/pull/2063

PR #2063, opened 2023-01-25
Prohibit to use procent m in pr  functions

Snorch:
As our pr_* functions are complex and can call different system calls
inside before actual printing (e.g. gettimeofday for timestamps) actual
errno at the time of printing may be changed.
    
Let's just use %s + strerror(errno) instead of %m with pr_* functions to
be explicit that errno to string transformation happens before calling
anything else.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -441,7 +441,7 @@ lint:
 	# Do not append \n to pr_perror or fail
 	! git --no-pager grep -E '^\s*\<(pr_perror|fail)\>.*\\n"'
 	# Do not use %m with pr_perror or fail
-	! git --no-pager grep -E '^\s*\<(pr_perror|fail)\>.*%m'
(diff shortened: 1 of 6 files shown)


Snorch:
I'm not sure that I didn't break PIE with my change in some unexpected way, does not feel safe to include:
```
+++ b/compel/include/log.h
+#include <errno.h>
+#include <string.h>
```
Would appreciate someone more experienced with PIE to take a look.

adrianreber (approved):

adrianreber:
> I'm not sure that I didn't break PIE with my change in some unexpected way, does not feel safe to include:
> 
> ```
> +++ b/compel/include/log.h
> +#include <errno.h>
> +#include <string.h>
> ```
> 
> Would appreciate someone more experienced with PIE to take a look.

I don't, but looking at the output of `nm` of `compel.a` `strerror` is already referenced by compel even without your changes. So I would say that you are not introducing a new dependency on an external symbol in compel which did not exist before.

avagin:
lgtm

PR #2064, opened 2023-01-25
dump: Show task comm early

Snorch:
When error happens on file dumping stage the only information about the task we dumping is its PID. For debug purpose show task's @comm early.

It proves useful when trying to understand which of dumped applications is "guilty" in brokern dump when pid is not there anymore.

--

This is a port of https://src.openvz.org/projects/OVZ/repos/criu/commits/487a62763
CC @cyrillos

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -24,6 +24,46 @@
 #include "xmalloc.h"
 #include "util.h"
 
+char *task_comm_info(pid_t pid, char *comm, size_t size)
+{
+	int ret = 0;
+
+	if (!pr_quelled(LOG_INFO)) {
+		int saved_errno = errno;
+		char path[32];
+		int fd;
+
+		snprintf(path, sizeof(path), "/proc/%d/comm", pid);
+		fd = open(path, O_RDONLY);
+		if (fd >= 0) {
+			ssize_t n = read(fd, comm, size);
+			if (n > 0)
+				comm[n - 1] = '\0';
+			else
+				ret = -1;
+			close(fd);
+		} else {
+			ret = -1;
+		}
+		errno = saved_errno;
+	}
+
+	if (ret || (pr_quelled(LOG_INFO) && comm[0]))
+		comm[0] = '\0';
+
+	return comm;
+}
+
+/*
+ * NOTE: Don't run simultaneously, it uses local static buffer!
+ */
+char *__task_comm_info(pid_t pid)
+{
+	static char comm[32];
+
+	return task_comm_info(pid, comm, sizeof(comm));
+}
+
 #define NR_ATTEMPTS 5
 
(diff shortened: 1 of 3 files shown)


Snorch:
note: linter errors are intentional, I believe splitting print messages where format string is on the first line and arguments are on second line is preferable to putting them all on the same line

adrianreber (approved):

rst0git (approved):
LGTM

criu/seize.c
```diff
@@ -24,6 +24,46 @@
...
+		int saved_errno = errno;
+		char path[32];
+		int fd;
+
+		snprintf(path, sizeof(path), "/proc/%d/comm", pid);
+		fd = open(path, O_RDONLY);
+		if (fd >= 0) {
+			ssize_t n = read(fd, comm, size);
+			if (n > 0)
+				comm[n - 1] = '\0';
```
avagin, line 41: Why do we override the last comm symbol?

criu/seize.c
```diff
@@ -24,6 +24,46 @@
 #include "xmalloc.h"
 #include "util.h"
 
+char *task_comm_info(pid_t pid, char *comm, size_t size)
+{
+	int ret = 0;
```
avagin, line 29: It looks more reasonable to initialize it with -1.

criu/seize.c
```diff
@@ -24,6 +24,46 @@
...
+		fd = open(path, O_RDONLY);
+		if (fd >= 0) {
+			ssize_t n = read(fd, comm, size);
+			if (n > 0)
+				comm[n - 1] = '\0';
+			else
+				ret = -1;
+			close(fd);
+		} else {
+			ret = -1;
```
avagin, line 46: it is better to print an error here

criu/seize.c
Snorch: This is to replace `\n` with `\0`. https://github.com/torvalds/linux/blob/f3a2439f20d918930cc4ae8f76fe1c1afd26958f/fs/proc/base.c#L1702

criu/seize.c
Snorch: Reworked it in:
https://github.com/checkpoint-restore/criu/pull/2095

criu/seize.c
Snorch: Added warnings in scope of https://github.com/checkpoint-restore/criu/pull/2095

criu/seize.c
avagin: It isn't obvious. I suggest to add a comment to explain this.

Snorch: Sure, done in:
https://github.com/checkpoint-restore/criu/pull/2095

PR #2074, opened 2023-02-03
net/iptables: remove libnft dependency for legacy iptables binary search

Snorch:
We don't use libnft in get_legacy_iptables_bin, so we can safely remove libnft dependency and always search for right legacy binary.


Snorch:
Also while on it let's use get_legacy_iptables_bin also on restore, to be consistent that we restore legacy->legacy only.

Snorch:
Closing and will resend 2-nd patch separately, first patch breaks centos7 where `iptables-save` does not have `-V` option, so it's probably reasonable to protect iptables legacy fallback by presence of libnft.

PR #2075, opened 2023-02-03
net: use get_legacy_iptables_bin also on restore

Snorch:
Without this we might try to restore iptables legacy images with iptables nft.

### DIFF
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -1193,7 +1193,7 @@ static int check_ipt_legacy(void)
 	char *ipt_legacy_bin;
 	char *ip6t_legacy_bin;
 
-	ipt_legacy_bin = get_legacy_iptables_bin(false);
+	ipt_legacy_bin = get_legacy_iptables_bin(false, false);
 	if (!ipt_legacy_bin) {
 		pr_warn("Couldn't find iptables version which is using iptables legacy API\n");
 		return -1;
@@ -1204,7 +1204,7 @@ static int check_ipt_legacy(void)
 	if (!kdat.ipv6)
 		return 0;
 
-	ip6t_legacy_bin = get_legacy_iptables_bin(true);
+	ip6t_legacy_bin = get_legacy_iptables_bin(true, false);
 	if (!ip6t_legacy_bin) {
 		pr_warn("Couldn't find ip6tables version which is using iptables legacy API\n");
 		return -1;
--- a/criu/include/util.h
+++ b/criu/include/util.h
@@ -384,7 +384,7 @@ static inline void print_stack_trace(pid_t pid)
 
 extern int mount_detached_fs(const char *fsname);
 
-extern char *get_legacy_iptables_bin(bool ipv6);
+extern char *get_legacy_iptables_bin(bool ipv6, bool restore);
 
 extern int set_opts_cap_eff(void);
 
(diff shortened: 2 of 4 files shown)


Snorch:
Note: linter error is intentional.

avagin:
@mihalicyn could you review this one?

mihalicyn (approved):
LGTM. Thanks!

It's my bad that I've not taken this into account in https://github.com/checkpoint-restore/criu/commit/d8821d9a8f2edb54c2f592e95fc80c62ccaa2ce9

Do I understand you correctly that this issue was found during migration between different nodes? AFAIU, it should not be reproducible with vzctl suspend/resume or snapshot or staff like that.

I think it's also worth to add some WARN on criu restore if we have nftables dump image, but have no nftables tool on the destination node:
https://github.com/checkpoint-restore/criu/blame/9fc413e360f1283d7c1c86907efb68db0bfd112e/criu/net.c#L2811
I'm ready to prepare PR. What do you think about that?

avagin:
> I think it's also worth to add some WARN on criu restore if we have nftables dump image, but have no nftables tool on the destination node:

It should be a fatal error.

> https://github.com/checkpoint-restore/criu/blame/9fc413e360f1283d7c1c86907efb68db0bfd112e/criu/net.c#L2811
> I'm ready to prepare PR. What do you think about that?

Go ahead.

Snorch:
> Do I understand you correctly that this issue was found during migration between different nodes? AFAIU, it should not be reproducible with vzctl suspend/resume or snapshot or staff like that.

No, found it while looking into code when I was rebasing VZ criu to 3.17.1.

Snorch:
Please let me know if something else is needed in scope of this pr, AFAICS now we are free to merge it.

Side note, not related to this PR, I believe that what we did in CRIU with libnft was a bad design choice:

nft tool can't dump some nft rules created with iptables-nft tool, for instance:
```
    ip6tables-nft -A INPUT -p ipv6-icmp -m icmp6 --icmpv6-type 130 -j ACCEPT  # Multicast Listener Query [RFC2710]
    ip6tables-nft -A INPUT -p ipv6-icmp -m icmp6 --icmpv6-type 131 -j ACCEPT  # Multicast Listener Report [RFC2710]
    ip6tables-nft -A INPUT -p ipv6-icmp -m icmp6 --icmpv6-type 132 -j ACCEPT  # Multicast Listener Done [RFC2710]
```
or `physdev` matches which are deprecated in nft.

Also we know that custom nft tool created table names (not "filter", "nat" or "mangle", etc.) are not dumped with iptables-nft.

And I even found an article where it is said that "Mixing iptables-nft and nft on the other hand is not unproblematic, either. A good rule of thumb to avoid issues is to not touch the tables iptables-nft creates with nft." https://www.redhat.com/en/blog/using-iptables-nft-hybrid-linux-firewall

So for now in Virtuozzo criu where we have all kinds of rules in containers both created with nft and iptables-nft I've prepared a fix:
https://github.com/OpenVZ/vzcriu/commit/d986e6df1aa9013556cc86440e8b7616c07299f0
https://github.com/OpenVZ/vzcriu/commit/8baea43807c725f57975b567f8a269158720b208
And it works somehow.

But what I wanted to state here is nft and iptables-nft are not good/robust enough for criu to use to save nftables. Instead they seem to be a big mess.

mihalicyn:
> Go ahead.

https://github.com/checkpoint-restore/criu/pull/2102 :)

Snorch:
> > Side note, not related to this PR, I believe that what we did in CRIU with libnft was a bad design choice.
> 
> pls file a new issue and assign it to @mihalicyn.

Ok, done https://github.com/checkpoint-restore/criu/issues/2105

PR #2089, opened 2023-02-23
crit: fix info action

rst0git:
In Python 3 `b'' == ''` is `False`. This causes the `crit info` action to fail with the following error.
```
  File "/usr/lib/python3.11/site-packages/crit-3.17-py3.11.egg/pycriu/images/images.py", line 178, in count
    size, = struct.unpack('i', buf)
            ^^^^^^^^^^^^^^^^^^^^^^^
  struct.error: unpack requires a buffer of 4 bytes
```
Reported-by: Sankalp Acharya (@sankalp-12)

### DIFF
--- a/lib/py/images/images.py
+++ b/lib/py/images/images.py
@@ -98,7 +98,7 @@ class entry_handler:
             # Read payload
             pbuff = self.payload()
             buf = f.read(4)
-            if buf == b'':
+            if len(buf) == 0:
                 break
             size, = struct.unpack('i', buf)
             pbuff.ParseFromString(f.read(size))
@@ -172,7 +172,7 @@ class entry_handler:
 
         while True:
             buf = f.read(4)
-            if buf == '':
+            if len(buf) == 0:
                 break
             size, = struct.unpack('i', buf)
             f.seek(size, 1)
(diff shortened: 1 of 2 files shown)


lib/py/images/images.py
```diff
@@ -172,7 +172,7 @@ def count(self, f):
 
         while True:
             buf = f.read(4)
-            if buf == '':
+            if len(buf) == 0:
```
Snorch, line 175: Alternatively we can use `b' '` on the right side of the assignment to have similar code to: 841edaf31 ("lib: a few fixes to be compatible with python3").

Also shouldn't we also update all other `if buf == ' ':` places:
ipc_msg_queue_handler:load
ipc_msg_queue_handler:skip

rst0git: Thank you Pavel for the code review!

I've updated the pull request to fix all other places as well. IMHO, `len(buf) == 0` is more readable and reliable solution for this problem (e.g., `buf == ' '` is not the same as `buf == ''`)

adrianreber (approved):

lib/py/images/images.py
Snorch: Sorry for `b' '`, this was a misprint, I meant `b''` =) And obviously you are right - len check is more readable.

Snorch:
@rst0git please look on images.py:pagemap_handler:load it seems you've missed one more `if buf == b'':` there.

rst0git:
@Snorch Thank you for pointing this out! I've updated the pull request.

Snorch (approved):

PR #2090, opened 2023-02-23
Add documentation for --ghost-fiemap

rst0git:
The `--ghost-fiemap` option was introduced with #1963. It enables an optimized algorithm based on fiemap ioctl that can reduce the number of syscalls used to checkpoint highly sparse ghost files.


Snorch:
@rst0git Thank you for documentating this option!

One thing I would like to add to your message is that "The option --ghost-fiemap is ON by default and can be disabled with --no-ghost-fiemap when using SEEK_HOLE/SEEK_DATA is preferred."

rst0git:
@Snorch I've updated the pull request with your input. Thanks!

Issue #2091, opened 2023-02-24
Podman Test fails in CI


Snorch:
https://github.com/checkpoint-restore/criu/actions/runs/4251213238/jobs/7403160954
```
2023-02-23T18:03:42.3288827Z make[3]: Nothing to be done for 'criu/string.c'.
2023-02-23T18:03:42.3405699Z   DEP      criu/unittest/unit.d
2023-02-23T18:03:42.3581977Z   DEP      criu/unittest/mock.d
2023-02-23T18:03:42.3734610Z   CC       criu/unittest/mock.o
2023-02-23T18:03:42.4287772Z   CC       criu/unittest/unit.o
2023-02-23T18:03:42.5524255Z   LINK     criu/unittest/built-in.o
2023-02-23T18:03:42.5668520Z   LINK     criu/unittest/unittest
2023-02-23T18:03:42.6077260Z Error (criu/config.c:127): Unsupported configuration file format. Please consult man page criu(8)
2023-02-23T18:03:42.6077639Z OK
2023-02-23T18:03:42.6080543Z Error (criu/config.c:238): Unsupported configuration file format. Please consult man page criu(8)
2023-02-23T18:03:42.6081414Z Error (criu/config.c:172): Unsupported configuration file format. Please consult man page criu(8)
2023-02-23T18:03:42.6082123Z Error (criu/config.c:238): Unsupported configuration file format. Please consult man page criu(8)
2023-02-23T18:03:42.6082876Z make[1]: Leaving directory '/home/runner/work/criu/criu'
...
2023-02-23T18:05:06.2180634Z + podman container restore cr
2023-02-23T18:05:06.6449222Z Error: OCI runtime error: CRIU restoring failed -52.  Please check CRIU logfile /var/lib/containers/storage/overlay-containers/8b660891c2e6de28d600f1468d4f4de0a2cfecd1d5c88645359bc4a1d5657ba2/userdata/restore.log
2023-02-23T18:05:06.6511420Z make: *** [Makefile:83: podman-test] Error 125
2023-02-23T18:05:06.6512145Z make: Leaving directory '/home/runner/work/criu
```

**Additional information you deem important (e.g. issue happens only occasionally):**
Happens on each PR.

Snorch:
@adrianreber It may be connected with bad criu configuration file put on this system:
```
 124                 if (input[offset] == 0) {
 125                         /* The value for the parameter was a single quote,      this is not supported. */
[...]

PR #2094, opened 2023-02-25
ci: disable c/r of cgroups with podman

rst0git:
This PR disables the checkpoint/restore of cgroups for the tests using Podman as a temporary workaround for https://github.com/checkpoint-restore/criu/issues/2091


adrianreber:
Works for me. Switching to ignore cgroups is anyway something that we want to do in Podman. The corresponding runc PR has been recently merged. If you are switching to runc you can remove the crun installation from the script.

adrianreber (approved):

PR #2095, opened 2023-02-27
dump: improve error printing and readability of task_comm_info

Snorch:
This addresses Andrei comments from
https://github.com/checkpoint-restore/criu/pull/2064

- Add comment about '\n' fixing
- Replace ret with more self explainting is_read
- Print warings if we failed to print comm for some reason


avagin:
LGTM

rst0git (approved):

PR #2098, opened 2023-02-28
pie/restorer: Fix fd leaking on error path

Snorch:
Nothing serious since OS will close it anyway but still to be precise.

It is a port of https://src.openvz.org/projects/OVZ/repos/criu/commits/459dfa2ca
CC: @cyrillos


adrianreber (approved):

rst0git (approved):
LGTM

PR #2102, opened 2023-02-28
net: fail restore if nftables isn't supported but image is present

mihalicyn:
Fixes: e1c4871 ("net: add nftables c/r")

### DIFF
--- a/criu/net.c
+++ b/criu/net.c
@@ -2391,29 +2391,18 @@ out:
 }
 
 #if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
-static inline int restore_nftables(int pid)
+static inline int do_restore_nftables(struct cr_img *img)
 {
-	int ret = -1;
-	struct cr_img *img;
+	int exit_code = -1;
 	struct nft_ctx *nft;
 	off_t img_data_size;
 	char *buf;
 
-	img = open_image(CR_FD_NFTABLES, O_RSTR, pid);
-	if (img == NULL)
-		return -1;
-	if (empty_image(img)) {
-		/* Backward compatibility */
-		pr_info("Skipping nft restore, no image\n");
-		ret = 0;
-		goto image_close_out;
-	}
-
 	if ((img_data_size = img_raw_size(img)) < 0)
-		goto image_close_out;
+		goto out;
 
 	if (read_img_str(img, &buf, img_data_size) < 0)
-		goto image_close_out;
+		goto out;
 
 	nft = nft_ctx_new(NFT_CTX_DEFAULT);
 	if (!nft)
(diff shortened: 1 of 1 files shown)


criu/net.c
```diff
@@ -2437,12 +2426,38 @@ static inline int restore_nftables(int pid)
...
 buf_free_out:
 	xfree(buf);
+out:
+	return ret;
+}
+#endif
+
+static inline int restore_nftables(int pid)
+{
+	int ret = -1;
```
Snorch, line 2436: nit: naming convention for variable which we return from current function is `exit_code`, please do same thing in do_restore_nftables too.

criu/net.c
```diff
@@ -2437,12 +2426,38 @@ static inline int restore_nftables(int pid)
...
+		return -1;
+	if (empty_image(img)) {
+		/* Backward compatibility */
+		pr_info("Skipping nft restore, no image\n");
+		ret = 0;
+		goto image_close_out;
+	}
+
+#if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
+	ret = do_restore_nftables(img);
```
Snorch, line 2450: Returning the return value of nested function is a bad practice, this way understanding allowed return values range of the function becomes harder, as you need to investigate nested functions recursively. I would advise:
```
if (!do_restore_nftables(img))
        exit_code = 0;
```
Meaning that any bad return from do_restore_nftables converts to -1 explicitly.

criu/net.c
mihalicyn: Have done.

criu/net.c
mihalicyn: Thanks! Have done.

Snorch (approved):

Snorch:
Vagrant Fedora based test (no VDSO) :
```
################### 1 TEST(S) FAILED (TOTAL 456/SKIPPED 35) ####################
 * zdtm/static/socket-tcp-reuseport(uns)
 
 ================= Run zdtm/static/socket-tcp-reuseport in uns ==================
Start test
./socket-tcp-reuseport --pidfile=socket-tcp-reuseport.pid --outfile=socket-tcp-reuseport.out
Run criu dump
Run criu restore
Send the 15 signal to  154
Wait for zdtm/static/socket-tcp-reuseport(154) to die for 0.100000
########## Test zdtm/static/socket-tcp-reuseport FAIL at result check ##########
22:14:02.186:     4: ERR: tcp.c:95: accept() failed (errno = 11 (Resource temporarily unavailable))
22:14:02.186:     4: ERR: tcp.c:95: accept() failed (errno = 11 (Resource temporarily unavailable))
22:14:02.186:     4: ERR: socket-tcp-reuseport.c:160: can't accept client connection 0
```
This CI error does not seem to be related to this patch.

PR #2103, opened 2023-03-01
coredump: two small fixes

adrianreber:
This fixes errors with long command-lines:
```
  File "/home/criu/coredump/criu_coredump/coredump.py", line 320, in _gen_prpsinfo
    prpsinfo.pr_psargs = self._gen_cmdline(pid)
    ^^^^^^^^^^^^^^^^^^
ValueError: bytes too long (88, maximum length 80)
```
and gives a cleaner error (without a complete Python backtrace) if files are missing:

New message:
```
 ERROR: Required file /usr/lib64/libcrypto.so.3.0.1 not found.
 Exiting
```
Old message:
```
   File "/home/criu/coredump/criu_coredump/coredump.py", line 693, in _gen_mem_chunk
     f = open(fname, 'rb')
 FileNotFoundError: [Errno 2] No such file or directory: '/usr/lib64/libcrypto.so.3.0.1'
```
Test has also been adapted verify that long command-lines do not break the coredump script.

### DIFF
--- a/coredump/coredump.py
+++ b/coredump/coredump.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import sys
 
 import criu_coredump
 
@@ -34,7 +35,12 @@ def main():
 
     opts = vars(parser.parse_args())
 
-    coredump(opts)
+    try:
+        coredump(opts)
+    except SystemExit as error:
+        print('ERROR: %s' % error)
+        print('Exiting')
+        sys.exit(1)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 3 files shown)


rst0git:
@adrianreber Would you be able to split this change into two commits?
(one that fixes the problem with long command-lines and another that handles the `SystemExit` exception)

adrianreber:
> @adrianreber Would you be able to split this change into two commits? (one that fixes the problem with long command-lines and another that handles the `SystemExit` exception)

commits have been split

PR #2108, opened 2023-03-06
Remove circleci configuration

h0lyalg0rithm:
Looking at the ci checks it seems like we are performing the same tests on both github workflow as well as circle ci.
https://github.com/checkpoint-restore/criu/blob/criu-dev/.github/workflows/x86-64-clang-test.yml
https://github.com/checkpoint-restore/criu/blob/criu-dev/.github/workflows/x86-64-gcc-test.yml.


adrianreber:
There are two checks related to try which do not work in GitHub actions. So we need circle CI for that at least.

h0lyalg0rithm:
Got it I wasnt aware about this.I will close the PR

PR #2109, opened 2023-03-07
pstree: fix shell job pgid replacement

Snorch:
Imagine situation entering prepare_pstree_for_shell_job we have:

  root_item->sid == 10
  root_item->pgid == 10
  current_sid == 20
  current_gid == 21

Then sid replacement loop would replace 10 to 20 everywhere in sids and pgids for all tasks including root_item so we have:

  root_item->sid == 20
  root_item->pgid == 20
  current_sid == 20
  current_gid == 21

Then pgid replacement loop would be entered and replace 20 to 21 everywhere in pgid and we would have:

  root_item->sid == 20
  root_item->pgid == 21

So 10,10 switched to 20,21, it does not feel right to me, better if restore would just ignore 21 in such case.

Alternatively we can just prohibit sid/pgid replacement `if ((root_item->sid == oot_item->pgid) != (current_sid == current_gid))`, but I believe it would be harder to use --shell-job option with this strict check.

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -397,7 +397,7 @@ static int prepare_pstree_for_shell_job(pid_t pid)
 		return 0;
 
 	old_gid = root_item->pgid;
-	if (old_gid != current_gid) {
+	if (old_gid != current_gid && old_gid != current_sid) {
 		pr_info("Migrating process tree (GID %d->%d)\n", old_gid, current_gid);
 
 		tmp = pstree_pid_by_virt(current_gid);


Snorch:
```
  CC       criu/mount.o
In file included from criu/include/cr_options.h:7,
                 from criu/mount.c:13:
In function '__list_add',
    inlined from 'list_add' at include/common/list.h:41:2,
    inlined from 'mnt_tree_for_each' at criu/mount.c:1977:2:
include/common/list.h:35:19: error: storing the address of local variable 'postpone' in '((struct list_head *)((char *)start + 8))[24].prev' [-Werror=dangling-pointer=]
   35 |         new->prev = prev;
      |         ~~~~~~~~~~^~~~~~
criu/mount.c: In function 'mnt_tree_for_each':
criu/mount.c:1972:19: note: 'postpone' declared here
 1972 |         LIST_HEAD(postpone);
      |                   ^~~~~~~~
include/common/list.h:23:42: note: in definition of macro 'LIST_HEAD'
   23 | #define LIST_HEAD(name) struct list_head name = LIST_HEAD_INIT(name)
      |                                          ^~~~
criu/mount.c:1972:19: note: 'start' declared here
 1972 |         LIST_HEAD(postpone);
      |                   ^~~~~~~~
include/common/list.h:23:42: note: in definition of macro 'LIST_HEAD'
   23 | #define LIST_HEAD(name) struct list_head name = LIST_HEAD_INIT(name)
      |                                          ^~~~
make[2]: *** [/criu/scripts/nmk/scripts/build.mk:119: criu/mount.o] Error 1
```
This is unrelated and should be fixed in terms of  https://github.com/checkpoint-restore/criu/pull/2110

criu/pstree.c
```diff
@@ -397,7 +397,7 @@ static int prepare_pstree_for_shell_job(pid_t pid)
 		return 0;
 
 	old_gid = root_item->pgid;
-	if (old_gid != current_gid) {
+	if (old_gid != current_gid && old_gid != current_sid) {
```
avagin, line 400: > So 10,10 switched to 20,21, it does not feel right to me, better if restore would just ignore 21 in such case.

This doesn't convince me. If I remember right, each shell job has own process group. It means that with this change, criu restores process in another shell job?

Snorch: Ok, looks like I was wrong -> will close. Thank you!

PR #2110, opened 2023-03-07
make: disable '-Wdangling-pointer' warning with gcc 12

Snorch:
The patch is similar to what has been done in linux kernel, as this warning effectively prevents us from adding list elements to local list head. See https://github.com/torvalds/linux/commit/49beadbd47c2

Note: it fixes CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,15 @@ DEFINES			+= -D_GNU_SOURCE
 
 WARNINGS		:= -Wall -Wformat-security -Wdeclaration-after-statement -Wstrict-prototypes
 
+# -Wdangling-pointer results in false warning when we add a list element to
+# local list head variable. It is false positive because before leaving the
+# function we always check that local list head variable is empty, thus
+# insuring that pointer to it is not dangling anywhere, but gcc can't
+# understand it.
+# Note: There is similar problem with kernel list, where this warning is also
+# disabled: https://github.com/torvalds/linux/commit/49beadbd47c2
+WARNINGS		+= -Wno-dangling-pointer -Wno-unknown-warning-option
+
 CFLAGS-GCOV		:= --coverage -fno-exceptions -fno-inline -fprofile-update=atomic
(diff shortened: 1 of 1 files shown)


Snorch:
V2 changes:
- check that `-Wno-dangling-pointer` exists before using it
- add compilation error to commit message

Makefile.config
```diff
@@ -54,6 +54,11 @@ else
         $(warning Warn: Building without nftables support)
 endif
 
+ifeq ($(call try-cc,$(FEATURE_TEST_NO_DANGLING_POINTER),-Wno-dangling-pointer),true)
```
avagin, line 57: pls add a comment that explains why -Wno-dangling-pointer should be set.

Snorch: done

adrianreber (approved):
I re-triggered the failing CI jobs and now everything looks good.

Makefile.config
```diff
@@ -54,6 +54,18 @@ else
...
+# -Wdangling-pointer results in false warning when we add a list element to
+# local list head variable. It is false positive because before leaving the
+# function we always check that local list head variable is empty, thus
+# insuring that pointer to it is not dangling anywhere, but gcc can't
+# understand it.
+# Note: There is similar problem with kernel list, where this warning is also
+# disabled: https://github.com/torvalds/linux/commit/49beadbd47c2
+ifeq ($(call try-cc,$(FEATURE_TEST_NO_DANGLING_POINTER),-Wno-dangling-pointer),true)
+        $(info Note: Building criu with -Wno-dangling-pointer.)
+        FEATURE_DEFINES += -Wno-dangling-pointer
```
avagin, line 66: I don't like the idea to use FEATURE_DEFINES for  anything else except defines. Here we can avoid try-cc by using -Wno-unknown-warning-option.

Snorch: Ok, I've updated PR accordingly.

PR #2106, opened 2023-03-06
util: fix code indent

rst0git:
This pull request fixes a code indentation issue introduced in https://github.com/checkpoint-restore/criu/pull/2075

PR #2111, opened 2023-03-07
ci/lint: disable 'make indent'

rst0git:
Following the discussions in https://github.com/checkpoint-restore/criu/pull/2044, https://github.com/checkpoint-restore/criu/pull/2035 and [[3]](https://github.com/checkpoint-restore/criu/pull/2106) we have decided that we should not blindly follow clang-format autoformat rules. In particular, if clang-format autoformat changes decrease code readability then we skip clang-format errors. Thus, in this patch we drop the `make indent` check to avoid false positive errors.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -19,17 +19,3 @@ jobs:
 
     - name: Run make lint
       run: make lint
-
-    - name: Run make indent
-      run: >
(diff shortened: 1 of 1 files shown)


Snorch:
This is also wrong see my comment in https://github.com/checkpoint-restore/criu/pull/2106#issuecomment-1457901266

PR #2112, opened 2023-03-07
CONTRIBUTING.md: document make lint / indent

rst0git:
This pull request documents how do we use `make lint` and `make indent` and adds a note about their integration with CI.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,6 +60,36 @@ When you change the source code, please keep in mind the following code conventi
 Other conventions can be learned from the source code itself. In short, make sure your new code
 looks similar to what is already there.
 
+The following command can be used to automatically run a code linter for Python files (flake8), Shell scripts (shellcheck),
+text spelling (codespell), and a number of CRIU-specific checks (usage of print macros and EOL whitespace for C files).
+
+```
+         make lint
+```
+
+In addition, we have adopted a [clang-format configuration file](https://www.kernel.org/doc/Documentation/process/clang-format.rst)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -60,6 +60,25 @@ When you change the source code, please keep in mind the following code conventi
...
+```
+
+In addition, we have adopted a [clang-format configuration file](https://www.kernel.org/doc/Documentation/process/clang-format.rst)
+based on the kernel source tree. However, compliance with the clang-format autoformat rules is optional. In cases where the use of
+clang-format to automatically format code results in reduced readability, we choose to ignore the errors generated by clang-format.
+
+Run the following command to check if your changes are compliant with the clang-format rules:
+
+```
+         make indent
```
Snorch, line 76: From commit message of new make indent option:
```
    This adds a BASE option for "make indent" where all commits in range
    BASE..HEAD would be checked with git-clang-format for codding style
    flaws. For instance when developing on top of criu-dev, one can use
    "make BASE=origin/criu-dev indent" to check all their commits for
    compliance with the clang-format rules. Default base is HEAD~1 to make
    last commit checked when "make indent" is called. The closest thing to
    the old behaviour would then be "make indent BASE=init", note that only
    commited files would be checked.
    
    Extra options to git-clang-format may be passed through OPTS variable.
```
we can also explain these options, maybe just give an example of how to use them, I tend to use a lot `make indent OPTS=--diff BASE=HEAD~N` to show me clang-format errors in last N commits without modifying the codebase.

Snorch (commented):
@rst0git Thank you for doing this documentation! I've added some comments with which we can extend your explanation further.

Snorch (approved):

Issue #1171, opened 2020-08-06
Live migration of an application result in lower performance


planetA:
Hello,

I have quite a weird performance issue. 

I have an MPI multiprocess application running on two nodes. The application consists of 4 processes (ranks). Each node starts with 2 processes. Soon, I live-migrate one of the processes to another node. This results in significant performance degradation that cannot be explained by the migration time. For example, if a normal run takes around 35 seconds, with migration, the program runs for around 65 seconds. The migration time is less than 2 seconds.

If I start the application with 3 processes on one node from the very beginning, the runtime is still around 35 seconds (I have 4 cores per node). This indicates that the problem is not about resource exhaustion.

I found that the problem is somehow connected to the usage of vector instructions. To confirm my hypotheses I was running `perf stat` repeatedly for 200 ms while the application was running. Here is a typical sample after one of the processes migrates:

```
Performance counter stats for process id '154166':

       775,326,343      cycles                                                        (32.58%)
       611,253,031      instructions              #    0.79  insn per cycle           (42.54%)
        61,378,465      branch-instructions                                           (44.53%)
               156      avx_insts.all                                                 (46.51%)
                 0      other_assists.avx_to_sse                                      (33.57%)
                 0      other_assists.sse_to_avx                                      (31.92%)

       0.200672137 seconds time elapsed


 Performance counter stats for process id '154165':

       775,240,259      cycles                                                        (32.99%)
       616,100,734      instructions              #    0.79  insn per cycle           (42.95%)
        61,929,959      branch-instructions                                           (44.95%)
               127      avx_insts.all                                                 (46.93%)
                 0      other_assists.avx_to_sse                                      (33.15%)
                 0      other_assists.sse_to_avx                                      (31.91%)

       0.200609032 seconds time elapsed


 Performance counter stats for process id '154234':

       774,536,014      cycles                                                        (33.41%)
       324,489,301      instructions              #    0.42  insn per cycle           (43.38%)
        32,593,842      branch-instructions                                           (45.37%)
         4,951,097      avx_insts.all                                                 (47.36%)
         2,481,913      other_assists.avx_to_sse                                      (32.71%)
         2,487,357      other_assists.sse_to_avx                                      (31.90%)

       0.200649176 seconds time elapsed
```

[...]

PR #2113, opened 2023-03-07
x86/xsave: Set only used XFEATURE_* in xstate_bv

mihalicyn:
Just rebasing on top of `criu-dev`.

All credit to Dmitry @0x7f454c46 for this incredible investigation and fix. (-:
```
Setting all supported by CPU features in xstate_bv may bring it into dirty-upper-state as documented in specs, resulting in lower performance. Let's not do this and set only those have been used by dumpee.

P.S.
Off course it has to be a one-liner!

Fixes: #1171
```

Fixes: #1171


0x7f454c46:
LGTM, though, it would be good to have a test.
But having in mind that I didn't create it and as a result, was avoiding preparing a pull request for 2 years, probably it can be merged without a test :-D

0x7f454c46 (approved):

Issue #1961, opened 2022-08-23
Checkpointing fails for a process with "uprobes" VMA


mironov-stripe:
**Description**
Hey folks,

Whilst trying out CRUI today I noticed that docker container checkpointing fails for (containerized) processes with `[uprobes]` VMA's.  I was able to "mitigate" the issue by making this area known to CRIU and skipping the `madivse()` bits on restore. I found [this piece of code](https://github.com/checkpoint-restore/criu/blob/ce1b705b157019c873fba33a5b427bafa2af356c/criu/cr-check.c#L686) mentioning uprobes though it's commented-out and the original commit doesn't say much about it.

**Steps to reproduce the issue:**
1.  A machine with `CONFIG_UPROBE_EVENTS=y` (and/or BPF) enabled and some global uprobes attached (e.g. through https://osquery.io/)
2. `docker checkpoint create CONTAINER_ID CHECKPOINT_NAME`

**Describe the results you received:**
```
Error response from daemon: Cannot checkpoint container CHECKPOINT_NAME: runc did not terminate successfully: criu failed: type NOTIFY errno 0 path= /run/containerd/io.containerd.runtime.v2.task/moby/<SHA>/criu-dump.log: unknown
```

**Describe the results you expected:**
Successful checkpoint.

**CRIU logs and information:**
From `criu-dump.log`:

```
(00.026560) Collecting mappings (pid: 123)
....
(00.028257) Error (criu/proc_parse.c:703): Unsupported mapping found 00007fffffffe000-00007ffffffff000
(00.028266) Error (criu/cr-dump.c:1524): Collect mappings (pid: 123) failed with -1
```

Looking at `/proc/123/smaps`:
```
7fffffffe000-7ffffffff000 --xp 00000000 00:00 0                          [uprobes]
Size:                  4 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:                   4 kB
Pss:                   4 kB
Shared_Clean:          0 kB
Shared_Dirty:          0 kB
Private_Clean:         4 kB
Private_Dirty:         0 kB
Referenced:            4 kB
Anonymous:             0 kB
LazyFree:              0 kB
AnonHugePages:         0 kB
ShmemPmdMapped:        0 kB
FilePmdMapped:         0 kB
Shared_Hugetlb:        0 kB
Private_Hugetlb:       0 kB
Swap:                  0 kB
SwapPss:               0 kB
Locked:                0 kB
THPeligible:    0
[...]

PR #2115, opened 2023-03-08
Handle processes with uprobes

mihalicyn:
Linux kernel uprobes (userspace probes) subsystem, uses special
type of VMAs, so-called XOL (execution out of line) mappings.
( see ("uprobes/core: Allocate XOL slots for uprobes use") https://github.com/torvalds/linux/commit/d4b3b6384f98f8692ad0209891ccdbc7e78bbefe )
This mappings are service mappings for internal uprobes mechanism needs,
it stores the original instruction to execute.

uprobes mechanism works by replacing the original instruction in the
tracee's address space by some kind of a breakpoint [see set_swbp() in
kernel/events/uprobes.c].

To achieve that, uprobes modifies existing code mappings (VMAs),
and replaces the page, that corresponds to a tracable virtual address
by a new temporary page [see uprobe_write_opcode()].

CRIU ignores such code modifications because executable code lies in
a file mappings, but CRIU does nothing with the file mappings contents.

It means that after C/R we are running the original processes code,
without uprobes traps. Therefore, we can safely drop XOL mapping too,
because it's only needed when uprobe tracing is enabled, and used
only during the short period of time when uprobe trap is handled in the kernel.

Fixes: #1961

### DIFF
--- a/criu/include/image.h
+++ b/criu/include/image.h
@@ -60,6 +60,11 @@
  *  	memory map for socket
  *  - AIO ring
  *  	memory area serves AIO buffers
+ *  - Uprobes
+ *  	XOL (execution out of line) memory area installed to the
+ *  	process mm_struct by the kernel for internal uprobes needs.
+ *  	CRIU skips it on restore, but we save all metadata about it
+ *  	to the CRIU images just in case.
  *  - unsupported
  *  	stands for any unknown memory areas, usually means
  *  	we don't know how to work with it and should stop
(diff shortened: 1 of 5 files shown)


avagin:
> CRIU ignores such code modifications because executable code lies in a file mappings, but CRIU does nothing with the file mappings contents.

The code segment is mapped as MAP_FILE | MAP_PRIVATE. It means CRIU will dump all changed pages.

mihalicyn:
> > CRIU ignores such code modifications because executable code lies in a file mappings, but CRIU does nothing with the file mappings contents.
> 
> The code segment is mapped as MAP_FILE | MAP_PRIVATE. It means CRIU will dump all changed pages.

my bad. Thanks ;-) Will choose another way to support this.

mihalicyn:
I'll reopen that after rework :)

PR #2118, opened 2023-03-12
ci: cancel preceding workflows run

rst0git:
This pull request adds concurrency groups to the CI workflows to automatically cancel any in-progress workflows when a pull request has been updated. A [`concurrency`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency) group allows to ensure that a single job or workflow will run at a time. For example, when a pull request is updated with a force-push, the GiHub CI workflows currently in-progress will be automatically cancelled, and the CI would run only with the updated commits.

### DIFF
--- a/.github/workflows/alpine-test.yml
+++ b/.github/workflows/alpine-test.yml
@@ -2,6 +2,11 @@ name: Alpine Test
 
 on: [push, pull_request]
 
+# Cancel any preceding run on the pull request.
+concurrency:
+  group: alpine-test-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/criu-dev' }}
+
 jobs:
   build:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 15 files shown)


avagin:
The idea looks good to me, but I can't parse this githabish yaml things. @adrianreber could you take a look at this pr.

adrianreber:
@rst0git if this get merged into `master`, do we get a problem because of `criu-dev` being mentioned in the concurrency rules?

rst0git:
> if this get merged into master, do we get a problem because of criu-dev being mentioned in the concurrency rules?

No. The condition `cancel-in-progress: ${{ github.ref != 'refs/heads/criu-dev' }}` means that for any branch other than criu-dev,  the currently running job or workflow in the same concurrency group would be cancelled. In other words, in-progress jobs or workflows would not be cancelled for the criu-dev branch. For example, when two pull requests are merged one after another to the criu-dev branch, the CI tests would run for both pull requests.

Since we don't update the master branch very often, I assume that it is not necessary to include it in this condition.

Issue #332, opened 2017-06-29
Fixup scanf formats vs variables' types


xemul:
https://www.codacy.com/app/xemul/criu/issues?bid=4795042&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W251bGxdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOlsiQ29kZSBTdHlsZSJdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltudWxsXX0seyJpZCI6IlBhdHRlcm4iLCJ2YWx1ZXMiOlsyMjM5XX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W251bGxdfSx7InZhbHVlcyI6W119XQ==

mihalicyn:
https://app.codacy.com/gh/mihalicyn/criu/issues?&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W119LHsiaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6WyJFcnJvclByb25lIl19LHsiaWQiOiJMZXZlbCIsInZhbHVlcyI6W119LHsiaWQiOiJQYXR0ZXJuIiwidmFsdWVzIjpbIjIyMzkiXX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W119XQ==

mihalicyn:
https://app.codacy.com/gh/mihalicyn/criu/issues?&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W119LHsiaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6WyJFcnJvclByb25lIl19LHsiaWQiOiJMZXZlbCIsInZhbHVlcyI6W119LHsiaWQiOiJQYXR0ZXJuIiwidmFsdWVzIjpbIjIxNTIiXX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W119XQ==

mihalicyn:
[...]

PR #2122, opened 2023-03-13
Fixed scanf format vs variable type mismatch

Rajiv2605:
@mihalicyn This PR addresses the issue #332 .

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -195,7 +195,7 @@ static int kerndat_get_dev(dev_t *dev, char *map, size_t size)
 	char maps[128];
 	struct stat buf;
 
-	sprintf(maps, "/proc/self/map_files/%lx-%zu", (unsigned long)map, (unsigned long)map + size);
+	sprintf(maps, "/proc/self/map_files/%lx-%lx", (unsigned long)map, (unsigned long)map + (unsigned long)size);
 	if (stat(maps, &buf) < 0) {
 		int e = errno;
 		if (errno == EPERM) {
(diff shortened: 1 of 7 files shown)


Rajiv2605:
Sorry I forgot to sign the commits before pushing them. Is there a way to fix this? Also, I am not able to understand why test-local-gcc tests are failing.

criu/proc_parse.c
```diff
@@ -1190,7 +1190,7 @@ struct opt2flag {
 
 static bool sb_opt_cb(char *opt, char *unknown, size_t *uoff)
 {
-	unsigned int id;
+	int id;
```
avagin, line 1193: If you change a variable type, you need to create a separate patch and explain in a commit message why you think it has to be done. In this case, I don't think the change is correct. uid and gid are uint32.

Rajiv2605: Instead of typecasting, I thought I will change the variable type of id since it is a local variable and that was the mistake. I will use typecasting instead.

criu/kerndat.c
```diff
@@ -195,7 +195,7 @@ static int kerndat_get_dev(dev_t *dev, char *map, size_t size)
 	char maps[128];
 	struct stat buf;
 
-	sprintf(maps, "/proc/self/map_files/%lx-%lx", (unsigned long)map, (unsigned long)map + size);
+	sprintf(maps, "/proc/self/map_files/%lx-%zu", (unsigned long)map, (unsigned long)map + size);
```
avagin: How do you test your changes?

avagin:
You need to reformat patches. "git rebase -i" can be useful for this. Each PR has to have a clean set of change. Each separate change should not break any tests. If you find an issue in one of existing patch, you should not create a separate commit for the fix, you need to rework in the target patch.

Issue #330, opened 2017-06-29
Optimize shell code with <'s (instead of cat + |)


xemul:
https://www.codacy.com/app/xemul/criu/issues?bid=4795042&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W251bGxdfSx7ImlkIjoiQ2F0ZWdvcnkiLCJ2YWx1ZXMiOlsiQ29kZSBTdHlsZSJdfSx7ImlkIjoiTGV2ZWwiLCJ2YWx1ZXMiOltudWxsXX0seyJpZCI6IlBhdHRlcm4iLCJ2YWx1ZXMiOlsyMzk1XX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W251bGxdfSx7InZhbHVlcyI6W119XQ==

sankalp-12:
Could I take up this issue?

PR #2125, opened 2023-03-15
Optimized shell code with <'s (instead of cat + |)

sankalp-12:
This PR optimizes shell code as reading a single file as input using a 'cat' command to a program is considered to be a Useless Use of Cat [[UUOC](https://en.wikipedia.org/wiki/Cat_(Unix)#Useless_use_of_cat)]. It's more efficient to simply use redirection. Fixes the issue https://github.com/checkpoint-restore/criu/issues/330.

### DIFF
--- a/test/jenkins/criu-fault.sh
+++ b/test/jenkins/criu-fault.sh
@@ -9,7 +9,7 @@ prep
 ./test/zdtm.py run -t zdtm/static/maps00 --fault 3 --report report -f h || fail
 
 # FIXME: fhandles looks broken on btrfs
-cat /proc/self/mountinfo | grep -P "/.* / " | grep -q btrfs || NOBTRFS=$?
+grep -P "/.* / " /proc/self/mountinfo | grep -q btrfs || NOBTRFS=$?
 if [ $NOBTRFS -eq 1 ] ; then
 	./test/zdtm.py run -t zdtm/static/inotify_irmap --fault 128 --pre 2 -f uns || fail
 fi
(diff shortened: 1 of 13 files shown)


test/jenkins/criu-fault.sh
```diff
@@ -9,7 +9,7 @@ prep
 ./test/zdtm.py run -t zdtm/static/maps00 --fault 3 --report report -f h || fail
 
 # FIXME: fhandles looks broken on btrfs
-cat /proc/self/mountinfo | grep -P "/.* / " | grep -q btrfs || NOBTRFS=$?
+< /proc/self/mountinfo grep -P "/.* / " | grep -q btrfs || NOBTRFS=$?
```
adrianreber, line 12: Things like this can be changed to `grep -P "/.* / " /proc/self/mountinfo`. The extra `<` seems unnecessary in this case. There are couple of cases here in this PR like this. Not all of them.

sankalp-12: It seems that most of the cases in this PR have the extra redirection operator. I have incorporated them in the latest commit.

adrianreber:
Please squash the commits. We try to avoid fixup commits in PRs.

adrianreber:
I think you can leave out the Fixes line from the commit message and also try to ensure that the commit message has line breaks. If it is not for URLs or error messages we usually try to break lines around 80 characters.

The one failing CI run seems unrelated to your change.

rst0git:
This blog post has a good description on how to format the commit message: https://cbea.ms/git-commit/

adrianreber (approved):
Looks good.

PR #2128, opened 2023-03-15
cgroup-v2: make new fields in image protobuf optional

minhbq-99:
The newly added fields in image protobuf are currently marked as required. This causes backward compatibility problem when using the newer CRIU version to restore an dumped image from older version which doesn't have these fields. This commit marks these new fields optional and reworks the logic:
- Fallback to cg_set in task_core when it is not in thread_core
- Skip fixing up threaded cgroup controllers if there is no information in dumped image


adrianreber:
Thanks for the quick PR!

Snorch:
Please put fixes for is_threaded and cg_set into separate patches.

PR #2130, opened 2023-03-16
pstree: improve id intersection detection in prepare_pstree_for_shell_job

Snorch:
First, let's move lookup_create_item-s to the end so that on pgid replacement we don't have false positive pstree_pid_by_virt check founding item created by sid replacement. (note: we need those lookup_create_item-s for the sake of free pid selection mechanism)

Second, let's add checks for sid/pgid in images intersecting with current_sid/pgid, as this would also bring problems on restore.

(second part is similar to what we already do in Virtuozzo criu here https://src.openvz.org/projects/OVZ/repos/criu/commits/23448a763)

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -381,20 +381,26 @@ static int prepare_pstree_for_shell_job(pid_t pid)
 		}
 
 		for_each_pstree_item(pi) {
+			if (pi->sid == current_sid) {
+				pr_err("Current sid %d intersects with sid of (%d) in images\n", current_sid, vpid(pi));
+				return -1;
+			}
 			if (pi->sid == old_sid)
 				pi->sid = current_sid;
 
+			if (pi->pgid == current_sid) {
+				pr_err("Current sid %d intersects with pgid of (%d) in images\n", current_sid,
+				       vpid(pi));
+				return -1;
+			}
 			if (pi->pgid == old_sid)
 				pi->pgid = current_sid;
 		}
-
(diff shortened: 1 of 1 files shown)


criu/pstree.c
```diff
@@ -381,20 +381,26 @@ static int prepare_pstree_for_shell_job(pid_t pid)
...
 			if (pi->pgid == old_sid)
 				pi->pgid = current_sid;
 		}
-
-		if (lookup_create_item(current_sid) == NULL)
-			return -1;
 	}
 
 	/* root_item is a group leader */
 	if (root_item->pgid == vpid(root_item))
```
mihalicyn, line 402: I'm not sure, but don't we want to check this condition `root_item->pgid == vpid(root_item)` before altering `->pgid` in the previous `for_each_pstree_item` loop? Can it happen, so we change `->pgid` during previous traverse over pstree and then this condition becomes false but in fact it was true?

Snorch: > Can it happen, so we change ->pgid during previous traverse over pstree and then this condition becomes false but in fact it was true?

It can not. If we had changed root_item->pgid during previous traversal it means that it was equal to root_item->sid, and if the condition (root_item->pid == root_item->pgid) was also true it means that root_item->pid == root_item->pgid
== root_item->sid and we would not get here.

And vice versa, If the condition (root_item->pid == root_item->pgid) was not tr
ue it would not become true even pgid had been replaced due to pstree_pid_by_virt check before replacement.

mihalicyn (approved):

PR #2133, opened 2023-03-18
Codacy : Conditions always evaluate to T/F

TaeminHa:
#### criu/appamor.c line 210 (DELETED)
    - if (sb1.st_mtim.tv_sec == sb2.st_mtim.tv_sec)
    - This check is redundant as line 201 checks for this condition. 
    
#### criu/arch/mips/cpu.c line 43 & 45 (UNMODIFIED)
    - if (cpu_init())
    - if (cpu_dump_cpuinfo())
    - These conditions are always false as cpu_init(void) and cpu_dump_cpuinfo() unconditionally returns 0.
    - Left unmodified as it's not clear if cpu_init(void) and cpu_dump_cpuinfo() are function stubs.
 
#### criu/arch/x86/sigframe.c line 26 (MODIFIED)
    - } else if (!sigframe->is_native) {
    - The is_native field is a boolean. Therefore, else if() should can be changed to a simple else{}.

#### criu/kerndat.c line 900 (UNMODIFIED)
    - if (ret < 0)
    - ret is the return value from kdat_x86_has_ptrace_fpu_xsave_bug(void), which unconditionally returns 0.
    - Left unmodified as it's not clear if kdat_x86_has_ptrace_fpu_xsave_bug is a function stub.

#### criu/kerndat.c line 1161 (MODIFIED)
    - if (!ret && kerndat_loginuid()) {
    - ret initially 0, and the first modification happens inside the if statement. Therefore, we can delete !ret

#### criu/mem.c line 333 (UNMODIFIED)
    - if (tps == -1) {
    - tps retrives clock ticks through sysconf(_SC_CLK_TCK).
    - The condition logically evaluates to false, but is necessary as a safety check.

#### criu/page-xfer.c line 818 (MODIFIED)
    - if (ret == -1 || ret != bytes_read) {
    - ret is the return value from syscall vmsplice(), which returns -1 upon error.
    - However, it's modified to if (ret < 0...) to be consistent with the codebase (line 183 of pipe.c, line 466 of cr-check.c, line 90 of parasite.c)

#### criu/proc_parse.c line 1986 (MODIFIED)
    - else if (ret < 6)
    - In the case where ret == 3 || ret == 6, EventpollTfdEntry's fields are set to false or true.
    - The remaining cases could be grouped into an else statement instead of manually checking the values between 3 and 6.

#### test/zdtm/static/cow00.c line 32 (MODIFIED)
    - if (fd1 < 0) {
    - The condition meant to check fd2 instead of fd1, which is checked in line 24.

#### test/zdtm/static/maps03.c line 38 (UNMODIFIED)
    - if (mem[4L << 30] != 1 || mem[8L << 30] != 2) {
    - This condition checks for possible memory corruption. Logically, it should be false, but is necessary to be here.

#### test/zdtm/static/thread_different_uid_gid.c line 133 (MODIFIED)
    - } else if (ret < 0) {
    - line 131 checks if (ret >= 0). line 133 could be replaced by a simple else statem

Signed-off-by : Taemin Ha taemin.ha@utexas.edu

### DIFF
--- a/criu/apparmor.c
+++ b/criu/apparmor.c
@@ -207,8 +207,6 @@ static int by_time(const struct dirent **de1, const struct dirent **de2)
 	} else {
 		if (sb1.st_mtim.tv_sec < sb2.st_mtim.tv_sec)
 			return -1;
-		if (sb1.st_mtim.tv_sec == sb2.st_mtim.tv_sec)
-			return 0;
 		return 1;
 	}
 }
--- a/criu/arch/x86/sigframe.c
+++ b/criu/arch/x86/sigframe.c
@@ -23,7 +23,7 @@ int sigreturn_prep_fpu_frame(struct rt_sigframe *sigframe, struct rt_sigframe *r
 		}
 
 		sigframe->native.uc.uc_mcontext.fpstate = (uint64_t)addr;
-	} else if (!sigframe->is_native) {
+	} else {
 		unsigned long addr = (unsigned long)(void *)&fpu_state->fpu_state_ia32.xsave;
 		sigframe->compat.uc.uc_mcontext.fpstate = (uint32_t)(unsigned long)(void *)&fpu_state->fpu_state_ia32;
 		if ((addr % 64ul)) {
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1608,7 +1608,7 @@ static int root_only_init(void)
 	if (opts.unprivileged)
 		return 0;
 
-	if (!ret && kerndat_loginuid()) {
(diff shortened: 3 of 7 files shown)


adrianreber:
Thanks for the PR.

My recommendation would be to have each change, as they seem independent of each other, be a separate commit and include the description from the PR in the commit.

Having the information in the commit makes it easier to understand the change in the future as the PR description is not part of the git repository. It is only available in GitHub.

TaeminHa:
Thank you for the feedback! I'll separate these out into multiple commits and draft a new PR.

PR #2137, opened 2023-03-23
zdtm/lib: fix cwd path freeing

Snorch:
Fix cwd freeing on error path in get_cwd_check_perm and on non-error-path in unix_fill_sock_name.

This is a port of Virtuozzo patch:
https://src.openvz.org/projects/OVZ/repos/criu/commits/a7916411a

CC: @uravas

### DIFF
--- a/test/zdtm/lib/fs.c
+++ b/test/zdtm/lib/fs.c
@@ -108,6 +108,7 @@ int get_cwd_check_perm(char **result)
 		       "Bit 'x' should be set in all path components of "
 		       "this directory\n",
 		       cwd, getuid(), getgid(), errno, strerror(errno));
+		free(cwd);
 		return -1;
 	}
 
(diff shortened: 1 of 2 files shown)


adrianreber:
Looks good. There would also be `cleanup_free` to get this behaviour automatically.

Snorch:
@adrianreber I always forget about `cleanup_free`, thanks for pointing that out!

But note that I left simple free variant in get_cwd_check_perm, as it does not feel right to me using `cleanup_free` in function which returns allocated data up the stack.

adrianreber:
> But note that I left simple free variant in get_cwd_check_perm, as it does not feel right to me using `cleanup_free` in function which returns allocated data up the stack.

Sure, in that case `cleanup_free` does not sound useful.

adrianreber (approved):

Issue #350, opened 2017-06-30
Distinguish criu dump fail from criu dump crash in zdtm


xemul:
In zdtm .desc files there can be a flag called "crfail". It means, that the dump of this test should fail, we know it and are willing to fix. Unfortunately, the zdtm.py script doesn't tell criu dump command failure from crash (for sigsegv, for example). While the former result is what's expected, the latter one should be reported as test failure.

sankalp-12:
I would like to take up this issue. Is the fix just adding a 'crfail' tag to every ZDTM ,desc file?

mihalicyn:
> I would like to take up this issue. Is the fix just adding a 'crfail' flag to every ZDTM ,desc file?

nope. It's about distinguishing two cases when `criu dump` process exits with an error and when it crashes.
When `crfail` flag is set on test, and `criu dump` exits with an error that's correct (test is passed), but if `criu dump` crashes, currently, zdtm will report that test is passed, but in fact is should report a test failure.

sankalp-12:
In the `do_run_tests()` function in the `zdtm.py` script file, when the `dump` action fails but the `crfail` flag is set, then the test stops. Adding a *try and except* block in the `cr(cr_api, test, opts)` (CRIU API) function call to handle signal errors is where I'm currently at.

```
 if opts['norst']:
            try_run_hook(test, ["--pre-dump"])
            try:
                cr_api.dump("dump", opts=["--leave-running"])
            except subprocess.CalledProcessError as e:
                if e.returncode < 0:
                    raise test_fail_exc("'criu dump' crashed with signal %d" % -e.returncode)
                else:
                    raise test_fail_exc("'criu dump' failed with return code %d" % e.returncode)
        else:
            try_run_hook(test, ["--pre-dump"])
            try:
                cr_api.dump("dump")
            except subprocess.CalledProcessError as e:
                if e.returncode < 0:
                    raise test_fail_exc("'criu dump' crashed with signal %d" % -e.returncode)
                else:
                    raise test_fail_exc("'criu dump' failed with return code %d" % e.returncode)
``` 

Should this work?

mihalicyn:
@sankalp-12

you can easily test it by adding bug in CRIU code and run some `crfail` test ;-)

AFAIU you need to pay attention to this piece of code:
```
            rst_succeeded = os.access(
                os.path.join(__ddir, "restore-succeeded"), os.F_OK)
            if self.__test.blocking() or (self.__sat and action == 'restore' and
[...]

PR #2138, opened 2023-03-24
Distinguish CRIU crash from CRIU fail in ZDTM

sankalp-12:
This PR distinguishes `criu dump` crash from `criu dump` fail in the `zdtm.py` script. Now, *zdtm* reports `criu dump` crash as a test failure. 
Fixes the issue https://github.com/checkpoint-restore/criu/issues/350.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -900,6 +900,10 @@ class criu_cli:
             return cr
         return cr.wait()
 
+    @staticmethod
+    def signal_exit(ret):
+        return ret < 0
+
 
 class criu_rpc_process:
     def wait(self):
@@ -1031,6 +1035,11 @@ class criu_rpc:
             return p
 
         return ret
+
+    @staticmethod
+    def signal_exit(ret):
+        # TODO
+        return False
 
 
 class criu:
@@ -1233,8 +1242,7 @@ class criu:
                     return
             rst_succeeded = os.access(
                 os.path.join(__ddir, "restore-succeeded"), os.F_OK)
-            if self.__test.blocking() or (self.__sat and action == 'restore' and
-                                          rst_succeeded):
+            if (self.__test.blocking() and not self.__criu.signal_exit(ret)) or (self.__sat and action == 'restore' and rst_succeeded):
                 raise test_fail_expected_exc(action)
             else:
                 raise test_fail_exc("CRIU %s" % action)
--- a/test/zdtm/criu_config.py
+++ b/test/zdtm/criu_config.py
@@ -40,3 +40,7 @@ class criu_config:
         if nowait:
             return cr
         return cr.wait()
+
+    @staticmethod
+    def signal_exit(ret):
+        return ret < 0


test/zdtm.py
```diff
@@ -1233,8 +1233,7 @@ def __criu_act(self, action, opts=[], log=None, nowait=False):
                     return
             rst_succeeded = os.access(
                 os.path.join(__ddir, "restore-succeeded"), os.F_OK)
-            if self.__test.blocking() or (self.__sat and action == 'restore' and
-                                          rst_succeeded):
+            if (self.__test.blocking() and ret > 0) or (self.__sat and action == 'restore' and rst_succeeded):
```
mihalicyn, line 1236: Looks correct. The only question is the case when `self.__criu == criu_rpc`, the `run` static method in this case can return negative value for any error, not related to SEGFAULT / SIGBUS. I can suggest to add a new static method (name it like `signal_exit`?) to `criu_rpc` and `criu_cli` to detect "signal exit" case. You can implement this method for `criu_cli` properly and for `criu_rpc` just always return `False` (and mark as `TODO`) for now.

sankalp-12: Done. Thanks for the suggestion!

test/zdtm.py
```diff
@@ -1233,11 +1245,13 @@ def __criu_act(self, action, opts=[], log=None, nowait=False):
...
                 raise test_fail_expected_exc(action)
             else:
-                raise test_fail_exc("CRIU %s" % action)
+                if ((self.__criu == criu_cli and not self.__criu.signal_exit(ret)) or (self.__criu == criu_rpc and not self.__criu.signal_exit(ret))):
```
mihalicyn, line 1251: this hunk looks strange. Why you need to check `self.__criu == criu_cli`?

sankalp-12: I checked it just as a precaution. Now, that I think about it, I don't think `criu_cli` would ever return a negative return code  for a non-signal exit.

test/zdtm.py
mihalicyn: my idea was to move `ret > 0` check into `signal_exit` method. So, `if (self.__test.blocking() and ret > 0) or (self.__sat and action == 'restore' and rst_succeeded):` should be like `if (self.__test.blocking() and !self.__criu.signal_exit(ret)) or (self.__sat and action == 'restore' and rst_succeeded):`.

```
class criu_cli:
...
    @staticmethod
    def signal_exit(ret):
        return ret < 0
```

Reason is to keep all class-specific details in the class itself.

mihalicyn (changes requested):

test/zdtm.py
sankalp-12: I will draft a new PR with all the changes. Closing this one.

mihalicyn: you don't need to create a new PR every time. You can use `git rebase -i HEAD~N`, `git commit --amend` and `git push -f` to fix your changes.

sankalp-12: The new PR is at https://github.com/checkpoint-restore/criu/pull/2140. Okay, I'll try not drafting a new one unnecessarily now onwards.

PR #2139, opened 2023-03-26
feat: Add distinction between dump failure and dump crash in zdtm.py

### LINKED ISSUE #350: Distinguish criu dump fail from criu dump crash in zdtm

linktorahulraj:

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1508,6 +1508,18 @@ class criu:
         if not os.access(opts['criu_bin'], os.X_OK):
             print("CRIU binary not found at %s" % opts['criu_bin'])
             sys.exit(1)
+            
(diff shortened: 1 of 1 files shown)


linktorahulraj:
@mihalicyn please check this.

mihalicyn:
we already have similar PR from another student https://github.com/checkpoint-restore/criu/pull/2138

have you tested your changes? As far as I can see the code that you've added shouldn't work.

PR #2140, opened 2023-03-27
Distinguish criu fail from criu crash in ZDTM

sankalp-12:
This PR distinguishes `criu dump` crash from `criu dump` fail in the `zdtm.py` script. Now, *zdtm* reports `criu dump` crash as a test failure. 
Fixes the issue https://github.com/checkpoint-restore/criu/issues/350.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -900,6 +900,10 @@ class criu_cli:
             return cr
         return cr.wait()
 
+    @staticmethod
+    def signal_exit(ret):
+        return ret < 0
+
 
 class criu_rpc_process:
     def wait(self):
@@ -1018,8 +1022,11 @@ class criu_rpc:
             else:
                 raise test_fail_exc('RPC for %s required' % action)
         except crpc.CRIUExceptionExternal as e:
-            print("Fail", e)
-            ret = -1
+            if e.errno == errno.EINTR:
+                ret = -2
+            else:
+                print("Fail", e)
+                ret = -1
         else:
             ret = 0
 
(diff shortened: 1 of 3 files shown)


mihalicyn:
@sankalp-12 I think you are close to the final, but please read through the https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md and try to fix your commit according to it:
- commit message header like `zdtm: ...`
- Signed-off-by tag with *real* name

test/zdtm.py
```diff
@@ -1032,6 +1036,11 @@ def run(action,
 
         return ret
 
+    @staticmethod
+    def signal_exit(ret):
+        # TODO
```
avagin, line 1041: TODO should contain a comment. Why can't we implement this method for the rpc?

sankalp-12: @avagin AFAIU, `return ret < 0` can't be used here as the `wait()` function of the `criu_rpc` class sets `ret` as negative even in the case of non-Signal exits, as given in the `CRIUExceptionExternal` class in the `test/pycriu/criu.py` file. This would lead to false positives.

sankalp-12: @avagin An instance of the `CRIUExceptionExternal` class always has an *errno* attribute which is the error code returned by the process using the `errno` module in Python. One of these error codes is `EINTR: Interrupted system call` which is raised when a system call is interrupted by a signal. AFAIU, this should work, I would be glad to here any better solutions or alternatives :)

avagin: Have you tested your change? If the answer is yes, pls describe the process.

sankalp-12: No, I haven't been able to successfully run the `zdtm` tests.

PR #2146, opened 2023-03-30
netlink: fix netlink fd flags dump/restore failed

hdzhoujie:
During the restore process, netlink fd uses the flags in the NetlinkSkEntry structure to restore the file state, but during the dump process, the flags values is not saved to the structure.

Signed-off-by: zhoujie zhoujie133@huawei.com
Signed-off-by: hejingxian hejingxian@huawei.com

### DIFF
--- a/criu/sk-netlink.c
+++ b/criu/sk-netlink.c
@@ -161,7 +161,7 @@ static int dump_one_netlink_fd(int lfd, u32 id, const struct fd_parms *p)
 
 		ne.protocol = val;
 	}
-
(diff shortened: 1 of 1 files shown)


avagin:
@hdzhoujie the patch looks good to me. Could you wrap lines of the commit message at 72 characters? Thanks.

rst0git (approved):
LGTM

@hdzhoujie I've updated the commit message and the format of the Signed-off-by lines.

PR #2147, opened 2023-03-30
cr-dump: properly apply rseq fixup for all threads

mclapinski:
Previously fixup was done before threads' registers were dumped so it
didn't actually work. This commit splits rseq fixup into thread leader
fixup and other threads fixup and applies them after the entities are
seized.

### DIFF
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -906,6 +906,7 @@ static int dump_task_thread(struct parasite_ctl *parasite_ctl, const struct pstr
 
 	close_image(img);
 err:
+	compel_release_thread(tctl);
 	pr_info("----------------------------------------\n");
 	return ret;
 }
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -875,6 +875,69 @@ static int collect_file_locks(void)
 	return parse_file_locks();
 }
 
+static bool task_in_rseq(struct criu_rseq_cs *rseq_cs, uint64_t addr)
+{
+	return addr >= rseq_cs->start_ip && addr < rseq_cs->start_ip + rseq_cs->post_commit_offset;
+}
+
+static int fixup_thread_rseq(struct pstree_item *item, int i)
+{
+	CoreEntry *core = item->core[i];
+	struct criu_rseq_cs *rseq_cs = &dmpi(item)->thread_rseq_cs[i];
+	pid_t tid = item->threads[i].real;
+
+	/* equivalent to (struct rseq)->rseq_cs is NULL */
+	if (!rseq_cs->start_ip)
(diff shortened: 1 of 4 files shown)


criu/parasite-syscall.c
```diff
@@ -211,23 +211,17 @@ int parasite_dump_thread_seized(struct parasite_thread_ctl *tctl, struct parasit
...
 	}
 
-	compel_release_thread(tctl);
-
 	tid->ns[0].virt = args->tid;
 	return dump_thread_core(pid, core, args);
-
-err_rth:
-	compel_release_thread(tctl);
-	return -1;
```
mihalicyn, line 230: What do you think about splitting this change to a separate commit like "extend parasite_thread_ctl lifetime to dump_task_thread" or like that?

criu/cr-dump.c
```diff
@@ -1183,107 +1256,26 @@ static int dump_task_rseq(pid_t pid, struct pstree_item *item)
...
-		if (dump_task_thread(parasite_ctl, item, i))
-			return -1;
+		ret = dump_task_thread(parasite_ctl, item, i);
+		if (ret)
+			break;
 	}
 
-	return 0;
+	xfree(dmpi(item)->thread_rseq_cs);
+	dmpi(item)->thread_rseq_cs = NULL;
```
mihalicyn, line 1277: Not related to your change. But in fact there is a memory leak. And that's my bad in this case.

We are allocating `dmpi(item)->thread_rseq_cs` in `dump_task_rseq`. Then:
```

	ret = dump_task_rseq(pid, item); // <<-- allocated
	if (ret) {
		pr_err("Dump %d rseq failed %d\n", pid, ret);
		goto err;
	}

	parasite_ctl = parasite_infect_seized(pid, item, &vmas);
	if (!parasite_ctl) {
		pr_err("Can't infect (pid: %d) with parasite\n", pid);
		goto err;
	}

//... a lot of code...

	ret = dump_task_threads(parasite_ctl, item); // <-- freed
	if (ret) {
		pr_err("Can't dump threads\n");
		goto err_cure;
	}
```

If anything goes wrong in between `dump_task_rseq` and `dump_task_threads` then we have a memleak. And I think it's not only about `dmpi(item)->thread_rseq_cs` but also about `dmpi(item)->thread_ctls` and `dmpi(item)->thread_sp` too...

That's ToDo for us :-)

mihalicyn:
Hi, Michał

Thanks for investigating and fixing this!

LGTM, except a small nit about splitting your change touching `compel_release_thread(tctl)` into a separate commit (but maybe I'm too picky ;-) )

mihalicyn (approved):

PR #2149, opened 2023-04-03
action-scripts: add pre-stream hook

rst0git:
This pull request adds a new action-script hook `pre-stream`. This hook allows to start image streamer process from an action script.

### DIFF
--- a/criu/action-scripts.c
+++ b/criu/action-scripts.c
@@ -18,6 +18,7 @@
 #include "common/scm.h"
 
 static const char *action_names[ACT_MAX] = {
+	[ACT_PRE_STREAM] = "pre-stream",
 	[ACT_PRE_DUMP] = "pre-dump",
 	[ACT_POST_DUMP] = "post-dump",
 	[ACT_PRE_RESTORE] = "pre-restore",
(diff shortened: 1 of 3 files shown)


avagin:
Could you add a bit more details why it should be started from an action script?

rst0git:
> Could you add a bit more details why it should be started from an action script?

By default, when `criu` is used with `--stream` it opens `streamer-capture.sock` (or `streamer-serve.sock`) in the path specified with `--images-dir`. However, when checkpointing/restoring containers, this path is typically generated by the container runtime.

The `pre-stream` hook allows to start the streamer with `--images-dir $CRTOOLS_IMAGE_DIR` when checkpointing/restoring containers.

Example:
```bash
$ cat /etc/criu/default.conf 
stream
action-script /etc/criu/streamer-hook
```
```bash
$ sudo cat /etc/criu/streamer-hook 
#!/bin/bash

if [ "$CRTOOLS_SCRIPT_ACTION" == "pre-stream" ]; then
	IMG_FILE="$CRTOOLS_IMAGE_DIR/img.lz4"
	if [ ! -e "$IMG_FILE" ]; then 
		setsid bash -c "criu-image-streamer --images-dir $CRTOOLS_IMAGE_DIR capture | lz4 -f - $IMG_FILE" &
		sleep 2
	else
		setsid bash -c "lz4 -d $IMG_FILE - | criu-image-streamer --images-dir $CRTOOLS_IMAGE_DIR serve" &
		sleep 2
	fi
fi

exit 0
```

PR #2159, opened 2023-04-18
dump: increase fcntl call failure judgment

hdzhoujie:
The pipe_size type is unsigned int, when the fcntl call fails and 
return -1, it will cause a negative rollover problem.

Signed-off-by: zhoujie <zhoujie133@huawei.com>

### DIFF
--- a/criu/page-pipe.c
+++ b/criu/page-pipe.c
@@ -99,6 +99,7 @@ static struct page_pipe_buf *ppb_alloc(struct page_pipe *pp, unsigned int ppb_fl
 {
 	struct page_pipe_buf *prev = pp_prev_ppb(pp, ppb_flags);
 	struct page_pipe_buf *ppb;
+	int ppb_size = 0;
 
 	ppb = xmalloc(sizeof(*ppb));
 	if (!ppb)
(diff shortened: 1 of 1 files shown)


criu/page-pipe.c
```diff
@@ -120,7 +121,13 @@ static struct page_pipe_buf *ppb_alloc(struct page_pipe *pp, unsigned int ppb_fl
 		cnt_add(CNT_PAGE_PIPES, 1);
 
 		ppb->pipe_off = 0;
-		ppb->pipe_size = fcntl(ppb->p[0], F_GETPIPE_SZ, 0) / PAGE_SIZE;
```
mihalicyn, line 123: Just wonder if it's a real case when `fcntl(ppb->p[0], F_GETPIPE_SZ, 0)` fails? As far as I can see from the kernel code it can only fail if `ppb->p[0]` is not a pipe FD, but `ppb->p[0]` is always a pipe FD because it is created a few lines above.

hdzhoujie: Thank you for your reply.
From the perspective of using security, it should be necessary to increase the judgment, otherwise the failure will not be recognized.

criu/page-pipe.c
```diff
@@ -99,6 +99,7 @@ static struct page_pipe_buf *ppb_alloc(struct page_pipe *pp, unsigned int ppb_fl
 {
 	struct page_pipe_buf *prev = pp_prev_ppb(pp, ppb_flags);
 	struct page_pipe_buf *ppb;
+	int ppb_size = 0;
```
avagin, line 102: nit: s/int/long/

hdzhoujie: The return type of the fcntl function is int

avagin: The return type of the system call is long: https://elixir.bootlin.com/linux/latest/source/fs/fcntl.c#L442

avagin: you are right, the return type of the libc wrapper is int. Moreover, the pipe size is limited by (1<<31): https://elixir.bootlin.com/linux/latest/source/fs/pipe.c#L1236

Issue #2129, opened 2023-03-15
Cirrus CI / Vagrant Fedora based test fails 


sankalp-12:
Encountered this problem on one of my own PRs: https://github.com/checkpoint-restore/criu/pull/2125/checks?check_run_id=12029551674.

```================= Run zdtm/static/socket-tcp-nfconntrack in h ==================
Start test
Test is SUID
./socket-tcp-nfconntrack --pidfile=socket-tcp-nfconntrack.pid --outfile=socket-tcp-nfconntrack.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/socket-tcp-nfconntrack/59/1/restore.log
------------------------ grep Error ------------------------
b'(00.012179)     59: \tRunning ip rule delete table local'
b'(00.013981)     59: \tRunning ip rule restore'
b'(00.020002)     59: \tRunning iptables-restore -w for iptables-restore -w'
b'(00.024142)     59: \tRunning ip6tables-restore -w for ip6tables-restore -w'
b'(00.028625)     59: Error (criu/libnetlink.c:54): -16 reported by netlink: Device or resource busy'
b"(00.028669)     59: Error (criu/util.c:1413): Can't wait or bad status: errno=0, status=65280"
b'(00.028688) Error (criu/cr-restore.c:2543): Restoring FAILED.'
------------------------ ERROR OVER ------------------------
######### Test zdtm/static/socket-tcp-nfconntrack FAIL at CRIU restore #########
```
However, the test fails on each PR.

rst0git:
I was able to replicate this error on Fedora 37 with `./zdtm.py run -t zdtm/static/socket-tcp-nfconntrack`.
It looks like CRIU fails when [`restore_nf_ct(nsid, CR_FD_NETNF_CT)`](https://github.com/checkpoint-restore/criu/blob/bd21ecc2394841c784c931582c4e0902fa38c21b/criu/net.c#L2828) →[`nlmsg_receive(...)`](https://github.com/checkpoint-restore/criu/blob/bd21ecc2394841c784c931582c4e0902fa38c21b/criu/libnetlink.c#L117) is called.

```
(00.028451)     59: Error (criu/libnetlink.c:54): -16 reported by netlink: Device or resource busy
(00.028606)     59: Error (criu/util.c:1413): Can't wait or bad status: errno=0, status=65280
```

A link to the full log: [restore.log](https://github.com/checkpoint-restore/criu/files/11029003/restore.log.txt)

PR #2161, opened 2023-04-24
compel: support XSAVE on newer Intel CPUs

adrianreber:
Newer Intel CPUs (Sapphire Rapids) have a much larger xsave area than before. Looking at older CPUs I see 2440 bytes.
```
    # cpuid -1 -l 0xd -s 0
    ...
        bytes required by XSAVE/XRSTOR area     = 0x00000988 (2440)
```
On newer CPUs (Sapphire Rapids) it grows to 11008 bytes.
```
    # cpuid -1 -l 0xd -s 0
    ...
        bytes required by XSAVE/XRSTOR area     = 0x00002b00 (11008)
```
This increases the xsave area from one page to four pages.

Without this patch the fpu03 test fails, with this patch it works again.

### DIFF
--- a/compel/arch/x86/src/lib/include/uapi/asm/fpu.h
+++ b/compel/arch/x86/src/lib/include/uapi/asm/fpu.h
@@ -21,7 +21,28 @@
 #define XSTATE_YMM 0x4
 
 #define FXSAVE_SIZE 512
-#define XSAVE_SIZE  4096
+/*
+ * This used to be 4096 (one page). There is a comment below concerning
+ * this size:
+ *  "One page should be enough for the whole xsave state ;-)"
+ * Which is kind of funny as it is no longer enough ;-)
+ *
+ * Older CPUs:
+ * # cpuid -1 -l 0xd -s 0
+ * ...
+ *     bytes required by XSAVE/XRSTOR area     = 0x00000988 (2440)
+ *
+ * Newer CPUs (Sapphire Rapids):
+ * # cpuid -1 -l 0xd -s 0
+ * ...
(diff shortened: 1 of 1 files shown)


adrianreber:
CC: @0x7f454c46

adrianreber:
See #2162 for a fix for the "Run code linter" error.

Not sure what is up with the docker test. Something is broken during package installation. I was not able to reproduce that error locally.

adrianreber:
The "no VDSO" tests failures are tracked in #2129

0x7f454c46 (approved):
Seems reasonable to me.
Firstly, I thought why don't we define it as the kernel does with a variable-sized array:
```
struct xregs_state {
	struct fxregs_state		i387;
	struct xstate_header		header;
	u8				extended_state_area[];
} __attribute__ ((packed, aligned (64)));
```
but then, looking where its roots go: to `fpu_state_t`, and into `struct rt_sigframe`, which on its own goes to `RESTORE_STACK_SIGFRAME`. So, in theory, we could add a callback to `crtools.c` from
```
	memzone_size = round_up(sizeof(struct restore_mem_zone) * current->nr_threads, page_size());
```
which would give a proper mem_zone size for an image file..

That would avoid these `XSAVE_SIZE` adjustments but sounds too intricate for a simple fix like this one.

adrianreber:
@0x7f454c46 thanks for the review.

@avagin Not sure how far along the next release is, but it would be nice to have this also in the upcoming release if doable.

avagin:
@adrianreber we will need to do v3.18.1.

PR #2162, opened 2023-04-24
Fix CI errors (shellcheck,  codespell)

adrianreber:
Fixes an error in CI during `make lint`.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 skip = ./.git,./test/pki
-ignore-words-list = creat,fpr,fle,ue,bord,parms,nd,te,testng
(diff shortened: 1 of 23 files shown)


adrianreber:
Hmm, okay, spellcheck is also very unhappy. Will update the PR.

Snorch:
> `make indent` fails. @Snorch you have been involved in our `clang-format` setup recently. What is our current setup concerning formatting changes by `clang-format`.

`make indent` tries to realign everything in flog_genbit, trying to do stupid thing of putting consecutive comments on one line:
```
+		/* Not used for a while */ /* float:			12, */ /* double:			13, */ /* long double:		14, */ \
```
We can just ignore it (moreover version 15.0.7-1.fc37 on my laptop does not see it as a problem, probably something changed in newer versions), the error should not appear again.

PR #2114, opened 2023-03-08
criu-ns: Add tests for criu-ns script

warusadura:
These changes add test implementations for criu-ns script.

Fixes: #1909

PR #2163, opened 2023-04-25
criu-ns: make --pidfile option show pid in caller pidns

Snorch:
Using the fact that we know criu_pid and criu is a parent of restored process we can create pidfile with pid on caller pidns level.

We need to move mount namespace creation to child so that criu-ns can see caller pidns proc.

This should help with adding the test for repeated criu-ns c/r. See https://github.com/checkpoint-restore/criu/pull/2114#issuecomment-1517590581

### DIFF
--- a/scripts/criu-ns
+++ b/scripts/criu-ns
@@ -6,6 +6,7 @@ import sys
 import os
 import fcntl
 import termios
+import time
 
 # <sched.h> constants for unshare
 CLONE_NEWNS = 0x00020000
@@ -110,8 +111,8 @@ def wrap_restore():
     if '--restore-sibling' in restore_args:
         raise OSError(errno.EINVAL, "--restore-sibling is not supported")
 
-    # Unshare pid and mount namespaces
-    if _unshare(CLONE_NEWNS | CLONE_NEWPID) != 0:
+    # Unshare pid namespace
+    if _unshare(CLONE_NEWPID) != 0:
         _errno = ctypes.get_errno()
         raise OSError(_errno, errno.errorcode[_errno])
 
@@ -123,8 +124,32 @@ def wrap_restore():
         restore_detached = True
         restore_args.remove('--restore-detached')
 
+    restore_pidfile = None
+    if '--pidfile' in restore_args:
+        try:
+            opt_index = restore_args.index('--pidfile')
+            restore_pidfile = restore_args[opt_index + 1]
+            del restore_args[opt_index:opt_index + 2]
+        except (ValueError, IndexError, FileNotFoundError):
+            raise OSError(errno.ENOENT, "--pidfile missing argument")
+
+        if not restore_pidfile.startswith('/'):
+            for base_dir_opt in ['--work-dir', '-W', '--images-dir', '-D']:
+                if base_dir_opt in restore_args:
+                    try:
+                        opt_index = restore_args.index(base_dir_opt)
+                        restore_pidfile = os.path.join(restore_args[opt_index + 1], restore_pidfile)
+                        break
+                    except (ValueError, IndexError, FileNotFoundError):
+                        raise OSError(errno.ENOENT, base_dir_opt + " missing argument")
+
     criu_pid = os.fork()
     if criu_pid == 0:
+        # Unshare mount namespace
+        if _unshare(CLONE_NEWNS) != 0:
+            _errno = ctypes.get_errno()
+            raise OSError(_errno, errno.errorcode[_errno])
+
         os.setsid()
         # Set stdin tty to be a controlling tty of our new session, this is
         # required by --shell-job option, as for it CRIU would try to set a
@@ -138,6 +163,25 @@ def wrap_restore():
 
         _mount_new_proc()
         run_criu(restore_args)
+
+    if restore_pidfile:
+        restored_pid = None
+        retry = 5
+
+        while not restored_pid and retry:
+            with open('/proc/%d/task/%d/children' % (criu_pid, criu_pid)) as f:
+                line = f.readline().strip()
+                if len(line):
+                    restored_pid = line
+                    break
+            retry -= 1
+            time.sleep(1)
+
+        if restored_pid:
+            with open(restore_pidfile, 'w+') as f:
+                f.write(restored_pid)
+        else:
+            print("Warn: Search of restored pid for --pidfile option timeouted")
 
     if restore_detached:
         return 0


scripts/criu-ns
```diff
@@ -139,6 +155,23 @@ def wrap_restore():
...
+        retry = 5
+
+        while not restored_pid and retry:
+            with open('/proc/%d/task/%d/children' % (criu_pid, criu_pid)) as f:
+                line = f.readline().strip()
+                if len(line):
+                    restored_pid = line
+                    break
+            retry -= 1
+            time.sleep(1)
```
rst0git, line 169: @Snorch Would it make sense to show a warning or error message in the case of timeout?

scripts/criu-ns
```diff
@@ -122,9 +123,24 @@ def wrap_restore():
     if '--restore-detached' in restore_args:
         restore_detached = True
         restore_args.remove('--restore-detached')
+    restore_pidfile = None
+    if '--pidfile' in restore_args:
+        try:
+            opt_index = restore_args.index('--pidfile')
+            restore_pidfile = restore_args[opt_index + 1]
+            restore_args.pop(opt_index)
+            restore_args.pop(opt_index)
```
rst0git, line 132: `del` can remove multiple elements from a list. For example:

```Python
del restore_args[opt_index:opt_index + 2]
```

scripts/criu-ns
```diff
@@ -139,6 +155,23 @@ def wrap_restore():
...
+                line = f.readline().strip()
+                if len(line):
+                    restored_pid = line
+                    break
+            retry -= 1
+            time.sleep(1)
+
+        if restored_pid:
+            with open(restore_pidfile, 'w+') as f:
+                f.write(restored_pid)
```
rst0git, line 173: I believe that CRIU creates the pidfile in the path specified with `--work-dir`, or if not specified it uses the path from `--images-dir` path.

@Snorch Could we get this path from `/proc/<PID>/cwd`?
https://github.com/checkpoint-restore/criu/blob/166aac0fb2673336e3cfa2190fedface0443bc27/criu/crtools.c#L252

warusadura: I think the `pidfile` should be created in the path specified by `--images-dir`. However, changes introduced in this PR create the `pidfile` in `criu/test/others/criu-ns`. It should be in `criu/test/others/criu-ns/dumpdir`

scripts/criu-ns
```diff
@@ -122,9 +123,24 @@ def wrap_restore():
...
+    restore_pidfile = None
+    if '--pidfile' in restore_args:
+        try:
+            opt_index = restore_args.index('--pidfile')
+            restore_pidfile = restore_args[opt_index + 1]
+            restore_args.pop(opt_index)
+            restore_args.pop(opt_index)
+        except (ValueError, IndexError, FileNotFoundError):
+            raise OSError(errno.ENOENT, "--pidfile missing argument")
+
```
warusadura, line 135: nit: extra newline.

scripts/criu-ns
```diff
@@ -139,6 +155,23 @@ def wrap_restore():
...
 
+    if restore_pidfile:
+        restored_pid = None
+        retry = 5
+
+        while not restored_pid and retry:
+            with open('/proc/%d/task/%d/children' % (criu_pid, criu_pid)) as f:
+                line = f.readline().strip()
+                if len(line):
+                    restored_pid = line
```
warusadura, line 166: question: I have executed this version of `criu-ns` script multiple times with my tests. And I always ended up with the same (content of `pidfile`) `restored_pid`: `15`. Is this right? :)

scripts/criu-ns
Snorch: Relative paths handling in CRIU is really bad to my taste, normally the caller of some command expects relative paths given in arguments be relative to cwd.

In criu we actually can save cwd in fdstore and resolve relative path with openat relative to it. Maybe it would be more logical. Sounds like a good project for GSOC.

Currently predicting there criu will put some file is really hard, it can put it to workdir or images dir or / if we save it after setns(CLONE_NEWNS) or some other syscall which affects cwd....

I see in depopulate_roots_yard that we save and restore cwd around setns, but we don't restore root, meaning that instead absolute paths can break, e.g. If originally criu was called from chroot...

But yes it seems for me that currently pidfile is put as you say, will try to do same in criu-ns.

scripts/criu-ns
rst0git: I believe this is because we get the value of `restored_pid` after unsharing the PID namespace in `criu-ns`.

Snorch: Unsharing pid namespace does not change current pid namespace, it only changes pid namespace for children, current implementation in criu-ns in this pr gives pid in pidns of the caller, if you have pid 15 there each time maybe you caller is itself in some pidns...

scripts/criu-ns
```diff
@@ -122,9 +123,39 @@ def wrap_restore():
...
+        if restore_pidfile[0] != '/':
+            base_dir_opt = None
+            if '--work-dir' in restore_args:
+                base_dir_opt = '--work-dir'
+            elif '-W' in restore_args:
+                base_dir_opt = '-W'
+            elif '--images-dir' in restore_args:
+                base_dir_opt = '--images-dir'
+            elif '-D' in restore_args:
+                base_dir_opt = '-D'
```
rst0git, line 143: this could be simplified to something like:
```Python
if not restore_pidfile.startswith('/'):
    for base_dir_opt in ['--work-dir', '-W', '--images-dir', '-D']:
        if base_dir_opt in restore_args:
            try:
                opt_index = restore_args.index(base_dir_opt)
                restore_pidfile = os.path.join(restore_args[opt_index + 1], restore_pidfile)
                break
            except (ValueError, IndexError, FileNotFoundError):
                raise OSError(errno.ENOENT, base_dir_opt + " missing argument")
```

rst0git, line 143: Could we simplify this to something like the following?
```Python
if not restore_pidfile.startswith('/'):
    for base_dir_opt in ['--work-dir', '-W', '--images-dir', '-D']:
        if base_dir_opt in restore_args:
            try:
                opt_index = restore_args.index(base_dir_opt)
                restore_pidfile = os.path.join(restore_args[opt_index + 1], restore_pidfile)
                break
            except (ValueError, IndexError, FileNotFoundError):
                raise OSError(errno.ENOENT, base_dir_opt + " missing argument")
```

scripts/criu-ns
warusadura: understood, thanks @Snorch. Yes, I can dump pid 15 :)

scripts/criu-ns
Snorch: Done. Thanks it becomes much better this way!

warusadura:
I can confirm my PR #2114 works perfect (even the redump) with the changes introduced in this PR.

warusadura (approved):
I can confirm my PR #2114 works perfect (even the redump) with the changes introduced in this PR.

rst0git (approved):
LGTM

PR #2165, opened 2023-04-26
criu/plugin: Add environment variable to cap size of buffers.

fdavid-amd:
The amdgpu plugin would create a memory buffer at the size of the largest VRAM bo (buffer object). On some systems, VRAM size exceeds RAM size, so the largest bo might be larger than the available memory.

Add an environment variable KFD_MAX_BUFFER_SIZE, which caps the size of this buffer. By default, it is set to 0, and has no effect. When active, any bo larger than its value will be saved to/restored from file in multiple passes.


fdavid-amd:
Now finished internal review and precursory testing

PR #2167, opened 2023-04-28
criu/plugin: Add environment variable to cap size of buffers.

fdavid-amd:
The amdgpu plugin would create a memory buffer at the size of the largest VRAM bo (buffer object). On some systems, VRAM size exceeds RAM size, so the largest bo might be larger than the available memory.

Add an environment variable KFD_MAX_BUFFER_SIZE, which caps the size of this buffer. By default, it is set to 0, and has no effect. When active, any bo larger than its value will be saved to/restored from file in multiple passes.

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -107,6 +107,8 @@ extern bool kfd_vram_size_check;
 extern bool kfd_numa_check;
 extern bool kfd_capability_check;
 
+size_t kfd_max_buffer_size;
+
 /**************************************************************************************************/
 
 int write_fp(FILE *fp, const void *buf, const size_t buf_len)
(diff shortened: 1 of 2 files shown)


plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -449,6 +451,47 @@ void getenv_bool(const char *var, bool *value)
...
+	pr_info("Value str: %s\n", value_str);
+
+	if (value_str) {
+		size = (size_t)strtoul(value_str, &endp, 0);
+		if (errno || value_str == endp) {
+			pr_err("Ignoring invalid value for %s=%s, expecting a positive integer\n", var, value_str);
+			return;
+		}
+		switch (*endp) {
+		case 'k':
```
avagin, line 470: should it be the uppercase K?

avagin:
@dayatsin-amd  could you review this pr?

plugins/amdgpu/amdgpu_plugin.c
rst0git: https://en.wikipedia.org/wiki/Units_of_information#Systematic_multiples

> In the past, uppercase K has been used instead of lowercase k to indicate 1024 instead of 1000. However, this usage was never consistently applied.

dayatsin-amd: Should be accept both, lowercase and uppercase K to indicate 1024 then?

fdavid-amd: Base CRIU accepts 'K' but not 'k' for its size-parsing config options. The standard is 'k' for 1000. I'm willing to add both.

fdavid-amd:
I've set the default to off to be conservative. Should the default instead be some reasonable, large value - maybe 2G or 4G? I have seen large SDMA copies cause issues even when the amount of RAM is sufficient.

dayatsin-amd:
[AMD Official Use Only - General]

Hi David,
Sorry, I missed that msg. I'd recommend setting the default to 4Gb.


________________________________

avagin:
LGTM

@dayatsin-amd does it look good to you?

dayatsin-amd:
lgtm as well. 
Thanks!

PR #2191, opened 2023-06-06
test/other: add test for action-script

rst0git:
This pull request is introducing a test for the action-script functionality of CRIU to verify that `pre-dump`, `post-dump`, `pre-restore`, `pre-resume`, `post-restore`, and `post-resume` hooks are executed during dump/restore.


Snorch (approved):
Nice test.

Note: would be good to have other script points also covered later, like `setup-namespaces`, `post-setup-namespaces`, which are only called in case `root_ns_mask` is not 0, and maybe others https://github.com/checkpoint-restore/criu/blob/988a5f4816d5bf651f9cdfa8b85359da18a77a7c/criu/action-scripts.c#L21.

PR #2194, opened 2023-06-14
A set of assorted fixes in logging and error propagation.

osctobe:
The changes in this series are mostly independent and could be applied individually. If there are change requests I'll shuffle relevant commits to the back when addressing them.

### DIFF
--- a/criu/include/sk-inet.h
+++ b/criu/include/sk-inet.h
@@ -69,6 +69,7 @@ extern int inet_connect(int sk, struct inet_sk_info *);
 
 #ifdef CR_NOGLIBC
 #define setsockopt sys_setsockopt
+#define pr_perror(fmt, ...) pr_err(fmt ": errno %d\n", ##__VA_ARGS__, -ret)
 #endif
 static inline void tcp_repair_off(int fd)
 {
@@ -76,7 +77,7 @@ static inline void tcp_repair_off(int fd)
 
 	ret = setsockopt(fd, SOL_TCP, TCP_REPAIR, &aux, sizeof(aux));
 	if (ret < 0)
-		pr_err("Failed to turn off repair mode on socket\n");
+		pr_perror("Failed to turn off repair mode on socket %d", fd);
 }
 
 extern void tcp_locked_conn_add(struct inet_sk_info *);


criu/include/log.h
```diff
@@ -60,6 +60,8 @@ void flush_early_log_buffer(int fd);
 
 #define pr_perror(fmt, ...) pr_err(fmt ": %s\n", ##__VA_ARGS__, strerror(errno))
 
+#define pr_pwarn(fmt, ...) pr_warn(fmt ": %s\n", ##__VA_ARGS__, strerror(errno))
```
Snorch, line 63: As you add `pr_pwarn` new printing function, you also need to update `lint` accordingly.

criu/include/sk-inet.h
```diff
@@ -76,7 +76,13 @@ static inline void tcp_repair_off(int fd)
 
 	ret = setsockopt(fd, SOL_TCP, TCP_REPAIR, &aux, sizeof(aux));
 	if (ret < 0)
-		pr_err("Failed to turn off repair mode on socket\n");
+		pr_warn("Failed to turn off repair mode on socket %d: "
+#ifdef CR_NOGLIBC
+			"%d\n", fd, ret
+#else
+			"%m\n", fd
```
Snorch, line 83: General rule in CRIU is that we don't allow using %m, please don't. Also there is no point in printing ret as it is always -1.

I would advise dropping this patch.

criu/files-reg.c
```diff
@@ -1818,7 +1818,7 @@ int dump_one_reg_file(int lfd, u32 id, const struct fd_parms *p)
 	}
 
 	if (!skip_for_shell_job && mnt_is_overmounted(mi)) {
-		pr_err("Open files on overmounted mounts are not supported yet\n");
+		pr_err("Open files on overmounted mounts are not supported yet; %s fd=%d path=%s\n", ext_id, p->fd, link->name + 1);
```
Snorch, line 1821: Can you please replace ext_id printing with just mnt_id printing? The variable ext_id sole purpose is for external file lookup, using it in other places does not look good.

criu/sk-unix.c
```diff
@@ -878,7 +878,8 @@ static int __dump_external_socket(struct unix_sk_desc *sk, struct unix_sk_desc *
 
 	if (peer->type != SOCK_DGRAM) {
 		show_one_unix("Ext stream not supported", peer);
-		pr_err("Can't dump half of stream unix connection.\n");
+		pr_err("Can't dump half of stream unix connection. "
+		       " name: %s; peer name: %s\n", sk->name, peer->name);
```
Snorch, line 882: Double space, please remove.

criu/mount.c
```diff
@@ -1228,7 +1228,7 @@ int __open_mountpoint(struct mount_info *pm)
 		return -1;
 	}
 
-	if (check_mountpoint_fd(pm, mnt_fd)) {
+	if (__check_mountpoint_fd(pm, mnt_fd, false)) {
```
Snorch, line 1231: There is no point in doing this change, unless you want to remove check_mountpoint_fd completely...

criu/include/log.h
osctobe: Done.

criu/mount.c
Snorch: It seems you broke tests with this one. Please just drop hunk in that place.

Snorch:
One general note here, in criu there is a requirement to prefix subjects with component name. See https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes

Snorch:
```
Subject: [PATCH] util: Remove error logs for ignored or otherwise logged
diff --git a/criu/netfilter.c b/criu/netfilter.c
index <some index>
--- a/criu/netfilter.c
+++ b/criu/netfilter.c
index <some index>
--- a/criu/util.c
+++ b/criu/util.c
```
In this case right prefix would be `util/netfilter:` or just `log:`

criu/include/sk-inet.h
```diff
@@ -69,14 +69,15 @@ extern int inet_connect(int sk, struct inet_sk_info *);
...
+#define pr_pwarn(fmt, ...) pr_warn(fmt ": %d\n", ##__VA_ARGS__, ret)
 #endif
 static inline void tcp_repair_off(int fd)
 {
 	int aux = 0, ret;
 
 	ret = setsockopt(fd, SOL_TCP, TCP_REPAIR, &aux, sizeof(aux));
 	if (ret < 0)
-		pr_err("Failed to turn off repair mode on socket\n");
+		pr_pwarn("Failed to turn off repair mode on socket %d", fd);
```
avagin, line 80: It is the error. It is ignored only because we can't stop the restore at this point.

osctobe: Leaving at error level then.

0x7f454c46:
`Change-Id: I8e5e4953f0231ed935ef3dbbab800080e4ee549f` clean this from the commit.

You don't submit those to linux/kernel, neither you do to criu. We hold the same rule, see:
https://elixir.bootlin.com/linux/latest/source/Documentation/dev-tools/checkpatch.rst#L593
`The Change-Id line must be removed before submitting.`

PR #2196, opened 2023-06-15
Guarantee run-id uniqueness

osctobe:
The first patch improves on run-id generation to make it guaranteed to be unique on a running machine.
Second removes wrapping tests in an empty netns that was working around the lack of run-id uniqueness.

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -1880,11 +1880,16 @@ uint64_t criu_run_id;
 
 void util_init(void)
 {
-	struct timespec tp;
+	struct stat statbuf;
 
-	clock_gettime(CLOCK_MONOTONIC, &tp);
-	criu_run_id = ((uint64_t)getpid() << 32) + tp.tv_sec + tp.tv_nsec;
+	criu_run_id = getpid();
+	if (!stat("/proc/self/ns/pid", &statbuf))
+		criu_run_id |= (uint64_t)statbuf.st_ino << 32;
+	else if (errno != ENOENT)
+		pr_perror("Can't stat /proc/self/ns/pid - CRIU run id might not be unique");
+
 	compel_run_id = criu_run_id;
+	pr_info("CRIU run id = %#" PRIx64 "\n", criu_run_id);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


test/zdtm_ct.c
```diff
@@ -102,10 +102,9 @@ int main(int argc, char **argv)
 	/*
 	 * pidns is used to avoid conflicts
 	 * mntns is used to mount /proc
-	 * net is used to avoid conflicts of parasite sockets
```
avagin, line 105: This isn't the only reason. We want to isolate network tests from each other. All our tcp tests use hard-coded port numbers and they will start failing with this change.

criu/util.c
```diff
@@ -1880,11 +1880,13 @@ uint64_t criu_run_id;
...
 void util_init(void)
 {
-	struct timespec tp;
+	struct stat statbuf;
 
-	clock_gettime(CLOCK_MONOTONIC, &tp);
-	criu_run_id = ((uint64_t)getpid() << 32) + tp.tv_sec + tp.tv_nsec;
+	criu_run_id = getpid();
+	if (!stat("/proc/self/ns/pid", &statbuf))
+		criu_run_id |= (uint64_t)statbuf.st_ino << 32;
```
avagin, line 1887: fall back to the clock_gettime way if /proc/self/ns/pid isn't accessible.  Conflicts with clock_gettime() are almost impossible.

osctobe: If pidns is not enabled in kernel, then names based on CRIU's pid alone are already machine-level unique.

avagin: In this case, you need to check that errno is ENOENT and report errors in other cases.

test/zdtm_ct.c
avagin: If you want to run tests without network namespaces in your environment, you can introduce an environment variable for that.

osctobe: Ok, I didn't see any tcp test failures, but maybe because I ran them with test sharding enabled. I'll rework the comment then instead of this patch.

PR #2197, opened 2023-06-15
Allow skipping iptables/nftables invocation.

osctobe:
Make it possible to skip network lock to enable uses that break connections anyway to work without iptables/nftables being present. This is used in Google's environment where we drop all connections and reconnect after migration.


adrianreber:
I like this change. That would be also useful for containers I think. It needs additional changes in `cr-service.c` I think and I also think that the name `NETWORK_LOCK_DUMMY` is not the best choice. Maybe a name that includes `NO_LOCK` or so would be better.

osctobe:
Would  NONE instead of DUMMY be better? (Keeping it a single-word.)

rst0git:
> Would NONE instead of DUMMY be better? (Keeping it a single-word.)

`NETWORK_LOCK_NONE` looks better. It is similar to the `--manage-cgroups none` option.

criu/config.c
```diff
@@ -1036,6 +1036,8 @@ int parse_options(int argc, char **argv, bool *usage_error, bool *has_exec_cmd,
 				opts.network_lock_method = NETWORK_LOCK_IPTABLES;
 			} else if (!strcmp("nftables", optarg)) {
 				opts.network_lock_method = NETWORK_LOCK_NFTABLES;
+			} else if (!strcmp("none", optarg)) {
```
rst0git, line 1039: Would it be possible to update the man page as well?
https://github.com/checkpoint-restore/criu/blob/c75c017e/Documentation/criu.txt#L453

rst0git:
> Make it possible to skip network lock to enable uses that break connections anyway to work without iptables/nftables being present.

How does this option relate to `--tcp-close`? Should these two options be used together?

osctobe:
> > Make it possible to skip network lock to enable uses that break connections anyway to work without iptables/nftables being present.
> 
> How does this option relate to `--tcp-close`? Should these two options be used together?

It makes `--tcp-close` cheaper. Added a hint to the doc.

criu/cr-service.c
```diff
@@ -520,6 +520,9 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 		case CRIU_NETWORK_LOCK_METHOD__NFTABLES:
 			opts.network_lock_method = NETWORK_LOCK_NFTABLES;
 			break;
+		case CRIU_NETWORK_LOCK_METHOD__NONE:
```
rst0git, line 523: you might need to also update the `criu_network_lock_method` enum in `rpc.proto`:
https://github.com/checkpoint-restore/criu/blob/c75c017e/images/rpc.proto#L52

osctobe: Done.

images/rpc.proto
```diff
@@ -52,6 +52,7 @@ enum criu_cg_mode {
 enum criu_network_lock_method {
 	IPTABLES = 	1;
 	NFTABLES = 	2;
+	NONE =  	3;
```
rst0git, line 55: `NONE` appears to be already defined for the `sk_shutdown` enum in `sk-opts.proto`.
```
TypeError: Couldn't build proto file into descriptor pool!
Invalid proto descriptor for file "sk-opts.proto":
  NONE: "NONE" is already defined in file "rpc.proto".
  NONE: Note that enum values use C++ scoping rules, meaning that enum values are siblings of their type, not children of it.  Therefore, "NONE" must be unique within the global scope, not just within "sk_shutdown".
```

Documentation/criu.txt
```diff
@@ -457,6 +457,8 @@ The 'mode' may be one of the following:
 
     *nftables*::: Use nftables rules to drop the packets.
 
+    *none*::: Don't lock the network. (Useful with *--tcp-close*.)
```
avagin, line 460: We have to mention that CRIU can't dump tcp connections if the network isn't locked. If the `none` mode is used without `--tcp-close`, the user is responsible for locking the network.

osctobe: Done.

images/rpc.proto
osctobe: Changed to 'skip'.

PR #2198, opened 2023-06-15
ZDTM test assorted improvements

osctobe:
This series contains a mostly-independent patches that improve test debuggability and make running the tests easier to parallelize.

### DIFF
--- a/test/zdtm/static/cgroup_ifpriomap.checkskip
+++ b/test/zdtm/static/cgroup_ifpriomap.checkskip
@@ -1,3 +1,6 @@
 #!/bin/bash
+set -e
 
-! test -f /sys/fs/cgroup/cgroup.controllers
+test ! -f /sys/fs/cgroup/cgroup.controllers
+
+grep -q '^net_prio\s' /proc/cgroups


test/zdtm/static/cgroup_ifpriomap.checkskip
```diff
@@ -1,3 +1,9 @@
 #!/bin/bash
 
-! test -f /sys/fs/cgroup/cgroup.controllers
+test -f /sys/fs/cgroup/cgroup.controllers && exit 1
+
+grep -q 'cgroup rw,.*net_prio' /proc/mounts && exit 1
```
avagin, line 5: why do we want to skip the test if the net_prio cgroup hierarchy is mounted?

adrianreber:
It seems the Python import changes are breaking lot's of CI runs.

osctobe:
> It seems the Python import changes are breaking lot's of CI runs.

Dropped the patch for now.

test/zdtm/static/cgroup_ifpriomap.checkskip
osctobe: The test wants to mount the controller itself and if it is used by the system already the test fails.

test/zdtm.py
```diff
@@ -1491,6 +1492,7 @@ def restore(self):
             pstree_check_stopped(self.__test.getpid())
             pstree_signal(self.__test.getpid(), signal.SIGCONT)
 
+    @functools.cache
```
avagin, line 1501: https://github.com/checkpoint-restore/criu/actions/runs/5286913168/jobs/9566809091?pr=2198

```
+ ./test/zdtm.py run -a -p 2 --keep-going
3032
3033
  File "./test/zdtm.py", line 1046, in <module>
3034
    class criu:
3035
  File "./test/zdtm.py", line 1501, in criu
3036
    @functools.cache
3037
AttributeError: module 'functools' has no attribute 'cache'
```

avagin: pls add some numbers to show how much faster it is with this change.

osctobe: `functools.cache` is Python 3.9+ [1]. I'll drop this patch for now.

[1] https://docs.python.org/3/library/functools.html#functools.cache

test/zdtm/static/cgroup_ifpriomap.checkskip
Snorch: Mounting cgroup should be possible multiple times, maybe you can show real error you try to fix?

test/zdtm/static/cgroup04.checkskip
```diff
@@ -1,3 +1,10 @@
 #!/bin/bash
 
-! test -f /sys/fs/cgroup/cgroup.controllers
+test -f /sys/fs/cgroup/cgroup.controllers && exit 1
+
+grep -q 'cgroup rw,' /proc/mounts && exit 1
```
avagin, line 5: > cgroup04 test needs full control over mem and devices cgroup hierarchies.

In our CI, all cgroups are mounted and they are controlled by systemd, but the test is running without any issues.

test/zdtm/static/cgroup_ifpriomap.checkskip
```diff
@@ -1,3 +1,9 @@
...
+
+if [[ -r /proc/config.gz ]]; then
+  zcat /proc/config.gz | grep -q ^CONFIG_CGROUP_NET_PRIO= || exit 1
```
avagin, line 8: /proc/config.gz depends on CONFIG_IKCONFIG. Can we look at  /proc/cgroups?
```
$ cat /proc/cgroups  | grep net_prio || exit 1
```

test/zdtm.py
```diff
@@ -2019,12 +2019,23 @@ def __init__(self, opts, nr_tests):
...
-            if not opts["ignore_taint"] and os.getenv("ZDTM_IGNORE_TAINT") != '1':
-                raise Exception("The kernel is tainted: %r" % self.__taint)
+            self.__report_kernel_taint("The kernel is tainted: %r" % self.__taint)
+
+    @staticmethod
+    def __read_kernel_tainted():
+        with open("/proc/sys/kernel/tainted") as taintfd:
+            taint = taintfd.read()
+            if taint[-1:] == "\n":
+              taint = taint[:-1]
```
rst0git, line 2031: >test/zdtm.py:2031:15: E111 indentation is not a multiple of 4

test/zdtm.py
```diff
@@ -2019,12 +2019,23 @@ def __init__(self, opts, nr_tests):
...
-            if not opts["ignore_taint"] and os.getenv("ZDTM_IGNORE_TAINT") != '1':
-                raise Exception("The kernel is tainted: %r" % self.__taint)
+            self.__report_kernel_taint("The kernel is tainted: %r" % self.__taint)
+
+    @staticmethod
+    def __read_kernel_tainted():
+        with open("/proc/sys/kernel/tainted") as taintfd:
+            taint = taintfd.read()
+            if taint[-1:] == "\n":
+                taint = taint[:-1]
```
avagin, line 2031: taint = taint.strip()

test/zdtm.py
```diff
@@ -513,8 +513,14 @@ def kill(self, sig=signal.SIGKILL):
         self.__freezer.thaw()
         if self.__pid:
             print("Send the %d signal to  %s" % (sig, self.__pid))
-            os.kill(int(self.__pid), sig)
-            self.gone(sig == signal.SIGKILL)
+            try:
+                os.kill(int(self.__pid), sig)
+            except ProcessLookupError:
+                # Prevent gone() calling waitpid().
+                self.auto_reap = True
```
avagin, line 520: if auto_reap is false, we need to raise the exception, because this pid can't disappear before we call waitpid.

test/zdtm/static/cgroup04.checkskip
```diff
@@ -1,3 +1,14 @@
 #!/bin/bash
+set -e
 
-! test -f /sys/fs/cgroup/cgroup.controllers
+test ! -f /sys/fs/cgroup/cgroup.controllers
+
+grep -q '^devices\s' /proc/cgroups
+grep -q '^memory\s' /proc/cgroups
+
+if grep -q ' cgroup rw,.*devices' /proc/mounts; then
```
avagin, line 9: I think it is better to grep /proc/self/cgroup. The test doesn't use existing mounts and we need to check that this controller isn't combined with others.
I think it should be like this:
```
grep '^[0-9]*:devices:' /proc/self/cgroup
```

and pls add a comment that tells what we are checking here.

test/zdtm/static/cgroup04.checkskip
```diff
@@ -1,3 +1,14 @@
...
-! test -f /sys/fs/cgroup/cgroup.controllers
+test ! -f /sys/fs/cgroup/cgroup.controllers
+
+for ctl in devices memory; do
+	# Check that the controller is available and not co-mounted with any other.
+
+	grep -q "^${ctl}\\s" /proc/cgroups
+
+	if ! grep -q "^[0-9]*:${ctl}:" /proc/self/cgroup; then
+		grep -qE "^[0-9]*:([^:]*,)?${ctl}(,[^:]*)?:" /proc/self/cgroup && exit 1
```
avagin, line 12: pls add a comment with an example of string that we grep for

PR #2199, opened 2023-06-15
RPC control extensions

osctobe:
A set of independent features allowing more options to be controlled via RPC.

The series needs to be applied in sequence to limit conflict resolution due to RPC fields inserted.

### DIFF
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -155,6 +155,11 @@ not compatible with *--external* *dev*.
             notification message contains a file descriptor for
             the master pty
 
+        *query-ext-files*:::
+            called after the process tree is stopped and network is locked.
+            This hook is used only in the RPC mode. The notification reply
+            contains file ids to be added to external file list (may be empty).
+
 *--unprivileged*::
     This option tells *criu* to accept the limitations when running
     as non-root. Running as non-root requires *criu* at least to have
--- a/criu/action-scripts.c
+++ b/criu/action-scripts.c
@@ -31,6 +31,7 @@ static const char *action_names[ACT_MAX] = {
 	[ACT_POST_RESUME] = "post-resume",
 	[ACT_ORPHAN_PTS_MASTER] = "orphan-pts-master",
 	[ACT_STATUS_READY] = "status-ready",
+	[ACT_QUERY_EXT_FILES] = "query-ext-files",
 };
 
 struct script {
@@ -110,6 +111,20 @@ int rpc_send_fd(enum script_actions act, int fd)
 
 	pr_debug("\tRPC\n");
 	return send_criu_rpc_script(act, (char *)action, rpc_sk, fd);
+}
+
+int rpc_query_external_files(void)
+{
+	int rpc_sk;
+
+	if (scripts_mode != SCRIPTS_RPC)
+		return 0;
+
+	rpc_sk = get_service_fd(RPC_SK_OFF);
+	if (rpc_sk < 0)
+		return -1;
+
+	return exec_rpc_query_external_files((char *)action_names[ACT_QUERY_EXT_FILES], rpc_sk);
 }
 
 int run_scripts(enum script_actions act)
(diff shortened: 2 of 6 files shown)


criu/cr-service.c
```diff
@@ -339,8 +339,14 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	 */
 	if (imgs_changed_by_rpc_conf)
 		strncpy(images_dir_path, opts.imgs_dir, PATH_MAX - 1);
-	else
+	else if (req->has_images_dir_fd)
 		sprintf(images_dir_path, "/proc/%d/fd/%d", ids.pid, req->images_dir_fd);
+	else if (req->images_dir)
```
avagin, line 388: pls explain in the commit message why we need this rpc argument. Why can't you use images_dir_fd?

criu/cr-service.c
```diff
@@ -729,6 +729,9 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	if (req->orphan_pts_master)
 		opts.orphan_pts_master = true;
 
+	if (req->has_display_stats)
```
avagin, line 779: if (req->display_stats)

criu/cr-service.c
```diff
@@ -240,6 +240,50 @@ int send_criu_rpc_script(enum script_actions act, char *name, int sk, int fd)
...
+
+	if (req->type != CRIU_REQ_TYPE__NOTIFY || !req->notify_success) {
+		pr_err("RPC client reported script error\n");
+		return -1;
+	}
+
+	ret = 0;
+	if (req->opts)
+		for (i = 0; i < req->opts->n_external; i++) {
+			char *key = xstrdup(req->opts->external[i]);
```
avagin, line 272: xstrdup isn't required here.

criu/cr-service.c
```diff
@@ -240,6 +240,50 @@ int send_criu_rpc_script(enum script_actions act, char *name, int sk, int fd)
...
+
+	ret = send_criu_msg_with_fd(sk, &msg, -1);
+	if (ret < 0)
+		return ret;
+
+	ret = recv_criu_msg_with_fd(sk, &req, NULL);
+	if (ret < 0)
+		return ret;
+
+	if (req->type != CRIU_REQ_TYPE__NOTIFY || !req->notify_success) {
```
avagin, line 264: nit: I would prefer to have a separate type for this request.

criu/cr-service.c
osctobe: Our RPC client process is in a different pidns and so CRIU can't open its /proc/<pid>/fd/X. For `images_dir_fd` to be useful here it would need to refer to a passed or CRIU's fd.

images/rpc.proto
```diff
@@ -60,7 +60,8 @@ enum criu_pre_dump_mode {
 };
 
 message criu_opts {
-	required int32			images_dir_fd	= 1;
+	optional int32			images_dir_fd	= 1;
```
avagin, line 63: is it safe to convert `required` to `optinal`? rpc.proto has to be backward compatible.

criu/cr-service.c
avagin: Got it. We can consider sending fd-s in scm.

criu/action-scripts.c
```diff
@@ -31,6 +31,7 @@ static const char *action_names[ACT_MAX] = {
 	[ACT_POST_RESUME] = "post-resume",
 	[ACT_ORPHAN_PTS_MASTER] = "orphan-pts-master",
 	[ACT_STATUS_READY] = "status-ready",
+	[ACT_QUERY_EXT_FILES] = "query-ext-files",
```
avagin, line 34: pls add a description of this new hook in Documentation/criu.txt.

osctobe: Done.

images/rpc.proto
osctobe: The old CRIU with a library using newer rpc.proto would fail, but new CRIU with old library would work the same. I've changed it so that older-CRIU scenario will fail only if using the images_dir. (Assuming code does not set both)

Documentation/criu.txt
```diff
@@ -155,6 +155,11 @@ not compatible with *--external* *dev*.
             notification message contains a file descriptor for
             the master pty
 
+        *query-ext-files*:::
+            called after the processes are stoppend and network is locked.
```
rst0git, line 159: *stopped

PR #2200, opened 2023-06-15
Diagnose capability changes if migrating between kernels of different cap_last_cap.

osctobe:
Log when trying to restore capabilities unknown to the kernel.

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2940,12 +2940,6 @@ out:
 	return ret;
 }
 
-static inline int verify_cap_size(CredsEntry *ce)
-{
-	return ((ce->n_cap_inh == CR_CAP_SIZE) && (ce->n_cap_eff == CR_CAP_SIZE) && (ce->n_cap_prm == CR_CAP_SIZE) &&
-		(ce->n_cap_bnd == CR_CAP_SIZE));
-}
-
 static int prepare_mm(pid_t pid, struct task_restore_args *args)
 {
 	int exe_fd, i, ret = -1;
@@ -3360,16 +3354,30 @@ static bool groups_match(gid_t *groups, int n_groups)
 	return ret;
 }
 
+static void copy_caps(u32 *out_caps, u32 *in_caps, int n_words)
+{
+	int i, cap_end;
+
+	for (i = kdat.last_cap + 1; i < 32 * n_words; ++i) {
+		if (~in_caps[i / 32] & (1 << (i % 32)))
+			continue;
+
+		pr_warn("Dropping unsupported capability %d > %d)\n", i, kdat.last_cap);
+		/* extra caps will be cleared below */
+	}
+
+	n_words = min(n_words, (kdat.last_cap + 31) / 32);
+	cap_end = (kdat.last_cap & 31) + 1;
+	memcpy(out_caps, in_caps, sizeof(*out_caps) * n_words);
+	if ((cap_end & 31) && n_words)
+		out_caps[n_words - 1] &= (1 << cap_end) - 1;
+	memset(out_caps + n_words, 0, sizeof(*out_caps) * (CR_CAP_SIZE - n_words));
+}
+
 static struct thread_creds_args *rst_prep_creds_args(CredsEntry *ce, unsigned long *prev_pos)
 {
 	unsigned long this_pos;
 	struct thread_creds_args *args;
-
-	if (!verify_cap_size(ce)) {
-		pr_err("Caps size mismatch %d %d %d %d\n", (int)ce->n_cap_inh, (int)ce->n_cap_eff, (int)ce->n_cap_prm,
-		       (int)ce->n_cap_bnd);
-		return ERR_PTR(-EINVAL);
-	}
 
 	this_pos = rst_mem_align_cpos(RM_PRIVATE);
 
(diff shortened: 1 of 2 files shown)


criu/cr-restore.c
```diff
@@ -3377,7 +3377,15 @@ static struct thread_creds_args *rst_prep_creds_args(CredsEntry *ce, unsigned lo
 	if (!args)
 		return ERR_PTR(-ENOMEM);
 
-	args->cap_last_cap = kdat.last_cap;
+	/* When restoring capability bounding set, don't attempt to mask out
+	 * caps which weren't known at the dumping side but are known at the
+	 * restore side - the dumped process couldn't have set that mask, and
+	 * the default is to not mask them out. Also in the opposite direction:
```
avagin, line 3383: >  the dumped process couldn't have set that mask, and the default is to not mask them out.

hmm, it looks more logical to drop all new caps from restored bounding sets, doesn't it?

avagin, line 3383: >  the dumped process couldn't have set that mask, and the default is to not mask them out.

hmm, it looks more logical to drop all new caps from restored bounding sets, doesn't it?

avagin: The idea is that restored processes should not be able to use new capabilities.

osctobe: The reasoning here is that if a checkpointing kernel knew more caps than restoring, then the extra caps can't be used nor disabled on the restoring side; if the restoring one knows more caps, then the extra caps coundn't have been disabled on the checkpointing side. (Note: there is no way to add caps to the bounding set.)

avagin: >  if the restoring one knows more caps, then the extra caps coundn't have been disabled on the checkpointing side. (Note: there is no way to add caps to the bounding set.)

but the app doesn't rely on new caps and so they can be dropped from the bounding set without affecting the app behavior.

osctobe: Which app do you have in mind? If it calls a setuid (external machine-specific file) binary from time to time, it will break if the external binary needs those caps. Bounding set is not what modifies the C/Rd app itself but the processes it starts. (BTW, we use app-assisted C/R, so the app can delay the checkpoint until after some external command finishes if necessary.)

avagin: > Which app do you have in mind? 

@osctobe runc (used by docker, podman, k8s, etc) starts containers with bounding sets that are specified in configs:
https://github.com/opencontainers/runtime-spec/blob/48415de180cf7d5168ca53a5aa27b6fcec8e4d81/specs-go/config.go#L103

> (BTW, we use app-assisted C/R, so the app can delay the checkpoint until after some external command finishes if necessary.)

I understand the problem that you are trying to solve, but here is the rule for such cases: in case of any uncertainty, we choose the variant that doesn't weaken security.

Here is an example. We have a process that is running without capabilities and with the empty bonding set. Then we migrate it to another node with a new kernel that has one new capability. Should the process be able to run any tools that  use this new capability. In general, the answer is no (it is the default case). In some cases like you described, the answer could be yes and we can help to figure out how to handle such cases.

osctobe: I've dug into the source of this patch and it looks it was fixing something else entirely. I redid the PR. We'll probably need more changes to allow unprivileged restore in some cases.

criu/cr-restore.c
```diff
@@ -3360,16 +3354,35 @@ static bool groups_match(gid_t *groups, int n_groups)
...
+
+	for (i = last_cap + 1; i < 32 * n_words; ++i) {
+		if (~in_caps[i / 32] & (1 << (i % 32)))
+			continue;
+
+		if (!opts.unprivileged) {
+			pr_err("Can't restore capability %d > min(kernel=%d, criu=%d)", i, kdat.last_cap, 32 * CR_CAP_SIZE - 1);
+			return -EPERM;
+		}
+		pr_warn("Can't restore capability %d > min(kernel=%d, criu=%d)", i, kdat.last_cap, 32 * CR_CAP_SIZE - 1);
```
avagin, line 3370: The commit message (and comment in this case) has to explain why it must work this way. Right now, it is unclear why it doesn't fail if unprivileged is set.

avagin:
> Change-Id: Iee9efb7d46e4a40558ab21035d352d6dcb74dc6a

nit: pls remove these Change-Id tags.

criu/cr-restore.c
```diff
@@ -3360,16 +3354,31 @@ static bool groups_match(gid_t *groups, int n_groups)
 	return ret;
 }
 
+static int copy_caps(u32 *out_caps, u32 *in_caps, int n_words)
+{
+	int last_cap = min(kdat.last_cap, 32 * CR_CAP_SIZE - 1);
```
avagin, line 3359: sounds like a bug. You need to add a comment to explain when it can happen.

criu/cr-restore.c
```diff
@@ -3360,16 +3354,31 @@ static bool groups_match(gid_t *groups, int n_groups)
...
+	int i;
+
+	for (i = last_cap + 1; i < 32 * n_words; ++i) {
+		if (~in_caps[i / 32] & (1 << (i % 32)))
+			continue;
+
+		pr_err("Can't restore capability %d > min(kernel=%d, criu=%d)", i, kdat.last_cap, 32 * CR_CAP_SIZE - 1);
```
avagin, line 3366: Need to add \n to the end of the message.

criu/cr-restore.c
osctobe: Added a check to kerndat instead.

avagin:
Merged. Thanks

PR #2207, opened 2023-06-21
Make some no-op permission changes not requiring privilege.

osctobe:


criu/util.c
```diff
@@ -952,6 +952,83 @@ FILE *fopenat(int dirfd, char *path, char *cflags)
...
 
+int fchown_nopriv(int fd, uid_t new_uid, gid_t new_gid)
+{
+	struct stat st;
+
+	if (fstat(fd, &st) == -1)
+		return -1;
+	pr_debug("fstat(%d): uid %u gid %u\n", fd, st.st_uid, st.st_gid);
+
+	if (new_uid == st.st_uid && new_gid == st.st_gid)
```
avagin, line 963: let's do fstat and these checks only if fchown fails.

criu/pie/restorer.c
```diff
@@ -221,6 +219,20 @@ static int restore_creds(struct thread_creds_args *args, int procfd, int lsm_typ
 		}
 	}
 
+	/*
+	 * Compare xids with current values. If all match then we can skip
+	 * setting them (which requires extra capabilities). There's no system
+	 * call to retrieve the fs ids so we'll just compare with the effective
```
avagin, line 225: setfsuid returns the current fsuid. setfsuid((uid_t) -1) returns the current value without doing anything else.

criu/pie/restorer.c
```diff
@@ -200,10 +205,19 @@ static int restore_creds(struct thread_creds_args *args, int procfd, int lsm_typ
...
-			pr_err("Can't setup supplementary group IDs: %d\n", ret);
-			return -1;
+		/*
+		 * We may be in an unprivileged user namespace where setgroups
+		 * is disabled.  If the current list of groups is already what
+		 * we want, skip the call to setgroups.
+		 */
+		unsigned int gids[MAX_GETGROUPS_CHECKED];
+		int n = sys_getgroups(MAX_GETGROUPS_CHECKED, gids);
+		if (n != ce->n_groups || memcmp(gids, args->groups, n * sizeof(*gids))) {
```
avagin, line 213: do we have any guarantees that the groups list is sorted?

criu/util.c
```diff
@@ -952,6 +952,88 @@ FILE *fopenat(int dirfd, char *path, char *cflags)
 	return fdopen(tmp, cflags);
 }
 
+int fchown_nopriv(int fd, uid_t new_uid, gid_t new_gid)
```
avagin, line 955: imho, this nopriv postfix is misleading. I would suggest to use the cr prefix for all this wrappers (cr_fchodn, cr_fchperm, etc).

criu/pie/restorer.c
osctobe: According to a comment in `groups_match()` in 'cr-restore.c', `setgroups()` does sort the gids. I can't see that information in the manpage, though.

avagin:
I have merged this series with one minor fix of the lint warning. We need to figure out how to test these new calls.

PR #2208, opened 2023-06-21
pipes: Plug pipe fd leak in "Unable to set a pipe size" error case.

osctobe:
A trivial fix for error path leaking pipe fds.


mihalicyn:
LGTM

but please fix https://github.com/checkpoint-restore/criu/pull/2208/checks?check_run_id=14450206848

rst0git (approved):

PR #2209, opened 2023-06-21
sockets: Increase the size of sockets hashmap to 16K.

osctobe:
During dump, CRIU stores the structs representing sockets in a statically sized hashmap of size 32. We have some (admittedly crazy) tasks that use tens of thousands of sockets, and seem to spend most of the dump time iterating over the linked lists of the map.

16K is chosen arbitrarily, so that it reduces the lengths of the chains to few elements on average, while not introducing significant memory overhead.


mihalicyn:
please fix https://github.com/checkpoint-restore/criu/pull/2209/checks?check_run_id=14450551879

rst0git (approved):
LGTM

Issue #2203, opened 2023-06-19
zdtm/centos 7: wait() got an unexpected keyword argument 'timeout'


rst0git:
After commit https://github.com/checkpoint-restore/criu/commit/0bd5abe4ed331c96fcc6d4da5fc4a60ece0045e2, the CentOS 7 CI tests fail with the following error:

```
Traceback (most recent call last):
  File "./test/zdtm.py", line 2854, in <module>
    opts['action'](opts)
  File "./test/zdtm.py", line 2427, in run_tests
    features[feat] = criu.check(feat)
  File "./test/zdtm.py", line 1507, in check
    return criu_cli.run("check", args, opts['criu_bin']) == 0
  File "./test/zdtm.py", line 903, in run
    return cr.wait(timeout=timeout)
TypeError: wait() got an unexpected keyword argument 'timeout'
```
The reason for this error is because in CentOS 7 we run the tests with Python 2.7 and with this version [wait()](https://docs.python.org/2.7/library/subprocess.html#subprocess.Popen.wait) does not support a `timeout` option.

This version of python has been deprecated for a few years now. Should we introduce a workaround to fix this problem or simply drop compatibility with Python 2?

avagin:
To: @osctobe

avagin:
@rst0git @adrianreber  RHEL7/Centos7 is in the maintenance mode, I think we can stop testing new versions of CRIU on it. WDYT?

avagin:
python2 reached eof too, so I think we need to drop its support from criu too.

avagin:
python2 reached eof too, so I think we need to drop its support from criu too.

adrianreber:
I agree. Last time we talked about it @Snorch indicated that it might still be needed for openvz. From my point of view we can ignore CentOS 7 at this point in time.

Snorch:
Initially I was against dropping python2 as we still use it in VZ7 based on Centos7
https://github.com/checkpoint-restore/criu/issues/1405
But then we have ported newer protobuf-compiler package and python3 should work on VZ7 now.
https://github.com/checkpoint-restore/criu/pull/1733#discussion_r809887414

So I also agree. (though I would probably get some more pain on my ass on next rebase =))

rst0git:
Great! I will open a pull request to drop Python 2 and CentOS 7 :)

PR #2210, opened 2023-06-21
Remove support for Python 2

rst0git:
This pull request removes support for Python 2 in CRIU and disables the CI tests for CentOS 7, as we discussed in https://github.com/checkpoint-restore/criu/issues/2203.


rst0git:
> This is a huge change, but it looks correct.

Indeed, these changes clean-up Python 2 compatibility fixes we have introduced over the past few years.

> Not entirely sure if this might break packaging CRIU in some places. Hard to tell.

I don't expect these changes to break packaging as we have already added all necessary build dependencies for Python 3 in the 3.18 release.

lib/Makefile
```diff
@@ -56,7 +56,8 @@ install: lib-c lib-a lib-py lib/c/criu.pc.in
 	$(Q) install -m 644 lib/c/criu.pc $(DESTDIR)$(LIBDIR)/pkgconfig
 ifeq ($(PYTHON),python3)
 	$(E) "  INSTALL " crit
-	$(Q) $(PYTHON) -m pip install --upgrade --force-reinstall --prefix=$(DESTDIR)$(PREFIX) ./crit
+	$(Q) PIP_BREAK_SYSTEM_PACKAGES=1 $(PYTHON) -m pip install -r ./crit/requirements.txt
```
avagin, line 58: I don't think that this is a good idea to set such variables here.  Let's users to decide whether they wait to "break their systems".

rst0git: If we remove this variable `pip install` / `pip uninstall` fails on Arch linux.
Example: https://github.com/checkpoint-restore/criu/actions/runs/5382664987/jobs/9768409154?pr=2217
```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try 'pacman -S
    python-xyz', where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Arch-packaged Python package,
    create a virtual environment using 'python -m venv path/to/venv'.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.
    
    If you wish to install a non-Arch packaged Python application,
    it may be easiest to use 'pipx install xyz', which will manage a
    virtual environment for you. Make sure you have python-pipx
    installed via pacman.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
Error uninstalling package crit: Command '['/usr/sbin/python3', '-m', 'pip', 'uninstall', '-y', 'crit']' returned non-zero exit status 1.
```

Should we skip the install / uninstall of CRIT with externally managed Python environment when `PIP_BREAK_SYSTEM_PACKAGES` is not set?

avagin: If a user decides to ignore the error, they can set the variable and execute make:
$ PIP_BREAK_SYSTEM_PACKAGES=1 make install

or
$ export PIP_BREAK_SYSTEM_PACKAGES=1
$ make install

adrianreber: I agree, we should not set this variable. Maybe if the installation fails you could print a message about the existence of the variable.

rst0git: Thanks Adnrei and Adrian! I have updated the pull request.

PR #2212, opened 2023-06-22
cgroup: Remove deprecated property handling.

osctobe:
`memory.kmem.limit_in_bytes` knob is deprecated since Linux 5.4 and not writable since Linux 5.16.

### DIFF
--- a/criu/cgroup-props.c
+++ b/criu/cgroup-props.c
@@ -427,7 +427,6 @@ static int cgp_parse_builtins(void)
 						"\"memory.move_charge_at_immigrate\", "
 						"\"memory.oom_control\", "
 						"\"memory.use_hierarchy\", "
-						"\"memory.kmem.limit_in_bytes\", "
 						"\"memory.kmem.tcp.limit_in_bytes\" "
 						"]\n"
 						/*


avagin:
rhel8 hasn't reached EOL and it is based on the 4.18 kernel.

@osctobe Does memory.kmem.limit_in_bytes trigger any problems in your environment?

osctobe:
Other than the tests failing on 5.10-based kernel, currently no (as we migrate to the same kernel). But we'd have to check with other users that do migrate between kernel major version.

adrianreber:
This is still a thing in RHEL 8:
```
# cd /sys/fs/cgroup/memory/
# cat memory.kmem.limit_in_bytes
9223372036854771712
```

I am not sure but I think I have also seen warnings/error messages concerning this entry on newer kernels. Cannot really remember.

I found it. On a Fedora 6.2.14-300.fc38.x86_64 kernel with cgroup v1. The property exists during checkpointing:
```
(00.122069) cg: Dumping value 9223372036854771712 from /proc/self/fd/19/machine.slice/libpod-9bc60d19cfa3fb6b5ccc25dc7a866765e67a44040d381c34f69f113e166bb2fa.scope/container/memory.kmem.limit_in_bytes
```
But restoring of that property fails:
```
(00.001748) cg: Restoring cgroup property value [9223372036854771712] to [memory/machine.slice/libpod-9bc60d19cfa3fb6b5ccc25dc7a866765e67a44040d381c34f69f113e166bb2fa.scope/container/memory.kmem.limit_in_bytes]
(00.001769) Error (criu/cgroup.c:1357): cg: Failed writing 9223372036854771712 to memory/machine.slice/libpod-9bc60d19cfa3fb6b5ccc25dc7a866765e67a44040d381c34f69f113e166bb2fa.scope/container/memory.kmem.limit_in_bytes: Operation not supported
(00.001776) Error (criu/cgroup.c:1624): cg: Restoring memory.kmem.limit_in_bytes special property failed
```

Which is probably the reason for this PR. It is not writeable. Is there a way to probe if it is writeable?

Snorch:
Might it be that https://github.com/checkpoint-restore/criu/commit/21f5be91a90a420c7e183b61f0a64ab7f0f8f137 already fixed this problem?

osctobe:
> Might it be that [21f5be9](https://github.com/checkpoint-restore/criu/commit/21f5be91a90a420c7e183b61f0a64ab7f0f8f137) already fixed this problem?

It looks about right. I wanted to also avoid WARNs showing up in dmesg when reading this knob.

PR #2214, opened 2023-06-23
restore: remove unused `secbits` field.

osctobe:


criu/include/restorer.h
```diff
@@ -72,7 +72,6 @@ struct thread_creds_args {
 	u32 cap_eff[CR_CAP_SIZE];
 	u32 cap_bnd[CR_CAP_SIZE];
 
-	unsigned int secbits;
```
rst0git, line 75: The value of the `secbits` field appears to be used in the following line within the `restore_creds()` function.
https://github.com/checkpoint-restore/criu/blob/f5cd44f79737fb91cacd28bd4b6d0abdca1e180b/criu/pie/restorer.c#L259

osctobe: This uses `secbits` from `CredsEntry` not from `thread_creds_args`.

rst0git: It looks like the parameter `args` is a pointer to a structure of type `struct thread_creds_args`.
```C
static int restore_creds(struct thread_creds_args *args, int procfd, int lsm_type, uid_t uid)
{
	CredsEntry *ce = &args->creds;
	...
```

rst0git (approved):

PR #2215, opened 2023-06-23
dump: Don't unfreeze tasks on dump failure with --no-resume-on-error.

osctobe:
Make it possible to kill or leave stopped tasks if a dump failed after stopping the tree.


criu/crtools.c
```diff
@@ -394,6 +394,8 @@ int main(int argc, char *argv[], char *envp[])
 	       "  -d|--restore-detached detach after restore\n"
 	       "  -S|--restore-sibling  restore root task as sibling\n"
 	       "  -s|--leave-stopped    leave tasks in stopped state after checkpoint\n"
+	       "     --no-resume-on-error\n"
+	       "                        don't resume tasks on dump failure if they were stopped\n"
```
rst0git, line 398: Could you also add a description of this option in `Documentation/criu.txt`?

rst0git:
@osctobe Would it be possible to add a test for this functionality?

avagin:
> > @osctobe Would it be possible to add a test for this functionality?
> 
> There are no tests for --leave-stopped or --leave-running yet that could be extended with this case. 

Here is the test for --leave-stopped:
https://github.com/checkpoint-restore/criu/blob/criu-dev/test/jenkins/criu-stop.sh

> The change is tested in production (always enabled), though.

I am sorry, but it doesn't work this way. I think our fault injection engine can be used to introduce a test.  test/jenkins/criu-fault.sh contains all these tests.

0x7f454c46:
You keep adding `Change-Id: Ia8956063cdc130650cfcde86851ee6a14331f2c2` that pollute git logs and don't provide anything outside your company. Clean these up, please.

Documentation/criu.txt
```diff
@@ -213,6 +213,9 @@ In other words, do not use it unless really needed.
 *-s*, *--leave-stopped*::
     Leave tasks in stopped state after checkpoint, instead of killing.
 
+*--no-resume-on-error*::
+    Leave tasks in stopped state even if checkpoint completed unsuccessfully.
+
```
0x7f454c46, line 218: Could you add some lines about what situations it is useful and why? What benefit does it provide?
How should I as a user decide what is preferred for them in case of criu failure?
Current doc lines are copy'n'paste from above without adding any useful context.

criu/cr-dump.c
```diff
@@ -2065,7 +2065,9 @@ static int cr_dump_finish(int ret)
 	 *    consistency of the FS and other resources, we simply
 	 *    start rollback procedure and cleanup everything.
 	 */
-	if (ret || post_dump_ret || opts.final_state == TASK_ALIVE) {
+	if (opts.resume_on_dump_error && (ret || post_dump_ret))
+		opts.final_state = TASK_ALIVE;
+	if (opts.final_state == TASK_ALIVE) {
```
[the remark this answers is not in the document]
Snorch: Yes, it is kinda strange that we don't unlock network for --leave-stopped. One would not be able to just resume such stopped container with SIGCONT, but will also need to manually unlock network and remove link remaps and do other cleanups.

Snorch:
See the freezer_restore_state() related code. If before dump you put your processes in freezer cgroup and make it FROZEN, you can later decide after dump finishes if you want to make cgroup THAWED (no dump failure) or leave it frozen (on dump failure). This does effectively the same as you want to accomplish with this new option.

avagin:
@osctobe could you response to comments?

PR #2217, opened 2023-06-24
action-scripts: allow shell scripts in rpc mode

rst0git:
Container runtimes commonly use CRIU with RPC. However, this prevents the use of action-scripts set in a configuration file due to the explicit scripts mode introduced with commit ac78f13bdfaee260dd4234f054bf4c5d2a373783. This pull request enables container checkpoint/restore with action-scripts specified via configuration file.

### DIFF
--- a/criu/action-scripts.c
+++ b/criu/action-scripts.c
@@ -52,6 +52,9 @@ static int run_shell_scripts(const char *action)
 #define ENV_IMGDIR  0x1
 #define ENV_ROOTPID 0x2
 
+	if (list_empty(&scripts))
+		return 0;
+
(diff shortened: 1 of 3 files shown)


criu/action-scripts.c
```diff
@@ -124,10 +124,11 @@ int run_scripts(enum script_actions act)
 
 	if (scripts_mode == SCRIPTS_RPC) {
 		ret = rpc_send_fd(act, -1);
-		goto out;
+		if (ret)
+			goto out;
 	}
 
-	if (scripts_mode == SCRIPTS_SHELL) {
+	if (!list_empty(&scripts)) {
```
avagin, line 131: should we convert these if-s to switch?

PR #2219, opened 2023-07-01
ci: add workflow to ensure self-contained commits

### LINKED ISSUE #2216: Add CI workflow to ensure self-contained commits
Currently, podman and checkpointctl both use a CI workflow to determine the delta in binary size with and without the commits in a PR. If it is beyond a certain threshold, the workflow fails, alerting maintainers to take a closer look at the code. If this is not already present in CRIU, it would be beneficial to track it. A good side effect is that it also ensures every commit in a PR is self-contained.

cc @rst0git @adrianreber

snprajwal:
The workflow builds CRIU at each commit in the PR, and fails if any of the commits are not self-contained. This will ensure that we can reliably use `git bisect` while debugging anything in the repo.

Fixes #2216


.github/workflows/check-commits.yml
```diff
@@ -0,0 +1,30 @@
...
+        # Checkout pull request HEAD commit instead of merge commit
+        ref: ${{ github.event.pull_request.head.sha }}
+    - name: Install dependencies
+      run: sudo apt-get install -y libprotobuf-dev libprotobuf-c-dev protobuf-c-compiler protobuf-compiler python3-protobuf libnl-3-dev libnet-dev libcap-dev
+    - name: Configure git user details
+      run: |
+        git config --global user.email "checkpoint-restore@users.noreply.github.com"
+        git config --global user.name "checkpoint-restore"
+    - name: Configure base branch without switching current branch
+      run: git fetch origin ${GITHUB_BASE_REF}:${GITHUB_BASE_REF}
```
Snorch, line 28: Please see https://github.com/checkpoint-restore/criu/blob/criu-dev/.github/workflows/lint.yml#L30

I understand that both `${{github.base_ref}}` and `${GITHUB_BASE_REF}` are possible, but to have easier grep-ing maybe we can use one name everywhere?

Snorch:
Adding some more info on which commit the workflow failed would be nice, see https://github.com/checkpoint-restore/criu/actions/runs/5462129746/jobs/9941059153?pr=2220

snprajwal:
I hope this works, `git rebase` is slightly iffy about what can be passed as an exec command (no newlines, apparently).

Snorch (approved):
Looks good!

Issue #2222, opened 2023-07-06
CRIU dump/check fails for a process with "kerndat_has_ipv6_freebind failed when initializing kerndat"


yevzman-swe:
**Description**
Hi,
Today while using CRIU (**v3.18**) I got an error:
` "Unable to create a ipv6 dgram socket: Address family not supported by protocol"`, when dumping the process and also when trying to do 
```python
criu check
```

**Steps to reproduce the issue:**
1. A machine with disabled IPv6. 
Artificially can be reproduced by editing value of variables in **/etc/default/grub** file:
```
GRUB_CMDLINE_LINUX_DEFAULT="ipv6.disable=1"
# (If there were any other parameters here, such as "quiet splash," simply separate that parameter from the new one with a space (e.g., GRUB_CMDLINE_LINUX_DEFAULT="quiet spash ipv6.disable=1"))
GRUB_CMDLINE_LINUX="ipv6.disable=1"
```
  and furthere system rebooting:
```bash
sudo reboot
```
2. Run command:
```bash
criu check
```
or dump process
```bash
criu dump -t $PID
```

**Describe the results you received:**
for both `criu check` and `criu dump` i got
```bash
Error (criu/kerndat.c:1581): Unable to create a ipv6 dgram socket: Address family not supported by protocol
Error (criu/kerndat.c:1839): kerndat_has_ipv6_freebind failed when initializing kerndat.
Error (criu/crtools.c:263): Could not initialize kernel features detection.
```
**Describe the results you expected:**
for `criu check`
```
Looks good.
```
for `criu dump`: Successful dump

**CRIU logs and information:**

<details><summary>Output of `criu --version`:</summary>
<p>

```
Version: 3.18
GitID: v3.18-57-g504d05197
```

</p>
</details>

<details><summary>Output of `criu check --all`:</summary>
<p>

```
Error (criu/kerndat.c:1581): Unable to create a ipv6 dgram socket: Address family not supported by protocol
Error (criu/kerndat.c:1839): kerndat_has_ipv6_freebind failed when initializing kerndat.
Error (criu/crtools.c:263): Could not initialize kernel features detection.

```

</p>
</details>

**Additional environment details:**
result of command `uname -a`
```
Linux 5.19.0-46-generic #47~22.04.1-Ubuntu SMP Wed Jun 21 15:35:31 UTC 2 x86_64 x86_64 x86_64 GNU/Linux
```

PR #2223, opened 2023-07-06
kerndat: bind ipv6-socket only if ipv6 is enabled

yevzman-swe:
Fixes: #2222
Fixes: f1c8d38 ("kerndat: check if setsockopt IPV6_FREEBIND is supported")


avagin:
@avagin closed this as completed in a4bb3f9

PR #2226, opened 2023-07-07
kerndat: don't leak a socket file descriptor

avagin:
kerndat_has_ipv6_freebind creates a socket but doesn't close it.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1574,9 +1574,26 @@ static int kerndat_has_nftables_concat(void)
 #define IPV6_FREEBIND 78
 #endif
 
+static int __kerndat_has_ipv6_freebind(int sk)
+{
+	int val = 1;
+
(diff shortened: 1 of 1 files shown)


rst0git (approved):
LGTM

criu/kerndat.c
```diff
@@ -1574,20 +1574,9 @@ static int kerndat_has_nftables_concat(void)
...
-	}
-
-	sk = socket(AF_INET6, SOCK_DGRAM, IPPROTO_UDP);
-	if (sk == -1) {
-		pr_perror("Unable to create a ipv6 dgram socket");
-		return -1;
-	}
+	int val;
 
 	val = 1;
```
Snorch, line 1581: We can save one line with `int val = 1;`.

Snorch (approved):

PR #2230, opened 2023-07-27
C/R memfd file permissions

osctobe:


criu/memfd.c
```diff
@@ -279,8 +281,13 @@ static int memfd_open_inode_nocache(struct memfd_restore_inode *inode)
...
 
-	if (cr_fchown(fd, mie->uid, mie->gid)) {
-		pr_perror("Can't change uid %d gid %d of memfd:%s", (int)mie->uid, (int)mie->gid, mie->name);
+	if (mie->has_mode)
+		ret = cr_fchperm(fd, mie->uid, mie->gid, mie->mode);
+	else
+		ret = cr_fchown(fd, mie->uid, mie->gid);
+	if (ret) {
+		pr_perror("Can't set permissions { uid %d gid %d mode %#o } of memfd:%s", (int)mie->uid,
+			  (int)mie->gid, mie->has_mode ? (int)mie->mode : -1, mie->name);
```
rst0git, line 290: When `mie->has_mode` is false, the output of `printf("%#o\n", -1)` might not be very helpful.

avagin:
pls write in the commit message when memfd file permissions are matter. Commit messages can't be empty except very trivial patches.

criu/memfd.c
osctobe: It should be obvious, though, that this isn't a valid mode, and it could be generated in `!mie->has_mode` or a broken image cases.

osctobe:
Done.

PR #2231, opened 2023-07-27
Preserve memfd exec()ability

osctobe:
We have a user (Sandbox2) which creates a memfd, writes some binary there and then calls `execveat(fd, ... AT_EMPTY_PATH)`. Currently, if CRIU catches the process between `memfd_create()` and `execveat()`, then after restore the latter call will fail with `ETXTBUSY`, as fd points not to the original memfd, but a reopened `/proc/*/fd`.


Snorch:
Obviousely you need to (same to all other your PRs):
- fix CI
  - lint
  - your test which is failing, I don't feel like the test is right as it even fails without any checkpointing

test/zdtm/static/memfd04.c
```diff
@@ -0,0 +1,108 @@
...
+
+	if (status != 0) {
+		pr_err("child exited with status=%d\n", status);
+		return false;
+	}
+
+	return true;
+}
+
+static const char script[] = "#!/bin/true";
```
avagin, line 74: You need to add /bin/true to dependencies:
```
diff --git a/test/zdtm/static/memfd04.desc b/test/zdtm/static/memfd04.desc
new file mode 100644
index 000000000..0e3cbb9f0
--- /dev/null
+++ b/test/zdtm/static/memfd04.desc
@@ -0,0 +1,3 @@
+{
+  'deps': ['/bin/true'],
+}
```

criu/memfd.c
```diff
@@ -335,18 +335,23 @@ int memfd_open(struct file_desc *d, u32 *fdflags)
...
+		 * important though.
+		 */
+		_fd = __open_proc(PROC_SELF, 0, flags, "fd/%d", fd);
+		if (_fd < 0) {
+			pr_perror("Can't reopen memfd id=%d", mfe->id);
+			goto err;
+		}
+		close(fd);
+		fd = _fd;
+	} else if (fcntl(fd, F_SETFL, flags)) {
```
avagin, line 351: We can do this only once for each inode.  If this memfd_open is called more than one time for one memfd, it means that we have two non-shared file descriptors.

osctobe: Done.

criu/memfd.c
```diff
@@ -339,6 +341,21 @@ int memfd_open(struct file_desc *d, u32 *fdflags)
 
 	/* Reopen the fd with original permissions */
 	flags = fdflags ? *fdflags : mfe->flags;
+
+	if (!mfi->inode->was_opened_rw && (flags & O_ACCMODE) == O_RDWR) {
+		/* HACK WARNING *
```
avagin, line 346: DONT PANIC!

pls remove this warning and add an explanation why we are doing this.

avagin:
lgtm

PR #2237, opened 2023-08-04
memfd: test fd-sharing for memfds

osctobe:
Add a test that verifies that memfd's are properly unshared to validate incoming executable-memfd fix.


avagin:
@osctobe why do you decide to close this pr?

PR #2238, opened 2023-08-07
tty: don't restore locking with --unprivileged

rst0git:
`TIOCSLCKTRMIOS` requires [CAP_SYS_ADMIN](https://github.com/torvalds/linux/blob/52a93d39b17dc7eb98b6aa3edb93943248e03b2f/drivers/tty/tty_ioctl.c#L863). This causes `criu restore` to fail when used with `--unprivileged`.

### DIFF
--- a/criu/tty.c
+++ b/criu/tty.c
@@ -817,8 +817,10 @@ static int do_restore_tty_parms(void *arg, int fd, pid_t pid)
 	 * on termios too. Just to be on the safe side.
 	 */
 
-	if ((p->has & HAS_TERMIOS_L) && ioctl(fd, TIOCSLCKTRMIOS, &p->tl) < 0)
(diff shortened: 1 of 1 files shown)


avagin:
Pls add a full description what this ioctl is about and why it is safe to do what you are doing in this patch.

avagin:
@rst0git ping

rst0git:
> Pls add a full description what this ioctl is about and why it is safe to do what you are doing in this patch.

Thank you for the question. I investigated this further and while skipping `TIOCSLCKTRMIOS` for unprivileged restore seems to work, it might not be safe. Instead, it seems that allowing `CAP_CHECKPOINT_RESTORE` to be used with `TIOCSLCKTRMIOS` is more appropriate.

PR #2242, opened 2023-08-09
Put a cap on the size of single preadv in restore operation.

osctobe:
While each preadv() is followed by a fallocate() that removes the data range from image files on tmpfs, temporarily (between preadv() and fallocate()) the same data is in two places; this increases the memory overhead of restore operation by the size of a single preadv. Uncapped preadv() would read up to 2 GiB of data, thus we limit that to a smaller block size (128 MiB).

### DIFF
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -56,6 +56,12 @@
  */
 #define MAX_GETGROUPS_CHECKED (512 / sizeof(unsigned int))
 
+/*
+ * Memory overhead limit for reading VMA when auto_dedup is enabled.
+ * An arbitrarily chosen trade-off point between speed and memory usage.
+ */
+#define AUTO_DEDUP_OVERHEAD_BYTES (128 << 20)
+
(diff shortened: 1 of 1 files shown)


criu/pie/restorer.c
```diff
@@ -1724,7 +1758,11 @@ long __export_restore_task(struct task_restore_args *args)
 
 		while (nr) {
 			pr_debug("Preadv %lx:%d... (%d iovs)\n", (unsigned long)iovs->iov_base, (int)iovs->iov_len, nr);
-			r = sys_preadv(args->vma_ios_fd, iovs, nr, rio->off);
+			/*
+			 * If we're requested to punch holes in the file after reading we do
+			 * it to save memory. Limit the reads then to an arbitrary block size.
+			 */
+			r = preadv_limited(args->vma_ios_fd, iovs, nr, rio->off, args->auto_dedup ? 128 << 20 : 0);
```
avagin, line 1765: You need to define a constant for 128 << 20 and support it with a comment to explain its meaning and value.

PR #2247, opened 2023-08-17
memfd_secret: add memfd-secret file support

### LINKED ISSUE #2188: CRIU can't dump `memfd_secret` fd containing process
**dumpee src:**
```
#include <stdio.h>
#include <string.h>
#include <sys/mman.h>
#include <sys/syscall.h>
#include <unistd.h>

static int memfd_secret(unsigned int flags)
{
	return syscall(SYS_memfd_secret, flags);
}

static void *secret_init(size_t size)
{
	int fd;
	void *secret_mem = NULL;

	fd = memfd_secret(0);
	if (fd < 0)
		return secret_mem;

	if (ftruncate(fd, size) < 0) {
		close(fd);
		return secret_mem;
	}

	secret_mem = mmap(NULL, size, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
	if (secret_mem == MAP_FAILED) {
		close(fd);
		return secret_mem;
	}

	return secret_mem;
}

static void secret_fini(void *mem, size_t size)
{
	munmap(mem, size);
}

int main(int argc, char *argv[])
{
	const char secret_msg[] = "CRIU";
	size_t size = sizeof(secret_msg);
	void *secret_mem;

	fprintf(stdout, "ppid: %d\n", getppid());
    	fprintf(stdout, "pid: %d\n", getpid());

	secret_mem = secret_init(size);
	if (! secret_mem)
		perror("not supported operation");

	strncpy(secret_mem, secret_msg, size);
	fprintf(stdout, "%zu bytes of secret data stored successfully at %p\n",
			size, secret_mem);

	pause();

	secret_fini(secret_mem, size);
	fprintf(stdout, "secret data successfully discarded\n");

	return 0;
}
```

**dumpee**
```
./memfd_app 
ppid: 3883
pid: 3959
5 bytes of secret data stored successfully at 0x7fadc0ed2000
```

**criu dump:**
`sudo ./criu/criu dump --shell-job -D dumpdir/ -v4 -t 3959`

**to reproduce this:**
Please apply the following `diff` otherwise the dump will fail at: `vma_get_mapfile`
```
diff --git a/criu/include/image.h b/criu/include/image.h
index 9a275565f..fe1ab9761 100644
--- a/criu/include/image.h
+++ b/criu/include/image.h
@@ -84,6 +84,7 @@
 #define VMA_AREA_VVAR	 (1 << 12)
 #define VMA_AREA_AIORING (1 << 13)
 #define VMA_AREA_MEMFD	 (1 << 14)
+#define VMA_AREA_MEMFD_SECRET (1 << 15)
 
 #define VMA_EXT_PLUGIN	  (1 << 27)
 #define VMA_CLOSE	  (1 << 28)
index 61c1eee24..ad8c8f3e1 100644
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -79,6 +79,12 @@ static char *buf = __buf.buf;
 
 #define AIO_FNAME "/[aio]"
 
+/*
+ * memfd_secret fd
+ */
+
+#define SECRETMEM_NAME "/secretmem (deleted)"
+
 /* check the @line starts with "%lx-%lx" format */
 static bool __is_vma_range_fmt(char *line)
 {
@@ -440,13 +446,17 @@ static int vma_get_mapfile(const char *fname, struct vma_area *vma, DIR *mfd, st
 
 	*vm_file_fd = openat(dirfd(mfd), path, flags);
 	if (*vm_file_fd < 0) {
-		if (errno == ENOENT)
+		if (errno == ENOENT) {
 			/* Just mapping w/o map_files link */
+			pr_debug("memfd_secret: ENOENT\n");
 			return 0;
+		}
 
 		if (errno == ENXIO) {
 			struct stat buf;
 
+			pr_debug("memfd_secret: ENXIO\n");
+
 			if (fstatat(dirfd(mfd), path, &buf, 0))
 				return -1;
 
@@ -464,6 +474,14 @@ static int vma_get_mapfile(const char *fname, struct vma_area *vma, DIR *mfd, st
 				return 0;
 			}
 
+			/* memfd_secret */
+			if (!strncmp(fname, SECRETMEM_NAME, sizeof(SECRETMEM_NAME))) {
+				pr_debug("memfd_secret: fd identified\n");
+				/* VmaEntry *e; include/vma.h:32 */
+				vma->e->status = VMA_AREA_MEMFD_SECRET;
+				return 0;
+			}
+
 			pr_err("Unknown shit %o (%s)\n", buf.st_mode, fname);
 			return -1;
 		}
```

warusadura:
Fixes: https://github.com/checkpoint-restore/criu/issues/2188


images/vma.proto
```diff
@@ -24,4 +24,6 @@ message vma_entry {
 
 	/* file status flags */
 	optional uint32		fdflags	= 10 [(criu).hex = true];
+
+        required uint64         secretmem_id = 11;
```
rst0git, line 28: https://protobuf.dev/programming-guides/proto2/#specifying-rules
```Suggestion
        optional uint64         secretmem_id = 11;
```

warusadura: thanks @rst0git :) done!

criu/Makefile.crtools
```diff
@@ -40,6 +40,8 @@ obj-y			+= log.o
 obj-y			+= lsm.o
 obj-y			+= mem.o
 obj-y			+= memfd.o
+obj-y                   += secretmem.o
+obj-y                   += memfd-secret.o
```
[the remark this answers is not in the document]
warusadura: Thanks @Snorch :) just fixed the lint.

Re: other CI tests: most of the CI tests are failing because, they don't have access to `memfd_secret` sys call. To enable `memfd_secret` sys call, we have to set kernel boot parameter `secretmem.enable=1`. How can I do this in CI?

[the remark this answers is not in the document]
Snorch: That is a good question:

First, you should detect syscall availability in kerndat and skip zdtm tests for it and skip criu using it in case it is not available.

Second, we need to have test in CI with secretmem.enable=1, but I'm not sure how to do it, maybe someone more experienced with CI knows if we can control kernel boot options? @checkpoint-restore/maintainers

criu/kerndat.c
```diff
@@ -517,6 +551,25 @@ static bool kerndat_has_memfd_hugetlb(void)
...
+	int ret;
+
+	ret = memfd_secret(0);
+
+	if (ret > 0) {
+		kdat.has_memfd_secret = true;
+		close(ret);
+	} else if (ret == -1 && (errno == ENOSYS || errno == EINVAL || errno == EMFILE || errno == ENOMEM))
+		kdat.has_memfd_secret = false;
+	else {
```
Snorch, line 565: Either ALL branches of conditional should have {} or ALL branches should not have {}, there should not be any mixed thing.

criu/Makefile.crtools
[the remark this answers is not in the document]
warusadura: thanks @Snorch

criu/kerndat.c
warusadura: done, thanks!

criu/Makefile.crtools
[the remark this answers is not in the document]
warusadura: I have implemented the `.desc` [file](https://github.com/checkpoint-restore/criu/blob/4992514cbff0748622ca0abb4fc6b2ecc638b948/test/zdtm/static/memfd-secret00.desc) to skip zdtm tests if `memfd_secret` [feature](https://github.com/checkpoint-restore/criu/blob/4992514cbff0748622ca0abb4fc6b2ecc638b948/criu/cr-check.c#L1635) isn't available.

Also, to fix the failing CI tests due to the unavailability of `memfd_secret` syscall, I tried modifying the `FEATURES_LIST` in [Makefile.config](https://github.com/checkpoint-restore/criu/blob/4992514cbff0748622ca0abb4fc6b2ecc638b948/Makefile.config#L82). And updated the [`scripts/feature-tests.mak`](https://github.com/checkpoint-restore/criu/blob/4992514cbff0748622ca0abb4fc6b2ecc638b948/scripts/feature-tests.mak#L175). Then iteratively (looking at the failing CIs), included conditional compilation macros around the `memfd_secret` related code. For example see [`kerndat_get_secretmem_dev()`](https://github.com/checkpoint-restore/criu/blob/4992514cbff0748622ca0abb4fc6b2ecc638b948/criu/kerndat.c#L261).

I'm not this is the right approach to do this. But, these changes fix the important CI tests such as `ci/circleci: test-local-gcc` and ` ci/circleci: test-local-clang`. What do you think @Snorch :)

criu/include/magic.h
```diff
@@ -95,6 +96,7 @@
 #define AUTOFS_MAGIC	     0x49353943 /* Sochi */
 #define FILES_MAGIC	     0x56303138 /* Toropets */
 #define MEMFD_INODE_MAGIC    0x48453499 /* Dnipro */
+#define MEMFD_SECRET_INODE_MAGIC 0x12345678 /* TODO */
```
rppt, line 99: Let's make it Simferopol, my home town :)

criu/include/memfd-secret.h
```diff
@@ -0,0 +1,20 @@
...
+extern int is_memfd_secret(dev_t dev);
+extern const struct fdtype_ops memfd_secret_dump_ops;
+extern struct collect_image_info memfd_secret_cinfo;
+
+#ifdef CONFIG_HAS_MEMFD_SECRET
+#include <sys/syscall.h>
+#include <unistd.h>
+static inline int memfd_secret(unsigned int flags)
+{
+	return syscall(SYS_memfd_secret, flags);
```
rppt, line 16: Please use `__NR_memfd_secret`

criu/Makefile.crtools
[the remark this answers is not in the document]
warusadura: understood, thanks @rppt :)

criu/kerndat.c
```diff
@@ -531,6 +571,31 @@ static bool kerndat_has_memfd_hugetlb(void)
...
+{
+#if defined(CONFIG_HAS_MEMFD_SECRET)
+	int ret;
+
+	ret = memfd_secret(0);
+
+	if (ret > 0) {
+		kdat.has_memfd_secret = true;
+		close(ret);
+	} else if (ret == -1 && (errno == ENOSYS || errno == EINVAL || errno == EMFILE || errno == ENOMEM)) {
```
rppt, line 584: `ENOSYS` means that `memfd_secret` system call is not available. All the rest `errno` values mean that there was an actual error.

criu/include/memfd-secret.h
```diff
@@ -0,0 +1,20 @@
...
+#define __CR_MEMFD_SECRET_H__
+
+#include <sys/types.h>
+#include "common/config.h"
+
+extern int is_memfd_secret(dev_t dev);
+extern const struct fdtype_ops memfd_secret_dump_ops;
+extern struct collect_image_info memfd_secret_cinfo;
+
+#ifdef CONFIG_HAS_MEMFD_SECRET
```
rppt, line 11: As I've mentioned earlier, we can detect availability of `memfd_secret` in runtime. If `syscall(__NR_memfd_secret, 0)` returns `-ENOSYS`, the system call is not available. Otherwise secretmem is enabled.

criu/kerndat.c
```diff
@@ -257,6 +258,45 @@ static int kerndat_get_shmemdev(void)
...
+
+	secretmem = mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+	if (secretmem == MAP_FAILED) {
+		close(fd);
+		return -1;
+	}
+
+	if (kerndat_get_dev(&dev, secretmem, PAGE_SIZE)) {
+		munmap(secretmem, PAGE_SIZE);
+		return -1;
```
rppt, line 285: Missing `close(fd)`. Consider using `goto` for error path.

criu/kerndat.c
```diff
@@ -257,6 +258,45 @@ static int kerndat_get_shmemdev(void)
...
+	}
+
+	munmap(secretmem, PAGE_SIZE);
```
rppt, line 288: `close(fd)`

criu/kerndat.c
```diff
@@ -531,6 +571,31 @@ static bool kerndat_has_memfd_hugetlb(void)
...
+
+	if (ret > 0) {
+		kdat.has_memfd_secret = true;
+		close(ret);
+	} else if (ret == -1 && (errno == ENOSYS || errno == EINVAL || errno == EMFILE || errno == ENOMEM)) {
+		kdat.has_memfd_secret = false;
+	} else {
+		pr_perror("Unexpected error from memfd_secret(0)");
+		return -1;
+	}
```
rppt, line 589: I think it's better to structure the conditions here so that we return on errors and continue when everything is fine. And we can call `kerndat_get_secretmem_dev` from this function or probably even merge these two.
E.g something along these lines:

```
kdat.has_memfd_secret = false;
fd = memfd_secret(0);
if (fd < 0) {
	if (errno == ENOSYS)
		return 0;
	return fd;
}

ret = kerndat_get_secretmem_dev(fd);
if (ret) {
	close(fd);
	return ret;
}
...
```

criu/memfd-secret.c
```diff
@@ -0,0 +1,282 @@
...
+	}
+
+	inode->fdstore_id = fdstore_add(fd);
+	if (inode->fdstore_id < 0)
+		goto out;
+
+	ret = fd;
+	fd = -1;
+
+out:
```
rppt, line 197: Consider using kernel style error handling: https://docs.kernel.org/process/coding-style.html#centralized-exiting-of-functions

criu/proc_parse.c
```diff
@@ -651,7 +662,8 @@ static int handle_vma(pid_t pid, struct vma_area *vma_area, const char *file_pat
 
 				close_safe(vm_file_fd);
 				return 0;
-			}
+			} else /* for memfd_secret case */
```
rppt, line 665: Are we certain that previous cases cover all possibilities? And even if they do, for future compatibility I think it should be
`else if (is_memfd_secret())` here

criu/secretmem.c
rppt, line 1: I think this file can be merged into `criu/memfd_secret.c`

test/zdtm/static/memfd-secret00.c
```diff
@@ -0,0 +1,80 @@
...
+
+#define SECRET "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+#define SIZE   26
+
+const char *test_doc = "memfd_secret file descriptor";
+const char *test_author = "Dhanuka Warusadura <csx@tuta.io>";
+
+static int _memfd_secret(unsigned int flags)
+{
+#if defined(CONFIG_HAS_MEMFD_SECRET)
```
rppt, line 17: Please use dynamic detection. If `syscall(__NR_memfd_secret, 0)` returns `-ENOSYS` just call `skip()` for the test

Snorch: I agree that we don't need CONFIG_HAS_MEMFD_SECRET. But we also don't need a dynamic detection, we have desc file which will get this information from kerndat.

Snorch:
Patch naming is not confirming to our usual stile of patches to zdtm:
`memfd_secret: add test for memfd_secret` should be `zdtm: add test for memfd_secret`

criu/kerndat.c
warusadura: understood, thanks!

criu/include/memfd-secret.h
warusadura: done, thanks!

test/zdtm/static/memfd-secret00.c
warusadura: Implemented the `desc` file for test.

criu/kerndat.c
warusadura: done, thanks!

criu/kerndat.c
warusadura: done, thanks!

criu/proc_parse.c
warusadura: I have thought of this before and I agree with you Mike. But, the issue is, call to `is_memfd_secret(st_buf->st_dev)` here always returns false. I'm not sure why.

Furthermore, the dump works even without this else block. But, the problem with this approach is `vma_area->e->status` is to `VMA_FILE_PRIVATE` not `VMA_AREA_MEMFD_SECRET` which is what we want.

Moreover, we are already in a parent `else` [block](https://github.com/checkpoint-restore/criu/blob/8ee13b6536cb441c4c515b32dec357a6fca9affb/criu/proc_parse.c#L642) without any condition, and currently at the end of it. So, it kind of make sense. Plus, there is already another if [block](https://github.com/checkpoint-restore/criu/blob/8ee13b6536cb441c4c515b32dec357a6fca9affb/criu/proc_parse.c#L668) to handle other cases if needed.

criu/kerndat.c
warusadura: @rppt I have updated the [`kerndat_get_secretmem_dev()`](https://github.com/checkpoint-restore/criu/blob/2e8fba61b9683904a1faf45ac53fef38dfdb648b/criu/kerndat.c#L261) and [`kerndat_has_memfd_secret()`](https://github.com/checkpoint-restore/criu/blob/2e8fba61b9683904a1faf45ac53fef38dfdb648b/criu/kerndat.c#L576) according to the previously requested changes (now resolved) . And I'm not sure why we need this change? Is it because, we are calling `memfd_secret()` syscall twice in both functions?

warusadura: done.

criu/include/magic.h
warusadura: done.

criu/proc_parse.c
warusadura: done!

images/memfd-secret.proto
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+
+syntax = "proto2";
+
+import "fown.proto";
+
+message memfd_secret_file_entry {
+        required uint32		id		= 1;
+	required uint32		flags		= 2;
```
rppt, line 9: nit: 
```
required uint32         flags           = 2 [(criu).flags = "rfile.flags"];
```

test/zdtm/static/memfd-secret00.desc
```diff
@@ -0,0 +1 @@
+{'feature': 'memfd_secret', 'flags': 'noauto'}
```
rppt, line 1: I think this belongs to the commit that adds the test.

rppt (commented):
Overall LGTM, there a few nits here and there

test/zdtm/static/memfd-secret00.c
```diff
@@ -0,0 +1,80 @@
...
+const char *test_doc = "memfd_secret file descriptor";
+const char *test_author = "Dhanuka Warusadura <csx@tuta.io>";
+
+static int _memfd_secret(unsigned int flags)
+{
+#ifdef __NR_memfd_secret
+	return syscall(__NR_memfd_secret, flags);
+#else
+	return -1;
+#endif
```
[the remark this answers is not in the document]
warusadura: Understood. But, how can I gracefully skip the test? There is no [API](https://criu.org/ZDTM_API) to skip the test.  I think the feature check (.desc file) approach is the only way to skip the test.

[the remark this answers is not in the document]
rppt: You can take a look at test/zdtm/static/arm-neon00.c and test/zdtm/static/fpu02.c to see how they skip tests based on an `#ifdef`. 

For the run-time, you can check if `syscall(__NR_memfd_secret)` returned `-ENOSYS` and then just print message with `skip()` and exit with success.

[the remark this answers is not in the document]
Snorch: Can we just define __NR_memfd_secret? (E.g. like we define __NR_rseq. For all arches except alpha it should be 447, and we don't support arch/alpha as far as I remember.)

criu/kerndat.c
```diff
@@ -259,6 +260,33 @@ static int kerndat_get_shmemdev(void)
...
+static int kerndat_get_secretmem_dev(int fd)
+{
+	void *secretmem = NULL;
+	dev_t dev;
+
+	if (ftruncate(fd, PAGE_SIZE) < 0)
+		goto err;
+
+	secretmem = mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+	if (secretmem == MAP_FAILED)
```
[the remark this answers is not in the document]
warusadura: done, thanks!

criu/memfd-secret.c
```diff
@@ -0,0 +1,462 @@
...
+static int memfd_secret_open_inode(struct memfd_secret_restore_inode *inode)
+{
+	int fd;
+
+	if (inode->fdstore_id != -1)
+		return fdstore_get(inode->fdstore_id);
+
+	mutex_lock(&inode->lock);
+	if (inode->fdstore_id != -1)
+		fd = fdstore_get(inode->fdstore_id);
```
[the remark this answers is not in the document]
warusadura: Yeah, my bad. Just removed the call to `fdstore_get()` before the lock.

images/memfd-secret.proto
warusadura: done, thanks!

test/zdtm/static/memfd-secret00.desc
warusadura: I agree, thanks!

test/zdtm/static/memfd-secret00.c
```diff
@@ -0,0 +1,80 @@
...
+	test_init(argc, argv);
+
+	secretmem = secret_init(SIZE);
+	if (!secretmem) {
+		fail("memfd_secret: not supported operation");
+		return 1;
+	}
+
+	for (int i = 0; i < SIZE; i++)
+		secretmem[i] = 'A' + i;
```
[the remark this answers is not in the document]
warusadura: done, thanks!

test/zdtm/static/memfd-secret00.c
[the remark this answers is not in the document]
warusadura: @rppt included the conditional compilation and call to `skip()`.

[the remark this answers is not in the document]
warusadura: @Snorch thanks. But, I'm not sure it will work. Because, as I understand, even if we define the ` __NR_memfd_secret`, the `secretmem` feature needs to be enable to use the syscall. I think, without explicitly enabling  `secretmem`, exposing/defining `__NR_memfd_secret` has no use.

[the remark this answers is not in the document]
Snorch: @warusadura Let me explain:
1) You need a kerndat check which will test if `secretmem` feature is supported by current kernel. And based on the result of this check you can skip the test as we do in multiple other tests.
2) You need to define __NR_memfd_secret as local headers may not provide it even if the current kernel supports it. I imagine that your ifdef is trying to do similar thing, just defining is much more clear solution.
Those are two separate things.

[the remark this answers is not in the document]
warusadura: Understood, thanks you @Snorch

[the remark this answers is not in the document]
warusadura: implemented,
```
#ifndef __NR_memfd_secret
#define __NR_memfd_secret 447
#endif
```

criu/kerndat.c
```diff
@@ -259,6 +260,33 @@ static int kerndat_get_shmemdev(void)
...
+		return -1;
+
+	if (kerndat_get_dev(&dev, secretmem, PAGE_SIZE))
+		goto err;
+
+	munmap(secretmem, PAGE_SIZE);
+	kdat.secretmem_dev = dev;
+	pr_info("Found secret-memory device at %ld\n", kdat.secretmem_dev);
```
Snorch, line 280: See get_tty_driver and how PRIx64 is used to print dev_t, you should use it here to fix "Cross Compile Tests".

warusadura: done, thanks!

PR #2250, opened 2023-08-24
Lint add warnings instead of check failure

Snorch:
There are multiple cases where good human readable code block is
converted to an unreadable mess by clang-format, so we don't want to
rely on clang-format completely. Also there is no way, as far as I can
see, to make clang-format only fix what we want it to fix without
breaking something.

So let's just display hints inline where clang-format is unhappy. When
reviewer sees such a warning it's a good sign that something is broken
in codding style around this warning.

### DIFF
--- a/scripts/github-indent-warnings.py
+++ b/scripts/github-indent-warnings.py
@@ -1,0 +1,33 @@
+#!/usr/bin/python3
+import sys
+import re
+
+re_file = r'^diff --git a/(\S\S*)\s.*$'
+re_line = r'^@@ -(\d\d*)\D.*@@.*$'
+
+if __name__ == '__main__':
+    if len(sys.argv) != 1 and len(sys.argv) != 2:
+        print(f'usage: {sys.argv[0]} <path/to/file>')
+        print(f'usage: <command> | {sys.argv[0]}')
+        exit(1)
+
+    input_file = sys.stdin.fileno()
+    if len(sys.argv) == 2:
+        input_file = sys.argv[1]
+
+    with open(input_file, 'r') as fi:
+        file_name = None
+        line_number = None
+        for line in fi:
+            file_matches = re.findall(re_file, line)
+            if len(file_matches) == 1:
+                file_name = file_matches[0]
+                continue
+
+            if file_name is None:
+                continue
+
+            line_matches = re.findall(re_line, line)
+            if len(line_matches) == 1:
+                line_number = int(line_matches[0]) + 3
+                print(f'::warning file={file_name},line={line_number}::clang-format: Possible coding style problem (https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#automatic-tools-to-fix-coding-style)')
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -26,15 +26,15 @@ jobs:
       run: make lint
 
     - name: Run make indent
-      run: >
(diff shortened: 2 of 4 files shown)


Snorch:
It looks like that:

image

Please remove https://github.com/checkpoint-restore/criu/pull/2250/commits/584515db3903ded398da012f781b4b1e93fcd344 when merging this PR, this commit is needed just to illustrate the concept.

criu/mount.c
```diff
@@ -95,12 +95,12 @@ int ext_mount_parse_auto(char *key)
 static char *ext_mount_lookup(char *key)
 {
 	char *v;
-	int len = strlen(key);
+int len = strlen(key);
```
avagin, line 98: > clang-format: Possible codding-style problem

can it give more info about a problem?

Snorch: I feel that hiding clang-generated auto-formats is a good thing, else we would end up blindly applying them again and code readability would suffer 😀 It should be enough to point the line in file to make manual review a bit easier.

Moreover, multiline warnings seem to be not working on github https://github.com/actions/toolkit/issues/193#issuecomment-1691046058 . And also there is no way to get info from clang about type of error.

scripts/github-indent-warnings.py
```diff
@@ -0,0 +1,28 @@
...
+                file_name = file_matches[0]
+                continue
+
+            if file_name is None:
+                continue
+
+            line_matches = re.findall(re_line, line)
+            if len(line_matches) == 1:
+                line_number = int(line_matches[0]) + 3
+                print(f'::warning file={file_name},line={line_number}::clang-format: Possible codding-style problem')
```
rst0git, line 28:
```Suggestion
                print(f'::warning file={file_name},line={line_number}::clang-format: Possible coding-style problem')
```

scripts/github-indent-warnings.py
```diff
@@ -0,0 +1,28 @@
...
+re_file = r'^diff --git a/(\S\S*)\s.*$'
+re_line = r'^@@ -(\d\d*)\D.*@@.*$'
+
+if __name__ == '__main__':
+    if len(sys.argv) != 1 and len(sys.argv) != 2:
+        print(f'usage: {sys.argv[0]} <path/to/file>')
+        print(f'usage: <command> | {sys.argv[0]}')
+        exit(1)
+
+    with open(sys.stdin.fileno() if len(sys.argv) == 1 else sys.argv[1], 'r') as fi:
```
rst0git, line 13: Would it be possible to split this into multiple lines? For example, something like the following:
```Python
input_file = sys.stdin.fileno()
if len(sys.argv) == 1:
    input_file = sys.argv[1]

with open(input_file, 'r') as fi:
   ...
```
([*simple is better than complex*](https://peps.python.org/pep-0020/))

scripts/github-indent-warnings.py
Snorch: Thanks, done!

scripts/github-indent-warnings.py
Snorch: Sure, you are right, done.

Snorch:
I've added changes to CONTRIBUTING.md with bad clang-formating examples to PR and a link from warnings to CONTRIBUTING.md, so that it would be easier to understand how this warning should be handled.

Snorch:
I've removed test commit to trigger warnings (it works), so it's now ready to merge.

PR #2251, opened 2023-08-24
ztdm: inhfd: handle child hanging

osctobe:
Let's kill the child when the test is hanging.  Due to PEP 475 the SIGALRM handler needs to throw an exception to be able to interrupt wait().  To improve debuggig, close the fd in the child after reading it and detect that in the parent to show whether the child hung part way.


osctobe:
So it seems some tests didn't expect timeouts to work...

PR #2252, opened 2023-08-24
kerndat: Skip clone3(set_tid) when unprivileged.

osctobe:
A set of fixes for kerndat tests and a few debug logging improvements.


criu/kerndat.c
```diff
@@ -1384,12 +1384,15 @@ static bool kerndat_has_clone3_set_tid(void)
 	pid_t pid;
 	struct _clone_args args = {};
 
+	kdat.has_clone3_set_tid = false;
+	if (opts.unprivileged && !has_cap_checkpoint_restore(opts.cap_eff))
```
avagin, line 1402: > Pretend it's not available instead of failing restore later when running unprivileged.

Can restore succeed if  kdat.has_clone3_set_tid isn't set and criu is running w/o cap_sys_admin and cap_cr?

criu/kerndat.c
```diff
@@ -72,11 +73,24 @@ static int check_pagemap(void)
...
 	}
 
-	/* Get the PFN of some present page. Stack is here, so try it :) */
-	ret = pread(fd, &pfn, sizeof(pfn), (((unsigned long)&ret) / page_size()) * sizeof(pfn));
-	if (ret != sizeof(pfn)) {
-		pr_perror("Can't read pagemap");
-		return -1;
+	retry = 3;
+	while (retry--) {
+		++dummy_var;
```
avagin, line 78: dummy_var++

criu/proc_parse.c
```diff
@@ -842,6 +842,7 @@ int parse_smaps(pid_t pid, struct vm_area_list *vma_area_list, dump_filemap_t du
 			goto err;
 		}
 
+		pr_info("Handling VMA with the following smaps entry: %s\n", str);
```
avagin, line 844: I don't think that pr_info is a good choice in this case.  smaps can contain a lot of entries, but printing log messages isn't so fast, so it can effect performance.

criu/kerndat.c
```diff
@@ -72,11 +73,24 @@ static int check_pagemap(void)
...
+		/* Get the PFN of a page likely to be present. */
+		ret = pread(fd, &pfn, sizeof(pfn), PAGE_PFN((uintptr_t)&dummy_var) * sizeof(pfn));
+		if (ret != sizeof(pfn)) {
+			pr_perror("Can't read pagemap");
+			return -1;
+		}
+		/* The page can be swapped out by the time the read occurs,
+		 * in which case the rest of the bits are a swap offset,
+		 * and can't be used to determine whether PFNs are visible.
```
avagin, line 87: If the kernel doesn't show pfn-s, it doesn't show swap offsets too, doesn't it?

criu/kerndat.c
osctobe: Yes. It does it using the old method: by using `ns_last_pid` (via a RPC helper).

criu/kerndat.c
osctobe: Added another patch to reword the comment a bit.

criu/kerndat.c
avagin: Let's write a comment to explain this in the code.

avagin: I think this isn't a right solution for the problem:
* process can be restored in a separate user and pid namespaces and clone3() works fine in this case.
* this function checks whether clone3 is supported or not, and it is separate from the dump/restore code. I think we have to move these checks to the code that restores processes with specified pids.

avagin: yes, you are right.

Snorch:
Hello @avagin and @osctobe, I had one idea, which I like a lot, but guys in mainstream Linux faced it with total silence.

The idea was to allow clone3 syscall to alter owner user namespace of newly created namespaces (e.g. new pid namespace owner if CLONE_NEWPID is specified). https://lore.kernel.org/all/20210402155131.119872-1-ptikhomirov@virtuozzo.com/

This way using clone3 CRIU is able to create all restored processes in topmost user namespace available, while preserving namespace ownership topology. So at each clone3 call we would have all permissions needed by clone3_set_tid functionality. (Later we can switch to proper user namespace for each process to also preserve task's user namespaces.)

I believe my fix would help in this case too. @osctobe Can you, please, give it a try on your environment?

PR #2255, opened 2023-08-30
clang-format: disable column limit constraint

Snorch:
The "ColumnLimit: 120" is not only allowing lines to be longer than 80 characters but it also forces line wrapping at 120 characters. If total expression length is more than 120 characters, clang-format will try to wrap it as close to 120 as it can, it would not even allow to wrap at 80 characters if we really want it. But as we all know 80 characters is Linux kernel coding style default and as far as our coding style is based on it it is really strange to prohibit wrapping lines at 80 characters...


Snorch:
@checkpoint-restore/maintainers I'm not sure about it, but as there seems to be no way to say "<120 is also ok" to clang-format, let's just disable it.

PR #2259, opened 2023-09-05
criu: Add pidfd support

h0lyalg0rithm:


warusadura:
@h0lyalg0rithm Is it ok, if I write the zdtm test for this?

h0lyalg0rithm:
@warusadura There is one bug that is to be fixed.Currently the patch lets you dump pidfd for processes which are not part of the current process tree.

I have the test prepared where I fork the test and have pidfd on the child process.
But once you try to restore the process it fails as the child process is not created yet.
One way to get around this is to wait until all the whole process tree is restored and then restore the pidfd.

I cannot think about any thing which doesnt sound hacky. What do you think?

warusadura:
> I cannot think about any thing which doesnt sound hacky. What do you think?

sorry @h0lyalg0rithm I'm not sure :)

criu/pidfd.c
```diff
@@ -0,0 +1,81 @@
...
+};
+
+static int open_pidfd_fd(struct file_desc *d, int *new_fd)
+{
+	int fd = -1;
+	struct pidfd_info *info = container_of(d, struct pidfd_info, d);
+	PidfdEntry *pidfde = info->pidfde;
+
+	pr_info("Creating new pidfd %" PRId64 "\n", pidfde->pid);
+	fd = pidfd_open(pidfde->pid, 0);
```
avagin, line 53: a task linked to fd can be dead and pidfd_open will fail in this case.

test/zdtm/static/pidfd00.c
```diff
@@ -0,0 +1,64 @@
...
+int main(int argc, char **argv)
+{
+	int pidfd, errcode = 42;
+	int status;
+	int ret;
+	struct pollfd pollfd;
+
+	test_init(argc, argv);
+
+	pidfd = pidfd_open(1, 0);
```
avagin, line 33: The test has to create more than one pidfd linked to different tasks and then checks that they are linked to proper tasks after C/R.

test/zdtm/static/pidfd00.c
```diff
@@ -0,0 +1,64 @@
...
+	pollfd.fd = pidfd;
+	pollfd.events = POLLIN;
+
+	ret = poll(&pollfd, 1, -1);
+	if (ret == -1) {
+		pr_perror("Poll error");
+		fail();
+	}
+
+	if (pollfd.revents & POLLIN) {
```
avagin, line 51: why can it be triggered?

criu/image.c
```diff
@@ -31,12 +31,14 @@ int check_img_inventory(bool restore)
...
 	InventoryEntry *he;
+	int loaded;
 
 	img = open_image(CR_FD_INVENTORY, O_RSTR);
 	if (!img)
 		return -1;
 
-	if (pb_read_one(img, &he, PB_INVENTORY) < 0)
+	loaded = pb_read_one(img, &he, PB_INVENTORY);
+	if (loaded < 0)
```
rst0git, line 41: This change looks unnecessary.

criu/include/magic.h
```diff
@@ -100,6 +100,7 @@
 #define BPFMAP_FILE_MAGIC    0x57506142 /* Alapayevsk */
 #define BPFMAP_DATA_MAGIC    0x64324033 /* Arkhangelsk */
 #define APPARMOR_MAGIC	     0x59423047 /* Nikolskoye */
+#define PIDFD_MAGIC	     0x59423447
```
rst0git, line 103: we tend to use location coordinates of a city as a magic number. For example, Nikolskoye is a city with the following coordinates (from Wikipedia) 59°42'N 30°47'E ->  0x59423047.

test/zdtm/static/pidfd00.c
h0lyalg0rithm: @avagin I tried to write a test that would create a fork of a process and the parent process would monitor the pid of the child process. The dumping and restore process would first restore the parent process and the child process would still not be restored, hence the restore would fail.
I was thinking of restoring the pidfd after all the child processes have been restored, does that make sense.
To do this I would have to dump the pidfd seperate from the regular fd dump process.

rst0git:
closing in favour of https://github.com/checkpoint-restore/criu/pull/2449

PR #2260, opened 2023-09-07
support clang build

lfeng14:


adrianreber:
You need to add a signed-off-by to your commit and please also add some details what and why you are changing. We already do clang based builds in CI.

Please take a look at the CI results the GCC based and the CLANG based builds all seem to be failing now.

PR #2263, opened 2023-09-12
dump: use MEMBARRIER_CMD_GET_REGISTRATIONS when available

mclapinski:
MEMBARRIER_CMD_GET_REGISTRATIONS can tell us whether or not the process used MEMBARRIER_CMD_REGISTER_GLOBAL_EXPEDITED unlike the old probing method.

Falls back to the old method when MEMBARRIER_CMD_GET_REGISTRATIONS is unavailable.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -17,6 +17,7 @@
 #include <sys/inotify.h>
 #include <sched.h>
 #include <sys/mount.h>
+#include <linux/membarrier.h>
 
 #if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
 #include <nftables/libnftables.h>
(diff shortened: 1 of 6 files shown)


criu/kerndat.c
```diff
@@ -1636,6 +1637,23 @@ static int kerndat_has_ipv6_freebind(void)
 	return ret;
 }
 
+static int kerndat_has_membarrier_get_registrations(void)
+{
+	const int cmd_get_registrations = 1 << 9;
```
avagin, line 1642: 1 << MEMBARRIER_CMDBIT_GET_REGISTRATIONS

avagin:
can we add a test to check that MEMBARRIER_CMD_REGISTER_GLOBAL_EXPEDITED is C/R-ed?

mclapinski:
I can't check it the same way we check other membarrier commands because invoking MEMBARRIER_CMD_GLOBAL_EXPEDITED without MEMBARRIER_CMD_REGISTER_GLOBAL_EXPEDITED is valid.

I could try to check that the feature itself works (the membarrier is executed) but I don't know of a good way to write such a test.

I could check it using MEMBARRIER_CMD_GET_REGISTRATIONS but it would be duplicating the criu code so I'm not sure if it's valuable. LMK.

avagin:
> I could check it using MEMBARRIER_CMD_GET_REGISTRATIONS but it would be duplicating the criu code so I'm not sure if it's valuable. LMK.

@mclapinski I think it is valuable, we need to check that the criu part works as expected.

mclapinski:
@avagin Done. I think the CI failures are unrelated to my PR.

avagin:
Merged. Thanks.

PR #2267, opened 2023-09-23
memfd: don't reopen file descriptors to restore memory mappings

avagin:


avagin:
This change fixes false-positive warnings that are reported by many tests, because memfd-s are used to restore shared memory regions:
```
Run criu restore
=[log]=> dump/zdtm/static/sched_prio00/52/1/restore.log
b'(00.003271)     53: No parent images directory provided'
b'(00.003277)     53: Opened local page read 3 (parent 0)'
b'(00.003278)     53: pr44632-3 Read 0 1 pages'
b'(00.003280)     53: \tpr44632-3 Read page from self 0/0'
b'(00.003316)     52: Warn  (criu/memfd.c:371): execveat(fd=5, ..., AT_EMPTY_PATH) might fail after restore; memfd id=3'
b'(00.003318)     54: Warn  (criu/memfd.c:371): execveat(fd=5, ..., AT_EMPTY_PATH) might fail after restore; memfd id=3'
b'(00.003325)     53: Restore via sigreturn'
b'(00.003326)     55: Warn  (criu/memfd.c:371): execveat(fd=5, ..., AT_EMPTY_PATH) might fail after restore; memfd id=3'
Send the 15 signal to  52
Wait for zdtm/static/sched_prio00(52) to die for 0.100000
Removing dump/zdtm/static/sched_prio00/52
```

criu/memfd.c
```diff
@@ -342,6 +342,9 @@ int memfd_open(struct file_desc *d, u32 *fdflags)
 	/* Reopen the fd with original permissions */
 	flags = fdflags ? *fdflags : mfe->flags;
 
+	if (filemap && (flags & O_ACCMODE) == O_RDWR)
```
osctobe, line 345: Isn't `filemap == !fdflags` always true?

test/zdtm/static/memfd04.c
```diff
@@ -73,20 +74,27 @@ static const size_t script_len = sizeof(script) - 1;
...
-	if (write(fd, script, script_len) != script_len) {
-		fail("write(memfd)");
+	addr = mmap(NULL, PAGE_SIZE, PROT_READ | PROT_WRITE, MAP_FILE | MAP_SHARED, fd, 0);
+	if (addr == MAP_FAILED) {
+		pr_perror("mmap");
 		return 1;
 	}
+	memcpy(addr, script, script_len);
 
 	if (!test_exec_fd(fd))
```
osctobe, line 99: I'm surprised this actually works to execute a fd that is RW-mmapped in another process. Can we make this a separate test in case this changes?

criu/memfd.c
avagin: No, it isn't:
https://github.com/checkpoint-restore/criu/blob/5e37ccf379ad7f47a720f64bced40cf2fe5bd6a2/criu/files-reg.c#L2512

test/zdtm/static/memfd04.c
avagin: Yes, we can. I am not sure that it is really worth the separate test...

osctobe: It is testing a different conditions, so I'd prefer to have a separate test for better bug signal. We could extract some common code if that's a concern.

avagin: Strictly speaking, it tests the same major condition is that an origin memfd is preserved when it is possible.

avagin: I have added memfd05...

PR #2268, opened 2023-09-24
ci: stop testing ubuntu overlayfs

avagin:
They break it with each kernel rebase. More details are here: https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257

Last time, it was fixed a few month ago and it has been broken again in 5.15.0-1046-azure.


adrianreber:
So the fix is basically to bind mount the criu directory from the outside in the test container and being independent of the file system used by the container. Took me a while to spot that change. Would have been good to mention the change in the commit message.

Besides that both changed tests are happy. LGTM.

PR #2270, opened 2023-09-27
uffd: Fix page fault address

yota9:
The page_size() returns unsigned int value that is after "bitwise not"
is promoted to unsigned long (msg->arg.pagefault.address) value. Sinc
e the value is unsigned promotion is done with 0 MSB that results in
lost of MSB pagefault address bits. Cast page_size to unsigned long
first to avoid such situation.

### DIFF
--- a/compel/plugins/std/infect.c
+++ b/compel/plugins/std/infect.c
@@ -27,7 +27,7 @@ static struct rt_sigframe *sigframe;
  */
 static unsigned __page_size;
 
-unsigned __attribute((weak)) page_size(void)
+unsigned long __attribute((weak)) page_size(void)
 {
 	return __page_size;
 }
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -104,7 +104,7 @@ bool fault_injected(enum faults f)
  * Hint: compel on aarch64 shall learn relocs for that.
  */
 static unsigned __page_size;
-unsigned page_size(void)
+unsigned long page_size(void)
 {
 	return __page_size;
 }
--- a/include/common/arch/aarch64/asm/page.h
+++ b/include/common/arch/aarch64/asm/page.h
@@ -10,7 +10,7 @@
 extern unsigned __page_size;
 extern unsigned __page_shift;
 
-static inline unsigned page_size(void)
+static inline unsigned long page_size(void)
 {
 	if (!__page_size)
 		__page_size = sysconf(_SC_PAGESIZE);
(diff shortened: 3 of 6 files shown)


adrianreber:
@rppt PTAL

minhbq-99:
It's a nice catch. page_size in defined in criu/include/common/arch/*/asm/page.h. Interestingly, on x86, s390, arm, it returns unsigned long which does not cause the above issue. However, on aarch64, mips, ppc64, it returns unsigned int. I wonder if changing the page_size to return unsigned long in all archs is a good solution.

rppt:
> It's a nice catch. page_size in defined in criu/include/common/arch/*/asm/page.h. Interestingly, on x86, s390, arm, it returns unsigned long which does not cause the above issue. However, on aarch64, mips, ppc64, it returns unsigned int. I wonder if changing the page_size to return unsigned long in all archs is a good solution.

I think it is. Adding a cast in uffd.c is more of a band aid and making page_size() return unsigned long is the proper solution IMO.

yota9:
> It's a nice catch. page_size in defined in criu/include/common/arch/*/asm/page.h. Interestingly, on x86, s390, arm, it returns unsigned long which does not cause the above issue. However, on aarch64, mips, ppc64, it returns unsigned int. I wonder if changing the page_size to return unsigned long in all archs is a good solution.

Thanks for further investigations, I've made page_size() to return unsigned long in all the arches and couple of other places.

rppt:
> > It's a nice catch. page_size in defined in criu/include/common/arch/*/asm/page.h. Interestingly, on x86, s390, arm, it returns unsigned long which does not cause the above issue. However, on aarch64, mips, ppc64, it returns unsigned int. I wonder if changing the page_size to return unsigned long in all archs is a good solution.
> 
> Thanks for further investigations, I've made page_size() to return unsigned long in all the arches and couple of other places.

It looks like you've missed loongarch64 :)

yota9:
True, done. Thanks!

> > > It's a nice catch. page_size in defined in criu/include/common/arch/*/asm/page.h. Interestingly, on x86, s390, arm, it returns unsigned long which does not cause the above issue. However, on aarch64, mips, ppc64, it returns unsigned int. I wonder if changing the page_size to return unsigned long in all archs is a good solution.
> > 
> > 
> > Thanks for further investigations, I've made page_size() to return unsigned long in all the arches and couple of other places.
> 
> It looks like you've missed loongarch64 :)

rppt (approved):

rst0git (approved):

yota9:
I'm not authorized to merge this pull request, so please do it for me, thanks!

PR #2272, opened 2023-09-28
vma: Add !VVAR condition to vma_entry_can_be_lazy

yota9:
Currently most of the times we don't have problems with VVAR segment and
lazy restore because when VDSO is parked there is an munmap call that
calls UFFDIO_UNREGISTER on the destination address.
But we don't want to enable userfaultfd for VDSO and VVAR at the first
place.

### DIFF
--- a/criu/include/vma.h
+++ b/criu/include/vma.h
@@ -122,8 +122,8 @@ static inline struct vma_area *vma_next(struct vma_area *vma)
 static inline bool vma_entry_can_be_lazy(VmaEntry *e)
 {
 	return ((e->flags & MAP_ANONYMOUS) && (e->flags & MAP_PRIVATE) && !(e->flags & MAP_LOCKED) &&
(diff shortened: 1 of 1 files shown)


avagin:
lgtm

rppt:
>  AFAIU Currently most of the times we don't have problems with VVAR segment due
> to during parking vdso on it's old place mremap is used and during it
> munmap is called to destination address which results in calling
> UFFDIO_UNREGISTER. But as for VDSO we don't want to enable uffd for these addresses at firse place.

I find the changelog a bit hard to parse, how about

Currently most of the times we don't have problems with VVAR segment and lazy restore because when VDSO is parked there is an munmap call that calls UFFDIO_UNREGISTER on the destination address.
But we don't want to enable userfaultfd for VDSO and VVAR at the first place.

Other than that, LGTM

yota9:
@rppt Thanks, done. Please merge it since I don't have rights.

rppt (approved):

Issue #2269, opened 2023-09-26
Criu Restore failing with sudo


rerrabolu:
master@hpe-sjc2-03:Tmp$ criu --version
Version: 3.18
GitID: v3.18-143-g4c845c885

Restore Cmd:  sudo criu restore -vvv -o restore.log --shell-job --link-remap && echo "Restore OK"

The restore is applied to a two process tree involving a GPU (amdgpu plugin). Each process generates 3 IDs for the various VMA associated with each process. The restore goes through the restoration of each ID file. Criu framework encounters an error as given below before invoking the two processes to "resume'

(01.355179)  11592: Further fle=0x7f48c9bdc748, pid=11592
(01.355185)  11592: Error (criu/util.c:246): fd 0 already in use (called at criu/files.c:992)
(01.355189)  11592: Error (criu/files.c:1177): Can't get fd=0, pid=11592
(01.355193)  11592: Error (criu/files.c:1215): Unable to open fd=0 id=0x11
(01.375880)  11591: Error (criu/cr-restore.c:1513): 11592 exited, status=1
(01.375984) Error (criu/cr-restore.c:2557): Restoring FAILED.

This error does not occur on another two process workload that generates only one ID file per process. The ID file refers to KFD device. In the failing case the two extra files refer to DRM devices.

Any suggestions on how to understand / debug this issue is appreciated

rerrabolu:
The file at the root of the error trigger is said to be:  **tty[8801:17]**

avagin:
Could you attach the full criu log? It should help to figure out the issue.

avagin:
I figured out where the problem is. I think we need to do something like:
https://github.com/avagin/criu/commit/31b4997b628e4a5556332fd12a5b49864e52aab1

Unfortunately, I don't have hardware to test the patch. @fdavid-amd and @dayatsin-amd could you review and finalize the change? It is probably requires some minor fixes.

rerrabolu:
Please find attached logs from the run that end up on "restore" failing

- Kfdtest-ProcMaps
- ram_criu_dump.log
- ram_criu_restore.log

Let me know if the logs have enough information. I will take a look at the patch provided and see if I can try it out.

dayatsin-amd:
Thanks for the patch @avagin 

From what I remember there was something we had to do to restore the tty devices with the --inherit-fd option:
https://criu.org/Inheriting_FDs_on_restore

@rerrabolu is also from AMD, will sync-up with up with him

[...]

PR #2273, opened 2023-09-28
util: allow to run criu under strace

avagin:
fork_and_ptrace_attach has to fork a child with CLONE_UNTRACED, so that strace doesn't trace it.

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -661,38 +661,52 @@ out:
 	return ret;
 }
 
+struct child_args {
+	int *sk_pair;
+	int (*child_setup)(void);
+};
+
+static int child_func(void *_args)
+{
+	struct child_args *args = _args;
+	int sk, *sk_pair = args->sk_pair;
+	char c = 0;
+
+	sk = sk_pair[1];
+	close(sk_pair[0]);
+
+	if (args->child_setup && args->child_setup() != 0)
+		exit(1);
+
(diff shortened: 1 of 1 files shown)


criu/util.c
```diff
@@ -661,40 +661,53 @@ int cr_system_userns(int in, int out, int err, char *cmd, char *const argv[], un
 	return ret;
 }
 
+struct child_args {
+	int *sk_pair;
+	int (*child_setup)(void);
+};
+
+static int child_func(void *_args) {
```
rst0git, line 669:
```diff
-static int child_func(void *_args) {
+static int child_func(void *_args)
+{
```

rst0git:
LGTM, but I was not able to replicate the problem in https://github.com/checkpoint-restore/criu/issues/2269#issuecomment-1739927149.

avagin:
@rst0git it is triggered each time when kerndat_has_ptrace_get_rseq_conf is called. If you can't reproduce it, you need to find out why kerndat_has_ptrace_get_rseq_conf isn't executed in your case.

https://github.com/checkpoint-restore/criu/blob/5bf7652504530b9baa04f411052f9c079c7f14eb/criu/kerndat.c#L978

rst0git:
@avagin Thanks! I was able to replicate it and confirm that it works.

rst0git (approved):

PR #2276, opened 2023-09-30
ci: enable build with amdgpu plugin

rst0git:
This pull request adds the `libdrm-dev` package to the list of CRIU dependencies installed in CI to build CRIU with amdgpu plugin.

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -6,7 +6,7 @@ CI_PKGS=(protobuf-c-compiler libprotobuf-c-dev libaio-dev libgnutls28-dev
 		libnl-3-dev gdb bash libnet-dev util-linux asciidoctor
 		libnl-route-3-dev time flake8 libbsd-dev python3-yaml
 		libperl-dev pkg-config python3-protobuf python3-pip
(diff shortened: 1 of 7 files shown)


scripts/ci/run-ci-tests.sh
```diff
@@ -326,3 +326,8 @@ make -C test/others/action-script run
 
 # compel testing
 make -C compel/test
+
+# amdgpu_plugin testing
+make amdgpu_plugin
+make -C plugins/amdgpu/ test_topology_remap
+./plugins/amdgpu/test_topology_remap
```
avagin, line 333: What does this test do? How does it work w/o amd devices?

rst0git: This is a unit test introduced in commit https://github.com/checkpoint-restore/criu/commit/ba9c62df24160743d8082cfc5f6af4e92a4e629a. There is a comment describing what the test does and how it works in [plugins/amdgpu/tests/test_topology_remap.c](https://github.com/checkpoint-restore/criu/blob/5de9040ee758f1fd1a2599b6f800013544c966b6/plugins/amdgpu/tests/test_topology_remap.c#L2).

PR #2278, opened 2023-10-05
tun: don't parse buffers that have not been filled with data

avagin:
read_ns_sys_file() can return an error, but we are trying to parse a buffer before checking a return code.

CID 417395 (#3 of 3): String not null terminated (STRING_NULL)
2. string_null: Passing unterminated string buf to strtol, which expects a null-terminated string.

### DIFF
--- a/criu/net.c
+++ b/criu/net.c
@@ -111,15 +111,18 @@ int read_ns_sys_file(char *path, char *buf, int len)
 	}
 
 	rlen = read(fd, buf, len);
+	if (rlen == -1)
+		pr_perror("Can't read ns' %s", path);
 	close(fd);
 
 	if (rlen == len) {
+		buf[0] = '\0';
 		pr_err("Too small buffer to read ns sys file %s\n", path);
 		return -1;
 	}
 
-	if (rlen > 0)
-		buf[rlen - 1] = '\0';
+	if (rlen >= 0)
+		buf[rlen] = '\0';
 
 	return rlen;
 }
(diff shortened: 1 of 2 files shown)


criu/net.c
```diff
@@ -111,15 +111,18 @@ int read_ns_sys_file(char *path, char *buf, int len)
 	}
 
 	rlen = read(fd, buf, len);
+	if (rlen == -1)
+		pr_perror("Can't read ns' %s", path);
 	close(fd);
```
kolyshkin, line 116: I would move it to after the `close` and add `return -1`. Being explicit always helps.

criu/net.c
```diff
@@ -111,15 +111,18 @@ int read_ns_sys_file(char *path, char *buf, int len)
...
 
 	if (rlen == len) {
+		buf[len - 1] = 0;
 		pr_err("Too small buffer to read ns sys file %s\n", path);
 		return -1;
```
kolyshkin, line 121: Does it make sense to return a partial buffer? I mean, maybe it's better to do `buf[0] = 0`.

criu/net.c
```diff
@@ -111,15 +111,18 @@ int read_ns_sys_file(char *path, char *buf, int len)
...
 	}
 
-	if (rlen > 0)
-		buf[rlen - 1] = '\0';
+	if (rlen >= 0)
+		buf[rlen] = '\0';
```
kolyshkin, line 125: ack

criu/net.c
```diff
@@ -111,15 +111,18 @@ int read_ns_sys_file(char *path, char *buf, int len)
 	}
 
 	rlen = read(fd, buf, len);
+	if (rlen == -1)
+		pr_perror("Can't read ns' %s", path);
 	close(fd);
 
 	if (rlen == len) {
+		buf[len - 1] = 0;
```
kolyshkin, line 119: nit: it's C so we don't have to be picky about types, but maybe use `\0` as it is more suitable here.

PR #2282, opened 2023-10-10
lib: use separate packages for pycriu and crit

rst0git:
Newer versions of pip use an isolated virtual environment when building Python projects. However, when the source code of CRIT is copied into the isolated environment, the symlink for `../lib/py` (pycriu) becomes invalid. As a workaround, we used the `--no-build-isolation` option for `pip install`. However, this functionality has issues in some versions of PIP (https://github.com/pypa/pip/pull/8221, https://github.com/pypa/pip/issues/8165#issuecomment-625401463). To fix this problem, this patch adds separate packages for pycriu and crit, and each package is installed independently.


adrianreber:
The most important thing from my point of view is that our DEB and RPM packages still work. The last time we changed something with the way the Python bindings are distributed it required some work to get the packaging effort going again.

I think this needs to be tested with RPM and DEB packaging to ensure this time it works better than last time.

I think the best way to ship the Python bindings would be a separate repository, but the way CRIU uses Python during testing would mean that things get much more complicated if the Python code is another repository.

To make sure RPM packaging works there is a packit service which we could run in CI to ensure there is always a working package.

rst0git:
Thank you Adrian, enabling buildability checks is a good idea. This pull request aims to fix the error we currently see in CI with CentOS Stream 9 (https://github.com/checkpoint-restore/criu/runs/17505949509).

Should we open a separate pull request for testing RPM and DEB packaging?

avagin:
> Thank you Adrian, enabling buildability checks is a good idea. This pull request aims to fix the error we currently see in CI with CentOS Stream 9 (https://github.com/checkpoint-restore/criu/runs/17505949509).

Do you understand why it fails only on CentOS 9?

rst0git:
> Do you understand why it fails only on CentOS 9?

The problem that causes this error was introduced in pip version [20.1.1](https://pip.pypa.io/en/stable/news/#v20-1-1)[^1] and fixed in version [21.3](https://pip.pypa.io/en/stable/news/#v21-3)[^2]. These versions of pip have a legacy behaviour that copies the entire project directory to a temporary location and break the [pycriu](https://github.com/checkpoint-restore/criu/blob/42c1c84b95f9ddceb0228416005aab4467650cc0/crit/pycriu) symbolic link. This legacy behaviour has been disabled by default and removed in newer versions of pip.

It currently fails on CentOS 9 because the `python3-pip` package provides version [21.2.3](https://pip.pypa.io/en/stable/news/#v21-2-3).

[^1]: Revert building of local directories in place, restoring the pre-20.1 behaviour of copying to a temporary directory. ([#7555](https://github.com/pypa/pip/issues/7555))
[^2]: In-tree builds are now the default. --use-feature=in-tree-build is now ignored. --use-deprecated=out-of-tree-build may be used temporarily to ease the transition. ([#10495](https://github.com/pypa/pip/issues/10495))

adrianreber (approved):

PR #2288, opened 2023-10-20
Makefile: introduce ARCHCFLAGS for arch specific cflags

marcusfolkesson:
Do not use $(USERCFLAGS) for anything other than what the user provide.

If $(USERCFLAGS) is provided by the user, then the make command will be unable to add machine specific options (used for ARM) resulting in build failures.

Instead, introduce $(ARCHCFLAGS) for such flags.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,18 +35,18 @@ ifeq ($(ARCH),arm)
         ARMV		:= $(shell echo $(SUBARCH) | sed -nr 's/armv([[:digit:]]).*/\1/p; t; i7')
 
         ifeq ($(ARMV),6)
-                USERCFLAGS += -march=armv6
+                ARCHCFLAGS += -march=armv6
         endif
 
         ifeq ($(ARMV),7)
-                USERCFLAGS += -march=armv7-a+fp
+                ARCHCFLAGS += -march=armv7-a+fp
         endif
 
         ifeq ($(ARMV),8)
                 # Running 'setarch linux32 uname -m' returns armv8l on travis aarch64.
                 # This tells CRIU to handle armv8l just as armv7hf. Right now this is
                 # only used for compile testing. No further verification of armv8l exists.
-                USERCFLAGS += -march=armv7-a
+                ARCHCFLAGS += -march=armv7-a
                 ARMV := 7
         endif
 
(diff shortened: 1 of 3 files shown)


adrianreber:
Thanks for the PR. I am not sure why the user should overwrite USERCFLAGS. 

I probably just don't get it totally. Is it somewhere documented that USERCFLAGS can be set or is it custom that a variable with this name is used for that?

I would expect, as a user, to be able to specify cflags via CFLAGS. Just looking for some additional clarification.

marcusfolkesson:
The USERCFLAGS is, what I can tell, not documented.

Based on its name, I assume the purpose is to let the user add additional flags to the compiler.
It is also reasonable to assume that is the purpose since it is used early in the Makefile (only possible input is from user):

ifeq ($(origin HOSTCFLAGS), undefined)
        HOSTCFLAGS := $(CFLAGS) $(USERCFLAGS)
endif

I'm completely okay with that, but reusing the variable and setting it in the Makefile does not work.
If the user sets USERCFLAGS, then the Makefile will be unable to override it, resulting in the -march=... for the ARM architecture is never set.

Similiar problems would be if the user sets CFLAGS on command line.

adrianreber:
For the Fedora package we do: `CFLAGS+=something make`

https://src.fedoraproject.org/rpms/criu/blob/rawhide/f/criu.spec#_115

Would that work for you?

Just had a closer look at `$USERCFLAGS`:
```
$ git grep USERCFLAGS
Makefile:        HOSTCFLAGS := $(CFLAGS) $(USERCFLAGS)
Makefile:                USERCFLAGS += -march=armv6
Makefile:                USERCFLAGS += -march=armv7-a+fp
Makefile:                USERCFLAGS += -march=armv7-a
Makefile:CFLAGS                 += $(USERCFLAGS) $(WARNINGS) $(DEFINES) -iquote include/
```

It doesn't seem to be ever overwritten. Just extended by `+=`, right?

Again, just trying to understand what is correct here. Calling it `$ARCHCFLAGS` sounds like a good idea, but I am still not sure what error you see.

marcusfolkesson:
Hum hum, making it part of the environment would work, IOW:

USERCFLAGS=foo make
works but

make USERFLAGS=foo
does not

Maybe this is no problem then :-)

I would still prefer to not reuse USERCFLAGS for other than user-provided flags though. So maybe just change the commit message?

rppt:
> Maybe this is no problem then :-)
> 
> I would still prefer to not reuse USERCFLAGS for other than user-provided flags though.

It makes sense to me to separate architecture specific flags from USERCFLAGS

> So maybe just change the commit message?

Yes :)
Also, test/zdtm/Makefile.inc should be updated as well

adrianreber (approved):
Looks like a useful change.

rppt (approved):
LGTM

PR #2289, opened 2023-10-20
gitignore: remove historical left-over files

rst0git:
In commit bbc2f133 was introduced a mechanism to auto-generate the files: sys-exec-tbl*.c, syscalls*.S, syscall-codes*.h, and syscall*.h. This commit also updated the gitignore rules to ignore auto-generated files. However, after commit 19fadee9, the path for these files has changed and the patterns specified in gitignore are no longer needed.

Reported-by: @felicitia

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,12 +25,6 @@ images/google/protobuf/*.h
 .gitid
 criu/criu
 criu/unittest/unittest
-criu/arch/*/sys-exec-tbl*.c
-# x86 syscalls-table is not generated
-!criu/arch/x86/sys-exec-tbl.c
-criu/arch/*/syscalls*.S
-criu/include/syscall-codes*.h
-criu/include/syscall*.h
 criu/include/version.h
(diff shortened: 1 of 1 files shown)


Snorch (approved):
Looks good. It seems for me that we have all proper rules for autogenerated syscall related files in compel/.gitignore. So we don't need them in global .gitignore

.gitignore
```diff
@@ -25,12 +25,8 @@ images/google/protobuf/*.h
 .gitid
 criu/criu
 criu/unittest/unittest
-criu/arch/*/sys-exec-tbl*.c
 # x86 syscalls-table is not generated
 !criu/arch/x86/sys-exec-tbl.c
```
0x7f454c46, line 29: Nit: while at here, this one only is needed with the line above.
So you can purge those two lines as well.

0x7f454c46 (approved):
LGTM

PR #2292, opened 2023-10-25
Use the PAGEMAP_SCAN ioctl when it is available 

avagin:


osctobe (approved):

criu/pagemap-cache.c
```diff
@@ -50,10 +51,19 @@ int pmc_init(pmc_t *pmc, pid_t pid, const struct list_head *vma_head, size_t siz
 	pmc->pid = pid;
 	pmc->map_len = PAGEMAP_LEN(map_size);
 	pmc->vma_head = vma_head;
+	pmc->regs_max_len = (map_size / PAGE_SIZE + 1) / 2;
```
osctobe, line 54: Won't this be huge for a big mapping (like ASan or TSan make)? If the cache refill call would be pushed to should_dump_page(), the pagemap cache could be made constant in size as the ioctl() can continue also from an address inside a VMA.

(I guess that since PAGEMAP_SCAN will also get to return SOFT_DIRTY bit, the special handling won't be needed.)

0x7f454c46:
Hmm, at some moment all kernels we regularly run CRIU tests on will have this `PAGEMAP_SCAN`. I'm wondering if it is worth testing pre-ioctl code by artificially setting `has_pagemap_scan = false`? That will make sure that the fallback for older kernels won't get broken.
Not only related to this PR, but probably to other options in kerndat as well. Maybe a wild idea to add a command-line option like `--kerndat-disable-pagemap-scan` and to other kerndat features (rather than adding more fault-injects)?
Likely can be addressed in a separate CRIU issue/PR.

adrianreber:
I had similar thoughts around how to test this. But as it is still marked as draft I was still waiting if tests will be part of this PR.

I like the idea of explicitly selecting kerndat features. It does not make too much sense to expose all the options to the users as it might be confusing. But it probably could be helpful in some cases.

Snorch:
>Maybe a wild idea to add a command-line option like --kerndat-disable-pagemap-scan and to other kerndat features (rather than adding more fault-injects)?

Would not it be cleaner to instead add an ability to read/write/edit kdat cache from crit tool? (Sounds like a simple task for GSOC students or any other newcomers.)

> Likely can be addressed in a separate CRIU issue/PR.

+1

0x7f454c46:
> Would not it be cleaner to instead add an ability to read/write/edit kdat cache from crit tool?

Agree, maybe another tool. Unsure about crit - then we would have to save kerndat in protobuf format and might do the same backward-compatibility as criu does. Hopefully, we can avoid keeping/maintaining compatibility as the intended use is testing-only.

avagin:
> Would not it be cleaner to instead add an ability to read/write/edit kdat cache from crit tool?
Usually, we use injectable faults in such cases. Look at the third patch.

criu/cr-check.c
```diff
@@ -1623,6 +1632,7 @@ static struct feature_list feature_list[] = {
 	{ "openat2", check_openat2 },
 	{ "get_rseq_conf", check_ptrace_get_rseq_conf },
 	{ "ipv6_freebind", check_ipv6_freebind },
+	{ "pagemap_scan", check_pagemap_scan},
```
Snorch, line 1635: missing space before bracket, detected by clang-format

criu/include/pagemap_scan.h
```diff
@@ -0,0 +1,66 @@
...
+#define PAGE_IS_SOFT_DIRTY      (1 << 7)
+
+/*
+ * struct page_region - Page region with flags
+ * @start:      Start of the region
+ * @end:        End of the region (exclusive)
+ * @categories: PAGE_IS_* category bitmask for the region
+ */
+struct page_region {
+        u64 start;
```
Snorch, line 24: spaces instead of tabs, detected by clang-format

criu/include/pagemap_scan.h
```diff
@@ -0,0 +1,66 @@
...
+        u64 end;
+        u64 categories;
+};
+
+#define PAGEMAP_SCAN    _IOWR('f', 16, struct pm_scan_arg)
```
Snorch, line 29: excess spaces "SCAN    _IOWR"

criu/include/pagemap_scan.h
```diff
@@ -0,0 +1,66 @@
...
+ * @vec:                Address of page_region struct array for output
+ * @vec_len:            Length of the page_region struct array
+ * @max_pages:          Optional limit for number of returned pages (0 = disabled)
+ * @category_inverted:  PAGE_IS_* categories which values match if 0 instead of 1
+ * @category_mask:      Skip pages for which any category doesn't match
+ * @category_anyof_mask: Skip pages for which no category matches
+ * @return_mask:        PAGE_IS_* categories that are to be reported in `page_region`s returned
+ */
+struct pm_scan_arg {
+        u64 size;
```
Snorch, line 52: spaces instead of tabs, detected by clang format

criu/mem.c
```diff
@@ -214,9 +243,9 @@ static int generate_iovs(struct pstree_item *item, struct vma_area *vma, struct
 		pages[st]++;
 	}
 
-	*off += pfn * PAGE_SIZE;
```
Snorch, line 246: excess newline, detected by clang-format

criu/mem.c
```diff
@@ -99,38 +99,62 @@ static inline bool __page_in_parent(bool dirty)
...
 
-bool should_dump_page(VmaEntry *vmae, u64 pme)
+static bool should_dump_entire_vma(VmaEntry *vmae)
 {
 	/*
 	 * vDSO area must be always dumped because on restore
 	 * we might need to generate a proxy.
 	 */
-	if (vma_entry_is(vmae, VMA_AREA_VDSO))
+	if (vma_entry_is(vmae, VMA_AREA_VDSO)) {
```
Snorch, line 108: excess bracket

criu/mem.c
```diff
@@ -390,7 +427,6 @@ static int generate_vma_iovs(struct pstree_item *item, struct vma_area *vma, str
 	 * pre-dump cycle has skipped processing & pagemap generation for
 	 * non-PROT_READ regions. So SPLICE mode throws error of missing
 	 * pagemap entry for encountered non-PROT_READ mapping.
-	 *
```
Snorch, line 393: likely accidental line removal

criu/shmem.c
```diff
@@ -206,23 +206,28 @@ static int expand_shmem(struct shmem_info *si, unsigned long new_size)
...
+		if (next != vaddr) {
+			vaddr = next - PAGE_SIZE;
 			continue;
+		}
 
 		shmem_pfn = vma_pfn + DIV_ROUND_UP(vma->pgoff, PAGE_SIZE);
-		if (map[vma_pfn] & PME_SOFT_DIRTY)
+		if (softdirty)
 			set_pstate(si->pstate_map, shmem_pfn, PST_DIRTY);
-		else if (page_is_zero(map[vma_pfn]))
```
Snorch, line 224: @avagin  Can you please explain why we need to remove this hunk of code?

criu/pagemap-cache.c
```diff
@@ -149,39 +160,75 @@ static int pmc_fill_cache(pmc_t *pmc, const struct vma_area *vma)
...
+			.flags = 0,
+			.start = pmc->start,
+			.end = pmc->end,
+			.vec = (long)pmc->regs,
+			.vec_len = pmc->regs_max_len,
+			.max_pages = 0,
+			.category_inverted = PAGE_IS_PFNZERO | PAGE_IS_FILE,
+			.category_mask = PAGE_IS_PFNZERO | PAGE_IS_FILE,
+			.category_anyof_mask = PAGE_IS_PRESENT | PAGE_IS_SWAPPED,
+			.return_mask = PAGE_IS_PRESENT | PAGE_IS_SWAPPED | PAGE_IS_SOFT_DIRTY,
```
Snorch, line 189: Adding a comment to explain what exactly are we querying/filtering for might be nice here.

Snorch:
Looks good.

I found this in kernel doc of PAGEMAP_SCAN ioctl:

> he ``PAGE_IS_WRITTEN`` flag can be considered as a better-performing alternative
of soft-dirty flag. It doesn't get affected by VMA merging of the kernel and hence
the user can find the true soft-dirty pages in case of normal pages. (There may
still be extra dirty pages reported for THP or Hugetlb pages.)

Should we use PAGE_IS_WRITTEN instead of PAGE_IS_SOFT_DIRTY, as it is called "better-performing"?

criu/include/pagemap_scan.h
```diff
@@ -0,0 +1,66 @@
+#ifndef __CR_PAGEMAP_SCAN_H__
+#define __CR_PAGEMAP_SCAN_H__
+
+#include <sys/types.h>
+#include "int.h"
+
```
adrianreber, line 6: Do you need something to handle if the kernel headers are already defining this? Something like `#ifndef PAGEMAP_SCAN` ?

avagin:
> Should we use PAGE_IS_WRITTEN instead of PAGE_IS_SOFT_DIRTY, as it is called "better-performing"?

It is not so simple. PAGE_IS_WRITTEN requires to register all regions in uffd-s and holds them between pre-dumps. It is a separate task.

PR #2293, opened 2023-10-26
plugin: Change default max buffer size to 4G

fdavid-amd:
This will, by default, prevent the amdgpu plugin from allocating more than 4G of memory at a time.

This was meant to be part of a previous pull, but got dropped somewhere.

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -521,7 +521,7 @@ int amdgpu_plugin_init(int stage)
 		getenv_bool("KFD_NUMA_CHECK", &kfd_numa_check);
 		getenv_bool("KFD_CAPABILITY_CHECK", &kfd_capability_check);
 	}
-	kfd_max_buffer_size = 0;
+	kfd_max_buffer_size = 4 << 30;
 	getenv_size_t("KFD_MAX_BUFFER_SIZE", &kfd_max_buffer_size);
 
 	return 0;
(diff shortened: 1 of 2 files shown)


plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -521,7 +521,7 @@ int amdgpu_plugin_init(int stage)
 		getenv_bool("KFD_NUMA_CHECK", &kfd_numa_check);
 		getenv_bool("KFD_CAPABILITY_CHECK", &kfd_capability_check);
 	}
-	kfd_max_buffer_size = 0;
+	kfd_max_buffer_size = 4 << 30;
```
avagin, line 524: amdgpu_plugin.c: In function 'amdgpu_plugin_init':
50499
amdgpu_plugin.c:524:33: error: result of '4 << 30' requires 34 bits to represent, but 'int' only has 32 bits [-Werror=shift-overflow=]
50500
  524 |         kfd_max_buffer_size = 4 << 30;

fdavid-amd:
@dayatsin-amd, could you look at this?

plugins/amdgpu/amdgpu_plugin.c
dayatsin-amd: @fdavid-amd I think size_t can be 32-bits depending on the compiler used. So probably safer to change 
size_t kfd_max_buffer_size;
to:
uint64_t kfd_max_buffer_size;
and most of the occurrences where we use size_t to uint64_t

rerrabolu: Mark the symbol as static If the symbol is used only in amdgpu_plugin.c

avagin:
Any updates?

rerrabolu:
I am not tracking this change.

plugins/amdgpu/amdgpu_plugin.c
dayatsin-amd: @fdavid-amd any ETA on these changes?

PR #2298, opened 2023-11-10
Refactor criu plugin baseline v2

rerrabolu:

### DIFF
--- a/plugins/amdgpu/Makefile
+++ b/plugins/amdgpu/Makefile
@@ -28,7 +28,7 @@ endif
 criu-amdgpu.pb-c.c: criu-amdgpu.proto
 		protoc-c --proto_path=. --c_out=. criu-amdgpu.proto
 
-amdgpu_plugin.so: amdgpu_plugin.c amdgpu_plugin_topology.c criu-amdgpu.pb-c.c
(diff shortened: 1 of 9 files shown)


rerrabolu:
My changelist fails for two scenarios:

1. Cirrus CI / Vagrant Fedora Rawhide based test
2. Cirrus CI / Vagrant Fedora based test (no VDSO)

Is it possible for me to re-run them as the "details" does not show me if my change is the cause. Furthermore, I see other changelists that are having failures in the same validating element.

avagin:
@dayatsin-amd friendly ping

rst0git:
@rerrabolu Would it be possible to add tests that can be used to verify the functionality on a system with GPU?

rerrabolu:
- Thanks for taking time to merge my issue.
- I see some lint errors, will try to provide a patch to address it next couple of weeks
- I will find out if we could add a small test case to validate CRIU on a system with AMD GPUs. This will take sometime and will require some coordination

rst0git:
@rerrabolu Some of the lint errors have been fixed in https://github.com/checkpoint-restore/criu/pull/2345.

I have been using the example applications from the following repository for testing locally with an MI100 GPU:
https://github.com/ROCm/HIP-Examples/tree/master/HIP-Examples-Applications

The amdgpu plugin is [already enabled](https://github.com/checkpoint-restore/criu/pull/2276) in CI and perhaps we could add a few simple test cases to the [ZDTM test suite](https://criu.org/ZDTM_test_suite).

PR #2301, opened 2023-11-28
compel: correct the syscall number of bind on ARM64

SallyKAN:
In the compel/arch/arm/plugins/std/syscalls/syscall.def, the syscall number of bind on ARM64 should be 200 instead of 235
Referring to [this](https://gpages.juszkiewicz.com.pl/syscalls-table/syscalls.html)


adrianreber:
Looks good. Thanks.

PR #2302, opened 2023-11-28
ci: fix docker test

rst0git:
The version of docker-ce on Ubuntu 20.04 has been recently updated to [25.0.0-beta.1](https://github.com/moby/moby/discussions/46816). However, with this version `docker start --checkpoint` fails with the following error:

```
$ docker start --checkpoint=c1 cr
Error response from daemon: failed to create task for container: content digest fdb1054b00a8c07f08574ce52198c5501d1f552b6a5fb46105c688c70a9acb45: not found: unknown
```

As a workaround, we install the most recent stable version of docker-ce. In addition, to improve code readability, the recursive function call used for retrying container restoration has been replaced with a loop.

### DIFF
--- a/scripts/ci/docker-test.sh
+++ b/scripts/ci/docker-test.sh
@@ -87,27 +87,25 @@ print_logs () {
 }
 
 declare -i max_restore_container_tries=3
-current_iteration=
 
 restore_container () {
 	CHECKPOINT_NAME=$1
 
-	docker start --checkpoint "$CHECKPOINT_NAME" cr 2>&1 | tee log || {
+	for i in $(seq $max_restore_container_tries); do
+		docker start --checkpoint "$CHECKPOINT_NAME" cr 2>&1 | tee log && break
+
 		# FIXME: There is a race condition in docker/containerd that causes
 		# docker to occasionally fail when starting a container from a
 		# checkpoint immediately after the checkpoint has been created.
 		# https://github.com/moby/moby/issues/42900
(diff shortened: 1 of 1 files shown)


0x7f454c46:
I remember there was quite a protest from vz guys for bashisms as they don't have bash in the testing setup (?).
But this file already has bashisms: `((current_iteration+=1))`, so it shouldn't be any worse than before.

Just for curiosity: is it also fixed by pre-initializing `current_iteration=0`?

avagin:
@0x7f454c46 I remember I was to tired with all these shellshit and replaced /bin/sh with /bin/bash everywhere, so bash is required to run tests.

scripts/ci/docker-test.sh
```diff
@@ -87,6 +87,8 @@ print_logs () {
 }
 
 declare -i max_restore_container_tries=3
+# $current_iteration is used to track the number of times we have tried to
+# restore a container. It is reset to 0 when a restore succeeds.
 current_iteration=
```
avagin, line 92: pls replace a recursive call with a loop and move current_iteration in the restore_container function.

Snorch:
> I remember there was quite a protest from vz guys for bashisms as they don't have bash in the testing setup (?).

Thanks for the note, I was not aware of it =)

But I know for sure that we don't run `scripts/ci/docker-test.sh` in vz ci, so adding bashisms here should be ok anyway.

PR #2304, opened 2023-11-29
ci: fix rawhide netlink error

adrianreber:
The rawhide netlink errors are fixed with a newer kernel than the default 6.2 available in Fedora 38.

### DIFF
--- a/scripts/ci/vagrant.sh
+++ b/scripts/ci/vagrant.sh
@@ -57,6 +57,11 @@ fedora-no-vdso() {
 }
 
 fedora-rawhide() {
+	# The 6.2 kernel of Fedora 38 in combination with rawhide userspace breaks
+	# zdtm/static/socket-tcp-nfconntrack. To activate the new kernel previously
+	# installed this reboots the VM.
+	vagrant reload
+	ssh default uname -a
 	#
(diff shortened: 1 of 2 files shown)


scripts/ci/vagrant.sh
```diff
@@ -57,6 +57,11 @@ fedora-no-vdso() {
 }
 
 fedora-rawhide() {
+	# The 6.2 kernel of Fedora 38 in combination with rawhide userspace breaks
+	# zdtm/static/socket-tcp-nfconntrack. To activate the new kernel previously
+	# installed this reboots the VM.
+	vagrant reload
+	ssh default uname -a
```
rst0git, line 64: Would it be better to move `vagrant reload` in `setup()` for consistency across tests?

adrianreber:
Sorry. I merged this. This was not my intention.

adrianreber:
@avagin Can you undo this while rebasing? I can also to a revert. Sorry.

scripts/ci/vagrant.sh
adrianreber: I was thinking about it, but if we do not need the latest kernel rebooting takes additional time.

PR #2307, opened 2023-11-30
ci: do not use 'tail' for skip-file-rwx-check test

adrianreber:
Newer versions of 'tail' rely on inotify and after a restore 'tail' is unhappy with the state of inotify and just stops.

This replaces 'tail' with a minimal test in C which just keeps a file open to still be able to test how CRIU reacts if the file mode has changed.

The skip-file-rwx-check test sometimes fails if `kill` is slower than `tail` ending itself.

### DIFF
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -288,11 +288,16 @@ ip net add test
 
 # Rootless tests
 # Check if cap_checkpoint_restore is supported and also if unshare -c is supported.
-if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true; then
+#
+# Do not run this test in a container (see https://github.com/checkpoint-restore/criu/issues/2312).
+# This is a temporary workaround until fixed in the kernel.
+# The kernel currently does not show correct device and inode numbers in /proc/pid/maps
+# for stackable file systems.
+if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true && [ ! -e /run/.containerenv ]; then
 	make -C test/zdtm/ cleanout
 	rm -rf test/dump
 	setcap cap_checkpoint_restore,cap_sys_ptrace+eip criu/criu
-	if [ -d /sys/fs/selinux ]; then
+	if [ -d /sys/fs/selinux ] && command -v getenforce &>/dev/null; then
 		# Note: selinux in Enforcing mode prevents us from calling clone3() or writing to ns_last_pid on restore; hence set to Permissive for the test and then set back.
 		selinuxmode=$(getenforce)
 		setenforce Permissive
@@ -301,7 +306,7 @@ if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true; then
 	# under those conditions. Note that the "... && true" part is necessary; we need at least one statement after the tests so that bash can reap zombies in the user namespace,
 	# otherwise it will exec the last statement and get replaced and nobody will be left to reap our zombies.
 	sudo --user=#65534 --group=#65534 unshare -Ucfpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true"
-	if [ -d /sys/fs/selinux ]; then
+	if [ -d /sys/fs/selinux ] && command -v getenforce &>/dev/null; then
 		setenforce "$selinuxmode"
 	fi
 	setcap -r criu/criu
--- a/scripts/ci/vagrant.sh
+++ b/scripts/ci/vagrant.sh
@@ -70,6 +70,10 @@ fedora-rawhide() {
 	#
 	ssh default 'sudo dnf remove -y crun || true'
 	ssh default sudo dnf install -y podman runc
+	# Some tests in the container need selinux to be disabled.
+	# In the container it is not possible to change the state of selinux.
+	# Let's just disable it for this test run completely.
+	ssh default 'sudo setenforce Permissive'
 	ssh default 'cd /vagrant; tar xf criu.tar; cd criu; sudo -E make -C scripts/ci fedora-rawhide CONTAINER_RUNTIME=podman BUILD_OPTIONS="--security-opt seccomp=unconfined"'
 }
 
--- a/scripts/ci/prepare-for-fedora-rawhide.sh
+++ b/scripts/ci/prepare-for-fedora-rawhide.sh
@@ -18,6 +18,7 @@ dnf install -y \
 	libnet-devel \
 	libnl3-devel \
 	libbsd-devel \
+	libselinux-utils \
 	make \
 	procps-ng \
 	protobuf-c-devel \


scripts/ci/vagrant.sh
```diff
@@ -70,6 +70,10 @@ fedora-rawhide() {
 	#
 	ssh default 'sudo dnf remove -y crun || true'
 	ssh default sudo dnf install -y podman runc
+	# Some tests in the container need selinux to be disabled.
+	# In the container it is not possible to change the state of selinux.
+	# Let's just disable it for this test run completely.
+	ssh default 'setenforce Permissive'
```
rst0git, line 76: Should it be `sudo setenforce Permissive`?
```
+ ssh default 'setenforce Permissive'
setenforce:  security_setenforce() failed:  Permission denied
make: *** [Makefile:73: vagrant-fedora-rawhide] Error 2
```

adrianreber: Yes. Thanks. Changed.

test/others/skip-file-rwx-check/run.sh
```diff
@@ -8,21 +8,22 @@ set -o xtrace
 source ../env.sh
 
 make clean
+make tester
 touch testfile
 chmod +w testfile
-tail --follow testfile &
-tailpid=$!
-if ! "$criu" dump --tree=$tailpid --shell-job --verbosity=4 --log-file=dump.log
+./tester testfile &
```
avagin, line 14: nit: you can use bash to open a file descriptor:
bash -c 'exec 3<testfile; while :; do sleep 1; done' &

test/others/skip-file-rwx-check/tester.c
```diff
@@ -0,0 +1,10 @@
+#include <fcntl.h>
+#include <unistd.h>
+
+int main(int argc, char *argv[])
+{
+	open(argv[1], O_RDONLY | O_APPEND);
```
avagin, line 6: need to check a return code and exit if open fails.

test/others/skip-file-rwx-check/run.sh
adrianreber: Thanks. Using that now and dropping the C file.

test/others/skip-file-rwx-check/tester.c
adrianreber: Removed C code.

scripts/ci/run-ci-tests.sh
```diff
@@ -301,7 +301,7 @@ if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true; then
 	# under those conditions. Note that the "... && true" part is necessary; we need at least one statement after the tests so that bash can reap zombies in the user namespace,
 	# otherwise it will exec the last statement and get replaced and nobody will be left to reap our zombies.
 	sudo --user=#65534 --group=#65534 unshare -Ucfpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true"
```
rst0git, line 303: This test currently fails with the following error:
```
=[log]=> dump/zdtm/static/maps00/66/1/dump.log
b'(00.000000) CRIU run id = 0xf000022600000043'
b'(00.000061) Version: 3.18 (gitid 0)'
b'(00.000070) Running on f4a8025d70c9 Linux 6.6.2-101.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Nov 22 21:31:13 UTC 2023 x86_64'
b'(00.000073) Warn  (criu/kerndat.c:1153): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.007422) cpu: fpu: AVX-512 ZMM_Hi256                xstate_offsets   1664 / 1408   xstate_sizes   1024 / 1024'
b'(00.007513) Effective capability 12 missing'
b'(00.008004) Found mmap_min_addr 0x10000'
b'(00.008018) files stat: fs/nr_open 1073741816'
b'(00.008022) Warn  (criu/kerndat.c:1153): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.011392) Handling VMA with the following smaps entry: 7fb21b401000-7fb21b404000 rw-p 00000000 00:00 0'
b'(00.011425) Handling VMA with the following smaps entry: 7fb21b404000-7fb21b42a000 r--p 00000000 00:22 88823                      /usr/lib64/libc.so.6'
b'(00.011434) Failed to open map_files/7fb21b404000-7fb21b42a000, try to go via [/usr/lib64/libc.so.6] path'
b"(00.011456) mnt: Path `/usr/lib64/libc.so.6' resolved to `./' mountpoint"
b'(00.011458) Error (criu/proc_parse.c:379): Failed to resolve mapping 7fb21b404000 filename'
b"(00.011561) Error (criu/proc_parse.c:693): Can't open 66's mapfile link 7fb21b404000: Operation not permitted"
b'(00.011570) Error (criu/cr-dump.c:1563): Collect mappings (pid: 66) failed with -1'
b'(00.011629) net: Unlock network'
b'(00.011634) Unfreezing tasks into 1'
b'(00.011637) \tUnseizing 66 into 1'
b'(00.011655) Error (criu/cr-dump.c:2098): Dumping FAILED.'
```

Would it make sense to move `[ ! -e /run/.containerenv ]` in the following if statement to skip the test when running in a container?
```
# Rootless tests
# Check if cap_checkpoint_restore is supported and also if unshare -c is supported.
if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true; then
```

adrianreber: Yes, that is what I want to do next.

scripts/ci/run-ci-tests.sh
```diff
@@ -288,7 +288,8 @@ ip net add test
 
 # Rootless tests
 # Check if cap_checkpoint_restore is supported and also if unshare -c is supported.
-if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true; then
+# Do not run this test in a container.
+if capsh --supports=cap_checkpoint_restore && unshare -c /bin/true && [ ! -e /run/.containerenv ]; then
```
avagin, line 292: I can't understand why we can run root tests but not non-root. Pls write more details in the comment and in the commit message.

adrianreber: As far as I understand this mount related:

```
mnt: Path `/usr/lib64/libc.so.6' resolved to `./' mountpoint"
```

I don't know if and how to correctly fix it, but it seems using a mount namespace in a mount namespace does not work as expected.

The reproducer for me is. This works without a container:
```
setcap cap_checkpoint_restore,cap_sys_ptrace+eip criu/criu
sudo --user=#65534 --group=#65534 unshare -Ucfpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -k always  -f h --rootless && true"
```
But it does not work in a container. @Snorch can you say if this should work and how?

Snorch: > mnt: Path `/usr/lib64/libc.so.6' resolved to `./' mountpoint"

This just says that the mentioned file is on the root filesystem. This is not an error message. Sharing some error message would be nice.

I didn't manage to get anywhere with the above reproduce:
```
[root@turmoil criu]# sudo --user=#65534 --group=#65534 unshare -Ucfpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -k always  -f h --rootless && true"
userns is supported
=== Run 1/1 ================ zdtm/static/maps00
Start test
  File "/criu/test/zdtm.py", line 2847, in <module>
    fork_zdtm()
  File "/criu/test/zdtm.py", line 2838, in fork_zdtm
  File "/criu/test/zdtm.py", line 1932, in do_run_test
    cr(cr_api, t, opts)
  File "/criu/test/zdtm.py", line 1577, in cr
    cr_api.set_test(test)
  File "/criu/test/zdtm.py", line 1121, in set_test
    os.makedirs(self.__dump_path)
  File "<frozen os>", line 215, in makedirs
  File "<frozen os>", line 225, in makedirs
PermissionError: [Errno 13] Permission denied: 'dump/zdtm/static/maps00'
```
Just strange file permission problems, no mount problems.

adrianreber: If you look at the run-ci.sh script you see a couple of chmod 777. You need to run those. Just before the sudo command there is also a rm which you probably need to run.

avagin: > I don't know if and how to correctly fix it, but it seems using a mount namespace in a mount namespace does not work as expected.

Strictly speaking, it is unacceptable to skip test without knowing exact reasons. In this case, the ability of C/R-ing in a container is one of target use-cases.

rst0git: > Sharing some error message would be nice.

The following error messages are from the logs in CI.
https://cirrus-ci.com/task/5101821836918784
```
b"(00.011456) mnt: Path `/usr/lib64/libc.so.6' resolved to `./' mountpoint"
b'(00.011458) Error (criu/proc_parse.c:379): Failed to resolve mapping 7fb21b404000 filename'
b"(00.011561) Error (criu/proc_parse.c:693): Can't open 66's mapfile link 7fb21b404000: Operation not permitted"
b'(00.011570) Error (criu/cr-dump.c:1563): Collect mappings (pid: 66) failed with -1'
```

@adrianreber Looking at the CI logs below, is it possible that the test fails because selinux is enabled on the host and we can't disable it from within the container?
```
++ getenforce
+ selinuxmode=Disabled
+ setenforce Permissive
setenforce: SELinux is disabled
```
https://cirrus-ci.com/task/4965372655304704?logs=build#L12622

adrianreber: > Strictly speaking, it is unacceptable to skip test without knowing exact reasons. In this case, the ability of C/R-ing in a container is one of target use-cases.

You are right. But my goal with this PR is to get CI green again. I think it is more important to have useful CI results than keeping broken tests running in CI. If a test is always red it might hide new errors introduced by new commits and becomes less useful. This test run alone had three broken tests and the first broken test was hiding the two follow-up errors. So this might have been merged without knowing that it does not work in all setups.

Anyway, if you prefer to not have this PR, no problem, I can close it.

adrianreber: > @adrianreber Looking at the CI logs below, is it possible that the test fails because selinux is enabled on the host and we can't disable it from within the container?

SELinux does not change anything on my SELinux enabled system when I reproduce it locally. Initially I thought it might be related to SELinux. Outside of a container it is, SELinux blocks checkpoint/restore in a user namespace. But creating a user namespace inside of a container SELinux does not seem to affect the results. Do you see other results?

rst0git: I was able to replicate the error locally and does not seem to be related to SELinux.
It occurs only when running the test in user namespace and CRIU fails in `phys_stat_dev_match()` because the value returned from `phys_stat_resolve_dev()` is not equal to `phys_dev`.

https://github.com/checkpoint-restore/criu/blob/b17a73b2ee21d64a7cc59968f39e784618837397/criu/mount.c#L297

The test seems to work without user namespace:
```
$ unshare -fpm --mount-proc -- bash -c "./test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true"
userns is supported
=== Run 1/1 ================ zdtm/static/maps00
Start test
Run criu dump
=[log]=> dump/zdtm/static/maps00/66/1/dump.log
b'(00.000000) CRIU run id = 0xf000058100000043'
b'(00.000141) Version: 3.18 (gitid 0)'
b'(00.000169) Running on e07aabb71f84 Linux 6.5.10-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Nov  2 19:59:55 UTC 2023 x86_64'
b'(00.000180) Warn  (criu/kerndat.c:1153): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.022561) cpu: fpu: MPX CSR                          xstate_offsets   1024 / 832    xstate_sizes     64 / 64'
b'(00.023064) Effective capability 12 missing'
b'(00.024829) Found mmap_min_addr 0x10000'
b'(00.024880) files stat: fs/nr_open 1073741816'
b'(00.024893) Warn  (criu/kerndat.c:1153): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
Run criu restore
=[log]=> dump/zdtm/static/maps00/66/1/restore.log
b'(00.000000) CRIU run id = 0xf00005810000004b'
b'(00.000109) Version: 3.18 (gitid 0)'
b'(00.000131) Running on e07aabb71f84 Linux 6.5.10-200.fc38.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Nov  2 19:59:55 UTC 2023 x86_64'
b'(00.000141) Warn  (criu/kerndat.c:1153): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.021152) cpu: fpu: MPX CSR                          xstate_offsets   1024 / 832    xstate_sizes     64 / 64'
b'(00.021723) Effective capability 12 missing'
b'(00.023582) Found mmap_min_addr 0x10000'
b'(00.023641) files stat: fs/nr_open 1073741816'
b'(00.023655) Warn  (criu/kerndat.c:1153): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
Send the 15 signal to  66
Wait for zdtm/static/maps00(66) to die for 0.100000
Removing dump/zdtm/static/maps00/66
========================= Test zdtm/static/maps00 PASS =========================
```

avagin: > It occurs only when running the test in user namespace and CRIU fails in phys_stat_dev_match() because the value returned from phys_stat_resolve_dev() is not equal to phys_dev.

It is one of the known overlayfs issues. Cc: @mihalicyn

avagin: These changes have to fix the issue in the kernel:
https://github.com/avagin/linux-task-diag/commits/overlay-vs-maps

adrianreber: See #2312. I will merge this with a comment mentioning the ticket and that it can be removed in the future.

adrianreber:
Failing alpine test is tracked here: https://github.com/checkpoint-restore/criu/issues/2313

PR #2311, opened 2023-12-08
tty: skip ioctl(TIOCSLCKTRMIOS) if possible

mihalicyn:
If ioctl(TIOCSLCKTRMIOS) fails with EPERM it means that a CRIU process lacks of CAP_SYS_ADMIN capability. But we can use ioctl(TIOCGLCKTRMIOS) to *read* current ->termios_locked value from the kernel and if it's the same as we already have we can skip failing ioctl(TIOCSLCKTRMIOS) safely.

Adrian has recently posted [1] a very good patch to allow ioctl(TIOCSLCKTRMIOS) for processes that have CAP_CHECKPOINT_RESTORE (right now it requires CAP_SYS_ADMIN).

[1] https://lore.kernel.org/all/20231206134340.7093-1-areber@redhat.com/

Suggested-by: Andrei Vagin

### DIFF
--- a/criu/tty.c
+++ b/criu/tty.c
@@ -817,8 +817,26 @@ static int do_restore_tty_parms(void *arg, int fd, pid_t pid)
 	 * on termios too. Just to be on the safe side.
 	 */
 
-	if ((p->has & HAS_TERMIOS_L) && ioctl(fd, TIOCSLCKTRMIOS, &p->tl) < 0)
-		goto err;
+	if ((p->has & HAS_TERMIOS_L) && ioctl(fd, TIOCSLCKTRMIOS, &p->tl) < 0) {
+		struct termios t;
+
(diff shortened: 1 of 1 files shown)


mihalicyn:
I have tested this with example from our wiki (https://criu.org/Simple_loop#A_shell_job):
```
echo 0 > /proc/sys/kernel/yama/ptrace_scope
setcap cap_checkpoint_restore+eip ./criu/criu

./criu/criu dump -vvvv -o dump.log -t 184826 --shell-job --unprivileged
./criu/criu restore -vvvv -o restore.log --shell-job --unprivileged
```

And it works perfectly fine for this case.

avagin:
lgtm. Thanks.

avagin (approved):

rst0git (approved):
LGTM

PR #2319, opened 2023-12-28
make: fix compilation on alpine

avagin:
Starting with the musl v1.2.4~69, _GNU_SOURCE doesn't set _LARGEFILE64_SOURCE.


rst0git (approved):
LGTM

Issue #2313, opened 2023-12-11
Alpine test/build fails


avagin:
```
#10 7.842   CC       criu/pie/restorer.o
#10 7.926 criu/pie/restorer.c: In function 'restore_rlims':
#10 7.926 criu/pie/restorer.c:453:43: error: invalid use of undefined type 'struct rlimit64'
#10 7.926   453 |                 krlim.rlim_cur = ta->rlims[r].rlim_cur;
#10 7.926       |                                           ^
#10 7.926 criu/pie/restorer.c:453:46: error: invalid use of undefined type 'struct rlimit64'
#10 7.926   453 |                 krlim.rlim_cur = ta->rlims[r].rlim_cur;
#10 7.926       |                                              ^
#10 7.927 criu/pie/restorer.c:454:43: error: invalid use of undefined type 'struct rlimit64'
#10 7.927   454 |                 krlim.rlim_max = ta->rlims[r].rlim_max;
#10 7.927       |                                           ^
#10 7.927 criu/pie/restorer.c:454:46: error: invalid use of undefined type 'struct rlimit64'
#10 7.927   454 |                 krlim.rlim_max = ta->rlims[r].rlim_max;
#10 7.927       |                                              ^
#10 7.969 At top level:
#10 7.969 cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics
#10 7.970 make[2]: *** [/criu/scripts/nmk/scripts/build.mk:118: criu/pie/restorer.o] Error 1
#10 7.970 make[2]: *** Waiting for unfinished jobs....
#10 8.040 make[1]: *** [criu/Makefile:59: pie] Error 2
#10 8.041 make: *** [Makefile:276: criu] Error 2
#10 ERROR: process "/bin/sh -c make mrproper && date && make -j $(nproc) CC=\"$CC\" && date" did not complete successfully: exit code: 2
```

https://github.com/checkpoint-restore/criu/actions/runs/7171767723/job/19527463844

avagin:
I was probably introduced by https://github.com/guidosarducci/musl/commit/25e6fee27f4a293728dd15b659170e7b9c7db9bc Cc: @richfelker

avagin:
```
2023-12-28T07:09:22.4352258Z =================== Run zdtm/static/netns_lock_iptables in h ===================
2023-12-28T07:09:22.4353045Z Start test
2023-12-28T07:09:22.4353418Z Test is SUID
[...]

PR #2321, opened 2023-12-30
build/alpine: install iptables-legacy

avagin:
It is required to C/R iptables rules.


rst0git:
@avagin I think that this is related to https://github.com/checkpoint-restore/criu/pull/1976

avagin:
@rst0git no, it isn't. It was an attempt to fix #2313.

PR #2323, opened 2024-01-05
net: fix network unlock with iptables-nft

rst0git:
When iptables-nft is used as backend for iptables, the rules for network locking are translated into the following nft rules:

```
$ iptables-restore-translate -f lock.txt
add table ip filter
add chain ip filter CRIU
insert rule ip filter INPUT counter jump CRIU
insert rule ip filter OUTPUT counter jump CRIU
add rule ip filter CRIU mark 0xc114 counter accept
add rule ip filter CRIU counter drop
```

These rules create the following chains:

```
table ip filter { # handle 1
	chain CRIU { # handle 1
		meta mark 0x0000c114 counter packets 16 bytes 890 accept # handle 6
		counter packets 1 bytes 60 drop # handle 7
		meta mark 0x0000c114 counter packets 0 bytes 0 accept # handle 8
		counter packets 0 bytes 0 drop # handle 9
	}

	chain INPUT { # handle 2
		type filter hook input priority filter; policy accept;
		counter packets 8 bytes 445 jump CRIU # handle 3
		counter packets 0 bytes 0 jump CRIU # handle 10
	}

	chain OUTPUT { # handle 4
		type filter hook output priority filter; policy accept;
		counter packets 9 bytes 505 jump CRIU # handle 5
		counter packets 0 bytes 0 jump CRIU # handle 11
	}
}
```

In order to delete the CRIU chain, we need to first delete all four jump targets. Otherwise, `-X CRIU` would fail with the following error:
```
iptables-restore v1.8.10 (nf_tables):
line 5: CHAIN_DEL failed (Resource busy): chain CRIU
```

Fixes: #2313


mihalicyn:
Great job, Radostin!

LGTM.

to discuss: probably at some point it makes sense to change `NETWORK_LOCK_DEFAULT` value to `NETWORK_LOCK_NFTABLES`.

mihalicyn (approved):

test/zdtm/static/socket-tcp.c
```diff
@@ -67,17 +67,38 @@ int main(int argc, char **argv)
...
+		pr_perror("unshare");
 		return 1;
-	if (system("iptables -w -A INPUT -j DROP"))
+	}
+	if (system("ip link set up dev lo"))
+		return 1;
+
+	if (system("nft add table ip filter"))
 		return 1;
+	if (system("nft add chain ip filter INPUT"))
```
Snorch, line 95: @rst0git should not we need to specify hook for it?

```
nft add chain ip filter INPUT { type filter hook input priority 0 ; }
```

criu/net.c
```diff
@@ -3178,19 +3190,53 @@ static inline int nftables_network_unlock(void)
 #endif
 }
 
+static int iptables_has_criu_jump_target(void)
```
Snorch, line 3193: This is a confusing name, it returns 0 on success, but in c zero is false, and from the name of the function it should imply that "iptables does not have criu jump target".

rst0git: Thank you Pavel! I have opened the following pull request to address this: https://github.com/checkpoint-restore/criu/pull/2337

PR #2331, opened 2024-01-15
Don't dump pages which only contain zero bytes

simonis:
I want to propose a new command line option `zero-pages` which, if enabled, detects pages which only contain zero bytes and skip their dumping to the image file. At restore, such pages will be replaced by the kernels zero page automatically.This can be useful for runtimes like Java, which often allocate large memory regions without fully using them (e.g. for the heap). For a simple Helloworld Java program, this new feature shrinks the image size be about 20% from 13mb to 11mb.

I've enabled GitHub Actions on my fork and all the test are green except the Alpine test which I see failing on upstream as well and the Code Linter test where I can't understand what he's objecting.

I'm a first time contributor to this project so please let me know if I've missed something.


rst0git:
> This can be useful for runtimes like Java, which often allocate large memory regions without fully using them (e.g. for the heap). For a simple Helloworld Java program, this new feature shrinks the image size be about 20% from 13mb to 11mb.

@simonis It might be more effective to address this problem using compression for memory pages, similar to the approach used in [CRaC](https://github.com/CRaC/criu).

simonis:
> > This can be useful for runtimes like Java, which often allocate large memory regions without fully using them (e.g. for the heap). For a simple Helloworld Java program, this new feature shrinks the image size be about 20% from 13mb to 11mb.
> 
> @simonis It might be more effective to address this problem using compression for memory pages, similar to the approach used in [CRaC](https://github.com/CRaC/criu).

I think compressing the image as done [criu-crac](https://github.com/CRaC/criu/pull/1) is somewhat orthogonal. It decreases the image size which is nice if you have to transfer it to another host. But before restore, it decompresses the image again and loads all its content (even the pages containing just zeros) into memory. The zeros will have to be copied into the target process which takes time and also increases the initial RSS of the target process.

With this proposed solution, fewer memory has to be copied to the target process and the target processes initial RSS will be smaller. The zero pages will be COWed gradually as they are needed which leads to a faster startup.

adrianreber:
I think the code linter wants a \n at the end of the warning.

Without looking too closely, the idea of this PR sounds like a good one.

simonis:
> I think the code linter wants a \n at the end of the warning.
> 

Thanks, I've fixed that now.

> Without looking too closely, the idea of this PR sounds like a good one.

Thanks.

criu/crtools.c
```diff
@@ -541,6 +541,12 @@ int main(int argc, char *argv[], char *envp[])
 	       "                        pages images of previous dump\n"
 	       "                        when used on restore, as soon as page is restored, it\n"
 	       "                        will be punched from the image\n"
+	       "  --zero-pages          don't dump pages containing only zero bytes. This is a\n"
```
rst0git, line 544: The name `--skip-zero-pages` might be more appropriate for this option.

criu/crtools.c
```diff
@@ -541,6 +541,12 @@ int main(int argc, char *argv[], char *envp[])
...
+	       "                        potentially expensive operation because it checks for\n"
+	       "                        every single process page if it contains only zeros but\n"
+               "                        it can significantly decrease the image size if many\n"
+               "                        such pages exist. It effectively replaces such pages\n"
+	       "                        which the kernel's zero-page on restore.\n"
```
rst0git, line 549: Would it be possible to move most of the details in the man page?
https://github.com/checkpoint-restore/criu/blob/criu-dev/Documentation/criu.txt

criu/mem.c
```diff
@@ -172,17 +175,58 @@ static int generate_iovs(struct pstree_item *item, struct vma_area *vma, struct
 	unsigned long pages[3] = {};
 	int ret = 0;
 
+	static char *ZERO_PAGE = NULL;
+	static char *REMOTE_PAGE = NULL;
```
rst0git, line 179: We tend to follow the [kernel coding style](https://www.kernel.org/doc/html/v4.10/process/coding-style.html) where variables are usually not capitalized.

criu/include/cr_options.h
```diff
@@ -157,6 +157,7 @@ struct cr_options {
 	int track_mem;
 	char *img_parent;
 	int auto_dedup;
+	int zero_pages;
```
rst0git, line 160: Would it be possible to add support for this option in [cr-service](https://github.com/checkpoint-restore/criu/blob/cda1c5c95aaca5affa28daa59d7f25ed64007a78/criu/cr-service.c#L288) and [libcriu](https://github.com/checkpoint-restore/criu/blob/criu-dev/lib/c/criu.c)?

criu/config.c
```diff
@@ -650,6 +650,7 @@ int parse_options(int argc, char **argv, bool *usage_error, bool *has_exec_cmd,
 		{ "ms", no_argument, 0, 1054 },
 		BOOL_OPT("track-mem", &opts.track_mem),
 		BOOL_OPT("auto-dedup", &opts.auto_dedup),
+		BOOL_OPT("zero-pages", &opts.zero_pages),
```
rst0git, line 653: To run the CI tests with this option, it also needs to be added in [`zdtm.py`](https://github.com/checkpoint-restore/criu/blob/cda1c5c95aaca5affa28daa59d7f25ed64007a78/test/zdtm.py#L2657)

criu/crtools.c
simonis: Agree.

criu/crtools.c
simonis: Done

criu/mem.c
simonis: Got it. Fixed.

criu/config.c
simonis: I think I've still not fully understood how your testing framework works, but I did my best to add support for `--skip-zero-pages` to it. Please let me know if there's still something missing.

criu/include/cr_options.h
simonis: I tried to do it but I must confess that I'm not familiar with this stuff so please take a second look and let me know if there's anything left?

criu/mem.c
```diff
@@ -172,17 +175,58 @@ static int generate_iovs(struct pstree_item *item, struct vma_area *vma, struct
...
+				remote[0].iov_base = (void*)vaddr;
+				nread = process_vm_readv(item->pid->real, local, 1, remote, 1, 0);
+				if (nread == PAGE_SIZE) {
+					zero = memcmp(zero_page, remote_page, PAGE_SIZE);
+					/*
+					 * If the page contains just zeros we can treat it like the zero page and skip it.
+					 * At restore it will be replaced by a reference to the zero page and COWed if accessed.
+					 */
+					if (zero == 0) {
+						pr_info("Zero page detected at virtual addr = %p\n", (void*)vaddr);
```
rst0git, line 222: Using `pr_info()` here is very expensive. Would it be possible to remove it?

rst0git:
> Code Linter test where I can't understand what he's objecting

@simonis Would it be possible to run `make indent` before committing your changes? This should fix the code style problems such as missing spaces. In addition, it would be great if you can add a commit message with a description that explains *what* changes are introduced and *why*. The following blog post describes how to write good commit messages: https://cbea.ms/git-commit/

test/javaTests/test-zero.xml
```diff
@@ -0,0 +1,89 @@
+<?xml version = "1.0" encoding = "UTF-8"?>
+<!DOCTYPE suite SYSTEM "http://testng.org/testng-1.0.dtd" >
+
+<suite name = "Suite2">
+	<parameter name="checkpointOpt" value="--zero-pages"/>
```
rst0git, line 5:
```Suggestion
	<parameter name="checkpointOpt" value="--skip-zero-pages"/>
```

test/javaTests/test-zero.xml
```diff
@@ -0,0 +1,89 @@
...
+	<test name = "test5-MultipleFileWrite">
+		<parameter name="testname" value="MultipleFileWrite"/>
+		<classes>
+			<class name = "org.criu.java.tests.CheckpointRestore"/>
+		</classes>
+	</test>
+
+	<test name = "test6-Sockets">
+		<parameter name="testname" value="Sockets"/>
+		<parameter name="checkpointOpt" value="--tcp-established --zero-pages"/>
```
rst0git, line 45: ditto

criu/mem.c
simonis: Sure, I can remove it. I mainly used it during development. Do you think `pr_debug()` would be more appropriate, oder better remove it altogether?

rst0git: Removing it altogether would be better because most CRIU users set verbosity level to debug.

criu/mem.c
```diff
@@ -172,17 +175,58 @@ static int generate_iovs(struct pstree_item *item, struct vma_area *vma, struct
...
-
 		vaddr = vma->e->start + *off + pfn * PAGE_SIZE;
+		/*
+		 * If should_dump_page() returns true, it means the page is in the dumpees resident memory
+		 * (i.e. bit 63 of the page frame number 'at[pfn]' is set) but it is not the zero-page.
+		 */
+		if (should_dump_page(vma->e, at[pfn])) {
+			if (opts.skip_zero_pages) {
+				remote[0].iov_base = (void*)vaddr;
+				nread = process_vm_readv(item->pid->real, local, 1, remote, 1, 0);
```
avagin, line 214: I don't like the idea to read process memory twice. We have to avoid this.

btw: https://github.com/checkpoint-restore/criu/pull/2292 solves the same problem in a more optimal way.

rst0git: I agree that it would be better if we can avoid reading process memory twice.

> https://github.com/checkpoint-restore/criu/pull/2292 solves the same problem in a more optimal way.

I think this is different. In https://github.com/checkpoint-restore/criu/pull/2292, we exclude pages with zero PFN (`PAGE_IS_PFNZERO`), while this option skips zero-filled memory (e.g., memory that has been filled with zeros using `memset()` would be skipped with this option).

test/javaTests/test-zero.xml
simonis: Good catch! Fixed.

criu/mem.c
simonis: Is this just a performance or a correctness issue? If it's just about performance, I think the benefit might justify the additional overhead and after all the feature is on by default.

In the case this is a correctness problem, do you have a suggestion how we can avoid this?

PS: and yes, @rst0git is right - this change is about skipping regular pages  which are filled with only zero bytes.

simonis:
> > Code Linter test where I can't understand what he's objecting
> 
> @simonis Would it be possible to run `make indent` before committing your changes? This should fix the code style problems such as missing spaces. In addition, it would be great if you can add a commit message with a description that explains _what_ changes are introduced and _why_. The following blog post describes how to write good commit messages: https://cbea.ms/git-commit/

Done (had to install a newer version of `clang-format`).

criu/mem.c
Snorch: > This can be useful for runtimes like Java, which often allocate large memory regions without fully using them (e.g. for the heap). For a simple Helloworld Java program, this new feature shrinks the image size be about 20% from 13mb to 11mb.

> we exclude pages with zero PFN (PAGE_IS_PFNZERO), while this option skips zero-filled memory

I believe we need a zdtm test, which can reproduce such a zero page without PAGE_IS_PFNZERO but with zero data. Maybe, we even want to fix kernel to report such a page as "PAGE_IS_PFNZERO" instead.

I agree with Andrei that manually checking page content with memcmp is an anti-pattern. Nack from me, at list in current state.

adrianreber: I also agree. Reading a page second time and using `memcmp()` sounds not optimal.

I still like the idea of not including empty pages in the checkpoint, but it sounds difficult.

If the kernel could track it, that would be nice. Not sure the kernel has a better alternative than `memcmp()` to find a zeroed memory page.

At this point I think it would be nice to see some numbers. How much faster is restoring if something like this PR is applied. Although I don't like the `memcmp()` it would only be used if the corresponding command-line option is explicitly selected by the user. Maybe that makes it acceptable. Can the second reading of the page be avoided?

Maybe some post-processing of the checkpoint image would be an alternative. Remove the empty pages after checkpointing and have support during restore to handle pages like this.

rst0git: > I believe we need a zdtm test, which can reproduce such a zero page without PAGE_IS_PFNZERO but with zero data.

The code snippet in https://github.com/checkpoint-restore/criu/pull/2331#discussion_r1459073455 can be used to reproduce it:
- `criu dump` creates `pages-1.img` with size 4.1GB
-  when `--skip-zero-pages` is used, `pages-1.img` should be 2.1GB.

> Maybe, we even want to fix kernel to report such a page as "PAGE_IS_PFNZERO" instead.

If understand correctly, page frame number (PFN) corresponds to the physical address of the page. If a process writes zeros to a memory page, the PFN for that page would not be zero.

simonis: > > This can be useful for runtimes like Java, which often allocate large memory regions without fully using them (e.g. for the heap). For a simple Helloworld Java program, this new feature shrinks the image size be about 20% from 13mb to 11mb.
> 
> > we exclude pages with zero PFN (PAGE_IS_PFNZERO), while this option skips zero-filled memory
> 
> I believe we need a zdtm test, which can reproduce such a zero page without PAGE_IS_PFNZERO but with zero data. Maybe, we even want to fix kernel to report such a page as "PAGE_IS_PFNZERO" instead.
> 

The kernel obviously can not do this, at least not until we run with [KSM and `use_zero_pages`](https://docs.kernel.org/admin-guide/mm/ksm.html) which comes with its own set of problems and overhead.

> I agree with Andrei that manually checking page content with memcmp is an anti-pattern. Nack from me, at list in current state.

That's actually exactly what this change is about - spending extra time during snapshot in order to decrease the resulting image size. It is basically KSM for pages containing only zero byte. I don't see how we could do that without reading the content of the page first. And again, as called out by @adrianreber, this is an optional feature which is guarded by its own command line option. So it will not impact anybody who is not using the new `--skip-zero-pages` option.

Snorch: > The code snippet in https://github.com/checkpoint-restore/criu/pull/2331#discussion_r1459073455 can be used to reproduce it:

> The kernel obviously can not do this, at least not until we run with [KSM](https://docs.kernel.org/admin-guide/mm/ksm.html)

Ok, I see that in the mentioned example we have this problem. But that is because the example's code is written sub-optimally, isn't it?

If we slightly change your example like this:
```
diff --git a/malloc-test.c b/malloc-test.c
index b959134..75d13ec 100644
--- a/malloc-test.c
+++ b/malloc-test.c
@@ -6,7 +6,7 @@
 int main() {
        /* Allocate 4GB of memory */
        size_t size = 4ULL * 1024 * 1024 * 1024;
-       char *memory = (char *)malloc(size);
+       char *memory = (char *)calloc(1, size);
 
        if (memory == NULL) {
                fprintf(stderr, "Memory allocation failed\n");
@@ -15,7 +15,6 @@ int main() {
 
        /* Fill the first half with 1 and the second half with 0 */
        memset(memory, 1, size / 2);
-       memset(memory + size / 2, 0, size / 2);
 
        while(1)
                sleep(1);
```
we will have the same result, but avoid zero-filled pages in RSS.

For instance (sorry I don't have free 4Gb on my laptop now so I changed allocation size to 2Gb, but it should not matter), in your example I get this picture in smaps:
```
7f9983374000-7f9a03377000 rw-p 00000000 00:00 0 
Size:            2097164 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:             2097160 kB
```
if we change to calloc it saves us half of the memory:
```
7f565d25e000-7f56dd261000 rw-p 00000000 00:00 0
Size:            2097164 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:             1048584 kB
```

And if I dump the latter I get:
```
du -h /tmp/images-dir/pages-1.img 
1.1G	/tmp/images-dir/pages-1.img
```

So here is next question: Do we really need to cover malloc+memset(0) case?

I'm not sure here, but maybe...

simonis: This is certainly not useful for everyone (and that's why it is a command line option which is off by default), but as I already wrote in the initial PR:

> *This can be useful for runtimes like Java, which often allocate large memory regions without fully using them (e.g. for the heap, code cache, metaspace). For a simple Helloworld Java program, this new feature shrinks the image size be about 20% from 13mb to 11mb.*

I don't think we should mandate specific programming styles in order for applications to work well with CRIU but instead try to optimize them all.

simonis: It's not just the "malloc+memset(0)" use case. Java does [mmap and pretouch memory](https://www.ibm.com/docs/en/semeru-runtime-ce-z/17?topic=options-xx-alwayspretouch) so we can have a lot of pages with only zero content (but not the kernel zero page) in some scenarios.

simonis:
> I believe we need a zdtm test, which can reproduce such a zero page without PAGE_IS_PFNZERO but with zero data.

I've now added a new zdtm test which verifies that the `--skip-zero-pages` option does indeed optimize pages which only contain zero bytes.

criu/mem.c
avagin: > Is this just a performance or a correctness issue? If it's just about performance, I think the benefit might justify the additional overhead and after all the feature is on by default.

I don't understand why we need to read process pages to do this check? Why can't we do that before dumping this pages into the image (page_xfer_dump_pages)?

Snorch: > I believe memset(0) is used in specific scenarios where applications/libraries are dealing with sensitive data and/or use custom memory management.

> It's not just the "malloc+memset(0)" use case. Java does mmap and pretouch memory so we can have a lot of pages with only zero content (but not the kernel zero page) in some scenarios.

1. First,
<details><summary>in current version of code it does not seem to work on my system:</summary>
<p>

I use the example dumpee code from above comments (2G instead of 4G):
```
#include <stdio.h>
#include <unistd.h>
#include <stdlib.h>
#include <string.h>

int main() {
	/* Allocate 2GB of memory */
	size_t size = 2ULL * 1024 * 1024 * 1024;
	char *memory = (char *)malloc(size);

	if (memory == NULL) {
		fprintf(stderr, "Memory allocation failed\n");
		return 1;
	}

	/* Fill the first half with 1 and the second half with 0 */
	memset(memory, 1, size / 2);
	memset(memory + size / 2, 0, size / 2);

	printf("Enter any char to stop\n");
	scanf("%*c");

	free(memory);
	return 0;
}
-------------------------------------------------
[root@turmoil tmp]# gcc -o malloc-test malloc-test.c
[root@turmoil tmp]# ./malloc-test 
Enter any char to stop
```
Here is what smaps show after I run the above:
```
[root@turmoil snorch]# grep 2097164 -A3 -B1 /proc/$(pidof malloc-test)/smaps
7fd2fbee9000-7fd37beec000 rw-p 00000000 00:00 0 
Size:            2097164 kB
KernelPageSize:        4 kB
MMUPageSize:           4 kB
Rss:             2097160 kB
```
Let's dump it:
```
[root@turmoil tmp]# /home/snorch/devel/ms/criu/criu/criu dump -v4 -o dump.log -t $(pidof malloc-test) -j -D /images-dir/
[root@turmoil tmp]# du -h /images-dir/pages-1.img 
2.1G	/images-dir/pages-1.img
```

CRIU is at b23ad2242 ("Don't dump pages which only contain zero bytes").

</p>
</details>

Second, If Java put so much effort to have those zeroed pages in RSS isn't it a bad idea to restore those pages like they are "PAGE_IS_PFNZERO" ones? =)

simonis: > Second, If Java put so much effort to have those zeroed pages in RSS isn't it a bad idea to restore those pages like they are "PAGE_IS_PFNZERO" ones? =)

@Snorch, Java (or OpenJDK based JVMs to be more specific) was not designed and optimized for cloud/container use cases but rather for large, monolithic application servers. In the old days, pretouching/zeroing memory was a way to pre-allocate physical memory and not potentially get it from swap later. For current cloud/container use cases the huge memory footprint can be problem. With check-pointing, a small image is more important and COWing a PAGE_IS_PFNZERO page is much faster then loading and populating it from disk.

simonis: > I don't understand why we need to read process pages to do this check? Why can't we do that before dumping these pages into the image (page_xfer_dump_pages)?

Thanks a lot for your suggestion @avagin. I'm short of time for the next week because of FOSDEM/Jfokus but I'll try to come up with a new version which moves the zero check to `page_xfer_dump_pages()` afterwards.

test/zdtm.py
```diff
@@ -2697,6 +2701,9 @@ def get_cli_args():
     rp.add_argument("--noauto-dedup",
                     help="Manual deduplicate images on iterations",
                     action='store_true')
+    rp.add_argument("--skip-zero-pages",
```
rst0git, line 2704: @simonis Would you be able to also enable testing with `--skip-zero-pages` in [run-ci-tests.sh](https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/ci/run-ci-tests.sh)?

criu/mem.c
avagin: @simonis FOSDEM was my favorite conference when I lived close by. btw @mihalicyn are there too, he is one of criu maintainers. He will be happy to help with any questions.

rppt:
\On Thu, Feb 1, 2024 at 9:47 PM Andrei Vagin ***@***.***> wrote:
>
> @avagin commented on this pull request.
>
> ________________________________
>
> In criu/mem.c:
>
> >   vaddr = vma->e->start + *off + pfn * PAGE_SIZE;
> + /*
> + * If should_dump_page() returns true, it means the page is in the dumpees resident memory
> + * (i.e. bit 63 of the page frame number 'at[pfn]' is set) but it is not the zero-page.
> + */
> + if (should_dump_page(vma->e, at[pfn])) {
> + if (opts.skip_zero_pages) {
> + remote[0].iov_base = (void*)vaddr;
> + nread = process_vm_readv(item->pid->real, local, 1, remote, 1, 0);
>
> @simonis FOSDEM was my favorite conference when I lived close by. btw @mihalicyn are there too, he is one of criu maintainers. He will be happy to help with any questions.

Both @mihalicyn and @rst0git are in the Containers devroom


-- 
Sincerely yours,
Mike.

test/zdtm.py
rst0git: @simonis  It was great meeting you at FOSDEM and your talk was very good!

We can use something like the following patch to run the existing ZDTM tests with `--skip-zero-pages`: https://github.com/rst0git/criu/commit/45a8ca12b5e708e9dcb72e208621ee2405e27a87

rst0git: @avagin Thank you for the review! Would it be sufficient to run the following tests?
```diff
diff --git a/scripts/ci/run-ci-tests.sh b/scripts/ci/run-ci-tests.sh
index ef7e869e0..8f5e25d03 100755
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -268,6 +268,9 @@ make -C test/others/rpc/ run
 ./test/zdtm.py run -t zdtm/transition/maps007 --pre 2 --page-server --dedup
 ./test/zdtm.py run -t zdtm/transition/maps007 --pre 2 --pre-dump-mode read
 
+# Run tests with --skip-zero-pages
+./test/zdtm.py run --skip-zero-pages -T '.*maps0.*'
+
 ./test/zdtm.py run -t zdtm/transition/pid_reuse --pre 2 # start time based pid reuse detection
 ./test/zdtm.py run -t zdtm/transition/pidfd_store_sk --rpc --pre 2 # pidfd based pid reuse detection

```

simonis:
Still on my ToDo list, so post this to avoid auto-closing..

rst0git:
Similar functionality was introduced in https://github.com/checkpoint-restore/criu/pull/2331

PR #2333, opened 2024-01-18
ci: update flake8

rst0git:
This pull request updates the version of flake8 we use to run `make lint` in CI.


rst0git:
We discussed this with Adrian, and it looks like replacing flake8 with [ruff](https://github.com/astral-sh/ruff) would be a better option. I have updated the pull request to fix only the lint error, and I'll open another PR to replace flake8 with ruff.

PR #2336, opened 2024-01-22
zdtm: socket-tcp-nft-nfconntrack: add a hook to the chain in nft case

Snorch:
Let's use hooked nft chain which actually affects packets.

Fixes: e5f4d8c6f ("test/nfconntrack: use nft or iptables-legacy")


rst0git (approved):
LGTM

mihalicyn (approved):

PR #2340, opened 2024-01-30
plugin/amdgpu: Don't print error for "No such process" during resume

fdavid-amd:
During the late stages of restore, each process being resumed gets an ioctl call to KFD_CRIU_OP_RESUME. If the process has no kfd process info, this call with fail with -ESRCH. This is normal behaviour, so we shouldn't print an error message for it.

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -1999,7 +1999,10 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
 	args.op = KFD_CRIU_OP_RESUME;
 	pr_info("Calling IOCTL to start notifiers and queues\n");
 	if (kmtIoctl(fd, AMDKFD_IOC_CRIU_OP, &args) == -1) {
-		pr_perror("restore late ioctl failed");
+		if (errno == ESRCH)
+			pr_info("Pid %d has no kfd process info\n", target_pid);
+		else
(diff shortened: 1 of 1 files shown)


plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1999,7 +1999,10 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
 	args.op = KFD_CRIU_OP_RESUME;
 	pr_info("Calling IOCTL to start notifiers and queues\n");
 	if (kmtIoctl(fd, AMDKFD_IOC_CRIU_OP, &args) == -1) {
-		pr_perror("restore late ioctl failed");
+		if (errno == ESRCH)
+			pr_info("Pid %d has no kfd process info\n", target_pid);
```
Snorch, line 2003: We don't print error here, but we still return -1, I feel that this behavior is kind of inconsistent? Do I miss something?

avagin: criu "ignores" this error code: https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/cr-restore.c#L2487

I think @Snorch is right, we have to return 0 in this case.

Snorch: Ok, then let's do it:

https://github.com/checkpoint-restore/criu/pull/2343

PR #2346, opened 2024-02-09
make: replace flake8 with ruff

rst0git:
This pull request is rebased on https://github.com/checkpoint-restore/criu/pull/2333 and https://github.com/checkpoint-restore/criu/pull/2345.

[Ruff](https://github.com/astral-sh/ruff) is a Python linter written in Rust, designed to replace flake8. It is significantly faster and actively maintained.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,7 @@ Other conventions can be learned from the source code itself. In short, make sur
 
 Important: These tools are there to advise you, but should not be considered as a "source of truth", as tools also make nasty mistakes from time to time which can completely break code readability.
 
-The following command can be used to automatically run a code linter for Python files (flake8), Shell scripts (shellcheck),
+The following command can be used to automatically run a code linter for Python files (ruff), Shell scripts (shellcheck),
 text spelling (codespell), and a number of CRIU-specific checks (usage of print macros and EOL whitespace for C files).
 
 ```
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -41,7 +41,6 @@ RUN apk add \
 	go \
 	e2fsprogs \
 	py-yaml \
-	py3-flake8 \
 	py3-importlib-metadata \
 	asciidoctor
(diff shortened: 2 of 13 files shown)


CONTRIBUTING.md
```diff
@@ -96,7 +96,7 @@ Here are some bad examples of clang-format-ing:
 ```
 @@ -58,8 +59,7 @@ static int register_membarriers(void)
          }
- 
+
```
Snorch, line 99: This " " space at the end of line is intended to be there as it is a part of diff format.

scripts/build/Dockerfile.alpine
```diff
@@ -41,13 +41,12 @@ RUN apk add \
...
 	py-yaml \
-	py3-flake8 \
 	py3-importlib-metadata \
 	asciidoctor
 
 # The rpc test cases are running as user #1000, let's add the user
 RUN adduser -u 1000 -D test
 
-RUN pip3 install junit_xml --break-system-packages
+RUN pip3 install junit_xml ruff --break-system-packages
```
Snorch, line 50: Why do we need ruff here? (Do we call lint somewhere here?)

scripts/ci/prepare-for-fedora-rawhide.sh
```diff
@@ -23,13 +23,13 @@ dnf install -y \
...
 	protobuf-c-devel \
 	protobuf-devel \
-	python3-flake8 \
 	python3-PyYAML \
 	python3-protobuf \
 	python3-junit_xml \
 	python3-pip \
 	python3-importlib-metadata \
 	python-unversioned-command \
+	ruff \
```
Snorch, line 32: Same. Do we need ruff here?

Snorch:
I have no objections to switching flake8 -> ruff as far as it works =) It claims `Drop-in parity with Flake8` so it should not be worse.

rst0git:
@Snorch Thank you for the code review! Indeed, it seems that we only need to install Ruff for the lint workflow. I have updated the pull request.

Issue #335, opened 2017-06-29
Move timers dump/restore code into separate file


xemul:
In `criu/cr-restore.c` file there's a set of functions starting from `prepare_posix_timers()` and `prepare_itimers()`. In `criu/parasite-syscall.c` there's a set starting from `parasite_dump_itimers_seized()` and `parasite_dump_posix_timers_seized()`.

Both are better to be moved to separate .c file in the criu/ directory.

Rajiv2605:
This issue seems unresolved/ open. Can I work on this issue?

PR #2356, opened 2024-02-29
criu: move timers dump/restore code into separate file

ccccrrrr:
Fixes: #335

move timers dump/restore code into separate file


criu/include/timer.h
```diff
@@ -0,0 +1,14 @@
+#ifndef __ITIMER_H__
+#define __ITIMER_H__
+
+#include "types.h"
+#include "parasite.h"
```
avagin, line 5: why do we need to include these two headers here?

criu/include/timer.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __CR_TIMER_H__
+#define __CR_TIMER_H__
+
+#include "restorer.h"
```
avagin, line 4: why do we need to include restorer.h?

criu/include/timer.h
```diff
@@ -0,0 +1,12 @@
...
+
+extern int prepare_itimers(int pid, struct task_restore_args *args, CoreEntry *core);
```
avagin, line 6: We need to include a header that declares CoreEntry.

ccccrrrr: thanks for your tip, my timer.h is not very well written

avagin:
Merged, thanks.

PR #2358, opened 2024-03-03
sk-tcp: Add test cases for TCP socket options and support for TCP_QUICKACK

juntongdeng:
Add test cases for TCP_CORK, TCP_NODELAY, TCP_QUICKACK socket options. 
Add support for restoring TCP_QUICKACK socket option.


test/zdtm/static/sock_tcp_opts00.c
```diff
@@ -0,0 +1,97 @@
...
+		for (j = 0; j < n_opts; j++) {
+			len = sizeof(int);
+			if (getsockopt(sk_confs[i].sk, opts[j].level, opts[j].opt, &val, &len) == -1) {
+				pr_perror("getsockopt(%d, %d) failed", opts[j].level, opts[j].opt);
+				goto close;
+			}
+
+			if (val != opts[j].val) {
+				fail("Unexpected value socket(%d,%d,%d) opts(%d,%d)", sk_confs[i].domain,
+				     sk_confs[i].type, sk_confs[i].protocol, opts[j].level, opts[j].opt);
```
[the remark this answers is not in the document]
juntongdeng: > The test is failing with the following error in CI:
> 
> > FAIL: sock_tcp_opts00.c:84: Unexpected value socket(2,1,6) opts(6,3) (errno = 11 (Resource temporarily unavailable))

This error is because, in the current CRIU implementation, TCP_CORK, TCP_NODELAY, TCP_QUICKACK socket options are dumped only after the TCP socket establish connection. In the test case I added, it did not make the test socket establish connection, causing the error to occur (but strangely this test case worked when I used zdtm, which caused me to not notice the problem).

I think the current CRIU implementation is wrong because the application can set TCP socket options before TCP connection is established, so I think we should move the part of dump_tcp_conn_state() that dump TCP socket options to dump_one_tcp().

What do you think?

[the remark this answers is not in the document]
Snorch: 1) I agree it looks like an error in tcp socket options dumping/restoring code that your test found. Options just disappear on non-connected sockets.

2) I don't see any problem with dumping these tcp socket options separately from connections similar to what we do for ip options here https://github.com/checkpoint-restore/criu/commit/c9a61a205f TcpStreamEntry is just not the right place for those options.

Snorch:
The `f10f8f6a1 ("Merge branch 'checkpoint-restore:criu-dev' into sk-tcp-socket-options")` commit is redundant, please remove it.

images/sk-opts.proto
```diff
@@ -26,13 +26,10 @@ message sk_opts_entry {
 	optional bool		so_reuseport	= 17;
 	optional bool		so_broadcast	= 18;
 	optional bool		so_keepalive	= 19;
-	optional uint32		tcp_keepcnt	= 20;
-	optional uint32		tcp_keepidle	= 21;
-	optional uint32		tcp_keepintvl	= 22;
-	optional uint32		so_oobinline	= 23;
-	optional uint32		so_linger	= 24;
+	optional uint32		so_oobinline	= 20;
```
Snorch, line 29: This is fundamentally wrong. Image format between old and new criu should be backward-(and if possible also forward-)compatible:
1) we can't change id of the existing field
2) we can't reuse id of previously existing field.

Basically you should just mark fields 20-22 as deprecated with a comment and leave them there.

juntongdeng: This PR is obsolete and the latest revision is at https://github.com/checkpoint-restore/criu/pull/2380.

images/tcp-stream.proto
```diff
@@ -16,9 +25,6 @@ message tcp_stream_entry {
 	optional uint32		rcv_wscale	= 8;
 	optional uint32		timestamp	= 9;
 
-	optional bool		cork		= 10;
-	optional bool		nodelay		= 11;
```
Snorch, line 20: Same here, we should leave those lines in the image description, so that new criu can read older images.

criu/sk-tcp.c
```diff
@@ -397,18 +412,6 @@ static int restore_tcp_conn_state(int sk, struct libsoccr_sk *socr, struct inet_
 	if (libsoccr_restore(socr, &data, sizeof(data)))
 		goto err_c;
 
-	if (tse->has_nodelay && tse->nodelay) {
-		aux = 1;
-		if (restore_opt(sk, SOL_TCP, TCP_NODELAY, &aux))
-			goto err_c;
-	}
```
Snorch, line 404: We might also want to consider leaving this for backward compatibility.

criu/sockets.c
```diff
@@ -585,18 +585,6 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 		pr_debug("\tset keepalive for socket\n");
 		ret |= restore_opt(sk, SOL_SOCKET, SO_KEEPALIVE, &val);
 	}
-	if (soe->has_tcp_keepcnt) {
-		pr_debug("\tset keepcnt for socket\n");
-		ret |= restore_opt(sk, SOL_TCP, TCP_KEEPCNT, &soe->tcp_keepcnt);
```
Snorch, line 590: We might also want to consider leaving this for backward compatibility.

Snorch:
> This PR is obsolete and the latest revision is at https://github.com/checkpoint-restore/criu/pull/2380.

Let's close this one then.

Issue #2359, opened 2024-03-04
Cannot restore unix socket of X11: No such file or directory


muvaf:
**Description**

I have a VNC server and a browser running inside a container. Dumping works fine, but when I restore, it fails saying that it cannot find `/tmp/.X11-unix/X0`, however, `Xtigervnc` uses that file as its display socket and I believe the browser communicates through that and all processes are included in the process tree. So, it should be able to restore as both ends of the socket are included.

**Describe the results you received:**

```
Warn  (criu/kerndat.c:1593): CRIU was built without libnftables support
Warn  (criu/kerndat.c:1243): Can't keep kdat cache on non-tempfs
  9296: Error (criu/sk-unix.c:1651): unix: Can't bind id 0x81 ino 75376689 addr /tmp/.X11-unix/X0: No such file or directory
  9296: Error (criu/files.c:1213): Unable to open fd=5 id=0x81
  9289: Error (criu/cr-restore.c:1513): 9296 exited, status=1
Error (criu/cr-restore.c:2557): Restoring FAILED.
```

**Describe the results you expected:**

Restored and continuing to work. I don't expect the TCP connections to continue from where they were since IP of container does change, hence `--tcp-close`

**Additional information you deem important (e.g. issue happens only occasionally):**

I'm using swrk mode with a wrapper to take the checkpoint but CLI to restore. The options boil down to the following:

<details><summary>Checkpoint options:</summary>
<p>

```
		TcpEstablished:  proto.Bool(true),
		ShellJob:        proto.Bool(false),
		FileLocks:       proto.Bool(false),
		LogFile:         proto.String("dump.log"),
		AutoDedup:       proto.Bool(false),
		Pid:             proto.Int32(int32(pid)),
		ImagesDirFd:     proto.Int32(int32(fd)), // To make it use ImagesDir.
		OrphanPtsMaster: proto.Bool(true),
		NotifyScripts:   proto.Bool(true),
		LeaveRunning:    proto.Bool(false),
		LogLevel:        proto.Int32(4),
		LazyPages:       proto.Bool(false),
		GhostLimit:      proto.Uint32(500 * 1048576), // 500MB
		Root:            proto.String("/"),
		TcpClose:        proto.Bool(true),
		ExtMnt: []*rpc.ExtMountMap{
		{
			Key: proto.String("/proc/kcore"),
			Val: proto.String("/dev/null"),
		},
		{
			Key: proto.String("/proc/keys"),
[...]

PR #2360, opened 2024-03-04
Add support for reset-on-fork scheduling flag

rst0git:
This pull request extends CRIU with checkpoint/restore support for the reset-on-fork scheduling flag ([`SCHED_RESET_ON_FORK`](https://lwn.net/Articles/335203/)). When this flag is set, the following rules apply for subsequently created children:

- If the calling thread has a scheduling policy of `SCHED_FIFO` or `SCHED_RR`, the policy is reset to `SCHED_OTHER` in child processes.

- If the calling process has a negative nice value, the nice value is reset to zero in child processes.

(See `man 7 sched`)

Fixes: #2359

### DIFF
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -157,6 +157,11 @@ static int dump_sched_info(int pid, ThreadCoreEntry *tc)
 	tc->has_sched_policy = true;
 	tc->sched_policy = ret;
 
+	/* The reset-on-fork flag might be used in combination
+	 * with SCHED_FIFO or SCHED_RR to reset the scheduling
+	 * policy/priority in child processes.
+	 */
+	ret &= ~SCHED_RESET_ON_FORK;
 	if ((ret == SCHED_RR) || (ret == SCHED_FIFO)) {
 		ret = syscall(__NR_sched_getparam, pid, &sp);
 		if (ret < 0) {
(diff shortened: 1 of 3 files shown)


criu/cr-dump.c
```diff
@@ -157,7 +157,8 @@ static int dump_sched_info(int pid, ThreadCoreEntry *tc)
 	tc->has_sched_policy = true;
 	tc->sched_policy = ret;
 
-	if ((ret == SCHED_RR) || (ret == SCHED_FIFO)) {
```
avagin, line 160: ret  &= ~SCHED_RESET_ON_FORK;

criu/cr-restore.c
```diff
@@ -3060,6 +3060,8 @@ static int validate_sched_parm(struct rst_sched_param *sp)
 	switch (sp->policy) {
```
avagin, line 3060: switch (sp->policy & ~SCHED_RESET_ON_FORK)

test/zdtm/static/sched_policy00.c
```diff
@@ -51,7 +51,7 @@ int main(int argc, char **argv)
 	}
 
 	p.sched_priority = param;
-	if (sched_setscheduler(pid, SCHED_RR, &p)) {
+	if (sched_setscheduler(pid, SCHED_RR | SCHED_RESET_ON_FORK, &p)) {
```
avagin, line 54: Usually, we don't mix test and criu changes in the same commit...

rst0git: Thank you Andrei for the code review! I have updated the pull request.

muvaf:
I built this PR and can confirm [the error](https://github.com/checkpoint-restore/criu/issues/2359#issuecomment-1976384465) I was getting is resolved.

PR #2371, opened 2024-03-24
ci: try to fix broken docker test

adrianreber:
Upgrade to 22.04 base image and skip pinning docker to a certain version.


rst0git (approved):
LGTM

rst0git:
@adrianreber Would it make sense to also upgrade the workflow for Podman to 22.04?

adrianreber:
> @adrianreber Would it make sense to also upgrade the workflow for Podman to 22.04?

This PR tries to fix a broken CI test, so I would not do it in this PR. In another PR, sure, why not.

PR #2376, opened 2024-03-28
zdtm: Distinguish between fail and crash of dump

bsach64:
This PR distinguishes `criu dump` crash from `criu dump` fail in the `zdtm.py` script. Now, `zdtm` reports `criu dump` crash as a test failure.

Very similiar to the work done in this PR (#2140)

But, to detect a rpc crash we see that if the response from the RPC is invalid, then we assume 
the RPC crashed.

For detecting the crash for the CLI we can check the `Popen.returncode`
https://docs.python.org/3/library/subprocess.html#subprocess.Popen.returncode

To simulate the crash, I added a call to `abort()` before `line 2104 criu/cr-dump.c` 

Fixes: #350


test/zdtm.py
```diff
@@ -1032,6 +1039,12 @@ def run(action,
 
         return ret
 
+    @staticmethod
+    def exit_signal(ret):
+        if ret == -2:
+            return True
+        return False
```
rst0git, line 1046: This could be simplified to:
```Python
def exit_signal(ret):
    return ret == -2
```

bsach64: Should've written that in the first place. I will fix it. Thank You :)

If I understand your question correctly,
The run static method returns -1 if criu dump fails.
I changed the run method to return -2 if rpc sends an invalid response (CRIU crashed?).

exit_signal function checks for a crash. So, if ret == -2 then it means CRIU crashed.

test/zdtm.py
```diff
@@ -1233,7 +1244,7 @@ def __criu_act(self, action, opts=[], log=None, nowait=False):
                     return
             rst_succeeded = os.access(
                 os.path.join(__ddir, "restore-succeeded"), os.F_OK)
-            if self.__test.blocking() or (self.__sat and action == 'restore' and
+            if (self.__test.blocking() and not self.__criu.exit_signal(ret)) or (self.__sat and action == 'restore' and
                                           rst_succeeded):
```
Snorch, line 1248: Bad alignment. Let's wrap after `or`

Snorch:
> To simulate the crash, I added a call to abort() before line 2104 criu/cr-dump.c

It looks like we need to add a new fault injection there. See f1714ccce714093170a2616474cfc7b33298c75e as an example. This way we would be able to add a test for it to test/jenkins/criu-fault.sh with `--fault <id>` option.

test/zdtm.py
bsach64: Changed alignment to:
``` 
           if (self.__test.blocking() and not self.__criu.exit_signal(ret)) or \
              (self.__sat and action == 'restore' and rst_succeeded):
```

bsach64:
> > To simulate the crash, I added a call to abort() before line 2104 criu/cr-dump.c
> 
> It looks like we need to add a new fault injection there. See [f1714cc](https://github.com/checkpoint-restore/criu/commit/f1714ccce714093170a2616474cfc7b33298c75e) as an example. This way we would be able to add a test for it to test/jenkins/criu-fault.sh with `--fault <id>` option.

Hello @snorch,
I added a new fault injection for CRIU crashing on dump. Please let me know if this was done correctly. 
On adding a test(s) to `test/jenkins/criu-fault.sh`, any test that specifies the crfail flag can (should?) be added to the `criu-fault.sh`.
There are ~15 tests that specify the `crfail` flag should all these tests be added to criu-fault.sh? Is there a way that we can run all tests that specify a specific flag?

Snorch:
> > > To simulate the crash, I added a call to abort() before line 2104 criu/cr-dump.c
> > 
> > 
> > It looks like we need to add a new fault injection there. See [f1714cc](https://github.com/checkpoint-restore/criu/commit/f1714ccce714093170a2616474cfc7b33298c75e) as an example. This way we would be able to add a test for it to test/jenkins/criu-fault.sh with `--fault <id>` option.
> 
> Hello @Snorch, I added a new fault injection for CRIU crashing on dump. Please let me know if this was done correctly.

Looks correct. But from zdtm.py point of view it may be not so simple, see how __criu_act retries action without fault, we likely don't want our fault to be retry-able by zdtm, so maybe just make it's id >=128 ?

> On adding a test(s) to `test/jenkins/criu-fault.sh`, any test that specifies the crfail flag can (should?) be added to the `criu-fault.sh`. There are ~15 tests that specify the `crfail` flag should all these tests be added to criu-fault.sh? Is there a way that we can run all tests that specify a specific flag?

Yeh, it's a good question. Let's just use one random test with `crfail` (e.g.: vfork00). I don't see easy alternative to this approach, maybe someone else does...

bsach64:
> > > > To simulate the crash, I added a call to abort() before line 2104 criu/cr-dump.c
> > > 
> > > 
> > > It looks like we need to add a new fault injection there. See [f1714cc](https://github.com/checkpoint-restore/criu/commit/f1714ccce714093170a2616474cfc7b33298c75e) as an example. This way we would be able to add a test for it to test/jenkins/criu-fault.sh with `--fault <id>` option.
> > 
> > 
> > Hello @Snorch, I added a new fault injection for CRIU crashing on dump. Please let me know if this was done correctly.
> 
> Looks correct. But from zdtm.py point of view it may be not so simple, see how __criu_act retries action without fault, we likely don't want our fault to be retry-able by zdtm, so maybe just make it's id >=128 ?

I don't think giving the fault injection an id >= 128 is what we want.
If I understand correctly, ids >= 128 refer to non-fatal fault injections (`criu/include/fault-injection.h`). Giving this fatal fault injection an id >= 128 causes tests to fail. I think currently zdtm retries all fatal fault injections. I will try to figure out a different way :)
Can you elaborate on why we don't want our fault to be retry-able?
> > On adding a test(s) to `test/jenkins/criu-fault.sh`, any test that specifies the crfail flag can (should?) be added to the `criu-fault.sh`. There are ~15 tests that specify the `crfail` flag should all these tests be added to criu-fault.sh? Is there a way that we can run all tests that specify a specific flag?
> 
> Yeh, it's a good question. Let's just use one random test with `crfail` (e.g.: vfork00). I don't see easy alternative to this approach, maybe someone else does...

I have added `vfork00` to `criu-fault.sh`.

Snorch:
> > > > > To simulate the crash, I added a call to abort() before line 2104 criu/cr-dump.c
> > > > 
> > > > 
> > > > It looks like we need to add a new fault injection there. See [f1714cc](https://github.com/checkpoint-restore/criu/commit/f1714ccce714093170a2616474cfc7b33298c75e) as an example. This way we would be able to add a test for it to test/jenkins/criu-fault.sh with `--fault <id>` option.
> > > 
> > > 
> > > Hello @Snorch, I added a new fault injection for CRIU crashing on dump. Please let me know if this was done correctly.
> > 
> > 
> > Looks correct. But from zdtm.py point of view it may be not so simple, see how __criu_act retries action without fault, we likely don't want our fault to be retry-able by zdtm, so maybe just make it's id >=128 ?
> 
> I don't think giving the fault injection an id >= 128 is what we want. If I understand correctly, ids >= 128 refer to non-fatal fault injections (`criu/include/fault-injection.h`). Giving this fatal fault injection an id >= 128 causes tests to fail.

I may misunderstand something, correct me if I'm wrong, but isn't a test fail exactly the behavior we expect in case of criu crash? In other words: if criu fails to dump/restore `crfail` test zdtm reports PASS, if criu crashes zdtm should report FAIL, isn't it? (And in criu-fault.sh we should expect FAIL from zdtm.)

> I think currently zdtm retries all fatal fault injections. I will try to figure out a different way :) Can you elaborate on why we don't want our fault to be retry-able?
> 
> > > On adding a test(s) to `test/jenkins/criu-fault.sh`, any test that specifies the crfail flag can (should?) be added to the `criu-fault.sh`. There are ~15 tests that specify the `crfail` flag should all these tests be added to criu-fault.sh? Is there a way that we can run all tests that specify a specific flag?
> > 
> > 
> > Yeh, it's a good question. Let's just use one random test with `crfail` (e.g.: vfork00). I don't see easy alternative to this approach, maybe someone else does...
> 
> I have added `vfork00` to `criu-fault.sh`.

test/jenkins/criu-fault.sh
```diff
@@ -39,3 +39,5 @@ fi
 ./test/zdtm.py run -t zdtm/static/fpu03 --fault 134 -f h --norst || fail
 # also check for the main thread corruption
 ./test/zdtm.py run -t zdtm/static/fpu00 --fault 134 -f h --norst || fail
+
+./test/zdtm.py run -t zdtm/static/vfork00 --fault 7 --report report -f h || fail
```
Snorch, line 43: It should likely instead be `&& fail` checking that zdtm reports an error in case of criu crash.

bsach64: Done!

Snorch (approved):
Looks good.

bsach64:
> > > > > > To simulate the crash, I added a call to abort() before line 2104 criu/cr-dump.c
> > > > > 
> > > > > 
> > > > > It looks like we need to add a new fault injection there. See [f1714cc](https://github.com/checkpoint-restore/criu/commit/f1714ccce714093170a2616474cfc7b33298c75e) as an example. This way we would be able to add a test for it to test/jenkins/criu-fault.sh with `--fault <id>` option.
> > > > 
> > > > 
> > > > Hello @Snorch, I added a new fault injection for CRIU crashing on dump. Please let me know if this was done correctly.
> > > 
> > > 
> > > Looks correct. But from zdtm.py point of view it may be not so simple, see how __criu_act retries action without fault, we likely don't want our fault to be retry-able by zdtm, so maybe just make it's id >=128 ?
> > 
> > 
> > I don't think giving the fault injection an id >= 128 is what we want. If I understand correctly, ids >= 128 refer to non-fatal fault injections (`criu/include/fault-injection.h`). Giving this fatal fault injection an id >= 128 causes tests to fail.
> 
> I may misunderstand something, correct me if I'm wrong, but isn't a test fail exactly the behavior we expect in case of criu crash? In other words: if criu fails to dump/restore `crfail` test zdtm reports PASS, if criu crashes zdtm should report FAIL, isn't it? (And in criu-fault.sh we should expect FAIL from zdtm.)
> 
> > I think currently zdtm retries all fatal fault injections. I will try to figure out a different way :) Can you elaborate on why we don't want our fault to be retry-able?
> > > > On adding a test(s) to `test/jenkins/criu-fault.sh`, any test that specifies the crfail flag can (should?) be added to the `criu-fault.sh`. There are ~15 tests that specify the `crfail` flag should all these tests be added to criu-fault.sh? Is there a way that we can run all tests that specify a specific flag?
> > > 
> > > 
> > > Yeh, it's a good question. Let's just use one random test with `crfail` (e.g.: vfork00). I don't see easy alternative to this approach, maybe someone else does...
> > 
> > 
> > I have added `vfork00` to `criu-fault.sh`.

Sorry, the misunderstanding was on my part. You are correct; the tests previously passed as they were re-run without the fault injection. Thank you for all your help @Snorch :)

Snorch:
Hm, test still does not pass with `&& fail` as though we don't call fail function we still return non-zero exit code and with `set -e` it fails the script.

This seem to do the right trick:
```
if ./test/zdtm.py run -t zdtm/static/vfork00 --fault 136 -f h ; then
        fail
fi
```
Also maybe someone else has better ideas?

bsach64:
> A friendly reminder that this PR had no activity for 30 days.

Can we remove the "stale-pr" label?

adrianreber:
> > A friendly reminder that this PR had no activity for 30 days.
> 
> Can we remove the "stale-pr" label?

Removed.

PR #2380, opened 2024-04-02
sk-tcp: Improve TCP socket options handling

juntongdeng:
This pull request improves the dumping and restoring of TCP socket options in CRIU.

Currently some of the TCP socket option information is stored in the TcpStreamEntry, but the information in the TcpStreamEntry is only restored after the TCP socket has established connection, which results in these TCP socket options not being restored for unconnected TCP sockets.

Move the TCP socket options from TcpStreamEntry to TcpOptsEntry and add dump_tcp_opts() and restore_tcp_opts() for TCP socket options dump and restore.

Currently some TCP socket option information is stored in SkOptsEntry, which is a little confusing. SkOptsEntry should only contain socket options that are common to all sockets.

Move the TCP-specific socket options from SkOptsEntry to TcpOptsEntry.

Currently there are no socket option test cases for TCP_CORK and TCP_NODELAY, this commit adds related test cases.

The socket option test cases for TCP_KEEPCNT, TCP_KEEPIDLE, and TCP_KEEPINTVL already exist in socket-tcp_keepalive.c, so they are not included in this pull request.


images/sk-opts.proto
```diff
@@ -26,13 +26,10 @@ message sk_opts_entry {
...
-	optional uint32		tcp_keepcnt	= 20;
-	optional uint32		tcp_keepidle	= 21;
-	optional uint32		tcp_keepintvl	= 22;
-	optional uint32		so_oobinline	= 23;
-	optional uint32		so_linger	= 24;
+	optional uint32		so_oobinline	= 20;
+	optional uint32		so_linger	= 21;
 
-	optional uint32		so_buf_lock	= 25;
+	optional uint32		so_buf_lock	= 22;
```
rst0git, line 32: @juntongdeng These changes would break backward compatibility.
https://criu.org/Postulates#Protobuf_images_compatibility

juntongdeng: Sorry, I will fix it

juntongdeng: fixed

images/tcp-stream.proto
```diff
@@ -16,9 +21,6 @@ message tcp_stream_entry {
 	optional uint32		rcv_wscale	= 8;
 	optional uint32		timestamp	= 9;
 
-	optional bool		cork		= 10;
```
avagin, line 19: Doesn't it break backward compatiblity? Have you tried to restore from old images?

juntongdeng: Yes, this break backwards compatibility.

But if we don't move the TCP socket options out of the TcpStreamEntry, it is difficult to restore the TCP socket options for unconnected TCP sockets.

How about this? We continue to support restoring TCP socket options from TcpStreamEntry and SkOptsEntry, but dumping TCP socket options to TcpOptsEntry in new versions of CRIU.

avagin: > Yes, this break backward compatibility.

We can't do that.

> How about this? We continue to support restoring TCP socket options from TcpStreamEntry and SkOptsEntry, but dumping TCP socket options to TcpOptsEntry in new versions of CRIU.

It is what we are usually doing in such cases.

juntongdeng: fixed

criu/sk-inet.c
```diff
@@ -581,9 +583,20 @@ static int do_dump_one_inet_fd(int lfd, u32 id, const struct fd_parms *p, int fa
...
+			err = dump_tcp_opts(lfd, &tcpopts);
+			if(err < 0)
+				goto err;
+			
+			err = dump_one_tcp(lfd, sk, &skopts);
+			if(err < 0)
+				goto err;
+		} else {
+			err = 0;
+		}
```
rst0git, line 599: Would it be possible to reverse the order of the if statement?

```C
if (type == SOCK_RAW) {
	err = 0;
} else {
...
}
```

juntongdeng:
I see that lint fails, but I don't seem to find the EOL whitespace problem in sk-inet.c?

criu/sk-inet.c
```diff
@@ -581,9 +583,20 @@ static int do_dump_one_inet_fd(int lfd, u32 id, const struct fd_parms *p, int fa
...
 		if (sk->shutdown)
 			sk_encode_shutdown(&ie, sk->shutdown);
+
+		if (type == SOCK_RAW) {
+			err = 0;
+		} else {
+			err = dump_tcp_opts(lfd, &tcpopts);
+			if(err < 0)
+				goto err;
+
```
adrianreber, line 595: This line probably triggers the lint warning. Looking at the raw patch this line contains three tabs and nothing else.

juntongdeng: fixed

juntongdeng:
It seems that the CRIU Image Streamer Test is stuck in the zdtm/transition/socket_loop00 test case.

But the zdtm/transition/socket_loop00 test case can end normally and pass on my computer.

adrianreber:
Don't worry. The image streamer tests are known to timeout sometimes.

Snorch (approved):

images/sk-opts.proto
```diff
@@ -26,9 +26,9 @@ message sk_opts_entry {
 	optional bool		so_reuseport	= 17;
 	optional bool		so_broadcast	= 18;
 	optional bool		so_keepalive	= 19;
-	optional uint32		tcp_keepcnt	= 20;
-	optional uint32		tcp_keepidle	= 21;
-	optional uint32		tcp_keepintvl	= 22;
+	optional uint32		tcp_keepcnt	= 20; /* obsolete */
+	optional uint32		tcp_keepidle	= 21; /* obsolete */
+	optional uint32		tcp_keepintvl	= 22; /* obsolete */
```
rst0git, line 31: Would you be able to add a comment like "deprecated in favour of tcp_opts_entry message"?

rst0git:
> We continue to support restoring TCP socket options from TcpStreamEntry and SkOptsEntry, but dumping TCP socket options to TcpOptsEntry in new versions of CRIU.

@juntongdeng Could you please add a comment in the code before restoring TcpStreamEntry and SkOptsEntry that this is done for backward comparability, and new versions of CRIU use TcpOptsEntry?

images/sk-opts.proto
juntongdeng: How about I add a line of comment at the top that says "These three TCP options are deprecated, please use the tcp_opts_entry message instead"? and then there is no need to add "obsolete" to each one?

juntongdeng:
> @juntongdeng Could you please add a comment in the code before restoring TcpStreamEntry and SkOptsEntry that this is done for backward comparability, and newer versions of CRIU use TcpOptsEntry?

Yes, will do

images/sk-opts.proto
rst0git: The comment could be something similar to other deprecated flags:
```
$ ag deprecated
core.proto
44:	/* This is deprecated, should be per-thread */
49:	/* These two are deprecated, should be per-thread */

rpc.proto
88:	repeated criu_veth_pair		veths		= 19; /* DEPRECATED, use external instead */
94:	repeated ext_mount_map		ext_mnt		= 23; /* DEPRECATED, use external instead */
108:	repeated unix_sk                unix_sk_ino     = 33; /* DEPRECATED, use external instead */

```

criu/sk-tcp.c
```diff
@@ -449,6 +441,18 @@ int prepare_tcp_socks(struct task_restore_args *ta)
 	return 0;
 }
 
+int restore_tcp_opts(int sk, TcpOptsEntry *toe)
```
avagin, line 432: toe can be null here and we need to handle this case. Do I miss something?

juntongdeng: fixed

juntongdeng:
> > We continue to support restoring TCP socket options from TcpStreamEntry and SkOptsEntry, but dumping TCP socket options to TcpOptsEntry in new versions of CRIU.
> 
> @juntongdeng Could you please add a comment in the code before restoring TcpStreamEntry and SkOptsEntry that this is done for backward comparability, and newer versions of CRIU use TcpOptsEntry?

fixed

images/sk-opts.proto
juntongdeng: fixed

juntongdeng:
How to remove "stale-pr" label?

Snorch (approved):
Looks good.

Test fails seem to be unrelated:

1) `Cirrus CI / Vagrant Fedora Rawhide based test`
https://github.com/checkpoint-restore/criu/pull/2380/checks?check_run_id=25148861695
Error is `setenforce: SELinux is disabled`, seems like we just have selinux disabled in kernel in this env. Does not seem to be related socket opritons or this PR.

2) `Cross Compile Tests / build (true, mips64el-unstable-cross)`
https://github.com/checkpoint-restore/criu/actions/runs/9147441768/job/25160880160?pr=2380
Error is `ERROR: failed to solve: process "/bin/sh -c apt-install ...`, seems like a package install problem. Does not seem to be related to this PR.

3) `Compat Tests / build (GCC)`
https://github.com/checkpoint-restore/criu/actions/runs/9147441773/job/25160879538?pr=2380
Error is `FAIL: vdso01.c:434: Failed to call vdso handlers from symtable after C/R (errno = 11 (Resource temporarily unavailable))`, I see same error in other PR CI. (https://github.com/checkpoint-restore/criu/issues/2390)

rst0git (approved):
LGTM

avagin:
Merged. Thanks a lot.

PR #2388, opened 2024-04-16
sk-tcp: cleanup dump_tcp_conn_state error handling

Snorch:
1) In dump_tcp_conn_state, if return from libsoccr_save is >=0, we check that sizeof(struct libsoccr_sk_data) returned from libsoccr_save is equal to sizeof(struct libsoccr_sk_data) we see in dump_tcp_conn_state (probably to check if we use the right library version). And if sizes are different we go to err_r, which just returns ret, which can teoretically be 0 (if size in library is zero) and that would lead dump_one_tcp treat this as success though it is obvious error.

2) In case of dump_opt or open_image fails we don't explicitly set ret and rely that sizeof(struct libsoccr_sk_data) previously set to ret is not 0, I don't really like it, it makes reading code too complex.

3) We have a lot of err_* labels which do exactly the same thing, there is no point in having all of them, also it is better to choose the name of the label based on what it really does.

So let's refactor error handling to avoid these inconsistencies.

Found this while reviewing https://github.com/checkpoint-restore/criu/pull/2358

### DIFF
--- a/criu/sk-tcp.c
+++ b/criu/sk-tcp.c
@@ -135,6 +135,7 @@ void cpt_unlock_tcp_connections(void)
 static int dump_tcp_conn_state(struct inet_sk_desc *sk)
 {
 	struct libsoccr_sk *socr = sk->priv;
+	int exit_code = -1;
 	int ret, aux;
 	struct cr_img *img;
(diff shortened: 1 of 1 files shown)


rst0git (approved):
LGTM

Issue #2404, opened 2024-05-18
CRIU package for Ubuntu 24.04


yang:
I saw that criu was removed from the Ubuntu 24's apt repos - is there another easy package based way to install on Ubuntu, or is build chain the only way from here on? Thanks.

https://launchpad.net/ubuntu/noble/amd64/criu/3.17.1-3

https://launchpad.net/~criu/+archive/ubuntu/ppa (no noble releases)

mihalicyn:
It's not removed from official Ubuntu repos, we just don't have PPA builds for Noble yet.

mihalicyn:
Related to https://github.com/checkpoint-restore/criu/issues/2398

rst0git:
I will update the OBS/PPA packages with the workaround from https://github.com/checkpoint-restore/criu/issues/2398#issuecomment-2079198028 for the time being and enable support for Ubuntu 24.04.

mihalicyn:
Huge thanks, Radostin!

I'll try to get CRIU into the official repos too. I was able to workaround compile-time failure, but there is another problem related to crit and python (see https://bugs.launchpad.net/ubuntu/+source/criu/+bug/2066148).

rst0git:
>  Please avoid running ``setup.py`` and ``easy_install``.

Should be fixed by https://github.com/checkpoint-restore/criu/pull/2282

In addition, have also applied the following patch in https://github.com/rst0git/criu-deb-packages/commit/80ddb47974f541b99a7afc3e4e226e850648f27d

rst0git:
@yang I can confirm that Ubuntu 24.04 support has been enabled for the packages in OBS and PPA:
- https://build.opensuse.org/package/show/devel:tools:criu/criu
- https://launchpad.net/~criu/+archive/ubuntu/ppa

mihalicyn:
As I can see from https://launchpad.net/~criu/+archive/ubuntu/ppa/+packages we have no builds for Noble, but only for Jammy. I guess we need to change the last entry in `debian/changelog` file and specify `noble` in there.

I'm currently stuggling with this install-stage failure during binary package build:
https://launchpadlibrarian.net/730947573/buildlog_ubuntu-noble-amd64.criu_3.17.1-3ubuntu1_BUILDING.txt.gz

I have applied these patches:
https://github.com/checkpoint-restore/criu/commit/7f0f07599a680258092a7790fb3046856f486f2f
[...]

PR #2414, opened 2024-05-26
criu: use proper format-specified to accommodate time_t 64-bit change

mihalicyn:
See also:
https://wiki.debian.org/ReleaseGoals/64bit-time

Needed for:
https://github.com/checkpoint-restore/criu/issues/2404


mihalicyn:
Looks like we have 3 unrelated test failures here.

PR #2415, opened 2024-05-30
irmap: duplicate string in irmap_scan_path_add

liusdu:
Duplicate string in irmap_scan_path_add, otherwise it will free before parsing next configuration input.

### DIFF
--- a/criu/irmap.c
+++ b/criu/irmap.c
@@ -500,7 +500,7 @@ int irmap_scan_path_add(char *path)
 		return -1;
 	}
 
-	o->ir->path = path;
+	o->ir->path = strdup(path);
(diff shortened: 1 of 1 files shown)


criu/irmap.c
```diff
@@ -500,7 +500,7 @@ int irmap_scan_path_add(char *path)
 		return -1;
 	}
 
-	o->ir->path = path;
+	o->ir->path = strdup(path);
```
avagin, line 503: strdup can return NULL and we need to handle this case.

avagin:
Merged: https://github.com/checkpoint-restore/criu/commit/fac8d640aa040624d9c371ec2be1df32fb57f42e

PR #2416, opened 2024-06-03
criu/plugin: Add NVIDIA CUDA plugin

jesus-ramos:
Adding support for the NVIDIA cuda-checkpoint utility, requires the use of an r555 or higher driver along with the cuda-checkpoint binary.


jesus-ramos:
For the CRIU changes commit would you just like the 2 changes to cr-dump.c and cr-restore.c or should I include the new plugin hooks/calls in that one as well?

adrianreber:
> For the CRIU changes commit would you just like the 2 changes to cr-dump.c and cr-restore.c or should I include the new plugin hooks/calls in that one as well?

One goal is to have each commit still compile. If your changes are not directly dependent on each other moving it into extra commits sounds correct. It really depends and is not always a yes or no decision. But if you already ask the question then it would probably make sense to split it, because that seems to indicate to me you also see it as two independent things.

plugins/cuda/README.md
```diff
@@ -0,0 +1,58 @@
...
+  state and release the process back to it's running state
+
+These actions are facilitated by a CUDA checkpoint+restore thread that the CUDA
+plugin will re-wake when needed.
+
+# Known Limitations
+* Currently GPU memory contents are brought into main system memory and CRIU
+  then checkpoints that as part of the normal procedure. On systems with many
+  GPU's with high GPU memory usage this can cause memory thrashing. A future
+  release will add support for dumping the memory contents to files to alleviate
```
rst0git, line 35: Would it be possible to clarify if this will be a future CRIU release, CUDA driver release or both?

plugins/cuda/README.md
```diff
@@ -0,0 +1,58 @@
...
+  this.
+* There's currently a small race between when a PAUSE_DEVICES hook is called on
+  a running process and a process calls cuInit() and finishes initializing CUDA
+  after the PAUSE is issued but before the process is frozen to checkpoint. This
+  will cause cuda-checkpoint to report that the process is in an illegal state
+  for checkpointing and it's recommended to just attempt the CRIU procedure
+  again, this should be very rare.
```
rst0git, line 42: Would it be possible to handle this case in CRIU? For example, we could print a debug message and retry checkpointing?

plugins/cuda/README.md
```diff
@@ -0,0 +1,58 @@
...
+  accessed again and unlikely to change.
+* CUDA applications that fork() but don't call exec() but also don't issue any
+  CUDA API calls will have some leftover references to /dev/nvidia* and fail to
+  checkpoint as a result. This can be worked around in a similar fashion to the
+  NVML case where the leftover references can be ignored as CUDA is not fork()
+  safe anyway.
+* Restore currently requires that you restore on a system with similar GPU's and
+  same GPU count.
+* NVIDIA UVM Managed Memory, MIG (Multi Instance GPU), and MPS (Multi-Process
+  Service) are currently not supported for checkpointing.
```
rst0git, line 58: Would it be possible to add more information about MIG / MPS? Are they going to be supported in a future release?

Makefile
```diff
@@ -348,6 +352,10 @@ amdgpu_plugin: criu
 	$(Q) $(MAKE) -C plugins/amdgpu all
 .PHONY: amdgpu_plugin
 
+cuda_plugin: criu
```
rst0git, line 355: In contrast to amdgpu plugin, the cuda plugin does not have additional build dependencies.
Would it be possible to add `cuda_plugin` in the list of default targets?
```Makefile
# Default target
all: flog criu lib crit cuda_plugin
```

plugins/cuda/Makefile
```diff
@@ -0,0 +1,41 @@
...
+CC := gcc
+PLUGIN_CFLAGS := -g -Wall -Werror -shared -nostartfiles -fPIC
+
+__nmk_dir ?= ../../scripts/nmk/scripts/
+include $(__nmk_dir)msg.mk
+
+all: $(DEPS_CUDA)
+
+cuda_plugin.so: cuda_plugin.c
+	$(CC) $(PLUGIN_CFLAGS) $(shell $(COMPEL) includes) $^ -o $@ $(PLUGIN_INCLUDE) $(PLUGIN_LDFLAGS)
```
rst0git, line 22:
```Suggestion
	$(call msg-gen, $@)
	$(Q) $(CC) $(PLUGIN_CFLAGS) $(shell $(COMPEL) includes) $^ -o $@ $(PLUGIN_INCLUDE) $(PLUGIN_LDFLAGS)
```

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,380 @@
...
+}
+
+static bool cuda_checkpoint_supports_flag(const char *flag)
+{
+	char msg_buf[2048];
+	const char *args[] = { CUDA_CHECKPOINT, "-h", NULL };
+	int ret = launch_cuda_checkpoint(args, msg_buf, sizeof(msg_buf));
+	if (ret != 0) {
+		pr_err("Failed to launch cuda-checkpoint utility, check that the utility is present in your $PATH\n");
+		return false;
```
rst0git, line 143: Instead of failing with an error during `cuda_plugin_init()` when `cuda-checkpoint` is missing, would it be possible to disable the CUDA plugin functionality?

rst0git:
@jesus-ramos It would be great if we could add tests for this functionality. We currently don't have GPUs available in CI, but we could add tests to verify that the `cuda-checkpoint` tool is called with the correct options during checkpoint/restore.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,380 @@
...
+
+	if (child_pid == 0) { // child
+		if (dup2(fd[WRITE], STDOUT_FILENO) == -1) {
+			return -1;
+		}
+		if (dup2(fd[WRITE], STDERR_FILENO) == -1) {
+			return -1;
+		}
+		close(fd[READ]);
+		return execvp(args[0], (char **)args);
```
rst0git, line 114: When the target process is running in a different namespace (e.g., container), we might need to execute the `cuda-checkpoint` tool within the namespace of the target process.

Would it be possible to use an approach similar to [`cr_system`](https://github.com/checkpoint-restore/criu/blob/b5e2025765b9dfc0b2f583478b7a6380452956b3/criu/util.c#L555) here?

The following is an example of how this functionality is used with tar to checkpoint tmpfs:
https://github.com/checkpoint-restore/criu/blob/b5e2025765b9dfc0b2f583478b7a6380452956b3/criu/filesystems.c#L419

plugins/cuda/README.md
jesus-ramos: Will be a future CUDA release, I'll update that.

plugins/cuda/README.md
jesus-ramos: We could try unfreezing that one process and attempt a lock with timeout and re-freeze it, the other option would be to unfreeze all processes and run through the PAUSE_DEVICES hook again before re-freezing. The first option would at least keep all that contained to the cuda plugin code. I've only been able to repro this by hacking up the driver itself as it is quite tight.

plugins/cuda/cuda_plugin.c
jesus-ramos: I've tested checkpoint/restore of containers running CRIU + cuda-checkpoint outside of the container, at least at the moment the cuda-checkpoint utility doesn't need to run in the container namespace to function.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+	} else { // parent
+		close(fd[WRITE]);
+
+		int bytes_read = read(fd[READ], buf, buf_size);
+		if (bytes_read > 0) {
+			buf[bytes_read - 1] = '\0';
+		}
```
avagin, line 126: nit: should we check that we read all data from the pipe? Otherwise , the process can block on writing into the pipe and never exits.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+
+		int status;
+		waitpid(child_pid, &status, 0);
```
avagin, line 129: nit: it is better to check errors of waitpid, otherwise status can contain some trash from stack.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+		if (!WIFEXITED(status)) {
+			close(fd[READ]);
+			return -1;
```
avagin, line 132: should we print an error here?

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+	if (restore_pid == -1) {
+		pr_info("No need to checkpoint devices on pid %d\n", pid);
+		return 0;
+	}
+
+	pr_info("Checkpointing CUDA devices on pid %d restore_pid %d\n", pid, restore_pid);
+	/* We need to resume the resume the checkpoint thread to prepare the
+	 * mappings for checkpointing
+	 */
+	if (ptrace(PTRACE_CONT, restore_pid, NULL, 0)) {
```
avagin, line 247: we probably need to block signals before resuming a thread. You can look at parasite_run to figure out how it can be done.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+
+	if (strstr(msg_buf, flag) == NULL) {
+		return false;
+	}
+
+	return true;
+}
+
+/* Retrieve the cuda restore thread TID from the root pid */
+static int get_cuda_restore_pid(int root_pid)
```
avagin, line 159: nit: get_cuda_restore_tid probably a better name for this function. Otherwise, it is unclear that it returns a thread ID.

plugins/cuda/cuda_plugin.c
avagin: btw: we can consider adding compel calls to resume and suspend threads. It can be done in a separate pr.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return resume_device(pid, 1);
+}
+CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_devices_late)
+
+int cuda_plugin_init(int stage)
+{
+	if (!cuda_checkpoint_supports_flag("--action")) {
+		pr_err("cuda-checkpoint --action flag not supported, an r555 or higher version driver is required. Disabling CUDA plugin\n");
```
avagin, line 368: It probably should be pr_warn and I think we need to print it only if we have the nvidia device on the host.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	return 0;
+}
+
+int cuda_plugin_checkpoint_devices(int pid)
+{
+	int restore_pid;
+	char msg_buf[CUDA_CKPT_BUF_SIZE];
+	int ret = 0;
```
avagin, line 224: nit: I think it is better to initialize it to -1 and set it to 0 after all actions. In this case, you will not forget to set it to -1 on all error path.  We usually call it `exit_code`, because `ret` is often used for intermediate calls.

avagin:
Thank you for contributing to the project. Overall, this PR looks good to me.

plugins/cuda/cuda_plugin.c
avagin: One more thing is here. If I am not mistaken, seccomp has been suspended to that moment (ptrace_suspend_seccomp). If it is right, we need to resume it before resuming the thread. Otherwise, we can have security concerns here.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,406 @@
...
+	}
+
+	/* compel_wait_task() will fail here as expected but the work we need it to
+	 * do is actually done before the sigstop check that fails. Without this
+	 * compel_stop_on_syscall() will hang waiting for the task to hit a
+	 * sigreturn when cleaning up the parasite
+	 */
+	struct proc_status_creds creds;
+	if (compel_wait_task(restore_pid, -1, parse_pid_status, NULL, &creds.s, NULL)) {
+		pr_debug("wait_task failed as expected. ignoring\n");
```
avagin, line 214: I was trying to figure out why compel_wait_task can fail here, but it always returned COMPEL_TASK_ALIVE (1). It is expected here, isn't it? Does it fail with another code in your environment?

avagin:
Here is a POC patch how we can test the cuda plugin with a mocked cuda-checkpoint tool:
https://github.com/avagin/criu/commit/89df45dd2ca17b019117ef0e1f46b7eeffbbecc7

plugins/cuda/cuda_plugin.c
jesus-ramos: COMPEL_TASK_ALIVE is correct, I think I had just left this from an earlier driver+plugin mix from when I was still changing things up on both sides and CHECKPOINT_DEVICES was called at at a different time/order. I'll clean that up along with the other comments. Thanks for pointing it out.

plugins/cuda/cuda_plugin.c
jesus-ramos: Fixed that up, let me know if it looks okay.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,459 @@
...
+	 * case cuda-checkpoint will report that we're in an invalid state to
+	 * checkpoint
+	 */
+	if (restore_tid == -1) {
+		pr_info("No need to checkpoint devices on pid %d\n", pid);
+		return 0;
+	}
+
+	pr_info("Checkpointing CUDA devices on pid %d restore_tid %d\n", pid, restore_tid);
+	/* We need to resume the resume the checkpoint thread to prepare the
```
rst0git, line 299: _We need to resume the checkpoint thread to ..._

jesus-ramos:
#11 17.04 In file included from ../../compel/include/uapi/compel/infect.h:6,
#11 17.04                  from ../../criu/include/proc_parse.h:6,
#11 17.04                  from cuda_plugin.c:6:
#11 17.04 ../../compel/include/uapi/compel/asm/sigframe.h:5:10: fatal error: asm/elf.h: No such file or directory
#11 17.04     5 | #include <asm/elf.h>
#11 17.04       |          ^~~~~~~~~~~
#11 17.04 compilation terminated.

For the ppc and arm builds do I need to add anything to the plugin Makefile to include the proper directory for this?

rst0git:
> For the ppc and arm builds do I need to add anything to the plugin Makefile to include the proper directory for this?

This error does not seem to be related to the changes in this pull request. It also occurs on the criu-dev branch.

plugins/cuda/cuda_plugin.c
```diff
@@ -0,0 +1,459 @@
...
+		return -1;
+	}
+
+	struct proc_status_creds creds;
+	if (compel_wait_task(restore_tid, -1, parse_pid_status, NULL, &creds.s, NULL) != COMPEL_TASK_ALIVE) {
+		pr_err("compel_wait_task failed after interrupt\n");
+		return -1;
+	}
+
+	if (ptrace(PTRACE_SETOPTIONS, restore_tid, NULL, PTRACE_O_SUSPEND_SECCOMP)) {
```
avagin, line 230: We probably need to set PTRACE_O_TRACESYSGOOD here too. compel expects it to be set:
```
compel/src/lib/ptrace.c:        if (ptrace(PTRACE_SETOPTIONS, pid, NULL, PTRACE_O_SUSPEND_SECCOMP | PTRACE_O_TRACESYSGOOD) < 0) {
```

avagin:
@ jesus-ramos  I think we can merge this pr and do other improvements in separate pr-s. Does it sound good to you?

jesus-ramos:
That works for me.

rst0git (approved):
LGTM

Snorch:
It breaks build in CI:
```
#11 17.14 In file included from ../../compel/include/uapi/compel/asm/sigframe.h:4,
#11 17.14                  from ../../compel/include/uapi/compel/infect.h:6,
#11 17.14                  from ../../criu/include/proc_parse.h:6,
#11 17.14                  from cuda_plugin.c:6:
#11 17.14 /usr/include/x86_64-linux-gnu/asm/sigcontext.h:40:8: error: redefinition of 'struct _fpx_sw_bytes'
#11 17.14    40 | struct _fpx_sw_bytes {
#11 17.14       |        ^~~~~~~~~~~~~
#11 17.14 In file included from /usr/include/signal.h:301,
#11 17.14                  from ../../criu/include/util.h:7,
#11 17.14                  from cuda_plugin.c:3:
```

https://github.com/checkpoint-restore/criu/actions/runs/9754988212/job/26922763707?pr=2427

jesus-ramos:
Is it an issue with the include order or something else? I can see about getting some time on an aarch64 machine to check it out.

Snorch:
> Is it an issue with the include order or something else? I can see about getting some time on an aarch64 machine to check it out.

Likely, in ./plugins/cuda/cuda_plugin.c:
```
#include "criu-log.h"
#include "plugin.h"
#include "util.h"
#include "cr_options.h"
#include "pid.h"
#include "proc_parse.h"

#include <common/list.h>
#include <compel/infect.h>

#include <ctype.h>
#include <fcntl.h>
#include <stdio.h>
#include <unistd.h>
#include <sys/ptrace.h>
#include <sys/wait.h>
```

criu headers should probably go AFTER system headers, but I'm not fully sure how it causes the problem.

rst0git:
> It breaks build in CI:

Andrei opened a pull request to fix this problem: https://github.com/checkpoint-restore/criu/pull/2428

PR #2420, opened 2024-06-09
ci: upgrade to Fedora 40 Vagrant images (38 is EOL)

adrianreber:


rst0git (approved):

adrianreber:
Without this PR all vargant based tests seems to fail really early because of missing vagrant images. This updates it to the latest Fedora release and it seems to work.

Issue #2037, opened 2023-01-04
CRIU crashes process and fails to checkpoint when process leverages Syscall User Dispatch


gmprice:
Syscall User Dispatch is a new feature for translating foreign system calls into linux system calls via a new interposition mechanism.

https://docs.kernel.org/admin-guide/syscall-user-dispatch.html

it does this on entry to the kernel, somewhat similar to SECCOMP

```
bool syscall_user_dispatch(struct pt_regs *regs)
{
... snip ...
  sd->on_dispatch = true;
  syscall_rollback(current, regs);
  trigger_sigsys(regs);
... snip ...
```

Presently, CRIU creates two issues in the compel and parasite code that causes a crash and failure to checkpoint.

1. The compel code forcibly blocks SIGSYS from being delivered.  Upon any future syscall (outside the Syscall User Dispatch exception area) will cause the SIGSYS to be delivered to CRIU, and CRIU will fail (usually looking for SIGTRAP).  CRIU either attempts to bail out, or it delivers the signal and waits for a TRAP.  The SIGSYS is then delivered to the process, but since SIGSYS is now masked, it crashes and everything fails.
2. In a test, I unblocked SIGSYS manually just to see what would happen.  CRIU progresses further, but upon the first compelled syscall, Syscall User Dispatch hooks that system call and attempts to deliver a SIGSYS.  At this point CRIU fails and bails out.

I believe the "correct" way to do deal with all of this is to try to get something akin to the SUSPEND_SECCOMP option merged into ptrace code

```
int ptrace_suspend_seccomp(pid_t pid)
{
  if (ptrace(PTRACE_SETOPTIONS, pid, NULL, PTRACE_O_SUSPEND_SECCOMP | PTRACE_O_TRACESYSGOOD) < 0) {
    pr_perror("suspending seccomp failed");
    return -1;
  }

  return 0;
}
```

A similar SUSPEND_SYSCALL_USER_DISPATCH option would allow all of this to happen cleanly.

mihalicyn:
@gmprice thanks for reporting this!

Do you want to work on this feature?

gmprice:
@mihalicyn I just pushed an RFC patch up to linux-kernel requesting SUSPEND_SYSCALL_USER_DISPATCH

https://lkml.org/lkml/2023/1/4/1044

[...]

PR #2423, opened 2024-06-24
[RFC] Dump/restore syscall user dispatch via ptrace

svetly-todorov:
Following up on issue #2037.

Use the PTRACE_GET/SET_SYSCALL_USER_DISPATCH flags upstreamed in kernel 6.4 to preserve SUD settings for a target process.

The majority of the implementation mirrors that of the seccomp dump/restore. The most significant difference is that, unlike seccomp, there's no flag for disabling SUD while a target is being ptraced. Therefore restoring it in the parasite blob is dangerous, because remote syscalls can generate a SIGSYS. Instead, restoration happens right before the final PTRACE_DETACH, and not in the parasite code.


images/core.proto
```diff
@@ -106,10 +106,13 @@ message thread_core_entry {
 	optional seccomp_mode		seccomp_mode	= 11;
 	optional uint32			seccomp_filter	= 12;
 
-	optional string			comm		= 13;
-	optional uint64			blk_sigset_extended	= 14;
-	optional rseq_entry		rseq_entry	= 15;
-	optional uint32			cg_set		= 16;
```
rst0git, line 112: Changing the field numbers here would break backward compatibility: https://criu.org/Postulates

rst0git:
@svetly-todorov Thank you for the pull request! Would you be able to add tests for this feature?

compel/src/lib/ptrace.c
```diff
@@ -21,6 +23,68 @@
...
+{
+	char task[32];
+	int tid;
+	sud_config_t disable;
+	struct dirent *de;
+	DIR *dp;
+
+	/* Setup SUD-disable struct */
+	memset(&disable, 0, sizeof(disable));
+	disable.mode = PR_SYS_DISPATCH_OFF;
```
rst0git, line 54:
```
#10 2.352 compel/src/lib/ptrace.c:54:17: error: ‘PR_SYS_DISPATCH_OFF’ undeclared (first use in this function)
#10 2.352    54 |  disable.mode = PR_SYS_DISPATCH_OFF;
#10 2.352       |                 ^~~~~~~~~~~~~~~~~~~
```

criu/include/sud.h
```diff
@@ -0,0 +1,38 @@
...
+#define SYS_DISPATCH_OFF 0
+#endif
+
+#ifndef SYS_DISPATCH_ON
+#define SYS_DISPATCH_ON 1
+#endif
+
+struct sys_dispatch_entry {
+	struct rb_node node;
+    struct sys_dispatch_entry *next;
```
rst0git, line 17: It looks like this file contains mixed indentation.

criu/sud.c
```diff
@@ -0,0 +1,269 @@
...
+	return entry;
+}
+
+/* This must run after ptrace freeze but *before* parasite code. */
+int sud_collect_entry(pid_t tid_real)
+{
+    struct sys_dispatch_entry *entry;
+    sud_config_t config;
+
+	entry = sud_lookup(tid_real, true, false);
```
rst0git, line 74: The indentation is also mixed here.

images/core.proto
svetly-todorov: Got it. I'll move the sud_mode and _setting to the bottom of the struct

criu/include/sud.h
svetly-todorov: Surprised I missed this. Will address.

compel/src/lib/ptrace.c
svetly-todorov: This is probably a case of kernel headers not always being defined. I'll go back and explicitly add the #ifdefs to sud.h.

svetly-todorov:
@rst0git 

Working on a rebase that adds some bugfixes and addresses the whitespace issues.

We (@gmprice and I) did write a pair of tests for this. Where should they be included?

rst0git:
> We (@gmprice and I) did write a pair of tests for this. Where should they be included?

Ideally, the tests should be in a separate commit and it would be great if it is possible to add them in [ZDTM](https://criu.org/ZDTM_test_suite). For example, the following commits have been used to add ZDTM tests: https://github.com/checkpoint-restore/criu/commit/96a30f068c3c8bd19c2377666a367bd030b83dde, https://github.com/checkpoint-restore/criu/commit/d4902182b0bf9b69c11e4159110bba6a65a846ba, https://github.com/checkpoint-restore/criu/commit/516b3698a7613b9d2d02b1d62a32fd1f015a32e6.

criu/sud.c
```diff
@@ -0,0 +1,92 @@
...
+#include "util.h"
+#include "rst-malloc.h"
+
+#include "protobuf.h"
+#include "images/sud.pb-c.h"
+
+#undef LOG_PREFIX
+#define LOG_PREFIX "sys-dispatch: "
+
+static struct rb_root sud_tid_rb_root = RB_ROOT;
```
avagin, line 25: pls add more info in the commit message to describe what is going on here.

avagin:
> The majority of the implementation mirrors that of the seccomp dump/restore. The most significant difference is that, unlike seccomp, there's no flag for disabling SUD while a target is being ptraced. Therefore restoring it in the parasite blob is dangerous, because remote syscalls can generate a SIGSYS. Instead, restoration happens right before the final PTRACE_DETACH, and not in the parasite code.

CRIU injects a parasite blob while a process is being dumped. How do you handle where to place a parasite blob?

svetly-todorov:
> > The majority of the implementation mirrors that of the seccomp dump/restore. The most significant difference is that, unlike seccomp, there's no flag for disabling SUD while a target is being ptraced. Therefore restoring it in the parasite blob is dangerous, because remote syscalls can generate a SIGSYS. Instead, restoration happens right before the final PTRACE_DETACH, and not in the parasite code.
> 
> CRIU injects a parasite blob while a process is being dumped. How do you handle where to place a parasite blob?

We don't worry about the parasite blob during dump because we collect the SUD settings before it is injected, and then disable SUD during the compel_wait_task that precedes the infection. It's been a while since I wrote this code, but if I'm remembering correctly, collection of SUD entries happens in this callback: https://github.com/checkpoint-restore/criu/pull/2423/files#diff-b5c20127a71a84d42a655e83714ec5358f2f3343f6c0359580ee2cbb1ea1e940R972

This callback triggers [here](https://github.com/checkpoint-restore/criu/pull/2423/files#diff-2649cc2dccddbcc4810ad5317ee58795dbf3ca2f390d2102a5587963a2d32f29R269) in 

The callback is necessary for SUD, but not necessary for seccomp, because [ptrace can disable seccomp without overwriting its settings](https://github.com/checkpoint-restore/criu/blob/criu-dev/compel/src/lib/ptrace.c#L26). So for seccomp, you can SUSPEND_SECCOMP during compel_wait_task and then read its settings later. But for SUD, you can only read its settings while it is active; ptrace only lets you enable/disable it without a graceful suspend.

svetly-todorov:
I have pushed up a kind of v2 for this RFC containing some fixes and the beginnings of a test case. The test does not work quite yet, but I have to drop work on this for now, because of other pressing obligations. Forgive me!

PR #2427, opened 2024-07-02
cgroupd: unblock SIGTERM to make stop_cgroupd actually work

Snorch:
Sometimes due to sigblockmask inheritance cgroupd can inherit SIGTERM blocked. That will lead cgroupd ignoring SIGTERM from stop_cgroupd() and CRIU will get stuck due to waiting for never-stopping cgroupd.

I see this happening in lxc-checkpoint, also saw this in OpenVZ jenkins on cgroup_inotify00 test.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1947,6 +1947,21 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
 	return 0;
 }
 
+static int cgroupd_unblock_sigterm(void)
+{
+	sigset_t unblockmask;
+
+	sigemptyset(&unblockmask);
+	sigaddset(&unblockmask, SIGTERM);
+
+	if (sigprocmask(SIG_UNBLOCK, &unblockmask, NULL)) {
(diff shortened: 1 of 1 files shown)


criu/cgroup.c
```diff
@@ -1955,6 +1969,9 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
  */
 static int cgroupd(int sk)
 {
+	if (cgroupd_unblock_sigterm())
```
avagin, line 1972: pls write a comment to explain why we need to unblock sigterm here

criu/cgroup.c
```diff
@@ -1947,6 +1947,20 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
 	return 0;
 }
```
avagin, line 1949: static int cgroupd_unblock_sigterm(void)
{

avagin:
LGTM

criu/cgroup.c
Snorch: Thanks for pointing this out, fixed.

PR #2428, opened 2024-07-02
plugins/cuda: fix cross compilation

avagin:


jesus-ramos (approved):

plugins/cuda/Makefile
```diff
@@ -10,7 +10,7 @@ PLUGIN_INCLUDE  	+= -iquote../../
 
 COMPEL := ../../compel/compel-host
 
-CC := gcc
+CC := $(CROSS_COMPILE)$(HOSTCC)
```
rst0git, line 13: We can remove this line. `CC` is defined in `tools.mk`:
https://github.com/checkpoint-restore/criu/blob/a11e944996d4a71ed275cc12bd6eefa795015100/scripts/nmk/scripts/tools.mk#L12

rst0git (approved):
LGTM

0x7f454c46:
makes sense to me as well :-)

PR #2429, opened 2024-07-03
apparmor: get_suspend_policy must return NULL in error cases

avagin:
Before this fix, it could return MAP_FAILED which is ((void *) -1).


rst0git (approved):
LGTM

PR #2435, opened 2024-07-07
cuda: fix launch cuda-checkpoint

rst0git:
When the `cuda-checkpoint` tool is not installed, `execvp()` is expected to fail and return `-1`. In this case, we need to call `exit()` to terminate the child process that was created earlier with `fork()`.

This fix also resolves the errors currently occurring with Podman in CI.


jesus-ramos (approved):

scripts/ci/podman-test.sh
```diff
@@ -20,6 +20,7 @@ sed -i 's/#runtime\s*=\s*.*/runtime = "runc"/' /usr/share/containers/containers.
 # Test checkpoint/restore with action script
 echo "action-script /usr/bin/true" | sudo tee /etc/criu/default.conf
 
+cat /proc/mounts
```
avagin, line 23: Usually, I use /proc/self/mountinfo. It contains more info.

plugins/cuda/cuda_plugin.c
```diff
@@ -116,7 +116,8 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
 			return -1;
 		}
 		close(fd[READ]);
-		return execvp(args[0], (char **)args);
+		execvp(args[0], (char **)args);
```
avagin, line 119: I think we need to print  error/warning here.

scripts/ci/podman-test.sh
rst0git: Thanks! I've updated the patch to use `/proc/self/mountinfo`.

plugins/cuda/cuda_plugin.c
rst0git: In the child process STDOUT and STDERR are set to a pipe fd. However, we handle the exit code of the process and show appropriate messages. I've updated the log messages so that when cuda-checkpoint is not installed, CRIU will show the following debug message:
```
cuda_plugin: cuda-checkpoint not present in $PATH
```

avagin: logfd isn't STDOUT or STRERR, is it?

rst0git: yes, but for some reason it appears to be closed after `execvp()`.
Should we use `close_fds()` here to avoid leaking file descriptors (e.g., https://github.com/checkpoint-restore/criu/commit/6c8ea60491305482eb60d9b17a8635ab9248a5ab)?

rst0git: I've updated the pull request to call `close_fds()` before `execvp()` and to show the following messages when `cuda-checkpoint` is not installed:
```
(00.008034) cuda_plugin: execvp("cuda-checkpoint") failed: No such file or directory
(00.008055) cuda_plugin: cuda-checkpoint not present in $PATH
```

plugins/cuda/cuda_plugin.c
```diff
@@ -419,7 +432,15 @@ CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_
 
 int cuda_plugin_init(int stage)
 {
-	if (!cuda_checkpoint_supports_flag("--action")) {
+	int ret = cuda_checkpoint_supports_flag("--action");
+
+	if (ret == -1) {
+		pr_debug("%s not present in $PATH\n", CUDA_CHECKPOINT);
```
avagin, line 438: pr_warning

avagin:
LGTM

rst0git:
Rebased on the criu-dev branch to resolve merge conflict with https://github.com/checkpoint-restore/criu/commit/dcb577b0dcfa8d6ef4bc744d2343fecb18f31d8b.

test/zdtm.py
```diff
@@ -895,7 +902,9 @@ def run(action,
             timeout=60):
         env = dict(
             os.environ,
-            ASAN_OPTIONS="log_path=asan.log:disable_coredump=0:detect_leaks=0")
+            ASAN_OPTIONS="log_path=asan.log:disable_coredump=0:detect_leaks=0",
+            CRIU_LIBS_DIR=PLUGINS_DIR
```
avagin, line 906: I don't think it is a good idea to run all tests with these plugins by default... I think it should be optional.

rst0git: Thank you for the code review Andrei!

I have updated the pull request to make running ZDTM tests with plugins optional. However, I noticed that, for some reason, when running the tests only with the `amdgpu_plugin`, we still see messages from the `cuda_plugin`:

```
sudo ./zdtm.py run -t zdtm/static/busyloop00 --criu-plugin amdgpu -f h
userns is supported
=== Run 1/1 ================ zdtm/static/busyloop00
======================= Run zdtm/static/busyloop00 in h ========================
Start test
./busyloop00 --pidfile=busyloop00.pid --outfile=busyloop00.out
Run criu dump
=[log]=> dump/zdtm/static/busyloop00/56/1/dump.log
b'(00.008948) cg:     `- [net_cls] -> [/] [0]'
b'(00.008950) cg: Set 1 is criu one'
b'(00.008963) Detected cgroup V1 freezer'
b"(00.008966) plugin: `cuda_plugin' hook 10 -> 0x7f0204737ede"
b'(00.052279) cuda_plugin: Error initializing CUDA: "no CUDA-capable device is detected"'
b'(00.052309) Error (cuda_plugin.c:201): cuda_plugin: Failed to launch cuda-checkpoint to retrieve restore tid: Error initializing CUDA: "no CUDA-capable device is detected"'
b'(00.052408) Collected (4 attempts, 0 in_progress)'
b'(00.052424) Collected (4 attempts, 0 in_progress)'
b'(00.052432) Collected 56 in 1 state'
b"(00.052434) plugin: `cuda_plugin' hook 11 -> 0x7f0204737cbf"
b'(00.072347) cuda_plugin: Error initializing CUDA: "no CUDA-capable device is detected"'
b'(00.072381) Error (cuda_plugin.c:201): cuda_plugin: Failed to launch cuda-checkpoint to retrieve restore tid: Error initializing CUDA: "no CUDA-capable device is detected"'
```

rst0git: I've updated this patch to add `--criu-plugin` option that allows to run tests with plugins.

plugins/cuda/cuda_plugin.c
```diff
@@ -118,7 +118,12 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
 
 		close_fds(STDERR_FILENO + 1);
 
-		return execvp(args[0], (char **)args);
```
avagin, line 119: two other child "returns" have to be replaced with exit too

plugins/cuda/cuda_plugin.c
```diff
@@ -151,27 +156,35 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
 			return -1;
 		}
 
+		if (WEXITSTATUS(status) != EXIT_SUCCESS)
```
avagin, line 159: It can be killed by signal and we probably want to see its output in this case too. Actually, I think we can always print it to the log.

plugins/cuda/cuda_plugin.c
rst0git: Thank you for pointing this out. I have updated the patch to use `_exit()` instead of `return`.

plugins/cuda/cuda_plugin.c
rst0git: I have updated the patch to also print the output when the child process is killed by signal.

This function is also used to get the output from `cuda-checkpoint --help`. To prevent cluttering the logs, we skip printing this output if `exit_code == EXIT_SUCCESS`.
```
$ cuda-checkpoint --help
CUDA checkpoint and restore utility.
Version 555.42.06. Copyright (C) 2024 NVIDIA Corporation. All rights reserved.

Operations:
--get-state --pid <pid>
        Prints the current checkpoint state of the process specified by <pid>

--action lock | checkpoint | restore | unlock --pid <pid> [--timeout <ms>]
        Performs the specified action on <pid>.
        For the lock action a timeout can be provided, the lock operation will wait up to <ms> milliseconds for the operation to succeed.

--toggle --pid <pid>
        Toggles the CUDA state in the specified process between the running and checkpointed states

--get-restore-tid --pid <pid>
        Retrieves the CUDA restore thread ID of the process specified by <pid>

Options:
--pid|-p <pid>
        The pid upon which to perform the operation

--timeout|-t <timeout>
        Optional timeout that can be specified for the lock action in milliseconds

--help|-h
        Print this help message
```

plugins/cuda/cuda_plugin.c
```diff
@@ -110,15 +110,20 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
 
 	if (child_pid == 0) { // child
 		if (dup2(fd[WRITE], STDOUT_FILENO) == -1) {
-			return -1;
+			_exit(EXIT_FAILURE);
 		}
 		if (dup2(fd[WRITE], STDERR_FILENO) == -1) {
-			return -1;
+			_exit(EXIT_FAILURE);
```
avagin, line 116: need to print an error message

plugins/cuda/cuda_plugin.c
```diff
@@ -145,33 +150,41 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
 		}
 
 		int status;
-		if (waitpid(child_pid, &status, 0) == -1 || !WIFEXITED(status)) {
-			pr_err("cuda-checkpoint exited improperly, couldn't complete operation\n");
-			close(fd[READ]);
-			return -1;
-		}
+		int exit_code = waitpid(child_pid, &status, 0);
+		if (exit_code == -1 || !WIFEXITED(status))
```
avagin, line 154: 1. we use the `exit_code` for the very specific pattern, so pls choose another name for this variable. 
2. you print error here, but you don't change exit_code, so this function return 0 that doesn't look tight.

plugins/cuda/cuda_plugin.c
```diff
@@ -145,33 +150,41 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
...
+			pr_err("cuda-checkpoint exited improperly\n");
```
avagin, line 155: If it has been killed, we need to print a signal.

plugins/cuda/cuda_plugin.c
rst0git: Thank you Andrei! I have updated the pull request with these changes.

plugins/cuda/cuda_plugin.c
```diff
@@ -144,34 +151,47 @@ static int launch_cuda_checkpoint(const char **args, char *buf, int buf_size)
 			bytes_read = read(fd[READ], scratch, sizeof(scratch));
 		}
 
-		int status;
+		int status, exit_code = -1;
 		if (waitpid(child_pid, &status, 0) == -1 || !WIFEXITED(status)) {
-			pr_err("cuda-checkpoint exited improperly, couldn't complete operation\n");
-			close(fd[READ]);
-			return -1;
+			if (WIFSIGNALED(status)) {
```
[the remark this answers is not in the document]
rst0git: Thank you Andrei! I have applied the changes from https://github.com/checkpoint-restore/criu/commit/710e1802ba62f0237573c9aa172d3aeb595979b9 and updated the pull request.

PR #2438, opened 2024-07-10
plugins: set executable bit on .so files

rst0git:
For historical reasons, some tools like [rpm](https://docs.fedoraproject.org/en-US/package-maintainers/CommonRpmlintIssues/#unstripped_binary_or_object) or [ldd](https://sourceware.org/git/?p=glibc.git;a=blob;f=elf/ldd.bash.in;h=d6b640df;hb=HEAD#l154) may expect the executable bit to be present for the correct identification of shared libraries. The executable bit on .so files is set by default by compilers (e.g., GCC). It is not strictly necessary but primarily a convention.

```
$ sudo ldd /usr/lib/criu/*.so
/usr/lib/criu/amdgpu_plugin.so:
ldd: warning: you do not have execution permission for `/usr/lib/criu/amdgpu_plugin.so'
	linux-vdso.so.1 (0x00007fd0a2a3e000)
	libdrm.so.2 => /lib64/libdrm.so.2 (0x00007fd0a29eb000)
	libdrm_amdgpu.so.1 => /lib64/libdrm_amdgpu.so.1 (0x00007fd0a29de000)
	libc.so.6 => /lib64/libc.so.6 (0x00007fd0a27fc000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd0a2a40000)
/usr/lib/criu/cuda_plugin.so:
ldd: warning: you do not have execution permission for `/usr/lib/criu/cuda_plugin.so'
	linux-vdso.so.1 (0x00007f1806e13000)
	libc.so.6 => /lib64/libc.so.6 (0x00007f1806c08000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f1806e15000)
```

### DIFF
--- a/plugins/amdgpu/Makefile
+++ b/plugins/amdgpu/Makefile
@@ -54,7 +54,7 @@ install:
 ifeq ($(CONFIG_AMDGPU),y)
 	$(Q) mkdir -p $(DESTDIR)$(PLUGINDIR)
 	$(E) "  INSTALL " $(PLUGIN_NAME)
-	$(Q) install -m 644 $(PLUGIN_SOBJ) $(DESTDIR)$(PLUGINDIR)
+	$(Q) install -m 755 $(PLUGIN_SOBJ) $(DESTDIR)$(PLUGINDIR)
 endif
 .PHONY: install
 
--- a/plugins/cuda/Makefile
+++ b/plugins/cuda/Makefile
@@ -31,11 +31,10 @@ mrproper: clean
 install:
 	$(Q) mkdir -p $(DESTDIR)$(PLUGINDIR)
 	$(E) "  INSTALL " $(PLUGIN_NAME)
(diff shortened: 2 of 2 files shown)


avagin:
I am not sure that it is a common practice. I listed shared libraries on my laptop and none of them except libc has executable permissions. Do we solve any other problems except the ldd warning here?

rst0git:
> Do we solve any other problems except the ldd warning here?

Building an RPM package fails with the following when the installed *.so files are not executable.
```
W: unstripped-binary-or-object /usr/lib64/criu/amdgpu_plugin.so
```
The change in this pull request is similar to how we install `libcompel.so`:
https://github.com/checkpoint-restore/criu/blob/1b3ba301f9874eb20addf65304283f717ef8d515/compel/Makefile#L61

PR #2440, opened 2024-07-10
Adjust to glibc __rseq_size semantic change

fweimer-rh:
We plan to backport this glibc change all the way to glibc 2.35. Most distributions use the stable release branches, so they will get this change eventually. I hope this is still fine from a CRIU perspective because the non-`ptrace` fallback path is not used.

### DIFF
--- a/test/zdtm/static/rseq00.c
+++ b/test/zdtm/static/rseq00.c
@@ -46,12 +46,15 @@ static inline void *__criu_thread_pointer(void)
 static inline void unregister_glibc_rseq(void)
 {
 	struct rseq *rseq = (struct rseq *)((char *)__criu_thread_pointer() + __rseq_offset);
+	unsigned int size = __rseq_size;
 
 	/* hack: mark glibc rseq structure as failed to register */
 	rseq->cpu_id = RSEQ_CPU_ID_REGISTRATION_FAILED;
 
 	/* unregister rseq */
-	syscall(__NR_rseq, (void *)rseq, __rseq_size, 1, RSEQ_SIG);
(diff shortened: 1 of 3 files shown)


test/zdtm/static/rseq00.c
```diff
@@ -51,7 +51,10 @@ static inline void unregister_glibc_rseq(void)
 	rseq->cpu_id = RSEQ_CPU_ID_REGISTRATION_FAILED;
 
 	/* unregister rseq */
-	syscall(__NR_rseq, (void *)rseq, __rseq_size, 1, RSEQ_SIG);
+	unsigned int size = __rseq_size;
```
rst0git, line 54: Would you be able to move the declaration for `unsigned int size;` at the beginning of the `unregister_glibc_rseq()` function?
```
.701 rseq00.c:54:9: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
8.701    54 |         unsigned int size = __rseq_size;
8.701       |         ^~~~~~~~
```

mihalicyn:
Hi Florian,

huge thanks for notifying us about this glibc change and your fix!

Can you please split your change to zdtm test and cr-restore change into a separate commits, something like:
- `cr-restore: adjust to glibc __rseq_size semantic change`
- `zdtm: rseq: adjust to glibc __rseq_size semantic change`

But anyways, LGTM.

mihalicyn (approved):

avagin:
@fweimer-rh I decided that it's not critical. It is the one logical change that fixes the same issue in criu and tests. Thank you for contributing and taking care about CRIU.

Issue #2441, opened 2024-07-10
Massive testsuite failures when testing from a subdirectory of /tmp on Fedora (starting with zdtm/static/aio00)


fweimer-rh:
Fedora 39 mounts `/tmp` like this:

```
tmpfs on /tmp type tmpfs (rw,nosuid,nodev,seclabel,nr_inodes=1048576,inode64)
```

This seems to cause all tests to fail if the CRIU Git checkout is located in `/tmp`. It's a bit non-obvious, so maybe the testsuite could check the mount flags of the mount point? At the very least, this ticket should be easier to find due to the `zdtm/static/aio00` string …

avagin:
Our tests creates a test root file system to run tests in a container-like environment:
https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm.py#L251

We probably can do something like this:
https://github.com/avagin/criu/commit/0e477f336fcc6b857cfb24334986dec28ebad7c1

@adrianreber could you try to run these tests with my patch?

avagin:
@adrianreber I just found that my change is incomplete. I will update it and push the new version tomorrow.

avagin:
My origin idea doesn't work. I think the simplest way to fix the issue is to do something like this:
```diff --git a/tests/run-zdtm.sh b/tests/run-zdtm.sh
index 56fc8ec..45b1d3e 100755
--- a/tests/run-zdtm.sh
+++ b/tests/run-zdtm.sh
@@ -24,7 +24,7 @@ EXCLUDES=" \
        -x zdtm/static/cgroup02 "
 
 run_test() {
-       ./zdtm.py run --criu-bin /usr/sbin/criu ${EXCLUDES} \
+       ./test/zdtm.py run --criu-bin /usr/sbin/criu ${EXCLUDES} \
                -a --ignore-taint --keep-going
 
        RESULT=$?
@@ -42,11 +42,14 @@ rm -f /var/lib/sss/pipes/nss
 
 cd source
 
+tar -czf test.tar.gz -C test .
+mount -t tmpfs criu-test test
+trap "umount -l test" EXIT
+tar -xzf test.tar.gz -C test
+
 echo "Build CRIU"
 make
 
-cd test
-
 echo "Run the actual CRIU test suite"
 run_test
```

rst0git:
> maybe the testsuite could check the mount flags of the mount point?

It sounds like a good idea.

> I think the simplest way to fix the issue is to do something like this

[...]

PR #2444, opened 2024-07-15
zdtm: verify that criu repo file system allows resolving device files

Snorch:
We put several device files into zdtm container through its root file system which is the same file system criu git resides in.

We do it like this to overcome inability to create character and block device files in user namespaces. So even if we put those device files into zdtm container in any other way, in 'uns' flavor CRIU would not be able to restore those device files.

Other option can be - creating auxiliary tmpfs mount and add it into zdtm container as external mount, but that looks like an overkill for this problem. Let's print a clear error so that user can either mount current file system without 'nodev" or put criu somewhere else.

Fixes: #2441

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -283,6 +283,14 @@ class ns_flavor:
                     self.__construct_root()
                     os.mknod(self.root + "/.constructed", stat.S_IFREG | 0o600)
 
+        try:
+            with open(self.root + "/dev/null"):
+                pass
+        except IOError as e:
+            if e.errno != errno.EACCES:
+                raise e
+            raise test_fail_exc("The filesystem of CRIU repo should not have 'nodev' flag. Consider putting CRIU on appropriate file system to run ZDTM.")
+
         for b in l_bins:
             self.__copy_libs(b)
         for b in x_bins:


test/zdtm.py
```diff
@@ -283,6 +283,14 @@ def init(self, l_bins, x_bins):
                     self.__construct_root()
                     os.mknod(self.root + "/.constructed", stat.S_IFREG | 0o600)
 
+        try:
+            with open(self.root + "/dev/null"):
+                pass
+        except IOError as e:
+            if e.errno != errno.EACCES:
+                raise e
+            raise test_fail_exc("The filesystem of CRIU repo should not have 'nodev' flag. Consider putting CRIU on appropriate file system to run ZDTM.")
```
rst0git, line 292: It might be good to add `\n`:
```Suggestion
            raise test_fail_exc("\nThe filesystem of CRIU repo should not have 'nodev' flag. Consider putting CRIU on appropriate file system to run ZDTM.")
```

```
[root@thinkpad:/tmp/criu/test] <base> 1 # python3 ./zdtm.py run -t zdtm/static/busyloop00 -f ns
userns is supported
=== Run 1/1 ================ zdtm/static/busyloop00
======================= Run zdtm/static/busyloop00 in ns =======================
 Test zdtm/static/busyloop00 FAIL at 
The filesystem of CRIU repo should not have 'nodev' flag. Consider putting CRIU on appropriate file system to run ZDTM. 
```

Snorch:
"External mount" approach is more robust, so closing it.

PR #2451, opened 2024-07-24
plugin: enable multiple plugins for the same hook

rst0git:
CRIU provides two plugins for checkpoint/restore of GPU applications: amdgpu and cuda. Both plugins use the `RESUME_DEVICES_LATE` hook to enable restore:

    CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, amdgpu_plugin_resume_devices_late)
    CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_devices_late)

However, CRIU currently [does not support](https://github.com/checkpoint-restore/criu/blob/93746eb25014f21287f2edc5f5dd7545ce19ab52/criu/include/plugin.h#L42) running more than one plugin for the same hook. As a result, when both plugins are installed, the resume function for CUDA applications is not executed. To fix this, we need to make sure that both `plugin_resume_devices_late()` functions return `-ENOTSUP` when restore is not supported.

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -1809,7 +1809,7 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
 	fd = open(AMDGPU_KFD_DEVICE, O_RDWR | O_CLOEXEC);
 	if (fd < 0) {
 		pr_perror("failed to open kfd in plugin");
-		return -1;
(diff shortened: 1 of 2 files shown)


jesus-ramos (approved):
LGTM

PR #2461, opened 2024-08-06
support user set remote mmap vma address

duguhaotian:
1. os auto assignment vma addr maybe conflict with vma in gpu living migrate scene;
2. so, we should give choice to user;
3. default, user_set_remote_map is 0L, so we donot change default behave;

user set vma usecase:
```
// we use compel/test/infect/spy.c to show example

static int do_infection(int pid)
{
       ... ....
	ictx = compel_infect_ctx(ctl);
	ictx->log_fd = STDERR_FILENO;

       // TODO: use should set assignment vma addr at here
      ictx->user_set_remote_map = xxxxL;

	parasite_setup_c_header(ctl);
       .... .....
}
```

### DIFF
--- a/compel/include/uapi/infect.h
+++ b/compel/include/uapi/infect.h
@@ -120,6 +120,7 @@ struct infect_ctx {
 	open_proc_fn open_proc;
 
 	int log_fd; /* fd for parasite code to send messages to */
+	unsigned long remote_map_addr; /* User-specified address where to mmap parasitic code, default not set */
 };
 
 extern struct infect_ctx *compel_infect_ctx(struct parasite_ctl *);
--- a/compel/src/lib/infect.c
+++ b/compel/src/lib/infect.c
@@ -816,7 +816,7 @@ static int parasite_memfd_exchange(struct parasite_ctl *ctl, unsigned long size,
 	uint8_t orig_code[MEMFD_FNAME_SZ] = MEMFD_FNAME;
 	pid_t pid = ctl->rpid;
 	long sret = -ENOSYS;
-	int ret, fd, lfd;
+	int ret, fd, lfd, remote_flags;
 
 	if (ctl->ictx.flags & INFECT_NO_MEMFD)
 		return 1;
@@ -860,7 +860,11 @@ static int parasite_memfd_exchange(struct parasite_ctl *ctl, unsigned long size,
 		goto err_cure;
 	}
 
-	ctl->remote_map = remote_mmap(ctl, NULL, size, remote_prot, MAP_FILE | MAP_SHARED, fd, 0);
+	remote_flags = MAP_FILE | MAP_SHARED;
+	if (ctl->ictx.remote_map_addr){
+		remote_flags |= MAP_FIXED_NOREPLACE;
+	}
+	ctl->remote_map = remote_mmap(ctl, (void *)ctl->ictx.remote_map_addr, size, remote_prot, remote_flags, fd, 0);
 	if (!ctl->remote_map) {
 		pr_err("Can't rmap memfd for parasite blob\n");
 		goto err_curef;


rst0git:
> os auto assignment vma addr maybe conflict with vma in gpu living migrate scene;

There are several different methods for checkpoint/restore of GPU state (e.g., [cuda-checkpoint](https://github.com/NVIDIA/cuda-checkpoint), [amdgpu plugin](https://github.com/checkpoint-restore/criu/blob/criu-dev/plugins/amdgpu/README.md), [cricket](https://github.com/RWTH-ACS/cricket), [cedana](https://github.com/cedana/cedana), a [device-proxy](https://arxiv.org/abs/2202.07848) mechanism from Microsoft, and a very similar [solution](https://www.youtube.com/watch?v=YNjN8S9P8Ic) from Tencent).

@duguhaotian Would you be able to provide more information about this use-case and why this change is necessary?

duguhaotian:
> > os auto assignment vma addr maybe conflict with vma in gpu living migrate scene;
> 
> There are several different methods for checkpoint/restore of GPU state (e.g., [cuda-checkpoint](https://github.com/NVIDIA/cuda-checkpoint), [amdgpu plugin](https://github.com/checkpoint-restore/criu/blob/criu-dev/plugins/amdgpu/README.md), [cricket](https://github.com/RWTH-ACS/cricket), [cedana](https://github.com/cedana/cedana), a [device-proxy](https://arxiv.org/abs/2202.07848) mechanism from Microsoft, and a very similar [solution](https://www.youtube.com/watch?v=YNjN8S9P8Ic) from Tencent).
> 
> @duguhaotian Would you be able to provide more information about this use-case and why this change is necessary?

 similar solution from Microsoft, log and reply. 
1. But, i encountered a log: mmap vma address and do somethings.
2. So, i should reply this mmap vma and do somethings.
3. And then, probabilistic failure if criu vma address same to log vma address;

duguhaotian:
@rst0git

compel/src/lib/infect.c
```diff
@@ -860,7 +860,7 @@ static int parasite_memfd_exchange(struct parasite_ctl *ctl, unsigned long size,
 		goto err_cure;
 	}
 
-	ctl->remote_map = remote_mmap(ctl, NULL, size, remote_prot, MAP_FILE | MAP_SHARED, fd, 0);
+	ctl->remote_map = remote_mmap(ctl, (void *)ctl->ictx.user_set_remote_map, size, remote_prot, MAP_FILE | MAP_SHARED, fd, 0);
```
avagin, line 863: If ctl->ictx.user_set_remote_map isn't zero, you probably need to consider setting MAP_FIXED_NOREPLACE.

duguhaotian: yes, thks. fixed it

duguhaotian: > If ctl->ictx.user_set_remote_map isn't zero, you probably need to consider setting MAP_FIXED_NOREPLACE.

yes, thks. fixed it

duguhaotian:
@avagin @rst0git please check this pr, thks.

compel/include/uapi/infect.h
```diff
@@ -120,6 +120,7 @@ struct infect_ctx {
 	open_proc_fn open_proc;
 
 	int log_fd; /* fd for parasite code to send messages to */
+	unsigned long user_set_remote_map;
```
avagin, line 123: nit: I think it is better to rename it to remote_map_addr and add a comment to explain how it can be used.

compel/src/lib/infect.c
```diff
@@ -860,7 +860,11 @@ static int parasite_memfd_exchange(struct parasite_ctl *ctl, unsigned long size,
 		goto err_cure;
 	}
 
-	ctl->remote_map = remote_mmap(ctl, NULL, size, remote_prot, MAP_FILE | MAP_SHARED, fd, 0);
+	remote_flags = MAP_FILE | MAP_SHARED;
+	if (ctl->ictx.user_set_remote_map != 0L){
```
avagin, line 864: nit: if (ctl->ictx.user_set_remote_map)

avagin:
LGTM pls resolve all items and we will merge it.

btw: how and where do you use this?

compel/include/uapi/infect.h
duguhaotian: fixed

compel/src/lib/infect.c
duguhaotian: fixed

duguhaotian:
> LGTM pls resolve all items and we will merge it.
> 
> btw: how and where do you use this?

1. pr comments show how to use it;
2. log/reply to restore state of xpu, We encountered a mmap address conflict, causing the parasitic code execution to get stuck

PR #2469, opened 2024-08-16
cuda: unlock on timeout error

rst0git:
When attempting to checkpoint a container with CUDA processes, CRIU could fail with the following error:
```
Error (criu/cr-dump.c:1791): Timeout reached. Try to interrupt: 1
Error (cuda_plugin.c:143): cuda_plugin: Unable to read output of cuda-checkpoint: Interrupted system call
Error (cuda_plugin.c:384): cuda_plugin: PAUSE_DEVICES failed with
``` 
In this situation, the target process is locked, but CRIU fails due to a timeout and exits with an error. We need to make sure that the target PID is unlocked in such case.

### DIFF
--- a/plugins/cuda/cuda_plugin.c
+++ b/plugins/cuda/cuda_plugin.c
@@ -4,6 +4,7 @@
 #include "cr_options.h"
 #include "pid.h"
 #include "proc_parse.h"
+#include "seize.h"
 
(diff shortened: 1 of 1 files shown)


jesus-ramos (approved):
LGTM. Thanks for all the fixes.

PR #2470, opened 2024-08-17
Enable CUDA checkpointing with multiple processes

rst0git:
When checkpointing a container with multiple CUDA processes, CRIU currently fails with the following error:

```
cuda_plugin: pausing devices on pid 802521
Error (criu/cr-dump.c:1791): Timeout reached. Try to interrupt: 0
Error (cuda_plugin.c:141): cuda_plugin: Unable to read output of cuda-checkpoint: Interrupted system call
Error (cuda_plugin.c:382): cuda_plugin: PAUSE_DEVICES failed with
```

This error occurs because CRIU uses the cgroup specified by the container runtime to freeze all processes running in the container. However, if there are multiple processes that have CUDA state, we need to "lock" all these processes before freezing the cgroup; otherwise the `cuda-checkpoint` tool hangs.

To address this problem, this pull request updates the `collect_pstree` function to run the `PAUSE_DEVICES` hook for all processes in the container cgroup prior to freezing.

This error can be replicated with the following example:
```
echo -e "tcp-established\nghost-limit=100M" | sudo tee -a /etc/criu/runc.conf

sudo podman run -d --privileged --name=hlp --device nvidia.com/gpu=all --security-opt=label=disable \
        nvcr.io/nvidia/hpc-benchmarks:24.06 \
        mpirun --bind-to none -np 2 \
        ./hpl.sh --no-multinode --dat /workspace/hpl-linux-x86_64/sample-dat/HPL-1GPU.dat

sudo podman container checkpoint -l -e /tmp/test.tar
```


criu/seize.c
```diff
@@ -983,7 +1032,17 @@ int collect_pstree(void)
...
 
-	ret = run_plugins(PAUSE_DEVICES, pid);
+	/*
+	 * To create a checkpoint of a container, we use the cgroup specified by the
+	 * container runtime (e.g., runc) to freeze all processes in the container.
+	 * However, if there are processes with CUDA state, we need to "lock" them
+	 * before freezing; otherwise, the cuda-checkpoint tool may hang.
+	 */
+	if (opts.freeze_cgroup)
+		ret = pause_devices_in_cgroup();
```
avagin, line 1042: Here is a race window between pause_devices_in_cgroup and freeze_processes.

criu/seize.c
```diff
@@ -967,6 +968,54 @@ static int cgroup_version(void)
...
+		char *ptr, *end_ptr;
+		/* Null-terminate the buffer */
+		buffer[bytes_read] = '\0';
+		ptr = buffer;
+
+		/* Process each PID */
+		while (*ptr) {
+			pid = strtol(ptr, &end_ptr, 10);
+			if (ptr != end_ptr) {
+				run_plugins(PAUSE_DEVICES, pid);
```
avagin, line 998: check errors

criu/seize.c
```diff
@@ -967,6 +968,54 @@ static int cgroup_version(void)
...
+				ptr = end_ptr;
+			} else {
+				/* Move to the next line if the current line was invalid */
```
avagin, line 1001: report a error if the line is "invalid"

criu/seize.c
```diff
@@ -967,6 +968,54 @@ static int cgroup_version(void)
...
+	/* Open the cgroup.procs file */
+	snprintf(procs_path, sizeof(procs_path), "%s/%s", opts.freeze_cgroup, "cgroup.procs");
+	procs_fd = open(procs_path, O_RDONLY);
+	if (procs_fd == -1) {
+		pr_perror("Failed to open cgroup.procs file");
+		return -1;
+	}
+
+	/* Read cgroup.procs into a buffer */
+	while ((bytes_read = read(procs_fd, buffer, sizeof(buffer) - 1)) > 0) {
```
avagin, line 988: you must understand the last line in the buffer can be incomplete and you need to handle it properly

avagin:
Any update?

rst0git:
Closing in favour of https://github.com/checkpoint-restore/criu/pull/2475

PR #2474, opened 2024-09-04
[WIP] keep consistent of pid in all pid namespace

duguhaotian:
Currently, containerd uses criu for container live migration, but there is an inconsistency issue with the root pid namespace of the container process. This PR aims to fix this problem.

current logs：
```
[root@localhost CRIU]# ctr run -d --net-host openeuler:22.03 test bash
[root@localhost CRIU]# cat /sys/fs/cgroup/cpu/default/test/tasks
67635
[root@localhost CRIU]# ctr t exec -t --exec-id=1 test ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   3624  2604 ?        Ss   17:22   0:00 bash
root           7  0.0  0.0   5900  2452 pts/0    Rs+  17:23   0:00 ps aux
[root@localhost CRIU]# ctr c checkpoint --rw --task test checkpoint/test
[root@localhost CRIU]# ctr t del -f test
WARN[0000] task test exit with non-zero exit code 137
[root@localhost CRIU]# ctr c del test
[root@localhost CRIU]# ctr c restore --rw --live test checkpoint/test
[root@localhost CRIU]# cat /sys/fs/cgroup/cpu/default/test/tasks
67850
[root@localhost CRIU]# ctr t exec -t --exec-id=1 test ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   3624   244 ?        Ss   17:24   0:00 bash
root          22  0.0  0.0   5900  2348 pts/0    Rs+  17:24   0:00 ps aux
```

fixed log
```
[root@localhost CRIU]# ctr run -d --net-host openeuler:22.03 test bash
[root@localhost CRIU]# cat /sys/fs/cgroup/cpu/default/test/tasks
71699
[root@localhost CRIU]# ctr t exec -t --exec-id=1 test ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   3624  2616 ?        Ss   17:25   0:00 bash
root           7  0.0  0.0   5900  2344 pts/0    Rs+  17:25   0:00 ps aux
[root@localhost CRIU]# ctr c checkpoint --rw --task test checkpoint/test
[root@localhost CRIU]# ctr t del -f test
WARN[0000] task test exit with non-zero exit code 137
[root@localhost CRIU]# ctr c del test
[root@localhost CRIU]# ctr c restore --rw --live test checkpoint/test
[root@localhost CRIU]#  cat /sys/fs/cgroup/cpu/default/test/tasks
71699
[root@localhost CRIU]#  ctr t exec -t --exec-id=1 test ps aux
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0   3624   244 ?        Ss   17:26   0:00 bash
root          22  0.0  0.0   5900  2416 pts/0    Rs+  17:26   0:00 ps aux
```

### DIFF
--- a/criu/clone-noasan.c
+++ b/criu/clone-noasan.c
@@ -46,9 +46,10 @@ int clone_noasan(int (*fn)(void *), int flags, void *arg)
 	return clone(fn, stack_ptr, flags, arg);
 }
 
-int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags, int exit_signal, pid_t pid)
+int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags, int exit_signal, pid_t *ns_tids, size_t ns_tids_len)
 {
 	struct _clone_args c_args = {};
+	pid_t pid;
 
 	BUG_ON(flags & CLONE_VM);
 
@@ -75,8 +76,8 @@ int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags, int exit_sig
 		c_args.exit_signal = exit_signal;
 	}
 	c_args.flags = flags;
-	c_args.set_tid = ptr_to_u64(&pid);
-	c_args.set_tid_size = 1;
+	c_args.set_tid = ptr_to_u64(ns_tids);
+	c_args.set_tid_size = ns_tids_len;
 	pid = syscall(__NR_clone3, &c_args, sizeof(c_args));
 	if (pid == 0)
 		exit(fn(arg));
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -1203,9 +1203,15 @@ static inline int fork_with_pid(struct pstree_item *item)
 	}
 
 	if (kdat.has_clone3_set_tid) {
+		pid_t ns_tids[2] = {pid};
+		size_t ns_tids_len = 1;
+		if (item->pid->real > 0 && item->pid->real != pid) {
+			ns_tids[ns_tids_len] = item->pid->real;
+			ns_tids_len++;
+		}
 		ret = clone3_with_pid_noasan(restore_task_with_children, &ca,
 					     (ca.clone_flags & ~(CLONE_NEWNET | CLONE_NEWCGROUP | CLONE_NEWTIME)),
-					     SIGCHLD, pid);
+					     SIGCHLD, ns_tids, ns_tids_len);
 	} else {
 		/*
 		 * Some kernel modules, such as network packet generator
--- a/criu/include/clone-noasan.h
+++ b/criu/include/clone-noasan.h
@@ -2,6 +2,6 @@
 #define __CR_CLONE_NOASAN_H__
 
 int clone_noasan(int (*fn)(void *), int flags, void *arg);
-int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags, int exit_signal, pid_t pid);
+int clone3_with_pid_noasan(int (*fn)(void *), void *arg, int flags, int exit_signal, pid_t *ns_tids, size_t ns_tids_len);
 
 #endif /* __CR_CLONE_NOASAN_H__ */
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -313,6 +313,7 @@ int dump_pstree(struct pstree_item *root_item)
 		e.pgid = item->pgid;
 		e.sid = item->sid;
 		e.n_threads = item->nr_threads;
+		e.realpid = item->pid->real;
 
 		e.threads = xmalloc(sizeof(e.threads[0]) * e.n_threads);
 		if (!e.threads)
@@ -577,6 +578,7 @@ static int read_one_pstree_item(struct cr_img *img, pid_t *pid_max)
 	if (e->sid > *pid_max)
 		*pid_max = e->sid;
 	pi->pid->state = TASK_ALIVE;
+	pi->pid->real = e->realpid;
 
 	if (e->ppid == 0) {
 		if (root_item) {
--- a/images/pstree.proto
+++ b/images/pstree.proto
@@ -8,4 +8,5 @@ message pstree_entry {
 	required uint32			pgid		= 3;
 	required uint32			sid		= 4;
 	repeated uint32			threads		= 5;
+	required uint32			realpid		= 6;
 }


adrianreber:
This is actually by design. We do not want to have the same PID outside of the container to avoid PID collisions when restoring the container multiple times.

If we actually want this feature it needs to be configurable. Not sure if we want this feature or not.

duguhaotian:
> This is actually by design. We do not want to have the same PID outside of the container to avoid PID collisions when restoring the container multiple times.
> 
> If we actually want this feature it needs to be configurable. Not sure if we want this feature or not.

1. yes, it needs to be configurable.
2. The scenario I've encountered involves the XPU driver using the hostpid as an identifier and then saving these states. This leads to issues when restoring.

adrianreber:
What is a XPU driver?

duguhaotian:
> What is a XPU driver?

like gpu

duguhaotian:
@adrianreber So, do you think we need to support this feature? If so, I will add configurable interfaces.

adrianreber:
Can the XPU driver be changed? This is the first time that we have this requirement.

duguhaotian:
> This is actually by design. We do not want to have the same PID outside of the container to avoid PID collisions when restoring the container multiple times.
> 
one question: which scenario we should restoring the container multiple times ?

duguhaotian:
> Can the XPU driver be changed? This is the first time that we have this requirement.

It involves a huge amount of work, so it probably won't be modified.

adrianreber:
> > This is actually by design. We do not want to have the same PID outside of the container to avoid PID collisions when restoring the container multiple times.
> 
> one question: which scenario we should restoring the container multiple times ?

For me that is a very common scenario. Fast startup of a pre-initialized container.

> > Can the XPU driver be changed? This is the first time that we have this requirement.
> 
> It involves a huge amount of work, so it probably won't be modified. 😂

Can't the driver offer an interface where the PID can be updated? At this point it seems still pretty unlikely that CRIU will default to restoring the container with the same host PID outside of the container. Being independent of the host PID is a very important feature as PID collisions are one of the main problems users see when using CRIU. What happens if another unrelated process takes one PID of the container. This makes the life for the user much harder. I am still very sceptical.

As far as I can tell neither the nvidia nor the AMD GPU support rely on the host PID, so maybe the XPU driver also needs to think about the migration scenario and how it can be supported even if it is a lot of work. Which vendor are we talking about. Can you be more specific about what you mean by XPU driver and how it supports checkpoint/restore.

duguhaotian:
> > > This is actually by design. We do not want to have the same PID outside of the container to avoid PID collisions when restoring the container multiple times.
> > 
> > 
> > one question: which scenario we should restoring the container multiple times ?
> 
> For me that is a very common scenario. Fast startup of a pre-initialized container.
> 
> > > Can the XPU driver be changed? This is the first time that we have this requirement.
> > 
> > 
> > It involves a huge amount of work, so it probably won't be modified. 😂
> 
> Can't the driver offer an interface where the PID can be updated? At this point it seems still pretty unlikely that CRIU will default to restoring the container with the same host PID outside of the container. Being independent of the host PID is a very important feature as PID collisions are one of the main problems users see when using CRIU. What happens if another unrelated process takes one PID of the container. This makes the life for the user much harder. I am still very sceptical.
> 
> As far as I can tell neither the nvidia nor the AMD GPU support rely on the host PID, so maybe the XPU driver also needs to think about the migration scenario and how it can be supported even if it is a lot of work. Which vendor are we talking about. Can you be more specific about what you mean by XPU driver and how it supports checkpoint/restore.

1. I got, thks. Since the XPU driver has not been open-sourced yet, I won't post the relevant code here. The logic involves managing many resource states within the driver using the hostpid as a key, and these states need to be saved and restored.
2. If left unchanged by default, would the community accept this feature being controlled by a parameter?

avagin:
> If left unchanged by default, would the community accept this feature being controlled by a parameter?

If we understand that you have a sustained solution that really require this feature, we will accept it.

In Linux,  PIDs are assigned sequentially, starting from a reserved range and going up to a configurable maximum (/proc/sys/kernel/pid_max). Once the maximum PID is reached, the allocation wraps around to the beginning of the range, skipping any PIDs that are still in use.

It means that the pid conflict problem exists almost in all scenarios. It isn't only about the case when you need to restore multiple instances from one snapshot. 

@duguhaotian what is your target scenario? Is it live-migration? When you migrate a workload to a remote host, PID-s of migrated processes can be already used by other processes on the remote host. How are you going to handle this case?

duguhaotian:
> > If left unchanged by default, would the community accept this feature being controlled by a parameter?
> 
> If we understand that you have a sustained solution that really require this feature, we will accept it.
> 
> In Linux, PIDs are assigned sequentially, starting from a reserved range and going up to a configurable maximum (/proc/sys/kernel/pid_max). Once the maximum PID is reached, the allocation wraps around to the beginning of the range, skipping any PIDs that are still in use.
> 
> It means that the pid conflict problem exists almost in all scenarios. It isn't only about the case when you need to restore multiple instances from one snapshot.
> 
> @duguhaotian what is your target scenario? Is it live-migration? When you migrate a workload to a remote host, PID-s of migrated processes can be already used by other processes on the remote host. How are you going to handle this case?

1. yes,  live-migration;
2. if conflict with remote host, we have two choice: 1.) just return failed; 2.) Downgrade to not guarantee consistent host PIDs;

duguhaotian:
> btw: If you start a container without a pid namespace, criu will preserve process pid-s.

Thks. This is an evasion tactic, but we cannot restrict the behavior of users.

adrianreber:
As the XPU driver seems to be also under your (organization) control I think it would make more sense to have the support for live migration in your driver.

> > btw: If you start a container without a pid namespace, criu will preserve process pid-s.
> 
> Thks. This is an evasion tactic, but we cannot restrict the behavior of users.

This doesn't sound different than what you are proposing. We say the user should use `--pid=host` and you say the user should use `--same-pid-on-host`. Both approaches require user interaction to adapt to your driver. Fixing the driver to handle the live migration scenario, which seems important to you, make more sense from my point of view. Especially as your XPU driver is so far the only one requiring this change.

duguhaotian:
> As the XPU driver seems to be also under your (organization) control I think it would make more sense to have the support for live migration in your driver.
> 
> > > btw: If you start a container without a pid namespace, criu will preserve process pid-s.
> > 
> > 
> > Thks. This is an evasion tactic, but we cannot restrict the behavior of users.
> 
> This doesn't sound different than what you are proposing. We say the user should use `--pid=host` and you say the user should use `--same-pid-on-host`. Both approaches require user interaction to adapt to your driver. Fixing the driver to handle the live migration scenario, which seems important to you, make more sense from my point of view. Especially as your XPU driver is so far the only one requiring this change.

> `--pid=host`  is set to container run command, but `--same-pid-on-host` set to criu dump/restore. this is different.
First, should container user to set. Second, can automatically handled by the PaaS platform.

But, i agree with you. The best way is to change XPU driver.

duguhaotian:
close this pr

PR #2475, opened 2024-09-12
criu: Allow disabling freeze cgroups

avagin:
Some plugins (e.g., CUDA) may not function correctly when processes are frozen using cgroups. This change introduces a mechanism to disable the use of freeze cgroups during process seizing, even if explicitly requested via the --freeze-cgroup option.

Fixes #2470


jesus-ramos (approved):
LGTM from the CUDA side, thanks Andrei.

criu/seize.c
```diff
@@ -993,12 +1006,14 @@ int collect_pstree(void)
...
-		goto err;
-
-	if (!opts.freeze_cgroup && compel_interrupt_task(pid)) {
-		set_cr_errno(ESRCH);
-		goto err;
+	if (opts.freeze_cgroup && !freeze_cgroup_disabled) {
+		if (freeze_processes())
+			goto err;
+	} else {
+		if (compel_interrupt_task(pid)) {
```
rst0git, line 1013: I was also experimenting with this approach but checkpointing fails with the following error:
```
sudo podman run -d \
	--rm \
	--privileged \
	--name=hlp \
	--device nvidia.com/gpu=all \
	--security-opt=label=disable \
		nvcr.io/nvidia/hpc-benchmarks:24.06 \
			mpirun --bind-to none -np 4 \
		   	./hpl.sh --dat /workspace/hpl-linux-x86_64/sample-dat/HPL-1GPU.dat

podman container checkpoint -l -e /tmp/test.tar
```
```
(33.240446) cuda_plugin: no need to pause devices on pid 620824
(33.240477) Detected cgroup V2 freezer
(33.240575) Seized task 620824, state 1
(33.240585) seccomp: Collected tid_real 620824 mode 0
(33.240613)     Seizing 620824's 622114 thread
(33.240655) Error (compel/src/lib/infect.c:263): Unseizable non-zombie 622114 found, state S, err -1/10
(33.240658)     Seizing 620824's 622115 thread
(33.240689) Error (compel/src/lib/infect.c:263): Unseizable non-zombie 622115 found, state S, err -1/10
(33.240691)     Seizing 620824's 622118 thread
(33.240721) Error (compel/src/lib/infect.c:263): Unseizable non-zombie 622118 found, state S, err -1/10
(33.240735)     Seizing 620824's 622114 thread
(33.240759) Error (compel/src/lib/infect.c:263): Unseizable non-zombie 622114 found, state S, err -1/10
(33.240760)     Seizing 620824's 622115 thread
(33.240784) Error (compel/src/lib/infect.c:263): Unseizable non-zombie 622115 found, state S, err -1/10
(33.240794)     Seizing 620824's 622118 thread
(33.240817) Error (compel/src/lib/infect.c:263): Unseizable non-zombie 622118 found, state S, err -1/10
(33.240819) Collected (-1 attempts, 3 in_progress)
```
[`dump.log`](https://paste.centos.org/view/420ad0f3)

rst0git: Hi Andrei, I have applied the changes from this pull request in https://github.com/checkpoint-restore/criu/pull/2470 and the example above seems to work:
- [dump.log](https://paste.centos.org/view/raw/e02768ec)
- [restore.log](https://paste.centos.org/view/raw/843e84b7)
```
$ checkpointctl show /tmp/test.tar 

Displaying container checkpoint data from /tmp/test.tar

+-----------+-------------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| CONTAINER |                IMAGE                |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
+-----------+-------------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| hlp       | nvcr.io/nvidia/hpc-benchmarks:24.06 | b4556b41b433 | runc    | 2024-09-12T14:19:02+01:00 | Podman | 2.7 GiB    | 136.1 MiB         |
+-----------+-------------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
```

rst0git: @avagin The following changes seems to fix the error above: https://github.com/rst0git/criu/commit/b6a97ba834a35d308ca5cfb434b6549dcd4ed1c0

avagin: You are right, we need to check freeze_cgroup_disabled in a few other places. Thanks.

avagin: > (If we ignore the race window between pause_devices_in_cgroup and freeze_processes, checkpoint/restore works for this container with the patch in #2470)

We can't ignore it and need to fix it. If we don't take into account random user workloads that can behave like a fork-bomb,using of a freeze cgroup doesn't bring a lot of value.

criu/seize.c
```diff
@@ -993,12 +1034,16 @@ int collect_pstree(void)
...
-		goto err;
-
-	if (!opts.freeze_cgroup && compel_interrupt_task(pid)) {
-		set_cr_errno(ESRCH);
-		goto err;
+	if (opts.freeze_cgroup && !freeze_cgroup_disabled) {
+		if (freeze_processes())
+			goto err;
+	} else {
+		if (opts.freeze_cgroup && unfreeze_cgroup())
```
Snorch, line 1041: Correct me if I'm wrong: It looks like here we try to unfreeze the freezer cgroup, which was frozen before running criu ("pre-frozen")?

I don't think this is correct as the "container-manager" which did this "pre-freeze" might do it not just for fun but for some deep synchronization reasons (e.g. filesystem is already checkpointed and by freezing processes it tries to guarantee consistency between fs checkpoint and criu checkpoint).

I believe we should just fail if cgroup is "pre-frozen".

avagin: Good point. Fixed. Thanks.

rst0git (approved):
LGTM

criu/seize.c
rst0git: > I don't think this is correct as the "container-manager" which did this "pre-freeze" might do it not just for fun but for some deep synchronization reasons (e.g. filesystem is already checkpointed and by freezing processes it tries to guarantee consistency between fs checkpoint and criu checkpoint).

This is correct. For example, in the case of CRI-O we freeze the container (https://github.com/cri-o/cri-o/pull/6977) to create a consistent snapshot of the container file system. We would need to find an alternative solution to enable GPU checkpointing.

PR #2479, opened 2024-09-14
plugin/cuda: disable CUDA plugin if /dev/nvidiactl isn't present

avagin:
The presence of /dev/nvidiactl indicates that the system has a compatible NVIDIA GPU driver installed and that the GPU is accessible to the operating system.


avagin:
@jesus-ramos is it the right assumption that nvidiactl should always exist to run cuda workloads? I am going to release v4.0 next week with the cuda plugin. Do you have anything that we need to merge into that release?

jesus-ramos (approved):
LGTM.

Yeah we expect /dev/nvidiactl to exist so that should be fine.

Nothing else to merge from me just yet. Looking forward to the release and thanks to you and Radostin on your help and work so far.

rst0git:
> Do you have anything that we need to merge into that release?

@avagin Currently we don't have a way of identifying if a checkpoint contains GPU state (e.g., CUDA is enabled during dump but disabled during restore). It might be useful to add a boolean flag in the inventory image.

avagin:
@rst0git it should be done in a generic way for all plugins. The first idea that comes to mind is that we need to have an array of required plugins in the inventory image. Plugins should be able to add new items into that arrays. Then on restore, we have to check that all required plugins are loaded.

plugins/cuda/cuda_plugin.c
```diff
@@ -460,8 +461,15 @@ CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_
 
 int cuda_plugin_init(int stage)
 {
-	int ret = cuda_checkpoint_supports_flag("--action");
+	int ret;
 
+	if (!fault_injected(FI_PLUGIN_CUDA_FORCE_ENABLE) && access("/dev/nvidiactl", F_OK)) {
```
rst0git, line 466: @avagin The [gpu-operator](https://github.com/NVIDIA/gpu-operator) for Kubernetes uses [`/run/nvidia/driver`](https://github.com/NVIDIA/gpu-operator/blob/d38a36f6b5d7a1e0dd6da600be4e061c7c83a183/controllers/object_controls.go#L175) as default path for driver installation.

In this case, `/dev/nvidiactl` does not exist:
```
# ls -al /dev/nvidiactl
ls: cannot access '/dev/nvidiactl': No such file or directory
# ls -al /run/nvidia/driver/dev/nvidiactl
crw-rw-rw- 1 root root 195, 255 Nov  1 15:45 /run/nvidia/driver/dev/nvidiactl
```

PR #2489, opened 2024-09-27
images/inventory: add field for enabled plugins

rst0git:
This pull request extends the inventory image with a `plugins` field that contains an array of names indicating which plugins were used during checkpointing, for example, to save GPU state. In particular, the CUDA and AMDGPU plugins are added to this field *only* when the checkpoint contains GPU state. This allows to disable unnecessary plugins during restore, show appropriate error messages if required CRIU plugins are missing, and migrate a process that does not use GPU from a GPU-enabled system to CPU-only environment.

Examples:
1. The checkpoint contains AMDGPU state and requires `amdgpu_plugin` for restore.
```json
{
    "magic": "INVENTORY",
    "entries": [
        {
            "plugins_entry": {
                "plugins": [
                    "amdgpu_plugin"
                ]
            }
        }
    ]
}
```
2. The checkpoint contains CUDA state and requires `cuda_plugin` for restore.
```json
{
    "magic": "INVENTORY",
    "entries": [
        {
            "plugins_entry": {
                "plugins": [
                    "cuda_plugin"
                ]
            }
        }
    ]
}
```
3.  The checkpoint does not contain GPU state (i.e., no plugins are required for restore).
```json
{
    "magic": "INVENTORY",
    "entries": [
        {
            "plugins_entry": {}
        }
    ]
}
```

### DIFF
--- a/criu/image.c
+++ b/criu/image.c
@@ -26,6 +26,14 @@ TaskKobjIdsEntry *root_ids;
 u32 root_cg_set;
 Lsmtype image_lsm;
 
+struct inventory_plugin {
+	struct list_head node;
+	char *name;
+};
+
+struct list_head inventory_plugins_list = LIST_HEAD_INIT(inventory_plugins_list);
+static int n_inventory_plugins;
+
 int check_img_inventory(bool restore)
 {
 	int ret = -1;
@@ -99,6 +107,19 @@ int check_img_inventory(bool restore)
 		} else {
 			opts.network_lock_method = he->network_lock_method;
 		}
+
+		if (!he->plugins_entry) {
+			/* backwards compatibility: if the 'plugins_entry' field is missing,
+			 * all plugins should be enabled during restore.
+			 */
+			n_inventory_plugins = -1;
+		} else {
+			PluginsEntry *pe = he->plugins_entry;
+			for (int i = 0; i < pe->n_plugins; i++) {
+				if (add_inventory_plugin(pe->plugins[i]))
+					goto out_err;
+			}
+		}
 	}
 
 	ret = 0;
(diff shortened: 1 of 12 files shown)


criu/plugin.c
```diff
@@ -247,6 +251,18 @@ int cr_plugin_init(int stage)
 		if (len < 3 || strncmp(de->d_name + len - 3, ".so", 3))
 			continue;
 
+		if (enabled_plugins_inventory && stage == CR_PLUGIN_STAGE__RESTORE) {
+			if (len == 16 && !strncmp(de->d_name, "amdgpu_plugin.so", 16)) {
```
avagin, line 255: the criu generic code should not be aware about specific plugins.

images/inventory.proto
```diff
@@ -21,4 +26,5 @@ message inventory_entry {
 	optional uint32			pre_dump_mode	= 9;
 	optional bool			tcp_close	= 10;
 	optional uint32			network_lock_method	= 11;
+	optional plugins_entry		plugins = 12;
```
avagin, line 29: I think it should be just an array of strings. Plugins should be able to add entries on the dump stage and they should be able to delete entries on the restore stage. After processing all plugins, criu checks whether it is empty or not. If it isn't empty, criu reports a error that some plugins haven't been loaded? Does it make sense?

rst0git: It makes sense. Protobuf doesn't have `optional array` type but we can use `repeated string` instead.

>  After processing all plugins, criu checks whether it is empty or not.

Would it make sense to load only the plugins listed in the `plugins` field during restore? In other words, if the field is empty, we don't need to load any plugins.

avagin: > Would it make sense to load only the plugins that are listed in the plugins field during restore? In other words, if the field is empty, we don't need to load any plugins.

It might better to make this on the plugin level. On restore, a plugin checks if it is in the list and makes a decision what hooks to register.

rst0git: Thanks @avagin! I've updated the pull request with these changes.

criu/image.c
```diff
@@ -110,8 +131,87 @@ int check_img_inventory(bool restore)
...
+{
+	if (n_inventory_plugins == -1)
+		return true; /* backwards compatibility */
+
+	if (n_inventory_plugins > 0) {
+		struct inventory_plugin *p, *tmp;
+		list_for_each_entry_safe(p, tmp, &inventory_plugins_list, node) {
+			if (!strncmp(name, p->name, n)) {
+				xfree(p->name);
+				list_del(&p->node);
```
avagin, line 149: xfree(p)?

rst0git: Thank you for catching this! I've updated the pull request.

plugins/cuda/cuda_plugin.c
```diff
@@ -354,6 +356,15 @@ int cuda_plugin_checkpoint_devices(int pid)
 			pr_err("Failed to restore process after error %s on pid %d\n", msg_buf, pid);
 		}
 	}
+
+	if (!plugin_added_to_inventory) {
+		status = add_inventory_plugin(CR_PLUGIN_DESC.name);
+		if (status)
+			pr_err("Failed to add CUDA plugin to inventory image\n");
```
avagin, line 363: should we return a non-zero code in this case?

avagin:
I think we should introduce a test for this change.

criu/image.c
```diff
@@ -110,8 +131,91 @@ int check_img_inventory(bool restore)
...
+ * the specified plugin name. If found, the plugin is removed
+ * from the linked list.
+ */
+bool check_and_remove_inventory_plugin(const char *name, size_t n)
+{
+	if (n_inventory_plugins == -1)
+		return true; /* backwards compatibility */
+
+	if (n_inventory_plugins > 0) {
+		struct inventory_plugin *p, *tmp;
```
avagin, line 145: nit: add an empty line between declarations and the code block.

test/plugins/Makefile
```diff
@@ -12,6 +20,12 @@ amdgpu_plugin.so: $(SRC_DIR)/amdgpu/amdgpu_plugin.so
 cuda_plugin.so: $(SRC_DIR)/cuda/cuda_plugin.so
 	$(Q) cp $< $@
 
+inventory_test_enabled_plugin.so: inventory_test_enabled_plugin.c
+	gcc $(PLUGIN_CFLAGS) $< -o $@ $(PLUGIN_INCLUDE)
```
avagin, line 24: s/gcc/$(Q) $(CC)/

avagin:
LGTM

jesus-ramos (approved):
LGTM

Issue #2490, opened 2024-10-08
zdtm/static/pidfd_dead fails in Fedora Rawhide:


avagin:
It looks like this test fails in Fedora Rawhide:

```
======================= Run zdtm/static/pidfd_dead in h ========================
Start test
./pidfd_dead --pidfile=pidfd_dead.pid --outfile=pidfd_dead.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/pidfd_dead/168/1/restore.log
------------------------ grep Error ------------------------
b'(00.004951)    168: \t\t\tGoing to dup 1 into 2'
b'(00.004953)    168: \tReceive fd for 2'
b'(00.004961)    168: \t\tCreate fd for 3'
b'(00.005264)    168: \t\tCreate fd for 4'
b'(00.005486)    168: Error (criu/cr-restore.c:1261): 175 killed by signal 9: Killed'
b'(00.005504) Error (criu/cr-restore.c:2314): Restoring FAILED.'
------------------------ ERROR OVER ------------------------
############### Test zdtm/static/pidfd_dead FAIL at CRIU restore ###############
Test output: ================================
```
https://github.com/checkpoint-restore/criu/runs/31040419148

_Originally posted by @rst0git in https://github.com/checkpoint-restore/criu/pull/2449#discussion_r1790787295_

PR #2491, opened 2024-10-09
pidfd: block SIGCHLD during tmp process creation

rst0git:
This pull request updates pidfd to block `SIGCHLD` during temporary process creation to prevent a race condition between `kill()` and `waitpid()` where `sigchld_handler()` causes `criu restore` to fail with an error.

Fixes: #2490


criu/pidfd.c
```diff
@@ -158,6 +159,9 @@ static int free_dead_pidfd(struct dead_pidfd *dead)
 		goto err;
 	}
 
+	/* Restore the original signal mask after tmp process has terminated */
+	sigprocmask(SIG_SETMASK, &dead->oldmask, NULL);
```
bsach64, line 163: We should probably check return value of `sigprocmask`

criu/pidfd.c
```diff
@@ -199,6 +204,14 @@ static int open_one_pidfd(struct file_desc *d, int *new_fd)
 	BUG_ON(dead->count == 0);
 	dead->count--;
 	if (dead->pid == -1) {
+		/* Block SIGCHLD to prevent interfering from sigchld_handler()
+		 * and to properly handle the tmp process termination without
+		 * a race condition. A similar approach is used in cr_system().
+		 */
+		sigemptyset(&blockmask);
+		sigaddset(&blockmask, SIGCHLD);
+		sigprocmask(SIG_BLOCK, &blockmask, &dead->oldmask);
```
bsach64, line 213: Similiarly, here

bsach64 (commented):
Thank you for the PR @rst0git!
I was about to raise one myself!

mihalicyn:
Hi Radostin!

Thanks a lot for looking into it! We had a call with Bhavik today and he wants to see if we can use pstree helpers to solve this problem. So, let's wait a bit with merging this.

rst0git:
> Thank you for the PR @rst0git! I was about to raise one myself!

No worries, I just wanted to confirm that the pidfd_dead test will pass with these changes in CI.

> We had a call with Bhavik today and he wants to see if we can use pstree helpers to solve this problem.

I'm not sure how easy it would be to do that but the messages in the following commits might help: https://github.com/checkpoint-restore/criu/commit/c09ba04c487cacacc4945039740c98096b6ef0ab https://github.com/checkpoint-restore/criu/commit/5f729636b49c6756e2f9ac8459787315c933c118

criu/pidfd.c
```diff
@@ -199,6 +207,18 @@ static int open_one_pidfd(struct file_desc *d, int *new_fd)
 	BUG_ON(dead->count == 0);
 	dead->count--;
 	if (dead->pid == -1) {
+		/* Block SIGCHLD to prevent interfering from sigchld_handler()
+		 * and to properly handle the tmp process termination without
+		 * a race condition. A similar approach is used in cr_system().
+		 */
+		sigemptyset(&blockmask);
+		sigaddset(&blockmask, SIGCHLD);
+		if (sigprocmask(SIG_BLOCK, &blockmask, &dead->oldmask) == -1) {
```
avagin, line 216: sigprocmask should be called right before killing this process. If I remember it right, a process can be created in one process, but killed in another process.

rst0git: It makes sense. I've updated the pull request.

criu/pidfd.c
```diff
@@ -145,6 +145,19 @@ static int create_tmp_process(void)
 static int free_dead_pidfd(struct dead_pidfd *dead)
 {
 	int status;
+	sigset_t blockmask, oldmask;
+
+	/* Block SIGCHLD to prevent interfering from sigchld_handler()
```
avagin, line 150: nit: We usually use a slightly different format for comments:
/*
 * Block SIGCHLD to prevent interfering from sigchld_handler()

avagin:
LGTM

PR #2492, opened 2024-10-12
uffd: Disable image deduplication after fork

liusdu:
After a fork, both the child and parent processes may trigger a page fault (#PF) at the same virtual address, referencing the same position in the page image. If deduplication is enabled, the last process to trigger the page fault will fail.

Therefore, deduplication should be disabled after a fork to prevent this issue.

### DIFF
--- a/criu/include/pagemap.h
+++ b/criu/include/pagemap.h
@@ -58,6 +58,9 @@ struct page_read {
 	/* Whether or not pages can be read in PIE code */
 	bool pieok;
 
+	/* Whether or not disable image deduplication*/
+	bool disable_dedup;
+
 	/* Private data of reader */
 	struct cr_img *pmi;
 	struct cr_img *pi;
@@ -111,6 +114,8 @@ int pagemap_render_iovec(struct list_head *from, struct task_restore_args *ta);
  * maintains its own set of references to those structures.
  */
 extern void dup_page_read(struct page_read *src, struct page_read *dst);
+
+extern void page_read_disable_dedup(struct page_read *pr);
 
 extern int dedup_one_iovec(struct page_read *pr, unsigned long base, unsigned long len);
 
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -261,7 +261,7 @@ static int read_local_page(struct page_read *pr, unsigned long vaddr, unsigned l
 			break;
 	}
 
-	if (opts.auto_dedup) {
+	if (opts.auto_dedup && !pr->disable_dedup) {
 		ret = punch_hole(pr, pr->pi_off, len, false);
 		if (ret == -1)
 			return -1;
@@ -792,6 +792,7 @@ int open_page_read_at(int dfd, unsigned long img_id, struct page_read *pr, int p
 	pr->bunch.iov_base = NULL;
 	pr->pmes = NULL;
 	pr->pieok = false;
+	pr->disable_dedup = false;
 
 	pr->pmi = open_image_at(dfd, i_typ, O_RSTR, img_id);
 	if (!pr->pmi)
@@ -851,6 +852,14 @@ int open_page_read(unsigned long img_id, struct page_read *pr, int pr_flags)
 }
 
 #define DUP_IDS_BASE 1000
+
+void page_read_disable_dedup(struct page_read *pr)
+{
+	pr_debug("disable dedup, id: %d\n", pr->id);
+	pr->disable_dedup = true;
+	if (pr->parent)
+		page_read_disable_dedup(pr->parent);
+}
 
 void dup_page_read(struct page_read *src, struct page_read *dst)
 {
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -1098,6 +1098,8 @@ static int handle_fork(struct lazy_pages_info *parent_lpi, struct uffd_msg *msg)
 
 	lpi_get(lpi->parent);
 
+	page_read_disable_dedup(&parent_lpi->pr);
+	page_read_disable_dedup(&lpi->pr);
 	return 1;
 
 out:


criu/include/pagemap.h
```diff
@@ -58,6 +58,9 @@ struct page_read {
 	/* Whether or not pages can be read in PIE code */
 	bool pieok;
 
+    /* Whether or not disable image deduplication*/
+    bool disable_dedup;
+
```
rst0git, line 63: @liusdu Would you be able to fix this patch to use tabs instead of spaces for indentation?

rst0git:
>If deduplication is enabled, the last process to trigger the page fault will fail.

Would it be possible to add a ZDTM test for this use-case?

criu/include/pagemap.h
liusdu: ok, @rst0git sorry to this~

liusdu:
> > If deduplication is enabled, the last process to trigger the page fault will fail.
> 
> Would it be possible to add a ZDTM test for this use-case?

I will add a ZDTM  testcase later. The follow test case can reproduce this issue.

- source code
```
#define _GNU_SOURCE
#include <stdio.h>
#include <stdlib.h>
#include <fcntl.h>
#include <unistd.h>
#include <sys/mman.h>
#include <string.h>
#include <signal.h>
#include <sched.h>


#define VALUE 0x1001
#define STACK_SIZE (1024 * 1024)

int validate_memory(void *arg)
{
    if  (*(int *)arg != VALUE) {
        printf("memory is corrupted\n");
        return -1;
    }
    printf("memory: %x\n", *(int *)arg);
    return 0;
}

int main() {
    int pid = 0;
    char *stack;
    int *buffer;;
     int flags = CLONE_VFORK;
    stack = malloc(STACK_SIZE + 4096);
    if (stack == NULL) {
        perror("malloc");
        exit(EXIT_FAILURE);
    }
    buffer = (int *)(stack + STACK_SIZE);
    buffer[0] = VALUE;

    kill(getpid(), 19);
    if (clone(validate_memory, buffer, flags, (void *)buffer) == -1) {
        perror("clone");
        free(stack);
        exit(EXIT_FAILURE);
    }
    validate_memory(buffer);

    return 0;
}
```
- steps
```
  gcc -o clone clone.c 
  setsid ./clone
  criu dump --shell-job  -vvvv --tree `pidof ./clone` -o dumo.log  --images-dir ./ckp 
  /data/home/liusdu/work/source/criu/criu/criu lazy-pages -vvvv -o lp.log  --auto-dedup --images-dir ckp/ &
  /data/home/liusdu/work/source/criu/criu/criu  restore -vvvv -o restore.log  --images-dir  ./ckp/ --lazy-pages -j --no-background-fetch  &
  kill -s SIGCONT `ps aux | grep '\[clone\]' | awk '{print $2}'`
 
```

criu/config.c
```diff
@@ -703,6 +703,7 @@ int parse_options(int argc, char **argv, bool *usage_error, bool *has_exec_cmd,
 		BOOL_OPT("mntns-compat-mode", &opts.mntns_compat_mode),
 		BOOL_OPT("unprivileged", &opts.unprivileged),
 		BOOL_OPT("ghost-fiemap", &opts.ghost_fiemap),
+		BOOL_OPT("no-background-fetch", &opts.no_background_fetch),
```
rst0git, line 706: @liusdu What is the use-case for this option?

cc: @rppt

liusdu: This is for the test case. We know that CRIU can set up pages in the background or on demand in lazy-page mode. We must disable the background mode to ensure that both the child and parent processes trigger two page faults (#PF). Otherwise, by the time the child and parent processes start running, the lazy page server may have already completed its tasks and exited

duguhaotian:
can you show some error message of this issuse?

criu/config.c
rst0git: Ah, I see. We typically add command-line options to enable new CRIU features and document these options in the man page and `criu --help`. However, having a large number of options may lead to confusion for users who are not familiar with the available CRIU features.

Would it be possible to use an environment variable instead? For example, something like the following:
```
CRIU_DISABLE_BACKGROUND_LAZY_PAGE_FETCH=1  criu restore ...
```

liusdu: ok~

criu/cr-restore.c
```diff
@@ -2213,7 +2213,7 @@ static int restore_root_task(struct pstree_item *init)
 		goto out_kill_network_unlocked;
 	}
 
-	if (lazy_pages_finish_restore())
+	if (!opts.no_background_fetch && lazy_pages_finish_restore())
```
rppt, line 2216: I'm not sure I understand how this would prevent background fetch

criu/config.c
rppt: > Ah, I see. We typically add command-line options to enable new CRIU features and document these options in the man page and `criu --help`. However, having a large number of options may lead to confusion for users who are not familiar with the available CRIU features.
> 
> Would it be possible to use an environment variable instead? For example, something like the following:
> 
> ```
> CRIU_DISABLE_BACKGROUND_LAZY_PAGE_FETCH=1  criu restore ...
> ```

Or maybe even fault injection mechanism, because normally we don't want to disable the background fetch

criu/cr-restore.c
liusdu: Yes, I found I can not prevent background fetch.  Even I apply the following patch.
```
diff --git a/criu/uffd.c b/criu/uffd.c
index 98c2b7e07..0ca4cdbaa 100644
--- a/criu/uffd.c
+++ b/criu/uffd.c
@@ -68,6 +68,7 @@
 #define MAX_XFER_LEN    (4 << 20)
 
 static mutex_t *lazy_sock_mutex;
+static bool disable_background_lazy_page_fetch;
 
 struct lazy_iov {
        struct list_head l;
@@ -1391,11 +1392,15 @@ static int prepare_uffds(int listen, int epollfd)
                        goto close_uffd;
        }
 
-       lazy_sk_rfd.fd = client;
-       lazy_sk_rfd.read_event = lazy_sk_read_event;
-       lazy_sk_rfd.hangup_event = lazy_sk_hangup_event;
-       if (epoll_add_rfd(epollfd, &lazy_sk_rfd))
-               goto close_uffd;
+    if (!disable_background_lazy_page_fetch) {
+           lazy_sk_rfd.fd = client;
+           lazy_sk_rfd.read_event = lazy_sk_read_event;
+           lazy_sk_rfd.hangup_event = lazy_sk_hangup_event;
+           if (epoll_add_rfd(epollfd, &lazy_sk_rfd))
+                   goto close_uffd;
+    } else {
+           restore_finished = true;
+    }
 
        close(listen);
        return 0;
@@ -1475,3 +1480,8 @@ int cr_lazy_pages(bool daemon)
        xfree(events);
        return ret;
 }
+
+static void __attribute__((constructor)) background_lazy_page_fetch_init(void)
+{
+    disable_background_lazy_page_fetch = (getenv("CRIU_DISABLE_BACKGROUND_LAZY_PAGE_FETCH") != NULL);
+}

```

liusdu:
@rst0git @rppt I found it very difficult to disable background fetch. I cannot reliably reproduce this issue in the tests because the memory may be fetched by the background process before deduplication. However, I can reliably reproduce this issue with a large-memory process, which is not suitable for testing. Could you please consider merging this patch without the test?

liusdu:
> can you show some error message of this issuse?

Please refer to the above testcase.

rst0git (approved):
LGTM

@liusdu Thank you for working on this!

rppt (approved):

Issue #2499, opened 2024-10-21
error: Can't bind transport socket /crtools-fd-3-0: Address in use


fntlnz:
**Description**
 Hey team! 

When restoring a dump in a new mount + pid namspace in an environment where multiple dumps sharing the same network namespace
can happen simultaneously CRIU was observed to create an anonymous  unix socket named `crtools-fd-3-0`.

Note: I already prepared a patch with what I think would fix this, submitting it right after opening this.

The log line that appears in a `restore.log` when this happens will look something like this

```
1606:(01.379765)      3: Error (criu/files.c:1695): Can't bind transport socket /crtools-fd-3-0: Address in use
```

The pattern pattern `crtools-fd-%d-%d` is defined in[ files.c](https://github.com/checkpoint-restore/criu/blob/0a7c5fd1bd8d1e49e273b51ff39af473d6c68cbc/criu/files.c#L973).

In this case `3` is always the restore PID, even when multiple restores are running, because the same process is being restored in a pid namespace whith PID 3, however I was surprised to see that `criu_run_id` was always `0`.

At that point, we investigated a bit and noticed that the last element of that print is the id of the pid namespace, which is obviously not zero in our case so something was off.

It looks like that bit is set in [`util_init`](https://github.com/checkpoint-restore/criu/blob/208f60f04638d44594dfe3845e6a2aec82e16e1a/criu/util.c#L1977), however, in `crtools.c` `util_init` is called after `cr_service_work` is started so srwk will never have that bit set.

**Steps to reproduce the issue:**

Start a new process + mount namespace with a process to dump inside of it

```
unshare --pid --mount --fork /bin/bash
top # <--- this is the process we want to dump
```

Now nsenter that namespace and dump top

```
nsenter -t <bash pid> -p -m
criu dump -t $TOP_PID  -D /tmp/criu-dump -v4 --shell-job
```

Now try to restore that process multiple times simultaneously in two different dump folders

```
cp /tmp/criu-dump /tmp/criu-dump-1
cp /tmp/criu-dump /tmp/criu-dump-2
```

For example, in two different terminals, start a new pid namespace again and restore the process.

```
unshare --pid --mount --fork --mount-proc criu restore -d -D /tmp/criu-dump-1 -v4 --shell-job
```
and

```
unshare --pid --mount --fork --mount-proc criu restore -d -D /tmp/criu-dump-2 -v4 --shell-job
```

[...]

PR #2500, opened 2024-10-21
criu: util_init before the criu service worker is started

fntlnz:
When restoring dumps in new mount + pid namespaces where multiple dumps
share the same network namespace, CRIU may fail due to conflicting
unix socket names. This happens because the service worker creates
sockets using a pattern that includes criu_run_id, but util_init()
is called after cr_service_work() starts.

The socket naming pattern "crtools-fd-%d-%d" uses the restore PID
and criu_run_id, however criu_run_id is always 0 when not initialized,
leading to conflicts when multiple restores run simultaneously either
in the same CRIU process or because of multiple CRIU processes
doing the same operation in different PID namespaces.

Fix this by:

- Moving util_init() before cr_service_work() starts
- Adding a second util_init() call in the service worker fork
to ensure unique IDs across multiple worker runs
- Making sure that dump and restore operations have util_init() called
early to generate unique socket names

With this fix, socket names always include the namespace ID, preventing
conflicts when multiple processes with the same pid share a network
namespace.

Fixes https://github.com/checkpoint-restore/criu/pull/2499


criu/crtools.c
```diff
@@ -170,6 +170,7 @@ int main(int argc, char *argv[], char *envp[])
 		goto usage;
 	}
 
+	util_init();
```
avagin, line 173: pls write a comment why it should be initialized here and update the commit message with more details.

rst0git: The following commit is a good example: https://github.com/checkpoint-restore/criu/commit/e8087fcff6feae2384c00b0145bf13e168785cc2

avagin: @rst0git This is a bad example. It has an empty commit message and zero comments;).

criu/crtools.c
```diff
@@ -254,8 +255,6 @@ int main(int argc, char *argv[], char *envp[])
 		return 1;
 	}
 
-	util_init();
```
avagin, line 257: I think we need to call it separately from cr_service_work. Each new service work should have an uniq criu_run_id and compel_run_id, because several service works can run concurrently.

fntlnz: you are talking about this right? https://github.com/checkpoint-restore/criu/blob/6cfe7aa1142093f93a38a127a8a063e1470bca65/criu/cr-service.c#L1531

I've been digging a bit as well, looks like having it at the begginging of `cr_service_work` would be enough in all cases.

criu/crtools.c
fntlnz: Ok I'll update that one :) Thanks!

criu/crtools.c
fntlnz: After testing a little bit more I realized that we need to init in multiple places, once when  `cr_service_work` is forked and in the location I was already proposing in this patch.

criu/crtools.c
fntlnz: Done in 3c4e1cb73ea72a984ddac917a19dd40dddf6fa55

rst0git:
> Fixes https://github.com/checkpoint-restore/criu/pull/2500

@fntlnz Would you be able to replace '2500' with '2499'?

fntlnz:
@rst0git done! Sorry for the oversight.

rst0git:
@fntlnz I'm not able to replicate this error locally.

Would it be possible to add a test that fails due to socket name conflict and passes when this patch is applied?
(e.g., you can add a new 'fault' (see [`fault-injection.h`](https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/include/fault-injection.h)) that causes CRIU to always use the same socket name)

avagin:
@fntlnz I pushed this change into the criu-dev branch. I have removed the util_init call right before cr_service_work, because it is called from cr_service_work and I don't think that we need to do this twice. Let me know if you have any concerns. Thank you for debugging and fixing the problem.

PR #2502, opened 2024-10-24
pass real fd when restore and dump ext file

duguhaotian:
**Changes：**
1. dump ext file: should pass real fd in dumped process;
2. restore ext file: should pass target fd which will open;

**Background:**
During the restoration process of an NPU device, it is necessary to distinguish between file descriptors (fd) to perform different operations.
1. Record operations targeting the device file descriptor (fd);
2. Replay the corresponding operations based on the fd.


avagin:
pls write detailed commit message:
https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#describe-your-changes

duguhaotian:
> pls write detailed commit message: https://www.kernel.org/doc/html/v4.10/process/submitting-patches.html#describe-your-changes

done

Issue #2496, opened 2024-10-18
pidfd: Cannot restore multiple processes pointing to a common dead pidfd


bsach64:
**Description**
Please read the reproducer code: https://gist.github.com/bsach64/fcbdaec357fd7ed4212a86ac81b2e8bb

``` C

#define _GNU_SOURCE

#include <sys/statfs.h>
#include <stdio.h>
#include <sys/syscall.h>
#include <unistd.h>
#include <sys/wait.h>
#include <fcntl.h>
#include <sys/stat.h>

#ifndef PID_FS_MAGIC
#define PID_FS_MAGIC 0x50494446
#endif

/*
 * main
 *	`- child
 *		`- grandchild
 *
 * main and child open a pidfd for grandchild.
 * Before C/R we kill grandchild.
 * We end up with two pidfds in different process of
 * pointing to a common dead process.
 */

static int pidfd_open(pid_t pid, unsigned int flags)
{
	return syscall(__NR_pidfd_open, pid, flags);
}

static int pidfd_send_signal(int pidfd, int sig, siginfo_t* info, unsigned int flags)
{
	return syscall(__NR_pidfd_send_signal, pidfd, sig, info, flags);
}

int main(int argc, char* argv[])
{
	#define READ 0
	#define WRITE 1

	int child, ret, gchild, p[2], status, pidfd;

	if (pipe(p)) {
		perror("pipe");
		return 1;
	}

	child = fork();
	if (child < 0) {
		perror("fork");
		return 1;
	} else if (child == 0) {
		int gchild = fork();
		close(p[READ]);
		if (gchild < 0) {
			perror("fork");
			return 1;
		} else if (gchild == 0) {
			close(p[WRITE]);
			while(1)
				sleep(1000);
		} else {
			int pidfd;
			pidfd = pidfd_open(gchild, 0);
			if (pidfd < 0) {
				perror("pidfd_open");
				return 1;
			}

			if (write(p[WRITE], &gchild, sizeof(int)) != sizeof(int)) {
				printf("write");
				return 1;
			}

			close(p[WRITE]);
			if (waitpid(gchild, &status, 0) != gchild) {
				printf("waitpid");
				return 1;
			}

			while(1)
				sleep(1000);
		}
	}

	close(p[WRITE]);
	if (read(p[READ], &gchild, sizeof(int)) != sizeof(int)) {
		printf("write");
		return 1;
	}
	close(p[READ]);

[...]

PR #2505, opened 2024-10-28
pidfd: tmp process is waited upon by correct process

bsach64:
Fixes: #2496
This patch ensures that the process that created the tmp process is the one that kills and waits for it after all pidfds have been opened. We do this by keeping track of each process that has to open a pidfd for the tmp process and the number of pidfds it has to open.

Thing I am unsure about need to dump the pid of the process that opened the pidfd. (This might be available from some where else)

### DIFF
--- a/criu/pidfd.c
+++ b/criu/pidfd.c
@@ -21,32 +21,26 @@
 struct pidfd_info {
 	PidfdEntry *pidfe;
 	struct file_desc d;
+
+	struct dead_pidfd *dead;
+	struct pidfd_info *next;
 };
 
 struct dead_pidfd {
 	unsigned int ino;
-	int pid;
-	size_t count;
-	mutex_t pidfd_lock;
+	int creator_id;
+
 	struct hlist_node hash;
+	struct pidfd_info *list;
 };
 
 #define DEAD_PIDFD_HASH_SIZE 32
 static struct hlist_head dead_pidfd_hash[DEAD_PIDFD_HASH_SIZE];
-static mutex_t *dead_pidfd_hash_lock;
 
-int init_dead_pidfd_hash(void)
+void init_dead_pidfd_hash(void)
 {
 	for (int i = 0; i < DEAD_PIDFD_HASH_SIZE; i++)
 		INIT_HLIST_HEAD(&dead_pidfd_hash[i]);
-
-	dead_pidfd_hash_lock = shmalloc(sizeof(*dead_pidfd_hash_lock));
-	if (!dead_pidfd_hash_lock)
-		return -1;
-
-	mutex_init(dead_pidfd_hash_lock);
-
-	return 0;
 }
 
 static struct dead_pidfd *lookup_dead_pidfd(unsigned int ino)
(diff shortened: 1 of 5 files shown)


criu/pidfd.c
```diff
@@ -23,11 +25,19 @@ struct pidfd_info {
 	struct file_desc d;
 };
 
+struct dead_count {
```
avagin, line 28: honestly, I don't understand what is going on here and I don't want to turn on the sherklock mode to figure out what you mean here. Pls write comments and a detailed commit message.

criu/pidfd.c
```diff
@@ -49,6 +59,30 @@ int init_dead_pidfd_hash(void)
 	return 0;
 }
 
+static struct dead_count *init_dead_count(int pidfd_owner)
+{
+	struct dead_count *spec_count;
```
avagin, line 64: nit: empty line between code and declarations.

criu/pidfd.c
avagin: why one process can't create a helper, open all required pidfd-s and send them to others? Cc: @mihalicyn

avagin: Here is a draft version of what I mean in the previous comment:
https://github.com/avagin/criu/commit/f14adde2c4d7da5ad8cd6066c4867e6d6130c420

bsach64: This is a much better approach to solve this problem! Thanks @avagin 
Shouldn't the allocation for `dead` still be `shmalloc()` ? https://github.com/avagin/criu/commit/f14adde2c4d7da5ad8cd6066c4867e6d6130c420#diff-0b8ec9ac559a1c3dc6868b3f4ffaae306e633c93c387fc5867ddc83922546b9eR278

criu/pidfd.c
bsach64: no longer needed

criu/pidfd.c
avagin: All dead structures are allocated from one process before forking other processes, the we just read them. I don't see any reason to use shmalloc.

criu/pidfd.c
```diff
@@ -269,32 +266,31 @@ static int collect_one_pidfd(void *obj, ProtobufCMessage *msg, struct cr_img *i)
...
 
-	INIT_HLIST_NODE(&dead->hash);
-	dead->ino = info->pidfe->ino;
-	dead->count = 1;
-	dead->pid = -1;
-	mutex_init(&dead->pidfd_lock);
+	info->dead = dead;
+	info->next = dead->list;
+	dead->list = info;
+	if (dead->creator_id > info->pidfe->id)
```
mihalicyn, line 291: We have discussed that with @avagin and a tricky thing here is that we need to ensure that there are no deadlock when process A waits on process B to restore some fd1, while process B waits on process A to restore (another) fd2. To prevent those deadlocks we have special way of deciding who will restore an fd (see `fdinfo_rst_prio`) and `mark_pipe_master_cb` (good example). I think we need to implement something like that here.

mihalicyn: @bsach64 if you want I can do this on top of this PR so you can squash it later.

bsach64: Okay Alex!

avagin:
LGTM

mihalicyn (approved):
I played with this PR locally and tried to make a deadlock. By combining pipe+pidfd in that way so one process became "master" (the one who restores pidfd or pipe pair) while another one became "slave" (the one who received fd through `TRANSPORT_FD` socket). But taking into account the fact that `recv_desc_from_peer` is non-blocking call and it can return `1` (which means "please retry later") we are on the safe side.

LGTM.

mihalicyn:
I'll do some small adjustments in this PR to fix code style issues so we can merge it today.

rst0git (approved):
LGTM

Issue #2509, opened 2024-11-01
The gpu-operator for Kubernetes uses /run/nvidia/driver as default path for driver installation.


avagin:
The  [gpu-operator](https://github.com/NVIDIA/gpu-operator) for Kubernetes uses [`/run/nvidia/driver`](https://github.com/NVIDIA/gpu-operator/blob/d38a36f6b5d7a1e0dd6da600be4e061c7c83a183/controllers/object_controls.go#L175) as default path for driver installation.

In this case, `/dev/nvidiactl` does not exist:
```
# ls -al /dev/nvidiactl
ls: cannot access '/dev/nvidiactl': No such file or directory
# ls -al /run/nvidia/driver/dev/nvidiactl
crw-rw-rw- 1 root root 195, 255 Nov  1 15:45 /run/nvidia/driver/dev/nvidiactl

# mount | grep run/nvidia/driver/dev 
tmpfs on /run/nvidia/driver/dev type tmpfs (rw,nosuid,size=65536k,mode=755,inode64)
devpts on /run/nvidia/driver/dev/pts type devpts (rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=666)
mqueue on /run/nvidia/driver/dev/mqueue type mqueue (rw,nosuid,nodev,noexec,relatime)
tmpfs on /run/nvidia/driver/dev/log type tmpfs (rw,nosuid,nodev,noexec,relatime,size=158496212k,mode=755,inode64)
/dev/md127 on /run/nvidia/driver/dev/termination-log type ext4 (rw,relatime,discard,quota,usrquota,stripe=1024)
shm on /run/nvidia/driver/dev/shm type tmpfs (rw,nosuid,nodev,noexec,relatime,size=65536k,inode64)
```

The CUDA plugin for CRIU is disabled as a result, leading to checkpoint failure.

_Originally posted by @rst0git in https://github.com/checkpoint-restore/criu/pull/2479#discussion_r1826133061_

PR #2510, opened 2024-11-02
cuda: fix check for GPU device availability

rst0git:
The check for `/dev/nvidiactl` to determine if the CUDA plugin can be used is unreliable because in some cases the default path for driver installation is different [^1]. This pull request changes the logic to check if a GPU device is available in `/proc/driver/nvidia/gpus/`. This approach is similar to [`torch.cuda.is_available()`](https://pytorch.org/docs/stable/_modules/torch/cuda.html#is_available) and it is a more accurate indicator. The subsequent check for support of the `cuda-checkpoint --action` option would confirm if the driver supports checkpoint/restore.

[^1]: https://github.com/NVIDIA/gpu-operator

Fixes: #2509

### DIFF
--- a/plugins/cuda/cuda_plugin.c
+++ b/plugins/cuda/cuda_plugin.c
@@ -470,6 +470,20 @@ int cuda_plugin_resume_devices_late(int pid)
 }
 CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_devices_late)
 
+/**
+ * Check if a CUDA device is available on the system
+ */
+static bool is_cuda_device_available(void)
+{
+	const char *gpu_path = "/proc/driver/nvidia/gpus/";
+	struct stat sb;
+
+	if (stat(gpu_path, &sb) != 0)
+		return false;
+
+	return S_ISDIR(sb.st_mode);
+}
+
 int cuda_plugin_init(int stage)
 {
 	int ret;
@@ -481,8 +495,8 @@ int cuda_plugin_init(int stage)
 		}
 	}
 
-	if (!fault_injected(FI_PLUGIN_CUDA_FORCE_ENABLE) && access("/dev/nvidiactl", F_OK)) {
-		pr_info("/dev/nvidiactl doesn't exist. The CUDA plugin is disabled.\n");
+	if (!fault_injected(FI_PLUGIN_CUDA_FORCE_ENABLE) && !is_cuda_device_available()) {
+		pr_info("No GPU device found; CUDA plugin is disabled\n");
 		plugin_disabled = true;
 		return 0;
 	}


jesus-ramos (approved):

plugins/cuda/cuda_plugin.c
```diff
@@ -470,6 +471,18 @@ int cuda_plugin_resume_devices_late(int pid)
 }
 CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_devices_late)
 
+/**
+ * The cuda-checkpoint tool requires the libcuda library.
```
avagin, line 475: * does it really require it? If I remember right, cude-checkpoint initially was a static-linked binary. Did I miss something?

* The idea was to find out whether a host has a required hardware. For example, user can use one vm image for all their vm-s, but only some of them will have nvidia cards. This check will always pass even when a host doesn't have any nvidia cards.

rst0git: >does it really require it?

yes, the `cuda-checkpoint` tool is linked against `libcuda.so.1` and it fails if this library is not installed.

> This check will always pass even when a host doesn't have any nvidia cards.

We can also check the content of `/proc/driver/nvidia/gpus/*`:

```
# System without GPU
$ ls /proc/driver/nvidia/gpus/*
ls: cannot access '/proc/driver/nvidia/gpus/*': No such file or directory

# System with GPU
$ ls /proc/driver/nvidia/gpus/
0000:01:00.0

$ ls  /proc/driver/nvidia/gpus/
0000:82:00.0
```

rst0git: I've updated the pull request to check only for `/proc/driver/nvidia/gpus/*`.

plugins/cuda/cuda_plugin.c
```diff
@@ -470,6 +470,22 @@ int cuda_plugin_resume_devices_late(int pid)
 }
 CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__RESUME_DEVICES_LATE, cuda_plugin_resume_devices_late)
 
+static bool has_gpu()
+{
+	DIR *dir = opendir("/proc/driver/nvidia/gpus/");
```
avagin, line 475: nit: I think it is enough to check an existence of  /proc/driver/nvidia/gpus. I don't think that someone loads a module without having any nvidia card on the host.

plugins/cuda/cuda_plugin.c
```diff
@@ -481,10 +497,12 @@ int cuda_plugin_init(int stage)
 		}
 	}
 
-	if (!fault_injected(FI_PLUGIN_CUDA_FORCE_ENABLE) && access("/dev/nvidiactl", F_OK)) {
-		pr_info("/dev/nvidiactl doesn't exist. The CUDA plugin is disabled.\n");
-		plugin_disabled = true;
-		return 0;
+	if (!fault_injected(FI_PLUGIN_CUDA_FORCE_ENABLE)) {
```
avagin, line 500: nit: !fault_injected() && !has_gpu()

avagin:
LGTM. Thanks.

plugins/cuda/cuda_plugin.c
rst0git: Thanks, I also noticed that (if I understand correctly), setting `FI_PLUGIN_CUDA_FORCE_ENABLE` should skip the GPU device check. In other words, the condition should be:
```
!fault_injected(FI_PLUGIN_CUDA_FORCE_ENABLE) || !has_nvidia_gpu()
```
I've also updated the zdtm test to run with `--criu-plugin cuda` to test this change.

Issue #1589, opened 2021-08-23
CRIU Pre-Dump failing. It's showing a Waring.


Avijit009:
CRIU Pre-Dump failing. It's showing some Waring. I am not able to understand why this one is showing like this. Please help me out to get out of this problem.
after running  **"criu pre-dump --tree <pid> --images-dir <path-to-existing-directory-A>"** this command its showing
exact code I have executed **"criu pre-dump --tree 46292 --images-dir /home/mypc/Desktop/criu/checkpoint"**
**Warn  (compel/arch/x86/src/lib/infect.c:281): Will restore 46292 with interrupted system call**

Snorch:
Some times when criu dumps a process it can stop the process in the middle of syscall, e.g. in the middle of sys_nanosleep and criu just makes this process to receive -EINTR return value from this syscall after restore.

Normally all well-coded applications check return values of syscalls and for blocking syscalls check for -EINTR and restart the syscall when needed. But badly written code can break on it.

For instance if your code would just call sys_nanosleep(1min) once without any checking/restarting logic you can end up sleeping much less after CRIU migration. Though same would happen if you get some non-killing signal in the middle.

Please take a look on https://github.com/checkpoint-restore/criu/commit/dd71cca58a commit description for more info.

Avijit009:
@Snorch Thank You

adrianreber:
Closing as it seems solved.

PR #2512, opened 2024-11-04
compel: when pre-dump, handle ERESTART_RESTARTBLOCK as ERESTARTNOINTR

anatasluo:
When pre-dump, if thread is in syscall like sleep or poll, it will return -ERESTART_RESTARTBLOCK. In that case, if we set ax register to -EINTR, it will break the original syscall. Instead, we can restore original syscall.

### DIFF
--- a/compel/arch/aarch64/src/lib/infect.c
+++ b/compel/arch/aarch64/src/lib/infect.c
@@ -60,7 +60,7 @@ int sigreturn_prep_fpu_frame_plain(struct rt_sigframe *sigframe, struct rt_sigfr
 }
 
 int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct_t *fpsimd, save_regs_t save,
-			 void *arg, __maybe_unused unsigned long flags)
+			 void *arg, __maybe_unused unsigned long flags, __maybe_unused bool pre_dump)
 {
 	struct iovec iov;
 	int ret;
--- a/compel/arch/x86/src/lib/infect.c
+++ b/compel/arch/x86/src/lib/infect.c
@@ -398,7 +398,7 @@ static int corrupt_extregs(pid_t pid)
 }
 
 int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct_t *xs, save_regs_t save,
-			 void *arg, unsigned long flags)
+			 void *arg, unsigned long flags, bool pre_dump)
 {
 	int ret = -1;
 
@@ -415,8 +415,13 @@ int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct
 			set_user_reg(regs, ip, get_user_reg(regs, ip) - 2);
 			break;
 		case -ERESTART_RESTARTBLOCK:
-			pr_warn("Will restore %d with interrupted system call\n", pid);
-			set_user_reg(regs, ax, -EINTR);
+			if (pre_dump) {
+				set_user_reg(regs, ax, get_user_reg(regs, orig_ax));
+				set_user_reg(regs, ip, get_user_reg(regs, ip) - 2);
+			} else {
+				pr_warn("Will restore %d with interrupted system call\n", pid);
+				set_user_reg(regs, ax, -EINTR);
+			}
 			break;
 		}
 	}
--- a/compel/arch/arm/src/lib/infect.c
+++ b/compel/arch/arm/src/lib/infect.c
@@ -66,7 +66,7 @@ int sigreturn_prep_fpu_frame_plain(struct rt_sigframe *sigframe, struct rt_sigfr
 
 #define PTRACE_GETVFPREGS 27
 int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct_t *vfp, save_regs_t save,
-			 void *arg, __maybe_unused unsigned long flags)
+			 void *arg, __maybe_unused unsigned long flags, __maybe_unused bool pre_dump)
 {
 	int ret = -1;
 
--- a/compel/arch/loongarch64/src/lib/infect.c
+++ b/compel/arch/loongarch64/src/lib/infect.c
@@ -49,7 +49,7 @@ int sigreturn_prep_fpu_frame_plain(struct rt_sigframe *sigframe, struct rt_sigfr
 }
 
 int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct_t *ext_regs, save_regs_t save,
-			 void *arg, __maybe_unused unsigned long flags)
+			 void *arg, __maybe_unused unsigned long flags, __maybe_unused bool pre_dump)
 {
 	user_fpregs_struct_t tmp, *fpregs = ext_regs ? ext_regs : &tmp;
 	struct iovec iov;
--- a/compel/arch/mips/src/lib/infect.c
+++ b/compel/arch/mips/src/lib/infect.c
@@ -120,7 +120,7 @@ int sigreturn_prep_fpu_frame_plain(struct rt_sigframe *sigframe, struct rt_sigfr
 }
 
 int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct_t *xs, save_regs_t save,
-			 void *arg, __maybe_unused unsigned long flags)
+			 void *arg, __maybe_unused unsigned long flags, __maybe_unused bool pre_dump)
 {
 	int ret = -1;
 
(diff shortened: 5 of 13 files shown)


anatasluo:
I tested two cases for this patch:
1. sleep syscall
`#include <unistd.h>
#include <stdio.h>
#include <time.h>

const unsigned int interval = 60;

int main()
{
	time_t begin, end;
	printf("main thread works - %d for %u seconds \n", getpid(), interval);
	time(&begin);
	sleep(interval);
	time(&end);
	printf("it is ok - %f seconds \n", difftime(end, begin));
	return 0;
}
`

2. poll syscall
`#include <stdio.h>
#include <unistd.h>
#include <sys/poll.h>

#define TIMEOUT 25

int main (void)
{
	struct pollfd fds[2];
	int ret;

    printf("main thread works - %d \n", getpid());

	/* watch stdin for input */
	fds[0].fd = STDIN_FILENO;
	fds[0].events = POLLIN;

	/* watch stdout for ability to write */
	fds[1].fd = STDOUT_FILENO;
	fds[1].events = POLLOUT;

	ret = poll(fds, 1, TIMEOUT * 1000);

	if (ret == -1) {
		perror ("poll");
		return 1;
	}

	if (!ret) {
		printf ("%d seconds elapsed.\n", TIMEOUT);
		return 0;
	}

	if (fds[0].revents & POLLIN)
		printf ("stdin is readable\n");

	// if (fds[1].revents & POLLOUT)
	// 	printf ("stdout is writable\n");

	return 0;

}`

compel/arch/x86/src/lib/infect.c
```diff
@@ -415,8 +415,13 @@ int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct
 			set_user_reg(regs, ip, get_user_reg(regs, ip) - 2);
 			break;
 		case -ERESTART_RESTARTBLOCK:
-			pr_warn("Will restore %d with interrupted system call\n", pid);
-			set_user_reg(regs, ax, -EINTR);
+			if (pre_dump) {
```
rst0git, line 418: Would it make sense to use `ERESTART_RESTARTBLOCK` also when `criu dump` is used with `--leave-running`?

anatasluo: Yes, I think --leave-running should do the same. I have made some changes to the patch.
```
parasite_ctl = parasite_infect_seized(pid, item, &vmas, opts.final_state == TASK_ALIVE);
```

compel/arch/aarch64/src/lib/infect.c
```diff
@@ -60,7 +60,7 @@ int sigreturn_prep_fpu_frame_plain(struct rt_sigframe *sigframe, struct rt_sigfr
 }
 
 int compel_get_task_regs(pid_t pid, user_regs_struct_t *regs, user_fpregs_struct_t *fpsimd, save_regs_t save,
-			 void *arg, __maybe_unused unsigned long flags)
+			 void *arg, __maybe_unused unsigned long flags, __maybe_unused bool pre_dump)
```
rst0git, line 63: Since this argument is now used with both `criu pre-dump` and `criu dump --leave-running`, it might be better to use more generic name than `pre_dump`.

rst0git (commented):
It looks good to me.

@Snorch This change seems related to the discussion in https://github.com/checkpoint-restore/criu/issues/1589#issuecomment-904650828. Would you be able to review this pull request?

Snorch:
This does not sound correct.

We specifically set aside the -ERESTART_RESTARTBLOCK case, e.g. see commit message of dd71cca58ada1, it says:

>Just change this code to enforce -EINTR after restore, this is what
>we actually want until we teach criu to handle ERESTART_RESTARTBLOCK.

So we explicitly want this behavior of not restarting syscall in this case.

One of the reasons why we might want not to simply restart sleep syscall is that this approach does not consider the passed time before restart and may lead to wrong sleep times (possibly even endless sleep due to repeated dump).

In other words: Application which uses sleep (nanosleep) is in charge of restarting sleep with proper time adjustment, as sleep can be interrupted by signal even without CRIU involved.

anatasluo:
@Snorch Yes, I agree with you. As you said, interrupted time will not be considered as elcasped time. So it is good to make ERESTART_RESTARTBLOCK case return -EINTR explicitly. 
But for the case of pre-dump or --leave-running, there are some different things:
1. interrupted time is really short
2. for the poll syscall, interrupt directly may change the program behavior

Anyway, I think two strategies are all acceptable, just one advice.

avagin:
> interrupted time is really short, sleep itself is not that accurate. Since program could be affeched by things like schedule.

@anatasluo syscalls like poll and sleep have restart blocks in the kernel:
https://elixir.bootlin.com/linux/v6.11.6/source/include/linux/restart_block.h#L25

so they can be proper restarted only if the kernel doesn't drop their restart blocks. Otherwise, it can be the case, that a process called sleep(1h) and has been sleeping for 59 minutes, then we restart the syscall and it starts sleeping for another hour.

anatasluo:
@avagin This patch is used for pre-dump or --leave-running. In this case, kernel should not drop anything. And also, check the kernel code, the remanning time is saved in userspace:
```
int nanosleep_copyout(struct restart_block *restart, struct timespec64 *ts)
{
	switch(restart->nanosleep.type) {
#ifdef CONFIG_COMPAT_32BIT_TIME
	case TT_COMPAT:
		if (put_old_timespec32(ts, restart->nanosleep.compat_rmtp))
			return -EFAULT;
		break;
#endif
	case TT_NATIVE:
		if (put_timespec64(ts, restart->nanosleep.rmtp))
			return -EFAULT;
		break;
	default:
		BUG();
	}
	return -ERESTART_RESTARTBLOCK;
}
```

anatasluo:
> ## Test cases
> I tested two cases for this patch:
> 
> 1. sleep syscall
> 
> ```
> #include <unistd.h>
> #include <stdio.h>
> #include <time.h>
> 
> const unsigned int interval = 60;
> 
> int main()
> {
> 	time_t begin, end;
> 	printf("main thread works - %d for %u seconds \n", getpid(), interval);
> 	time(&begin);
> 	sleep(interval);
> 	time(&end);
> 	printf("it is ok - %f seconds \n", difftime(end, begin));
> 	return 0;
> }
> ```
> 
> 2. poll syscall
> 
> ```
> #include <stdio.h>
> #include <unistd.h>
> #include <sys/poll.h>
> 
> #define TIMEOUT 25
> 
> int main (void)
> {
> 	struct pollfd fds[2];
> 	int ret;
> 
>     printf("main thread works - %d \n", getpid());
> 
> 	/* watch stdin for input */
> 	fds[0].fd = STDIN_FILENO;
> 	fds[0].events = POLLIN;
> 
> 	/* watch stdout for ability to write */
> 	fds[1].fd = STDOUT_FILENO;
> 	fds[1].events = POLLOUT;
> 
> 	ret = poll(fds, 1, TIMEOUT * 1000);
> 
> 	if (ret == -1) {
> 		perror ("poll");
> 		return 1;
> 	}
> 
> 	if (!ret) {
> 		printf ("%d seconds elapsed.\n", TIMEOUT);
> 		return 0;
> 	}
> 
> 	if (fds[0].revents & POLLIN)
> 		printf ("stdin is readable\n");
> 
> 	// if (fds[1].revents & POLLOUT)
> 	// 	printf ("stdout is writable\n");
> 
> 	return 0;
> 
> }
> ```
> 
> ## Test command and result
> > sudo ./criu/criu pre-dump -vvv --shell-job -t
> 
> Without this patch, pre-dump of criu will break the sleep or poll syscall

@avagin Test cases have considered this situation.

anatasluo:
@avagin I found the problem here, restore_sigcontext would not affect this patch. Because this patch doesn't use restart_syscall. Instead, this patch restarts the original syscall. The problem here is that timeout for sleep syscall is read from memory while timeout for poll syscall is a value. So this patch works for sleep syscall, but not works for poll syscall.
```
SYSCALL_DEFINE3(poll, struct pollfd __user *, ufds, unsigned int, nfds,
		int, timeout_msecs)
{
	struct timespec64 end_time, *to = NULL;
	int ret;

	if (timeout_msecs >= 0) {
		to = &end_time;
		poll_select_set_timeout(to, timeout_msecs / MSEC_PER_SEC,
			NSEC_PER_MSEC * (timeout_msecs % MSEC_PER_SEC));
	}

	ret = do_sys_poll(ufds, nfds, to);

	if (ret == -ERESTARTNOHAND) {
		struct restart_block *restart_block;

		restart_block = &current->restart_block;
		restart_block->poll.ufds = ufds;
		restart_block->poll.nfds = nfds;

		if (timeout_msecs >= 0) {
			restart_block->poll.tv_sec = end_time.tv_sec;
			restart_block->poll.tv_nsec = end_time.tv_nsec;
			restart_block->poll.has_timeout = 1;
		} else
			restart_block->poll.has_timeout = 0;

		ret = set_restart_fn(restart_block, do_restart_poll);
	}
	return ret;
}
```

anatasluo:
Since this patch still has problems to fix, i will close it now.

Issue #2508, opened 2024-11-01
seccomp: Can't find entry on tid_real


rst0git:
**Steps to reproduce the issue:**
1. Create a Pod with the following manifest:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: counter
spec:
  securityContext:
    runAsNonRoot: true
    runAsUser: 1000
    fsGroup: 101
    seccompProfile:
      type: RuntimeDefault
  containers:
    - name: counter
      image: busybox:latest
      command:
        - "/bin/sh"
        - "-c"
        - "i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done"
      securityContext:
        allowPrivilegeEscalation: false
        capabilities:
          drop:
            - ALL
      resources:
        requests:
          memory: "1Gi"
          cpu: "1"
        limits:
          memory: "1Gi"
          cpu: "1"
```
2. Create a container checkpoint using the [kubelet Checkpoint API](https://kubernetes.io/docs/reference/node/kubelet-checkpoint-api/)

**Describe the results you received:**
CRIU fails with the following error:

```
(00.122028) Collecting pidns 9/501089
(00.122037) Error (criu/seccomp.c:61): seccomp: Can't find entry on tid_real 501089
(00.122039) Error (criu/seccomp.c:278): seccomp: Can't collect filter on leader tid_real 501089
(00.122098) net: Unlock network
(00.122100) Running network-unlock scripts
(00.122101) 	RPC
(00.196082) cuda_plugin: finished cuda_plugin stage 0 err -1
(00.196155) Unfreezing tasks into 1
(00.196157) 	Unseizing 501089 into 1
(00.196165) Error (compel/src/lib/infect.c:418): Unable to detach from 501089: No such process
(00.196174) Error (criu/cr-dump.c:2111): Dumping FAILED
```
[dump.log](https://github.com/user-attachments/files/17603061/criu-dump.log.txt)

PR #2513, opened 2024-11-04
seize: fix error handling for check_freezer_cgroup

rst0git:
When `check_freezer_cgroup()` has non-zero return value, `goto err` calls `return ret`. However, the value of `ret` has been set to `0` in the lines above and CRIU does not handle the error properly.

This problem is related to https://github.com/checkpoint-restore/criu/issues/2508

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -1009,7 +1009,7 @@ static int cgroup_version(void)
 int collect_pstree(void)
 {
 	pid_t pid = root_item->pid->real;
-	int ret = -1;
+	int ret, exit_code = -1;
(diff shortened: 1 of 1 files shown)


criu/seize.c
```diff
@@ -1026,6 +1026,7 @@ int collect_pstree(void)
 	if (ret < 0 && ret != -ENOTSUP) {
 		goto err;
 	}
+	ret = -1;
```
avagin, line 1029: could you rework it to the exit_code pattern?

avagin:
Merged. Thanks a lot.

PR #2514, opened 2024-11-04
seize: enable support for frozen containers

rst0git:
Container runtimes like CRI-O and containerd utilize the freezer cgroup to create a consistent snapshot of container rootfs changes. In this case, the container is frozen before invoking CRIU. Once CRIU successfully completes, a copy of the container rootfs diff is saved, and then the container is unfrozen. To enable GPU checkpointing support with these runtimes, we need to unfreeze the cgroup and restore it to its original state at the end.

When the CUDA plugin is installed, container checkpointing with Kubernetes fails, even for containers that don't use GPUs. This patch aims to resolve this issue.

Fixes: #2508


criu/seize.c
```diff
@@ -517,17 +518,22 @@ static int check_freezer_cgroup(void)
 		return -1;
 
 	state = get_freezer_state(fd);
-	close(fd);
 	if (state == FREEZER_ERROR) {
-		return -1;
+		goto err;
 	}
 
+	origin_freezer_state = state == FREEZING ? FROZEN : state;
```
avagin, line 525: freeze_processes() initializes the origin_freezer_state again, but we have to be sure that it contains the state that has been initially.

adrianreber:
Doesn't this break the expectations of the container engines. You wrote they freeze the container to avoid changes to the container file-system. Does the container now continue to run with your change?

criu/seize.c
rst0git: > freeze_processes() initializes the origin_freezer_state again

When the CUDA plugin is used and `freeze_cgroup_disabled` is set to `true`, CRIU calls `check_freezer_cgroup()` instead of `freeze_processes()`:
```C
	if (opts.freeze_cgroup && !freeze_cgroup_disabled) {
		if (freeze_processes())
			goto err;
	} else {
		if (opts.freeze_cgroup && check_freezer_cgroup())
			goto err;
		if (compel_interrupt_task(pid)) {
			set_cr_errno(ESRCH);
			goto err;
		}
	}

```

rst0git:
>Does the container now continue to run with your change?

No, we use process seizing without freezer cgroup during checkpointing (see https://github.com/checkpoint-restore/criu/pull/2475 and https://github.com/checkpoint-restore/criu/pull/2470).

criu/seize.c
avagin: It is even more sophisticated. Pls rename freeze_cgroup_disabled and check_freezer_cgroup to something more suitable for the new reality.

rst0git: Thank you Andrei! I've updated the pull request to rename `freeze_cgroup_disabled` to `compel_interrupt_only_mode` and `check_freezer_cgroup` to `prepare_freezer_for_interrupt_only_mode`.

test/jenkins/criu-fault.sh
```diff
@@ -40,7 +40,7 @@ fi
 # also check for the main thread corruption
 ./test/zdtm.py run -t zdtm/static/fpu00 --fault 134 -f h --norst || fail
 
-# check dont_use_freeze_cgroup
+# check set_compel_interrupt_only_mode
 ./test/zdtm.py run -t zdtm/static/env00 --freezecg zdtm:t --fault 137
```
avagin, line 44: should we add one more case when a test cgroup is frozen?

avagin:
> Doesn't this break the expectations of the container engines. You wrote they freeze the container to avoid changes to the container file-system. Does the container now continue to run with your change?

Strictly speaking, this expectation was not right even before this change. CRIU does file system changes while dumping processes. For example, it creates ghost files.

I think the right expectation here is that file systems are not changed after dumping processes and this statement isn't affected by this change.

avagin:
LGTM. Thanks.

PR #2517, opened 2024-11-12
cuda: enable checkpoint support for paused tasks

rst0git:
If a CUDA process is already in a "locked" or "checkpointed" state during `criu dump`, the CUDA plugin fails with an error because it attempts an unnecessary "lock" action using the `cuda-checkpoint` tool. This pull request extends the CUDA plugin to handle such case by first checking the original state of the CUDA process and skipping unnecessary "lock" and "checkpoint" actions if the process was already locked or checkpointed before CRIU was invoked.

### DIFF
--- a/plugins/cuda/cuda_plugin.c
+++ b/plugins/cuda/cuda_plugin.c
@@ -26,6 +26,13 @@
 #define ACTION_RESTORE	  "restore"
 #define ACTION_UNLOCK	  "unlock"
 
+typedef enum {
+	CUDA_TASK_RUNNING = 0,
+	CUDA_TASK_LOCKED,
+	CUDA_TASK_CHECKPOINTED,
+	CUDA_TASK_UNKNOWN = -1
+} cuda_task_state_t;
+
 #define CUDA_CKPT_BUF_SIZE (128)
 
 #ifdef LOG_PREFIX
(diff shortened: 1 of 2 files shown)


plugins/cuda/cuda_plugin.c
```diff
@@ -397,7 +460,7 @@ int cuda_plugin_pause_devices(int pid)
 		return -1;
 	}
 
-	if (add_pid_to_buf(&cuda_pids, pid)) {
+	if (add_pid_to_buf(&cuda_pids, restore_tid, CUDA_TASK_RUNNING)) {
```
jesus-ramos, line 463: Shouldn't we be tracking the PID and not the restore thread id here?

jesus-ramos (approved):
LGTM

plugins/cuda/cuda_plugin.c
rst0git: You're right, tracking the PID would indeed be more reliable here. Thank you for catching that!

rst0git:
> Is there any real use-case for that?

The use-case is similar to https://github.com/checkpoint-restore/criu/pull/2514 -- the CUDA tasks may be in a "locked" or "checkpointed" state before `criu dump` is invoked to ensure consistent checkpoint/restore, particularly in distributed model training where multiple containers are running across different cluster nodes.

rst0git:
@avagin I've updated the commit message with a brief description of the use-cases for this functionality.

PR #2518, opened 2024-11-17
RISC-V port fixes (part I)

mihalicyn:
Some follow-up fixes for CRIU [port on RISC-V](https://github.com/checkpoint-restore/criu/pull/2234).

cc @ancientmodern @Cryolitia @felicitia


mihalicyn:
Earlier I [mentioned](https://github.com/checkpoint-restore/criu/pull/2234#issuecomment-2433504111) that I had experienced not only `TASK_SIZE`-related issue but also crashes on later stages. But after I've got an upstream 6.12-rc7 kernel running on my board it started to work. Most likely I was experiencing some kernel bug and also it was kernel without https://github.com/torvalds/linux/commit/ce4f78f1b53d3327fbd32764aa333bf05fb68818 fix which can also affect my earlier testing.

compel/arch/riscv64/src/lib/infect.c
```diff
@@ -181,20 +181,28 @@ int arch_fetch_sas(struct parasite_ctl *ctl, struct rt_sigframe *s)
...
- * - 0x800000000000 ( 128TB) for RV64 using SV48 mmu
+ * -        0x9fc00000	(~2.5GB) for RV32.
+ * -      0x4000000000	( 256GB) for RV64 using SV39 mmu
+ * -    0x800000000000	( 128TB) for RV64 using SV48 mmu
+ * - 0x100000000000000	(  64PB) for RV64 using SV57 mmu
  *
  * Note that PGDIR_SIZE must evenly divide TASK_SIZE since "RISC-V
  * Instruction Set Manual Volume II: Privileged Architecture" states that
  * "load and store effective addresses, which are 64bits, must have bits
  * 63–48 all equal to bit 47, or else a page-fault exception will occur."
```
ancientmodern, line 192: This comment was copied from kernel `arch/riscv/include/asm/pgtable.h`, where it explains the rationale behind `#define TASK_SIZE_64 (PGDIR_SIZE * PTRS_PER_PGD / 2)`. I think we can either remove or replace this comment as it looks out of place in our context :)

mihalicyn: Hi Haorong,

thanks, fixed!

ancientmodern (approved):
LGTM, thanks!

compel/arch/riscv64/src/lib/infect.c
```diff
@@ -181,20 +181,22 @@ int arch_fetch_sas(struct parasite_ctl *ctl, struct rt_sigframe *s)
...
+#define TASK_SIZE_MIN (1UL << 38)
+#define TASK_SIZE_MAX (1UL << 56)
 
 unsigned long compel_task_size(void)
 {
-	return TASK_SIZE;
+	unsigned long task_size;
+
+	for (task_size = TASK_SIZE_MIN; task_size < TASK_SIZE_MAX; task_size <<= 1)
+		if (munmap((void *)task_size, page_size()))
```
avagin, line 197: I merged it because I was thinking that I'd reviewed it already.

I think this munamp can unmap something useful and trigger sigsegv.

mihalicyn: Hmm, actually, we are using the same approach for other architectures:
https://github.com/checkpoint-restore/criu/blob/1452c76f6541721a77e886a41c6ee63bb5571c0c/compel/arch/arm/src/lib/infect.c#L184
https://github.com/checkpoint-restore/criu/blob/1452c76f6541721a77e886a41c6ee63bb5571c0c/compel/arch/ppc64/src/lib/infect.c#L507

But that's an interesting point and probably I should try to find a cleaner way to determine a task size across all architectures we have then.

But taking into account that it's not a new and dirty hack I introduced but rather reusing an existing and time-proven method I think it's good enough to be there.

avagin: We can do something like this: https://cs.opensource.google/gvisor/gvisor/+/master:pkg/abi/linux/mm.go;l=142?q=tasksize&ss=gvisor%2Fgvisor

PR #2527, opened 2024-11-22
AMDGPU: add parallel restore of BO content to accelerate restore

wweewrwer:
**TL;DR:**

This pull request extends CRIU to support parallel restore of AMDGPU buffer object content alongside other restore operations to accelerate the restoration.

**The target issue:** 

In the current restore procedure of AMDGPU applications, the content of the AMDGPU buffer object (BO) is restored synchronously in `CR_PLUGIN_HOOK__RESTORE_EXT_FILE`. This procedure usually takes a significant amount of time, and during this time the target process cannot perform any other restore operations. However, this restoration has no logical dependencies with other restore operations. Parallelizing this part with other restore operations can speed up the restoration.

**The parallel restore approach in this PR:** 

The core idea of these patch series is to offload the restore of the BO content from the target process to the main CRIU process (the main CRIU process refers to the parent process, and the target process refers to the child process created during the fork). To achieve this, we introduce a new hook, `CR_PLUGIN_HOOK__RESTORE_ASYNCHRONOUS`, in the main CRIU process. For the AMDGPU plugin, the target process will no longer restore BO contents in `CR_PLUGIN_HOOK__RESTORE_EXT_FILE` and just send the relevant BOs to the main CRIU process. the main CRIU process will receive the corresponding BOs in `CR_PLUGIN_HOOK__RESTORE_ASYNCHRONOUS` and begin the restoration. Meanwhile, the target process can continue with other parts of the restoration without being blocked by the BO content restoration. The full design of the idea can also be referred with the ACM SoCC'24 paper: [On-demand and Parallel Checkpoint/Restore for GPU Applications](https://dl.acm.org/doi/10.1145/3698038.3698510).

**Tests:**

We evaluated the performance according to the following settings. The results show that parallel restore can speed up by 34.3% when images cached in the page cache, and 7.6% when restoring from disk.

Results:

|                    | From disk | From page cache |
| ------------------ | --------- | --------------- |
| Sequential restore | 1728ms    | 254ms           |
| Parallel restore   | 1596ms    | 167ms           |
| Speed up           | 7.6%      | 34.3%           |

Settings:

CPU: Intel(R) Core(TM) i7-10700 CPU @ 2.90GHz

Memory: DDR4, 2x8GB

GPU: AMD MI50

Disk: 512GB, Samsung SSD 860

Docker image: rocm/pytorch:rocm5.6_ubuntu20.04_py3.8_pytorch_1.12.1

Example program：

`example.py`: a ResNet18 application. Enter 'y' to exit, or press any other key to perform inference.

```
import time
import os
import sys
import torch
import torchvision.models as models
import torchvision.transforms as transforms
torch.set_grad_enabled(False)

device = "cuda:0"

model = models.resnet18(weights='DEFAULT')
model = model.to(device)
model.eval()

batch_size = 1
channels = 3
height = 224
width = 224
input_tensor = torch.randn(batch_size, channels, height, width)
preprocess = transforms.Compose([
    transforms.Normalize(mean=[0.485, 0.456, 0.406], std=[0.229, 0.224, 0.225])
])
input_tensor = preprocess(input_tensor)

while input()!="y":
    st = time.time()
    input_tensor = input_tensor.to(device)
    output = model(input_tensor)
    output = output.to("cpu")
    _, predicted_idx = torch.max(output, 1)
    torch.cuda.synchronize()
    ed = time.time()
    print("test time:",ed-st)
    sys.stdout.flush()
```

Steps：

0. Install CRIU

   Follow the standard CRIU installation process. Ensure you set the environment variable `CRIU_LIBS_DIR` to the `plugins/amdgpu` path.

1. Dump checkpoint image

   ```
   #In one shell
   python3 example.py
   #In another shell
   mkdir -p /tmp/criu-dump
   criu dump -t $(pgrep python3) -D /tmp/criu-dump -j --file-locks
   ```

2. Restore from disk

   Test for sequential restore:

   ```
   #Clear page cache
   sync; sudo sh -c "echo 3 > /proc/sys/vm/drop_caches" 
   criu restore -D /tmp/criu-dump -j --file-locks
   cat stats-restore | crit decode --pretty | grep restore_time
   ```

   Test for parallel restore:

   ```
   sync; sudo sh -c "echo 3 > /proc/sys/vm/drop_caches" 
   criu restore -D /tmp/criu-dump -j --file-locks --parallel
   cat stats-restore | crit decode --pretty | grep restore_time
   ```

3. Restore from page cache

   Install vmtouch for caching images:

   ```
   sudo apt install vmtouch
   ```

   Test:

   ```
   sync; sudo sh -c "echo 3 > /proc/sys/vm/drop_caches" 
   #Cache image in memory
   vmtouch -l criu-dump
   #Warm up environment 
   criu restore -D /tmp/criu-dump -j --file-locks
   #Begin to Test
   criu restore -D /tmp/criu-dump -j --file-locks
   cat stats-restore | crit decode --pretty | grep restore_time
   criu restore -D /tmp/criu-dump -j --file-locks --parallel
   cat stats-restore | crit decode --pretty | grep restore_time
   ```

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2365,41 +2365,47 @@ int cr_restore_tasks(void)
 		return 1;
 
 	if (check_img_inventory(/* restore = */ true) < 0)
-		goto err;
-
-	if (cr_plugin_init(CR_PLUGIN_STAGE__RESTORE))
 		return -1;
 
 	if (init_stats(RESTORE_STATS))
-		goto err;
+		return -1;
 
 	if (lsm_check_opts())
-		goto err;
+		return -1;
 
 	timing_start(TIME_RESTORE);
 
 	if (cpu_init() < 0)
-		goto err;
+		return -1;
 
 	if (vdso_init_restore())
-		goto err;
+		return -1;
 
 	if (tty_init_restore())
-		goto err;
+		return -1;
 
 	if (opts.cpu_cap & CPU_CAP_IMAGE) {
 		if (cpu_validate_cpuinfo())
-			goto err;
+			return -1;
 	}
 
 	if (prepare_task_entries() < 0)
-		goto err;
+		return -1;
 
 	if (prepare_pstree() < 0)
-		goto err;
+		return -1;
 
 	if (fdstore_init())
-		goto err;
+		return -1;
+
+	/*
+	 * For the AMDGPU plugin, its parallel restore feature needs to use fdstore to store
+	 * its socket file descriptor. This allows the main process and the target process to
+	 * communicate with each other through this file descriptor. Therefore, cr_plugin_init
+	 * must be initialized after fdstore_init.
+	 */
+	if (cr_plugin_init(CR_PLUGIN_STAGE__RESTORE))
+		return -1;
 
 	if (inherit_fd_move_to_fdstore())
 		goto err;
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2131,6 +2131,9 @@ static int restore_root_task(struct pstree_item *init)
 	__restore_switch_stage(CR_STATE_FORKING);
 
 skip_ns_bouncing:
+	ret = run_plugins(POST_FORKING);
+	if (ret < 0 && ret != -ENOTSUP)
+		goto out_kill;
 
 	ret = restore_wait_inprogress_tasks();
 	if (ret < 0)
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -182,7 +182,7 @@ void free_pstree(struct pstree_item *root_item)
 	struct pstree_item *item = root_item, *parent;
 
 	while (item) {
-		if (!list_empty(&item->children)) {
+		if (has_children(item)) {
 			item = list_first_entry(&item->children, struct pstree_item, sibling);
 			continue;
 		}
@@ -244,10 +244,15 @@ int init_pstree_helper(struct pstree_item *ret)
 	return 0;
 }
 
+bool has_children(struct pstree_item *item)
+{
+	return !list_empty(&item->children);
+}
+
 /* Deep first search on children */
 struct pstree_item *pstree_item_next(struct pstree_item *item)
 {
-	if (!list_empty(&item->children))
+	if (has_children(item))
 		return list_first_entry(&item->children, struct pstree_item, sibling);
 
 	while (item->parent) {
--- a/plugins/amdgpu/README.md
+++ b/plugins/amdgpu/README.md
@@ -3,7 +3,8 @@ Supporting ROCm with CRIU
 
 _Felix Kuehling <Felix.Kuehling@amd.com>_<br>
 _Rajneesh Bardwaj <Rajneesh.Bhardwaj@amd.com>_<br>
-_David Yat Sin <David.YatSin@amd.com>_
+_David Yat Sin <David.YatSin@amd.com>_<br>
+_Yanning Yang <yangyanning@sjtu.edu.cn>_
 
 # Introduction
 
@@ -223,6 +224,26 @@ to resume execution on the GPUs.
 
 *This new plugin is enabled by the new hook `__RESUME_DEVICES_LATE` in our RFC
 patch series.*
+
+## Restoring BO content in parallel
+
+Restoring the BO content is an important part in the restore of GPU state and
+usually takes a significant amount of time. A possible location for this
+procedure is the `cr_plugin_restore_file` hook. However, restoring in this hook
+blocks the target process from performing other restore operations, which
+hinders further optimization of the restore process.
+
+Therefore, a new plugin hook that runs in the master restore process is
+introduced, and it interacts with the `cr_plugin_restore_file` hook to complete
+the restore of BO content. Specifically, the target process only needs to send
+the relevant BOs to the master restore process, while this new hook handles all
+the restore of buffer objects. Through this method, during the restore of the BO
+content, the target process can perform other restore operations, thus
+accelerating the restore procedure. This is an implementation of the gCROP
+method proposed in the ACM SoCC'24 paper: [On-demand and Parallel
+Checkpoint/Restore for GPU Applications](https://dl.acm.org/doi/10.1145/3698038.3698510).
+
+*This optimization technique is enabled by the `__POST_FORKING` hook.*
 
 ## Other CRIU changes
 
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -28,11 +28,13 @@
 #include "xmalloc.h"
 #include "criu-log.h"
 #include "files.h"
+#include "pstree.h"
 
 #include "common/list.h"
 #include "amdgpu_plugin_drm.h"
 #include "amdgpu_plugin_util.h"
 #include "amdgpu_plugin_topology.h"
+#include "amdgpu_socket_utils.h"
 
 #include "img-streamer.h"
 #include "image.h"
@@ -64,6 +66,18 @@ bool plugin_added_to_inventory = false;
 
 bool plugin_disabled = false;
 
+/*
+ * In the case of a single process (common case), this optimization can effectively
+ * reduce the restore latency with parallel restore. In the case of multiple processes,
+ * states are already restored in parallel within different processes. Therefore, this
+ * optimization does not introduce further improvement and will be disabled by default
+ * in this case. The flag, parallel_disabled, is used to control whether the
+ * optimization is enabled or disabled.
+ */
+bool parallel_disabled = false;
+
+pthread_t parallel_thread = 0;
+int parallel_thread_result = 0;
 /**************************************************************************************************/
 
 /* Call ioctl, restarting if it is interrupted */
@@ -351,6 +365,15 @@ int amdgpu_plugin_init(int stage)
 	maps_init(&restore_maps);
 
 	if (stage == CR_PLUGIN_STAGE__RESTORE) {
+		if (has_children(root_item)) {
+			pr_info("Parallel restore disabled\n");
+			parallel_disabled = true;
+		} else {
+			if (install_parallel_sock() < 0) {
+				pr_err("Failed to install parallel socket\n");
+				return -1;
+			}
+		}
 		/* Default Values */
 		kfd_fw_version_check = true;
 		kfd_sdma_fw_version_check = true;
@@ -1439,14 +1462,9 @@ static int restore_bos(struct kfd_ioctl_criu_args *args, CriuKfd *e)
 
 static int restore_bo_data(int id, struct kfd_criu_bo_bucket *bo_buckets, CriuKfd *e)
 {
-	struct thread_data *thread_datas;
+	struct thread_data *thread_datas = NULL;
 	int thread_i, ret = 0;
-
-	thread_datas = xzalloc(sizeof(*thread_datas) * e->num_of_gpus);
-	if (!thread_datas) {
-		ret = -ENOMEM;
-		goto exit;
-	}
+	int offset = 0;
 
 	for (int i = 0; i < e->num_of_bos; i++) {
 		struct kfd_criu_bo_bucket *bo_bucket = &bo_buckets[i];
@@ -1489,56 +1507,101 @@ static int restore_bo_data(int id, struct kfd_criu_bo_bucket *bo_buckets, CriuKf
 		}
 	}
 
-	thread_i = 0;
-	for (int i = 0; i < e->num_of_gpus + e->num_of_cpus; i++) {
-		struct tp_node *dev;
-		int ret_thread = 0;
-		uint32_t target_gpu_id;
+	if (!parallel_disabled) {
+		parallel_restore_cmd restore_cmd;
+		pr_info("Begin to send parallel restore cmd\n");
+		ret = init_parallel_restore_cmd(e->num_of_bos, id, e->num_of_gpus, &restore_cmd);
+		if (ret)
+			goto exit_parallel;
 
-		if (!e->device_entries[i]->gpu_id)
-			continue;
+		for (int i = 0; i < e->num_of_gpus + e->num_of_cpus; i++) {
+			uint32_t target_gpu_id;
+			struct tp_node *dev;
 
-		/* e->device_entries[i]->gpu_id is user_gpu_id, target_gpu_id is actual_gpu_id */
-		target_gpu_id = maps_get_dest_gpu(&restore_maps, e->device_entries[i]->gpu_id);
+			if (!e->device_entries[i]->gpu_id)
+				continue;
 
-		/* We need the fd for actual_gpu_id */
-		dev = sys_get_node_by_gpu_id(&dest_topology, target_gpu_id);
-		if (!dev) {
-			pr_err("Failed to find node with gpu_id:0x%04x\n", target_gpu_id);
-			ret = -ENODEV;
+			target_gpu_id = maps_get_dest_gpu(&restore_maps, e->device_entries[i]->gpu_id);
+			dev = sys_get_node_by_gpu_id(&dest_topology, target_gpu_id);
+			if (!dev) {
+				pr_err("Failed to find node with gpu_id:0x%04x\n", target_gpu_id);
+				ret = -ENODEV;
+				goto exit_parallel;
+			}
+			parallel_restore_gpu_id_add(e->device_entries[i]->gpu_id, dev->drm_render_minor, &restore_cmd);
+
+			for (int j = 0; j < e->num_of_bos; j++) {
+				if (bo_buckets[j].gpu_id != e->device_entries[i]->gpu_id)
+					continue;
+				if (bo_buckets[j].alloc_flags &
+				    (KFD_IOC_ALLOC_MEM_FLAGS_VRAM | KFD_IOC_ALLOC_MEM_FLAGS_GTT)) {
+					parallel_restore_bo_add(bo_buckets[j].dmabuf_fd, bo_buckets[j].gpu_id,
+								bo_buckets[j].size, offset, &restore_cmd);
+					offset += bo_buckets[j].size;
+				}
+			}
+		}
+		ret = send_parallel_restore_cmd(&restore_cmd);
+exit_parallel:
+		free_parallel_restore_cmd(&restore_cmd);
+	} else {
+		thread_datas = xzalloc(sizeof(*thread_datas) * e->num_of_gpus);
+		if (!thread_datas) {
+			ret = -ENOMEM;
 			goto exit;
 		}
 
-		thread_datas[thread_i].id = id;
-		thread_datas[thread_i].gpu_id = e->device_entries[i]->gpu_id;
-		thread_datas[thread_i].bo_buckets = bo_buckets;
-		thread_datas[thread_i].bo_entries = e->bo_entries;
-		thread_datas[thread_i].pid = e->pid;
-		thread_datas[thread_i].num_of_bos = e->num_of_bos;
+		thread_i = 0;
+		for (int i = 0; i < e->num_of_gpus + e->num_of_cpus; i++) {
+			struct tp_node *dev;
+			int ret_thread = 0;
+			uint32_t target_gpu_id;
 
-		thread_datas[thread_i].drm_fd = node_get_drm_render_device(dev);
-		if (thread_datas[thread_i].drm_fd < 0) {
-			ret = -thread_datas[thread_i].drm_fd;
-			goto exit;
+			if (!e->device_entries[i]->gpu_id)
+				continue;
+
+			/* e->device_entries[i]->gpu_id is user_gpu_id, target_gpu_id is actual_gpu_id */
+			target_gpu_id = maps_get_dest_gpu(&restore_maps, e->device_entries[i]->gpu_id);
+
+			/* We need the fd for actual_gpu_id */
+			dev = sys_get_node_by_gpu_id(&dest_topology, target_gpu_id);
+			if (!dev) {
+				pr_err("Failed to find node with gpu_id:0x%04x\n", target_gpu_id);
+				ret = -ENODEV;
+				goto exit;
+			}
+
+			thread_datas[thread_i].id = id;
+			thread_datas[thread_i].gpu_id = e->device_entries[i]->gpu_id;
+			thread_datas[thread_i].bo_buckets = bo_buckets;
+			thread_datas[thread_i].bo_entries = e->bo_entries;
+			thread_datas[thread_i].pid = e->pid;
+			thread_datas[thread_i].num_of_bos = e->num_of_bos;
+
+			thread_datas[thread_i].drm_fd = node_get_drm_render_device(dev);
+			if (thread_datas[thread_i].drm_fd < 0) {
+				ret = -thread_datas[thread_i].drm_fd;
+				goto exit;
+			}
+
+			ret_thread = pthread_create(&thread_datas[thread_i].thread, NULL, restore_bo_contents,
+						    (void *)&thread_datas[thread_i]);
+			if (ret_thread) {
+				pr_err("Failed to create thread[%i] ret:%d\n", thread_i, ret_thread);
+				ret = -ret_thread;
+				goto exit;
+			}
+			thread_i++;
 		}
 
-		ret_thread = pthread_create(&thread_datas[thread_i].thread, NULL, restore_bo_contents,
-					    (void *)&thread_datas[thread_i]);
-		if (ret_thread) {
-			pr_err("Failed to create thread[%i] ret:%d\n", thread_i, ret_thread);
-			ret = -ret_thread;
-			goto exit;
-		}
-		thread_i++;
-	}
+		for (int i = 0; i < e->num_of_gpus; i++) {
+			pthread_join(thread_datas[i].thread, NULL);
+			pr_info("Thread[0x%x] finished ret:%d\n", thread_datas[i].gpu_id, thread_datas[i].ret);
 
-	for (int i = 0; i < e->num_of_gpus; i++) {
-		pthread_join(thread_datas[i].thread, NULL);
-		pr_info("Thread[0x%x] finished ret:%d\n", thread_datas[i].gpu_id, thread_datas[i].ret);
-
-		if (thread_datas[i].ret) {
-			ret = thread_datas[i].ret;
-			goto exit;
+			if (thread_datas[i].ret) {
+				ret = thread_datas[i].ret;
+				goto exit;
+			}
 		}
 	}
 exit:
@@ -1546,8 +1609,8 @@ exit:
 		if (bo_buckets[i].dmabuf_fd != KFD_INVALID_FD)
 			close(bo_buckets[i].dmabuf_fd);
 	}
-
-	xfree(thread_datas);
+	if (thread_datas)
+		xfree(thread_datas);
 	return ret;
 }
 
@@ -1836,6 +1899,24 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
 	if (plugin_disabled)
 		return -ENOTSUP;
 
+	if (!parallel_disabled) {
+		pr_info("Close parallel restore server\n");
+		if (close_parallel_restore_server()) {
+			pr_err("Close parallel restore server fail\n");
+			return -1;
+		}
+
+		exit_code = pthread_join(parallel_thread, NULL);
+		if (exit_code) {
+			pr_err("Failed to join parallel thread ret:%d\n", exit_code);
+			return -1;
+		}
+		if (parallel_thread_result) {
+			pr_err("Parallel restore fail\n");
+			return parallel_thread_result;
+		}
+	}
+
 	pr_info("Inside %s for target pid = %d\n", __func__, target_pid);
 
 	fd = open(AMDGPU_KFD_DEVICE, O_RDWR | O_CLOEXEC);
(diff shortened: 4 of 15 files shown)


criu/crtools.c
```diff
@@ -422,6 +422,8 @@ int main(int argc, char *argv[], char *envp[])
 	       "                        can be 'nftables' or 'iptables' (default).\n"
 	       "  --unprivileged        accept limitations when running as non-root\n"
 	       "                        consult documentation for further details\n"
+	       "  --parallel            enable parallel restore of AMDGPU buffer object content\n"
+	       "                        with other restore operations\n"
```
rst0git, line 426: @wweewrwer Is there any reason why parallel restore of buffer objects should not be enabled by default?

wweewrwer: To the best of my ability, I can't think of any scenarios where parallel restore wouldn't be applicable. I think the default method is simpler to implement, while the parallel method is more efficient. The state of processes restored through parallel restore has no differences from those restored using the default method. If most of the criu team prefer to enable the optimization by default, we can make the change :)

rst0git: I tested this functionality with Podman, but it causes container restore to hang:
```
(00.000986) Forking task with 1 pid (flags 0x6c028000)
(00.000987) Creating process using clone3()
(00.001105) PID: real 389586 virt 1
(00.001120) Run restore asynchronous hook from criu master for external devices
(00.001121) plugin: `amdgpu_plugin' hook 12 -> 0x7fcb95a16f83
(00.001122) amdgpu_plugin: Begin to recv parallel restore cmd
(00.001132) Parallel restore: begin to recv cmd_head
(00.001140)      1: Found id extRootNetNS (fd 14) in inherit fd list
(00.001219)      1: timens: monotonic -602 193779164
(00.001223)      1: timens: boottime -602 193774845
```
Is it intended to work with containers?

criu/cr-restore.c
```diff
@@ -2051,6 +2051,17 @@ static int restore_root_task(struct pstree_item *init)
...
 
+	pr_info("Run restore asynchronous hook from criu master for external devices\n");
+	ret = run_plugins(RESTORE_ASYNCHRONOUS);
+	/*
+	 * This may not really be an error. Only certain plugin hooks
+	 * (if available) will return success such as amdgpu_plugin.
+	 * Most of the times, it'll be -ENOTSUP and in few cases, it
+	 * might actually be a true error code but that won't be 
+	 * -ENOTSUP.
+	 */
```
avagin, line 2062: pls don't add comments that don't add any new information.

criu/cr-restore.c
```diff
@@ -2051,6 +2051,17 @@ static int restore_root_task(struct pstree_item *init)
 	if (ret < 0)
 		goto out;
 
+	pr_info("Run restore asynchronous hook from criu master for external devices\n");
+	ret = run_plugins(RESTORE_ASYNCHRONOUS);
```
avagin, line 2055: I think it isn't a good name for this hook. It can be a bit more descriptive about the exact moment when it is called.

criu/crtools.c
avagin: It can be enabled by default.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1897,142 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+	FILE *bo_contents_fp = NULL;
+	void *buffer = NULL;
+	int ret = 0;
+
+	pr_info("Begin to recv parallel restore cmd\n");
+	ret = recv_parallel_restore_cmd();
+	if (ret)
+		return ret;
+
+	int *vis = (int *)malloc(restore_cmd.cmd_head.entry_num * sizeof(int));
```
avagin, line 1977: pls use xmalloc and check its return code.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1897,142 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+	memset(vis, 0, restore_cmd.cmd_head.entry_num * sizeof(int));
```
avagin, line 1978: xzalloc return zeroed memory region.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1897,142 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+			ret = -ENOMEM;
+			goto err_sdma;
+		}
+
+		//Enumerate restore_cmd for the same gpu_id
+		for (int j = i; j < restore_cmd.cmd_head.entry_num; j++) {
+			if (restore_cmd.entries[i].gpu_id == restore_cmd.entries[j].gpu_id) {
+				vis[j] = 1;
+				fseek(bo_contents_fp, restore_cmd.entries[j].read_offset, SEEK_SET);
+				ret = sdma_copy_bo_helper(restore_cmd.entries[j].size, restore_cmd.fds_write[restore_cmd.entries[j].write_id], bo_contents_fp, buffer, buffer_size, h_dev, max_copy_size, SDMA_OP_VRAM_WRITE);
```
avagin, line 2016: This line exceeds the recommended 80-character limit.  While it's not a strict rule, keeping lines shorter generally improves readability.

criu/crtools.c
Ddnirvana: @avagin Thanks for the confirmation! We will make the change.

Ddnirvana:
Thanks for the above comments @avagin @rst0git , we are fixing and polishing the PR. Will update ASAP.

rst0git:
@Ddnirvana  @wweewrwer Thank you for contributions! It might be good to also update the content of the following file to reflect these changes: https://github.com/checkpoint-restore/criu/blob/criu-dev/plugins/amdgpu/README.md

Ddnirvana:
> @Ddnirvana @wweewrwer Thank you for your contributions! It might be good to also update the content of the following files to reflect these changes:
> 
> * https://github.com/checkpoint-restore/criu/blob/criu-dev/plugins/amdgpu/README.md
> * https://github.com/checkpoint-restore/criu/blob/criu-dev/Documentation/criu-amdgpu-plugin.txt

@rst0git No problem. We will add proper description in the next version.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,159 @@ int install_parallel_sock()
...
+
+void free_parallel_restore_cmd()
+{
+	free(restore_cmd.fds_write);
+	free(restore_cmd.entries);
+}
+
+void init_parallel_restore_cmd_by_head()
+{
+	restore_cmd.fds_write = (int *)malloc(restore_cmd.cmd_head.fd_write_num * sizeof(int));
```
dayatsin-amd, line 139: pls use xmalloc and check its return code.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,159 @@ int install_parallel_sock()
...
+	restore_cmd.entries = (parallel_restore_entry *)malloc(restore_cmd.cmd_head.entry_num * sizeof(parallel_restore_entry));
```
dayatsin-amd, line 140: pls use xmalloc and check its return code.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,159 @@ int install_parallel_sock()
...
+	return ret;
+}
+
+void init_parallel_restore_cmd(int num, int id)
+{
+	restore_cmd.cmd_head.id = id;
+	restore_cmd.cmd_head.fd_write_num = 0;
+	restore_cmd.cmd_head.entry_num = 0;
+
+	restore_cmd.fds_write = (int *)malloc(num * sizeof(int));
```
dayatsin-amd, line 124: pls use xmalloc and check its return code

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,159 @@ int install_parallel_sock()
...
+	restore_cmd.entries = (parallel_restore_entry *)malloc(num * sizeof(parallel_restore_entry));
```
dayatsin-amd, line 125: pls use xmalloc and check its return code.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,159 @@ int install_parallel_sock()
...
+
+	memset(restore_cmd.fds_write, 0, num * sizeof(int));
```
dayatsin-amd, line 127: Will not need these two lines with xmalloc

dayatsin-amd (commented):
Thank you @wweewrwer. Some minor nit picks, but overall the code looks good to me.

wweewrwer:
@rst0git @avagin @dayatsin-amd hi maintainers, thanks for your prior reviews and comments. We have fixed all the issues, as the following:
1. Use the proper APIs to allocate (xmalloc, etc.)
2. Enable the optimizations by default
3. Change the name of hook
4. Fix the issues to run in Podman containers
5. Other fixes (line width, comments, etc.) 
6. Add descriptions in README to explain the optimizations.

Please let us know if you have any further comments

rst0git:
@wweewrwer Would you be able to merge the fixup commits into the previous commits using `git rebase`?
https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#submit-your-work-upstream

wweewrwer:
> @wweewrwer Would you be able to merge the fixup commits into the previous commits using `git rebase`? https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#submit-your-work-upstream

Thanks for your comment! I have merged the fixup commits into the previous commits using git rebase. Please let me know if you have any further comments.

criu/cr-restore.c
```diff
@@ -2125,6 +2125,10 @@ static int restore_root_task(struct pstree_item *init)
 	__restore_switch_stage(CR_STATE_FORKING);
 
 skip_ns_bouncing:
+	pr_info("Run restore device parallel hook from criu master for external devices\n");
+	ret = run_plugins(RESTORE_DEVICE_PARALLEL);
```
avagin, line 2129: RESTORE_DEVICE_PARALLEL doesn't look like a good name. You are using it for restore device states, but other plugins can use it for other things. In this case, I think a name has to describe a point where it is called.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -0,0 +1,43 @@
...
+	addr->sun_family = AF_UNIX;
+	snprintf(addr->sun_path, UNIX_PATH_MAX, "x/criu-amdgpu-parallel-%" PRIx64, criu_run_id);
+	*len = SUN_LEN(addr);
+	*addr->sun_path = '\0';
+}
+
+int install_parallel_sock()
+{
+	int ret = 0;
+	int sock_fd;
```
avagin, line 27: nit: empty line after declarations

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -0,0 +1,43 @@
...
+	if ((sock_fd = socket(PF_UNIX, SOCK_DGRAM | SOCK_CLOEXEC, 0)) < 0) {
+		pr_perror("socket creation failed");
+		return -1;
+	}
+	amdgpu_socket_name_gen(&sock_addr, &addr_len);
+	if ((ret = bind(sock_fd, (struct sockaddr *)&sock_addr, addr_len)) < 0) {
```
avagin, line 33: don't do assignments in if statements

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -0,0 +1,43 @@
...
+#include "criu-log.h"
+#include "common/scm.h"
+#include "fdstore.h"
+#include <time.h>
+#include <semaphore.h>
+#include <fcntl.h>
+#include "util-pie.h"
+#include "util.h"
+
+int addr_len;
```
avagin, line 11: These names are too generic to be in the global space.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,171 @@ int install_parallel_sock()
...
+	restore_cmd.fds_write[restore_cmd.cmd_head.fd_write_num] = dmabuf_fd;
+
+	restore_cmd.cmd_head.entry_num += 1;
+	restore_cmd.cmd_head.fd_write_num += 1;
+}
+
+static int send_metadata(int sock_fd)
+{
+	if (sendto(sock_fd, &restore_cmd.cmd_head, sizeof(parallel_restore_cmd_head), 0, (struct sockaddr *)&sock_addr, addr_len) < 0) {
+		pr_perror("write");
```
avagin, line 65: pls add a more details error message

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -40,4 +41,171 @@ int install_parallel_sock()
...
+int recv_parallel_restore_cmd()
+{
+	int sock_fd;
+	int ret = 0;
+
+	if ((sock_fd = fdstore_get(sock_id)) < 0)
+		return -1;
+
+	pr_info("Parallel restore: begin to recv cmd_head\n");
+	ret = recv_metadata(sock_fd);
```
avagin, line 188: How do you handle a case when two processes send commands concurrently?

criu/cr-restore.c
wweewrwer: This hook is called after `CR_STATE_FORKING` and before `CR_STATE_RESTORE_SIGCHLD`. Its purpose is to allow the plugin to perform certain operations in the CRIU main process in parallel with the restore operations in the target processes during `CR_STATE_RESTORE`. So could I name it `RESTORE_PARALLEL_WITH_TASKS`, or are there any other name suggestions?

rst0git: >This hook is called after CR_STATE_FORKING and before CR_STATE_RESTORE_SIGCHLD

I also agree with Andrei that it would be better for the name to describe the *place* where this plugin hook is being invoked, rather than the *purpose* of the operations performed with the AMD GPU plugin. Other plugins may use this hook for different purposes.

What about something like `PRE_RESTORE_SIGCHLD` or `POST_FORKING`?

wweewrwer: OK. We will use `POST_FORKING`. If there are other name suggestions, please let us know.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1836,147 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+  amdgpu_device_handle h_dev;
+	uint64_t max_copy_size;
+	size_t total_bo_size = 0, max_bo_size = 0, buffer_size = 0;
+	FILE *bo_contents_fp = NULL;
+	parallel_restore_entry *entry;
+	void *buffer = NULL;
+	int ret = 0;
+
+	pr_info("Begin to recv parallel restore cmd\n");
+	ret = recv_parallel_restore_cmd();
```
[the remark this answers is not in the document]
wweewrwer: Sorry for this bug. I've fixed it by rewriting this hook function in the AMDGPU plugin. Compared to the previous implementation, there are some changes:

- Replaced UDP with TCP to distinguish messages between different processes and commands.
- Instead of receiving the command only once, the hook function now launches a thread to receive commands indefinitely until all tasks finish their restore stage. The main thread in this hook uses `restore_wait_inprogress_tasks` to determine when tasks have finished. Once they are completed, it sends an exit command to the parallel restore server to stop receiving commands.

plugins/amdgpu/README.md
```diff
@@ -224,6 +225,27 @@ to resume execution on the GPUs.
...
 
+## Restoring BO content in parallel
+
+Restoring the BO content is an important part in the restore of GPU state and
+usually takes a significant amount of time. A possible location for this
+procedure is the `cr_plugin_restore_file` plugin. However, restoring in this
+plugin blocks the target process from performing other restore operations, which
+hinders further optimization of the restore process.
+
+Therefore, a new plugin that runs in the master restore process is introduced,
```
rst0git, line 236: ... a new plugin _hook_?

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -49,6 +50,233 @@ int install_parallel_sock()
...
+	restore_cmd.fds_write = xzalloc(restore_cmd.cmd_head.fd_write_num * sizeof(int));
+	if (!restore_cmd.fds_write)
+		return -ENOMEM;
+	restore_cmd.entries = xzalloc(restore_cmd.cmd_head.entry_num * sizeof(parallel_restore_entry));
+	if (!restore_cmd.entries)
+		return -ENOMEM;
+	return 0;
+}
+
+static int check_quit_cmd()
```
rst0git, line 174:
```Suggestion
static int check_quit_cmd(void)
```

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -49,6 +50,233 @@ int install_parallel_sock()
...
+
+void free_parallel_restore_cmd()
+{
+	if (restore_cmd.fds_write)
+		xfree(restore_cmd.fds_write);
+	if (restore_cmd.entries)
+		xfree(restore_cmd.entries);
+}
+
+static int init_parallel_restore_cmd_by_head()
```
rst0git, line 163:
```Suggestion
static int init_parallel_restore_cmd_by_head(void)
```

wweewrwer:
@rst0git @avagin 
Dear maintainers,

We have pushed the V4 version of the PR, completing all mentioned issues since the last version. Specifically, we: (1) support multiple commands (from a single process), (2) support multiple processes restore, and (3) fix other minor issues mentioned.

Details:

- Replaced UDP with TCP to distinguish messages between different processes and commands.
- Multiple-command support: Instead of receiving the command only once, the hook function now launches a dedicated thread to receive commands indefinitely until all tasks finish their restore stage. The main thread in this hook uses `restore_wait_inprogress_tasks` to determine when tasks have finished. Once completed, it sends an exit command to the parallel restore thread to stop receiving commands.
- Multi-process support: In the case of multiple processes, they are restored in parallel (with different processes) by default, which will not benefit from the parallel optimizations. Therefore, we introduce a flag (called `parallel_disabled`) to only enable the optimization for single-process (which is the common case) as a fast path, and fallback to original restore otherwise.
- Multi-GPU parallel restore support: In the original restore, when a process has multiple GPUs, the content on each GPU is restored in parallel. In this version, we have supported multi-GPU parallel restore utilizing the original design.
- Other issues: Big thanks to Andrei and Radostin for other issues and suggestions, which are all fixed accordingly.

We have performed all the tests with the above changes. The PR can still bring 31% decrease for the restore latency in the case of single process, and achieves the same results for mutlti-process scenarios.

Please let me know if you have any further comments.

wweewrwer:
@rst0git @avagin Just a friendly reminder about the updates in this PR (in case maintainers miss the prior notifications)

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -351,6 +356,14 @@ int amdgpu_plugin_init(int stage)
...
 
 	if (stage == CR_PLUGIN_STAGE__RESTORE) {
+		if (has_children(root_item)) {
+			pr_info("Parallel restore disabled\n");
+			parallel_disabled = true;
+		}
+
+		if (!parallel_disabled && install_parallel_sock() < 0) {
+			return -1;
+		}
```
rst0git, line 366:
```C
if (has_children(root_item)) {
    pr_info("Parallel restore disabled\n");
    parallel_disabled = true;
} else {
    if (install_parallel_sock() < 0) {
        pr_err("Failed to install parallel socket\n");
        return -1;
    }
}
````

File: plugins/amdgpu/amdgpu_plugin.c [hunk 23]
```diff
@@ -64,6 +67,8 @@ bool plugin_added_to_inventory = false;
 
 bool plugin_disabled = false;
 
+bool parallel_disabled = false;
```
rst0git, line 70: Could you please add a comment to clarify why/when parallel restore is disabled?

wweewrwer: OK, I will add it in the next version.

wweewrwer:
> Have you investigated other approaches of parallel restoring of BO? For example, it is possible to fork a thread and restoring BO asynchronously in context of its process. In this case, two BO will be restored concurrently.

Yes, we have investigated the approach of forking a thread in the background, but it cannot work as it conflicts with the restore logic of CRIU.

Specifically, when CRIU tries to restore its memory state, it will unmap all old mappings. However, some mappings may be needed by the background thread for BO restoring. Therefore, a thread can only run in parallel with shorter procedures (possibly before entering the restorer blob), while offloading the restore of BO content to a new process (in this PR) can be parallelized with almost the entire restore procedure.

Below figure shows the issues that BO restore must be finished before the CPU memory state restore:

71b554be0823fc028c0fd9a8f9c554f

wweewrwer:
@rst0git @avagin Dear maintainers/reviewers, just want to know if there are any further issues/concerns about the latest version?

wweewrwer:
> I have requested this PR to be validated on a multi-GPU set-up internally at AMD. Can you give us a few days to confirm there is no regression.
> 
> Thank you for this patch!

Sure. Thanks!

Ddnirvana:
> I have requested this PR to be validated on a multi-GPU set-up internally at AMD. Can you give us a few days to confirm there is no regression.
> 
> Thank you for this patch!

Dear David @dayatsin-amd , just want to know are there any progress/results about the internal regression test. Thank you again for the assistance and happy new year btw :)

avagin:
> > Have you investigated other approaches of parallel restoring of BO? For example, it is possible to fork a thread and restoring BO asynchronously in context of its process. In this case, two BO will be restored concurrently.
> 
> Yes, we have investigated the approach of forking a thread in the background, but it cannot work as it conflicts with the restore logic of CRIU.
> 
> Specifically, when CRIU tries to restore its memory state, it will unmap all old mappings. However, some mappings may be needed by the background thread for BO restoring. Therefore, a thread can only run in parallel with shorter procedures (possibly before entering the restorer blob), while offloading the restore of BO content to a new process (in this PR) can be parallelized with almost the entire restore procedure.

Everything what is happening in the restore blob should be fast, the main restore process is happening before switching to the restorer blob. I am still not convinced that we need to offload that to another process... Maybe I miss something, but I want to see numbers and clear explanation why the current approach is good.

This new approach solves one parallelism issue by introducing another one. In this new version, all buffer objects are restored consequently. It means that it can be even slower for workloads with many buffer objects (BO).

Another problem here is that a plugin callback is running through the entire restore process

avagin:
> > Have you investigated other approaches of parallel restoring of BO? For example, it is possible to fork a thread and restoring BO asynchronously in context of its process. In this case, two BO will be restored concurrently.
> 
> Yes, we have investigated the approach of forking a thread in the background, but it cannot work as it conflicts with the restore logic of CRIU.
> 
> Specifically, when CRIU tries to restore its memory state, it will unmap all old mappings. However, some mappings may be needed by the background thread for BO restoring. Therefore, a thread can only run in parallel with shorter procedures (possibly before entering the restorer blob), while offloading the restore of BO content to a new process (in this PR) can be parallelized with almost the entire restore procedure.

Everything what is happening in the restore blob should be fast. All mappings are restored before switching into the restore blob. There, the restored mappings are just remapped to proper addresses. I am still not convinced that the idea of restoring buffer objects from the main process is really what we need here.. I can miss something, but I want to see a clear explanation with numbers why the proposed solution is a valuable one.

Additionally, I see two potential issues:
* Sequential Restoration:  This change seems to introduce a new bottleneck by restoring buffer objects sequentially. Could this cause performance problems for workloads with many buffer objects across multiple processes? It would be helpful to understand how this approach scales.
* Plugin Hook Execution: Running the plugin hook in the main CRIU process for an extended period and making it dependent on other processes is problematic. This deviates from the expectation that multiple plugins should operate independently with equal capabilities.

wweewrwer:
> > > Have you investigated other approaches of parallel restoring of BO? For example, it is possible to fork a thread and restoring BO asynchronously in context of its process. In this case, two BO will be restored concurrently.
> > 
> > 
> > Yes, we have investigated the approach of forking a thread in the background, but it cannot work as it conflicts with the restore logic of CRIU.
> > Specifically, when CRIU tries to restore its memory state, it will unmap all old mappings. However, some mappings may be needed by the background thread for BO restoring. Therefore, a thread can only run in parallel with shorter procedures (possibly before entering the restorer blob), while offloading the restore of BO content to a new process (in this PR) can be parallelized with almost the entire restore procedure.
> 
> Everything what is happening in the restore blob should be fast. All mappings are restored before switching into the restore blob. There, the restored mappings are just remapped to proper addresses. I am still not convinced that the idea of restoring buffer objects from the main process is really what we need here.. I can miss something, but I want to see a clear explanation with numbers why the proposed solution is a valuable one.
> 
> Additionally, I see two potential issues:
> 
> * Sequential Restoration:  This change seems to introduce a new bottleneck by restoring buffer objects sequentially. Could this cause performance problems for workloads with many buffer objects across multiple processes? It would be helpful to understand how this approach scales.
> * Plugin Hook Execution: Running the plugin hook in the main CRIU process for an extended period and making it dependent on other processes is problematic. This deviates from the expectation that multiple plugins should operate independently with equal capabilities.

Thank you for your comment.

> Everything what is happening in the restore blob should be fast. All mappings are restored before switching into the restore blob. There, the restored mappings are just remapped to proper addresses.

In the case of GPU apps (AMD GPUs in our case), not everything in the restore blob is fast, and not all mappings are restored before switching into the restore blob. Some mappings are still restored within the restore blob (e.g., `criu/pie/restorer.c:1807`), and these operations take a significant amount of time.

In our tests, the restore time of the restore blob even takes longer than the GPU buffer objects restore time (this can be easily reproduced using the same test application and approach as in this pull request). Here is the breakdown of our test.



This is the motivation of the PR and we believe it is valuable to parallelize the restoration of GPU buffer objects with other CRIU restore operations for optimized restore latency.

> It would be helpful to understand how this approach scales.

This PR will not lead to sequential restoration issue. The optimization focuses on the single-process situation (common case), as shown in the following table. In other scenarios, it will turn to the original method. This is achieved with the new `parallel_disabled` flag.

|                  | Our method | Original method |
| ---------------- | ---------- | --------------- |
| Single process   | Y          |                 |
| Multiple process |            | Y               |

And for a single process, its restore logic is the same as the original method except that it is offloaded to the main CRIU process.

> Running the plugin hook in the main CRIU process for an extended period and making it dependent on other processes is problematic.

It will not introduce new dependency. Currently the main CRIU process must wait for other processes to complete their memory state restoration before proceeding with additional restoration logic. Besides, most of the code is in separate files and related to AMD GPUs for optimized performance, and will not break anything in the main path.

Please let me know if you have any further comments.

avagin:
> In the case of GPU apps (AMD GPUs in our case), not everything in the restore blob is fast, and not all mappings are restored before switching into the restore blob. Some mappings are still restored within the restore blob (e.g., criu/pie/restorer.c:1807), and these operations take a significant amount of time.

I forgot that we moved restore of anon-vma-s to the restorer (91388fce03). Now, I understand the problem. Thanks.

> It will not introduce new dependency. Currently the main CRIU process must wait for other processes to complete their memory state restoration before proceeding with additional restoration logic. Besides, most of the code is in separate files and related to AMD GPUs for optimized performance, and will not break anything in the main path.

I think you misunderstand me here. I don't mean code dependencies. The problem here is that the amd plugin completely occupied the main criu process. It means if there will be another plugin with a similar post-fork hook, these plugin will not work together. If you want to introduce a hook handling events from child processes, it should be running in context of a separate thread.

ps, I like all these detailed comments, but why all these details have not been explained on the commit messages. It would speed up the review process and in future, it would help to understand ideas behind the implementation.

wweewrwer:
Hi Andrei,

> I think you misunderstand me here. I don't mean code dependencies. The problem here is that the amd plugin completely occupied the main criu process. It means if there will be another plugin with a similar post-fork hook, these plugin will not work together. If you want to introduce a hook handling events from child processes, it should be running in context of a separate thread.

Thanks for the clarification! Got the concern.

We can only launch a separate thread in `POST_FORKING` hook (but not blocking), and wait for the thread in a latter hook  `RESUME_DEVICES_LATE` to avoid occupy the main CRIU process and other plugins in `POST_FORKING`. 


Specifically, to avoid occupying the main CRIU process, we only launch a thread in `POST_FORKING` without waiting for it to complete. Instead, we wait for this thread in the next hook, `RESUME_DEVICES_LATE`.  In `RESUME_DEVICES_LATE`, since every child process has completed its restore logic, no new commands will be sent to the thread. Therefore, we can then immediately send a stop command to the thread and wait for it to finish processing and terminate.

This design won't add extra waiting time to the main CRIU process, as when the GPU data is restored in child processes, the main CRIU process also needs to wait for the child processes to finish before calling the `RESUME_DEVICES_LATE` hook.

This is a not a big change, but can make the whole thing more clean and avoid the occupation issue.

We are making the change and will post updates once done. Please feel free to let us know if you have any further comments/suggestions on this.

> ps, I like all these detailed comments, but why all these details have not been explained on the commit messages. It would speed up the review process and in future, it would help to understand ideas behind the implementation.

Big thanks for the suggestion. Will add more key information during the discussion/comments to the commit messages in the next version. And will also follow the principle in the future contributions :)

criu/pstree.c
```diff
@@ -241,6 +241,11 @@ int init_pstree_helper(struct pstree_item *ret)
 	return 0;
 }
 
+bool has_children(struct pstree_item *item)
```
rst0git, line 244: Could you also replace `!list_empty(&item->children)` with has_children() in `cr-dump.c`, `pstree.c` and `seize.c`?

plugins/amdgpu/README.md
```diff
@@ -224,6 +225,27 @@ to resume execution on the GPUs.
...
+the restore of BO content. Specifically, the target process only needs to send
+the relevant BOs to the master restore process, while this new hook handles all
+the restore of buffer objects. Through this method, during the restore of the BO
+content, the target process can perform other restore operations, thus
+accelerating the restore procedure. It is an implementation of gCROP from the
+ACM SoCC'24 paper: [On-demand and Parallel Checkpoint/Restore for GPU
+Applications](https://dl.acm.org/doi/10.1145/3698038.3698510).
+
+*This new plugin is enabled by the new hook `__POST_FORKING` in our patch
+series.*
```
rst0git, line 247: Once this pull request has been merged, "new plugin" and "our patch series" might become slightly confusing to the reader. What about something like the following?

_This optimization technique is enabled by the `__POST_FORKING` hook._

plugins/amdgpu/README.md
```diff
@@ -224,6 +225,27 @@ to resume execution on the GPUs.
...
+hinders further optimization of the restore process.
+
+Therefore, a new plugin hook that runs in the master restore process is
+introduced, and it interacts with the `cr_plugin_restore_file` hook to complete
+the restore of BO content. Specifically, the target process only needs to send
+the relevant BOs to the master restore process, while this new hook handles all
+the restore of buffer objects. Through this method, during the restore of the BO
+content, the target process can perform other restore operations, thus
+accelerating the restore procedure. It is an implementation of gCROP from the
+ACM SoCC'24 paper: [On-demand and Parallel Checkpoint/Restore for GPU
```
rst0git, line 243: _This is an implementation of the gCROP method proposed in the ACM SoCC'24 paper ..._

wweewrwer:
The previous suggestions have been updated, and both the regression test and related tests have been completed. Please let us know if you have any further comments. :)

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1935,214 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+
+		for (int i = 0; i < restore_cmd.cmd_head.gpu_num; i++) {
+			thread_datas[i].gpu_id = restore_cmd.gpu_ids[i].gpu_id;
+			thread_datas[i].minor = restore_cmd.gpu_ids[i].minor;
+			thread_datas[i].restore_cmd = &restore_cmd;
+
+			ret = pthread_create(&thread_datas[i].thread, NULL, parallel_restore_bo_contents,
+					     (void *)&thread_datas[i]);
+			if (ret) {
+				pr_err("Failed to create thread[%i] ret:%d\n", i, ret);
```
avagin, line 2106: pthread_create sets errno, so it is better to use pr_perror

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1935,214 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+		ret = recv_parallel_restore_cmd(&restore_cmd);
+		if (ret) {
+			if (ret == 1) {
+				*(int *)arg = 0;
+				goto exit;
+			}
+			goto err;
+		}
+
+		thread_datas = xzalloc(sizeof(*thread_datas) * restore_cmd.cmd_head.gpu_num);
```
avagin, line 2092: where do you free thread_datas?

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1935,214 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+}
+
+void *restore_device_parallel_worker(void *arg)
+{
+	while (1) {
+		parallel_restore_cmd restore_cmd = { 0 };
+		struct parallel_thread_data *thread_datas;
+		int ret;
+
+		pr_info("Begin to recv parallel restore cmd\n");
```
avagin, line 2082: nit: pr_debug looks more reasonable in this case.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1836,6 +1897,18 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
 	if (plugin_disabled)
 		return -ENOTSUP;
 
+	if (!parallel_disabled) {
+		pr_info("Close parallel restore server\n");
+		if (close_parallel_restore_server())
+			return -1;
+
+		pthread_join(parallel_thread, NULL);
```
avagin, line 1905: Need to check a return code

criu/cr-restore.c
```diff
@@ -2125,6 +2125,10 @@ static int restore_root_task(struct pstree_item *init)
 	__restore_switch_stage(CR_STATE_FORKING);
 
 skip_ns_bouncing:
+	pr_info("Run post forking hook from criu master\n");
```
avagin, line 2128: pr_debug or even remove it. The last is preferred.

wweewrwer:
@avagin Thanks for your comments and all suggestions have been updated! Please let us know if you have any further comments. :)

Ddnirvana:
@dayatsin-amd Dear David, could you help have a look at this PR to check whether the changes since your last approval are good to you? Yanning (wweewrwer) is always active to respond any requests/suggestions in this PR and take many efforts to make it better. Thanks!!!

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1836,6 +1898,22 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
 	if (plugin_disabled)
 		return -ENOTSUP;
 
+	if (!parallel_disabled) {
+		pr_info("Close parallel restore server\n");
+		if (close_parallel_restore_server())
```
dayatsin-amd, line 1903: Minor, can you please add an an error print here.

dayatsin-amd (changes requested):
Sorry for late review. 
Just one minor comment. Otherwise it looks good to me.

wweewrwer:
> Sorry for late review. Just one minor comment. Otherwise it looks good to me.

@dayatsin-amd Thanks for your comment! It has been updated now. Please let us know if you have any further comments. :)

dayatsin-amd:
@wweewrwer  Looks good to me. 

Sorry one more, one of the CI's (ci/circleci:test-local-clang) have a warning.

Can you initialize 
struct parallel_thread_data *thread_datas;

to null in amdgpu_plugin.c:2086

wweewrwer:
> @wweewrwer Looks good to me.
> 
> Sorry one more, one of the CI's (ci/circleci:test-local-clang) have a warning.
> 
> Can you initialize struct parallel_thread_data *thread_datas;
> 
> to null in amdgpu_plugin.c:2086

OK, it is fixed now. I also rebased our code to the newest commit.  Besides that, there are no updates to the code.

dayatsin-amd (approved):
Looks good to me.
Thank you!

Ddnirvana:
@dayatsin-amd Thank you David for the approval!!!

Dear @rst0git Radostin @avagin  Andrei, the PR has been rebased to the latest commit and there are no pending requests for further change. Just let us know if there are any other concerns blocking it. Thanks!!

avagin:
@Ddnirvana, I was busy at work this week. I will find time next week to review this PR.

Ddnirvana:
@avagin just a friendly reminder about the PR.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -0,0 +1,55 @@
+#include "amdgpu_socket_utils.h"
+#include "criu-log.h"
+#include "common/scm.h"
+#include "fdstore.h"
+#include "util-pie.h"
+#include "util.h"
+#include <time.h>
+#include <semaphore.h>
```
rst0git, line 8: It looks like `time.h` and `semaphore.h` are not used in this file.

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -49,6 +50,278 @@ int install_parallel_sock(void)
...
+	pr_debug("Parallel restore: begin to recv entries\n");
+	ret = recv_cmds(client_fd, restore_cmd);
+	if (ret) {
+		goto err;
+	}
+
+	pr_debug("Parallel restore: begin to recv fds\n");
+	ret = recv_dmabuf_fds(client_fd, restore_cmd);
+
+	pr_debug("Parallel restore: recv over\n");
```
rst0git, line 292: perhaps we no longer need these debug messages?

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -49,6 +50,278 @@ int install_parallel_sock(void)
...
+{
+	restore_cmd->gpu_ids = xzalloc(restore_cmd->cmd_head.gpu_num * sizeof(parallel_gpu_info));
+	if (!restore_cmd->gpu_ids)
+		return -ENOMEM;
+	restore_cmd->fds_write = xzalloc(restore_cmd->cmd_head.fd_write_num * sizeof(int));
+	if (!restore_cmd->fds_write)
+		return -ENOMEM;
+	restore_cmd->entries = xzalloc(restore_cmd->cmd_head.entry_num * sizeof(parallel_restore_entry));
+	if (!restore_cmd->entries)
+		return -ENOMEM;
```
rst0git, line 200: Do we need to `free_parallel_restore_cmd()` here?

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -49,6 +50,278 @@ int install_parallel_sock(void)
...
+	sock_fd = socket(PF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
+	if (sock_fd < 0) {
+		pr_perror("Socket creation failed");
+		return -1;
+	}
+
+	ret = connect(sock_fd, (struct sockaddr *)&parallel_socket_addr, parallel_socket_addr_len);
+	if (ret < 0) {
+		pr_perror("Connect failed");
+		goto err;
```
rst0git, line 316: `close_parallel_restore_server()` will `return 0` in this case. Would it make sense to use `int ret = -1;`?

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1942,235 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+								    SDMA_LINEAR_COPY_MAX_SIZE - 1;
+	return 0;
+err:
+	amdgpu_device_deinitialize(*h_dev);
+	return ret;
+}
+
+FILE *get_bo_contents_fp(int id, int gpu_id, size_t tot_size)
+{
+	char img_path[40];
```
rst0git, line 1987: `char img_path[PATH_MAX];`

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1942,235 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+	if (plugin_disabled)
+		return -ENOTSUP;
+
+	if (parallel_disabled)
+		return 0;
+
+	if (back_thread_create(&parallel_thread, restore_device_parallel_worker, &parallel_thread_result))
+		return -1;
+
+	return 0;
```
rst0git, line 2174: We can simplify this to
```
return back_thread_create(&parallel_thread, restore_device_parallel_worker, &parallel_thread_result);
```

plugins/amdgpu/amdgpu_socket_utils.c
```diff
@@ -49,6 +50,278 @@ int install_parallel_sock(void)
...
+
+	pr_debug("Parallel restore: begin to recv cmd_head\n");
+	ret = recv_metadata(client_fd, restore_cmd);
+	if (ret) {
+		goto err;
+	}
+
+	ret = init_parallel_restore_cmd_by_head(restore_cmd);
+	if (ret) {
+		goto err;
```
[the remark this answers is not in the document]
wweewrwer: Sorry, I don't quite understand this point. If `init_parallel_restore_cmd_by_head()` fails, I think it will go to `err` and return `ret`, which is `-ENOMEM`. Why will `recv_parallel_restore_cmd()` return 0?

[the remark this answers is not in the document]
rst0git: Sorry, I missed that. We usually use `return exit_code` instead of `ret`.

plugins/amdgpu/amdgpu_socket_utils.c
wweewrwer: No, this function is paired with `free_parallel_restore_cmd` externally. Regardless of whether it allocates successfully, `free_parallel_restore_cmd` will be called, so there's no need to free it here.

criu/cr-restore.c
```diff
@@ -2394,6 +2391,9 @@ int cr_restore_tasks(void)
 	if (fdstore_init())
 		goto err;
 
+	if (cr_plugin_init(CR_PLUGIN_STAGE__RESTORE))
```
avagin, line 2394: pls add a comment why it is called here.

criu/cr-restore.c
```diff
@@ -2394,6 +2391,9 @@ int cr_restore_tasks(void)
 	if (fdstore_init())
 		goto err;
```
avagin, line 2392: it seems that goto err should be used only after cr_plugin_init, because it calls cr_plugin_fini.

plugins/amdgpu/amdgpu_plugin.c
avagin: It is still unclear why just uniprocess workloads will benefit from this optimization.

criu/cr-restore.c
```diff
@@ -2358,41 +2358,47 @@ int cr_restore_tasks(void)
...
 	if (prepare_pstree() < 0)
-		goto err;
+		return -1;
 
 	if (fdstore_init())
-		goto err;
+		return -1;
+
+	/*
+   * For the AMDGPU plugin, its parallel restore feature needs to use fdstore to store
```
rst0git, line 2395: Could you please use tabs for indentation here?

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1862,3 +1941,233 @@ int amdgpu_plugin_resume_devices_late(int target_pid)
...
+
+			if (thread_datas[i].ret) {
+				ret = thread_datas[i].ret;
+				goto err;
+			}
+		}
+
+err:
+		if (thread_datas)
+			xfree(thread_datas);
```
avagin, line 2130: thread_datas can't be freed until all threads have exited. In the previous loop, you go to the err label when one thread fails, so some other threads can be still running.

plugins/amdgpu/amdgpu_plugin.c
wweewrwer: Sorry for the confusion earlier. We have now updated the comment to provide further clarification.

Ddnirvana:
Dear @rst0git  Radostin and  @avagin  Andrei, could you take a look again and merge it if there are no pending issues? It is already blocked for 1 month after Yanning's last update.

PR #2533, opened 2024-11-26
Support criu-image-streamer through RPC

lianakoleva:
Programs calling CRIU through RPC may want access to image streaming
for faster data transfer. 

To use image streaming through RPC, 
(1) criu-image-streamer must be started by the calling program.
(2) The RPC call must enable stream in `criu_opts`.


criu/cr-service.c
```diff
@@ -905,7 +901,7 @@ static int check(int sk, CriuOpts *req)
 		__setproctitle("check --rpc");
 
 		opts.mode = CR_CHECK;
-		if (setup_opts_from_req(sk, req))
+		if (setup_opts_from_req(sk, req, -1))
```
rst0git, line 904: Would it be possible to add a comment that explains why we use `-1` here?

criu/cr-service.c
```diff
@@ -1015,7 +1011,7 @@ static int start_page_server_req(int sk, CriuOpts *req, bool daemon_mode)
 		close(start_pipe[0]);
 
 		opts.mode = CR_PAGE_SERVER;
-		if (setup_opts_from_req(sk, req))
+		if (setup_opts_from_req(sk, req, -1))
```
rst0git, line 1014: It would be good to add a similar comment here.

criu/cr-service.c
```diff
@@ -285,7 +285,7 @@ int exec_rpc_query_external_files(char *name, int sk)
 
 static char images_dir[PATH_MAX];
 
-static int setup_opts_from_req(int sk, CriuOpts *req)
+static int setup_opts_from_req(int sk, CriuOpts *req, int mode)
```
rst0git, line 288: It looks like another variable with the same name is used in this function:

```C
	/* Override the manage_cgroup if mode is set explicitly */
	if (req->has_manage_cgroups_mode) {
		unsigned int mode;
```

rst0git:
@lianakoleva Would it be possible to run tests for this functionality with [ZDTM](https://github.com/checkpoint-restore/criu/blob/1452c76f6541721a77e886a41c6ee63bb5571c0c/scripts/ci/run-ci-tests.sh#L282)?

For example, with something similar to `./zdtm.py run -t zdtm/static/env00 --rpc --stream`?

PR #2536, opened 2024-12-06
Support criu-image-streamer through RPC

lianakoleva:
Programs calling CRIU through RPC may want access to image streaming for faster data transfer.

To use image streaming through RPC,
(1) criu-image-streamer must be started by the calling program.
(2) The RPC call must enable stream in `criu_opts`.

The `image_dir_mode` function was modified to be used by RPC as well. This required some changes to accommodate for `CR_CPUINFO`, which has a subcommand `dump` or `check`.


criu/crtools.c
```diff
@@ -54,19 +54,17 @@ void flush_early_log_to_stderr(void)
...
+	case CR_CPUINFO_DUMP:
+		/* fallthrough */
 	case CR_PRE_DUMP:
 		return O_DUMP;
 	case CR_RESTORE:
 		return O_RSTR;
-	case CR_CPUINFO:
-		if (!strcmp(argv[optind + 1], "dump"))
-			return O_DUMP;
-		/* fallthrough */
```
rst0git, line 69: Would it be possible to describe these changes in your commit message?

criu/cr-service.c
```diff
@@ -1261,7 +1257,7 @@ static int handle_cpuinfo(int sk, CriuReq *msg)
 	if (pid == 0) {
 		int ret = 1;
 
-		opts.mode = CR_CPUINFO;
+		opts.mode = (msg->type == CRIU_REQ_TYPE__CPUINFO_DUMP) ? CR_CPUINFO_DUMP : CR_CPUINFO_CHECK;
```
rst0git, line 1260: Why do we change `opts.mode` here?

criu/crtools.c
```diff
@@ -96,8 +94,12 @@ static int parse_criu_mode(char *mode)
...
 	else if (!strcmp(mode, "dedup"))
 		opts.mode = CR_DEDUP;
-	else if (!strcmp(mode, "cpuinfo"))
-		opts.mode = CR_CPUINFO;
+	else if (!strcmp(mode, "cpuinfo") && subcommand == NULL)
+		return -2;
+	else if (!strcmp(mode, "cpuinfo") && !strcmp(subcommand, "dump"))
+		opts.mode = CR_CPUINFO_DUMP;
+	else if (!strcmp(mode, "cpuinfo") && !strcmp(subcommand, "check"))
+		opts.mode = CR_CPUINFO_CHECK;
```
rst0git, line 102: It might be better to separate the changes for cpuinfo and opts.stream in different commits:

criu/cr-service.c
lianakoleva: The command "cpuinfo" requires a subcommand "dump" or "check".
- `CR_CPUINFO` is removed; `CR_CPUINFO_DUMP` and `CR_CPUINFO_CHECK` are added in its place.
- All parsing is now done in `parse_criu_mode`, removing need for passing string subcommand to `image_dir_mode`.
- Check for valid subcommand is done once at the start with only the enum `CR_CPUINFO_DUMP` or `CR_CPUINFO_CHECK` propagated on success, rather than propagating both `CR_CPUINFO` and the subcommand string while doing `strcmp` on each call to `image_dir_mode`.

criu/crtools.c
lianakoleva: Will do!

PR #2540, opened 2024-12-10
cuda: Fix return value from CHECKPOINT_DEVICES

jesus-ramos:
When cuda-checkpoint fails an operation it always returns the positive CUDA error code. In CHECKPOINT_DEVICES that was getting passed up as the failure code from one of the calls so if the operation failed we would write the failure to the log but not stop the dump since that checked for error values less than 0.


avagin:
LGTM. You need to add sob in the commit message: https://github.com/checkpoint-restore/criu/pull/2540/checks?check_run_id=34217028525

jesus-ramos:
> LGTM. You need to add sob in the commit message: https://github.com/checkpoint-restore/criu/pull/2540/checks?check_run_id=34217028525

Fixed.

PR #2543, opened 2024-12-11
test/java: increate the ghost file limit

avagin:
Right now, this test fails with this error:
```
Error (criu/files-reg.c:1031): Can't dump ghost file
  /criu/test/javaTests/omrvmem_000000626_Mlm48x of 2097152 size,
  increase limit
```


rst0git (approved):
LGTM

ymanton (approved):
LGTM. We've run into this before and adjusted the limit as well.

PR #2545, opened 2024-12-13
Freeze fixes and v1 kludges

kolyshkin:
## 1. freeze_processes: fix logic
There are a few issues with the freeze_processes logic:

1. Commit https://github.com/checkpoint-restore/criu/commit/9fae23fbe292be7f041c599c03b1f84dadea5216 grossly (by 1000x) miscalculated the number of
   attempts required, as a result, we are seeing something like this:

> (00.000340) freezing processes: 100000 attempts with 100 ms steps
> (00.000351) freezer.state=THAWED
> (00.000358) freezer.state=FREEZING
> (00.100446) freezer.state=FREEZING
> ...close to 100 lines skipped...
> (09.915110) freezer.state=FREEZING
> (10.000432) Error (criu/cr-dump.c:1467): Timeout reached. Try to interrupt: 0
> (10.000563) freezer.state=FREEZING

   For 10s with 100ms steps we only need 100 attempts, not 100000.

2. When the timeout is hit, the "failed to freeze cgroup" error is not
   printed, and the log_unfrozen_stacks is not called either.

3. The nanosleep at the last iteration is useless (this was hidden by
   issue 1 above, as the timeout was hit first).

Fix all these.

While at it,

4. Amend the error message with the number of attempts, sleep duration,
   and timeout.

5. Modify the "freezing cgroup" debug message to be in sync with the
   above error.

   Was:

   > freezing processes: 100000 attempts with 100 ms steps

   Now:

   > freezing cgroup some/name: 100 x 100ms attempts, timeout: 10s

## 2. freeze_processes: implement kludges for cgroup v1

Cgroup v1 freezer has always been problematic, failing to freeze a
cgroup.

In runc, we have implemented a few kludges to increase the chance of
succeeding, but those are used when runc freezes a cgroup for its own
purposes (for "runc pause" and to modify device properties for cgroup
v1).

When criu is used, it fails to freeze a cgroup from time to time
(see [1], [2]). Let's try adding kludges similar to ones in runc.

Alas, I have absolutely no way to test this, so please review carefully.

[1]: https://github.com/opencontainers/runc/issues/4273
[2]: https://github.com/opencontainers/runc/issues/4457

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -87,7 +87,10 @@ static const char frozen[] = "FROZEN";
 static const char freezing[] = "FREEZING";
 static const char thawed[] = "THAWED";
 
-enum freezer_state { FREEZER_ERROR = -1, THAWED, FROZEN, FREEZING };
+enum freezer_state { FREEZER_ERROR = -1,
+		     THAWED,
+		     FROZEN,
+		     FREEZING };
 
 /* Track if we are running on cgroup v2 system. */
 static bool cgroup_v2 = false;
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -539,6 +539,34 @@ err:
 	return exit_code;
 }
 
+static void cgroupv1_freezer_kludges(int fd, int iter, const struct timespec *req) {
+	/* As per older kernel docs (freezer-subsystem.txt before
+	 * the kernel commit ef9fe980c6fcc1821), if FREEZING is seen,
+	 * userspace should either retry or thaw. While current
+	 * kernel cgroup v1 docs no longer mention a need to retry,
+	 * even recent kernels can't reliably freeze a cgroup v1.
+	 *
+	 * Let's keep asking the kernel to freeze from time to time.
+	 * In addition, do occasional thaw/sleep/freeze.
+	 *
+	 * This is still a game of chances (the real fix belongs to the kernel)
+	 * but these kludges might improve the probability of success.
+	 *
+	 * Cgroup v2 does not have this problem.
+	 */
+	switch (iter % 32) {
+		case 9:
+		case 20:
+			freezer_write_state(fd, FROZEN);
+			break;
+		case 31:
+			freezer_write_state(fd, THAWED);
+			nanosleep(req, NULL);
+			freezer_write_state(fd, FROZEN);
+			break;
+	}
+}
+
 static int freeze_processes(void)
 {
 	int fd, exit_code = -1;
@@ -596,6 +624,9 @@ static int freeze_processes(void)
 
 			if (state == FROZEN || i++ == nr_attempts || alarm_timeouted())
 				break;
+
+			if (!cgroup_v2)
+				cgroupv1_freezer_kludges(fd, i, &req);
 
 			nanosleep(&req, NULL);
 		}
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -545,7 +545,8 @@ static int freeze_processes(void)
 	enum freezer_state state = THAWED;
 
 	static const unsigned long step_ms = 100;
-	unsigned long nr_attempts = (opts.timeout * 1000000) / step_ms;
+	/* Since opts.timeout is in seconds, multiply it by 1000 to convert to milliseconds. */
+	unsigned long nr_attempts = (opts.timeout * 1000) / step_ms;
 	unsigned long i = 0;
 
 	const struct timespec req = {
@@ -554,14 +555,12 @@ static int freeze_processes(void)
 	};
 
 	if (unlikely(!nr_attempts)) {
-		/*
-		 * If timeout is turned off, lets
-		 * wait for at least 10 seconds.
-		 */
-		nr_attempts = (10 * 1000000) / step_ms;
+		/* If the timeout is 0, wait for at least 10 seconds. */
+		nr_attempts = (10 * 1000) / step_ms;
 	}
 
-	pr_debug("freezing processes: %lu attempts with %lu ms steps\n", nr_attempts, step_ms);
+	pr_debug("freezing cgroup %s: %lu x %lums attempts, timeout: %us\n",
+		 opts.freeze_cgroup, nr_attempts, step_ms, opts.timeout);
 
 	fd = freezer_open();
 	if (fd < 0)
@@ -588,22 +587,22 @@ static int freeze_processes(void)
 		 * not read @tasks pids while freezer in
 		 * transition stage.
 		 */
-		for (; i <= nr_attempts; i++) {
+		while (1) {
 			state = get_freezer_state(fd);
 			if (state == FREEZER_ERROR) {
 				close(fd);
 				return -1;
 			}
 
-			if (state == FROZEN)
+			if (state == FROZEN || i++ == nr_attempts || alarm_timeouted())
 				break;
-			if (alarm_timeouted())
-				goto err;
+
 			nanosleep(&req, NULL);
 		}
 
-		if (i > nr_attempts) {
-			pr_err("Unable to freeze cgroup %s\n", opts.freeze_cgroup);
+		if (state != FROZEN) {
+			pr_err("Unable to freeze cgroup %s (%lu x %lums attempts, timeout: %us)\n",
+			       opts.freeze_cgroup, i, step_ms, opts.timeout);
 			if (!pr_quelled(LOG_DEBUG))
 				log_unfrozen_stacks(opts.freeze_cgroup);
 			goto err;


criu/seize.c
```diff
@@ -555,7 +555,7 @@ static int freeze_processes(void)
 		 * If timeout is turned off, lets
 		 * wait for at least 10 seconds.
 		 */
-		nr_attempts = (10 * 1000000) / step_ms;
+		nr_attempts = (10 * 1000) / step_ms;
```
rst0git, line 558: Do we need to update the comment above?

criu/seize.c
```diff
@@ -542,7 +542,7 @@ static int freeze_processes(void)
 	enum freezer_state state = THAWED;
 
 	static const unsigned long step_ms = 100;
-	unsigned long nr_attempts = (opts.timeout * 1000000) / step_ms;
+	unsigned long nr_attempts = (opts.timeout * 1000) / step_ms;
```
rst0git, line 545: @kolyshkin Could you please add a comment here to explain why we are multiplying the timeout by 1000?

criu/seize.c
```diff
@@ -599,6 +599,34 @@ static int freeze_processes(void)
...
+				continue;
+
+			/* As per older kernel docs (freezer-subsystem.txt before
+			 * the kernel commit ef9fe980c6fcc1821), if FREEZING is seen,
+			 * userspace should either retry or thaw. While current
+			 * kernel cgroup v1 docs no longer mention a need to retry,
+			 * even recent kernels can't reliably freeze a cgroup v1.
+			 *
+			 * Let's keep asking the kernel to freeze from time to time.
+			 * In addition, do occasional thaw/sleep/freeze.
```
rst0git, line 613: ould it be possible to add a reference to the implementation in runc?
https://github.com/opencontainers/runc/commit/d1007b08a33ae63bb695fcb3244300c9bba2658a

avagin: I am not sure why we may need it here...

criu/seize.c
```diff
@@ -599,6 +599,34 @@ static int freeze_processes(void)
...
+			 *
+			 * This is still a game of chances (the real fix belongs to the kernel)
+			 * but these kludges might improve the probability of success.
+			 *
+			 * Cgroup v2 does not have this problem.
+			 */
+			switch (i%32) {
+				case 30:
+					freezer_write_state(fd, THAWED);
```
avagin, line 622: I think it is better to do that before nanosleep, otherwise we waste one time slice.

criu/seize.c
kolyshkin: I think it's clear (s to ms conversion), but will add.

criu/seize.c
kolyshkin: We don't waste anything, we just spend one time slice in THAWED state, letting the processes to run for some time, hoping this will change their state to one in which they will have a better chance to freeze.

The only "wasteful" check is `get_freezer_state` after this, since it will return `THAWED`, and I don't consider it heavy enough to worth optimizing out and making the code more complicated than it should be.

Or, the code can be changed like this, to be more straightforward in what we're trying to do here:

```diff
@@ -620,10 +621,11 @@ static int freeze_processes(void)
                        switch (i%32) {
                                case 30:
                                        freezer_write_state(fd, THAWED);
+                                       nanosleep(&req, NULL);
+                                       freezer_write_state(fd, FROZEN);
                                        break;
                                case 9:
                                case 20:
-                               case 31:
                                        freezer_write_state(fd, FROZEN);
                                        break;
                        }
```

criu/seize.c
kolyshkin: There are 2 or 3 runc commits which change this code, and it's still failing some time. Not sure which commit to refer to...

kolyshkin:
Addressed review comments; rebased.

criu/seize.c
```diff
@@ -599,6 +600,35 @@ static int freeze_processes(void)
...
+			 *
+			 * Cgroup v2 does not have this problem.
+			 */
+			switch (i % 32) {
+			case 9:
+			case 20:
+				freezer_write_state(fd, FROZEN);
+				break;
+			case 31:
+				freezer_write_state(fd, THAWED);
```
avagin, line 627: Kir, there is no sense to do the previous nanosleep and then thaw the cgroup.

criu/seize.c
avagin:
```
			nanosleep(&req, NULL); <---- we should not sleep here if we are going to thaw the cgroup.

			if (cgroup_v2)
				continue;

			/* As per older kernel docs (freezer-subsystem.txt before
```

criu/seize.c
rst0git: I thought it would be useful to reference the runc implementation here, as if it changes in the future it might be good to apply similar changes in CRIU. Perhaps we could add something like
> This approach is similar to the one used in `libcontainer/cgroups/fs/freezer.go` in runc.

criu/seize.c
kolyshkin: Yes, my bad. Moved nanosleep to before checking the status -- the difference is the minimum sleep time is 100ms now (was 0). Not sure if it makes any practical difference. If it does, we can do
```c
if i != 0
   nanosleep(&req, NULL);
```

avagin: You can do 1M syscalls for 100ms. We probably need to reconsider this timeout. In many cases, the full container dump should take a shorter period.

let's do something like that:
```
+static void cgroupv1_freezer_kludges(int fd, int i, const struct timespec *req)
+{
+       /* As per older kernel docs (freezer-subsystem.txt before
+        * the kernel commit ef9fe980c6fcc1821), if FREEZING is seen,
+        * userspace should either retry or thaw. While current
+        * kernel cgroup v1 docs no longer mention a need to retry,
+        * even recent kernels can't reliably freeze a cgroup v1.
+        *
+        * Let's keep asking the kernel to freeze from time to time.
+        * In addition, do occasional thaw/sleep/freeze.
+        *
+        * This is still a game of chances (the real fix belongs to the kernel)
+        * but these kludges might improve the probability of success.
+        *
+        * Cgroup v2 does not have this problem.
+        */
+       switch (i % 32) {
+       case 9:
+       case 20:
+               freezer_write_state(fd, FROZEN);
+               break;
+       case 31:
+               freezer_write_state(fd, THAWED);
+               nanosleep(req, NULL);
+               freezer_write_state(fd, FROZEN);
+               break;
+       }
+}
+
 static int freeze_processes(void)
 {
        int fd, exit_code = -1;
        enum freezer_state state = THAWED;
 
        static const unsigned long step_ms = 100;
+       /* Since opts.timeout is in seconds, multiply it by 1000 to convert to milliseconds. */
        unsigned long nr_attempts = (opts.timeout * 1000) / step_ms;
        unsigned long i = 0;
 
@@ -598,6 +628,10 @@ static int freeze_processes(void)
                                pr_err("Unable to freeze cgroup %s (timed out)\n", opts.freeze_cgroup);
                                goto err;
                        }
+
+                       if (!cgroup_v2)
+                               cgroupv1_freezer_kludges(fd, i, &req);
+
                        nanosleep(&req, NULL);
               
```

kolyshkin: Makes sense, although in this case we have a wasteful nanosleep when `i` reaches `nr_attempts` and the timeout is not hit. I.e. we check the state, sleep, and then print an error. Not a new problem, but let me fix this, too.

criu/seize.c
kolyshkin: Right, I got it now.

Refactored to eliminate this (and some other) issues.

kolyshkin:
Refactored to fix some more issues.

kolyshkin:
> 1. Commit 9fae23f grossly (by 1000x) miscalculated the number of attempts required
<...>
> 2. When the timeout is hit, the "failed to freeze cgroup" error is not printed, and the log_unfrozen_stacks is not called either.

As this is now fixed, we may find additional issues with the `log_unfrozen_stacks` (which, I guess, was never called before due to nr_attempts miscalculation).

kolyshkin:
Testing this in https://github.com/opencontainers/runc/pull/4559, no luck so far (can't reproduce the issue).

kolyshkin:
> Testing this in [opencontainers/runc#4559](https://github.com/opencontainers/runc/pull/4559)

I have concluded the testing, these kludges definitely help (i.e. can easily reproduce the issue without the fix, and can not reproduce it with the fix). See [opencontainers/runc#4559](https://github.com/opencontainers/runc/pull/4559) for more details.

kolyshkin:
@avagin PTAL

PR #2547, opened 2024-12-14
cuda: prevent task lockup on timeout error

rst0git:
When creating a checkpoint of large models, the `checkpoint` action of `cuda-checkpoint` can exceed the CRIU timeout. This causes CRIU to fail with the following error, leaving the CUDA task in a locked state:

	cuda_plugin: Checkpointing CUDA devices on pid 84145 restore_tid 84202
	Error (criu/cr-dump.c:1791): Timeout reached. Try to interrupt: 0
	Error (cuda_plugin.c:139): cuda_plugin: Unable to read output of cuda-checkpoint: Interrupted system call
	Error (cuda_plugin.c:396): cuda_plugin: CHECKPOINT_DEVICES failed with
	net: Unlock network
	cuda_plugin: finished cuda_plugin stage 0 err -1
	cuda_plugin: resuming devices on pid 84145
	cuda_plugin: Restore thread pid 84202 found for real pid 84145
	Unfreezing tasks into 1
		Unseizing 84145 into 1
	Error (criu/cr-dump.c:2111): Dumping FAILED.

To fix this, we set `task_info->checkpointed` before invoking the `checkpoint` action to ensure that the CUDA task is resumed even if CRIU times out.

### DIFF
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -2192,6 +2192,9 @@ int cr_dump_tasks(pid_t pid)
 	if (collect_pstree())
 		goto err;
 
+	if (checkpoint_devices())
+		goto err;
+
 	if (collect_pstree_ids())
 		goto err;
 
--- a/plugins/cuda/cuda_plugin.c
+++ b/plugins/cuda/cuda_plugin.c
@@ -565,6 +565,12 @@ int cuda_plugin_init(int stage)
 {
 	int ret;
 
+	/* Disable CUDA checkpointing with pre-dump */
+	if (stage == CR_PLUGIN_STAGE__PRE_DUMP) {
+		plugin_disabled = true;
+		return 0;
+	}
+
 	if (stage == CR_PLUGIN_STAGE__RESTORE) {
 		if (!check_and_remove_inventory_plugin(CR_PLUGIN_DESC.name, strlen(CR_PLUGIN_DESC.name))) {
 			plugin_disabled = true;
--- a/plugins/cuda/cuda_plugin.c
+++ b/plugins/cuda/cuda_plugin.c
@@ -391,13 +391,13 @@ int cuda_plugin_checkpoint_devices(int pid)
 	if (resume_restore_thread(restore_tid, &save_sigset)) {
 		return -1;
 	}
+
+	task_info->checkpointed = 1;
 	status = cuda_process_checkpoint_action(pid, ACTION_CHECKPOINT, 0, msg_buf, sizeof(msg_buf));
 	if (status) {
 		pr_err("CHECKPOINT_DEVICES failed with %s\n", msg_buf);
 		goto interrupt;
 	}
-
-	task_info->checkpointed = 1;
 
 interrupt:
 	int_ret = interrupt_restore_thread(restore_tid, &save_sigset);
--- a/criu/include/seize.h
+++ b/criu/include/seize.h
@@ -2,6 +2,7 @@
 #define __CR_SEIZE_H__
 
 extern int collect_pstree(void);
+extern int checkpoint_devices(void);
 struct pstree_item;
 extern void pstree_switch_state(struct pstree_item *root_item, int st);
 extern const char *get_real_freezer_state(void);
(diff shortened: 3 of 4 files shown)


avagin:
LGTM

I think we need to move run_plugins(CHECKPOINT_DEVICES) out of collect_pstree(). collrect_pstree shoould just freezer processes.

jesus-ramos:
Unfortunately this problem also requires a driver fix. Due to the way cuda-checkpoint works at the moment killing it via ctrl-c or alarm timeout in the middle of operations can make it get out of sync with later invocations of cuda-checkpoint. Essentially you will get stale operation responses from previous invocations so this may not always be working as intended. Ex: killing it in the middle of a checkpoint operation that then completes behind the scenes the following call to cuda-checkpoint to restore will return the status of the checkpoint rather than the restore. Fixing it requires restarting the target application at the moment which is not very user friendly.

I'll forward the issue internally along though as it's been on the radar to fix for a while.

Patch itself LGTM, and also agree with Andrei's point to move the checkpoint plugin call out of the pstree walk/freeze.

rst0git:
@avagin @jesus-ramos I've updated the pull request with this change, and I was able to confirm that CRIU no longer fails with a timeout error when checkpointing large models.

criu/cr-dump.c
```diff
@@ -1963,6 +1963,9 @@ int cr_pre_dump_tasks(pid_t pid)
 	if (collect_pstree())
 		goto err;
 
+	if (checkpoint_devices())
```
avagin, line 1966: I don't understand why checkpoint_devices() should be called on pre-dump? How should it work?

rst0git: If we remove this line, `criu pre-dump` fails with the following error:

```
(00.381784) Handling VMA with the following smaps entry: 00405000-00406000 rw-p 00004000 fd:00 530240                             /root/cuda-criu-benchmarks/cuda-counter/main
(00.381787) vma 405000 borrows vfi from previous 404000
(00.381791) Handling VMA with the following smaps entry: 1822c000-18da5000 rw-p 00000000 00:00 0                                  [heap]
(00.382372) Handling VMA with the following smaps entry: 200000000-200200000 ---p 00000000 00:00 0
(00.382379) Handling VMA with the following smaps entry: 200200000-200400000 rw-s 00000000 00:06 895                              /dev/nvidia0
(00.382400) Error (criu/proc_parse.c:116): handle_device_vma plugin failed: No such file or directory
(00.382402) Error (criu/proc_parse.c:632): Can't handle non-regular mapping on 705693's map 200200000
(00.382407) Error (criu/cr-dump.c:1486): Collect mappings (pid: 705693) failed with -1
```

rst0git: > How should it work?

To ensure consistency during checkpointing, we need to synchronize the CPU and GPU execution. `checkpoint_devices()` ensures that the GPU state is then locked and copied into host memory. This creates a consistent and unified checkpoint that contains both CPU and GPU states.

avagin: maybe I don't understand something. If we call checkpoint_devices(), the gpu state is loaded to the process memory. Then the process is resumed and so the gpu state is loaded back to GPU and it is cleaned up from the process memory. Then we do another pre-dump or the full dump. The GPU state is loaded again to the process memory and CRIU will dump it again. How can it make any sense?

avagin: I think we need to skip nvidia mappings during pre-dump.

rst0git: You are correct. I will look into how we can skip nvidia mappings during pre-dump.

avagin: It doesn't matter. Here you can remove checkpoint_devices, pause_devices for pre-dump. The actual pre-dump work can be done in a separate pr.

plugins/cuda/cuda_plugin.c
```diff
@@ -565,6 +565,12 @@ int cuda_plugin_init(int stage)
 {
 	int ret;
 
+	/* Disable CUDA checkpointing with pre-dump */
+	if (stage == CR_PLUGIN_STAGE__PRE_DUMP) {
```
avagin, line 569: Form this comment and from the commit message, it is unclear why we need to disable the plugin.

PR #2549, opened 2024-12-17
net: redirect nftables stdout and stderr to log CRIU's log file

adrianreber:
When using the nftables network locking backend and restoring a process a second time the network locking has already been deleted by the first restore. The second restore will print out to the console text like:
```
Error: Could not process rule: No such file or directory
delete table inet CRIU-202621
```
With this change CRIU's log FD is used by linbftables stdout and stderr.


criu/net.c
```diff
@@ -3081,6 +3082,14 @@ static inline int nftables_lock_network_internal(void)
 	if (!nft)
 		return -1;
 
+	fp = fdopen(log_get_fd(), "w");
+	if (!fp) {
+		pr_perror("fdopen() failed");
+		goto err3;
+	}
+	nft_ctx_set_output(nft, fp);
+	nft_ctx_set_error(nft, fp);
```
rst0git, line 3091: The code in `nftables_lock_network_internal()` and `nftables_network_unlock()` looks very similar.

Would it make sense to create a function?
```C
static inline int nftables_set_logfd(struct nft_ctx *nft)
{
	FILE *fp;
	fp = fdopen(log_get_fd(), "w");
	if (!fp) {
		pr_perror("fdopen() failed");
		return -1;
	}
	nft_ctx_set_output(nft, fp);
	nft_ctx_set_error(nft, fp);

	fflush(fp);
	fclose(fp);
	return 0;
}
```
We can use this function in `nftables_lock_network_internal()`:
```C
if (nftables_set_logfd(nft)) {
	goto err2;
}
```
and in `nftables_network_unlock()`:
```C
if (nftables_set_logfd(nft)) {
	nft_ctx_free(nft);
	return -1;
}
```

avagin:
```
The nft_ctx_set_output() and nft_ctx_set_error() functions set the output or error file pointer in ctx to the value of fp. They return the previous value to aid in temporary file pointer overrides. On error, these functions return NULL. 
```

nftables_set_logfd can't close fp...

rst0git:
It looks like the changes in this pull request are also applied in https://github.com/checkpoint-restore/criu/pull/2550
@adrianreber Would it make sense to close this PR?

adrianreber:
@avagin Anything missing here from your point of view. It is not exactly the same as #2550. It is related but could be done independently of the discussion in #2550

criu/net.c
```diff
@@ -3107,6 +3116,9 @@ static inline int nftables_lock_network_internal(void)
 	snprintf(buf, sizeof(buf), "delete table %s", table);
 	NFT_RUN_CMD(nft, buf);
 err2:
+	fflush(fp);
+	fclose(fp);
```
avagin, line 3120: fclose() closes the descriptor that has been returned by log_get_fd(). log_get_fd returns the actual log file descriptor and doesn't dup it, so we should not close it. Do I miss something?

adrianreber: Interesting question. I just followed existing code in CRIU. Not sure. Let me check.

adrianreber: corrected

adrianreber:
Moved everything to a separate function as suggested by @rst0git and removed the `fclose()` as suggested by @avagin.

criu/net.c
```diff
@@ -3066,6 +3066,32 @@ static int iptables_restore(bool ipv6, char *buf, int size)
 	return ret;
 }
 
+#if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
+static inline int redirect_nftables_output(struct nft_ctx *nft)
```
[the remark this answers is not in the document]
adrianreber: The function is only called from code within the same ifdefs.

rst0git (approved):
LGTM

avagin:
> Moved everything to a separate function as suggested by @rst0git and removed the fclose() as suggested by @avagin.

Ooops, I didn't mean that. Without fclose(), it introduces a memory leak, because fdopen allocates a new FILE object. Unfortunately, it is impossible to do fclose() without closing a file description.  The only option is to dup a logging file descriptor and operates with a new descriptor.

adrianreber:
> > Moved everything to a separate function as suggested by @rst0git and removed the fclose() as suggested by @avagin.
> 
> Ooops, I didn't mean that. Without fclose(), it introduces a memory leak, because fdopen allocates a new FILE object. Unfortunately, it is impossible to do fclose() without closing a file description. The only option is to dup a logging file descriptor and operates with a new descriptor.

Thanks for the explanation. Reworked to close the FILE pointers using a `dup()`ed FD.

avagin:
LGTM. Thanks for working on this.

adrianreber:
> > Rebasing the commits of this branch on top of the base branch cannot be performed automatically as this would create a different result than a regular merge.
> 
> @adrianreber could you rebase it?

There is nothing to rebase from what I see. Nothing happens if I try to rebase locally.

rst0git:
@adrianreber GitHub was showing the message `This branch cannot be rebased safely` and merge button was disabled. I rebased the branch on criu-dev and it seems to work now.

PR #2550, opened 2024-12-17
net: remember the name of the lock chain (nftables)

adrianreber:
Using libnftables the chain to lock the network is composed of ("CRIU-%d", real_pid). This leads to around 40 zdtm tests failing with errors like this:
```
Error: No such file or directory; did you mean table 'CRIU-62' in family inet?
delete table inet CRIU-86
```
The reason is that as soon as a process is running in a namespace the real PID can be anything and only the PID in the namespace is restored correctly. Relying on the real PID does not work for the chain name.

Using the PID of the innermost namespace would lead to the chain be called 'CRIU-1' most of the time which is also not really unique.

The uniqueness of the name was always problematic. With this change all tests are working again which rely on network locking if the nftables backend is used for network locking.

### DIFF
--- a/compel/include/uapi/infect-util.h
+++ b/compel/include/uapi/infect-util.h
@@ -3,11 +3,20 @@
 
 #include "common/compiler.h"
 
+/**
+ * The length of the hash is based on what libuuid provides.
+ * According to the manpage this is:
+ *
+ * The uuid_unparse() function converts the supplied UUID uu from the binary
+ * representation into a 36-byte string (plus trailing '\0')
+ */
+#define RUN_ID_HASH_LENGTH 37
+
 /*
  * compel_run_id is a unique value of the current run. It can be used to
  * generate resource ID-s to avoid conflicts with other processes.
  */
-extern uint64_t compel_run_id;
+extern char compel_run_id[RUN_ID_HASH_LENGTH];
 
 struct parasite_ctl;
 extern int __must_check compel_util_send_fd(struct parasite_ctl *ctl, int fd);
--- a/criu/netfilter.c
+++ b/criu/netfilter.c
@@ -299,7 +299,25 @@ int nftables_lock_connection(struct inet_sk_desc *sk)
 
 int nftables_get_table(char *table, int n)
 {
-	if (snprintf(table, n, "inet CRIU-%d", root_item->pid->real) < 0) {
+	int ret;
+
+	switch(dump_criu_run_id[0]) {
+	case 0:
+		/* This is not a restore.*/
+		ret = snprintf(table, n, "inet CRIU-%s", criu_run_id);
+		break;
+	case NO_DUMP_CRIU_RUN_ID:
+		/**
+		 * This is a restore from an older image with no
+		 * dump_criu_run_id available. Let's use the old ID.
+		 */
+		ret = snprintf(table, n, "inet CRIU-%d", root_item->pid->real);
+		break;
+	default:
+		ret = snprintf(table, n, "inet CRIU-%s", dump_criu_run_id);
+	}
+
+	if (ret < 0) {
 		pr_err("Cannot generate CRIU's nftables table name\n");
 		return -1;
 	}
--- a/criu/unittest/mock.c
+++ b/criu/unittest/mock.c
@@ -5,6 +5,8 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+#include "compel/infect-util.h"
+
 int add_external(char *key)
 {
 	return 0;
@@ -141,4 +143,4 @@ int check_mount_v2(void)
 	return 0;
 }
 
-uint64_t compel_run_id;
+char compel_run_id[RUN_ID_HASH_LENGTH];
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -36,7 +36,7 @@ task:
     ln -sf /usr/include/google/protobuf/descriptor.proto images/google/protobuf/descriptor.proto
     dnf config-manager --set-enabled crb # Same as CentOS 8 powertools
     dnf -y install epel-release epel-next-release
-    dnf -y install --allowerasing asciidoc gcc git gnutls-devel libaio-devel libasan libcap-devel libnet-devel libnl3-devel libbsd-devel libselinux-devel make protobuf-c-devel protobuf-devel python-devel python-PyYAML python-protobuf python-junit_xml python3-importlib-metadata xmlto libdrm-devel
+    dnf -y install --allowerasing asciidoc gcc git gnutls-devel libaio-devel libasan libcap-devel libnet-devel libnl3-devel libbsd-devel libselinux-devel make protobuf-c-devel protobuf-devel python-devel python-PyYAML python-protobuf python-junit_xml python3-importlib-metadata xmlto libdrm-devel libuuid-devel
     # The image has a too old version of nettle which does not work with gnutls.
     # Just upgrade to the latest to make the error go away.
     dnf -y upgrade nettle nettle-devel
--- a/.github/workflows/check-commits.yml
+++ b/.github/workflows/check-commits.yml
@@ -19,7 +19,7 @@ jobs:
         # Checkout pull request HEAD commit instead of merge commit
         ref: ${{ github.event.pull_request.head.sha }}
     - name: Install dependencies
-      run: sudo scripts/ci/apt-install libprotobuf-dev libprotobuf-c-dev protobuf-c-compiler protobuf-compiler python3-protobuf libnl-3-dev libnet-dev libcap-dev
+      run: sudo scripts/ci/apt-install libprotobuf-dev libprotobuf-c-dev protobuf-c-compiler protobuf-compiler python3-protobuf libnl-3-dev libnet-dev libcap-dev uuid-dev
     - name: Configure git user details
       run: |
         git config --global user.email "checkpoint-restore@users.noreply.github.com"
--- a/compel/src/lib/infect-util.c
+++ b/compel/src/lib/infect-util.c
@@ -7,7 +7,7 @@
 #include "infect-rpc.h"
 #include "infect-util.h"
 
-uint64_t compel_run_id;
+char compel_run_id[RUN_ID_HASH_LENGTH];
 
 int compel_util_send_fd(struct parasite_ctl *ctl, int fd)
 {
--- a/compel/src/lib/infect.c
+++ b/compel/src/lib/infect.c
@@ -427,7 +427,7 @@ static int gen_parasite_saddr(struct sockaddr_un *saddr, int key)
 	int sun_len;
 
 	saddr->sun_family = AF_UNIX;
-	snprintf(saddr->sun_path, UNIX_PATH_MAX, "X/crtools-pr-%d-%" PRIx64, key, compel_run_id);
+	snprintf(saddr->sun_path, UNIX_PATH_MAX, "X/crtools-pr-%d-%s", key, compel_run_id);
 
 	sun_len = SUN_LEN(saddr);
 	*saddr->sun_path = '\0';
--- a/criu/Makefile.packages
+++ b/criu/Makefile.packages
@@ -6,6 +6,7 @@ REQ-RPM-PKG-NAMES	+= protobuf-devel
 REQ-RPM-PKG-NAMES	+= protobuf-python
 REQ-RPM-PKG-NAMES	+= libnl3-devel
 REQ-RPM-PKG-NAMES	+= libcap-devel
+REQ-RPM-PKG-NAMES	+= libuuid-devel
 
 REQ-RPM-PKG-TEST-NAMES  += libaio-devel
 
@@ -16,6 +17,7 @@ REQ-DEB-PKG-NAMES	+= protobuf-compiler
 REQ-DEB-PKG-NAMES	+= $(PYTHON)-protobuf
 REQ-DEB-PKG-NAMES	+= libnl-3-dev
 REQ-DEB-PKG-NAMES	+= libcap-dev
+REQ-DEB-PKG-NAMES	+= uuid-dev
 
 REQ-DEB-PKG-TEST-NAMES	+= $(PYTHON)-yaml
 REQ-DEB-PKG-TEST-NAMES	+= libaio-dev
@@ -25,7 +27,7 @@ REQ-DEB-PKG-TEST-NAMES	+= libaio-dev
 REQ-RPM-PKG-TEST-NAMES	+= $(PYTHON)-PyYAML
 
 
-export LIBS		+= -lprotobuf-c -ldl -lnl-3 -lsoccr -Lsoccr/ -lnet
+export LIBS		+= -lprotobuf-c -ldl -lnl-3 -lsoccr -Lsoccr/ -lnet -luuid
 
 check-packages-failed:
 	$(warning Can not find some of the required libraries)
--- a/criu/fdstore.c
+++ b/criu/fdstore.c
@@ -58,7 +58,7 @@ int fdstore_init(void)
 	}
 
 	addr.sun_family = AF_UNIX;
-	addrlen = snprintf(addr.sun_path, sizeof(addr.sun_path), "X/criu-fdstore-%" PRIx64 "-%" PRIx64, st.st_ino,
+	addrlen = snprintf(addr.sun_path, sizeof(addr.sun_path), "X/criu-fdstore-%" PRIx64 "-%s", st.st_ino,
 			   criu_run_id);
 	addrlen += sizeof(addr.sun_family);
 
--- a/criu/files.c
+++ b/criu/files.c
@@ -978,7 +978,7 @@ static int receive_fd(struct fdinfo_list_entry *fle);
 static void transport_name_gen(struct sockaddr_un *addr, int *len, int pid)
 {
 	addr->sun_family = AF_UNIX;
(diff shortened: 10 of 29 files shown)


adrianreber:
> A first question I have after going through this is "How this has worked before?".

It probably never did. We are not running all the tests on a system without iptables with nftables locking backend. Only two or four tests are running with the nftables backend.

mihalicyn:
Ah, thanks for clarifications!

I wonder if we can do something like this:
```
$ git diff
diff --git a/criu/netfilter.c b/criu/netfilter.c
index 9e78dc4b0..c558f9bf1 100644
--- a/criu/netfilter.c
+++ b/criu/netfilter.c
@@ -299,7 +299,7 @@ int nftables_lock_connection(struct inet_sk_desc *sk)
 
 int nftables_get_table(char *table, int n)
 {
-       if (snprintf(table, n, "inet CRIU-%d", root_item->pid->real) < 0) {
+       if (snprintf(table, n, "inet CRIU-%d", root_item->ids->pid_ns_id) < 0) {
                pr_err("Cannot generate CRIU's nftables table name\n");
                return -1;
        }
```

Yes, it's not a forward-compatible change and will break restore of images which were dumped with an older CRIU. In this form only works for experimental purposes (and have to check for `root_item->ids->has_pid_ns_id` too). But I'm curious if it helps.

mihalicyn:
My idea is that instead of introducing a new field in `inventory_entry` just for a single purpose, we can use `inventory_entry->root_ids->pid_ns_id` or `root_item->ids->pid_ns_id` as a source of unique criu run id. And we already do something like that when generate `criu_run_id` value:
```
void util_init(void)
{
...
	criu_run_id = getpid();
	if (!stat("/proc/self/ns/pid", &statbuf))
		criu_run_id |= (uint64_t)statbuf.st_ino << 32;
```

adrianreber:
@mihalicyn I am happy to use whatever makes most sense.

What is `pid_ns_id`? Is that basically the inode of the PID NS? Or more? It still sounds like something we need to save somewhere in the image, right?

> Yes, it's not a forward-compatible change and will break restore of images which were dumped with an older CRIU.

I don't think we have to worry about this. Currently it doesn't work at all.

Let me know which ID makes most sense and I can rework this PR. I think the important part is that is has to come from some value of the checkpoint image and not be generated during restore.

adrianreber:
@mihalicyn I think I understood your proposal now. The PR could be really simple as pid_ns_id is already in the image. Let me try it out.

adrianreber:
With this line it also passes all the zdtm test cases (besides a couple of tests which call iptables (which I did not install)) if I switch to the nftables locking backend:
```
if (snprintf(table, n, "inet CRIU-%d", root_item->ids->pid_ns_id) < 0) {
```
That brings it down to a one line change. Very good idea @mihalicyn. Thanks.

How long can the pid_ns_id be? Currently the variable `table` is set to 32 characters.

adrianreber:
@mihalicyn Tests are happy, but `root_item->ids->pid_ns_id` always returns 1 when running in the host PID namespace.

So that is not really a good idea I think as it not really unique.

mihalicyn:
Hey Adrian,

>Is that basically the inode of the PID NS?

Yes, precisely.

>It still sounds like something we need to save somewhere in the image, right?

We don't as we already have it in the image anyways.

>I don't think we have to worry about this. Currently it doesn't work at all.

Are we 100% percent sure that it doesn't work and never worked in any circumstances?

>How long can the pid_ns_id be? Currently the variable table is set to 32 characters.

Hmm, it's `uint32` so in a string representation I guess it's like 10 chars.

>Tests are happy, but root_item->ids->pid_ns_id always returns 1 when running in the host PID namespace.

That's my bad, actually, to get pid namespace inode number you need something like:
```
		ns = lookup_ns_by_id(root_item->ids->pid_ns_id, &pid_ns_desc);
		if (ns) {
			ns->kid // << this thing is inode number of pid namespace
```

But yes, I don't think that even with this change having pid_ns_id would be enough, I think we still need to add a new field to `inventory_entry` but my point is to make it universal for different cases like this one and name it like `criu_run_id` or something like that. Also, to clearly document when it's unique and for what purposes must be used.

mihalicyn:
Also, we have `inventory_entry.dump_uptime` field we can consume to get a certain degree of uniqueness.

adrianreber:
Ah, okay. So let's use the `criu_run_id` and store it in the inventory.

> > I don't think we have to worry about this. Currently it doesn't work at all.

> Are we 100% percent sure that it doesn't work and never worked in any circumstances?

I don't know. All tests with open TCP connections are just hanging during restore because the network locking cannot be disabled. According to zdtm it is so broken that it doesn't work currently.

> Also, we have inventory_entry.dump_uptime field we can consume to get a certain degree of uniqueness.

As an additional field in the nft table name? `"CRIU-%d-%" PRIx64 " ", criu_run_id, inventory_entry.dump_uptime)`

Or instead of `criu_run_id`? Currently we collect the uptime rather late in the process of checkpointing. Definitely after the network locking. It seems to be only used during `detect_pid_reuse()` and that is never directly use, but it seems is only relevant during pre-dump and looking at the parent process. So it seems like we could move the uptime detection to an earlier point and then also use it in the network locking chain. During restore we would need to look at the `criu_run_id` from the checkpointing run and at the uptime of the checkpointing run. That could work.

rst0git:
> > A first question I have after going through this is "How this has worked before?".
> 
> It probably never did. We are not running all the tests on a system without iptables with nftables locking backend. Only two or four tests are running with the nftables backend.

Would it be possible to add a workflow or modify an existing one to run all tests with the nftables backend?

adrianreber:
@mihalicyn What do you think about the latest version. This works in my tests just as good as the previous version. Now using criu_run_id as suggested.

mihalicyn:
Hi Adrian!

> What do you think about the latest version. This works in my tests just as good as the previous version. Now using criu_run_id as suggested.

Looks great! The only thing that worries me is that idea behind `criu_run_id` was to make it relatively unique within one machine/system, but here we try to use it as "unique enough" for purpose of migration between different machines, right? It's not a question to you, but in general. Maybe, we should consider adding more randomness to it? And what is even more important is that as we add `criu_run_id` to the images we loose ability to increase it size (from `uint64_t` to something bigger) in the future.

Also, I tried to play with nftables-based locking mechanism on my machine, and found, that for `host`-flavor test we have no problem with locking:
```
sudo ./test/zdtm.py run --ignore-taint -t zdtm/static/socket-tcp-local -f h
userns is supported
The kernel is tainted: '12288'
=== Run 1/1 ================ zdtm/static/socket-tcp-local
==================== Run zdtm/static/socket-tcp-local in h =====================
Start test
Running zdtm/static/socket-tcp-local.hook(--post-start)
./socket-tcp-local --pidfile=socket-tcp-local.pid --outfile=socket-tcp-local.out
Running zdtm/static/socket-tcp-local.hook(--pre-dump)
State      Recv-Q     Send-Q         Local Address:Port          Peer Address:Port     Process     
LISTEN     0          1                    0.0.0.0:8880               0.0.0.0:*                    
ESTAB      0          0                  127.0.0.1:8880             127.0.0.1:56452                
ESTAB      0          0                  127.0.0.1:56452            127.0.0.1:8880                 
Run criu dump
Running zdtm/static/socket-tcp-local.hook(--pre-restore)
Run criu restore
=[log]=> dump/zdtm/static/socket-tcp-local/55/1/restore.log
b'(00.006320) pie: 55: 55: Restored'
b'(00.006413) Running post-restore scripts'
b'(00.006475) pidfile: Wrote pid 55 to /home/alex/storage/dev/criu/test/zdtm/static/socket-tcp-local.pid (2 bytes)'
b'(00.006480) net: Unlock network'
Running zdtm/static/socket-tcp-local.hook(--post-restore)
State      Recv-Q     Send-Q         Local Address:Port          Peer Address:Port     Process     
LISTEN     0          1                    0.0.0.0:8880               0.0.0.0:*                    
ESTAB      0          0                  127.0.0.1:8880             127.0.0.1:56452                
ESTAB      0          0                  127.0.0.1:56452            127.0.0.1:8880                 
Check TCP images
Send the 15 signal to  55
Wait for zdtm/static/socket-tcp-local(55) to die for 0.100000
Running zdtm/static/socket-tcp-local.hook(--clean)
Removing dump/zdtm/static/socket-tcp-local/55
==================== Test zdtm/static/socket-tcp-local PASS ====================
```

While for `-f ns`/`-f uns` we have the problem you described. BUT! When `-f ns`/`-f uns` is used, it means that locking happens inside net namespace and we can't expect any issues with uniqueness of nftable table names, right? My point is that we can simplify things for the case when net namespace is dumped, while for the case when we don't use namespaces we can still use this simple `pid`-based approach. Tell me if I'm saying something stupid.

mihalicyn:
Just for the sake of demonstrating my point. Somethine like this:
```
--- a/criu/netfilter.c
+++ b/criu/netfilter.c
@@ -14,6 +14,7 @@
 #include "util.h"
 #include "common/list.h"
 #include "files.h"
+#include "namespaces.h"
 #include "netfilter.h"
 #include "sockets.h"
 #include "sk-inet.h"
@@ -299,7 +300,19 @@ int nftables_lock_connection(struct inet_sk_desc *sk)
 
 int nftables_get_table(char *table, int n)
 {
-       if (snprintf(table, n, "inet CRIU-%d", root_item->pid->real) < 0) {
+       int table_id = 0;
+
+       if (root_ns_mask & CLONE_NEWNET) {
+               table_id = 0;
+       } else if (!(root_ns_mask & CLONE_NEWPID)) {
+               table_id = root_item->pid->real;
+       } else {
+               // here we need something unique
+               // as we don't have a net namespace and table name conflict is possible
+               // also, we *do* have a pid namespace and root_item->pid->real makes no sense.
+       }
+
+       if (snprintf(table, n, "inet CRIU-%d", table_id) < 0) {
                pr_err("Cannot generate CRIU's nftables table name\n");
                return -1;
        }
```
fixes tests for `-f h` and `-f ns`/`-f uns`. At the same time, it's broken if you make a setup where process use pid namespace and don't use net namespace.

adrianreber:
> Also, I tried to play with nftables-based locking mechanism on my machine, and found, that for host-flavor test we have no problem with locking:

Right, because `pid_real` is restored correctly.

> While for `-f ns`/`-f uns` we have the problem you described. BUT! When `-f ns`/`-f uns` is used, it means that locking happens inside net namespace and we can't expect any issues with uniqueness of nftable table names, right? My point is that we can simplify things for the case when net namespace is dumped, while for the case when we don't use namespaces we can still use this simple `pid`-based approach. Tell me if I'm saying something stupid.

You are right. It works for `-f h`. I am not sure that having different code paths for host network namespace and separate namespaces is such a good idea. It doesn't seem to break anything if we always use the `criu_run_id`. Having one less `if` is not much, but if we do not gain anything from it, we could leave it away.

> Maybe, we should consider adding more randomness to it?

No real opinion here, but it might be good. No idea.

> And what is even more important is that as we add criu_run_id to the images we loose ability to increase it size (from uint64_t to something bigger) in the future.

We can just deprecate the protobuf field at some point in the future and use a new one if we feel that is necessary.

criu/netfilter.c
```diff
@@ -297,9 +297,9 @@ int nftables_lock_connection(struct inet_sk_desc *sk)
 	return ret;
 }
 
-int nftables_get_table(char *table, int n)
+int nftables_get_table(char *table, int n, uint64_t id)
 {
-	if (snprintf(table, n, "inet CRIU-%d", root_item->pid->real) < 0) {
+	if (snprintf(table, n, "inet CRIU-%" PRIx64, id) < 0) {
```
mihalicyn, line 302:
```
	/*
	 * Keep compatibility with images
	 * without he->dump_criu_run_id field.
	 */
	if (!id) {
		if (!(root_ns_mask & CLONE_NEWPID)) {
			id = root_item->pid->real;
		} else {
			pr_err("Cannot generate CRIU's nftables table name because of issue #2550\n");
			return -1;
		}
		
	}

	if (snprintf(table, n, "inet CRIU-%" PRIx64, id) < 0) {
```
What do you think about this?

mihalicyn (changes requested):

mihalicyn:
> We can just deprecate the protobuf field at some point in the future and use a new one if we feel that is necessary.

I agree.

>I am not sure that having different code paths for host network namespace and separate namespaces is such a good idea.

Yeah, I agree. I just wanted to ensure that I understood problem right and this code example is the best way to show different scenarios we have and when it works and when it doesn't.

But we still need some extra checks for compatibility reasons, IMHO.

criu/net.c
```diff
@@ -3171,18 +3185,44 @@ static inline int nftables_network_unlock(void)
...
+	/**
+	 * Network unlocking can happen during restore, but also
+	 * during checkpointing. It happens during checkpointing
+	 * if the process is kept on running or if there was some
+	 * error. In that case, checkpointing, dump_criu_run_id
+	 * is not set and we have to use criu_run_id.
+	 */
 
-	if (nftables_get_table(table, sizeof(table)))
+	if (dump_criu_run_id == 0)
```
mihalicyn, line 3199: I think we need to introduce a boolean parameter here to determine if we are on restore/dump codepath as `dump_criu_run_id == 0` might be in two different cases: when we deal with an old image (without `dump_criu_run_id` field) or if we are on the restore codepath.

avagin:
> > A first question I have after going through this is "How this has worked before?".
> 
> It probably never did. We are not running all the tests on a system without iptables with nftables locking backend. Only two or four tests are running with the nftables backend.

It was broken by 9838d34ded.

avagin:
> > And what is even more important is that as we add criu_run_id to the images we loose ability to increase it size (from uint64_t to something bigger) in the future.
> 
> We can just deprecate the protobuf field at some point in the future and use a new one if we feel that is necessary.

We can do that, but we are trying to avoid that. If we see the problem, it is better to handle it right away. The obvious solution here is to add a `true` uuid, isn't it?

adrianreber:
> > > And what is even more important is that as we add criu_run_id to the images we loose ability to increase it size (from uint64_t to something bigger) in the future.
> > 
> > 
> > We can just deprecate the protobuf field at some point in the future and use a new one if we feel that is necessary.
> 
> We can do that, but we are trying to avoid that. If we see the problem, it is better to handle it right away. The obvious solution here is to add a `true` uuid, isn't it?

What do you exactly suggest? Can you be more specific? Not sure what you mean by `true`. I assume you mean it should be really unique, right? Should we base that on the existing `criu_run_id`?

avagin:
> > We can do that, but we are trying to avoid that. If we see the problem, it is better to handle it right away. The obvious solution here is to add a `true` uuid, isn't it?
> 
> What do you exactly suggest? Can you be more specific? Not sure what you mean by `true`. I assume you mean it should be really unique, right? Should we base that on the existing `criu_run_id`?

I mean we can consider to use the standard way to generate uuid (libuuid).

compel/include/uapi/infect-util.h
```diff
@@ -7,7 +7,7 @@
  * compel_run_id is a unique value of the current run. It can be used to
  * generate resource ID-s to avoid conflicts with other processes.
  */
-extern uint64_t compel_run_id;
+extern char compel_run_id[37];
```
rst0git, line 10: Would it be possible to define `37` somewhere as a constant?

adrianreber: added a define

compel/include/uapi/infect-util.h
```diff
@@ -3,11 +3,20 @@
...
 #include "common/compiler.h"
 
+/**
+ * The length of the hash is based on what libuuid provides.
+ * According to the manpage this is:
+ *
+ * The uuid_unparse() function converts the supplied UUID uu from the binary
+ * representation into a 36-byte string (plus trailing '\0')
+ */
+#define RUN_ID_HASH_LENGTH 36+1
```
avagin, line 13: brackets around 36+1 are required (37 should work too). Otherwise expressions like 2 * RUN_ID_HASH_LENGTH don't work as expected.

criu/unittest/mock.c
```diff
@@ -141,4 +141,4 @@ int check_mount_v2(void)
 	return 0;
 }
 
-uint64_t compel_run_id;
+char compel_run_id[37];
```
avagin, line 144: RUN_ID_HASH_LENGTH

avagin:
LGTM

compel/include/uapi/infect-util.h
adrianreber: done

criu/unittest/mock.c
adrianreber: done

adrianreber:
@mihalicyn PTAL

criu/netfilter.c
```diff
@@ -299,7 +299,25 @@ int nftables_lock_connection(struct inet_sk_desc *sk)
 
 int nftables_get_table(char *table, int n)
 {
-	if (snprintf(table, n, "inet CRIU-%d", root_item->pid->real) < 0) {
+	int ret;
+
+	switch(dump_criu_run_id[0]) {
+	case 0:
+		/* This is not a restore.*/
+		ret = snprintf(table, n, "inet CRIU-%s", criu_run_id);
```
rst0git, line 307: What does `"This is not a restore"` mean here?

adrianreber: If doing a dump the name of the table is based on the current criu_run_id and not on the one from the previous dump. It means, that this is not a restore but a dump.

PR #2553, opened 2024-12-19
seize: Adjust the position of the log message

yummypeng:
Based on the code, the `ret` variable at this point does not represent the task state, so this log message should be moved to a position after the `compel_wait_task()` function.


avagin:
LGTM. thanks.

rst0git (approved):

PR #2554, opened 2024-12-19
cr: Task CapAmb support

liuchao173:
criu only dump and restore Ing, Prm, Eff, Bnd. Add the process of dumping and restoring the CapAmb by referring to the existing implementation.


images/creds.proto
```diff
@@ -16,13 +16,14 @@ message creds_entry {
 	repeated uint32	cap_prm	= 10;
 	repeated uint32	cap_eff	= 11;
 	repeated uint32	cap_bnd	= 12;
+	repeated uint32 cap_amb = 13;
```
adrianreber, line 19: You cannot change protobuf IDs. That breaks backwards compatibility. Always add additional fields at the end.

adrianreber:
Take a look at the CI failures. Looks like your proposed change breaks a couple of our tests. They should not break.

images/creds.proto
liuchao173: done

liuchao173:
> Take a look at the CI failures. Looks like your proposed change breaks a couple of our tests. They should not break.

Fixed. The rest tests do not seem to be related to this pull request.

avagin:
You need to introduce a test (test/zdtm/static/) and write a commit message.

criu/pie/restorer.c
```diff
@@ -347,6 +347,21 @@ static int restore_creds(struct thread_creds_args *args, int procfd, int lsm_typ
...
+		for (i = 0; i < 32; i++) {
+			if (b * 32 + i > args->cap_last_cap)
+				break;
+			if ((args->cap_amb[b] & (1 << i)) == 0)
+				/* don't set */
+				continue;
+			ret = sys_prctl(PR_CAP_AMBIENT, PR_CAP_AMBIENT_RAISE, i + b * 32, 0, 0);
+			if (!ret)
+				continue;
+			pr_warn("Unable to raise ambient capability %d: %d\n", i + b * 32, ret);
```
avagin, line 360: why do we ignore errors here?

liuchao173:
> You need to introduce a test (test/zdtm/static/) and write a commit message.

done

PR #2562, opened 2025-01-08
Fix one shot itimer

hckuo:


avagin:
The test passes without the fix:
```
$ git log HEAD~3... --pretty=oneline
36bf4a056f3c153baf2b130cac6df8f62835e3b4 (HEAD) Revert "Update timer handling to properly respect the it_value field"
c20a4cfa3a403558dd702942cf02b51b36c01f30 test: zdtm: Enhance timers test with multiple timer configurations
3249a8308806ba9e5a0dc0501268c5b40833d721 Update timer handling to properly respect the it_value field
$ make -j 4
$ python test/zdtm.py run -t zdtm/static/timers --ignore-taint -f h
userns is supported
The kernel is tainted: '12352'
=== Run 1/1 ================ zdtm/static/timers
========================= Run zdtm/static/timers in h ==========================
Start test
./timers --pidfile=timers.pid --outfile=timers.out
Run criu dump
Run criu restore
Send the 15 signal to  52
Wait for zdtm/static/timers(52) to die for 0.100000
Wait for zdtm/static/timers(52) to die for 0.200000
Wait for zdtm/static/timers(52) to die for 0.400000
Wait for zdtm/static/timers(52) to die for 0.800000
Wait for zdtm/static/timers(52) to die for 1.600000
Removing dump/zdtm/static/timers/52
========================= Test zdtm/static/timers PASS =========================
```

I think the intention was to introduce the regression test, wasn't it?

hckuo:
I think it has something to do with how the test is setup. If I only test
the ones hot timers it would fail as expected. I can duplicate the test
file so one for regular timers and one for one shot timers.  However this
is something that I want to avoid to not have duplicate code. Please
instruct on a better way if any. Or I can just duplicate the test file and
make it two tests.

hckuo:
```
ubuntu@ip-172-31-13-102:~/criu$ git diff
diff --git a/test/zdtm/static/timers.c b/test/zdtm/static/timers.c
index 5a919f2b5..95004f384 100644
--- a/test/zdtm/static/timers.c
+++ b/test/zdtm/static/timers.c
@@ -87,7 +87,7 @@ int main(int argc, char **argv)
        };
 
        test_init(argc, argv);
-       for (int i = 0; i < 2; i++) {
+       for (int i = 1; i < 2; i++) {
                setup_timers(&tvs[i]);
 
                test_daemon();
ubuntu@ip-172-31-13-102:~/criu$ sudo python3 test/zdtm.py run -t zdtm/static/timers --ignore-taint -f h
userns is supported
=== Run 1/1 ================ zdtm/static/timers
========================= Run zdtm/static/timers in h ==========================
 DEP       timers.d
 CC        timers.o
 DEP       parseargs.d
 CC        parseargs.o
 AR        libzdtmtst.a
 LINK      groups
 LINK      timers
Start test
./timers --pidfile=timers.pid --outfile=timers.out
Run criu dump
Run criu restore
Send the 15 signal to  116
Wait for zdtm/static/timers(116) to die for 0.100000
Wait for zdtm/static/timers(116) to die for 0.200000
Wait for zdtm/static/timers(116) to die for 0.400000
Wait for zdtm/static/timers(116) to die for 0.800000
Wait for zdtm/static/timers(116) to die for 1.600000
Wait for zdtm/static/timers(116) to die for 3.200000
Wait for zdtm/static/timers(116) to die for 6.400000
==== ALARM ====
################# Test zdtm/static/timers FAIL at result check #################
22:21:52.061:   116: FAIL: timers.c:68: timer 0 stuck (errno = 11 (Resource temporarily unavailable))

```

criu/timer.c
```diff
@@ -29,17 +29,8 @@ static inline int decode_itimer(char *n, ItimerEntry *ie, struct itimerval *val)
 		return -1;
 	}
 
-	if (ie->vsec == 0 && ie->vusec == 0) {
```
avagin, line 32: This code handles the case when the timer is armed, but it_value is set to 0:
https://man7.org/linux/man-pages/man2/setitimer.2.html
```
 If  both fields of it_value are zero, then this timer is currently disarmed (inactive).
```

criu/pie/restorer.c
```diff
@@ -2210,7 +2210,7 @@ __visible long __export_restore_task(struct task_restore_args *args)
 	 * code below doesn't fail due to bad timing values.
 	 */
 
-#define itimer_armed(args, i) (args->itimers[i].it_interval.tv_sec || args->itimers[i].it_interval.tv_usec)
+#define itimer_armed(args, i) (args->itimers[i].it_interval.tv_sec || args->itimers[i].it_interval.tv_usec || args->itimers[i].it_value.tv_sec || args->itimers[i].it_value.tv_usec)
```
avagin, line 2213: nit: The limit on the length of lines is 80 columns and this is a strongly preferred limit.

avagin: actually, you need to check just it_value.tv_sec and it_value.tv_usec:
```
#define itimer_armed(args, i) (args->itimers[i].it_value.tv_sec || args->itimers[i].it_value.tv_usec)
```

If this two fields are zero, setitimer disarms the timer.

test/zdtm/static/timers.c
```diff
@@ -77,13 +73,27 @@ static void check_timers(void)
...
 
-	setup_timers();
+	test_init(argc, argv);
+	for (int i = 0; i < 2; i++) {
+		setup_timers(&tvs[i]);
 
-	test_daemon();
-	test_waitsig();
+		test_daemon();
+		test_waitsig();
```
avagin, line 94: It will not work. test_waitsig does its job when it is called first time. If you call it second time, it returns immediately:
https://github.com/checkpoint-restore/criu/blob/27a5b9aa87d6f54ff837a406d4ab6a20ab170302/test/zdtm/lib/test.c#L365

I suggest to add a new test. Here is my version of it: https://github.com/avagin/criu/commit/c02e31345a90e37a615fa9a60b18c24b61837bce

It fails without the your fix and it passes with it:
```

# ./test/zdtm.py run -t zdtm/static/timers01 --ignore-taint
userns is supported
The kernel is tainted: '12352'
=== Run 1/1 ================ zdtm/static/timers01
======================= Run zdtm/static/timers01 in uns ========================
Start test
./timers01 --pidfile=timers01.pid --outfile=timers01.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/timers01/64/1/restore.log
b'(00.009848)      1: No ipcns-sem-11.img image'
b'(00.018559)      1: net: Try to restore a link 10:1:lo'
b'(00.018566)      1: net: Restoring link lo type 1'
b'(00.021284)      1: net: \tRunning ip addr restore'
Send the 15 signal to  92
Wait for zdtm/static/timers01(92) to die for 0.100000
################ Test zdtm/static/timers01 FAIL at result check ################
17:44:11.873:     4: FAIL: timers01.c:54: 0 isn't in [3600, 3300] (errno = 11 (Resource temporarily unavailable))

```

criu/pie/restorer.c
hckuo: yes

test/zdtm/static/timers.c
hckuo: thanks!

avagin:
Merged. Thanks.

PR #2569, opened 2025-01-22
.gitignore: Add compilation database file and directory

haiyuewa:
The tool 'bear'[1] can be used to generate the C language server protocol database file while compiling: 'bear -- make'

Add the database file into .gitignore as Linux Kernel [2] does.

The clangd will parse the compile_commands.json language metadata into the .cache directory. Add it for clean git status track.

[1]: https://github.com/rizsotto/Bear
[2]: https://git.kernel.org/torvalds/c/26c4c71bcd9a

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,7 @@ compel/include/asm
 include/common/asm
 include/common/config.h
 build/**
+
+# Clang's compilation database file and directory.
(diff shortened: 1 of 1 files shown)


haiyuewa:
No Reviewers, and looks like no need for this PR.

PR #2583, opened 2025-02-04
make: allow setting the default network locking backend

adrianreber:


INSTALL.md
```diff
@@ -1,3 +1,23 @@
...
+
+First, you need to install compile-time dependencies. Check [Installation dependencies](https://criu.org/Installation#Dependencies) for more info.
+
+To compile CRIU, run:
+```
+    make
+```
+This should create the `./criu/criu` executable.
+
+To change the default behaviour of CRIU following variables can be
```
rst0git, line 11:
```Suggestion
To change the default behaviour of CRIU, the following variables can be
```

INSTALL.md
```diff
@@ -1,3 +1,23 @@
...
+passed to the make command:
+
+ * **NETWORK_LOCK_DEFAULT**, can be set to one of the following
+   values: *NETWORK_LOCK_IPTABLES*, *NETWORK_LOCK_NFTABLES*,
```
rst0git, line 15:
```Suggestion
   values: `NETWORK_LOCK_IPTABLES`, `NETWORK_LOCK_NFTABLES`,
```

INSTALL.md
```diff
@@ -1,3 +1,23 @@
...
+   *NETWORK_LOCK_SKIP*. CRIU defaults to *NETWORK_LOCK_IPTABLES*
```
rst0git, line 16:
```Suggestion
   `NETWORK_LOCK_SKIP`. CRIU defaults to `NETWORK_LOCK_IPTABLES`
```

INSTALL.md
```diff
@@ -1,3 +1,23 @@
...
+   if nothing is specified. If another network locking backend is
+   needed `make` can be called like this:
```
rst0git, line 18:
```Suggestion
   needed, `make` can be called like this:
```

INSTALL.md
```diff
@@ -1,3 +1,23 @@
+## Building CRIU from source code
+
+First, you need to install compile-time dependencies. Check [Installation dependencies](https://criu.org/Installation#Dependencies) for more info.
+
+To compile CRIU, run:
+```
+    make
```
rst0git, line 7: Would it be possible to remove the 4-space indent in the code blocks?

adrianreber:
@mihalicyn That adds a compile time option to easily switch between backends.

adrianreber:
@danishprakash this is a PR which should help you building RPMs with a different network locking backend.

danishprakash:
Thanks for this @adrianreber, this is quite helpful.

---

I remember we had this discussion in https://github.com/containers/podman/issues/24799 but if we were to make criu "intelligent" and have it figure out the network locking backend on its own, what issues could we have encountered?

criu/include/cr_options.h
```diff
@@ -70,7 +70,15 @@ enum NETWORK_LOCK_METHOD {
...
 };
 
+/**
+ * CRIU currently defaults to the iptables locking backend.
+ *
+ * It is, however, possible to change this by defining
+ * NETWORK_LOCK_DEFAULT to a different value on the command-line.
+ */
+#ifndef NETWORK_LOCK_DEFAULT
 #define NETWORK_LOCK_DEFAULT NETWORK_LOCK_IPTABLES
```
avagin, line 80: should we change the default to nftable?

avagin:
LGTM. Thanks!

criu/include/cr_options.h
adrianreber: I don't know. Currently Fedora >= 42 defaults to nft and CentOS Stream >= 10 (RHEL >= 10).

@danishprakash mentioned that Suse is also looking into switching to nft. Not sure what other distributions are planing. 

Maintaining the RHEL and Fedora packages it would make my life a bit easier, but I still need to deal with older Fedora and RHEL releases. It feels a bit too early. Maybe if Debian or Ubuntu also switches to nft, the we would have most of the bigger distributions.

criu/cr-service.c
```diff
@@ -701,6 +702,8 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	if (req->lsm_profile) {
 		opts.lsm_supplied = true;
 		SET_CHAR_OPTS(lsm_profile, req->lsm_profile);
+		pr_debug("opts.lsm_supplied %d\n", opts.lsm_supplied);
```
rst0git, line 705: It seems like `opts.lsm_supplied` is always `true` here.

adrianreber: I don't think this belongs in this PR at all. I have to remove this.

adrianreber: removed

PR #2584, opened 2025-02-04
optimize: make mntinfo_add_list O(1) by adding a tail pointer

ankushT369:
The `mntinfo_add_list` function previously traversed the entire linked list to find the tail node, resulting in a **time complexity** of **O(n)**. This commit introduces a `mntinfo_tail` pointer to track the last node of the list, reducing the **time complexity** of insertion to **O(1)**.

This change addresses the `FIXME` comment in the code and improves performance for large lists.

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -110,18 +110,17 @@ static char *ext_mount_lookup(char *key)
  * Single linked list of mount points get from proc/images
  */
 struct mount_info *mntinfo;
+struct mount_info *mntinfo_tail;
 
 static void mntinfo_add_list(struct mount_info *new)
 {
-	if (!mntinfo)
+	if (!mntinfo) {
 		mntinfo = new;
+        mntinfo_tail = mntinfo;
+    }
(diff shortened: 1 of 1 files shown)


ankushT369:
Could someone please review this PR? If there are any issues, could you kindly point them out and suggest how I can resolve them?

adrianreber:
If you look at the output of the CI runs you can see that all mount related tests are now failing.

ankushT369:
Thankyou for replying @adrianreber yea I am getting some mount related test failing. But in my system it worked fine and tests are passed. How will I know about other systems and rectify that according to that (apologies if it's a very beginner type question I am new to this).

adrianreber:
Good to heat that tests are passing on your system, but make sure they also work in our different CI setups.

> How will I know about other systems

Just have a look what is running in CI

> and rectify that according to that

Unfortunately I do not know. Maybe use a container to reproduce the different setups we have in CI.

PR #2587, opened 2025-02-06
lsm: use the user provided lsm label

adrianreber:
Currently CRIU has the possibility to specify a LSM label during restore. Unfortunately the information is completely ignored in the case of SELinux.

This change selects the lsm label from the user if it is provided and else the label from the checkpoint image is used.


adrianreber:
Fedora Rawhide currently is broken due to the branching of Fedora 42.

PR #2593, opened 2025-02-07
mnt: Optimize mntinfo_add_list to O(1) using a tail pointer 

ankushT369:
This patch optimizes the `mntinfo_add_list` function from O(n) to O(1) by introducing a tail pointer (`tailbuffer`) to track the end of the linked list. Previously, adding a new `mount_info` node required traversing the entire list to find the tail, which was inefficient for large lists.

Changes:
1. Added a `struct list_head tail` to the `mount_info` struct to maintain the tail pointer.
2. Introduced a global `tailbuffer` to store the current tail of the list.
3. Modified `mntinfo_add_list` to use `tailbuffer` for O(1) insertion at the tail.
4. Updated `mntinfo_add_list_before` to initialize the tail pointer and maintain consistency when adding nodes at the head.

The optimization significantly improves performance for operations involving large lists of mount information. This change also ensures that the list remains consistent and correctly updated during insertions.

### DIFF
--- a/criu/include/mount.h
+++ b/criu/include/mount.h
@@ -90,6 +90,7 @@ struct mount_info {
 	int deleted_level;
 	struct list_head deleted_list;
 	struct mount_info *next;
+        struct list_head tail;
 	struct ns_id *nsid;
 
 	char *external;
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -110,23 +110,29 @@ static char *ext_mount_lookup(char *key)
  * Single linked list of mount points get from proc/images
  */
 struct mount_info *mntinfo;
+struct mount_info *tailbuffer;
 
 static void mntinfo_add_list(struct mount_info *new)
 {
 	if (!mntinfo)
 		mntinfo = new;
 	else {
-		struct mount_info *pm;
+                struct list_head *_tail = mntinfo->tail.next;
+                tailbuffer = list_entry(_tail, struct mount_info, tail);
 
-		/* Add to the tail. (FIXME -- make O(1) ) */
-		for (pm = mntinfo; pm->next != NULL; pm = pm->next)
-			;
-		pm->next = new;
+                tailbuffer->next = new;
+                mntinfo->tail = new->tail;
 	}
 }
 
 void mntinfo_add_list_before(struct mount_info **head, struct mount_info *new)
 {
+        if (!*head)
+                tailbuffer = new;
+
+        INIT_LIST_HEAD(&new->tail);
+        list_add_tail(&tailbuffer->tail, &new->tail);
+
 	new->next = *head;
 	*head = new;
 }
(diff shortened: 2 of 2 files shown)


criu/mount.c
```diff
@@ -113,21 +113,25 @@ struct mount_info *mntinfo;
...
+                mntinfo->tail = new->tail;
 	}
 }
 
+struct mount_info *tail_buffer;
+
 void mntinfo_add_list_before(struct mount_info **head, struct mount_info *new)
 {
-	new->next = *head;
+    if(!*head)
```
avagin: pls use tab indents.

criu/include/mount.h
```diff
@@ -90,6 +90,7 @@ struct mount_info {
 	int deleted_level;
 	struct list_head deleted_list;
 	struct mount_info *next;
+	struct mount_info *tail;
```
avagin, line 93: I think it is better to user the list_head structure here.

criu/mount.c
```diff
@@ -1959,7 +1959,7 @@ static int dump_mnt_ns(struct ns_id *ns, struct mount_info *pms)
 				_mi = list_entry(_mi->children._el, struct mount_info, siblings); \
 				continue;                                                         \
 			}                                                                         \
-		up:                                                                               \
+up:                                                                                               \
```
avagin: you should not change unrelated code...

ankushT369:
@adrianreber can you review this PR? I tried to do some small optimization

criu/mount.c
ankushT369: I'm sorry for that it was a mistake

criu/include/mount.h
ankushT369: You mean in place of mount_info *tail should I make it list_head tail

ankushT369:
@avagin all over what I have to change in the main logic??

criu/include/mount.h
avagin: I mean we can use the standard list here instead of reinventing a bicycle.

ankushT369: I'm sorry if this conversation is taking long..  the thing is the `list_head` is of different type in order to store the tail of a `struct mount_info*` shouldn't we use the  `struct mount_info*` type?? because in the code mount.c (where I have changed code) the `struct mount_info *mntinfo` was the single linked list as it was told in the comment (file : mount.c, line : 110)  and the `struct mount_info *new` has to traverse the `struct mount_info *mntinfo` (file : mount.c, line : 122) in order to find the tail of it ( if the `*mntinfo` wasn't empty) thats why I created a `struct mount_info* tail` in order to hold the pointer to tail. Any node of type `struct mount_info*` can have its tail stored in it so that it can fetch in O(1) time. I know its reinventing a bicycle but the thing is I can't understand how can I use the `struct list_head` cause there is a mismatch in type in order to hold the tail of anything of type `struct mount_info*`. Maybe I can wrong here by misunderstanding I you tried to say you can rectify me I am a beginner currently learning things.

ankushT369:
ok got it I am gonna fix this!!

criu/include/mount.h
```diff
@@ -90,6 +90,7 @@ struct mount_info {
 	int deleted_level;
 	struct list_head deleted_list;
 	struct mount_info *next;
+        struct list_head tail;
```
avagin, line 93: You need to replace `next` and `tail` with `struct list_head node` and then use standard operation like list_add, list_del, list_for_each_entry (./include/common/list.h).

ankushT369: okk I will

PR #2599, opened 2025-02-17
mnt: Optimize mntinfo_add_list to O(1) using a tail pointer

ankushT369:
This patch optimizes the `mntinfo_add_list` function from O(n) to O(1) by introducing a tail pointer (`tailbuffer`) to track the end of the linked list. Previously, adding a new `mount_info` node required traversing the entire list to find the tail, which was inefficient for large lists.

Changes:
1. Added a `struct list_head tail` to the `mount_info` struct to maintain the tail pointer.
2. Introduced a global `tailbuffer` to store the current tail of the list.
3. Modified `mntinfo_add_list` to use `tailbuffer` for O(1) insertion at the tail.
4. Updated `mntinfo_add_list_before` to initialize the tail pointer and maintain consistency when adding nodes at the head.

The optimization significantly improves performance for operations involving large lists of mount information. This change also ensures that the list remains consistent and correctly updated during insertions.

### DIFF
--- a/criu/include/mount.h
+++ b/criu/include/mount.h
@@ -90,6 +90,7 @@ struct mount_info {
 	int deleted_level;
 	struct list_head deleted_list;
 	struct mount_info *next;
+        struct list_head tail;
 	struct ns_id *nsid;
 
 	char *external;
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -110,23 +110,29 @@ static char *ext_mount_lookup(char *key)
  * Single linked list of mount points get from proc/images
  */
 struct mount_info *mntinfo;
+struct mount_info *tailbuffer;
 
 static void mntinfo_add_list(struct mount_info *new)
 {
 	if (!mntinfo)
 		mntinfo = new;
 	else {
-		struct mount_info *pm;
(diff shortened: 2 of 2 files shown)


ankushT369:
@avagin can you please review this PR?

mihalicyn:
Hi @ankushT369,

as far as I understand you work on the same fix in https://github.com/checkpoint-restore/criu/pull/2593 
Why you created a second PR?

mihalicyn:
hint: you can use `git rebase -i` command to interactively edit your commits and then `git push -f` to update your existing PR the way you want.

criu/include/mount.h
```diff
@@ -90,6 +90,7 @@ struct mount_info {
 	int deleted_level;
 	struct list_head deleted_list;
 	struct mount_info *next;
+        struct list_head tail;
```
mihalicyn, line 93: if you add this `struct list_head` member here, then you don't need to have `struct mount_info *next` member anymore as you can use `tail.next` to access to the next `struct mount_info` pointer, isn't it?

ankushT369:
@mihalicyn ok I will!!

Issue #2607, opened 2025-03-03
make uninstall fails with ModuleNotFoundError: signal


rst0git:
When running with Python 3.13, `make uninstall PREFIX=/usr` fails with the following error:
```
 UNINSTALL pycriu
Traceback (most recent call last):
  File "/criu/./scripts/uninstall_module.py", line 66, in <module>
    uninstall_module(args.module_name, args.prefix)
  File "/criu/./scripts/uninstall_module.py", line 49, in uninstall_module
    subprocess.check_call(command, env=os.environ)
  File "/usr/lib64/python3.13/subprocess.py", line 416, in check_call
    retcode = call(*popenargs, **kwargs)
  File "/usr/lib64/python3.13/subprocess.py", line 397, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/usr/lib64/python3.13/subprocess.py", line 1038, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/lib64/python3.13/subprocess.py", line 1859, in _execute_child
    self._posix_spawn(args, executable, env, restore_signals, close_fds,
  File "/usr/lib64/python3.13/subprocess.py", line 1775, in _posix_spawn
    import signal
ModuleNotFoundError: No module named 'signal'
```
This error occurs because the `signal` module is located in `/usr/lib64/python3.13/signal.py` and this path is not included in the list containing site-packages directories used by [uninstall_module.py](https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/uninstall_module.py).

avagin:
How do you install python 3.13?

rst0git:
Fedora 41 comes with Python 3.13

PR #2608, opened 2025-03-03
scripts/uninstall_module: import signal module

rst0git:
With Python 3.13, the `subprocess` module now uses the `posix_spawn()` function [^1], which requires the `signal` module to be imported.

Fixes: #2607

[^1]: https://docs.python.org/3/whatsnew/3.13.html#subprocess

### DIFF
--- a/scripts/uninstall_module.py
+++ b/scripts/uninstall_module.py
@@ -10,6 +10,16 @@ import site
 import subprocess
 import sys
 
+# With Python 3.13 the subprocess module now uses the `posix_spawn()`
+# function which requires loading the `signal` module:
+#     https://docs.python.org/3/whatsnew/3.13.html#subprocess
+#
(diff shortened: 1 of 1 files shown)


scripts/uninstall_module.py
```diff
@@ -10,6 +10,12 @@
 import subprocess
 import sys
 
+# With Python 3.13 the subprocess module uses `posix_spawn()` function
+# which requires loading the `signal` module:
+#     https://docs.python.org/3/whatsnew/3.13.html#subprocess
+# flake8: noqa: F401
```
avagin, line 16: From this comment, it is unclear why this module should be imported here.

rst0git: I've updated the comment to clarify the reason why this module should be imported here.

PR #2613, opened 2025-03-10
Dmabuf IPC support for amdgpu plugin (v2)

fdavid-amd:
This patch set, in combination with the accompanying kernel patch set, is the
second draft of work to support amdgpu dmabuf IPC. Since the first post, the
patches have been split to make them more readable and made
backwards-compatible.


criu/files.c
```diff
@@ -836,6 +836,9 @@ struct fdinfo_list_entry *collect_fd_to(int pid, FdinfoEntry *e, struct rst_info
 {
 	struct fdinfo_list_entry *new_le;
 
+	if (fdesc->ops->type == FD_TYPES__EXT)
+		run_plugins(COLLECT_FILE, pid, fdesc->id);
```
avagin, line 840: need to check a return code

criu/cr-dump.c
```diff
@@ -2225,6 +2225,9 @@ int cr_dump_tasks(pid_t pid)
 			goto err;
 	}
 
+	if(run_plugins(DUMP_DEVICE_LATE, pid))
```
avagin, line 2228: run_plugins() returns -ENOTSUP if no hook is registered. This is why ci tests fail on this pr.

criu/servicefd.c
```diff
@@ -312,5 +313,15 @@ int clone_service_fd(struct pstree_item *me)
 	service_fd_id = id;
 	ret = 0;
 
+	next_high_fd = service_fd_base + 1024;
```
avagin, line 316: Where does this magic number 1024 come from? Perhaps I've missed something, but it doesn't seem safe to assume that next_high_fd will be available and won't conflict with other descriptors. I feel like a detailed comment is needed here.

criu/servicefd.c
```diff
@@ -312,5 +313,15 @@ int clone_service_fd(struct pstree_item *me)
...
+
 	return ret;
 }
+
+int get_unused_high_fd(void)
+{
+	if (next_high_fd > service_fd_rlim_cur)
+		return -1;
+	next_high_fd += 1;
```
avagin, line 325: Why don't you use the fdstore for temporary file descriptors? It was created specifically for this purpose.

avagin:
@fdavid-amd how is it going? Let me know if you need help with this pr.

plugins/amdgpu/criu-amdgpu.proto
```diff
@@ -61,6 +62,29 @@ message criu_kfd {
...
+message drm_vm_entry {
+	required uint64 start = 1;
+	required uint64 last = 2;
+	required uint64 offset = 3;
+	required uint64 flags = 4;
+}
+
 message criu_render_node {
 	required uint32 gpu_id = 1;
+	required uint32 id = 2;
```
rst0git, line 87: It looks like these changes break compatibility - a newer version of CRIU would not be able to restore a checkpoint created with a previous version.

plugins/amdgpu/amdgpu_plugin_util.c
```diff
@@ -49,37 +51,149 @@ struct tp_system dest_topology;
...
-void init_gpu_count(struct tp_system *topo)
-{
-	if (dev_file_cnt != 0)
-		return;
+bool shared_bo_has_exporter(int handle) {
+	struct shared_bo *bo;
 
-	/* We add ONE to include checkpointing of KFD device */
-	dev_file_cnt = 1 + topology_gpu_count(topo);
+	if (handle == -1) {
```
dayatsin-amd: We do not need the curly braces here.

plugins/amdgpu/amdgpu_plugin_util.c
```diff
@@ -49,37 +51,149 @@ struct tp_system dest_topology;
...
+
+int read_fp(FILE *fp, void *buf, const size_t buf_len)
+{
+       size_t len_read;
+
+       len_read = fread(buf, 1, buf_len, fp);
+       if (len_read != buf_len) {
+               pr_err("Unable to read file (read:%ld buf_len:%ld)\n", len_read, buf_len);
+               return -EIO;
+
```
dayatsin-amd: extra lines

criu/include/criu-plugin.h
```diff
@@ -64,6 +64,8 @@ enum {
 
 	CR_PLUGIN_HOOK__COLLECT_FILE = 13,
 
+	CR_PLUGIN_HOOK__DUMP_DEVICE_LATE = 14,
```
dayatsin-amd: nit pick. This should be CR_PLUGIN_HOOK__DUMP_DEVICES_LATE to match the other callbacks using plural.

dayatsin-amd (changes requested):

criu/servicefd.c
fdavid-amd: Thanks, didn't know about that. It seems like it would work for the dmabuf_fds, but I would still need to put the socket fd itself somewhere.

avagin: fdstore is already there, you can use it (fdid = fdstore_add(fd); ....; fd =  fdstore_get(fdis). In another comment, I explained why the "high_fd" hack will not work in some case.

fdavid-amd:
@avagin 
Sorry for leaving this untouched for so long, I was revising the kernel side of the changes. Could you take a look at this again?

criu/cr-dump.c
fdavid-amd: Fixed

criu/files.c
fdavid-amd: Fixed

plugins/amdgpu/criu-amdgpu.proto
fdavid-amd: That is true - I wasn't aware that was a type of backwards-compatibility we committed to.

plugins/amdgpu/amdgpu_plugin_util.c
fdavid-amd: Fixed

plugins/amdgpu/amdgpu_plugin_util.c
fdavid-amd: Fixed

fdavid-amd:
Some of the build failures I don't understand - they might be an issue with this patch set or they might be an issue with the CI. There's a lot of 'Error: ipv6: address already assigned.', which could possibly be a result of the way that I'm allocating the plugin socket but seems unlikely. 
The linter errors I do understand - they're flagging real grammar and syntax errors in the amdgpu-drm.h and drm.h files - but the point of those files is that they're exact copies of the kernel header files of the same names, so I don't want to change them.

criu/include/criu-plugin.h
fdavid-amd: Fixed

fdavid-amd:
@avagin sorry for the repeat ping, but there is something of a rush on my part to get kernel patch https://lists.freedesktop.org/archives/dri-devel/2025-July/514922.html merged in the next week or so (it adds the CHANGE_HANDLE ioctl). As part of merging that patch, I need to demonstrate that it has written and reasonably stable userspace code that uses it (which would be these CRIU patches). It would be a big help to me if you could say if the general approach taken by these patches is acceptable or unacceptable. 
Thanks,
David

avagin:
@fdavid-amd The general approach looks good to me. We just need to adjust how we manage file descriptors, but it is just a criu internal thing.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1030,29 +1034,153 @@ int restore_hsakmt_shared_mem(const uint64_t shared_mem_size, const uint32_t sha
...
+int amdgpu_make_socket(int pid)
+{
+	int ret = 0;
+	struct amdgpu_process *p;
+	struct sockaddr_un saddr;
+	int sock, slen;
+
+	list_for_each_entry(p, &amdgpu_processes, l) {
+		if (p->pid == pid) {
+			dmabuf_socket_fd = find_unused_fd_pid(pid);
```
avagin: find_unused_fd_pid will not work without adding a few file entry (look at add_fake_fle). We can use fdstore, but I think it might be unneeded. Pls read other comment first.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1030,29 +1034,153 @@ int restore_hsakmt_shared_mem(const uint64_t shared_mem_size, const uint32_t sha
...
+	while (true) {
+		ret = __recv_fds(dmabuf_socket_fd, &fd, 1, (void *)&handle, sizeof(handle), MSG_DONTWAIT);
+		int id;
+
+		if (ret == -EAGAIN || ret == -EWOULDBLOCK)
+			return 0;
+		else if (ret)
+			return -1;
+
+		id = fdstore_add(fd);
```
avagin: Once you put a file descriptor in the fdstore, it becomes available to all other processes. This means you don't need to send it to each process separately; you can just put it in the fdstore once. You can use shared memory to synchronize/communicate between processes, so you probably don't need to create dmabuf_socket_fd sockets.

avagin:
> It would be a big help to me if you could say if the general approach taken by these patches is acceptable or unacceptable.

BTW we can merge the current version to the criu-dev branch and do all other adjustments on top of it.

plugins/amdgpu/amdgpu_plugin.c
fdavid-amd: Thanks for pointing this out. Is there an example of shared memory between CRIU processes without sockets that I could reference?

fdavid-amd:
This code is currently not functional because it relies on ioctls that are not merged into any kernel branch; is is still okay to merge?

rst0git:
>This code is currently not functional because it relies on ioctls that are not merged into any kernel branch; is is still okay to merge?

The way we usually handle this problem in CRIU is by checking if the required kernel functionality is available and disabling features that are not supported. However, it might be better to wait until the ioctls are merged.

fdavid-amd:
Some of these compilation tests seem to be failing at a install phase; is that something my code is doing or is it a problem with the test?

fdavid-amd:
@avagin There's now an additional patch in this patch set (Support for Open dmabuf handles) to handle the case where a dmabuf fd is open during checkpoint. It involves some changes to core CRIU, what do you think of the approach?

criu/files.c
```diff
@@ -329,7 +329,9 @@ uint32_t make_gen_id(uint32_t st_dev, uint32_t st_ino, uint64_t pos)
 	return st_dev ^ st_ino ^ pos_hi ^ pos_low;
 }
 
-int do_dump_gen_file(struct fd_parms *p, int lfd, const struct fdtype_ops *ops, FdinfoEntry *e)
+/* Use "force" to override cache */
+int do_dump_gen_file(struct fd_parms *p, int lfd,
+					const struct fdtype_ops *ops, FdinfoEntry *e, bool force)
```
rst0git: Would it be possible to provide more information about why do we need to "override cache"? For example, can you share with us an example where CRIU needs to handle open dmabuf file descriptors. We usually implement tests to very that such functionality works as expected.

fdavid-amd: CRIU has always dealt with dependencies between restoring files, but dmabuf fd adds dependencies between dumping files. A dmabuf fd can't be dumped until we've dumped the device file it was exported from. For this reason this patch set adds a retry system to file dumping. Files that are being retried should have the appropriate dump call made even if there is no new file descriptor for them.

I have implemented tests for internal testing, which can be found here. The DmabufDelayTest in this branch pauses while a dmabuf fd is active to trigger this code.
https://github.com/fdavid-amd/ROCT-Thunk-Interface-CRIU#

criu/files.c
```diff
@@ -654,9 +661,32 @@ int dump_task_files_seized(struct parasite_ctl *ctl, struct pstree_item *item, s
...
+				ret = 0; //* Reset ret to continue the loop */
+				continue;
+			} else if (ret)
+				break;
+
+			ret = pb_write_one(img, &e, PB_FDINFO);
 			if (ret)
 				break;
+		}
+		/* Dmabuf FDs cannot be dumped unless we have first dumped the device file that exported them.
```
avagin: Could the proper device file already be closed? Could it be opened in another process?

criu/files.c
```diff
@@ -654,9 +661,32 @@ int dump_task_files_seized(struct parasite_ctl *ctl, struct pstree_item *item, s
 		for (i = 0; i < nr_fds; i++) {
 			FdinfoEntry e = FDINFO_ENTRY__INIT;
 
-			ret = dump_one_file(item->pid, dfds->fds[i + off], lfds[i], opts + i, ctl, &e, dfds);
+			ret = dump_one_file(item->pid, dfds->fds[i + off], lfds[i], opts + i, ctl, &e, dfds, false);
+			if (ret == -EAGAIN) {
+				retry_indices[retry_count++] = i;
```
avagin: i+off?

criu/files.c
fdavid-amd: I was under the impression that CRIU didn't close the files it was dumping with until later in the process. 
The earlier patches in this pr allow the amdgpu plugin to find the dmabuf fd even if it was exported by another process.

fdavid-amd: I was under the impression that CRIU didn't close the files it was dumping with until later in the process. 
The earlier patches in this pr allow the amdgpu plugin to find the dmabuf fd even if it was exported by another process.

criu/files.c
fdavid-amd: I believe this addition is performed a few lines later

avagin:
> To handle this, the patch "Allow for retry of dumped files" racks files that failed to dump and tries to dump them again after all other dumps have succeeded. It uses a variable "is_retry" so that dump will be re-attempted even if it failed the first time.

Why do we need to do that in the criu generic code? Why can't we cache these file descriptors in the "plugin" and then dump them when their device file descriptor is dumped?

criu/files.c
avagin: I think something is wrong here. parasite_drain_fds_seized can grub up to PARASITE_MAX_FDS file descriptors from a dumped process, so parasite_drain_fds_seized can be called a few time (look at the top level loop). `i` is the index in the opts array, so it is reinitilized on each iteration of the top level loop. retry_indices isn't reinitialized on each iterration, so it can't store 'i' indexies. Have I missed something?

avagin:
LGTM

rst0git:
@fdavid-amd Would it be possible to provide more information on how we can test these patches?

criu/include/criu-plugin.h
```diff
@@ -62,6 +62,8 @@ enum {
 
 	CR_PLUGIN_HOOK__POST_FORKING = 12,
 
+	CR_PLUGIN_HOOK__RESTORE_INIT = 13,
```
rst0git, line 65: It would be good to update [plugins/amdgpu/README.md](https://github.com/checkpoint-restore/criu/blob/a000d89c70611e83dcc113ae59d0e4fdd244237d/plugins/amdgpu/README.md) with information about the new plugin callbacks.

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1403,7 +1407,17 @@ int amdgpu_plugin_dump_file(int fd, int id)
 		return -1;
 	}
 
-	/* Check whether this plugin was called for kfd or render nodes */
+	/* Check whether this plugin was called for kfd, dmabuf or render nodes */
+	ret = get_dmabuf_info(fd, &st);
+	if (ret < 0) {
+		pr_perror("Failed to get dmabuf info");
+		return -1;
+	} else if (ret == 0) {
```
rst0git: nit: since we use `return` here, we can remove `else`:
```
ret = get_dmabuf_info(fd, &st);
if (ret < 0) {
    pr_perror("Failed to get dmabuf info");
	return -1;
}

if (ret == 0) {
	pr_info("Dumping dmabuf fd = %d\n", fd);
	ret = amdgpu_plugin_dmabuf_dump(fd, id);
	return 0;
}
```

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1417,7 +1431,7 @@ int amdgpu_plugin_dump_file(int fd, int id)
 		ret = record_dumped_fd(fd, true);
 		if (ret)
 			return ret;
-
+		ret = try_dump_dmabuf_list();
 		/* Need to return success here so that criu can call plugins for renderD nodes */
 		return ret;
```
rst0git: Can we use `return try_dump_dmabuf_list();` here?

plugins/amdgpu/amdgpu_plugin_dmabuf.c
```diff
@@ -0,0 +1,201 @@
...
+        return 1;
+    }
+    int dmabuf_fd = fdstore_get(fd_id);
+    if (dmabuf_fd == -1) {
+        pr_err("Failed to find dmabuf_fd for GEM handle = %d\n",
+                                                            rd->gem_handle);
+        return 1;  /* Retry needed */
+    } else {
+        pr_info("Restored dmabuf_fd = %d for GEM handle = %d\n",
+                                                dmabuf_fd, rd->gem_handle);
```
rst0git: Could you please run `make indent` to fix the indentation?

plugins/amdgpu/amdgpu_plugin_dmabuf.c
```diff
@@ -0,0 +1,201 @@
...
+
+int try_dump_dmabuf_list()
+{
+   struct dmabuf *b, *t;
+   list_for_each_entry_safe(b, t, &dmabuf_list, node) {
+       int ret = __amdgpu_plugin_dmabuf_dump(b->dmabuf_fd, b->id);
+       if (ret == -EAGAIN)
+           continue;
+       else if (ret)
+           return ret;
```
rst0git:
```
if (ret == -EAGAIN)
    continue;

if (ret)
    return ret;
```

plugins/amdgpu/amdgpu_plugin_dmabuf.c
```diff
@@ -0,0 +1,201 @@
...
+   return 0;
+}
+
+int post_dump_dmabuf_check()
+{
+   if (!list_empty(&dmabuf_list)) {
+       pr_err("Not all dma buffers have been dumped\n");
+       return -1;
+   }
+   return 1;
```
rst0git: Why do we use `return 1;` here?

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1826,12 +1839,17 @@ int amdgpu_plugin_restore_file(int id, bool *retry_needed)
...
-		pr_info("Restoring RenderD %s\n", img_path);
 
 		img_fp = open_img_file(img_path, false, &img_size);
-		if (!img_fp)
-			return -EINVAL;
-
+		if (!img_fp) {
+			ret = amdgpu_plugin_dmabuf_restore(id);
+			if (ret == 1) {
+				*retry_needed = true;
```
rst0git, line 1849: Would it be possible to add a comment that explains why we use `retry_needed`?

fdavid-amd: Yes, done

criu/include/criu-plugin.h
fdavid-amd: That README doesn't seem to contain information about plugin callbacks; I think the appropriate place for such documentation might be the callback functions themselves or criu-plugin.h
In any case I hope it is clear what RESTORE_INIT does

plugins/amdgpu/amdgpu_plugin_dmabuf.c
fdavid-amd: Changed to tabs; hopefully now up to standard

criu/include/criu-plugin.h
rst0git: https://github.com/checkpoint-restore/criu/blob/a000d89c70611e83dcc113ae59d0e4fdd244237d/plugins/amdgpu/README.md?plain=1#L203

https://github.com/checkpoint-restore/criu/blob/a000d89c70611e83dcc113ae59d0e4fdd244237d/plugins/amdgpu/README.md?plain=1#L225

https://github.com/checkpoint-restore/criu/blob/a000d89c70611e83dcc113ae59d0e4fdd244237d/plugins/amdgpu/README.md?plain=1#L246

rst0git: The purpose of this documentation is to help new users to understand how CRIU and the AMD GPU Plugin work. Most users don't read the comments in header files like `criu-plugin.h`.

rst0git:
@fdavid-amd I've applied the patches here with @avagin's changes from the `wip/amdgpu` branch in the following pull request: https://github.com/checkpoint-restore/criu/pull/2810

PR #2620, opened 2025-03-13
Add a lock around cgroupd communication.

### LINKED ISSUE #2614: Restoring threads very slow.
I'm seeing slowness in restoring a Java process. In one instance, it took 49 seconds to restore a process with ~5G of heap.

Looking through the restore.log file, the majority of time seems to be taken up by actions related to rseq. From seconds 12 to 48.9, the restore.log looks like:

```
(30.874907) pie: 9380: 9380: Restored
(31.090798) pie: 10821: Restoring scheduler params 0.0.0
(31.090826) pie: 10821: rseq: rseq_abi_pointer = 0x7f0559898fe0 signature = 0x53053053
(31.090834) pie: 10821: 10821: Restored
(31.211756) pie: 10929: Restoring scheduler params 0.0.0
(31.211783) pie: 10929: rseq: rseq_abi_pointer = 0x7f0550bcbfe0 signature = 0x53053053
(31.211792) pie: 10929: 10929: Restored
(31.610783) pie: 628: Restoring scheduler params 0.0.0
(31.610816) pie: 628: rseq: rseq_abi_pointer = 0x7f05649cafe0 signature = 0x53053053
(31.610824) pie: 628: 628: Restored
(31.626826) pie: 9381: Restoring scheduler params 0.0.0
(31.626839) pie: 10822: Restoring scheduler params 0.0.0
(31.626849) pie: 9381: rseq: rseq_abi_pointer = 0x7f053dc9dfe0 signature = 0x53053053
(31.626854) pie: 10822: rseq: rseq_abi_pointer = 0x7f054ca8afe0 signature = 0x53053053
(31.626859) pie: 9381: 9381: Restored
(31.626860) pie: 10822: 10822: Restored
(31.626897) pie: 590: Restoring scheduler params 0.0.0
(31.626904) pie: 590: rseq: rseq_abi_pointer = 0x7f059dedefe0 signature = 0x53053053
(31.626910) pie: 590: 590: Restored
(31.710782) pie: 591: Restoring scheduler params 0.0.0
(31.710808) pie: 591: rseq: rseq_abi_pointer = 0x7f059dbdbfe0 signature = 0x53053053
(31.710816) pie: 591: 591: Restored
(32.062808) pie: 9382: Restoring scheduler params 0.0.0
```

I am puzzled why it is getting so few work items done. Is this doing serialized work, sleeps and/or is it possible that there is lock contention? 

This is with CRIU v4.0-75-g53d7921b7 running on an AWS c5d.2xlarge instance running Debian 12.

```
Linux ip-172-31-7-56 6.1.0-31-cloud-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.128-1 (2025-02-07) x86_64 GNU/Linux
```

CPU

```
processor	: 7
vendor_id	: GenuineIntel
cpu family	: 6
model		: 85
model name	: Intel(R) Xeon(R) Platinum 8275CL CPU @ 3.00GHz

```

hanwen-flow:
Threads are put into cgroups through the cgroupd thread, which communicates with other threads using a socketpair.  Each thread received a dup'd copy of the socket, and did the following

    sendmsg(socket_dup_fd, my_cgroup_set);

    // wait for ack.
    while (1) {
        recvmsg(socket_dup_fd, &h, MSG_PEEK);
        if (h.pid != my_pid) continue;
        recvmsg(socket_dup_fd, &h, 0);
    }
    close(socket_dup_fd);

When restoring many threads, most threads would be spinning in the above loop waiting for their PID to appear.

In my test-case, restoring a process with a 11.5G heap and 491 threads could take anywhere between 10 seconds and 60 seconds to complete.

To avoid the spinning, we drop the loop and MSG_PEEK, and add a lock around the above code. This does not decrease parallelism, as the cgroupd daemon uses a single thread anyway.

With the lock in place, the same restore takes a consistent 9.2 seconds on my machine (Thinkpad P14s, AMD Ryzen 8840HS).

There is a similar "daemon" thread for user namespaces. That already is protected with a similar userns_sync_lock in __userns_call().

Fixes #2614

### DIFF
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -704,9 +704,8 @@ static int send_cg_set(int sk, int cg_set)
 }
 
 /*
- * As this socket is shared among threads, recvmsg(MSG_PEEK)
- * from the socket until getting its own thread id as an
- * acknowledge of successful threaded cgroup fixup
+ * As the cgroupd socket is shared among threads and processes, this
+ * should be called with task_entries->cgroupd_sync_lock held.
  */
 static int recv_cg_set_restore_ack(int sk)
 {
@@ -719,33 +718,22 @@ static int recv_cg_set_restore_ack(int sk)
 	h.msg_control = cmsg;
 	h.msg_controllen = sizeof(cmsg);
 
-	while (1) {
-		ret = sys_recvmsg(sk, &h, MSG_PEEK);
-		if (ret < 0) {
-			pr_err("Unable to peek from cgroupd %d\n", ret);
-			return -1;
-		}
+	ret = sys_recvmsg(sk, &h, 0);
+	if (ret < 0) {
+		pr_err("Unable to receive from cgroupd %d\n", ret);
+		return -1;
+	}
 
-		if (h.msg_controllen != sizeof(cmsg)) {
-			pr_err("The message from cgroupd is truncated\n");
-			return -1;
-		}
+	if (h.msg_controllen != sizeof(cmsg)) {
+		pr_err("The message from cgroupd is truncated\n");
+		return -1;
+	}
 
-		ch = CMSG_FIRSTHDR(&h);
-		cred = (struct ucred *)CMSG_DATA(ch);
-		if (cred->pid != sys_gettid())
-			continue;
-
-		/*
-		 * Actual remove message from recv queue of socket
-		 */
-		ret = sys_recvmsg(sk, &h, 0);
-		if (ret < 0) {
-			pr_err("Unable to receive from cgroupd %d\n", ret);
-			return -1;
-		}
-
-		break;
+	ch = CMSG_FIRSTHDR(&h);
+	cred = (struct ucred *)CMSG_DATA(ch);
+	if (cred->pid != sys_gettid()) {
+		pr_err("cred pid %d != gettid\n", cred->pid);
+		return -1;
 	}
 	return 0;
 }
@@ -782,12 +770,21 @@ __visible long __export_restore_thread(struct thread_restore_args *args)
 	rt_sigframe = (void *)&args->mz->rt_sigframe;
 
 	if (args->cg_set != -1) {
+		int err = 0;
+
+		mutex_lock(&task_entries_local->cgroupd_sync_lock);
+
 		pr_info("Restore cg_set in thread cg_set: %d\n", args->cg_set);
-		if (send_cg_set(args->cgroupd_sk, args->cg_set))
-			goto core_restore_end;
-		if (recv_cg_set_restore_ack(args->cgroupd_sk))
-			goto core_restore_end;
+
+		err = send_cg_set(args->cgroupd_sk, args->cg_set);
+		if (!err)
+			err = recv_cg_set_restore_ack(args->cgroupd_sk);
+
+		mutex_unlock(&task_entries_local->cgroupd_sync_lock);
 		sys_close(args->cgroupd_sk);
+
+		if (err)
+			goto core_restore_end;
 	}
 
 	if (restore_thread_common(args))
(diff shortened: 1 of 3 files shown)


criu/include/restorer.h
```diff
@@ -159,7 +159,8 @@ struct task_restore_args {
 	struct task_entries *task_entries;
 	void *rst_mem;
 	unsigned long rst_mem_size;
-
+	mutex_t cgroupd_sync_lock;
```
minhbq-99, line 162: As the processes can be restored in parallel. I think this local mutex is not enough, it just guarantee mutual exclusion between threads in the same process. I think we should add a mutex to this struct: https://github.com/checkpoint-restore/criu/blob/5b4c819d54f953197e721b4c6af67af42860e5f3/criu/include/rst_info.h#L11-L18

Later, in `__export_restore_thread` , we can use it by `mutex_lock(&task_entries_local->cgroupd_sync_lock)`

hanwen-flow: ah, there is a cgroupd shared among all processes? That might explain why my previous approach didn't work :)

hanwen-flow: does that mean we have to keep the MSG_PEEK , though? Can two different processes use a mutex in task_entries? (that sounds wrong in a normal programming, but maybe they all mmap the same piece of memory?)  the userns lock doesn't use MSG_PEEK, so I guess it's OK ?

minhbq-99: Underlying, the mutex is implemented by futex. AFAIK, the `task_entries` is MAP_SHARED between the CRIU process and restored process. So in 2 processes, the address is mapped to the same physical page which turns into the same futex key.

criu/pie/restorer.c
```diff
@@ -782,12 +770,22 @@ __visible long __export_restore_thread(struct thread_restore_args *args)
 	rt_sigframe = (void *)&args->mz->rt_sigframe;
 
 	if (args->cg_set != -1) {
+		int err = 0;
```
minhbq-99, line 773: nit: I prefer a new empty line between the variable declaration and following statements.

criu/pie/restorer.c
```diff
@@ -782,12 +770,22 @@ __visible long __export_restore_thread(struct thread_restore_args *args)
...
+		mutex_lock(&task_entries_local->cgroupd_sync_lock);
+		
 		pr_info("Restore cg_set in thread cg_set: %d\n", args->cg_set);
-		if (send_cg_set(args->cgroupd_sk, args->cg_set))
-			goto core_restore_end;
-		if (recv_cg_set_restore_ack(args->cgroupd_sk))
-			goto core_restore_end;
+		
+		err = send_cg_set(args->cgroupd_sk, args->cg_set);
+		if (!err) {
```
minhbq-99, line 779: nit: we can remove the curly braces here

criu/pie/restorer.c
```diff
@@ -782,12 +770,22 @@ __visible long __export_restore_thread(struct thread_restore_args *args)
...
+			err = recv_cg_set_restore_ack(args->cgroupd_sk);
+		}
+
+		mutex_unlock(&task_entries_local->cgroupd_sync_lock);
 		sys_close(args->cgroupd_sk);
+
+		if (err) {
```
minhbq-99, line 786: nit: we can remove the curly braces here

minhbq-99 (commented):
There are 3 empty lines which have the tabs at the beginning so the linter is complaining.

I think the commit title should be "restorer: Add a lock around cgroupd communication" to follow the project's convention.

These are nits. Overall, the PR LGTM.

hanwen-flow:
thanks, PTAL. 

I tried running 'make lint', but it wants to run 'ruff' which isn't available on Ubuntu 24.

rst0git:
> it wants to run 'ruff' which isn't available on Ubuntu 24.

Have you tried `pip3 install ruff`?

avagin:
Merged. Thanks a lot.

Issue #2483, opened 2024-09-19
criu core dump when checkpointing a normal user vnc application.


coldbloodx:
**Description**

**Steps to reproduce the issue:**
1. follow https://criu.org/VNC create vncserver.sh like below:
```
[root@laworks criutool]# cat vncserver.sh
#!/bin/bash
set -m
Xvnc :25 -v -geometry 1440x900 -interface 0.0.0.0 -SecurityTypes none &
pid=$!
trap "kill $pid; wait" EXIT
sleep 3
DISPLAY=:25 $@
```

2. run an xclock like below:
```
[leo@laworks 4cpu]$ unshare -c -i ./vncserver.sh xclock
_XSERVTransmkdir: Owner of /tmp/.X11-unix should be set to root

Xvnc TigerVNC 1.13.1 - built Apr 22 2024 00:00:00
Copyright (C) 1999-2022 TigerVNC Team and many others (see README.rst)
See https://www.tigervnc.org for information on TigerVNC.
Underlying X server release 12011000


Thu Sep 19 14:25:44 2024
 vncext:      VNC extension running!
 vncext:      Listening for VNC connections on 0.0.0.0 interface(s), port 5925
 vncext:      created VNC server for screen 0

```

3. run criu dump `pgrep vncserver.sh`, then coredump happens
```
[root@laworks criutool]# criu dump -t 3781956 -D temp -j
Error (criu/namespaces.c:931): Unable to convert uid or gid
Error (criu/namespaces.c:997): One or more namespaces doesn't belong to the target user namespace
munmap_chunk(): invalid pointer
Aborted (core dumped)

```

4. check system log, found below core dump info:
```
Sep 19 14:29:00 laworks systemd[1]: Started Process Core Dump (PID 3782082/UID 0).
Sep 19 14:29:00 laworks systemd-coredump[3782083]: [🡕] Process 3782079 (criu) of user 0 dumped core.

                                                   Stack trace of thread 3782079:
                                                   #0  0x00007fb4ee08b94c __pthread_kill_implementation (libc.so.6 + 0x8b94c)
                                                   #1  0x00007fb4ee03e646 raise (libc.so.6 + 0x3e646)
                                                   #2  0x00007fb4ee0287f3 abort (libc.so.6 + 0x287f3)
                                                   #3  0x00007fb4ee029130 __libc_message.cold (libc.so.6 + 0x29130)
[...]

PR #2627, opened 2025-03-19
namespace: skip cleaning up the uid/gid map in error cases

minhbq-99:
free_userns_maps is called to clean up uid/gid map when the dump finishes. If we try to clean up these maps in error cases, it can lead to double free panic. So just skip cleaning up these maps and let free_userns_maps do its job.

Fixes: #2483


avagin:
Merged. Thanks!

Issue #358, opened 2017-07-05
Signals: non-safe usage of pr_err() in handlers


0x7f454c46:
We have several handlers for signals for different cases.
At least two of them (others I didn't check) use `pr_err()` function.
This is not safe not only in means of messages corruptions (which is not neat, but a minor thing), but also in means of locking.
`sigchld_handler()`does:
```
	if (exit)
		pr_err("%d exited, status=%d\n", pid, status);
	else
		pr_err("%d killed by signal %d: %s\n",
			pid, status, strsignal(status));
```

And `usernsd_handler()` does:
```
		if (!ns)
			pr_err("Spurious pid ns helper: pid=%d\n", pid);

		pr_err("%d finished unexpected: status=%d\n", pid, status);
```

But it looks like we have such a great feature as keeping first error message for RPC.
It's done with `log_note_err()`, which does this:
```
		mutex_lock(&first_err->l);
		if (first_err->s[0] == '\0')
			strlcpy(first_err->s, msg, sizeof(first_err->s));
		mutex_unlock(&first_err->l);
```

TLDR; We may lockup in case of delivered signal while we've taken mutex in `log_note_err()` as the function also is called from inside pr_err() in signal handlers.

Daz-3ux:
My understanding of the issue is that when we call `pr_err()`, we will enter the `log_note_err()` function, and if `log_note_err()` encounters a signal, it will try to call `pr_err()` again to acquire a lock that has already been acquired. However, the first call to `log_note_err()` has already been interrupted by the signal and cannot release the lock. My idea is that perhaps we can use a spin lock to modify `log_note_err`. Does my idea make sense?

Daz-3ux:
I want to try to solve it. However, I am concerned that this may not be an easy bug to test. What tools should I use to debug the deadlock issue in CRIU?

Daz-3ux:
> My understanding of the issue is that when we call `pr_err()`, we will enter the `log_note_err()` function, and if `log_note_err()` encounters a signal, it will try to call `pr_err()` again to acquire a lock that has already been acquired. However, the first call to `log_note_err()` has already been interrupted by the signal and cannot release the lock. My idea is that perhaps we can use a spin lock to modify `log_note_err`. Does my idea make sense?

TLDR; Is this a 'Self deadlock' problem? And can it be solved using spin locks + disabling interrupts?

mihalicyn:
[...]

PR #2629, opened 2025-03-22
criu: fix log_keep_err signal deadlock

ivpravdin:
When using pr_err in signal handler, locking is used in an unsafe manner. If another signal happens while holding the lock, deadlock can happen.

To fix this, we can introduce mutex_trylock similar to pthread_mutex_trylock that returns immediately. Due to the fact that lock is used only for writing first_err, this change garantees that deadlock cannot happen.

Fixes: #358

### DIFF
--- a/include/common/lock.h
+++ b/include/common/lock.h
@@ -2,6 +2,7 @@
 #define __CR_COMMON_LOCK_H__
 
 #include <stdint.h>
+#include <stdbool.h>
 #include <linux/futex.h>
 #include <sys/time.h>
 #include <limits.h>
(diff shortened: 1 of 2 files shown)


include/common/lock.h
```diff
@@ -162,6 +162,13 @@ static inline void mutex_lock(mutex_t *m)
 	}
 }
 
+static inline uint32_t mutex_trylock(mutex_t *m)
```
avagin, line 165: s/uint32_t/bool/

avagin:
LGTM

PR #2633, opened 2025-03-26
Simplify "cpuinfo" subcommand check

lianakoleva:
The command "cpuinfo" requires a subcommand "dump" or "check".
- `CR_CPUINFO` is removed; `CR_CPUINFO_DUMP` and `CR_CPUINFO_CHECK` are added in its place.
- All parsing is now done in `parse_criu_mode`, removing need for passing string subcommand to `image_dir_mode`.
- Check for valid subcommand is done once at the start with only the enum `CR_CPUINFO_DUMP` or `CR_CPUINFO_CHECK` propagated on success, rather than propagating both `CR_CPUINFO` and the subcommand string while doing `strcmp` on each call to `image_dir_mode`.


avagin:
> CR_CPUINFO is removed; CR_CPUINFO_DUMP and CR_CPUINFO_CHECK are added in its place.

How does it affect existing users who use CR_CPUINFO?

lianakoleva:
@avagin Could you share an example?

avagin:
@lianakoleva Sorry for the long delay. The main problem was that I hadn't seen a significant difference between the current and new version. Yesterday, I decided to look at this PR more closely, which led to even more cleanups in crtools.c: https://github.com/lianakoleva/criu/pull/1. Could you please review this change and merge it into your branch?

rst0git:
@lianakoleva Thank you for working on this!

@avagin I've updated Liana's branch with the cleanups in `crtools.c` and added one more commit for `cpuinfo check` to show an error message when the image file is missing.

lianakoleva:
@avagin @rst0git anything else this needs?

PR #2634, opened 2025-03-27
restore: use the new kernel interface to restore timers

avagin:
Thomas Gleixner introduced the new interface to create posix timers with specified timer IDs:
https://github.com/torvalds/linux/commit/ec2d0c04624b3c8a7eb1682e006717fa20cfbe24

Previously, CRIU recreated timers by repeatedly creating and deleting them until the desired ID was reached. This approach isn't fast, especially for timers with large IDs. For example, restoring two timers with IDs 1000000 and 2000000 took approximately 1.5 seconds.

The new `prctl()` based interface allows direct creation of timers with specified IDs, reducing the restoration time to around 3 microseconds for the same example.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -1720,6 +1720,22 @@ static int kerndat_has_close_range(void)
 	return 0;
 }
 
+static int kerndat_has_timer_cr_ids(void)
+{
+	if (prctl(PR_TIMER_CREATE_RESTORE_IDS,
+		  PR_TIMER_CREATE_RESTORE_IDS_GET, 0, 0, 0) == -1) {
+		if (errno == EINVAL) {
+			pr_debug("PR_TIMER_CREATE_RESTORE_IDS isn't supported\n");
+			return 0;
+		}
+		pr_perror("prctl returned unexpected error code");
+		return -1;
+	}
+
+	kdat.has_timer_cr_ids = true;
+	return 0;
+}
+
 /*
  * Some features depend on resource that can be dynamically changed
  * at the OS runtime. There are cases that we cannot determine the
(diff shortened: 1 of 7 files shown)


criu/kerndat.c
```diff
@@ -1720,6 +1720,21 @@ static int kerndat_has_close_range(void)
 	return 0;
 }
 
+static int kerndat_has_timer_cr_ids(void)
+{
+	if (prctl(PR_TIMER_CREATE_RESTORE_IDS, PR_TIMER_CREATE_RESTORE_IDS_OFF, 0, 0, 0) == -1) {
```
cyrillos, line 1725: Please use `PR_TIMER_CREATE_RESTORE_IDS_GET` instead, which is more safe than disabling.

criu/pie/restorer.c
```diff
@@ -1235,9 +1235,23 @@ static int timerfd_arm(struct task_restore_args *args)
...
+
+	/* prctl returns EINVAL if PR_TIMER_CREATE_RESTORE_IDS isn't supported. */
+	ret = sys_prctl(PR_TIMER_CREATE_RESTORE_IDS,
+			PR_TIMER_CREATE_RESTORE_IDS_ON, 0, 0, 0);
+	if (ret == 0) {
+		create_restore_ids = true
+	} else if (ret != -EINVAL) {
+		pr_err("Can't enabled PR_TIMER_CREATE_RESTORE_IDS: %d\n", ret);
+		return -1;
+	}
```
cyrillos, line 1254: I don't remember from scratch, but can't we use `kdat.has_timer_cr_ids` here to not test this feature twice?

cyrillos (approved):
Except few nits looks good for me, thanks! Feel free to leave code as is, since my nits are negligible

Issue #2632, opened 2025-03-25
Does criu not restore the read-only attribute of the container's rootfs?


yummypeng:
I use Docker to start a container, ensuring the container's rootfs is read-only by specifying the `--read-only` flag. Then, I save and resume the container using Docker's integrated checkpoint command. After restoring the container, when I exec into it, I notice that the rootfs has become writable. Here are the simple steps to reproduce the issue:

```
# docker run -td --read-only ubuntu:22.04 sleep inf
8cd6084fc2aa2584443aa5d2151ae7884b6c4b889a8bd869545255c1856574a6

# docker exec -ti 8cd6084fc2aa2584443aa5d2151ae7884b6c4b889a8bd869545255c1856574a6 bash
bash-4.4#
bash-4.4# findmnt /
TARGET SOURCE  FSTYPE  OPTIONS
/      overlay overlay ro,relatime,lowerdir=/var/lib/docker/overlay2/l/LEQGJFKK7AP52BWBHFYWUTSY7H:/var/lib/docker/....
bash-4.4#
```

The root directory of the container can be seen as read-only at this moment. Next, we will perform checkpoint/restore on this container:
```
# docker checkpoint create 8cd6084fc2aa2584443aa5d2151ae7884b6c4b889a8bd869545255c1856574a6 checkpoint1
checkpoint1

# docker start --checkpoint checkpoint1 8cd6084fc2aa2584443aa5d2151ae7884b6c4b889a8bd869545255c1856574a6
# docker exec -ti 8cd6084fc2aa2584443aa5d2151ae7884b6c4b889a8bd869545255c1856574a6 bash
bash-4.4#
bash-4.4# findmnt  /
TARGET SOURCE  FSTYPE  OPTIONS
/      overlay overlay rw,relatime,lowerdir=/var/lib/docker/overlay2/l/LEQGJFKK7AP52BWBHFYWUTSY7H:/var/lib/docker/...
```

At this point, the container's rootfs attribute has become rw.

adrianreber:
That sounds like a docker problem. The rootfs is pretty sure mounted by docker and then CRIU is called.

yummypeng:
@adrianreber So CRIU doesn't save and restore any mount attributes for the container's rootfs, right?

adrianreber:
No, only mounts on top of the rootfs

yummypeng:
Understood, thank you.

[...]

PR #2637, opened 2025-04-02
Preserve root mount flags

Snorch:
mount-v2: restore root mount flags

Mount flags belong to mount and mount namespace of the Container, so we
should preserve them, as Container user will not expect mounts switching
between ro and rw over c/r.

Fixes: https://github.com/checkpoint-restore/criu/issues/2632

### DIFF
--- a/test/zdtm/lib/ns.c
+++ b/test/zdtm/lib/ns.c
@@ -57,6 +57,9 @@ static int prepare_mntns(void)
 	if (zdtm_bind) {
 		/*
 		 * Bindmount the directory to itself.
+		 * e.g.: The mnt_ro_root test makes "/" mount readonly, but we
+		 * still want to write logs to /zdtm/static/ so let's make it
+		 * separate writable bind mount.
 		 */
 		snprintf(bind_path, sizeof(bind_path),  "%s/%s", root, zdtm_bind);
 		if (mount(bind_path, bind_path, NULL, MS_BIND, NULL)) {
--- a/test/zdtm/lib/ns.c
+++ b/test/zdtm/lib/ns.c
@@ -28,8 +28,9 @@ extern int pivot_root(const char *new_root, const char *put_old);
 static int prepare_mntns(void)
 {
 	int dfd, ret;
-	char *root, *criu_path, *dev_path;
+	char *root, *criu_path, *dev_path, *zdtm_bind;
 	char path[PATH_MAX];
+	char bind_path[PATH_MAX];
 
 	root = getenv("ZDTM_ROOT");
 	if (!root) {
@@ -50,6 +51,18 @@ static int prepare_mntns(void)
 	if (mount(root, root, NULL, MS_BIND | MS_REC, NULL)) {
 		fprintf(stderr, "Can't bind-mount root: %m\n");
 		return -1;
+	}
+
+	zdtm_bind = getenv("ZDTM_BIND");
+	if (zdtm_bind) {
+		/*
+		 * Bindmount the directory to itself.
+		 */
+		snprintf(bind_path, sizeof(bind_path),  "%s/%s", root, zdtm_bind);
+		if (mount(bind_path, bind_path, NULL, MS_BIND, NULL)) {
+			fprintf(stderr, "Can't bind-mount ZDTM_BIND: %m\n");
+			return -1;
+		}
 	}
 
 	dev_path = getenv("ZDTM_DEV");
--- a/test/zdtm/static/mnt_ro_root.c
+++ b/test/zdtm/static/mnt_ro_root.c
@@ -1,0 +1,32 @@
+#include <sys/mount.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check if root mount remains read-only after c/r";
+const char *test_author = "Pavel Tikhomirov <ptikhomirov@virtuozzo.com>";
+
+char *dirname;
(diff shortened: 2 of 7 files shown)


Snorch:
I originally did statmount version https://github.com/Snorch/criu/tree/preserve-root-mount-flags-statmount but as long as flag inconsistency is detected by zdtm already like:
```
1: Old mounts lost: [('/home/turmoil/devel/ms/criu/test', '/', 'ro,relatime', None, None)]
1: New mounts appeared: [('/home/turmoil/devel/ms/criu/test', '/', 'rw,relatime', None, None)]
############# Test zdtm/static/mnt_ro_root FAIL at mounts compare ##############
```
So I removed statmount part for easier compatibility.

test/zdtm/static/mnt_ro_root.c
```diff
@@ -0,0 +1,25 @@
...
+	test_init(argc, argv);
+
+	if (mount(NULL, "/", NULL, MS_REMOUNT | MS_RDONLY | MS_BIND, NULL)) {
+		pr_perror("mount");
+		return 1;
+	}
+
+	test_daemon();
+	test_waitsig();
+
```
avagin, line 22: should we check that the mount is still read-only?

Snorch: zdtm already checks it for all the tests with mntns, please see https://github.com/checkpoint-restore/criu/pull/2637#issuecomment-2771652435

avagin: I had to read comments:). Let's add a comment in the test to explain why it doesn't check anything.

Snorch: done

rst0git (approved):
LGTM

test/zdtm/lib/ns.c
```diff
@@ -52,6 +53,17 @@ static int prepare_mntns(void)
 		return -1;
 	}
 
+	/*
+	 * The mnt_ro_root test makes "/" mount readonly, but we still
+	 * want to write logs to /zdtm/static/ so let's make it separate
+	 * writable mount.
+	 */
+	snprintf(testdir, sizeof(testdir),  "%s/zdtm/static/", root);
+	if (mount(testdir, testdir, NULL, MS_BIND, NULL)) {
```
rst0git, line 62:
```
sudo ./zdtm.py run -t zdtm/static/mnt_ext_auto -f uns
userns is supported
Checking feature mnt_id
mnt_id is supported
=== Run 1/1 ================ zdtm/static/mnt_ext_auto
===================== Run zdtm/static/mnt_ext_auto in uns ======================
Start test
./mnt_ext_auto --pidfile=mnt_ext_auto.pid --outfile=mnt_ext_auto.out --dirname=mnt_ext_auto.test
Can't bind-mount testdir: Invalid argument
The test failed (0, 1)
make: *** [Makefile:519: mnt_ext_auto.pid] Error 1
 Test zdtm/static/mnt_ext_auto FAIL at ['make', '--no-print-directory', '-C', 'zdtm/static', 'mnt_ext_auto.pid'] 
```

avagin, line 62: A few tests has to be adjusted. Here is the patch for mntns_shared_bind:
```
diff --git a/test/zdtm/static/mntns_shared_bind.c b/test/zdtm/static/mntns_shared_bind.c
index 7c745606e..6f300195f 100644
--- a/test/zdtm/static/mntns_shared_bind.c
+++ b/test/zdtm/static/mntns_shared_bind.c
@@ -42,7 +42,7 @@ int main(int argc, char **argv)
                return 1;
        }
 
-       if (mount(NULL, "/", NULL, MS_SHARED, NULL)) {
+       if (mount(NULL, "/zdtm/static", NULL, MS_SHARED, NULL)) {
                pr_perror("mount");
                return 1;
        }
```

Snorch: Thanks for pointing this out! I think adding this bindmount for all tests was a mistake. Let's add it to `mnt_ro_root` test only, thus we will know for sure that we didn't break other tests.

avagin:
```
2025-04-07T04:32:28.8557433Z === Run 152/474 =====----------- zdtm/static/mnt_root_ext
2025-04-07T04:32:28.8558304Z ====================== Run zdtm/static/mnt_ro_root in uns ======================
2025-04-07T04:32:28.8559088Z Start test
2025-04-07T04:32:28.8560246Z Test is SUID
2025-04-07T04:32:28.8562798Z ./mnt_ro_root --pidfile=mnt_ro_root.pid --outfile=mnt_ro_root.out --dirname=mnt_ro_root.test
2025-04-07T04:32:28.8563864Z Run criu dump
2025-04-07T04:32:28.8564362Z Run criu restore
2025-04-07T04:32:28.8565456Z 1: Old mounts lost: [('/home/runner/actions-runner/_work/criu/criu/test', '/', 'ro,relatime', None, None)]
2025-04-07T04:32:28.8567793Z 1: New mounts appeared: [('/home/runner/actions-runner/_work/criu/criu/test', '/', 'rw,relatime', None, None)]
2025-04-07T04:32:28.8569902Z ############# Test zdtm/static/mnt_ro_root FAIL at mounts compare ##############
```

Snorch:
The main patch is to mount-v2, so test should not be run in `--mntns-compat-mode`, adding `--no-mntns-compat-mode` to the test to overcome this.

PR #2646, opened 2025-04-14
feat: add Nix flake

### LINKED ISSUE #2332: Introduce a Nix flake for CRIU development
CRIU currently requires a decent number of dependencies in order to build from source. I also suspect that we are dependent on specific versions of certain dependencies in order to build correctly or support some features. Introducing a Nix flake will allow developers to spin up a dev env using `nix develop`, saving them the trouble of manually dealing with dependency installation across distros. It'll also prevent us from polluting the global package set on the machine in case we need to jump back and fix a regression that needs an older version of a dependency.

This would simply require adding a `flake.nix` file with the appropriate configuration to successfully build CRIU.

snprajwal:
Closes #2332

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,7 +34,7 @@ To clone CRIU repo and switch to the proper branch, run:
 
 ### Compile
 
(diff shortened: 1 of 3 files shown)


snprajwal:
cc @avagin

PR #2656, opened 2025-04-30
zdtm.py: add an option to change pycriu import path

Snorch:
By default zdtm expects that criu is built from source first and only then you can run zdtm tests against it. But what if you really want to run tests against a criu version installed on the system? Yes there is already a nice option for zdtm to change the criu binary it uses "--criu-bin", but it would still end up using the pycriu module from source and you would still have to build everything beforehand.

Let's add an option to change the path where zdtm searches for pycriu module "--pycriu-search-path". This way we can run zdtm tests on the criu installed on the system directly without building criu from source, e.g. on Fedora it works like:
```
test/zdtm.py run --criu-bin /usr/sbin/criu \
  --pycriu-search-path /usr/lib/python3.13/site-packages \
  -t zdtm/static/env00
```

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -22,11 +22,11 @@ import sys
 import tempfile
 import time
 import uuid
+import site
 from builtins import input, int, open, range, str, zip
 
 import yaml
 
-import pycriu as crpc
 from zdtm.criu_config import criu_config
 
 # File to store content of streamed images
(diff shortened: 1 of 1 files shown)


avagin:
LGTM

test/zdtm.py
```diff
@@ -2860,6 +2868,9 @@ def get_cli_args():
     rp.add_argument("--criu-bin",
                     help="Path to criu binary",
                     default='../criu/criu')
+    rp.add_argument("--pycriu-search-path",
+                    help="Path to where to search pycriu module first",
```
rst0git, line 2872:
```Python

import site
...

help=f"Path to search for pycriu module first (e.g., {site.getsitepackages()[0]})",
```

test/zdtm.py
```diff
@@ -1142,6 +1141,14 @@ def __init__(self, opts):
         self.__img_streamer_process = None
         self.__tls = self.__tls_options() if opts['tls'] else []
         self.__criu_bin = opts['criu_bin']
+
+        global crpc
+        if opts['pycriu_search_path']:
+            sys.path = [opts['pycriu_search_path']] + sys.path
```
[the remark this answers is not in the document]
Snorch: Thanks, yes this way it's much easier for the user. Also I've mentioned that criu binary availability check was broken, adding a separate patch to fix that and add similar comment there.

test/zdtm.py
Snorch: Very useful, added. Thanks!

rst0git (approved):

PR #2652, opened 2025-04-20
Support MADV_GUARD_INSTALL

mihalicyn:
Ready for review / merge.

Linux kernel 6.16+ is required to support dump for `MADV_GUARD_INSTALL`-protected pages.

Change-log:
05.08.2025: completely reworked to support file mappings and shared mappings

ToDo:
- [x] support file-backed and read-only mappings https://github.com/torvalds/linux/commit/f807123d578df4218e2580a1e1bb3436f4567c4a
- [x] handle VM_WIPEONFORK in CRIU & test
- [x] some more code cleanup needed
- [x] ensure that `./test/zdtm.py run -t zdtm/static/maps12 -f h --page-server` works

Closes #2626

PR #2659, opened 2025-05-04
mem: support VM_WIPEONFORK/MAP_DROPPABLE

mihalicyn:
It would be good to have this before https://github.com/checkpoint-restore/criu/pull/2652 as I would like to test how these two can co-exist.

ToDo:
- [x] add test with `fork()` to ensure that premap/cow detection logic is fine
- [x] decide if we have to forbid premapping for these mappings? (Nope, we don't.)
- [x] address code review comments from Bui Quang Minh and Andrei

### DIFF
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -160,6 +160,8 @@ static void __parse_vmflags(char *buf, u32 *flags, u64 *madv, int *io_pf,
 			*madv |= (1ul << MADV_HUGEPAGE);
 		else if (_vmflag_match(tok, "nh"))
 			*madv |= (1ul << MADV_NOHUGEPAGE);
+		else if (_vmflag_match(tok, "wf"))
+			*madv |= (1ul << MADV_WIPEONFORK);
 
 		/* vmsplice doesn't work for VM_IO and VM_PFNMAP mappings. */
 		if (_vmflag_match(tok, "io") || _vmflag_match(tok, "pf"))
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -144,6 +144,8 @@ static void __parse_vmflags(char *buf, u32 *flags, u64 *madv, int *io_pf,
 			*flags |= MAP_NORESERVE;
 		else if (_vmflag_match(tok, "ht"))
 			*flags |= MAP_HUGETLB;
+		else if (_vmflag_match(tok, "dp"))
+			*flags |= MAP_DROPPABLE;
 
 		/* madvise() block */
 		if (_vmflag_match(tok, "sr"))
@@ -206,6 +208,20 @@ static void parse_vma_vmflags(char *buf, struct vma_area *vma_area)
 
 	if (vma_area->e->madv)
 		vma_area->e->has_madv = true;
+
+	/*
+	 * We set MAP_PRIVATE flag on vma_area->e->flags right after parsing
+	 * a first line of VMA entry in /proc/<pid>/smaps file:
+	 * 7fa84fa70000-7fa84fa95000 rw-p 00000000 00:00 0
+	 * but it's too early and we can't distinguish between MAP_DROPPABLE
+	 * and MAP_PRIVATE mappings yet, as they both private mappings in nature
+	 * and at this point we haven't yet read "VmFlags:" line in smaps.
+	 *
+	 * Let's detect this situation and drop MAP_PRIVATE flag while keep
+	 * MAP_DROPPABLE, otherwise restorer's restore_mapping() helper will fail.
+	 */
+	if ((vma_area->e->flags & MAP_PRIVATE) && (vma_area->e->flags & MAP_DROPPABLE))
+		vma_area->e->flags &= ~MAP_PRIVATE;
 }
 
 static inline int is_anon_shmem_map(dev_t dev)
--- a/test/zdtm/static/maps11.c
+++ b/test/zdtm/static/maps11.c
@@ -1,0 +1,205 @@
+#include <stdint.h>
+#include <signal.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/wait.h>
+#include "zdtmtst.h"
+
+#ifndef MAP_DROPPABLE
+#define MAP_DROPPABLE 0x08
+#endif
+
+#ifndef MADV_WIPEONFORK
+#define MADV_WIPEONFORK 18
+#endif
+
+const char *test_doc = "Test MAP_DROPPABLE/MADV_WIPEONFORK mappings with 2 processes";
+const char *test_author = "Alexander Mikhalitsyn <aleksandr.mikhalitsyn@canonical.com>";
+
+bool mem_is_zero(const uint8_t *buffer, size_t length)
+{
+	size_t i;
+
+	for (i = 0; i < length; i++)
+		if (buffer[i] != 0)
+			return false;
+
+	return true;
+}
+
+int main(int argc, char **argv)
+{
+	uint8_t *p1, *p2;
+	pid_t pid;
+	int status;
+	const char data[] = "MADV_WIPEONFORK vma data";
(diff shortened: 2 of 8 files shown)


test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,112 @@
...
+
+	if (madvise(p2, sizeof(data), MADV_WIPEONFORK)) {
+		pr_perror("madvise failed");
+		return -1;
+	}
+
+	memcpy(p1, data, sizeof(data));
+	memcpy(p2, data, sizeof(data));
+
+	pid = test_fork();
```
avagin, line 55: why do we do fork before C/R (test_waitsig)?

test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,112 @@
...
+	if (pid < 0) {
+		pr_perror("fork failed");
+		return 1;
+	}
+
+	if (pid == 0) {
+		task_waiter_complete(&t, 1);
```
avagin, line 62: why do we need this waiter?

criu/proc_parse.c
```diff
@@ -206,6 +208,20 @@ static void parse_vma_vmflags(char *buf, struct vma_area *vma_area)
...
+	 * a first line of VMA entry in /proc/<pid>/smaps file:
+	 * 7fa84fa70000-7fa84fa95000 rw-p 00000000 00:00 0
+	 * but it's too early and we can't distinguish between MAP_DROPPABLE
+	 * and MAP_PRIVATE mappings yet, as they both private mappings in nature
+	 * and at this point we haven't yet read "VmFlags:" line in smaps.
+	 *
+	 * Let's detect this situation and drop MAP_PRIVATE flag while keep
+	 * MAP_DROPPABLE, otherwise restorer's restore_mapping() helper will fail.
+	 */
+	if ((vma_area->e->flags & MAP_PRIVATE) && (vma_area->e->flags & MAP_DROPPABLE))
```
avagin, line 223: do we need to dump content of MAP_DROPPABLE areas? The kernel commit message says:
```
b) Under memory pressure, mm can just drop the pages (so that they're
   zero when read back again).
```

mihalicyn: yes I think we can drop the entire contents of them, but what is more important is that right now (without these fixes) we restore `MAP_DROPPABLE` mappings just as `MAP_PRIVATE` ones and lose `VM_WIPEONFORK`. This can be fatal if application relies on the fact that after `fork()` mapping should be null-ified, because it won't be. I would classify this as a memory corruption-like issue.

criu/proc_parse.c
```diff
@@ -206,6 +208,20 @@ static void parse_vma_vmflags(char *buf, struct vma_area *vma_area)
...
+		vma_area->e->flags &= ~MAP_PRIVATE;
```
minhbq-99, line 224: At first, I'm quite concerned that this makes MAP_PRIVATE not set in `vma_area->e->flags` but `vma_area->e->status` might have VMA_ANON_PRIVATE as this is set in `handle_vma` already. Looking for the check of MAP_PRIVATE in `vma->e->flags` in the code, I only see 2 places `can_dump_with_memfd_hugetlb` and `vma_entry_can_be_lazy`
- can_dump_with_memfd_hugetlb: MAP_DROPPABLE cannot reach this path as MAP_DROPPABLE cannot be hugetlb backed
- vma_entry_can_be_lazy: must return false on MAP_DROPPABLE as we cannot use userfaultfd on MAP_DROPPABLE,. As we drop MAP_PRIVATE here, this will correctly return false

So this change looks good to me.

test/zdtm/static/maps11.c
mihalicyn: I have added an explanatory comment in there and also second `fork()` after C/R to make test even more strict.

Thanks!

test/zdtm/static/maps11.c
mihalicyn: Nope, my bad. Removed.

criu/proc_parse.c
mihalicyn: Have fixed this. Now we drop `MAP_DROPPABLE` mappings content after C/R.

Thanks!

test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,112 @@
...
+	test_waitsig();
+
+	kill(pid, SIGTERM);
+	wait(&status);
+	if (WIFEXITED(status)) {
+		if (WEXITSTATUS(status))
+			goto err;
+	} else {
+		goto err;
+	}
```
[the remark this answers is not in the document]
mihalicyn: Hmm, no, I would say that this logic is correct. Basically, we want to check if task is exited with zero exit code. All other situations we treat as error. I've done some small refactoring of this and collapsed it into one `if` condition, also replaced `wait()` with `waitpid()` to make it a bit more robust. What do you think?

test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,112 @@
...
+		wait(NULL);
+	}
+	return 1;
+
+skip:
+	test_daemon();
+	test_waitsig();
+	pass();
+	return 0;
+}
```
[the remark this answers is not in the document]
mihalicyn: Have fixed. Thanks! ;-)

criu/proc_parse.c
mihalicyn: yeah, I was thinking about this too. It is kinda tricky. Thanks for double checking and ensuring that everything is correct! ;-)

test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	if (pid == 0) {
+		if (!mem_is_zero(p1, sizeof(data)) || !mem_is_zero(p2, sizeof(data)))
+			return 1;
+
+		return 0;
+	}
+
+	waitpid(pid, &status, 0);
```
avagin, line 159: nit: pls check errors codes of system calls. It can save hours in the future...
if (waitpid(pid...) != pid) {
  pr_perror("waitpid");
  goto hell;
}

test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,178 @@
...
+			return 1;
+
+		return 0;
+	}
+
+	/*
+	 * A simple way to detect if C/R happened is to compare st_ino
+	 * fields of stat() on the procfs files of the current task.
+	 *
+	 * Hopefully, this terrible hack is never used in real-world applications ;-)
```
avagin, line 102: nit: 80 symbols warp is still preferred...

avagin:
LGTM. Thanks a lot!

test/zdtm/static/maps11.c
mihalicyn: You are right. My bad. Have fixed.

test/zdtm/static/maps11.c
```diff
@@ -0,0 +1,178 @@
...
+
+	if (pid == 0) {
+		if (!mem_is_zero(p1, sizeof(data)) || !mem_is_zero(p2, sizeof(data)))
+			return 1;
+
+		return 0;
+	}
+
+	waitpid(pid, &status, 0);
+	if (!WIFEXITED(status) || WEXITSTATUS(status) != 0)
```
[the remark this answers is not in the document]
mihalicyn: > * Always print an error if something goes wrong...

Have fixed. Thanks!



> * I prefer to check if status is non-zero. If is simple and works all the time. All these WABARAKADABRA looks the same.

you mean not to use `WIFEXITED`/`WEXITSTATUS` macros?

test/zdtm/static/maps11.c
mihalicyn: Have fixed!

test/zdtm/static/maps11.c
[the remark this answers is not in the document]
avagin: Yes, I mean that: `if (!status) {`

Issue #338, opened 2017-06-29
Report first error from RPC always!


xemul:
In RPC we report back the first error message met. But this is not always the case.

See commit f1a9cd4d. The problem is that pre_dump_using_req() calls fork() to do the pre-dump itself. The called function cr_pre_dump_tasks() calls log_keep_err() to keep the message, then the task exits. Its parent, which sits in the pre_dump_using_req() then tries to get the first err by calling log_first_err(), but there's no error there, it's been set in the child task.

Need some way to get the error back. E.g. by setting up the error message buffer in parent, or by sending the RPC response from child.

Daz-3ux:
Hi there, I have some questions regarding the description of this specific issue. 
The statement reads, "The called function cr_pre_dump_tasks() calls log_keep_err() to keep the message, then the task exits." This implies that the call chain is pre_dump_using_req() -> fork -> cr_pre_dump_tasks() -> log_keep_err(), which means that the child process initializes the error message, causing the parent process to only be able to return NULL when reporting an error(in log_first_err() ).

However, I found that log_keep_err() is only called by setup_opts_from_req(), with the call chain being pre_dump_using_req() -> fork -> setup_opts_from_req -> log_keep_err(). Therefore, the child process is the one setting the error message.

I would like to know 
if I am misunderstanding the problem \
if there is an error in the issue description \
if the issue has already been resolved? 

Thank you for all your help. :)

PR #2661, opened 2025-05-07
rpc/log: return first error always

ivpravdin:
Introduce interface to criu-log component to allocate 
first error buffer shared between a parent and its children.

Use shared first error buffer to return correct
first error in rpc.

Fixes: #338

### DIFF
--- a/criu/log.c
+++ b/criu/log.c
@@ -10,6 +10,7 @@
 #include <sys/time.h>
 #include <sys/resource.h>
 #include <sys/utsname.h>
+#include <sys/mman.h>
 
 #include <fcntl.h>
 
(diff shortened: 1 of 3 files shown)


criu/log.c
```diff
@@ -123,6 +127,23 @@ int log_keep_err(void)
 	return 0;
 }
 
+int log_keep_shr_err(void)
```
avagin, line 130: log_keep_err does the same thing, doesn't it?

ivpravdin: The new function creates new memory mapping with MAP_SHARED, this memory region is shared between a parent and its children. It's useful when a rpc server spawns a child but does not have access to the child's error buffer when something goes wrong

avagin: log_keep_err uses shmalloc, it creates shared regions too.

ivpravdin: Uh, I was looking at the wrong shmalloc definition. I've updated it to just use `log_keep_err`.

PR #2681, opened 2025-06-12
fix: link_remap files get left over on failed dumps, unique file names to avoid collisions

bsmithai:
## Problem:
`create_link_remap` logic specifies before creating the hard link that these file names can be truly unique. Currently these files are generated w/ ids matching sequentially with file descriptor ids. This change introduces generating a random suffix to prevent link_remap collisions when a dump fails and another dump is attempted on the same process that has open fds with links that were remapped.

## Tested:
Tested w/ a process that opens and uses semaphores, the test creates remap files in `/dev/shm` that would have been created by criu on dump. A dump is done on this process and collisions are avoided. A restore is then done to verify nothing was damaged in C/R. 

Very simple solution but I wanted to verify e2e.

## Results:
Original bug:
```
(00.022698) Strip ' (deleted)' tag from './dev/shm/sem.8NKp2J (deleted)'
(00.022703) Warn  (criu/files-reg.c:1881): Can't link  -> ./dev/shm/link_remap.5
(00.022707) Error (criu/files-reg.c:1159): Can't link remap to /dev/shm/sem.8NKp2J: File exists
(00.022711) Error (criu/cr-dump.c:1570): Collect mappings (pid: 1401204) failed with -1
(00.022753) net: Unlock network
```
This branch:
```
(00.043028) Found regular file mapping, OK
(00.043041) Dumping path for -3 fd via self 12 [/dev/shm/sem.8NKp2J (deleted)]
(00.043046) Strip ' (deleted)' tag from './dev/shm/sem.8NKp2J (deleted)'
(00.043081) Only file size could be stored for validation for file /dev/shm/sem.8NKp2J
(00.043087) Handling VMA with the following smaps entry: 745051de5000-745051de7000 r--p 00037000 fc:01 17565282                   /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
(00.043097) Found regular file mapping, OK
(00.043120) Handling VMA with the following smaps entry: 745051de7000-745051de9000 rw-p 00039000 fc:01 17565282                   /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
(00.043130) vma 745051de7000 borrows vfi from previous 745051de5000
(00.043134) Handling VMA with the following smaps entry: 7ffcde225000-7ffcde247000 rw-p 00000000 00:00 0                          [stack]
(00.043143) Handling VMA with the following smaps entry: ffffffffff600000-ffffffffff601000 --xp 00000000 00:00 0                  [vsyscall]
(00.043150) Collected, longest area occupies 405 pages
(00.043153) 0x64e115337000-0x64e115338000 (4K) prot 0x1 flags 0x2 fdflags 0 st 0x41 off 0 reg fp  shmid: 0x1
(00.043156) 0x64e115338000-0x64e115339000 (4K) prot 0x5 flags 0x2 fdflags 0 st 0x41 off 0x1000 reg fp  shmid: 0x1
(00.043158) 0x64e115339000-0x64e11533a000 (4K) prot 0x1 flags 0x2 fdflags 0 st 0x41 off 0x2000 reg fp  shmid: 0x1
```

```
Original CRIU (/usr/local/bin/criu): Exit code 1
Modified CRIU (./criu/criu):         Exit code 0
Modified CRIU Restore:               Exit code 0
```


criu/files-reg.c
```diff
@@ -1149,18 +1152,39 @@ static int create_link_remap(char *path, int len, int lfd, u32 *idp, struct ns_i
 	rfe.fown = &fwn;
 	rfe.name = link_name + 1;
 
-	/* Any 'unique' name works here actually. Remap works by reg-file ids. */
-	snprintf(tmp + 1, sizeof(link_name) - (size_t)(tmp - link_name) - 1, "link_remap.%d", rfe.id);
-
 	mntns_root = mntns_get_root_fd(nsid);
 
+	fd = open("/dev/urandom", O_RDONLY);
```
adrianreber, line 1157: Curious question. Why /dev/random and not a syscall like getrandom()?

What is with the branch when open() fails? Can you add a comment why that branch exists and why /dev/random?

adrianreber:
Do you think you could also add a zdtm test? You mentioned that you have a test that is able to trigger the problem you described.

criu/files-reg.c
bsmithai: I honestly didn't know about getrandom but have used /dev/random before, this can be changed.

And the check on open is just if there's security or containerized environment that is missing features. I can get rid of the fallback though if criu typically wouldn't run in a containerized environment or w/ security policies that would not allow use of these special files.

adrianreber: Right, containerized not having all devices is a good reason. I don't really know if the syscall is better but it seems definitely simpler.

You are also doing the time fallback twice. Not sure if there is a simple way to avoid the code duplication. Although it is just a couple of lines.

If using the syscall maybe the time fallback isn't necessary. Never used it myself.

bsmithai: CRIU docs says it supports >=3.11 but `getrandom` was released in 3.17, I can add `getrandom` but it would bump the support up for link remap to 3.17, unless we just have `/dev/urandom` be the fallback after `getrandom`.

I have another branch directly related to this PR and fix, it's a `--content-remap` where we serialize a file that would normally be link remapped. The purpose is to have support of migration of linked files whose original path had been moved. content remap would drop the file dependency in `/dev` for tmpfs files like semaphores. It's specific to files in memory and they the ceiling of file size is set to `memfd` threshold. Curious of what your thoughts are @adrianreber

bsmithai: nvm I think support for posix semaphores is needed, the content remap idea won't really work bc it's treating it as a regular file. I am thinking in terms of migration support for these files

avagin, line 1157: should we generate this uid once for the entire dump?

criu/files-reg.c
```diff
@@ -1149,18 +1152,39 @@ static int create_link_remap(char *path, int len, int lfd, u32 *idp, struct ns_i
...
+		close(fd);
+	} else {
+		struct timespec ts;
+		clock_gettime(CLOCK_MONOTONIC, &ts);
+		snprintf(random_suffix, sizeof(random_suffix), "%lx", 
+			(unsigned long)(ts.tv_sec ^ ts.tv_nsec));
+	}
+
+    // create file link_remap.fd_id.random_suffix id
+	snprintf(tmp + 1, sizeof(link_name) - (size_t)(tmp - link_name) - 1,
```
avagin, line 1178: can we customize a link name if linkat_hard returns EEXIST?

criu/files-reg.c
```diff
@@ -1149,18 +1152,39 @@ static int create_link_remap(char *path, int len, int lfd, u32 *idp, struct ns_i
...
-	snprintf(tmp + 1, sizeof(link_name) - (size_t)(tmp - link_name) - 1, "link_remap.%d", rfe.id);
-
 	mntns_root = mntns_get_root_fd(nsid);
 
+	fd = open("/dev/urandom", O_RDONLY);
+	if (fd >= 0) {
+		unsigned char rand_bytes[8];
+		if (read(fd, rand_bytes, sizeof(rand_bytes)) == sizeof(rand_bytes)) {
+			snprintf(random_suffix, sizeof(random_suffix), "%02x%02x%02x%02x",
+				rand_bytes[0], rand_bytes[1], rand_bytes[2], rand_bytes[3]);
```
avagin, line 1162: why do you read 8 bytes, if you use just four of them?

criu/files-reg.c
```diff
@@ -1149,18 +1152,39 @@ static int create_link_remap(char *path, int len, int lfd, u32 *idp, struct ns_i
...
+		}
+		close(fd);
+	} else {
+		struct timespec ts;
+		clock_gettime(CLOCK_MONOTONIC, &ts);
+		snprintf(random_suffix, sizeof(random_suffix), "%lx", 
+			(unsigned long)(ts.tv_sec ^ ts.tv_nsec));
+	}
+
+    // create file link_remap.fd_id.random_suffix id
```
avagin, line 1177: please use tabs for indentation

criu/files-reg.c
```diff
@@ -1149,18 +1152,39 @@ static int create_link_remap(char *path, int len, int lfd, u32 *idp, struct ns_i
...
+		"link_remap.%d.%s", rfe.id, random_suffix);
+
 	while (linkat_hard(lfd, "", mntns_root, link_name, ost->st_uid, ost->st_gid, AT_EMPTY_PATH) < 0) {
 		if (errno != ENOENT) {
 			pr_perror("Can't link remap to %s", path);
 			return -1;
 		}
-
-		/* Use grand parent, if parent directory does not exist. */
+
```
avagin, line 1186: white spaces...

rst0git:
>Tested w/ a process that opens and uses semaphores, the test creates remap files in /dev/shm that would have been created by criu on dump. A dump is done on this process and collisions are avoided. A restore is then done to verify nothing was damaged in C/R.

Would you be able to include this test, for example, as part of ZDTM or [`test/others`](https://github.com/checkpoint-restore/criu/tree/criu-dev/test/others)?

criu/files-reg.c
rst0git: >This change introduces generating a random suffix to prevent link_remap collisions when a dump fails and another dump is attempted on the same process

> should we generate this uid once for the entire dump?

We already do this for [`dump_criu_run_id`](https://github.com/checkpoint-restore/criu/blob/b24f6e25bb99211ddfe971bb7f47778928a2f737/criu/util.c#L2036). Can we use this value instead of generating a new one?

rst0git: > I have another branch directly related to this PR and fix ...

@bsmithai I believe that you are referring to the following branch: https://github.com/cedana/criu/pull/4

It would be good to open an issue or draft PR describing the use-case with an example (of a real) application that you are trying to migrate. This will help us to understand why this functionality is important and discuss what might be the best solution.

bsmithai:
https://github.com/checkpoint-restore/criu/pull/2683

I have a PR with a zdtm test open. The specific use case was a multi
process training job where multiple processes share the same semaphore, but
you’re migrating the whole training job.

PR #2688, opened 2025-07-11
images: remove symlink for descriptor.proto

rst0git:
The criu package's archive (criu-4.1.tar.gz) contains a symlink with an absolute path:

    criu-4.1/images/google/protobuf/descriptor.proto -> /usr/include/google/protobuf/descriptor.proto

This causes *fails to unpack* when using `tarfile.TarFile.extractall()` or `shutil.unpack_archive()` in Python 3.14.

This happens because Python 3.14 will switch the default behavior of extractall() from "fully trusting the content of archive" to "disallow common attack vectors while extracting the archive". With this new behavior, extractall() raises an exception when at least one file in the archive extracts or points to outside of the extraction directory (these are called path traversal attacks and zip slip attacks).

Reported-by: Dmitrii Kuvaiskii <dimakuv@amazon.de>


images/Makefile
```diff
@@ -91,6 +90,21 @@ endef
 
 makefile-deps := Makefile $(obj)/Makefile
 
+#
+# Generate descriptor.pb-c.c and descriptor.pb-c.h to compile opts.proto.
+PROTOBUF_DIR := images/google
+DESCRIPTOR_DIR := $(PROTOBUF_DIR)/protobuf
+$(DESCRIPTOR_DIR)/descriptor.pb-c.c: /usr/include/google/protobuf/descriptor.proto
```
avagin: should we do something like:
GOOGLE_INCLUDE=$(shell pkg-config protobuf --variable=includedir)/google/protobuf

rst0git: Thanks! I've updated the patch with this change.

dimakuv (approved):

rst0git:
@kolyshkin Thank you for fixing this!

Issue #2641, opened 2025-04-08
SIGILL during checkpoint


specht478:
Dear CRIU team,

When using CRIU 3.19 on debian bookworm to checkpoint a container, approximately 5% of the attempts fail with the following error messages in the log:

```
(00.246054) Parasite syscall_ip at 0x40002c0be000
(00.278562) Set up parasite blob using memfd
(00.278627) Putting parasite blob into 0xffff81c9f000->0x400137d16000
(00.278760) Dumping GP/FPU registers for 15222
(00.278795) Putting tsock into pid 15222
(00.279009) Wait for parasite being daemonized...
(00.279028) Wait for ack 2 on daemon socket
(00.279173) Error (criu/parasite-syscall.c:88): si_code=4 si_pid=15222 si_status=4
(00.279198) Error (criu/parasite-syscall.c:95): 15222 was stopped by 4 unexpectedly
```

According to the comments in parasite-syscall.c and the log output, this indicates that a SIGILL occurred while executing the parasite code.

The process running in the container has an installed custom signal handler for nearly all signals, including `SIGILL`. After CRIU sets the parasite blob into the process, will the old signal handler be still be trigerred? I am asking because if that is the case, I could set a breakpoint there with a hardware debugger (I am working with an ARM SoC) to have a look to the stack trace and determine what caused that `SIGILL`.

Do you have any other suggestions or insights on how can I proceed to debug and resolve this problem?

Thanks in advance and regards
gspecht478

avagin:
While the parasite code is running, `old` signal handlers should not be triggered. I think you need to modify CRIU to dump process registers and figure out what instruction triggers SIGILL.

specht478:
Hi @avagin,

Thanks for your reply. There is something I am not getting about the signal handlers here...

This is what the dump log shows:

```
00.222379) Parasite syscall_ip at 0x4000188a9000
(00.247969) Set up parasite blob using memfd
(00.248031) Putting parasite blob into 0xffffa1d5f000->0x4000e40a6000
(00.248150) Dumping GP/FPU registers for 28052
(00.248182) Putting tsock into pid 28052
(00.248356) Wait for parasite being daemonized...
(00.248373) Wait for ack 2 on daemon socket
pie: 1: Running daemon thread leader
(00.248427) Fetched ack: 2 2 0
(00.248444) Parasite 28052 has been switched to daemon mode
(00.248468) Sent msg to daemon 72 0 0
[...]

PR #2689, opened 2025-07-17
Fix SIGILL during checkpoint

specht478:
This PR addresses an issue where some parasited threads received SIGILL signals while communicating with the CRIU process in RPC mode. The problem was previously reported and analyzed in #2641

Following a suggestion by @adrianreber, this PR applies a similar fix as in commit 6be10a2, targeting locations where executable code is modified in memory. This helps prevent instruction cache incoherency, a behavior that is documented for the `arm` architecture.

### DIFF
--- a/compel/src/lib/infect.c
+++ b/compel/src/lib/infect.c
@@ -1054,6 +1054,12 @@ int compel_infect_no_daemon(struct parasite_ctl *ctl, unsigned long nr_threads, 
 
 	memcpy(ctl->local_map, ctl->pblob.hdr.mem, ctl->pblob.hdr.bsize);
 	compel_relocs_apply(ctl->local_map, ctl->remote_map, &ctl->pblob);
+	/*
+	 * Ensure the infected thread sees the updated code by flushing instruction and/or data caches as required by the
+	 * target architecture.
+	 * For arm64, it is important that the code in d-cache is flushed to RAM.
+ 	*/
+	__builtin___clear_cache(ctl->local_map, ctl->local_map + ctl->pblob.hdr.bsize);
 
 	p = parasite_size;
 
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2569,6 +2569,13 @@ static int remap_restorer_blob(void *addr)
 	restorer_setup_c_header_desc(&pbd, true);
 	compel_relocs_apply(addr, addr, &pbd);
 
+	/*
+	 * Ensure the infected thread sees the updated code by flushing instruction and/or data caches as required by the
+	 * target architecture.
+	 * For arm64, it is important that the code in d-cache is flushed to RAM.
+ 	*/
+	__builtin___clear_cache(addr, addr + pbd.hdr.bsize);
+
 	return 0;
 }
 


adrianreber:
Thanks for opening this PR.

avagin:
I'm quite skeptical about this change. In both cases, we're patching code that has never been executed, so it shouldn't be in a CPU instruction cache. One more thing about the dump change is that we're patching code from the CRIU process, but another process executes it.

Do you have a detailed explanation what is really going on here? Is there any examples of using cacheflush on arm64 in other projects?

specht478:
> The DCO checker is not happy with your commits:

Now I deselected the "keep my email private" option. I think that was the problem

specht478:
> > > The DCO checker is not happy with your commits:
> > 
> > 
> > Now I deselected the "keep my email private" option. I think that was the problem
> 
> You probably need to push your changes again to the branch.

Ok, thanks!

@avagin Thank you also for your comment. I am gathering the information and test results to provide a complete answer to it :)

specht478:
Hi @avagin,

our theory here is that a `SIGILL` is being generated because after modifying the executable code for the parasited process, the instruction cache contains old stale instructions. So the cache needs to be cleared after letting the parasited process to execute the code, as already done in a test by commit 6be10a2.

After calling `parasite_memfd_exange` (this happens in the CRIU process), the parasited thread (in the other process) will execute code under the virtual addresses of `ctl->remote_map`. Those addresses could map to physicall addresses that used to contain other code, and we did not invalidate the instruction cache. This would also explain that the error is so sporadic (about 6% of the times)

But, as you point out, this case is not quite the same as in commit 6be10a2, since we are talking here of two separated processes with different virtual memory spaces.

According to the [Arm Architecture Reference Manual for A-Profile architecture](https://developer.arm.com/documentation/ddi0487/lb/?lang=en), version `L.b`, section `G5.10.12`, the instruction caches can be VIPT or PIPT. For my board PIPT (Physically Indexed, Physically Tagged) is the case.
```
[    0.020280] Detected PIPT I-cache on CPU1
[    0.029068] Detected PIPT I-cache on CPU2
...
```
This means, calling `__builtin__clear_cache()` from the CRIU process invalidates cache lines based on physical addresses and this directly affects the parasited process.

I am not really sure if this would work for the VIPT (Virtually Indexed, Physically Tagged) case. But it might, according to[ this StackOverflow](https://stackoverflow.com/questions/70635862/synchronizing-caches-for-jit-self-modifying-code-on-arm/70684882#70684882) thread.

But even if the change had no effect for ARM with VIPT instruction caches or other architectures, I assume this call has a negligible performance cost.

If you wish, I can modify the PR title or the commit descriptions to make clear this change is for a specific architecture and a specific i-cache size.

This explanation is just theory, but my tests also verified (so far) this fix. Today I installed again a CRIU version without the fix, and after just 6 iterations of checkpoint, the SIGILL problem was there:

```
(02.172084) Set up parasite blob using memfd
(02.172111) Putting parasite blob into 0xffff86ce7000->0x400002b5f000
(02.172199) Dumping GP/FPU registers for 13639
(02.172225) Putting tsock into pid 13639
(02.172359) Wait for parasite being daemonized...
pie: 4: Running daemon thread leader
(02.172376) Wait for ack 2 on daemon socket
(02.172515) Error (criu/parasite-syscall.c:88): si_code=4 si_pid=13639 si_status=4
(02.172536) Error (criu/parasite-syscall.c:95): 13639 was stopped by 4 unexpectedly
pie: 4: Sent only -32 bytes while 12 expected
pie: 4: 4: new_sp=0x400002bee780 ip 0x400001e906c8
```

With the CRIU with this fix, after several tests this week and today, I did not see that problem anymore. Of course we will keep verifying this, but it looks good.

avagin:
The change looks good to me. @specht478, Could you update the comments with more details about what we believe is going on there and why we need to flush caches? Thanks again for working on this fix. @adrianreber, thank you for suggesting the idea of flushing caches.

rst0git:
> @specht478, Could you update the comments with more details about what we believe is going on there and why we need to flush caches?

@specht478 I agree with Andrei, it would be very helpful to include this information in the commit messages.

specht478:
Hi @avagin @rst0git , I added the description as discussed here to the commit messages

compel/src/lib/infect.c
```diff
@@ -1054,6 +1054,11 @@ int compel_infect_no_daemon(struct parasite_ctl *ctl, unsigned long nr_threads,
 
 	memcpy(ctl->local_map, ctl->pblob.hdr.mem, ctl->pblob.hdr.bsize);
 	compel_relocs_apply(ctl->local_map, ctl->remote_map, &ctl->pblob);
+	/*
+	 * Clear instruction cache to prevent SIGILL from stale instructions.
```
avagin: we clearing the data cache, instruction cache doesn't have any relevant entries, because the code hasn't been executed.

specht478: According to the [documentation](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html), `__builtin___clear_cache(start, end)` is used to flush the instruction cache for the specified address range. In our case, we call it from the CRIU process after modifying shared memory that contains the injected code. On architectures with PIPT (Physically Indexed, Physically Tagged) instruction caches, the virtual addresses passed to this function are translated to physical addresses. Since the memory is shared, the parasite process maps the same physical pages at different virtual addresses. When it executes the injected code, its virtual addresses resolve to the same physical memory, which has already been flushed from the instruction cache. This ensures the parasite process sees the updated instructions and avoids executing stale code.

specht478: Note: According to this stackoverflow comment: https://stackoverflow.com/a/77400268
`__builtin___clear_cache()` for arm64 will be expanded to `__aarch64_sync_cache_range()`, which first flushes the data cache

specht478: So actually the call to `__builtin___clear_cache` results on:
- Putting the code from d-cache to physicall memory (in case it was only in d-cache)
- Clearing the i-cache for that physical-memory range

avagin: again, nothing is in i-cache in this case. This code has never been executed, so it can't be in the i-cache... except I missed something...

avagin: If I don't miss anything, I think we just need to fix the comment to indicate that it is actually cleaning d-cache.

specht478: There are two processes here: for the CRIU Process, this data is in d-cache. But with the call to `__builtin___clear_cache`, we are also clearing the i-cache for the parasited thread. I am finishing a diagram to indicate what I mean. 
It is like  this situation:  https://community.arm.com/arm-community-blogs/b/architectures-and-processors-blog/posts/caches-and-self-modifying-code

But yes, I guess the comment must be updated to be more specific

specht478: This is what I mean. Of course, this situation does not always happen (the error appears sporadically). It is also only of relevance if everything happens on the same core.

image
-------

image

-------

image

avagin: The code can be cached in i-cache only when it is executed. Where do we execute the parasite code before patching it?

avagin: compel_infect_no_daemon creates a new anon shared vma (compel_map_exchange). Then it copies the parasite code into it and applies relocations (compel_relocs_apply). After that, we call __builtin___clear_cache.

specht478: Yes, but the problem arises when the parasited thread tries to execute the parasite code, and the cache has old stuff (the "blue" code of the diagram) in it. So the important thing is to clear the old stuff. It's not about putting the parasite code in the i-cache (that will happen when it is first executed). It is about clearing old stuff.

specht478: I updated the diagrams. I hope, i could make my point more clear.

I also updated the commit messages and the comment. Let me know what you think!

avagin: When you create a new mapping, the kernel guarantees that the i-cache has been cleaned up for the mmaped region. The initial state is no old stuff in i-cache. If you think is that the i-cache is still in effect, you need to call __builtin___clear_cache in the target process...

specht478: Hi, according to what I read these days, cache clearing is performed by the kernel in case the map is marked as PROT_EXEC, which is not the case in `infect.c:865`, `infect.c:871`. I am not even sure about that.

Calling `__builtin___clear_cache` after modifying code is mentioned in several websites for the code modification case.

Regarding your last point, it seems counterintuitive that one can affect the i-cache seen by the target process from the CRIU process. But this is the case for the `PIPT` i-cache, since the i-cache works on physical address level. This means: the CRIU process and the target process will see the mapping under different virtual addresses, but the call to `__builtin___clear_cache` will first translate the address to physical address and then clear. Therefore, the call in CRIU affects the target process.

Maybe this also works with VIPT i-cache, as explained here: https://stackoverflow.com/questions/70635862/synchronizing-caches-for-jit-self-modifying-code-on-arm/70684882#70684882 but there is no guarantee of this.

This is also mentioned in the commit message, since it is a bit confusing

avagin: @specht478 When a PTE is mapped to a process address space, both the i-cache and d-cache are dropped (see: https://github.com/torvalds/linux/blob/89be9a83ccf1f88522317ce02f854f30d6115c41/arch/arm64/include/asm/pgtable.h#L737). For the target process, pte_user_exec(pte) returns true for the parasite code, which means the i-cache will be correctly dropped for those parasite pages.

Frankly, I don't think we're on the same page. I've asked @mihalicyn to review this discussion and offer a third perspective.

mihalicyn: Dear friends,

first of all, @specht478 thanks for working on this fix!

Let's try to sync our understanding. I'll formulate a bunch of statements with numbering (please, mention which you think are not correct).

1. Instruction cache is filled **only** when CPU goes to fetch and execute an instruction.
2. Data cache is filled when CPU executes store/load instructions.

=> this means that unless the code in memory is executed i-cache is empty. This is what Andrei is saying.

3. When we allocate a memory from the kernel, kernel will drop all necessary caches for us (Andrei has given a link to the code where this happens).

Now let's analyze what we have in our case. You have a problem with a stable reproducer and fix for the problem which involves `__builtin___clear_cache`. Let's look into the implementation in GCC:
https://github.com/gcc-mirror/gcc/blob/c720869f0eed38b56d6fd4c6a7f7941e64be096c/libgcc/config/aarch64/sync-cache.c#L32

As we can see from the implementation this thing involves both d-cache and i-cache which is not surprising, because **when you modify a code in memory, these changes go to d-cache and not i-cache**.

In our case, in `parasite_memfd_exchange()` we allocate a fresh memory area and mmap it into the CRIU process address space with `PROT_READ | PROT_WRITE`, and also mmap to the victim's address space with `PROT_READ | PROT_EXEC` (also with `| PROT_WRITE` in some special circumstances).

4. Then, we copy a parasite code into that memory and this **fills d-cache**.
5. Then we apply relocations which can also only affect d-cache and not i-cache.
6. [ your fix ] You have added `__builtin___clear_cache(` calls which cleans up d-cache & i-cache, but what really important here is d-cache, because i-cache is empty as nobody yet tried to fetch this data from memory as a CPU instruction and execute it. Until this point our code in memory is just a data.
7. Then at some point this code is actually getting executed, CPU goes and fills i-cache with a consistent parasite code (with relocations applied, and so on).

Now let's try to think what happens without your fix (step 6). After step 5, we have some parts of parasite's code in physical memory, some parts in d-cache. i-cache is empty. We go next and try to execute parasite's code, what happens is CPU goes and tries to fetch an instruction. i-cache is empty and CPU goes to a physical memory to load an instruction and put it into i-cache. But physical **memory state is not consistent at this point**, as some parts of parasite's code **remains in d-cache**.

To conclude, your current fix code is completely makes sense and correct but what we are trying to do here is to agree on the explanation of this fix and write a good comment. CRIU is a pretty complex thing and in our project we are trying to keep the following statement true "For any given code line X in CRIU there is at least (!) one developer around CRIU project who can explain it :)".

mihalicyn: btw, if you have a time and want to have fun, you can try to copy `__aarch64_sync_cache_range` function from https://github.com/gcc-mirror/gcc/blob/c720869f0eed38b56d6fd4c6a7f7941e64be096c/libgcc/config/aarch64/sync-cache.c#L32 and call it directly instead of `__builtin___clear_cache`. Ensure that it still works and fixes the problem, and after that try to remove the part of code which cleans i-cache, to see if cleaning d-cache is enough. It is not a proposal for your fix change at all, your current approach with `__builtin___clear_cache` is great, but this would allows us to experimentally understand if i-cache is anyhow involved or not. ;)

specht478: Hi @avagin and @mihalicyn,

First of all, thank you very much for taking the time to discuss this topic in such detail. It's a somewhat complex issue, and it's also important for me to fully understand the implications of this fix :)

✅ **TL;DR:** I see your point now. I fully understand and agree with your reasoning. I’ll go ahead and update the comments and commit messages accordingly.

I was missing point 3 (thanks @avagin for the reference). My theory was: the i-cache contains old stuff that does not get updated and the parasited thread fetches that. But this is erroneous, since the i-cache is cleared after mmap. The correct explanation for the problem is, as written by @mihalicyn:

> Now let's try to think what happens without your fix (step 6). After step 5, we have some parts of parasite's code in physical memory, some parts in d-cache. i-cache is empty. We go next and try to execute parasite's code, what happens is CPU goes and tries to fetch an instruction. i-cache is empty and CPU goes to a physical memory to load an instruction and put it into i-cache. But physical memory state is not consistent at this point, as some parts of parasite's code remains in d-cache.

specht478: > btw, if you have a time and want to have fun, you can try to copy `__aarch64_sync_cache_range` function from https://github.com/gcc-mirror/gcc/blob/c720869f0eed38b56d6fd4c6a7f7941e64be096c/libgcc/config/aarch64/sync-cache.c#L32 and call it directly instead of `__builtin___clear_cache`. Ensure that it still works and fixes the problem, and after that try to remove the part of code which cleans i-cache, to see if cleaning d-cache is enough. It is not a proposal for your fix change at all, your current approach with `__builtin___clear_cache` is great, but this would allows us to experimentally understand if i-cache is anyhow involved or not. ;)

Update: I am testing this but it takes time. I hope I can post some results tomorrow. The good thing is: until now did not see any SIGILL!

specht478: Hi @mihalicyn @avagin @adrianreber,

i performed the suggested test, and I have some good news 🎉

| Testcase                              | Test result               |
|---------------------------------------|---------------------------|
| `__aarch64_sync_cache_range` copied 1:1 | ✅ No SIGILL                 |
| i-cache flush commented out, d-cache flush active           | ✅ No SIGILL                 |
| d-cache flush commented out, i-cache flush active           | ❌ SIGILL after 7 iterations |

This is the "empirical" confirmation of @avagin's and @mihalicyn's theory!

I adapted the comments and commit messages. Please let me know what you think.

mihalicyn: Thanks for doing this investigation! Really cool stuff. We all learned something new today :)

adrianreber: @mihalicyn @specht478 @avagin Indeed, that was real thorough and good review!

avagin:
Merged. Thanks!

PR #2691, opened 2025-07-23
Keep images/google/protobuf directory

kolyshkin:
Commit 68f92b551 removed `images/google/protobuf` directory, so it is re-created each time during the build process.

This resulted in a weird behavior change. Previously, one could do something like this:

	git clone $CRURL criu
	(cd criu && sudo make install-criu)
	rm -rf criu

This worked fine, including running rm -rf as a non-root user, since no new directories were created under criu -- all directories were still owned by the original user.

Since commit 68f92b551 the same sequence fails:

	rm: cannot remove '/home/runner/criu/images/google/protobuf/descriptor.pb-c.c': Permission denied
	rm: cannot remove '/home/runner/criu/images/google/protobuf/descriptor.pb-c.d': Permission denied
	rm: cannot remove '/home/runner/criu/images/google/protobuf/descriptor.pb-c.h': Permission denied

A workaround is to keep empty images/google/protobuf directory, which is what this commit does.


kolyshkin:
Added a second commit with a tiny nit to images/Makefile

PR #2693, opened 2025-07-23
fix mount invalid argument error in kernel 6.6

duguhaotian:

### DIFF
--- a/criu/mount-v2.c
+++ b/criu/mount-v2.c
@@ -933,12 +933,19 @@ static int move_mount_set_group(int src_id, char *source, int dst_id)
 
 static int restore_one_sharing(struct sharing_group *sg, struct mount_info *target)
 {
-	char target_path[PATH_MAX];
+	char tfd_path[PATH_MAX] = {0};
+	char target_path[PATH_MAX] = {0};
 	int target_fd;
 
 	target_fd = fdstore_get(target->mnt_fd_id);
 	BUG_ON(target_fd < 0);
-	snprintf(target_path, sizeof(target_path), "/proc/self/fd/%d", target_fd);
+	snprintf(tfd_path, sizeof(tfd_path), "/proc/self/fd/%d", target_fd);
+
+	if (readlink(tfd_path, target_path, sizeof(target_path)) < 0) {
+		pr_perror("Failed to readlink %s", tfd_path);
+		close(target_fd);
+		return -1;
+	}
 
 	/* Restore target's master_id from shared_id of the source */
 	if (sg->master_id) {


criu/mount-v2.c
```diff
@@ -933,12 +933,19 @@ static int move_mount_set_group(int src_id, char *source, int dst_id)
...
+	char tfd_path[PATH_MAX] = {0};
+	char target_path[PATH_MAX] = {0};
 	int target_fd;
 
 	target_fd = fdstore_get(target->mnt_fd_id);
 	BUG_ON(target_fd < 0);
-	snprintf(target_path, sizeof(target_path), "/proc/self/fd/%d", target_fd);
+	snprintf(tfd_path, sizeof(tfd_path), "/proc/self/fd/%d", target_fd);
+
+	if (readlink(tfd_path, target_path, sizeof(target_path)) < 0) {
```
mihalicyn, line 944: Can you explain how this helps with https://github.com/torvalds/linux/commit/12f147ddd6de7382dad54812e65f3f08d05809fc check? `check_mnt(mnt)` operates on a mount level, and procfs symlink should be resolved by this time already inside the kernel.

```

int do_mount(const char *dev_name, const char __user *dir_name,
		const char *type_page, unsigned long flags, void *data_page)
{
	struct path path;
	int ret;

	ret = user_path_at(AT_FDCWD, dir_name, LOOKUP_FOLLOW, &path); // here /proc/self/fd/<N> gets resolved into the real mount path. Isn't it?
	if (ret)
		return ret;
	ret = path_mount(dev_name, &path, type_page, flags, data_page);
```

mihalicyn:
Also, you mean 6.16+, not 6.6?

duguhaotian:
> Also, you mean 6.16+, not 6.6?

sorry, it's 6.16, but i backport to 6.6. so, have some problew.

criu/mount-v2.c
duguhaotian: we debug in kernel, found not resovled at check_mnt().
maybe it's bug of kernel, but we can use this pr to bypass this error.

duguhaotian: /proc/self/fd/${fd} is send from FDSTORE_SK_OFF, we guess maybe this fd related with root namespace in kernel

duguhaotian:
ZDTM error log
```
=== Run 154/478 =====----------- zdtm/static/mnt_root_ext
====================== Run zdtm/static/mnt_root_ext in ns ======================
Start test
Test is SUID
./mnt_root_ext --pidfile=mnt_root_ext.pid --outfile=mnt_root_ext.out --dirname=mnt_root_ext.test
Running zdtm/static/mnt_root_ext.hook(--post-start)
Running zdtm/static/mnt_root_ext.hook(--pre-dump)
Run criu dump
Running zdtm/static/mnt_root_ext.hook(--pre-restore)
Run criu restore
=[log]=> dump/zdtm/static/mnt_root_ext/48/1/restore.log
b'(00.003325)      1: No ipcns-sem-11.img image'
b'(00.004028)      1: net: Try to restore a link 10:1:lo'
b'(00.004033)      1: net: Restoring link lo type 1'
b'(00.004479)      1: net: \tRunning ip addr restore'
b'(00.040335)      1: mnt-v2: Move mount 522 from /tmp/.criu.mntns.aNlKfv/mnt-0000000522 to /tmp/.criu.mntns.aNlKfv/14-0000000000/dev/pts'
b'(00.040344)      1: mnt-v2: Move mount 554 from /tmp/.criu.mntns.aNlKfv/mnt-0000000554 to /tmp/.criu.mntns.aNlKfv/14-0000000000/proc'
b'(00.040352)      1: mnt-v2: Move mount 555 from /tmp/.criu.mntns.aNlKfv/mnt-0000000555 to /tmp/.criu.mntns.aNlKfv/14-0000000000/run'
b'(00.040357)      1: mnt: Move the root to /tmp/.criu.mntns.aNlKfv/14-0000000000'
b'(00.040563)      1: Error (criu/mount-v2.c:979): mnt-v2: Failed to make mount 552 slave: No such file or directory'
b"(00.041189) Error (criu/mount.c:3700): mnt: Can't remove the directory /tmp/.criu.mntns.aNlKfv: Device or resource busy"
b'(00.041203) Error (criu/cr-restore.c:2324): Restoring FAILED.'
############## Test zdtm/static/mnt_root_ext FAIL at CRIU restore ##############
```

criu/mount-v2.c
avagin: This fix doesn't look like the right one. We want to do changes on the specific mount, so we need to enter its mount namespace and do changes from it.

duguhaotian: > This fix doesn't look like the right one. We want to do changes on the specific mount, so we need to enter its mount namespace and do changes from it.

So, we this usecase is required in criu. Maybe kernel check patch is wrong.

PR #2695, opened 2025-07-24
mount-v2: enter the mount namesapce to propagation properties

### LINKED ISSUE #2694: restore failure on Linux 6.6+: mount ns setup fails due to kernel mount() syscall change
restore failure on Linux 6.6+, which commit is https://github.com/torvalds/linux/commit/12f147ddd6de7382dad54812e65f3f08d05809fc
current report error is:

(00.034502)      7: Error (criu/mount-v2.c:966): mnt-v2: Failed to make mount 476 slave: Invalid argument
(00.034694) Error (criu/mount.c:3689): mnt: Can't remove the directory /tmp/.criu.mntns.iLwsbF: Device or resource busy
(00.034723) Error (criu/cr-restore.c:2637): Restoring FAILED.
(00.035067) Error (criu/cr-restore.c:1536): 30584 killed by signal 9: Killed
(00.035153) Error (criu/namespaces.c:235): Can't open 30584/ns/net on procfs: No such file or directory
(00.035626) Error (criu/cgroup.c:1970): cg: cgroupd: recv req error: No such file or directory

avagin:
A kernel change (commit 12f147ddd6de, "do_change_type(): refuse to operate on unmounted/not ours mounts") modified how mount propagation properties can be changed. Previously, these properties could be changed from any mount namespace. Now, they can only be modified from the specific mount namespace where the target mount is actually mounted

This commit addresses this new restriction by ensuring that CRIU enters the correct mount namespace before attempting to restore mount propagation properties (MS_SLAVE or MS_SHARED) for a mount.

Fixes #2694

### DIFF
--- a/criu/mount-v2.c
+++ b/criu/mount-v2.c
@@ -933,8 +933,12 @@ static int move_mount_set_group(int src_id, char *source, int dst_id)
 
 static int restore_one_sharing(struct sharing_group *sg, struct mount_info *target)
 {
+	int nsfd = -1, orig_nsfd = -1, exit_code = -1;
 	char target_path[PATH_MAX];
-	int target_fd;
+	int target_fd = -1;
+
+	if (!sg->master_id && !sg->shared_id)
+		return 0;
 
 	target_fd = fdstore_get(target->mnt_fd_id);
 	BUG_ON(target_fd < 0);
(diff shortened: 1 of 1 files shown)


criu/mount-v2.c
```diff
@@ -950,7 +954,7 @@ static int restore_one_sharing(struct sharing_group *sg, struct mount_info *targ
 			if (move_mount_set_group(first->mnt_fd_id, NULL, target->mnt_fd_id)) {
 				pr_err("Failed to copy sharing from %d to %d\n", first->mnt_id, target->mnt_id);
 				close(target_fd);
-				return -1;
+				goto err;
```
Snorch, line 956: If we `close_safe(&target_fd)` in `err` label, maybe we can remove `close(target_fd)` everywhere before `goto err`?

Snorch (approved):
Looks good!

criu/mount-v2.c
avagin: You are right. Done. Thanks.

rst0git (approved):
LGTM

Issue #2705, opened 2025-08-10
Fix sockopts handling for 6.16+ kernels


mihalicyn:
On 6.16+ kernels the following tests are failing (https://github.com/checkpoint-restore/criu/pull/2652/checks?check_run_id=47767998459):
```
################### 78 TEST(S) FAILED (TOTAL 479/SKIPPED 41) ###################
 * zdtm/static/bridge(uns)
 * zdtm/static/net_lock_socket_iptables(h)
 * zdtm/static/net_lock_socket_iptables6(h)
 * zdtm/static/net_lock_socket_nftables(h)
 * zdtm/static/net_lock_socket_nftables6(h)
 * zdtm/static/packet_sock(uns)
 * zdtm/static/packet_sock_mmap(h)
 * zdtm/static/packet_sock_spkt(uns)
 * zdtm/static/sk-freebind(ns)
 * zdtm/static/sk-freebind-false(uns)
 * zdtm/static/sock_filter00(ns)
 * zdtm/static/sock_filter01(uns)
 * zdtm/static/sock_ip_opts00(ns)
 * zdtm/static/sock_ip_opts01(ns)
 * zdtm/static/sock_opts00(ns)
 * zdtm/static/sock_opts01(uns)
 * zdtm/static/sock_opts02(ns)
 * zdtm/static/sock_tcp_opts00(ns)
 * zdtm/static/sock_tcp_opts01(uns)
 * zdtm/static/socket-dump-tcp-close(uns)
 * zdtm/static/socket-linger(ns)
 * zdtm/static/socket-raw(ns)
 * zdtm/static/socket-tcp(h)
 * zdtm/static/socket-tcp-close0(uns)
 * zdtm/static/socket-tcp-close-wait(h)
 * zdtm/static/socket-tcp-close2(ns)
 * zdtm/static/socket-tcp-close1(uns)
 * zdtm/static/socket-tcp-closed-last-ack(uns)
 * zdtm/static/socket-tcp-closed(uns)
 * zdtm/static/socket-tcp-fin-wait1(h)
 * zdtm/static/socket-tcp-closing(uns)
 * zdtm/static/socket-tcp-fin-wait2(h)
 * zdtm/static/socket-tcp-ipt-nfconntrack(h)
 * zdtm/static/socket-tcp-keepalive(h)
 * zdtm/static/socket-tcp-listen(uns)
 * zdtm/static/socket-tcp-last-ack(uns)
[...]

PR #2711, opened 2025-08-13
Fix: getsockopt usage for SO_PASSCRED/SO_PASSSEC on Linux 6.16+Fix socket opts

DongSunchao:
Linux 6.16 changed the behavior of getsockopt for credential-related options and it restricted `SO_PASS{CRED,PIDFD,SEC}` to `AF_{UNIX,NETLINK,BLUETOOTH}` .
This patch updates the logic to correctly handle` SO_PASSCRED` and `SO_PASSSEC`,
ensuring compatibility with newer kernels while preserving support for older ones. 

Besides, **/test/zdtm/static/sock_opts00.c** has been changed to check the kernel version so that `PF_INET` can skip the  `SO_PASSCRED` and `SO_PASSSEC` options.

Tested with: ./test/zdtm.py run -t zdtm/static/sock_opts00 -f ns

Fixes: #2705


criu/sockets.c
```diff
@@ -688,13 +691,22 @@ int dump_socket_opts(int sk, SkOptsEntry *soe)
...
-	soe->so_passcred = val ? true : false;
-
-	ret |= dump_opt(sk, SOL_SOCKET, SO_PASSSEC, &val);
-	soe->has_so_passsec = true;
-	soe->so_passsec = val ? true : false;
+	/*
+	 * Restrict SO_PASS{CRED,PIDFD,SEC} to AF_{UNIX,NETLINK,BLUETOOTH} while
+	 * kernel >= 6.16.
+	 */
+	if (getsockname(sk, (struct sockaddr *)&addr, &addrlen) == -1) {
```
mihalicyn: instead of this, you can do:
```
	int family;
..
	if (dump_opt(sk, SOL_SOCKET, SO_DOMAIN, &family))
		return -1;
```

and then use `family` value to make a decision if you need to dump `SO_PASSCRED` and `SO_PASSSEC` or not (only for `AF_UNIX`, `AF_NETLINK`, `AF_BLUETOOTH`, but we don't support bluetooth sockets anyways). Please notice that if can also be a matter of kernel configuration, specifically, `SO_PASSSEC` is available only if kernel is built with `CONFIG_SECURITY_NETWORK=Y`.

I personally think that we should just check if `EOPNOTSUPP` is returned from `getsockopt` for these two options and ignore it. This is the simplest and clean enough solution, IMHO.

DongSunchao: Hi,
I've removed all kernel version checks and make some changes.
In **sockets.c**, I've restricted `SO_PASS{CRED,PIDFD,SEC}` to `AF_{UNIX,NETLINK}` based on the `family` value. 
And in test file **sock_opts00.c** I've ignore these two options according to `vname[i].opt`.
This seems to work correctly in my respository actions, but I'm unsure if this is the correct approach. I would appreciate any feedback or suggestions on potential issues.
Thank you!

criu/sockets.c
```diff
@@ -635,8 +635,9 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 int do_dump_opt(int sk, int level, int name, void *val, int len)
 {
 	socklen_t aux = len;
+	int result = getsockopt(sk, level, name, val, &aux);
 
-	if (getsockopt(sk, level, name, val, &aux) < 0) {
+	if (result < 0 && result != -ENOPROTOOPT) {
```
avagin: The error code will be in `errno`.

DongSunchao: Thank you for pointing it out. I have changed it.

avagin:
Hi @DongSunchao , please read the contribution guidelines at https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html and reformat your commits. The command `git rebase -i origin/criu-dev` should be helpful.

criu/sockets.c
```diff
@@ -635,8 +635,9 @@ int restore_socket_opts(int sk, SkOptsEntry *soe)
 int do_dump_opt(int sk, int level, int name, void *val, int len)
 {
 	socklen_t aux = len;
+	int result = getsockopt(sk, level, name, val, &aux);
 
-	if (getsockopt(sk, level, name, val, &aux) < 0) {
+	if (result < 0 && errno != -ENOPROTOOPT) {
```
avagin: why do you need to check ENOPROTOOPT here?

test/zdtm/static/sock_opts00.c
```diff
@@ -83,8 +83,13 @@ int main(int argc, char **argv)
 			pr_perror("can't verify %s", vname[i].name);
 			return 1;
 		}
-
+		//for kernel version >= 6.16.0 Restrict
```
avagin: pls fix the comment style:
```
/* single line */

/*
 * multiline
 * multiline
 */
```

criu/sockets.c
DongSunchao: I've removed the `ENOPROTOOPT` check. It seems I misunderstood something, so I'm reverting the code to its original state.

test/zdtm/static/sock_opts00.c
DongSunchao: Thanks for your patience. I know my previous commits had issues. This is my first open-source contribution, and I've learned a lot from this experience. I've noted down all my mistakes to avoid them in the future.

DongSunchao:
> @DongSunchao Would it be possible to reset the author of your commits to yourself (instead of `root`) and use messages similar to the commits in #2719?
> 
> The following blog post provides more information on how to write good commit messages and why this is important: https://cbea.ms/git-commit/#seven-rules

I've made the requested changes and have done my best to follow the guidelines. However, I've run into a new issue: my code passes the **Fedora Rawhide** test in my repository action but fails the **Vagrant Fedora Rawhide** test on this PR. I'm not sure how to resolve it.

Snorch:
Having a hunk adding lines in first patch
```
--- a/test/zdtm/static/sock_opts00.c
+++ b/test/zdtm/static/sock_opts00.c
@@ -83,8 +83,15 @@ int main(int argc, char **argv)
                        pr_perror("can't verify %s", vname[i].name);
                        return 1;
                }
-
+               /*
+               * for kernel version >= 6.16.0 Restrict
+               * SO_PASS{CRED,PIDFD,SEC} to AF_{UNIX,NETLINK,BLUETOOTH}
+               */
                if (val[i] != rval) {
+                       if (vname[i].opt == SO_PASSCRED || vname[i].opt == SO_PASSSEC) {
+                               continue;
+                       }
+
                        errno = 0;
                        fail("%s changed: %d -> %d", vname[i].name, val[i], rval);
                        return 1;
```
And then having hunk removing those same lines in second patch:
```
@@ -78,28 +85,16 @@ int main(int argc, char **argv)
        test_waitsig();
 
        for (i = 0; i < NOPTS; i++) {
-               ret = getsockopt(sock, SOL_SOCKET, vname[i].opt, &rval, &len);
+               sk = vname[i].opt == SO_PASSCRED || vname[i].opt == SO_PASSSEC ? usock : sock;
+               ret = getsockopt(sk, SOL_SOCKET, vname[i].opt, &rval, &len);
                if (ret) {
                        pr_perror("can't verify %s", vname[i].name);
                        return 1;
                }
-               /*
-               * for kernel version >= 6.16.0 Restrict
-               * SO_PASS{CRED,PIDFD,SEC} to AF_{UNIX,NETLINK,BLUETOOTH}
-               */
-               if (val[i] != rval) {
-                       if (vname[i].opt == SO_PASSCRED || vname[i].opt == SO_PASSSEC) {
-                               continue;
-                       }
-
-                       errno = 0;
-                       fail("%s changed: %d -> %d", vname[i].name, val[i], rval);
-                       return 1;
-               }
        }
 
        pass();
        close(sock);
```

Also by those hunks you remove important `if (val[i] != rval) {` value persistence check, which is not what we want, please return it.  
is generally discouraged. Please remove the hunk touching zdtm from the first patch completely.

criu/sockets.c
```diff
@@ -688,13 +688,21 @@ int dump_socket_opts(int sk, SkOptsEntry *soe)
 	soe->so_reuseport = val ? true : false;
 	soe->has_so_reuseport = true;
 
-	ret |= dump_opt(sk, SOL_SOCKET, SO_PASSCRED, &val);
-	soe->has_so_passcred = true;
-	soe->so_passcred = val ? true : false;
+	switch (family) {
+	case AF_UNIX:
+
```
Snorch: Excess newline between case statements, please remove.

DongSunchao: I have removed it.

test/zdtm/static/sock_opts00.c
```diff
@@ -78,21 +85,21 @@ int main(int argc, char **argv)
...
 		if (ret) {
 			pr_perror("can't verify %s", vname[i].name);
 			return 1;
 		}
 
-		if (val[i] != rval) {
-			errno = 0;
-			fail("%s changed: %d -> %d", vname[i].name, val[i], rval);
+		if (rval != val[i]) {
+			pr_err("%s changed: %d -> %d", vname[i].name, val[i], rval);
```
Snorch: Excess hunk -> please remove.

criu/sockets.c
```diff
@@ -688,13 +688,20 @@ int dump_socket_opts(int sk, SkOptsEntry *soe)
...
 
-	ret |= dump_opt(sk, SOL_SOCKET, SO_PASSSEC, &val);
-	soe->has_so_passsec = true;
-	soe->so_passsec = val ? true : false;
+		ret |= dump_opt(sk, SOL_SOCKET, SO_PASSSEC, &val);
+		soe->has_so_passsec = true;
+		soe->so_passsec = val ? true : false;
+		break;
+	default:
+		break;
```
Snorch: I don't really like switch case use, it leads to excess lines as opposed to simple if. Moreover "default" switch branch is completely unused.

DongSunchao: Thank you for pointing it out. I’ve made the change.

Snorch (approved):
Looks good. Thank you!

DongSunchao:
> Looks good. Thank you!

Hi, there is another issue related to the one I mentioned previously. The code fails the **Vagrant Fedora Rawhide based test** with an error in `vdso-proxy`. The specific issue is that it's creating three new VMAs and incorrectly incorporating one existing VMA. This behavior is unexpected.

However, I've tried running the same test on my local virtual machine (kernel version **6.17.0-0.rc1.250815gd7ee5bdce789.21.fc44.x86_64**, **Vagrant Fedora**), and it passes successfully.

I find this very strange and suspect the issue might be related to the CI environment.

Snorch:
> > Looks good. Thank you!
> 
> Hi, there is another issue related to the one I mentioned previously. The code fails the **Vagrant Fedora Rawhide based test** with an error in `vdso-proxy`. The specific issue is that it's creating three new VMAs and incorrectly incorporating one existing VMA. This behavior is unexpected.
> 
> However, I've tried running the same test on my local virtual machine (kernel version **6.17.0-0.rc1.250815gd7ee5bdce789.21.fc44.x86_64**, **Vagrant Fedora**), and it passes successfully.
> 
> I find this very strange and suspect the issue might be related to the CI environment.

1) I don't see `[Vagrant Fedora Rawhide based test](https://github.com/checkpoint-restore/criu/pull/2711/checks?check_run_id=48489724887)` fail with `vdso-proxy` now. And the fail I see is about `zombie00 segfault` which is likely unrelated as it happens in other PRs (e.g. https://github.com/checkpoint-restore/criu/pull/2681/checks?check_run_id=43934303793).

2) Changes requested here https://github.com/checkpoint-restore/criu/pull/2711#pullrequestreview-3114654710 are seemingly done.

 So probably we are clear to merge.

PR #2712, opened 2025-08-15
make: Disable branch-protection for PIE code on ARM64

### LINKED ISSUE #2709: criu 4.1.1 times out in runc c/r tests on GHA arm (ubuntu-24.04-arm)
Here's the situation with ubuntu-24.04-arm and criu packages:

 - when criu-dev HEAD is used, it works fine;
 - when 4.1-1 from https://download.opensuse.org/repositories/devel:/tools:/criu/xUbuntu is used, it ends up being stuck on restore, and the test is aborted via a timeout.
 
Here you can find all the c/r logs for a test which timed out: [test (ubuntu-24.04-arm, 1.24.x)](https://github.com/opencontainers/runc/actions/runs/16841823814/job/47714378172?pr=4844#logs), [raw logs](https://productionresultssa19.blob.core.windows.net/actions-results/9ce9f718-9e3c-4ff3-a81d-2a7e3c2e5fe7/workflow-job-run-2ed77145-8b4a-51bd-8fb3-6be534d2a39e/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-08-08T23%3A23%3A16Z&sig=eeD6kxa5eOmWoNRwlD7GBjlLnWlPamDt0Ei64Gmupw0%3D&ske=2025-08-09T10%3A58%3A17Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-08-08T22%3A58%3A17Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-08-08T23%3A13%3A11Z&sv=2025-05-05)
 
Here is a job with criu-dev which doesn't fail: https://github.com/opencontainers/runc/actions/runs/16841823814/job/47714378171?pr=4844

It's the same with criu-4.1.1 from opensuse build farm, see e.g. https://github.com/opencontainers/runc/actions/runs/16920767253/job/47946071066

I'll be happy to help with reproducing this, testing any fixes etc.

avagin:
Branch protection uses PAC. It cryptographically "signs" a function's return address before it is stored on the stack. Upon return, the address is authenticated using a secret key. If the signature is invalid, the program will fault.

The PIE code is used for the parasite and the restorer. In both cases, it runs in a foreign process. The case of the restorer is even trickier because it needs to restore the original PAC keys, which invalidates all previously "signed" pointers within the restorer itself.

Fixes #2709


rst0git:
@avagin Compiling CRIU with this patch fails on Debian 10 / Raspbian 10 with the following error:

```
gcc: error: unrecognized command line option '-mbranch-protection=none'; did you mean '-fcf-protection=none'?
```
The `branch-protection` option was introduced relatively recently and these distributions use gcc 8.3.0.
https://gcc.gnu.org/pipermail/gcc-patches/2022-April/593736.html

rst0git (approved):
LGTM

PR #2715, opened 2025-08-16
test/zdtm/static/maps12: fix pointer-to-int cast

rst0git:
The `offset` argument to `mmap()` was computed with a direct cast from pointer to `off_t`:
    
    (off_t)addr_hint - (off_t)map_base
    
This causes a build failure when compiling since pointers and `off_t` may differ in size on some platforms. The fix in this patch is to cast both pointers to `intptr_t`, perform the subtraction in that type, and then cast the result back to `off_t`.


mihalicyn (approved):
Thanks!

PR #2717, opened 2025-08-18
Handle processes with uprobes vma

shashank-mahadasyam:
Currently, uprobes vma are treated as unsupported which causes criu to error with "Unsupported mapping". This PR fixes that.

A uprobes vma is automatically mapped by the kernel when a uprobes location is hit. The TL;DR version of how uprobes works is, a software breakpoint is placed at the uprobe'd location. When execution reaches this place, control is transferred over to the kernel, which then sets up this special vma, if it hasn't done so already. The kernel executes whatever uprobe hooks it has to, and then executes the instruction which was replaced by the software breakpoint. This happens out-of-line in the special vma. For more details, checkout this commit: https://github.com/torvalds/linux/commit/d4b3b6384f98f8692ad0209891ccdbc7e78bbefe

This vma is not unmapped even when once active uprobes are deleted. So the presence of the vma does not necessarily mean an active uprobe, i.e. software breakpoint instruction replacement.

- Commits 1-3 add the new VMA_AREA_UPROBES flag.
- Commit 4 introduces the new `--allow-uprobes` option, and teaches criu to deal with the uprobes vma.
- Commit 5 adds the documentation for this new option.

I'm not sure if adding a new option is the best way to solve this problem or not. So, if there are improvements, then I'll rework this PR. If this is accepted, then I'll send another PR adding some tests for this option.

Fixes: checkpoint-restore#1961

### DIFF
--- a/test/zdtm/static/uprobes.c
+++ b/test/zdtm/static/uprobes.c
@@ -1,0 +1,295 @@
+#include <fcntl.h>
+#include <gelf.h>
+#include <libelf.h>
+#include <limits.h>
+#include <signal.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <tracefs.h>
+#include <unistd.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Test the --allow-uprobes option";
+const char *test_author = "Shashank Balaji <shashank.mahadasyam@sony.com>";
+
+#define UPROBE_GROUP_NAME	"zdtm"
+#define UPROBE_EVENT_NAME	"uprobes_test"
+#define UPROBED_FUNCTION	uprobe_target
+
+/*
+ * A uprobe can be set at the start of a function, but not all instructions
+ * will trigger the creation of a uprobes vma.
+ *
+ * Examples:
+ * - aarch64: if the function is a single `ret`, then no vma creation
+ * - x64: if the function is `nop; ret`, then no vma creation
+ *
+ * So to guarantee vma creation, create a volatile dummy variable (to prevent
+ * compiler optimization) and use it (to prevent "unused variable" warning)
+ */
+void UPROBED_FUNCTION(void) {
+	volatile int dummy = 0;
+	dummy += 1;
+}
+/* Calling via volatile function pointer ensures noinline at callsite */
+typedef void (*func_ptr)(void);
+volatile func_ptr uprobe_target_alias = UPROBED_FUNCTION;
+
+struct uprobe_context {
+	struct tracefs_instance *instance;
+	struct tracefs_dynevent *uprobe;
+};
+
+volatile bool got_sigtrap = false;
+
+/*
+ * Returns the file offset of a symbol in the executable of this program
+ * Returns 0 on failure
+*/
+uint64_t calc_sym_offset(const char *sym_name)
+{
+	GElf_Shdr section_header;
+	Elf_Scn *section = NULL;
+	Elf_Data *symtab_data;
+	uint64_t offset = 0;
+	char buf[PATH_MAX];
+	GElf_Sym symbol;
+	ssize_t n_bytes;
+	int n_entries;
+	Elf *elf;
+	int fd;
+	int i;
+
+	if (elf_version(EV_CURRENT) == EV_NONE) {
+		pr_err("ELF version of libelf is lower than that of the program\n");
+		return 0;
+	}
+
+	n_bytes = readlink("/proc/self/exe", buf, sizeof(buf));
+	if (n_bytes < 0) {
+		pr_perror("Failed to readlink /proc/self/exe");
+		return 0;
+	}
+	buf[n_bytes] = '\0';
+
+	fd = open(buf, O_RDONLY);
+	if (fd < 0) {
+		pr_perror("Failed to open self-executable");
+		return 0;
+	}
+
+	elf = elf_begin(fd, ELF_C_READ, NULL);
+	if (!elf) {
+		pr_err("%s\n", elf_errmsg(elf_errno()));
+		goto out_fd;
+	}
+
+	/* Look for the symbol table section and its header */
+	while ((section = elf_nextscn(elf, section)) != NULL) {
+		gelf_getshdr(section, &section_header);
+		if (section_header.sh_type == SHT_SYMTAB)
+			break;
+	}
+	if (!section) {
+		pr_err("Failed to find symbol table\n");
+		goto out_elf;
+	}
+	symtab_data = elf_getdata(section, NULL);
+	n_entries = section_header.sh_size / section_header.sh_entsize;
+
+	/* Look for a symbol with the required name */
+	for (i = 0; i < n_entries; i++) {
+		gelf_getsym(symtab_data, i, &symbol);
+		/* Symbol table's sh_link is the index of the string table section header */
+		if (!strcmp(sym_name,
+			    elf_strptr(elf, section_header.sh_link, symbol.st_name)))
+			break;
+	}
+	if (i == n_entries) {
+		pr_err("Failed to find symbol \"%s\"\n", sym_name);
+		goto out_elf;
+	}
+
+	/* Get the section the symbol belongs to (mostly .text) */
+	section = elf_getscn(elf, symbol.st_shndx);
+	gelf_getshdr(section, &section_header);
+	offset = symbol.st_value - section_header.sh_addr + section_header.sh_offset;
+
+out_elf:
+	elf_end(elf);
+out_fd:
+	close(fd);
+	return offset;
+}
+
+/*
+ * Set and enable a uprobe on the file at the given offset
+ * Returns struct uprobe_context with members set to NULL on failure
+*/
+struct uprobe_context enable_uprobe(const char *file, uint64_t offset)
+{
+	struct tracefs_instance *trace_instance;
+	struct tracefs_dynevent *uprobe;
+	struct uprobe_context context = {};
+
+	trace_instance = tracefs_instance_create("zdtm_uprobes_test");
+	if (!trace_instance) {
+		pr_perror("Failed to create tracefs instance");
+		return context;
+	}
+	tracefs_instance_reset(trace_instance);
+
+	uprobe = tracefs_uprobe_alloc(UPROBE_GROUP_NAME, UPROBE_EVENT_NAME, file, offset, NULL);
+	if (!uprobe) {
+		pr_perror("Failed to allocate uprobe");
+		goto instance_destroy;
+	}
+
+	if (tracefs_dynevent_create(uprobe)) {
+		pr_perror("Failed to create uprobe");
+		goto uprobe_free;
+	}
+
+	if (tracefs_event_enable(trace_instance, UPROBE_GROUP_NAME, UPROBE_EVENT_NAME)) {
+		pr_perror("Failed to enable uprobe");
+		goto uprobe_destroy;
+	}
+
+	context.instance = trace_instance;
+	context.uprobe   = uprobe;
+	return context;
+
+uprobe_destroy:
+	tracefs_dynevent_destroy(uprobe, false);
+uprobe_free:
+	tracefs_dynevent_free(uprobe);
+instance_destroy:
+	tracefs_instance_destroy(trace_instance);
+	tracefs_instance_free(trace_instance);
+	return context;
+}
+
+void destroy_uprobe(struct uprobe_context context)
+{
+	tracefs_dynevent_destroy(context.uprobe, true);
+	tracefs_dynevent_free(context.uprobe);
+	tracefs_instance_destroy(context.instance);
+	tracefs_instance_free(context.instance);
+}
+
+/*
+ * Check for the existence of the "[uprobes]" vma in /proc/self/maps
+ * Returns -1 on failure, 0 if not found, 1 if found
+*/
+int uprobes_vma_exists(void)
+{
+	FILE *f;
+	char buf[LINE_MAX];
+	int ret = 0;
+
+	f = fopen("/proc/self/maps", "r");
+	if (!f) {
+		pr_perror("Failed to open /proc/self/maps");
+		return -1;
+	}
+
+	while (fgets(buf, sizeof(buf), f)) {
+		if (strstr(buf, "[uprobes]")) {
+			ret = 1;
+			break;
+		}
+	}
+	if (ret == 0 && !feof(f)) {
+		pr_err("Failed to finish reading /proc/self/maps\n");
+		ret = -1;
+	}
+
+	fclose(f);
+	return ret;
+}
+
+/*
+ * SIGTRAP is sent if execution reaches a previously set uprobed location, and
+ * the corresponding uprobe is not active. We don't want this to happen on restore
+*/
+void sigtrap_handler(int signo, siginfo_t *info, void* context)
+{
+	if (info->si_code == SI_KERNEL) {
+		got_sigtrap = true;
+		fail("SIGTRAP on attempting to call uprobed function");
+	}
+}
+
+int main(int argc, char **argv)
+{
+	struct uprobe_context context;
+	struct sigaction sa;
+	char buf[PATH_MAX];
+	uint64_t offset;
+	int n_bytes;
+	int ret = 1;
+
+	test_init(argc, argv);
+
+	offset = calc_sym_offset(__stringify(UPROBED_FUNCTION));
+	if (!offset)
+		return 1;
+
+	n_bytes = readlink("/proc/self/exe", buf, sizeof(buf));
+	if (n_bytes < 0) {
+		pr_perror("Failed to readlink /proc/self/exe");
+		return 1;
+	}
+	buf[n_bytes] = '\0';
+
+	sa.sa_flags = SA_SIGINFO;
+	sa.sa_sigaction = sigtrap_handler;
+	sigemptyset(&sa.sa_mask);
+	if (sigaction(SIGTRAP, &sa, NULL)) {
+		pr_perror("Failed to set SIGTRAP handler");
+		return 1;
+	}
+
+	context = enable_uprobe(buf, offset);
+	if (!context.instance)
+		return 1;
+
+	/*
+	 * Execution must reach the uprobed location at least once
+	 * for the kernel to create the uprobes vma
+	*/
+	uprobe_target_alias();
+
+	switch (uprobes_vma_exists()) {
+	case -1:
+		goto out_uprobe;
+		break;
+	case 0:
+		pr_err("uprobes vma does not exist\n");
+		goto out_uprobe;
+		break;
+	case 1:
+		test_msg("Found uprobes vma\n");
+		break;
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	/*
+	 * Calling the uprobed function after restore should not cause
+	 * a SIGTRAP, since the uprobe is still active
+	*/
+	uprobe_target_alias();
+	if (!got_sigtrap) {
+		pass();
+		ret = 0;
+	}
+
+out_uprobe:
+	destroy_uprobe(context);
+	return ret;
+}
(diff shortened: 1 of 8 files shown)


criu/proc_parse.c
```diff
@@ -603,6 +606,13 @@ static int handle_vma(pid_t pid, struct vma_area *vma_area, const char *file_pat
 			goto err;
 	} else if (!strcmp(file_path, "[heap]")) {
 		vma_area->e->status |= VMA_AREA_REGULAR | VMA_AREA_HEAP;
+	} else if (!strcmp(file_path, "[uprobes]")) {
+		if (!opts.allow_uprobes) {
+			pr_err("PID %d has uprobes vma. Consider using --" OPT_ALLOW_UPROBES ".\n",
```
rst0git, line 614: Would it make sense to use a boolean flag for this option in [inventory](https://github.com/checkpoint-restore/criu/blob/criu-dev/images/inventory.proto) instead asking users to explicitly specify it during restore?

shashank-mahadasyam: Do you mean something like how the `--tcp-close` option is handled?

```c
int check_img_inventory(bool restore)
{
    // ...
	if (restore && he->tcp_close && !opts.tcp_close) {
		pr_err("Need to set the --tcp-close options.\n");
		goto out_err;
	}
    // ...
}

// ...

int prepare_inventory(InventoryEntry *he)
{
    // ...
	/* tcp_close has to be set on restore if it has been set on dump. */
	if (opts.tcp_close) {
		he->tcp_close = true;
		he->has_tcp_close = true;
	}
    // ...
}
```

If so, then this does sound like a better idea. When a uprobes vma is detected, we can check if the `--allow-uprobes` option is specified, and if so, set this flag in the inventory image, and entirely skip storing any information about the uprobes vma (currently it shows up in the mm image). Then on restore, we can check early on if this flag is set in the inventory image and if the user has specified it on the command line or not. How does that sound?

rst0git: >Then on restore, we can check early on if this flag is set in the inventory image and if the user has specified it on the command line or not.

It might be enough to just set `opts.allow_uprobes = he->allow_uprobes`  during restore and print info message in the logs, similar to the following:

https://github.com/checkpoint-restore/criu/blob/5c2f02b2ea5e131deec105a48a3b6a16d4e46633/criu/config.c#L1077

shashank-mahadasyam: The reason I required the user to specify `--allow-uprobes` on restore is because if we restore while the uprobes (the ones which were active on checkpoint) are not active, then the kernel will send a SIGTRAP if execution reaches one of the uprobe'd locations. This is a surprising behaviour for the user.

So, using `--allow-uprobes` manually gives the user a chance to restore the required uprobes, if any. Essentially, it's a guarantee by the user that "the required uprobes are active on restore, or I'll deal with any SIGTRAPs". This is explained further in this commit: https://github.com/checkpoint-restore/criu/pull/2717/commits/5de06ac6ac9911dac46b06c9c698984af7e652bf

rst0git: It makes sense, thank you for clarifying.

shashank-mahadasyam: > If so, then this does sound like a better idea. When a uprobes vma is detected, we can check if the `--allow-uprobes` option is specified, and if so, set this flag in the inventory image, and entirely skip storing any information about the uprobes vma (currently it shows up in the mm image). Then on restore, we can check early on if this flag is set in the inventory image and if the user has specified it on the command line or not. How does that sound?

You gave me this idea. So, thank you, @rst0git  :)

It makes sense too, because there's no point in storing any information about the uprobes vma when we aren't even using it on restore.

Introducing a new (optional) field to the inventory image should be backwards compatible. Older inventory images will be treated as `allowed-uprobes = false`. I guess it's not forwards compatible because older criu can't handle newer inventory images with `allowed-uprobes = true`. Is this okay, @rst0git? If so, I'll force push the commits for this change

rst0git: Sounds good to me.

shashank-mahadasyam:
Hi, @rst0git! As we discussed in the previous comment, I've force pushed the changes to add a new optional boolean `allow_uprobes` to the inventory proto file. With these changes, we no longer have to store any information about the uprobes vma in the images. Just whether we encountered one or not, indicated by the `allow_uprobes` field.

Documentation/criu.txt
```diff
@@ -465,6 +465,36 @@ The 'mode' may be one of the following:
...
     must be locked externally to allow CRIU to dump TCP connections.
 
+*--allow-uprobes*::
+    Allow dumping when uprobes vma is present. When used on dump, this option is
+    required on restore as well.
+
+    The kernel creates a uprobes mapping in a process when execution hits a
+    uprobe. This is created in order to execute the uprobe'd instruction
+    out-of-line. Refer to commit d4b3b6384f98 ("uprobes/core: Allocate XOL slots
+    for uprobes use") of the Linux kernel for more details. This mapping is not
```
rst0git: @shashank-mahadasyam We usually don't include such references in the CRIU man page. Instead, we provide more detailed documentation and usage examples on the wiki pages: https://criu.org/CLI

criu/crtools.c
```diff
@@ -427,6 +427,8 @@ int main(int argc, char *argv[], char *envp[])
 	       "                        can be 'nftables' or 'iptables' (default).\n"
 	       "  --unprivileged        accept limitations when running as non-root\n"
 	       "                        consult documentation for further details\n"
+	       "  --allow-uprobes       allow dump/restore with uprobes vma\n"
+	       "                        consult documentation for further details\n"
```
rst0git: I think we should remove `"consult documentation for further details"` in  both places above. Most users already know that the documentation provides further details.

Documentation/criu.txt
shashank-mahadasyam: Updated documentation commit: https://github.com/checkpoint-restore/criu/commit/3d47ded7c6f543412372bf49a0a738ea39304b21

criu/crtools.c
shashank-mahadasyam: Added commit: https://github.com/checkpoint-restore/criu/pull/2717/commits/12b65160bccc9e2b0b67f80b22b4adbb88f13c64

shashank-mahadasyam:
I see that making the uprobes test depend on libtracefs, libtraceevent, and libelf is breaking CI. Just to confirm, where do I need to add these dependencies? `contrib/debian/dev-packages.lst`, `scripts/ci/vagrant.sh`, `scripts/ci/prepare-for-fedora-rawhide.sh`, and where else?

mihalicyn:
Dear @shashank-mahadasyam,

thanks for working on this!

I'm sorry for the delay with review, but I'll be able to look into this over the weekend. Pretty busy week. :(

Andrey (@avagin) is on vacation now and (almost) without a computer, that's why he hasn't looked yet.

>I see that making the uprobes test depend on libtracefs, libtraceevent, and libelf is breaking CI. Just to confirm, where do I need to add these dependencies? 

I guess almost everywhere. Because some of these files are used to build docker images to run tests from it, some are used to prepare a GitHub runner and install necessary dependencies.

adrianreber:
Take a look at https://github.com/checkpoint-restore/criu/commit/d165b94bb51a1ee34c7101898a8f99409366272f

rst0git:
@avagin I was wondering if you have any comments about the changes in this PR?

avagin:
LGTM. Thanks a lot for this contribution. We need to fix CI tests before merging it.

avagin:
@shashank-mahadasyam 

> LGTM. Thanks a lot for this contribution. We need to fix CI tests before merging it.

https://github.com/avagin/criu/commit/31e5d370d4f9f936f0bf2a16f7873290621d2f9a should fix most of ci failures.

shashank-mahadasyam:
I just rebased this branch onto the latest main branch. Sorry about the mess.

scripts/build/Dockerfile.hotspot-alpine
```diff
@@ -24,6 +24,7 @@ RUN apk update && apk add \
 
 COPY . /criu
 WORKDIR /criu
+RUN apk add $(sed 's/\#.*$//' contrib/apk-packages.txt)
```
rst0git: Would it be possible to move the patches adding package-manager specific lists in a separate PR?
It is not directly related to the changes for enabling uprobes support and it would be better to review and merge it separately.

>I know I should have sent these CI related commits as another PR, but due to certain internal constraints, I would really prefer it to go through this PR. I hope it isn't too annoying :)

We can merge it first and later just rebase this PR on top of the criu-dev branch.

test/zdtm/static/uprobes.c
```diff
@@ -0,0 +1,281 @@
...
+*/
+struct uprobe_context enable_uprobe(const char *file, uint64_t offset)
+{
+	struct tracefs_instance *trace_instance;
+	struct tracefs_dynevent *uprobe;
+	struct uprobe_context context = {};
+
+	trace_instance = tracefs_instance_create("zdtm_uprobes_test");
+	if (!trace_instance) {
+		pr_err("Failed to create tracefs instance\n");
```
rst0git: Could please replace `pr_err` with `pr_perror` here and in other places where the number of last error is helpful?

test/zdtm/static/uprobes.desc
```diff
@@ -0,0 +1,6 @@
+{
+	'feature': 'cgroupns',
+	'flags': 'suid nouser',
+	'flavor': 'h ns uns',
```
rst0git: Running this test locally, it only works for the `host` flavour. For `ns` and `uns` it fails with:
```
02:56:11.289:     4: ERR: uprobes.c:126: Failed to create tracefs instance (errno = 2 (No such file or directory))
```
I believe this is because ZDTM is currently not adding the mount points that are needed for libtracefs.

test/zdtm/static/uprobes.c
```diff
@@ -0,0 +1,281 @@
...
+	tracefs_instance_free(trace_instance);
+	return context;
+}
+
+void destroy_uprobe(struct uprobe_context context)
+{
+	tracefs_dynevent_destroy(context.uprobe, true);
+	tracefs_dynevent_free(context.uprobe);
+	tracefs_instance_destroy(context.instance);
+	tracefs_instance_free(context.instance);
```
rst0git, line 180: We currently run some of our CI tests on Ubuntu 22.04. It provides `libtracefs-dev` [version 1.2.5](https://packages.ubuntu.com/jammy/libtracefs-dev) which appears to be too old for this test to compile:

```
uprobes.c:129:9: error: implicit declaration of function ‘tracefs_instance_reset’; did you mean ‘tracefs_instance_create’? [-Werror=implicit-function-declaration]
  129 |         tracefs_instance_reset(trace_instance);
      |         ^~~~~~~~~~~~~~~~~~~~~~
      |         tracefs_instance_create
gcc -g -O2 -Wall -Werror -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=0 -Wdeclaration-after-statement -Wstrict-prototypes   -D_GNU_SOURCE -D_LARGEFILE64_SOURCE -I../lib -iquote ../lib/arch/x86/include -c write_read01.c -o write_read01.o
uprobes.c:131:18: error: implicit declaration of function ‘tracefs_uprobe_alloc’; did you mean ‘tracefs_instance_alloc’? [-Werror=implicit-function-declaration]
  131 |         uprobe = tracefs_uprobe_alloc(UPROBE_GROUP_NAME, UPROBE_EVENT_NAME, file, offset, NULL);
      |                  ^~~~~~~~~~~~~~~~~~~~
      |                  tracefs_instance_alloc
uprobes.c:131:16: error: assignment to ‘struct tracefs_dynevent *’ from ‘int’ makes pointer from integer without a cast [-Werror=int-conversion]
  131 |         uprobe = tracefs_uprobe_alloc(UPROBE_GROUP_NAME, UPROBE_EVENT_NAME, file, offset, NULL);
      |                ^
uprobes.c:137:13: error: implicit declaration of function ‘tracefs_dynevent_create’; did you mean ‘tracefs_instance_create’? [-Werror=implicit-function-declaration]
  137 |         if (tracefs_dynevent_create(uprobe)) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~
      |             tracefs_instance_create
uprobes.c:152:9: error: implicit declaration of function ‘tracefs_dynevent_destroy’; did you mean ‘tracefs_instance_destroy’? [-Werror=implicit-function-declaration]
  152 |         tracefs_dynevent_destroy(uprobe, false);
      |         ^~~~~~~~~~~~~~~~~~~~~~~~
      |         tracefs_instance_destroy
uprobes.c:154:9: error: implicit declaration of function ‘tracefs_dynevent_free’; did you mean ‘tracefs_list_free’? [-Werror=implicit-function-declaration]
  154 |         tracefs_dynevent_free(uprobe);
      |         ^~~~~~~~~~~~~~~~~~~~~
      |         tracefs_list_free
```

scripts/build/Dockerfile.hotspot-alpine
adrianreber: I also thought that this does not really belong in this PR.

test/zdtm/static/uprobes.c
shashank-mahadasyam: The man page doesn't specify that a `tracefs_instance_create` error is returned via errno: https://man7.org/linux/man-pages/man3/tracefs_instance_create.3.html

Even the implementation never explicitly sets errno: https://github.com/rostedt/libtracefs/blob/0a2a28f94489ad0d8a2ac5d70f617fd87296cd61/src/tracefs-instance.c#L189

test/zdtm/static/uprobes.desc
shashank-mahadasyam: I see how using `pr_perror` instead of `pr_err` is useful here even though libtracefs never explicitly sets it. I guess the errno set by the last failing libc syscall gives us at least some clue

test/zdtm/static/uprobes.c
shashank-mahadasyam: This affects the CircleCI clang/gcc, build coverage, and the criu image streamer CI pipelines. The Java test CI pipeline is using Ubuntu 20.04, which doesn't even have libtracefs in its repo:

```
#10 17.96 E: Unable to locate package libtraceevent-dev
#10 17.96 E: Unable to locate package libtracefs-dev
```

What do we do then? Is it possible to dump up the versions?

scripts/build/Dockerfile.hotspot-alpine
shashank-mahadasyam: Here's the PR with the CI-specific commits: https://github.com/checkpoint-restore/criu/pull/2748

test/zdtm/static/uprobes.desc
shashank-mahadasyam: Since uprobes is a kernel feature that's doesn't have any dependence on any namespace (it just needs tracefs to be mounted), it's enough if the test passes just for the host flavour, right?

rst0git: Yes, would suggest removing `ns` and `uns` for now. We can extend ZDTM later if this is needed.

test/zdtm/static/uprobes.c
rst0git: We can check the version of `libtracefs` and if it is too old we can skip this test.

We currently have only `pkg-config-check `: https://github.com/checkpoint-restore/criu/blob/dcd9df7389f7cebb2e50f70b87d478a17b889589/test/zdtm/lib/Makefile#L10

We can add something similar to check also for the version number.

test/zdtm/static/uprobes.c
```diff
@@ -0,0 +1,281 @@
...
+	 * for the kernel to create the uprobes vma
+	*/
+	uprobe_target_alias();
+
+	switch (uprobes_vma_exists()) {
+	case -1:
+		goto out_uprobe;
+		break;
+	case 0:
+		pr_err("uprobes vma does not exist\n");
```
rst0git, line 271:
```
2025-09-26T14:50:05.4155273Z ========================= Run zdtm/static/uprobes in h =========================
2025-09-26T14:50:05.4155605Z Start test
﻿2025-09-26T14:50:05.4158464Z Test is SUID
2025-09-26T14:50:05.4158695Z ./uprobes --pidfile=uprobes.pid --outfile=uprobes.out
2025-09-26T14:50:05.4158984Z make[1]: *** [Makefile:522: uprobes.pid] Error 1
2025-09-26T14:50:05.4159413Z  Test zdtm/static/uprobes FAIL at ['make', '--no-print-directory', '-C', 'zdtm/static', 'uprobes.pid'] 
2025-09-26T14:50:05.4159815Z Test output: ================================
2025-09-26T14:50:05.4160094Z 14:50:03.244:    42: ERR: uprobes.c:257: uprobes vma does not exist
2025-09-26T14:50:05.4160417Z 14:50:03.320:    41: ERR: test.c:320: Test exited unexpectedly with code 1
2025-09-26T14:50:05.4160634Z 
2025-09-26T14:50:05.4160701Z  <<< ================================
```
https://github.com/checkpoint-restore/criu/actions/runs/18041064442/job/51339787104?pr=2717

rst0git: If necessary, we can add `.checkskip` file to handle this:

Example:
https://github.com/checkpoint-restore/criu/blob/criu-dev/test/zdtm/static/selinux00.checkskip

shashank-mahadasyam: That's quite odd. I'm not sure why that happened. The kernel should have added the upobes mapping after the uprobed function having been called. Either way, I don't think there's anything we can do about it. How about `skip()` instead of`pr_err()` when uprobes vma is not found?

avagin: @shashank-mahadasyam It fails on arm64 instances. I found that it works with this patch:
```
diff --git a/test/zdtm/static/uprobes.c b/test/zdtm/static/uprobes.c
index 0e06a47b5..99c4e366c 100644
--- a/test/zdtm/static/uprobes.c
+++ b/test/zdtm/static/uprobes.c
@@ -19,7 +19,10 @@ const char *test_author = "Shashank Balaji <shashank.mahadasyam@sony.com>";
 #define UPROBE_EVENT_NAME      "uprobes_test"
 #define UPROBED_FUNCTION       uprobe_target
 
-void UPROBED_FUNCTION(void) {}
+void UPROBED_FUNCTION(void) __attribute__((noinline));
+void UPROBED_FUNCTION(void) {
+       test_msg("uprobe trigger\n");
+}
 /* Calling via volatile function pointer ensures noinline */
 typedef void (*func_ptr)(void);
 volatile func_ptr uprobe_target_alias = UPROBED_FUNCTION;
@@ -108,7 +111,8 @@ out_elf:
        elf_end(elf);
 out_fd:
        close(fd);
-       return offset;
+       test_msg("offset - %lx\n", offset);
+       return offset + 4;
 }
 
 /*
@@ -247,7 +251,7 @@ int main(int argc, char **argv)
         * Execution must reach the uprobed location at least once
         * for the kernel to create the uprobes vma
        */
-       uprobe_target_alias();
+       UPROBED_FUNCTION();
 
        switch (uprobes_vma_exists()) {
        case -1:
```

The key thing here is (offset + 4). I don't know why it doesn't work if uprobe is set on the first function instruction. Could you try to figure out?

shashank-mahadasyam: As it turns out, not all instructions trigger the creation of the uprobes vma. On aarch64, for example, the [`arm_probe_decode_insn`](https://elixir.bootlin.com/linux/v6.11/source/arch/arm64/kernel/probes/decode-insn.c#L76) decides whether an instruction requires the creation of the vma or not. For the empty function body of the uprobe target function, clang generates a single ret instruction (while gcc generates `nop; ret`), which doesn't need the creation of the vma. A nop on aarch64 will trigger the vma creation. While on x64, by empirical testing, I found out that `nop; ret` doesn't trigger the vma creation.

So to fix all this, the latest force push modifies the uprobe target function body with the creation of a dummy volatile variable and its use to prevent "unused variable" warning

shashank-mahadasyam:
Thanks for the feedback, @rst0git , @avagin!

PR #2728, opened 2025-09-03
cr-service: Refactor RPC config parsing

### LINKED ISSUE #2727: Parsing runc.conf results in duplicate values
When evaluating a configuration file specified via RPC (e.g., `runc.conf`), CRIU will parse this config file twice:

https://github.com/checkpoint-restore/criu/blob/5c2f02b2ea5e131deec105a48a3b6a16d4e46633/criu/cr-service.c#L332

https://github.com/checkpoint-restore/criu/blob/5c2f02b2ea5e131deec105a48a3b6a16d4e46633/criu/cr-service.c#L784

This causes options such as `--action-script` to accumulate duplicate values, leading to unexpected behavior (e.g., when used with [criu-coordinator](https://github.com/checkpoint-restore/criu-coordinator)).

rst0git:
When an additional configuration file is specified via RPC, this file is parsed twice: first at an early stage to load options such as `--log-file`, `--work-dir`, and `--images-dir`; and again after all RPC options and configuration files have been evaluated.

This allows users to overwrite options specified via RPC by the container runtime (e.g., `--tcp-established`). However, processing the RPC config file twice leads to silently duplicating the values of repeatable options such as `--action-script`.

To address this problem, we adjust the order of options parsing so that the RPC config file is evaluated only once. This does not introduce any functional changes.

Fixes #2727

### DIFF
--- a/test/others/rpc/config_file.py
+++ b/test/others/rpc/config_file.py
@@ -92,29 +92,37 @@ def test_broken_configuration_file():
         sys.exit(-1)
 
 
-def search_in_log_file(log, message):
-    with open(os.path.join(args['dir'], log)) as f:
+def search_in_log_file(log_path, message):
+    with open(log_path) as f:
         if message not in f.read():
-            print(
-                'FAIL: Missing the expected error message (%s) in the log file'
-                % message)
+            print('FAIL: Missing the expected error message (%s) in the log file' % message)
             sys.exit(-1)
+
+
+def print_log_file(log_path):
+    print("\n--- Begin log file: %s ---" % log_path)
+    with open(log_path, 'r') as f:
+        print(f.read())
+    print("--- End log file ---\n")
 
 
 def check_results(resp, log):
     # Check if the specified log file exists
-    if not os.path.isfile(os.path.join(args['dir'], log)):
+    log_path = os.path.join(args['dir'], log)
+    if not os.path.isfile(log_path):
         print('FAIL: Expected log file %s does not exist' % log)
         sys.exit(-1)
     # Dump should have failed with: 'The criu itself is within dumped tree'
     if resp.type != rpc.DUMP:
         print('FAIL: Unexpected msg type %r' % resp.type)
+        print_log_file(log_path)
         sys.exit(-1)
     if 'The criu itself is within dumped tree' not in resp.cr_errmsg:
         print('FAIL: Missing the expected error message in RPC response')
+        print_log_file(log_path)
         sys.exit(-1)
     # Look into the log file for the same message
-    search_in_log_file(log, 'The criu itself is within dumped tree')
+    search_in_log_file(log_path, 'The criu itself is within dumped tree')
 
 
 def test_rpc_without_configuration_file():


criu/config.c
```diff
@@ -592,6 +592,71 @@ static int parse_file_validation_method(struct cr_options *opts, const char *opt
...
+int early_parse_rpc_cfg_file(void)
+{
+	char **_argv = NULL;
+	int _argc = 0;
+	int opt, idx;
+	int rc = 0;
+	int saved_opterr;
+
+	static const char short_opts[] = "D:W:o:";
+	static struct option long_opts[] = {
```
avagin: Strictly speaking, you can't do that. If you don't know all the available options, you can't distinguish which are options and which are arguments.

avagin:
Could you explain why we need to parse it twice? Why do we need to handle log-file, workdir before parsing other options? Why we can't handle these option on the second run?

rst0git:
@avagin The only information I found are the following commit messages:

[cr-service: add support for configuration files in RPC mode](https://github.com/checkpoint-restore/criu/commit/99d52ec7dd4543010860bec50a187d93eaa72634)

>Most CRIU options are correctly used by just writing the new values to
the corresponding fields of the opts structure. For the RPC case there
are, however, a few options (output, work_dir, imgs_dir) which need
special handling.
> So the RPC configuration file is parsed twice. First time to get output,
work_dir and imgs_dir. Once those are read and correctly used, the RPC
code overwrites all options again by values set by the RPC interface. At
the end the RPC configuration file is read a second time and finally
overwrites the values set via RPC.

[Fix RPC configuration file handling](https://github.com/checkpoint-restore/criu/commit/b1c9e7faaf6e8874aafd241f404a1a685ecc3c0b)

> While writing runc test cases to verify that runc correctly uses RPC
configuration files it became clear that some things were not working as
they are supposed to. Looking closer at the code to set log files
via RPC configuration files I discovered that the code seems wrong (at
least I did not understand it any more (or the intentions behind it)).

@adrianreber Do you remember by any chance why we need to parse the RPC configuration file twice?

rst0git:
> I think we need to rework how these options are handled.

@avagin I've updated the pull request with these changes.

criu/cr-service.c
```diff
@@ -781,13 +623,147 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
...
 	if (req->config_file) {
+		char *tmp_output = opts.output;
+		char *tmp_work = opts.work_dir;
+		char *tmp_imgs = opts.imgs_dir;
+
+		opts.output = NULL;
+		opts.work_dir = NULL;
+		opts.imgs_dir = NULL;
+
+		pr_debug("Would overwrite RPC settings with values from %s\n", req->config_file);
```
adrianreber, line 643: Where does this line end up? Not in the configuration file set via RPC, so maybe drop that line or do that later.

adrianreber:
Concerning this line from the commit:

> This does not introduce any functional changes.

This is not true. Probably better to remove that line. Because you will be loosing early log messages with this change. Output from a couple of functions (`check_caps()`, `kerndat_init()` and more) will be written to another log file if the log file is changed via an additional configuration file. This means the user will now have to log files.

rst0git:
Thanks for the comments Adrian! 
>Because you will be loosing early log messages with this change.

Early log messages are still preserved with this change. We handle these in [log_init()](https://github.com/checkpoint-restore/criu/commit/1ff233356405a0a17a44bdab357b5fb54c2fb32b).

criu/cr-service.c
rst0git: The following is the CRIU logs output for `podman container checkpoint`:
```
(00.000000) Unable to get $HOME directory, local configuration file will not be used.
(00.000000) CRIU run id = 7ab73789-af78-4160-8e60-3acd276fd820
(00.000000) CRIU run id = a535288f-66d3-441b-86c9-02cb2bc6216c
(00.000000) Parsing config file /etc/criu/runc.conf
(00.000251) Version: 4.0 (gitid v4.0-168-g4a2a11258)
(00.000286) Running on thinkpad Linux 6.16.5-100.vanilla.fc41.x86_64 #1 SMP PREEMPT_DYNAMIC Thu Sep  4 15:39:11 UTC 2025 x86_64
(00.000290) Would overwrite RPC settings with values from /etc/criu/runc.conf
(00.000319) File /run/criu.kdat does not exist
(00.000981) sockets: Probing sock diag modules
```

adrianreber:
> Thanks for the comments Adrian!
> 
> > Because you will be loosing early log messages with this change.
> 
> Early log messages are still preserved with this change. We handle these in [log_init()](https://github.com/checkpoint-restore/criu/commit/1ff233356405a0a17a44bdab357b5fb54c2fb32b).

That is good to know. Maybe add that as a comment in the code or commit.

criu/cr-service.c
adrianreber: I would expect that all lines that end up in the early log buffer to have the "(00.000000)" time stamp. What is in the `runc.conf` you are using?

rst0git:
```
$ cat /etc/criu/runc.conf 
log-file=/tmp/criu.log
```

criu/cr-service.c
```diff
@@ -781,13 +623,147 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
...
+		/* If this is non-NULL, the RPC configuration file had a value, use it.*/
+		if (opts.output)
+			output_changed_by_rpc_conf = true;
+		/* If this is NULL, use the old value if it was set. */
+		if (!opts.output && tmp_output) {
+			opts.output = tmp_output;
+			tmp_output = NULL;
+		}
+
+		if (opts.work_dir)
```
avagin: nit: I think this version is a bit readable:
```
if (opts.work_dir) {
  work_changed_by_rpc_conf = true;
} else {
  opts.work_dir = tmp_work;
  tmp_work = NULL;
}
```

criu/cr-service.c
```diff
@@ -781,13 +623,147 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	if (req->has_display_stats)
 		opts.display_stats = req->display_stats;
 
-	/* Evaluate additional configuration file a second time to overwrite
-	 * all RPC settings. */
+	/* Evaluate additional configuration file (e.g., runc.conf) to overwrite all RPC settings. */
 	if (req->config_file) {
+		char *tmp_output = opts.output;
+		char *tmp_work = opts.work_dir;
+		char *tmp_imgs = opts.imgs_dir;
```
avagin: it seems we can do xfree(opts.imgs_dir) right here.

criu/cr-service.c
rst0git: Good point! Thanks!

criu/cr-service.c
rst0git: Yes, it seems like we don't actually need `tmp_imgs` here.

Issue #2738, opened 2025-09-11
Third predump fails with "Missing in parent pagemap" when compiled with musl


fntlnz:
When compiled with musl-libc criu is not able to build the parent chain correctly. This never happened to me when compiling against glibc.

The versions I identified as affected are:

- master
- v4.1.1
- v4.1
- the current version shipped in the testing repository of Alpine Linux. (I didn't test any other version). At the moment of writing it's v3.19

To reproduce:

```bash
docker run -it --privileged alpine:edge sh
```

Inside the Alpine container, enable testing repo, then install criu

```
echo "@testing https://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
apk update
apk add criu@testing python3
```

Now create this script, call it dumpchain.sh

```
set -xeuo pipefail
pid=$1
rm -Rf /root/dumps
mkdir -p /root/dumps

echo "DOING FIRST"
mkdir /root/dumps/1 && criu pre-dump -t $pid --track-mem  --images-dir /root/dumps/1 -vvv
echo "DOING SECOND"
mkdir /root/dumps/2 && criu pre-dump -t $pid --track-mem  --prev-images-dir /root/dumps/1 --images-dir /root/dumps/2 -vvv
echo "DOING THIRD"
mkdir /root/dumps/3 && criu pre-dump -t $pid --track-mem  --prev-images-dir /root/dumps/2 --images-dir /root/dumps/3 -vvv
echo "DOING FOURTH"
mkdir /root/dumps/4 && criu pre-dump -t $pid --track-mem  --prev-images-dir /root/dumps/3 --images-dir /root/dumps/4 -vvv
```

Now create a script to fill enough memory to trigger the behavior, let's call it `memfiller.py`. 

```
size_in_gb = 5
size_in_bytes = size_in_gb * (1024**3)

try:
    large_list = [0] * (size_in_bytes // 8)
    print("Successfully allocated.")

    while True:
        pass

except MemoryError:
    print("Memory allocation failed. Not enough memory available.")
except KeyboardInterrupt:
    print("Program interrupted by user.")
```

At this point we can put the pieces together:

- run memfiller `python3 memfiller.py`
- once it says "Successfully allocated" you can run `dumpchain.sh` passing the PID of the python process as first argument.

Now in the logs you should see an error like this:

```
(00.123096) page-xfer: Transferring pages:
(00.123098) page-xfer:  buf 1/1
[...]

PR #2739, opened 2025-09-11
fix: lookup for page hole in all ancestors

fntlnz:
Changed the check pagehole logic to go through the entire parent chain to find missing pages. More details on how to reproduce in the linked issue.

Fixes https://github.com/checkpoint-restore/criu/issues/2738


fntlnz:
@avagin thanks for taking a look, I suspected that something else might be wrong because it seemed strange that we would need to do that. I didn't realize it was supposed to read only the previous pagemap. I'll test and report back soon.

Issue #2033, opened 2023-01-04
criu 3.17.1 does not work on Manjaro Linux with kernel version 6.1.1-1


Ralf12358:
**Description**

criu produces the following error on Manjaro Linux with kernel:

criu version: 3.17.1
Kernerl version: Linux 6.1.1-1

```
$ criu check
Looks good.
```

```
(00.022866) No parent images directory provided
(00.022893) Error (criu/lsm.c:37): err scanfing: Invalid argument
(00.022989) Unlock network
(00.022998) Unfreezing tasks into 1
(00.023000)     Unseizing 2745391 into 1
(00.023005)     Unseizing 2752834 into 1
(00.023011) Error (criu/cr-dump.c:2053): Dumping FAILED.
```

criu/lsm.c:37
```C
static int apparmor_get_label(pid_t pid, char **profile_name)
{
	FILE *f;
	char *space;

	f = fopen_proc(pid, "attr/current");
	if (!f)
		return -1;

	if (fscanf(f, "%ms", profile_name) != 1) {
                // !!! Error message generated here. f has unexpected format. Put content of f into log file?
		pr_perror("err scanfing"); 
		fclose(f);
		return -1;
	}
	
// ...
```

adrianreber:
Do you have apparmor enabled on your system? What is the content of `/proc/self/attr/current`?

Ralf12358:
> Do you have apparmor enabled on your system? What is the content of `/proc/self/attr/current`?
```
$ aa-enabled 
Yes
```
apparmor is enabled

```
$ cat /proc/self/attr/current
cat: /proc/self/attr/current: Invalid argument
```
Ok. I see. Seems it is a general problem and not related to criu.

I found this: 
https://unix.stackexchange.com/questions/570545/why-does-sudo-cat-proc-self-attr-current-return-invalid-argument 
... but unfortunately with no solution. 

If you have an advice how to fix it, this would of course help, anyway.

Thanks for the quick response!

adrianreber:
@brauner any ideas why `current` would return invalid argument?

Ralf12358:
> @brauner any ideas why `current` would return invalid argument?

I think I found something. But have to try it:

https://github.com/docker/for-linux/issues/1199
https://wiki.archlinux.org/title/AppArmor#Installation

```
cat /sys/kernel/security/lsm
capability,landlock,lockdown,yama,bpf,apparmor
```

I will try this from the arch wiki:

[...]

PR #2741, opened 2025-09-13
lsm: use attr/apparmor/current to get apparmor label

filip-hejsek:
On some kernels, attr/current can be intercepted by BPF LSM, causing errors (#2033). Using attr/apparmor/current is preferable, because it is guaranteed to return the apparmor label. attr/current will still be used as a fallback for older kernels.

Fixes: #2033


rst0git (approved):
LGTM

PR #2743, opened 2025-09-15
zdtm: stop importing junit_xml

avagin:
We are dropping support for generating JUnit XML reports in zdtm.py as we've migrated testing infrastructure entirely to `GitHub Actions` and other third-party test runners.

This package has been removed from some distribution repositories (e.g., Fedora), making it simpler to remove the dependency than to force installation via pip.


rst0git (approved):
LGTM

I think this is a good idea. The upstream [python-junit-xml](https://github.com/kyrus/python-junit-xml) project has not been updated since 2020.

Issue #2745, opened 2025-09-16
Unsupported Elf format on mips


chenx97:
I tried to run `make` and was greeted with `Error (compel/arch/mips/src/lib/handle-elf-host.c:20): Unsupported Elf format detected`. A quick debug print shows me that the ELF magic in question is 7f 45 4c 46 02 01 01 00 05 00 00 00 00 00 00 00, while handle_binary expects 7f 45 4c 46 02 01 01 00 00 00 00 00 00 00 00 00. The problematic ELF magic is the legitimate ELF magic of a shared library on MIPS platforms. How should I make changes to criu to address it?

chenx97:
To make matters worse, normal `.o` files match the traditional ELF magic verified by `handle_binary`.

chenx97:
By digging into binutils-gdb, I found that `EI_ABIVERSION` is set to 5 on MIPS when
```C
  /* Mark that we need support for .MIPS.xhash in the dynamic linker,
     if it is the only hash section that will be created.  */
  if (link_info && link_info->emit_gnu_hash && !link_info->emit_hash)
    i_ehdrp->e_ident[EI_ABIVERSION] = MIPS_LIBC_ABI_XHASH;
```
in bfd/elfxx-mips.c. Running readelf successfully finds `.MIPS.xhash` sections in such shared objects.

PR #2746, opened 2025-09-17
fix: Relax ELF magic check to support MIPS libraries

DongSunchao:
On **MIPS** platforms, shared libraries may use **EI_ABIVERSION** = 5 to indicate
support for **.MIPS.xhash** sections. The previous ELF header check in
`handle_binary()` strictly compared e_ident against a hardcoded value,
causing legitimate shared objects to be rejected.

This patch replaces the memcmp-based check with a structured validation
of ELF magic and class, and allows **EI_ABIVERSION** values beside 0.
fixes: #2745

### DIFF
--- a/compel/arch/mips/src/lib/handle-elf.c
+++ b/compel/arch/mips/src/lib/handle-elf.c
@@ -5,18 +5,31 @@
 #include "piegen.h"
 #include "log.h"
 
-static const unsigned char __maybe_unused elf_ident_64_le[EI_NIDENT] = {
-	0x7f, 0x45, 0x4c, 0x46, 0x02, 0x01, 0x01, 0x00, /* clang-format */
-	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
-};
-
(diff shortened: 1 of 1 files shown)


compel/arch/mips/src/lib/handle-elf.c
```diff
@@ -5,18 +5,31 @@
...
-
 extern int __handle_elf(void *mem, size_t size);
 
 int handle_binary(void *mem, size_t size)
 {
-	if (memcmp(mem, elf_ident_64_le, sizeof(elf_ident_64_le)) == 0)
-		return __handle_elf(mem, size);
+	Elf64_Ehdr *ehdr = (Elf64_Ehdr *)mem;
+
+	//check ELF magic
```
avagin: nit: could you reformat comments to the commonly used format: /* check ELF magic */

chenx97:
The new file appears to be missing an LF at the end.

DongSunchao:
> The new file appears to be missing an LF at the end.

Thanks, I have fixed it.

PR #2748, opened 2025-09-18
Easier CI Package Addition

shashank-mahadasyam:
To test the `--allow-uprobes` introduces by https://github.com/checkpoint-restore/criu/pull/2717 requires the addition of a few packages to be downloaded in the CI pipeline. The current way of doing this requires adding the package names to multiple Dockerfiles and scripts, which are not exhaustively listed, as far as I can see. This way, some files may end up being missed. As discussed in  https://github.com/checkpoint-restore/criu/pull/2717#discussion_r2357376276, this PR makes the Dockerfiles and scripts pull in common package names from package-manager specific files, making it easier to add new packages.

### DIFF
--- a/scripts/build/Dockerfile.archlinux
+++ b/scripts/build/Dockerfile.archlinux
@@ -5,6 +5,8 @@ ARG CC=gcc
 # Initialize machine ID
 RUN systemd-machine-id-setup
 
+COPY contrib/dependencies/pacman-packages.txt /tmp/pacman-packages.txt
+
 RUN pacman -Syu --noconfirm \
 	$CC \
 	bash \
@@ -35,7 +37,10 @@ RUN pacman -Syu --noconfirm \
 	python-importlib-metadata \
 	libdrm \
 	util-linux-libs \
-	diffutils
+	diffutils \
+	$(sed 's/\#.*$//' /tmp/pacman-packages.txt) \
+	&& \
+	rm /tmp/pacman-packages.txt
 
 COPY . /criu
 WORKDIR /criu
--- a/scripts/build/Dockerfile.archlinux
+++ b/scripts/build/Dockerfile.archlinux
@@ -9,35 +9,6 @@ COPY contrib/dependencies/pacman-packages.txt /tmp/pacman-packages.txt
 
 RUN pacman -Syu --noconfirm \
 	$CC \
-	bash \
-	make \
-	coreutils \
-	git \
-	gnutls \
-	libaio \
-	libcap \
-	libnet \
-	libnl \
-	nftables \
-	pkgconfig \
-	protobuf-c \
-	protobuf \
-	python-pip \
-	python-protobuf \
-	which \
-	sudo \
-	iptables \
-	nftables \
-	iproute2 \
-	tar \
-	bash \
-	go \
-	python-yaml \
-	asciidoctor \
-	python-importlib-metadata \
-	libdrm \
-	util-linux-libs \
-	diffutils \
 	$(sed 's/\#.*$//' /tmp/pacman-packages.txt) \
 	&& \
 	rm /tmp/pacman-packages.txt
--- a/scripts/build/Dockerfile.hotspot-alpine
+++ b/scripts/build/Dockerfile.hotspot-alpine
@@ -3,28 +3,7 @@ ARG CC=gcc
 
 COPY contrib/dependencies/apk-packages.txt /tmp/apk-packages.txt
 
-RUN apk add --no-cache \
-	bash \
-	build-base \
-	coreutils \
-	git \
-	gnutls-dev \
-	libaio-dev \
-	libcap-dev \
-	libnet-dev \
-	libnl3-dev \
-	pkgconfig \
-	protobuf-c-dev \
-	protobuf-dev \
-	python3 \
-	sudo \
-	maven \
-	ip6tables \
-	iptables \
-	util-linux-dev \
-	bash \
-	$(sed 's/\#.*$//' /tmp/apk-packages.txt) \
-	&& \
+RUN apk add --no-cache $(sed 's/\#.*$//' /tmp/apk-packages.txt) && \
 	rm /tmp/apk-packages.txt
 
 COPY . /criu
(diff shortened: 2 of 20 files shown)


adrianreber (approved):
Looks correct. Thanks.

scripts/build/Dockerfile.archlinux
```diff
@@ -5,40 +5,11 @@ ARG CC=gcc
...
-	asciidoctor \
-	python-importlib-metadata \
-	libdrm \
-	util-linux-libs \
-	diffutils
+RUN pacman -Syu --noconfirm $CC
 
 COPY . /criu
 WORKDIR /criu
+RUN pacman -Syu --noconfirm $(sed 's/\#.*$//' contrib/pacman-packages.txt)
```
rst0git: Would it be possible to create subdirectory called "dependencies" and add "pacman.txt", "apk.txt", "apt.txt" inside?

scripts/build/Dockerfile.hotspot-alpine
```diff
@@ -1,29 +1,11 @@
...
-	protobuf-c-dev \
-	protobuf-dev \
-	python3 \
-	sudo \
-	maven \
-	ip6tables \
-	iptables \
-	util-linux-dev \
-	bash
+RUN apk update
```
rst0git: Every `RUN` command creates a new layer in the container image. It might be better to use `RUN apk update && apk add $(sed 's/\#.*$//' contrib/apk-packages.txt)`

scripts/build/Dockerfile.hotspot-ubuntu
```diff
@@ -4,29 +4,22 @@ ARG CC=gcc
...
 	git \
 	pkg-config \
 	iptables \
 	gcc \
-	uuid-dev \
 	maven
 
 COPY . /criu
 WORKDIR /criu
+RUN apt-install $(sed 's/\#.*$//' contrib/apt-packages.txt)
```
rst0git: Most of the packages above (e.g., `protobuf-c-compiler`, `protobuf-compiler`, `libbsd-dev`, `make`, `pkg-config`, `gcc`, and `iptables`) are also required for CRIU. Should we include them as well?

scripts/build/Dockerfile.linux32.tmpl
```diff
@@ -3,29 +3,22 @@ ARG CC=gcc
...
 	libselinux-dev \
 	pkg-config \
 	protobuf-c-compiler \
 	protobuf-compiler \
-	uuid-dev \
 	python3-minimal
 
 COPY . /criu
 WORKDIR /criu
+RUN apt-install $(sed 's/\#.*$//' contrib/apt-packages.txt)
```
rst0git: Same comment here. It would be good to include all required dependencies in the TXT file. The following page provides more information: https://criu.org/Installation

rst0git:
@shashank-mahadasyam What do you think about about using shell scripts instead of `.txt` files? For example, these shell scripts will be useful also when testing CRIU manually in a new environment to install the required build dependencies.

scripts/build/Dockerfile.hotspot-ubuntu
shashank-mahadasyam: Sure, I can do that. The reason some of the obvious ones are not there in `apt-packages.txt` is because some Dockerfiles/scripts aren't installing those packages. I didn't think too much when populating these package list files. I just saw the common packages across all the Dockerfiles/scripts for each package manager and put the common ones in the package list files.

For example, `libbsd-dev` is used in `Dockerfile.openj9-ubuntu`, `Dockerfile.hotspot-ubuntu`, and `run-ci-tests.sh`, but not in `Dockerfile.[un]stable-cross.tmpl`, and `Dockerfile.linux32.tmpl`

scripts/build/Dockerfile.hotspot-alpine
shashank-mahadasyam: That's a good point. I'll combine the update and installation steps in all the Dockerfiles wherever possible

scripts/build/Dockerfile.hotspot-ubuntu
rst0git: This is because different people have introduced the Dockerfile files at different time. When consolidating the list of packages it would be good, if possible, to make them consistent.

shashank-mahadasyam: Then I'll add the packages mentioned on the criu installation page (https://criu.org/Installation) to all the package list files

scripts/build/Dockerfile.riscv64-stable-cross.tmpl
```diff
@@ -17,29 +17,18 @@ RUN dpkg --add-architecture ${DEBIAN_ARCH} && \
...
-	libnet-dev:${DEBIAN_ARCH}		\
-	libprotobuf-c-dev:${DEBIAN_ARCH}	\
-	libcap-dev:${DEBIAN_ARCH}		\
-	libaio-dev:${DEBIAN_ARCH}		\
-	uuid-dev:${DEBIAN_ARCH}			\
-	libnl-route-3-dev:${DEBIAN_ARCH}	\
-	libnftables-dev:${DEBIAN_ARCH}	\
-	libgnutls28-dev:${DEBIAN_ARCH}	\
-	iproute2:${DEBIAN_ARCH}	\
 	$(awk '!/^#/ && NF {print $0 ":'"${DEBIAN_ARCH}"'"}' /tmp/apt-packages.txt) \
```
avagin, line 32: apt-packages.txt contains protobuf-c-compiler, doesn't it? Do we really need to install protobuf-c-compiler:${DEBIAN_ARCH} package?

shashank-mahadasyam: Oops, I overlooked that. I'll filter out the host-only packages from `apt-packages.txt`

shashank-mahadasyam:
Closing this because https://github.com/checkpoint-restore/criu/pull/2755 is more complete

shashank-mahadasyam:
Closing this because https://github.com/checkpoint-restore/criu/pull/2755 is more complete

PR #2750, opened 2025-09-19
pagemap: change PagemapEntry.nr_pages to uint64 to support huge mappings

avagin:
Variables storing page counts were previously `unsigned int`, limiting
them to a maximum of 2^32 pages. With a 4k page size, this corresponds
to a 16TB memory mapping, which is insufficient for larger mappings.

This commit changes the type for these variables to `unsigned long` to
support larger memory mappings.


rst0git:
@avagin There is one `unsigned int nr_pages` in `xfer_pages()` which is used as an argument for `uffd_handle_pages()` and `pr.read_pages()`. Should we change this to `unsigned long` as well?

PR #2755, opened 2025-09-24
ci: use package-manager dependency install scripts

rst0git:
Currently, adding a package which is required either for development or testing requires it to be added in multiple places due to many duplicated Dockerfiles and installation scripts. This makes it difficult to ensure that all scripts are updated appropriately and can lead to some places being missed.

This patch consolidates the list of dependencies and adds installation scripts for each package-manager used in our CI (apk, apt, dnf, yum, pacman).

This change also replaces the `debian/dev-packages.lst` as this subfolder conflicts with the Ubuntu/Debian packing scripts used for CRIU (https://github.com/rst0git/criu-deb-packages)


scripts/build/Dockerfile.alpine
```diff
@@ -1,49 +1,12 @@
...
-	tar \
-	bash \
-	go \
-	e2fsprogs \
-	py-yaml \
-	py3-importlib-metadata \
-	asciidoctor
+RUN /criu/contrib/dependencies/apk-packages.sh
+
+RUN make mrproper && date && make -j $(nproc) CC="$CC" && date
```
shashank-mahadasyam, line 9: It's recommended to have static RUN command near the start of the Dockerfile. That ensures, any build cache invalidation by later COPY/ADD commands do not invalidate the RUN command ([reference](https://circleci.com/blog/tips-for-optimizing-docker-builds/)). In our case, the `COPY . /criu` will almost always trigger a cache invalidation

contrib/dependencies/apt-cross-packages.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/sh
+
+apt-install \
+	crossbuild-essential-"${DEBIAN_ARCH}" \
+	iproute2:"${DEBIAN_ARCH}" \
+	libaio-dev:"${DEBIAN_ARCH}" \
+	libbz2-dev:"${DEBIAN_ARCH}" \
+	libc6-"${DEBIAN_ARCH}"-cross \
```
shashank-mahadasyam: Now that we have scripts instead of just text files, `apt-packages.sh` can be modified to be used in both normal and cross-compilation targets:
- Two lists of packages: host-only (build-essential, pkg-config, etc.), and cross- versions
- Install the cross- versions with the arch prefix (empty if not set for normal targets), and the host-only packages as they are

contrib/dependencies/apk-packages.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/sh
```
shashank-mahadasyam: The rest of them are `#!/usr/bin/sh`. Any particular reason why this is just `#!/bin/sh`?

contrib/dependencies/yum-packages.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/bash
```
shashank-mahadasyam: The rest of them are using sh in their shebang. Any particular reason for the difference?

contrib/dependencies/apt-packages.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/sh
+
+apt-install \
+	"$CC" \
```
shashank-mahadasyam, line 4: It would be easier to reason about if this is used where it's defined, i.e. in Dockerfiles, and not rely on the implicit connection "this script will install whatever `$CC` the caller sets"

contrib/dependencies/pacman-packages.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/sh
+
+pacman -Syu --noconfirm \
+	"$CC" \
```
shashank-mahadasyam, line 4: Same comment about "use where defined"

scripts/build/Dockerfile.fedora.tmpl
```diff
@@ -1,11 +1,11 @@
 ARG CC=gcc
 
-COPY scripts/ci/prepare-for-fedora-rawhide.sh /bin/prepare-for-fedora-rawhide.sh
-RUN /bin/prepare-for-fedora-rawhide.sh
-
 COPY . /criu
 WORKDIR /criu
 
+RUN contrib/dependencies/dnf-packages.sh && \
+	scripts/ci/prepare-for-fedora-rawhide.sh
```
shashank-mahadasyam: `scripts/ci/prepare-for-fedora-rawhide.sh` calls `contrib/dependencies/dnf-packages.sh` anyway

scripts/ci/run-ci-tests.sh
```diff
@@ -60,6 +55,7 @@ ci_prep () {
 		CI_PKGS+=("${X86_64_PKGS[@]}")
 	fi
 
+	PATH="$PATH:scripts/ci/" contrib/dependencies/apt-packages.sh
```
shashank-mahadasyam: Why is the PATH assignment required?

.cirrus.yml
```diff
@@ -34,7 +34,11 @@ task:
   setup_script: |
     dnf config-manager --set-enabled crb # Same as CentOS 8 powertools
     dnf -y install epel-release epel-next-release
-    dnf -y install --allowerasing asciidoc gcc git gnutls-devel libaio-devel libasan libcap-devel libnet-devel libnl3-devel libbsd-devel libselinux-devel make protobuf-c-devel protobuf-devel python-devel python-PyYAML python-protobuf python3-importlib-metadata xmlto libdrm-devel libuuid-devel
+    dnf install -y --allowerasing \
+      asciidoc \
+      python-devel \
+      xmlto
+    bash contrib/dependencies/dnf-packages.sh
```
shashank-mahadasyam: Why mention the interpreter when it's mentioned in the shebang ;)

contrib/debian/dev-packages.lst
shashank-mahadasyam, line 1: `scripts/install-debian-pkgs.sh` references this file, so that script should be removed as well

scripts/build/Dockerfile.alpine
rst0git: I agree with you but in our case we don't actually use the build cache. Each Dockerfile is built in a clean CI environment from scratch.

scripts/ci/run-ci-tests.sh
rst0git: We use `COPY scripts/ci/apt-install /bin/apt-install` to do something similar in the Dockerfiles. Here we just update `PATH`.  Without this assignment `apt-install` would not be available.

scripts/build/Dockerfile.fedora.tmpl
rst0git: Thanks!

contrib/dependencies/apk-packages.sh
rst0git: `/usr/bin/sh` does not exist in the Alpine container image:
```
$ sudo podman run alpine:latest ls /usr/bin/sh
ls: /usr/bin/sh: No such file or directory

$ sudo podman run alpine:latest ls /bin/sh
/bin/sh
```

contrib/dependencies/yum-packages.sh
rst0git: This script is actually not needed as CentOS 8 is EOL.

contrib/dependencies/apt-cross-packages.sh
rst0git: The cross-compilation tests are simpler and require fewer dependencies. Keeping these dependencies as a separate list makes is it easier to maintain.

contrib/dependencies/apk-packages.sh
shashank-mahadasyam: `#!/usr/bin/env sh` would take care of that ;)

scripts/build/Dockerfile.hotspot-alpine
```diff
@@ -1,30 +1,11 @@
...
-	maven \
-	ip6tables \
-	iptables \
-	util-linux-dev \
-	bash
-
 COPY . /criu
 WORKDIR /criu
 
+RUN apk add maven "$CC" && contrib/dependencies/apk-packages.sh
```
shashank-mahadasyam: Missing `--no-cache`

shashank-mahadasyam (approved):

shashank-mahadasyam:
Closing this because https://github.com/checkpoint-restore/criu/pull/2755 is more complete

adrianreber:
Removing CentOS 8 makes sense as it is EOL, but maybe you should introduce tests on one of the EL_8 clones (AlmaLinux for example).

shashank-mahadasyam:
It would be nice to mention `contrib/` in `CONTRIBUTING.md` for installing dependencies as well as for adding new ones for CI. Maybe in https://criu.org/Installation#Dependencies as well

rst0git:
> Removing CentOS 8 makes sense as it is EOL, but maybe you should introduce tests on one of the EL_8 clones (AlmaLinux for example).

Thanks Adrian! I will create another pull request for the tests with AlmaLinux.

Issue #2756, opened 2025-09-24
armv7 cross compilation fails with uint64 nr_pages


rst0git:
I just noticed the following errors related to https://github.com/checkpoint-restore/criu/pull/2750:

```
#11 12.51 criu/page-xfer.c: In function 'page_server_get_pages':
#11 12.51 criu/page-xfer.c:1148:62: error: passing argument 4 of 'page_pipe_read' from incompatible pointer type [-Wincompatible-pointer-types]
#11 12.51  1148 |         ret = page_pipe_read(pp, &pipe_read_dest, pi->vaddr, &pi->nr_pages, PPB_LAZY);
#11 12.51       |                                                              ^~~~~~~~~~~~~
#11 12.51       |                                                              |
#11 12.51       |                                                              u64 * {aka long long unsigned int *}
#11 12.51 In file included from criu/page-xfer.c:20:
#11 12.51 criu/include/page-pipe.h:152:112: note: expected 'long unsigned int *' but argument is of type 'u64 *' {aka 'long long unsigned int *'}
#11 12.51   152 | extern int page_pipe_read(struct page_pipe *pp, struct pipe_read_dest *prd, unsigned long addr, unsigned long *nr_pages,
#11 12.51       |                                                                                                 ~~~~~~~~~~~~~~~^~~~~~~~
#11 12.52 At top level:
#11 12.52 cc1: note: unrecognized command-line option '-Wno-unknown-warning-option' may have been intended to silence earlier diagnostics
#11 12.52 make[2]: *** [/criu/scripts/nmk/scripts/build.mk:119: criu/page-xfer.o] Error 1
#11 12.52 make[2]: *** Waiting for unfinished jobs....
#11 13.22 make[1]: *** [criu/Makefile:76: criu/built-in.o] Error 2
#11 13.22 make: *** [Makefile:278: criu] Error 2
#11 ERROR: process "/bin/sh -c make mrproper && date && make -j $(nproc) zdtm && date" did not complete successfully: exit code: 2
------
 > [7/7] RUN make mrproper && date && make -j $(nproc) zdtm && date:
12.51 In file included from criu/page-xfer.c:20:
12.51 criu/include/page-pipe.h:152:112: note: expected 'long unsigned int *' but argument is of type 'u64 *' {aka 'long long unsigned int *'}
12.51   152 | extern int page_pipe_read(struct page_pipe *pp, struct pipe_read_dest *prd, unsigned long addr, unsigned long *nr_pages,
[...]

PR #2757, opened 2025-09-24
page-xfer: fix incompatible pointer type on armv7

rst0git:
`page_pipe_read()` expects an `unsigned long *`, but `pi->nr_pages` is `u64`. On 32-bit platforms (e.g., armv7), passing `&pi->nr_pages` directly causes a compiler error. To fix this we introduce a temporary variable and copy the result back to `pi->nr_pages`.

Fixes: #2756

### DIFF
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -1139,13 +1139,17 @@ static int page_server_get_pages(int sk, struct page_server_iov *pi)
 {
 	struct pstree_item *item;
 	struct page_pipe *pp;
-	unsigned long len;
+	unsigned long len, nr_pages;
 	int ret;
 
 	item = pstree_item_by_virt(pi->dst_id);
 	pp = dmpi(item)->mem_pp;
 
-	ret = page_pipe_read(pp, &pipe_read_dest, pi->vaddr, &pi->nr_pages, PPB_LAZY);
+	/* page_pipe_read() uses 'unsigned long *' but pi->nr_pages is u64.
+	 * Use a temporary variable to fix the incompatible pointer type
+	 * on 32-bit platforms (e.g. armv7). */
+	nr_pages = pi->nr_pages;
+	ret = page_pipe_read(pp, &pipe_read_dest, pi->vaddr, &nr_pages, PPB_LAZY);
 	if (ret)
 		return ret;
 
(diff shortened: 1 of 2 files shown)


criu/page-xfer.c
```diff
@@ -32,7 +32,7 @@ static int page_server_sk = -1;
 
 struct page_server_iov {
 	u32 cmd;
-	u64 nr_pages;
```
avagin, line 35: is page_server_iov sent between criu and page-server? If the answer is yes, it is better to use fixed-size types.

rst0git: > is page_server_iov sent between criu and page-server?

Yes, `struct page_server_iov` is used to send and receive messages between the CRIU process and the page-server.  I will update the patch to use fixed-size types instead.

criu/include/page-pipe.h
```diff
@@ -93,8 +93,8 @@ struct page_pipe_buf {
 	int p[2];		/* pipe with pages */
 	unsigned int pipe_size; /* how many pages can be fit into pipe */
 	unsigned int nr_segs;	/* how many iov-s are busy */
-	unsigned long pipe_off;	/* where this buf is started in a pipe */
-	unsigned long pages_in;	/* how many pages are there */
+	uint64_t pipe_off;	/* where this buf is started in a pipe */
```
avagin: I mean a bit different thing in my previous comment. The native pipe can't contain more that LONG_MAX pages, so I am not sure that we need to do this change.  When I wrote the previous comment, I was thinking about something like this:
```
@@ -1139,13 +1139,14 @@ static int page_server_get_pages(int sk, struct page_server_iov *pi)
 {
        struct pstree_item *item;
        struct page_pipe *pp;
-       unsigned long len;
+       unsigned long len, nr_pages;
        int ret;
 
        item = pstree_item_by_virt(pi->dst_id);
        pp = dmpi(item)->mem_pp;
 
-       ret = page_pipe_read(pp, &pipe_read_dest, pi->vaddr, &pi->nr_pages, PPB_LAZY);
+       nr_pages = pi->nr_pages;
+       ret = page_pipe_read(pp, &pipe_read_dest, pi->vaddr, &nr_pages, PPB_LAZY);
        if (ret)
                return ret;
 
@@ -1154,6 +1155,7 @@ static int page_server_get_pages(int sk, struct page_server_iov *pi)
         * .dst_id all remain intact.
         */
 
+       pi->nr_pages = nr_pages;
        if (pi->nr_pages == 0) {
                pr_debug("no iovs found, zero pages\n");
                return -1;
```

PR #2759, opened 2025-09-25
fix: add missing opts merge in pycriu.criu.check() request

herheliuk:
lib/pycriu/criu.py: fix missing opts merge in check()

The check() RPC request requires opts.images_dir_fd to be set.
However, since self.opts was never merged into req.opts, CRIU
saw no options at all and returned an "Unknown error" instead
of a proper failure.

Fix this by adding the missing req.opts.MergeFrom(self.opts),
so criu.check() now reports sane error codes when opts are
invalid or incomplete.


herheliuk:
sorry for all the force pushes, this is my first contribution. (now just one file is changed, as it was initially)

rst0git:
`--track-mem` and `opts.track_mem = True` don't do what you think.

As mentioned above, [`FEATURE_CHECK`](https://github.com/checkpoint-restore/criu/blob/b39a9c6da161fd969ede7d1c96867ceac7ee0df3/criu/cr-service.c#L1150) requests are used for this purpose.
Example:
> sudo criu check --feature mem_dirty_track

herheliuk:
I wanted to rename the branch. (didn't go well)

PR #2761, opened 2025-09-26
ci/java: update base image from focal to jammy

rst0git:
Ubuntu Focal Fossa (20.04) reached its end-of-life on 31 May 2025. So, move over to using Ubuntu Jammy (22.04) base images.
    
Also, focal repos do not have libtracefs, which is needed for the uprobes zdtm test in https://github.com/checkpoint-restore/criu/pull/2717.


shashank-mahadasyam:
It would be nice to get rid of the other Travis CI references as well while we're at it: https://github.com/search?q=repo%3Acheckpoint-restore%2Fcriu%20travis&type=code

PR #2706, opened 2025-08-10
CI: Consolidate arm64 tests on GitHub runners

avagin:
The arm64 tests are currently being executed on both actuated and GitHub runners. This change removes the actuated runner to avoid redundancy and streamline our CI process.

Thanks to @alexellis  and the actuated team for providing arm64 runners before they were available natively on GitHub.

Fixes #2704

PR #2763, opened 2025-09-27
ci: consolidate aarch64 tests on GitHub runners

rst0git:
Currently we run aarch64 tests on both Cirrus CI and GitHub runners. However, Cirrus CI fails with `"Monthly compute limit exceeded!"`. This change removes the redundant tests to streamline our CI process.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -89,28 +89,6 @@ task:
     make -C scripts/ci vagrant-fedora-non-root
 
 task:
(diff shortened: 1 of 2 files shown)


adrianreber:
Sounds good. The aarch64 CI systems were from a time when GitHub didn't have aarch64 systems.

rst0git:
@avagin The patch in this pull request is very similar to https://github.com/checkpoint-restore/criu/pull/2706.

PR #2767, opened 2025-09-30
criu/compel: fix compilation on musl/arm64

peppergrayxyz:
Compilation on gentoo/arm64 (llvm+musl) fails with:

```
In file included from compel/include/uapi/compel/asm/sigframe.h:4,
                 from compel/plugins/std/infect.c:14:
/usr/include/asm/sigcontext.h:28:8: error: redefinition of 'struct sigcontext'
   28 | struct sigcontext {
      |        ^~~~~~~~~~

In file included from criu/arch/aarch64/include/asm/restorer.h:4,
                 from criu/arch/aarch64/crtools.c:11:
/usr/include/asm/sigcontext.h:28:8: error: redefinition of 'struct sigcontext'
   28 | struct sigcontext {
      |        ^~~~~~~~~~

```
This is happening because <asm/sigcontext.h> and <signal.h> are mutually incompatible on Linux.
To fix, use  <signal.h> instead of <asm/sigcontext.h> for arm64 (like all others arches do).

Issue #2766, opened 2025-09-30
[ARM64] <asm/sigcontext.h> and <signal.h> are mutually incompatible on Linux


peppergrayxyz:
I'm running into the same issue as described here https://github.com/checkpoint-restore/criu/issues/934 and here [sourceware#24813](https://sourceware.org/bugzilla/show_bug.cgi?id=24813):

```
  CC       compel/plugins/std/infect.o
In file included from compel/include/uapi/compel/asm/sigframe.h:4,
                 from compel/plugins/std/infect.c:14:
/usr/include/asm/sigcontext.h:28:8: error: redefinition of 'struct sigcontext'
   28 | struct sigcontext {
      |        ^~~~~~~~~~
In file included from /usr/include/signal.h:48,
                 from compel/include/uapi/compel/plugins/std/syscall-types.h:12,
                 from compel/include/uapi/compel/plugins/std/syscall.h:4,
                 from compel/include/uapi/compel/plugins/std.h:5,
                 from compel/plugins/std/infect.c:1:
/usr/include/bits/signal.h:18:16: note: originally defined here
   18 | typedef struct sigcontext {
      |                ^~~~~~~~~~
/usr/include/asm/sigcontext.h:70:8: error: redefinition of 'struct _aarch64_ctx'
   70 | struct _aarch64_ctx {
      |        ^~~~~~~~~~~~
/usr/include/bits/signal.h:29:8: note: originally defined here
   29 | struct _aarch64_ctx {
      |        ^~~~~~~~~~~~
/usr/include/asm/sigcontext.h:77:8: error: redefinition of 'struct fpsimd_context'
   77 | struct fpsimd_context {
      |        ^~~~~~~~~~~~~~
/usr/include/bits/signal.h:33:8: note: originally defined here
   33 | struct fpsimd_context {
      |        ^~~~~~~~~~~~~~
/usr/include/asm/sigcontext.h:96:8: error: redefinition of 'struct esr_context'
   96 | struct esr_context {
      |        ^~~~~~~~~~~
/usr/include/bits/signal.h:39:8: note: originally defined here
   39 | struct esr_context {
      |        ^~~~~~~~~~~
/usr/include/asm/sigcontext.h:136:8: error: redefinition of 'struct extra_context'
  136 | struct extra_context {
      |        ^~~~~~~~~~~~~
/usr/include/bits/signal.h:43:8: note: originally defined here
   43 | struct extra_context {
      |        ^~~~~~~~~~~~~
[...]

PR #2770, opened 2025-10-02
ci: add alpine on arm test

kolyshkin:
Inspired by #2766 / #2767.


kolyshkin:
I think I've reproduced the issue reported in #2766. Will try the fix now.

PR #2742, opened 2025-09-15
files: fork helpers without CLONE_FILES | CLONE_FS

avagin:
On restore, CRIU needs to change mount namespaces to properly restore files and unix sockets. However, the kernel prevents this if a process is sharing its file system information (fs) with other processes.

Fixes #2687

PR #2771, opened 2025-10-03
zdtm: add sk-unix-restore-fs-share test

rst0git:
This pull request adds a ZDTM test case based on the patch in https://github.com/checkpoint-restore/criu/pull/2742#issuecomment-3327840816

### DIFF
--- a/test/zdtm/static/sk-unix-restore-fs-share.c
+++ b/test/zdtm/static/sk-unix-restore-fs-share.c
@@ -1,0 +1,196 @@
+#include <stdio.h>
+#include <string.h>
+#include <signal.h>
+#include <unistd.h>
+#include <sys/wait.h>
+#include <sys/mman.h>
+#include <sys/socket.h>
+#include <sys/un.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Test non-empty process group with terminated parent and unix socket";
+const char *test_author = "Qiao Ma <mqaio@linux.alibaba.com>";
+
+char *filename;
+TEST_OPTION(filename, string, "socket file name", 1);
+
+static int create_and_connect(void)
+{
+	struct sockaddr_un addr;
+	int client_fd;
+
+	client_fd = socket(AF_UNIX, SOCK_STREAM, 0);
+	if (client_fd == -1) {
+		pr_perror("socket");
+		return -1;
+	}
+
+	memset(&addr, 0, sizeof(addr));
+	addr.sun_family = AF_UNIX;
+	if (snprintf(addr.sun_path, sizeof(addr.sun_path), "%s", filename) >= (int)sizeof(addr.sun_path)) {
+		pr_err("Socket path too long\n");
+		close(client_fd);
+		return -1;
+	}
+
+	if (connect(client_fd, (struct sockaddr *)&addr, sizeof(addr)) == -1) {
+		pr_perror("connect");
+		close(client_fd);
+		return -1;
+	}
+
+	return 0;
+}
+
+static int child(int ready_fd)
+{
+	int listen_fd;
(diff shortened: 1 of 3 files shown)


test/zdtm/static/sk-unix-restore-fs-share.c
```diff
@@ -0,0 +1,185 @@
...
+	pid = fork();
+	if (pid < 0) {
+		pr_perror("Failed to fork zombie");
+		goto out;
+	}
+
+	if (pid == 0)
+		exit(zombie_leader(cpid));
+
+	if (waitpid(pid, &status, 0) < 0) {
```
avagin, line 168: nit: I think it should be better to check status.

test/zdtm/static/sk-unix-restore-fs-share.c
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// Signal parent that socket is ready
+	if (write(ready_fd, "1", 1) != 1) {
+		pr_perror("write ready_fd");
+		close(listen_fd);
+		return EXIT_FAILURE;
+	}
+
+	create_and_connect();
```
avagin: should it be before the previous write? Otherwise, it is undetermined whether it will be called before of after C/R, isn't it?

test/zdtm/static/sk-unix-restore-fs-share.c
```diff
@@ -0,0 +1,185 @@
...
+		return EXIT_FAILURE;
+	}
+
+	if (listen(listen_fd, 5) == -1) {
+		pr_perror("listen");
+		close(listen_fd);
+		return EXIT_FAILURE;
+	}
+
+	// Signal parent that socket is ready
```
avagin: nit: /* .... */

test/zdtm/static/sk-unix-restore-fs-share.c
```diff
@@ -0,0 +1,185 @@
...
+
+	ret = EXIT_SUCCESS;
+out:
+	if (*cpid)
+		kill(*cpid, SIGKILL);
+
+	munmap(cpid, sizeof(int));
+
+	if (ret == EXIT_SUCCESS)
+		pass();
```
avagin: nit: I think we can move pass() where we set ret to EXIT_SUCCESS.

test/zdtm/static/sk-unix-restore-fs-share.desc
```diff
@@ -0,0 +1 @@
+{'flavor': 'ns uns', 'flags': 'suid'}
```
avagin: why suid is required here?

rst0git: `suid` is not needed here.

test/zdtm/static/sk-unix-restore-fs-share.c
rst0git: Yes, you are right.

test/zdtm/static/sk-unix-restore-fs-share.c
```diff
@@ -0,0 +1,196 @@
...
+		return EXIT_FAILURE;
+	}
+
+	memset(&addr, 0, sizeof(addr));
+	addr.sun_family = AF_UNIX;
+	if (strlen(filename) >= sizeof(addr.sun_path)) {
+		pr_err("Socket path too long\n");
+		goto cleanup;
+	}
+	strncpy(addr.sun_path, filename, sizeof(addr.sun_path) - 1);
```
avagin: nit: -1 isn't required here. addr is on stack and it hasn't been zeroed out.

avagin:
LGTM. Thanks!

PR #2772, opened 2025-10-04
ci: consolidate action-script tests

rst0git:
This patch consolidates the action-script tests into `test/others/action-script` to ensure all tests are executed consistently and reduce duplication.


test/others/action-script/check_actions.py
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env python3
```
avagin, line 1: This is test/check_actions.py, isn't it? Should it be marked as moved?

rst0git: I've updated the test to run ZDTM only with `ns` flavour and modified `check_actions.py` to check that all action-script hooks are called in the correct order.

The test previously collected all action-script hooks in `actions_called.txt` for `h`, `ns`, and `uns`, then checked that each hook appeared at least once in the txt file.

Would it be better if I separate into two commits the change of (1) moving `check_actions.py` and (2) updating its content?

avagin: pls update the commit message with all these information. 

> Would it be better if I separate into two commits the change of (1) moving check_actions.py and (2) updating its content?

yes, this sounds like the right way to do this change.

rst0git: Thanks! I've updated the pull request with this change.

PR #2774, opened 2025-10-06
cirrus-ci: add tests for almalinux 8

rst0git:
The CI tests for CentOS 8 have been removed as it is EOL. This pull request adds AlmaLinux task in Cirrus-CI as a replacement for these tests to build and test CRIU in RHEL 8–compatible environment.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -49,6 +49,29 @@ task:
     make -C scripts/ci local SKIP_CI_PREP=1 CC=gcc CD_TO_TOP=1 ZDTM_OPTS="-x zdtm/static/socket-raw"
 
 task:
+  name: AlmaLinux 8
+  environment:
+    HOME: "/root"
+    CIRRUS_WORKING_DIR: "/tmp/criu"
+
+  compute_engine_instance:
+    image_project: almalinux-cloud
+    image: family/almalinux-8
+    platform: linux
+    cpu: 4
+    memory: 8G
+
+  setup_script: |
+    dnf install -y https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm dnf-plugins-core
+    dnf config-manager --set-enabled powertools
+    contrib/dependencies/dnf-packages.sh
+    setenforce 0
+
+  build_script: |
+    make -j $(nproc) CC=gcc
+    make -C scripts/ci/ local CC=gcc SKIP_CI_PREP=1 CD_TO_TOP=1 ZDTM_OPTS="-x zdtm/static/socket-raw"
+
+task:
   name: Vagrant Fedora Rawhide
   environment:
     HOME: "/root"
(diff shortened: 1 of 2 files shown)


contrib/dependencies/yum-packages.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env sh
+
+yum install -y --allowerasing \
+	asciidoc \
+	chkconfig \
+	coreutils \
```
shashank-mahadasyam: Could you please add `libtraceevent-devel`, `libtracefs-devel`, and `elfutils-libelf-devel`? They're needed for the uprobes test merged in https://github.com/checkpoint-restore/criu/commit/08fa6c369a82b9945867c9f06fb1a44ff7a023c0

shashank-mahadasyam: Maybe a separate yum packages file is not required after all: https://almalinux.pkgs.org/8/almalinux-powertools-x86_64/libtracefs-devel-1.3.1-3.el8.x86_64.rpm.html#:~:text=repo.almalinux.org-,Install%20Howto,-Install%20libtracefs%2Ddevel

contrib/dependencies/yum-packages.sh
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env sh
+
+yum install -y --allowerasing \
```
adrianreber: Switching to dnf would make sense. Then also the file should be renamed. So maybe something for another PR.

adrianreber (commented):
Looks. The only thing which could be changed is s/yum/dnf/ but not necessary for this PR.

contrib/dependencies/yum-packages.sh
rst0git: Thanks! The list of packages in the existing `dnf-packages.sh` script appears to be compatible with AlmaLinux 8, so I went ahead and used it instead of creating a separate yum packages file.

adrianreber:
> @rst0git rhel8 reached the eol. I don't think that we need to continue testing it and all its derivatives.

CentOS Stream 8 is EOL. RHEL 8 and clones are available and supported at least until 2029.

rst0git:
> > @rst0git rhel8 reached the eol. I don't think that we need to continue testing it and all its derivatives.
> 
> CentOS Stream 8 is EOL. RHEL 8 and clones are available and supported at least until 2029.

https://access.redhat.com/support/policy/updates/errata#RHEL8_Planning_Guide

test/zdtm/lib/pidfd.h
```diff
@@ -0,0 +1,26 @@
+#ifndef __NR_pidfd_open
+#define __NR_pidfd_open 434
```
avagin, line 2: __NR_pidfd_open is 5434 on mips

avagin:
@rst0git yeah, this sort of vulnerabilities is one in decade.... btw this case was caught by other jobs and rhel8 jobs had no effect how it was fixed.

As I said, I don't like the idea of supporting ancient kernels. If we really want to run these tests, we can create a separate branch with the rhel8 criu version that will run tests on Almalinux once a day.

PR #2777, opened 2025-10-09
libcriu: enable setting of RPC config file

rst0git:
Container runtimes that use libcriu (e.g., crun) need to specify a CRIU configuration file that allows to overwrite default options set via RPC. This is particularly useful to set options such as `--tcp-established` via `/etc/criu/runc.conf` in Kubernetes.

### DIFF
--- a/test/others/libcriu/.gitignore
+++ b/test/others/libcriu/.gitignore
@@ -8,3 +8,4 @@ test_pre_dump
 test_feature_check
 output/
 libcriu.so.*
+test_rpc_config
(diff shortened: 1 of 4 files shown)


lib/c/criu.c
```diff
@@ -2041,3 +2041,19 @@ void criu_set_empty_ns(int namespaces)
 {
 	criu_local_set_empty_ns(global_opts, namespaces);
 }
+
+int criu_local_set_config_file(criu_opts *opts, const char *path)
+{
+	if (path) {
```
avagin: nit: braces are not needed for a single statement body

rst0git: Thanks! I've updated the patch with this change. I also noticed that we need to call `free()` before assigning a new `strdup()` value to avoid memory leaks if `criu_set_config_file()` is called more than once.

PR #2783, opened 2025-10-17
pycriu: add missing protobuf dependency

herheliuk:
pycriu depends on protobuf to function correctly. Currently,
it raises an error if protobuf is not installed. Adding
protobuf to the dependencies ensures it is available after
installing pycriu.


herheliuk:
#2790, Sorry didn't do it in the first place

PR #2784, opened 2025-10-18
pycriu: prevent trailing "Unknown" in error messages

herheliuk:
Regardless of the actual error message, "Unknown" was always appended
to the end of the string, resulting in messages like:
"DUMP failed: Error(3): No process with such pid**Unknown**".

Fixed by changing standalone if statements to else-if blocks so
"Unknown" is only added when no specific error condition matches.


herheliuk:
if you don't like elifs:
```python
if not self.errno:
    s += "Unknown"
```
instead of:
```python
s += "Unknown"
```

rst0git (approved):

Issue #2758, opened 2025-09-25
pycriu `check` always errors


herheliuk:
Trying to use pycriu:

```
sudo criu service &
```

```python
from pycriu import criu as pycriu_criu

criu = pycriu_criu()

criu.use_sk('criu_service.socket')

criu.check()
```

But I get the following error:

```
pycriu.criu.CRIUExceptionExternal: CHECK failed: Error(0): Unknown
```

Output from `criu service`:

```
(13.259281) Connected.
(13.259432) Waiting for connection...
(13.259533) CRIU run id = 7ae68e84-7a16-4271-bb4b-fe017c9ff8e1
(13.313752) Worker(pid 1791) exited with 0
```

I tested this on an `Ubuntu Server` virtual machines for both `x86_64` and `aarch64`.

adrianreber:
As a reference look at https://github.com/checkpoint-restore/criu/tree/criu-dev/test/others/rpc

There is is used during test runs. Maybe that helps.

herheliuk:
I found an error in criu.py!
in a check function. will commit asap.

herheliuk:
New Python code:

```python
from pycriu import criu as _criu
from os import open as os_open, O_DIRECTORY

criu = _criu()

criu.use_sk('criu_service.socket')

criu.opts.images_dir_fd = os_open('dumps', O_DIRECTORY)

criu.check()
```

herheliuk:
idk, this is just how it works ATM... (with req.opts.MergeFrom(self.opts))

herheliuk:
actually, merging opts messes up images_dir_fd for some reason

even though it seems to be there:
```python
            print(req.opts.images_dir_fd)
            print(req.opts.HasField('images_dir_fd'))
            s.send(req.SerializeToString())
```

```bash
-1
False
Traceback (most recent call last):
  File "/home/user/criu_api/env/lib/python3.12/site-packages/pycriu/criu.py", line 250, in _send_req_and_recv_resp
(100.07187) Connected.
    s.send(req.SerializeToString())
           ^^^^^^^^^^^^^^^^^^^^^^^
google.protobuf.message.EncodeError: Message criu_req is missing required fields: opts.images_dir_fd
```

herheliuk:
https://github.com/checkpoint-restore/criu/blob/57251d8115ce58f14717d90b095901b6a1c50ea8/criu/cr-service.c#L696-L699

need to skip this else if req.type is CHECK

[...]

PR #2789, opened 2025-10-20
cr-service: refactor RPC opts parsing for check()

rst0git:
The check() functionality is very different from `dump`, `pre-dump`, and `restore`. It is used only to check if the kernel supports required features, and does not need the majority of options set via RPC.

This patch updates the RPC options parser so that it only handles the logging options when `check()` is used. In particular, logging to a file is required when `log_file` is explicitly set or no `log_to_stderr` is used. In such case, we also resolve `images_dir` and `work_dir` where the log file will be created.

Fixes: #2758


criu/cr-service.c
```diff
@@ -885,11 +885,7 @@ static int check(int sk, CriuOpts *req)
 
 	if (pid == 0) {
 		__setproctitle("check --rpc");
-
 		opts.mode = CR_CHECK;
-		if (setup_opts_from_req(sk, req))
```
avagin, line 890: We have check specific options like --feature, --extra, --experimental...

herheliuk: `cr-check.c` 1637: `int cr_check(void)`

`opts.check_extra_features`
`opts.check_experimental_features`

> but

```python
AttributeError: Protocol message criu_opts has no "check_extra_features" field.
```

avagin: I have not idea what it means... Please, be more detailed. In my comment, I am asking how we are going to handle these check specific options without calling `setup_opts_from_req`.

herheliuk: You're talking about `check_extra_features` and `check_experimental_features` right ?

They doesn't seem to be supported by `criu_opts` (rpc) protocol.

rst0git: @avagin The `--feature` option is implemented as a separate message type `FEATURE_CHECK` in [rpc.proto](https://github.com/checkpoint-restore/criu/blob/criu-dev/images/rpc.proto) and handled in [`handle_feature_check()`](https://github.com/checkpoint-restore/criu/blob/d25d829bde2fb44bfe5410c96e7909265022b81b/criu/cr-service.c#L1150C12-L1150C32).

`--extra` and `--experimental` are not currently exposed via RPC. I'm not sure if anyone needs these options via RPC, as most container runtimes use only the `FEATURE_CHECK` functionality (e.g., [runc](https://github.com/opencontainers/runc/blob/a7651e0ee0642bb8c1f9cb5ded484a5a7fdc9f99/libcontainer/criu_linux.go#L36) and [crun](https://github.com/containers/crun/blob/64611d7ac938b8397e8a00a0e69987583fadec7d/src/libcrun/criu.c#L254)).

avagin: > require images_dir_fd or images_dir to be specified.

Could you show in the code where we require them to be specified? I think we should not do that and it will be a proper fix for the issue.

herheliuk: @avagin no offence, but it is obvious that `dump` `pre-dump` and `restore` require `images_dir_fd` or `images_dir` to be specified.

herheliuk: @avagin The point of this PR is to fix `pycriu` and `criulib` `check()` function, since it's not supposed to use opts at all.

You can see this in `criu --help`, and as he mentioned, there are `feature_check()` in `libcriu` for this.

avagin: > @avagin The point of this PR is to fix `pycriu` and `criulib` `check()` function, since it's not supposed to use opts at all.

Who said that? There are a check specific option, so it is supposed to use opts with this command...

herheliuk: There is `feature_check` function for that.

You're not supposed to do `criu check --track-mem`, only `criu check --all`, `--extra` etc.

It does work this way in CLI, but it is unintended behaviour.

`criu --help` supports that point.

@avagin

rst0git: > > require images_dir_fd or images_dir to be specified.
> 
> Could you show in the code where we require them to be specified? I think we should not do that and it will be a proper fix for the issue.

Here:
https://github.com/checkpoint-restore/criu/blob/57251d8115ce58f14717d90b095901b6a1c50ea8/criu/cr-service.c#L697

I believe the intention is for `images_dir_fd` or `images_dir` to be required when `dump`, `pre-dump`, and `restore` are used via RPC.

avagin: > It does work this way in CLI, but it is unintended behaviour.

I always appreciate it when someone feels the need to explain my intentions back to me;).

herheliuk: rst0git has convinced me that I'm wrong. https://github.com/checkpoint-restore/criu/pull/2759#issuecomment-3419134934
Otherwise I was happily using `criu check --track-mem` 🙄

avagin: criu_opts contains properties like log_file, log_level and they are relevant for the check command.

herheliuk:
Allegedly, the 'Co-developed-by' tags in b64a3623bf7ec4b8ffa85825e8d98cc633b8402c and c2e1a88c05a32cf41aae8cd8d0e0689e3066d1bf also require my sign-off, unlike 'Co-authored-by'.

lib/pycriu/criu.py
```diff
@@ -204,6 +204,8 @@ class criu:
     def __init__(self):
         self.use_binary('criu')
         self.opts = rpc.criu_opts()
+        # req.SerializeToString requires images_dir_fd to be set
+        self.opts.images_dir_fd = -1
```
herheliuk: in case we do ever change -1 to something else, and to avoid hard coding:

```suggestion
        self.opts.images_dir_fd = self.opts.images_dir_fd
```

lib/pycriu/criu.py
```diff
@@ -204,6 +204,8 @@ class criu:
     def __init__(self):
         self.use_binary('criu')
         self.opts = rpc.criu_opts()
+        # req.SerializeToString requires images_dir_fd to be set
```
herheliuk:
```suggestion
        # Since images_dir_fd is a required field in rpc.proto, req.SerializeToString requires images_dir_fd to be explicitly set
```

lib/pycriu/criu.py
rst0git: I'm not sure if this makes sense - it assigns the attribute to itself.
`-1` is the default value for `images_dir_fd`

rst0git: >in case we do ever change -1 to something else

It is very unlikely that we are going to do that. If we do, we are going to update this code.

herheliuk: > `-1` is the default value for `images_dir_fd`

exactly. The default value is already `-1` and it will show `-1` if you print `self.opts.images_dir_fd`.

But if you print `self.opts` they will be empty. It is counter intuitive, but that's how protobuf works.

If you assign the value to itself and print `self.opts` (or merged `req.opts`) you'll be actually able to see it.

We need to do this, because `images_dir_fd` is marked as **required** in `rpc.proto`,
so we **must explicitly assign it** at all times.

herheliuk: > If we do, we are going to update this code.

It's always better not to hard code anything (because of a technical debt).

rst0git: I appreciate your comments but it is very unlikely for us to change the default value of this field. Readability is more important here - there is no way to know what `self.opts.images_dir_fd = self.opts.images_dir_fd` does just by reading the code.

rst0git:
@herheliuk No worries, I was just trying to give you credit for working on this.

criu/cr-service.c
```diff
@@ -285,6 +285,31 @@ int exec_rpc_query_external_files(char *name, int sk)
 
 static char images_dir[PATH_MAX];
 
+static int resolve_images_dir_path(char images_dir_path[PATH_MAX],
```
avagin: I think for the compiler it will be just a char pointer. There is no guarantee about a size of images_dir_path.

criu/cr-service.c
```diff
@@ -285,6 +285,31 @@ int exec_rpc_query_external_files(char *name, int sk)
...
+				   bool imgs_changed_by_rpc_conf,
+				   const CriuOpts *req,
+				   pid_t peer_pid)
+{
+	if (imgs_changed_by_rpc_conf) {
+		strncpy(images_dir_path, opts.imgs_dir, PATH_MAX - 1);
+		images_dir_path[PATH_MAX - 1] = '\0';
+	} else if (req->images_dir_fd != -1) {
+		if (snprintf(images_dir_path, PATH_MAX, "/proc/%d/fd/%d",
+			     peer_pid, req->images_dir_fd) >= PATH_MAX) {
```
avagin: this is actually impossible and I don't think that we need this check here....

criu/cr-service.c
```diff
@@ -285,6 +285,31 @@ int exec_rpc_query_external_files(char *name, int sk)
...
+			pr_err("images_dir path truncated\n");
+			return -1;
+		}
+	} else if (req->images_dir) {
+		strncpy(images_dir_path, req->images_dir, PATH_MAX - 1);
```
avagin, line 311: it is probably more valuable to check return code of strncpy then the previous snprintf.

criu/cr-service.c
```diff
@@ -299,8 +299,8 @@ static int setup_images_and_workdir(const char *images_dir_path,
 	}
 
 	/* get full path to images_dir to use in process title */
-	if (readlink(images_dir_path, images_dir, PATH_MAX) == -1) {
-		pr_perror("Can't readlink %s", images_dir_path);
+	if (!realpath(images_dir_path, images_dir)) {
```
avagin: I already commented that. I think we need to delete this buggy code and stop using this path to set proctitle.

criu/cr-service.c
```diff
@@ -895,6 +918,70 @@ static int restore_using_req(int sk, CriuOpts *req)
...
 }
 
+/*
+ * Lightweight parser for options used by the CHECK workflow.
+ * Only parse log_file, log_to_stderr, and log_level.
+ *
+ * If we will write a log file (explicit log_file or default log file),
+ * resolve images_dir and work_dir the same way as setup_opts_from_req().
+ */
+static int setup_check_opts_from_req(int sk, CriuOpts *req)
```
avagin: I don't think that we need a second parser...

criu/cr-service.c
rst0git: Sorry, I missed your earlier comment.

If we stop using `setproctitle()` it also means that we can remove the dependency for `libbsd-devel`. It might be better to do this in another pull request.

criu/cr-service.c
rst0git: I noticed that we don't currently check the return value of `strncpy` anywhere in the CRIU code, so I left this as it is for now. We can address this in a separate patch or pull request.

criu/cr-service.c
```diff
@@ -397,6 +411,42 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
...
 
+	/*
+	 * Parse only the options relevant to the CHECK workflow:
+	 * log_file, log_to_stderr, and log_level.
+	 *
+	 * If logging to a file, either explicitly specified with log_file
+	 * or when using the DEFAULT_LOG_FILENAME, also resolve the
+	 * images_dir and work_dir.
+	 */
+	if (opts.mode == CR_CHECK) {
```
avagin, line 418: yeah. Last time, when I said that we didn't need the separate parser for CR_CHECK, I mean the current parser should work for this command and any other commands. Maybe I don't understand something, but the commit message doesn't make it clear to me either. I know that some option are not applicable to the check command and users should not user them, but there is nothing wrong still processing them. As for image_dir, we need to open it only when it is specified. It is ok to check the `check` command in resolve_images_dir_path.

avagin:
please don't merge unrelated patches to one pull request. It is much easier to review smaller pr-s

criu/cr-service.c
rst0git: >As for image_dir, we need to open it only when it is specified.

The idea is that we don't need to open `image_dir` when running `check()` because this functionality doesn't create or process image files. In this case, `image_dir` is used as `work_dir`, only to create a log file.

rst0git: > I know that some option are not applicable to the check command and users should not use them, but there is nothing wrong still processing them.

I was trying to avoid the error messages that would appear when validating these options in [`check_options()`](https://github.com/checkpoint-restore/criu/blob/ae971c9fe069e3d24e461740ce936080bc7b6812/criu/config.c#L1076): e.g., https://github.com/checkpoint-restore/criu/pull/2759#issuecomment-3418805753

PR #2793, opened 2025-10-21
fix-tests: prevent pip from installing third-party dependencies 

herheliuk:
fix-tests: prevent pip from installing third-party dependencies 

Added `--no-deps` to the pip arguments since all required dependencies
(such as `python3-protobuf`) have already been installed system-wide.


rst0git:
@herheliuk Would it be okay with you if we merge https://github.com/checkpoint-restore/criu/pull/2792 instead?

PR #2795, opened 2025-10-22
lib/pycriu: changing the default behavior to use the system binary

herheliuk:
Use system-installed CRIU binary instead of a local file

Thanks to @avagin for suggesting this solution.

Co-authored-by: Andrei Vagin <avagin@gmail.com>

### DIFF
--- a/lib/pycriu/criu.py
+++ b/lib/pycriu/criu.py
@@ -103,7 +103,7 @@ class _criu_comm_bin(_criu_comm):
                 os.close(2)
 
                 css[0].send(struct.pack('i', os.getpid()))
-                os.execv(self.comm,
+                os.execvp(self.comm,
                          [self.comm, 'swrk',
                           "%d" % css[0].fileno()])
                 os._exit(1)


herheliuk:
Currently, pycriu attempts to use a local CRIU binary, but I believe it would make more sense to use a system-installed CRIU binary instead, as that would provide a more “works out of the box” experience.

lib/pycriu/criu.py
```diff
@@ -202,7 +203,7 @@ class criu:
     _comm = None  #Communication method
 
     def __init__(self):
-        self.use_binary('criu')
+        self.use_binary(shutil_which('criu'))
```
avagin: this is probably the better fix:
```
diff --git a/lib/pycriu/criu.py b/lib/pycriu/criu.py
index 5bd7ffecd..5973b4b91 100644
--- a/lib/pycriu/criu.py
+++ b/lib/pycriu/criu.py
@@ -103,7 +103,7 @@ class _criu_comm_bin(_criu_comm):
                 os.close(2)
 
                 css[0].send(struct.pack('i', os.getpid()))
-                os.execv(self.comm,
+                os.execvp(self.comm,
                          [self.comm, 'swrk',
                           "%d" % css[0].fileno()])
                 os._exit(1)
```

rst0git:
> Currently, pycriu attempts to use a local CRIU binary, but I believe it would make more sense to use a system-installed CRIU binary instead.

Note that this behavior is for testing. We want to run tests with a locally built binary, not a system-installed one.

herheliuk:
@rst0git won't `os.execvp` resolve to local binary if it is available? tests seem to be passing...

herheliuk:
Tests might have been failing because I accidentally pushed my fork instead of origin rebase.
So, maybe we don't need to add the binary to PATH in tests.

You know better, but I don't think there would be a criu installed system-wide on the CI anyway.

PR #2796, opened 2025-10-23
cr-service: fix path resolution for `images_dir`

herheliuk:
The old code always used `readlink()` to resolve `images_dir_path`, which wasn’t correct for normal directories, so if you’d ever try to use `images_dir` instead of `images_dir_fd` it would always fail there. 

Also, `readlink()` doesn’t add a `null terminator`, so now we explicitly add `\0` at the end to avoid potential string issues or overflows.

Fixes: 9089ce8 ("service: use setproctitle")

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -714,9 +714,22 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	}
 
 	/* get full path to images_dir to use in process title */
-	if (readlink(images_dir_path, images_dir, PATH_MAX) == -1) {
-		pr_perror("Can't readlink %s", images_dir_path);
-		goto err;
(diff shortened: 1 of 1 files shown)


avagin:
Should we just remove this global image_dir and stop passing it to setproctitle? This code looks buggy and seems never to work in real environments.

avagin:
@herheliuk please avoid closing a pull request if you're actively working on it. If it's not ready for review yet, you can simply mark it as a draft. To update the PR with new changes, you can use a force push to the branch.

If you need to reorder or squash changes, consider using `git rebase -i origin/criu-dev` before force pushing.

herheliuk:
#2789 Copies this PR.

PR #2797, opened 2025-10-23
rpc.proto: making `images_dir_fd` optional

herheliuk:
Currently, using `images_dir` requires explicitly setting `opts.images_dir_fd` to `-1` for **every
request**, because Protobuf treats the field as required, even though its default value is already `-1`.

This makes the field unnecessarily mandatory.

### DIFF
--- a/images/rpc.proto
+++ b/images/rpc.proto
@@ -61,7 +61,7 @@ enum criu_pre_dump_mode {
 };
 
 message criu_opts {
-	required int32			images_dir_fd	= 1 [default = -1];
+	optional int32			images_dir_fd	= 1 [default = -1];
 	optional string			images_dir	= 68; /* used only if images_dir_fd == -1 */
 	optional int32			pid		= 2; /* if not set on dump, will dump requesting process */
 
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -674,7 +674,7 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	}
 
 	/*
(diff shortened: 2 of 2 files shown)


herheliuk:
current example of `images_dir` usage:

```python
from pycriu import criu as pycriu_criu

criu = pycriu_criu()

criu.opts.images_dir_fd = -1
criu.opts.images_dir = dumps_dir
criu.opts.pid = int(input('pid: '))
criu.opts.shell_job = True
criu.dump()
```

if we don't set `criu.opts.images_dir_fd` to `-1`:

```python
google.protobuf.message.EncodeError: Message criu_req is missing required fields: opts.images_dir_fd
```

images/rpc.proto
```diff
@@ -61,7 +61,7 @@ enum criu_pre_dump_mode {
 };
 
 message criu_opts {
-	required int32			images_dir_fd	= 1 [default = -1];
+	optional int32			images_dir_fd	= 1 [default = -1];
```
rst0git, line 64: This change is likely to break existing users:
https://protobuf.dev/programming-guides/proto2/#required-deprecated

herheliuk:
tests have failed.

images/rpc.proto
herheliuk: we can hardcode `req.opts.images_dir_fd = self.opts.images_dir_fd` on every request, (in pycriu)
but it won't be that pretty.

I'm down for any solution, just to fix this inconvenience (and realistically barrier for newbies).

PR #2798, opened 2025-10-23
pycriu: making `images_dir_fd` optional

herheliuk:
Since `images_dir_fd` is a required filed in `rpc.proto` we must
explicitly pass it into `opts` in order to use `images_dir`.

Co-authored-by: Andrei Vagin <avagin@gmail.com>

### DIFF
--- a/lib/pycriu/criu.py
+++ b/lib/pycriu/criu.py
@@ -204,6 +204,9 @@ class criu:
     def __init__(self):
         self.use_binary('criu')
         self.opts = rpc.criu_opts()
+        # Since images_dir_fd is a required field in rpc.proto
+        # we need to explicitly pass it into opts
+        self.opts.images_dir_fd = self.opts.images_dir_fd
         self.sk = None
 
     def use_sk(self, sk_name):


lib/pycriu/criu.py
```diff
@@ -278,6 +278,7 @@ def dump(self):
         req = rpc.criu_req()
         req.type = rpc.DUMP
         req.opts.MergeFrom(self.opts)
+        req.opts.images_dir_fd = self.opts.images_dir_fd
```
avagin: it might be better to do something like that:
```
diff --git a/lib/pycriu/criu.py b/lib/pycriu/criu.py
index 5bd7ffecd..a496756d5 100644
--- a/lib/pycriu/criu.py
+++ b/lib/pycriu/criu.py
@@ -204,6 +204,9 @@ class criu:
     def __init__(self):
         self.use_binary('criu')
         self.opts = rpc.criu_opts()
+        # write a comment here why it is needed.
+        if self.opts.HasField("image_dir_fd"):
+          self.opts.images_dir_fd = -1
         self.sk = None
 
     def use_sk(self, sk_name):
```

herheliuk: Actually we can't do that, because we need to explicitly set `images_dir_fd` in `req.opts` not `self.opts`

P.S. `CopyFrom(self.opts)` doesn't work either, since it's a **required field** we need to pass it explicitly.

herheliuk: Oh wait, my bad

herheliuk:
#2789 Copies this PR.

PR #2799, opened 2025-10-23
cr-service: skip dump&restore specific optons when doing check

herheliuk:
Fixes pycriu check funciton


herheliuk:
with this fix:

```python
from pycriu import criu as pycriu_criu

criu = pycriu_criu()

criu.track_mem = True

criu.check()
```

doesn't raise an error, even though I'm on aarch64.

So it's likely that #2789 would actually be more appropriate.

PR #2802, opened 2025-10-23
pycriu: better `dump` errors

herheliuk:
Specify the reason for a dump fail (No dumps dir).

### DIFF
--- a/lib/pycriu/criu.py
+++ b/lib/pycriu/criu.py
@@ -181,6 +181,9 @@ class CRIUExceptionExternal(CRIUException):
         if self.errno == errno.EBADRQC:
             s += "Bad options"
 
+        elif self.typ == rpc.DUMP and self.errno == errno.ENOENT:
+            s += "Dump folder not found. Please create it first."
+
         elif self.typ == rpc.DUMP and self.errno == errno.ESRCH:
             s += "No process with such pid"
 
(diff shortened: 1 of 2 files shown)


lib/pycriu/criu.py
```diff
@@ -181,6 +181,9 @@ def _gen_error_str(self):
         if self.errno == errno.EBADRQC:
             s += "Bad options"
 
+        elif self.typ == rpc.DUMP and self.errno == errno.ENOENT:
+            s += "Dump folder not found. Please create it first."
```
avagin, line 185: It can be a workdir and any other file patch. I think it is better to add cr_errmsg into the error message.

herheliuk: I'm not sure how to find and send `pr_perror`'s value back...

But it would make sense indeed. (Currently pycriu receives "Unknown" most of the time)

As far as this commit goes though, I believe the change in `cr-service.c`
and python check is enough to nail down this specific error.

herheliuk: `cr-sercice.c` lines 137-142

```
static void set_resp_err(CriuResp *resp)
{
	resp->cr_errno = get_cr_errno();
	resp->has_cr_errno = resp->cr_errno ? true : false;
	resp->cr_errmsg = log_first_err();
}
```

`cr_errmsg` is here, but `pycriu` doesn't seem to receive/process this value

herheliuk: `log_first_err()` is returning `NULL` #2805

Edited: when using `use_binary`

unichronic:
Hey @herheliuk why was this PR #2805 closed? Was the issue solved?

herheliuk:
@unichronic
This PR (#2802) was closed because we need to approach this issue differently, fixing the issue mentioned in #2805.

Instead of duplicating already existing error messages, we need to figure out why they weren't sent in the first place.

PR #2806, opened 2025-10-27
pycriu: making sk_name optional

herheliuk:
so user can just call criu.use_sk() without any parameters to use the default socket

Co-authored-by: Radostin Stoyanov <rstoyanov@fedoraproject.org>

### DIFF
--- a/lib/pycriu/criu.py
+++ b/lib/pycriu/criu.py
@@ -8,6 +8,7 @@ import struct
 
 import pycriu.rpc_pb2 as rpc
 
+CR_DEFAULT_SERVICE_ADDRESS = "./criu_service.socket"
 
 class _criu_comm:
     """
@@ -213,7 +214,7 @@ class criu:
         self.opts = rpc.criu_opts()
         self.sk = None
 
-    def use_sk(self, sk_name):
+    def use_sk(self, sk_name=CR_DEFAULT_SERVICE_ADDRESS):
         """
         Access criu using unix socket which that belongs to criu service daemon.
         """


lib/pycriu/criu.py
```diff
@@ -213,7 +213,7 @@ def __init__(self):
         self.opts = rpc.criu_opts()
         self.sk = None
 
-    def use_sk(self, sk_name):
+    def use_sk(self, sk_name='criu_service.socket'):
```
rst0git: It would be better to use a constant similar to the definition in [include/cr-service-const.h](https://github.com/checkpoint-restore/criu/blob/ae971c9fe069e3d24e461740ce936080bc7b6812/criu/include/cr-service-const.h):

```python
CR_DEFAULT_SERVICE_ADDRESS = "./criu_service.socket"
```

This constant can be used as follows:

```python
def use_sk(self, sk_name=CR_DEFAULT_SERVICE_ADDRESS):
...
```

```python
print(pycriu.criu.CR_DEFAULT_SERVICE_ADDRESS)
```

herheliuk: Yeah, that's the right approach!

Same thing was with `self.opts.images_dir_fd` = `self.opts.images_dir_fd`,
perverting technical debt ^^

herheliuk: Unfortunately though you have to do `CR_DEFAULT_SERVICE_ADDRESS = "./criu_service.socket"`
on top of the file instead of using some environment variable...

herheliuk: So I don't think that we need `pycriu.criu.CR_DEFAULT_SERVICE_ADDRESS`
and it's more intuitive to have it as I did.

P.S. You've done the same with `self.opts = rpc.criu_opts(images_dir_fd=-1)` just 3 lines above

herheliuk: Although you're right, #2800 states that people can easily debug it themselves...

`pycriu.criu.CR_DEFAULT_SERVICE_ADDRESS` would allow them to do just that.

PR #2810, opened 2025-11-01
Dmabuf IPC support for amdgpu plugin (v2)

rst0git:
This patch set, in combination with the accompanying kernel patch set, is the second draft of work to support amdgpu dmabuf IPC. Since the first post, the patches have been split to make them more readable and made backwards-compatible.

This is pull request contains the patches from https://github.com/checkpoint-restore/criu/pull/2613 with changes from the [`wip/amdgpu`](https://github.com/checkpoint-restore/criu/tree/wip/amdgpu) branch.

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -20,7 +20,6 @@
 
 #include <xf86drm.h>
 #include <libdrm/amdgpu.h>
-#include <libdrm/amdgpu_drm.h>
 
 #include "criu-plugin.h"
 #include "plugin.h"
@@ -38,6 +37,7 @@
 #include "rst-malloc.h"
 
 #include "common/list.h"
+#include "amdgpu_drm.h"
 #include "amdgpu_plugin_dmabuf.h"
 #include "amdgpu_plugin_drm.h"
 #include "amdgpu_plugin_util.h"
--- a/plugins/amdgpu/amdgpu_plugin_drm.c
+++ b/plugins/amdgpu/amdgpu_plugin_drm.c
@@ -23,6 +23,8 @@
 #include "fdstore.h"
 
 #include "criu-amdgpu.pb-c.h"
+
+/* Define __user as empty for kernel headers in user-space */
 #define __user
 #include "drm.h"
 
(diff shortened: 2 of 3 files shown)


plugins/amdgpu/amdgpu_drm.h
```diff
@@ -348,7 +348,7 @@ union drm_amdgpu_ctx {
  * For operation AMDGPU_USERQ_OP_CREATE: user is expected
  *  to set all fields, excep the parameter 'queue_id'.
  * For operation AMDGPU_USERQ_OP_FREE: the only input parameter expected
- *  to be set is 'queue_id', eveything else is ignored.
+ *  to be set is 'queue_id', everything else is ignored.
```
fdavid-amd: I agree this is a mistake, but the point of this file is that it's a 1-for-1 copy of the original kernel header, and this mistake is in the original. Probably should be kept for that reason, but if it makes CI tools upset this can be fixed.

rst0git: As an alternative, we can add these files in the `skip` field of `.codespellrc`

I noticed that we also include `<libdrm/amdgpu_drm.h>` in `plugins/amdgpu/amdgpu_plugin.c` and `<libdrm/drm.h>` in `plugins/amdgpu/kfd_ioctl.h`. Should we replace the libdrm includes with these headers?

fdavid-amd: Yes, good catch

rst0git: Hmm, with this change CRIU fails to build with the following error:
```
In file included from kfd_ioctl.h:30,
                 from amdgpu_plugin_topology.c:18:
drm.h:1047:10: fatal error: drm_mode.h: No such file or directory
 1047 | #include "drm_mode.h"
      |          ^~~~~~~~~~~~
```
`drm_mode.h` is kernel header file that does not currently exist in the CRIU repository.

rst0git: @avagin Do you have any advice on how to handle this? Should we add `drm_mode.h` as well?

fdavid-amd: Let me check this

fdavid-amd: Not sure why this wasn't causing trouble before, but there sure is an include in the middle of drm.h. 
Looks like we'll have to add drm_mode.h as well (which thankfully doesn't include anything else)
Annoying but doable.

rst0git: I've updated the pull request with this change.

Issue #2781, opened 2025-10-16
runc tests fail with criu-dev


kolyshkin:
For the last few years, runc repo have a CI job to test runc against criu-dev branch, mostly to catch some regressions early.

We are seeing failures recently (this week), alas without too much diagnostics.

For example (from [here](https://productionresultssa6.blob.core.windows.net/actions-results/865121f5-ae53-4a5d-a72a-1f3aace783d2/workflow-job-run-15406e29-e965-5079-8b70-bbc24930c4cb/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-10-16T17%3A41%3A00Z&sig=F9TpGdybzQNGVNqU1oT0IKHCEtLAy%2BqSiUzOFRMHqt8%3D&ske=2025-10-17T05%3A10%3A53Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-10-16T17%3A10%3A53Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2025-10-16T17%3A30%3A55Z&sv=2025-11-05), look for `not ok 32`):
```
2025-10-16T01:03:57.5662977Z not ok 32 checkpoint and restore
2025-10-16T01:03:57.5673614Z # (from function `simple_cr' in file tests/integration/checkpoint.bats, line 131,
2025-10-16T01:03:57.5675071Z #  in test file tests/integration/checkpoint.bats, line 192)
2025-10-16T01:03:57.5698743Z #   `simple_cr' failed
2025-10-16T01:03:57.5712389Z # runc spec (status=0)
2025-10-16T01:03:57.5713588Z #
2025-10-16T01:03:57.5714281Z # runc run -d --console-socket /tmp/bats-run-kiFEqC/runc.CtdpiC/tty/sock test_busybox (status=0)
2025-10-16T01:03:57.5715346Z #
2025-10-16T01:03:57.5715677Z # runc state test_busybox (status=0)
2025-10-16T01:03:57.5716304Z # {
2025-10-16T01:03:57.5716613Z #   "ociVersion": "1.2.1+dev",
2025-10-16T01:03:57.5716999Z #   "id": "test_busybox",
2025-10-16T01:03:57.5717360Z #   "pid": 13992,
2025-10-16T01:03:57.5717695Z #   "status": "running",
[...]

PR #2818, opened 2025-11-11
cr-servce: move kerndat_init after log_init

rst0git:
`kerndat_init()` should be called after logging has been initialized.

Update: #2781


avagin:
CI tests fails because init_pidfd_store_sk is called before kerndat_init.

avagin:
you can use the following command to reproduce the issue locally:
```
./test/zdtm.py run -t zdtm/transition/pidfd_store_sk --rpc --pre 2 --ignore-taint
```

PR #2821, opened 2025-11-11
Fix docs for `--inherit-fd` to use `pipe:[INODE_NUM]` instead of `pipe[INODE_NUM]`

markasoftware-tc:
The latter will fail silently and doesn't remap any actual pipes. The `pipe.c` test uses the correct `pipe:[INODE_NUM]` syntax by calling `readlink` on `/proc/blah/fd/blah` which returns the syntax with colon.

Wiki page https://criu.org/Inheriting_FDs_on_restore#External_unnamed_pipes already uses the correct syntax.

I haven't checked if the syntax examples for the other types of resources are correct or not.

Would be cool to actually give an error if trying to remap something that doesn't exist.


criu/crtools.c
```diff
@@ -503,7 +503,7 @@ int main(int argc, char *argv[], char *envp[])
 	       "                        Inherit file descriptors, treating fd NUM as being\n"
 	       "                        already opened via an existing RES, which can be:\n"
 	       "                            tty[rdev:dev]\n"
-	       "                            pipe[inode]\n"
+	       "                            pipe:[inode]\n"
 	       "                            socket[inode]\n"
```
rst0git: It looks like the same change is also needed for `socket:[%u]`:
https://github.com/checkpoint-restore/criu/blob/a1fd7e629be95ca4263a8d911778c6e81be4b819/criu/sk-unix.c#L1986

rst0git:
@markasoftware-tc Would you be able to add a `Signed-off-by` line in your commit message?
https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes

markasoftware-tc:
Done for both.

Did you check that the other variants of the option are correct or should I check myself?

rst0git:
> Did you check that the other variants of the option are correct or should I check myself?

https://github.com/checkpoint-restore/criu/blob/a1fd7e629be95ca4263a8d911778c6e81be4b819/criu/tty.c#L1217
https://github.com/checkpoint-restore/criu/blob/a1fd7e629be95ca4263a8d911778c6e81be4b819/criu/files-reg.c#L1776
https://github.com/checkpoint-restore/criu/blob/a1fd7e629be95ca4263a8d911778c6e81be4b819/criu/memfd.c#L26

rst0git (approved):
LGTM

markasoftware-tc:
Not sure what the conventions are for CRIU but it seems I don't have permissions to merge this

avagin:
Merged. Thanks.

PR #2830, opened 2025-11-21
Util: close_safe must not clear fd on close() failure

shbhmexe:
This pull request addresses potential file descriptor leaks in the `criu/util.c` file by ensuring that file descriptors are properly closed if `install_service_fd` fails. This improves resource management and prevents leaking system resources.

Resource management improvements:

* In both `set_proc_fd` and `open_proc_sfd`, added logic to close the duplicated or opened file descriptor if `install_service_fd` fails, preventing file descriptor leaks. [[1]](diffhunk://#diff-1120c47e8f3ca1a8b18111630a63190d0930ad39a2c63ff40b899add44872640L357-R361) [[2]](diffhunk://#diff-1120c47e8f3ca1a8b18111630a63190d0930ad39a2c63ff40b899add44872640L374-R381)

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -354,8 +354,11 @@ int set_proc_fd(int fd)
 		pr_perror("dup() failed");
 		return -1;
 	}
-	if (install_service_fd(PROC_FD_OFF, _fd) < 0)
+	if (install_service_fd(PROC_FD_OFF, _fd) < 0) {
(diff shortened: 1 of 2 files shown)


adrianreber:
Looking at the code I see that `install_service_fd()` already closes the fd if an error happened. Now it is closed twice.

The commit description also does not really match what happens in the code.

This feels like some coding assistant generated changes which look wrong to me.

PR #2832, opened 2025-11-26
Fixing two small static code analysis warnings

adrianreber:
I got a static code analysis report and this fixes the warnings.

Both reports do not look like a real problem, but the fixes are simple at the same time. If this is seen as unnecessary I can close it again.


rst0git (approved):
LGTM

PR #2833, opened 2025-11-26
tty: fix compiler error

adrianreber:
At least on tests running on Fedora rawhide following error could be seen:

```
  criu/tty.c: In function 'pts_fd_get_index':
  criu/tty.c:262:21: error: initialization discards 'const' qualifier from pointer target type [-Werror=discarded-qualifiers]
    262 |         char *pos = strrchr(link->name, '/');
        |
```

This fixes it.


rst0git (approved):
LGTM

PR #2836, opened 2025-12-02
criu: add support for --gpu-device-map

rst0git:
This option allows to specify a GPU device map that will be used during restore. The provided map must contain all checkpointed GPU devices in the format `"oldUuid=newUuid,oldUuid=newUuid,..."` that will be used to remap old devices to new devices.

In the following example all process checkpointed on `$GPU_0` will be restored on `$GPU_1`:

```bash
i=0
for uuid in $(nvidia-smi --list-gpus | grep -oP 'UUID: \K[^)]+'); do
    export GPU_$i=$uuid
    i=$((i+1))
done

criu restore --gpu-device-map=$GPU_0=$GPU_1,$GPU_1=$GPU_0,$GPU_2=$GPU_2,$GPU_3=$GPU_3
```

Demo: https://youtu.be/8F6RAXb_GfU


rst0git:
Adding a note here that we discussed this with  @avagin and he suggested to extend the functionality of `--external` instead of introducing and a new option.

rst0git:
Closing  in favour of https://github.com/checkpoint-restore/criu/pull/3107

PR #2841, opened 2025-12-05
plugin: add UPDATE_INETSK hook to adjust inet socket IPs on restore

Rowen-Ye:
We need to migrate processes across nodes/subnets, and after restore the saved socket IPs may no longer make sense. This PR introduces a new plugin hook CR_PLUGIN_HOOK__UPDATE_INETSK.

The idea is to let a plugin update IPv4/IPv6 addresses in InetSkEntry during restore. The hook runs in collect_one_inetsk() before we do port reservation / bind / connect, so the plugin can adjust the addresses in-place. The old descriptor autogen is also updated so existing plugins keep working.

Usage is simple: the plugin rewrites src_ip / dst_ip (network byte order) and returns 0. -ENOTSUP means “not for me”—we try the next plugin. Any other error aborts restore for that socket. Only the IPs are allowed to change; we still enforce family, ports, ifname, and length constraints.

A note regarding TCP: for established connections you still need external coordination or some mechanism to preserve the original 4-tuple (NAT/overlay/etc.). This hook only changes addresses, it doesn’t touch TCP state. UDP and listen sockets work fine; connected TCP success depends on the network setup.

### DIFF
--- a/criu/include/criu-plugin.h
+++ b/criu/include/criu-plugin.h
@@ -66,6 +66,8 @@ enum {
 
 	CR_PLUGIN_HOOK__DUMP_DEVICES_LATE = 14,
 
+	CR_PLUGIN_HOOK__UPDATE_INETSK = 15,
+
 	CR_PLUGIN_HOOK__MAX
 };
 
@@ -87,6 +89,7 @@ DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__CHECKPOINT_DEVICES, int pid);
 DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__POST_FORKING, void);
 DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__RESTORE_INIT, void);
 DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__DUMP_DEVICES_LATE, int id);
+DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__UPDATE_INETSK, uint32_t family, uint32_t state, uint32_t *src_ip, uint32_t *dst_ip);
 
 enum {
 	CR_PLUGIN_STAGE__DUMP,
@@ -162,5 +165,6 @@ typedef int(cr_plugin_update_vma_map_t)(const char *path, const uint64_t addr, c
 					uint64_t *new_pgoff, int *plugin_fd);
 typedef int(cr_plugin_resume_devices_late_t)(int pid);
 typedef int(cr_plugin_post_forking_t)(void);
+typedef int(cr_plugin_update_inetsk_t)(uint32_t family, uint32_t state, uint32_t *src_ip, uint32_t *dst_ip);
 
 #endif /* __CRIU_PLUGIN_H__ */
--- a/criu/sk-inet.c
+++ b/criu/sk-inet.c
@@ -35,6 +35,7 @@
 #include "protobuf.h"
 #include "util.h"
 #include "namespaces.h"
+#include "plugin.h"
 
 #include "images/inventory.pb-c.h"
 
@@ -707,8 +708,14 @@ static inline int tcp_connection(InetSkEntry *ie)
 static int collect_one_inetsk(void *o, ProtobufCMessage *base, struct cr_img *i)
 {
 	struct inet_sk_info *ii = o;
+	int ret;
 
 	ii->ie = pb_msg(base, InetSkEntry);
+	
+	ret = run_plugins(UPDATE_INETSK, ii->ie->family, ii->ie->state, ii->ie->src_addr, ii->ie->dst_addr);
+	if (ret < 0 && ret != -ENOTSUP)
+		return -1;
+
 	if (tcp_connection(ii->ie))
 		tcp_locked_conn_add(ii);
 
--- a/criu/plugin.c
+++ b/criu/plugin.c
@@ -62,6 +62,7 @@ static cr_plugin_desc_t *cr_gen_plugin_desc(void *h, char *path)
 	__assign_hook(POST_FORKING, "cr_plugin_post_forking");
 	__assign_hook(RESTORE_INIT, "cr_plugin_restore_init");
 	__assign_hook(DUMP_DEVICES_LATE, "cr_plugin_dump_devices_late");
+	__assign_hook(UPDATE_INETSK, "cr_plugin_update_inetsk");
 
 #undef __assign_hook
 


adrianreber:
Interesting approach. Please also provide the plugin and some tests to see how it all works together.

adrianreber:
Also rework your commit message to have line breaks. Take a look at https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes and there are some recommendations (and a link) how to write commit messages.

rst0git:
@Rowan-Ye you can use a simple script to modify the content of `files.img`:
- https://github.com/stano45/p4containerflow/blob/main/scripts/edit_files_img.py
- https://github.com/p4lang/gsoc/tree/main/2024/projects/container_migration

You can run this as an [action-script](https://criu.org/Action_scripts) (instead of plugin) and use the `CRTOOLS_IMAGE_DIR` environment variable to find this image file.

The following project also uses similar approach for coordination:
https://github.com/checkpoint-restore/criu-coordinator

rst0git:
We use a similar approach to overwrite TCP listen socket ports in [criu-image-streamer](https://github.com/checkpoint-restore/criu-image-streamer/commit/fd83e8521af57af6f171a16205516c6ded63db22).

Rowen-Ye:
> @Rowan-Ye you can use a simple script to modify the content of `files.img`:
> 
> * https://github.com/stano45/p4containerflow/blob/main/scripts/edit_files_img.py
> * https://github.com/p4lang/gsoc/tree/main/2024/projects/container_migration
> 
> You can run this as an [action-script](https://criu.org/Action_scripts) (instead of plugin) and use the `CRTOOLS_IMAGE_DIR` environment variable to find this image file.
> 
> The following project also uses similar approach for coordination: https://github.com/checkpoint-restore/criu-coordinator

@rst0git  Thanks for the references — the scripts are indeed quite useful in setups where the environment is predictable.
In our case the restore happens inside Kubernetes, and that leads to a bit of a timing issue. The pod’s IP only becomes available after the new pod is created and the CNI finishes setting up the network namespace. Editing files.img ahead of time means we’d have to know that IP in advance, which unfortunately we don’t — hence the “chicken-and-egg” problem on our side.
Using a hook during restore avoids that. At that point the target netns already exists and has its final IP assigned, so the hook can read the actual value and patch the socket entries on the fly. This fits our workflow (checkpoint an inference service → fast-start it via K8s) much better.

Rowen-Ye:
> We use a similar approach to overwrite TCP listen socket ports in [criu-image-streamer](https://github.com/checkpoint-restore/criu-image-streamer/commit/fd83e8521af57af6f171a16205516c6ded63db22).

Thanks for bringing up the criu-image-streamer approach — the port-remap option there is a neat solution when the mappings are fixed and known ahead of time.
For our use case (migrating a vLLM inference service in K8s), the situation is a bit more varied. We run into multiple socket types and states. The streamer’s remap logic becomes harder to express the different socket rewrites we need, especially when the values depend on the runtime environment.
A hook during restore gives us more room to work with, we can apply whatever adjustments are needed on a per-socket basis.

Rowen-Ye:
> Also rework your commit message to have line breaks. Take a look at https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes and there are some recommendations (and a link) how to write commit messages.

thanks for the pointer. I’ll update the commit message soon.

Rowen-Ye:
> Interesting approach. Please also provide the plugin and some tests to see how it all works together.

@adrianreber  Thanks, that makes sense. I’ve put together a small plugin to show how the CR_PLUGIN_HOOK__UPDATE_INETSK hook is intended to be used. The plugin reads a target IPv4 address from an environment variable (for example INETSK_LOCAL_IPV4) at restore time and rewrites the src/dst addresses inside InetSkEntry in place. For AF_INET sockets it replaces the 32‑bit src/dst IP with this value; for AF_INET6 sockets it converts the same IPv4 into an IPv4‑mapped IPv6 (::ffff:w.x.y.z) and writes that into the 4×32‑bit address array. To avoid touching special cases, it skips sockets whose source address is “unspecified” or loopback (e.g. 0.0.0.0, 127.0.0.1, ::, ::1). Apart from that, it only adjusts the address words and leaves family/ports unchanged, then returns 0 so the normal restore logic continues.

 I’ll follow up with a proper C implementation of this plugin plus some tests, so it’s clear how it behaves end‑to‑end with CRIU.

```
// Convert IPv4 string into IPv4-mapped IPv6 form (::ffff:x.x.x.x)
struct in6_addr ipv4_to_ipv6_mapped(const char *ipv4_str)
{
    struct in6_addr ipv6;
    uint8_t *p;

    if (ipv4_str invalid)
        return IPv6_ZERO_ADDR;

    memset(&ipv6, 0, sizeof(ipv6));
    ipv6.s6_addr[10] = 0xff;
    ipv6.s6_addr[11] = 0xff;

    p = ipv4_string_to_bytes(ipv4_str);   // returns 4 bytes
    memcpy(&ipv6.s6_addr[12], p, 4);

    return ipv6;
}

// Check whether the given source IP should be ignored
bool should_skip(uint32_t family, uint32_t ip[4])
{
    if (family == AF_INET) {
        if (ip[0] == 0 || ip[0] == 0x7f000001)
            return true;
    } else {
        if (ip == (:: or ::1))
            return true;
    }
    return false;
}

// Update IPv4 src/dst
void update_ipv4(uint32_t src[1], uint32_t dst[1], const char *env_ip, uint32_t state)
{
    uint32_t new_ip = ipv4_to_u32(env_ip);
    src[0] = new_ip;

    if (state == 1)
        dst[0] = new_ip;
}

// Update IPv6 using IPv4‐mapped IPv6 (::ffff:w.x.y.z)
void update_ipv6(uint32_t src[4], uint32_t dst[4], const char *env_ip, uint32_t state)
{
    struct in6_addr mapped = ipv4_to_ipv6_mapped(env_ip);
    uint32_t *w = (uint32_t *)&mapped;

    for (int i = 0; i < 4; i++)
        src[i] = w[i];

    if (state == 1) {
        for (int i = 0; i < 4; i++)
            dst[i] = w[i];
    }
}

// Main update hook
int update_inetsk(uint32_t family, uint32_t state, uint32_t *src_ip, uint32_t *dst_ip)
{
    const char *env_ip;

    if (!src_ip)
        return 0;

    if (should_skip(family, src_ip))
        return 0;

    env_ip = getenv(INETSK_LOCAL_IPV4);
    if (!env_ip || env_ip empty)
        return 0;

    if (family == AF_INET)
        update_ipv4(src_ip, dst_ip, env_ip, state);
    else
        update_ipv6(src_ip, dst_ip, env_ip, state);

    return 0;
}

// Register hook
CR_PLUGIN_REGISTER_HOOK(CR_PLUGIN_HOOK__UPDATE_INETSK, update_inetsk);
```

Our main use case for this is a fairly complex end‑to‑end Kubernetes setup (CNI, pod lifecycle, service routing, vLLM inference traffic, etc.), so the “real” validation happens in that environment rather than in a small, self‑contained CRIU test. Because of that, I’m afraid I can’t easily provide a minimal automated test that fully reproduces our scenario.

rst0git:
> Editing files.img ahead of time means we’d have to know that IP in advance, which unfortunately we don’t — hence the “chicken-and-egg” problem on our side. Using a hook during restore avoids that.

@Rowan-Ye Using an "action-script" to modify the image would be the same as using a plugin hook - CRIU runs this script during restore. The only difference is that it works out-of-the-box and doesn't require changes in CRIU. The following page provides more information: https://criu.org/Action_scripts

rst0git:
>Our main use case for this is a fairly complex end‑to‑end Kubernetes setup (CNI, pod lifecycle, service routing, vLLM inference traffic, etc.), so the “real” validation happens in that environment rather than in a small, self‑contained CRIU test.

The use-case sounds similar to what we have been working on. It would be great if you can join our working group:
- https://criu.org/Kubernetes
- https://github.com/kubernetes/community/tree/master/wg-checkpoint-restore

Rowen-Ye:
> > Editing files.img ahead of time means we’d have to know that IP in advance, which unfortunately we don’t — hence the “chicken-and-egg” problem on our side. Using a hook during restore avoids that.
> 
> @Rowan-Ye Using an "action-script" to modify the image would be the same as using a plugin hook - CRIU runs this script during restore. The only difference is that it works out-of-the-box and doesn't require changes in CRIU. The following page provides more information: https://criu.org/Action_scripts

@rst0git Thanks for the pointer to action‑scripts, but in my setup, the checkpoint image is stored once in a distributed storage system and reused by multiple nodes; we don’t copy the image locally per node.

If we use an action‑script that edits the image on restore, different nodes would end up modifying and consuming the same shared image concurrently, which is quite hard to reason about and coordinate (races, partial edits, etc.). With the plugin hook we can keep the image immutable and only adjust the in‑memory socket addresses on each node at restore time, which avoids those concurrency issues and fits better with our “one shared image, many consumers” model.

Rowen-Ye:
> > Our main use case for this is a fairly complex end‑to‑end Kubernetes setup (CNI, pod lifecycle, service routing, vLLM inference traffic, etc.), so the “real” validation happens in that environment rather than in a small, self‑contained CRIU test.
> 
> The use-case sounds similar to what we have been working on. It would be great if you can join our working group:
> 
> * https://criu.org/Kubernetes
> * https://github.com/kubernetes/community/tree/master/wg-checkpoint-restore

That sounds great, thanks for sharing the links. The use case you’re working on is indeed very close to what we’re trying to do with vLLM on Kubernetes, so I’d be very interested in following and contributing where I can.

Rowen-Ye:
> > > Editing files.img ahead of time means we’d have to know that IP in advance, which unfortunately we don’t — hence the “chicken-and-egg” problem on our side. Using a hook during restore avoids that.
> > 
> > 
> > @Rowan-Ye Using an "action-script" to modify the image would be the same as using a plugin hook - CRIU runs this script during restore. The only difference is that it works out-of-the-box and doesn't require changes in CRIU. The following page provides more information: https://criu.org/Action_scripts
> 
> @rst0git Thanks for the pointer to action‑scripts, but in my setup, the checkpoint image is stored once in a distributed storage system and reused by multiple nodes; we don’t copy the image locally per node.
> 
> If we use an action‑script that edits the image on restore, different nodes would end up modifying and consuming the same shared image concurrently, which is quite hard to reason about and coordinate (races, partial edits, etc.). With the plugin hook we can keep the image immutable and only adjust the in‑memory socket addresses on each node at restore time, which avoids those concurrency issues and fits better with our “one shared image, many consumers” model.

@rst0git @adrianreber 

I shared our use case above and why, in our environment, a plugin hook fits better than action-scripts . If there’s anything missing, or if you’d like me to adjust the proposal to better fit the project’s design direction, I’m happy to iterate.

When you have a moment, I’d really appreciate another look or any guidance on next steps.

rst0git:
@Rowan-Ye Changing the IP address is only part of the solution - how do you tell the connected clients to use a new IP address for established connections?

The following page provides more information about the problem and possible solutions: https://criu.org/Change_IP_address

Rowen-Ye:
> @Rowan-Ye Changing the IP address is only part of the solution - how do you tell the connected clients to use a new IP address for established connections?
> 
> The following page provides more information about the problem and possible solutions: https://criu.org/Change_IP_address

@rst0git In our scenario, when we migrate across nodes, all endpoints involved in the established TCP connections are checkpointed and restored together. During restore we use a shared/global mapping to determine the new addresses for each endpoint. The UPDATE_INETSK hook reads that mapping and updates both src_ip and dst_ip accordingly, so both sides of the connection are restored with a consistent view of the new addressing.

Rowen-Ye:
> > @Rowan-Ye Changing the IP address is only part of the solution - how do you tell the connected clients to use a new IP address for established connections?
> > The following page provides more information about the problem and possible solutions: https://criu.org/Change_IP_address
> 
> @rst0git In our scenario, when we migrate across nodes, all endpoints involved in the established TCP connections are checkpointed and restored together. During restore we use a shared/global mapping to determine the new addresses for each endpoint. The UPDATE_INETSK hook reads that mapping and updates both src_ip and dst_ip accordingly, so both sides of the connection are restored with a consistent view of the new addressing.

 @rst0git  I did some additional experiments in our environment and found a practical split that works well for us. For established TCP connections, we use --tcp-close during restore to intentionally close them and let the application layer reconnect using our migration IP mapping. For the other TCP states (e.g., LISTEN), we keep them and rely on the UPDATE_INETSK plugin hook to rewrite the addresses at restore time.

With this approach we avoid the hardest part (preserving established TCP across IP changes) while still supporting seamless migration of non-established sockets, which is enough for our inference service use case.

criu/include/criu-plugin.h
```diff
@@ -87,6 +89,7 @@ DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__CHECKPOINT_DEVICES, int pid);
 DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__POST_FORKING, void);
 DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__RESTORE_INIT, void);
 DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__DUMP_DEVICES_LATE, int id);
+DECLARE_PLUGIN_HOOK_ARGS(CR_PLUGIN_HOOK__UPDATE_INETSK, uint32_t family, uint32_t state, uint32_t src_ip, uint32_t dst_ip);
```
avagin: only four byte address is passed to the plugin hook. How can it handled ipv6 addresses?

criu/sk-inet.c
```diff
@@ -707,8 +708,14 @@ static inline int tcp_connection(InetSkEntry *ie)
 static int collect_one_inetsk(void *o, ProtobufCMessage *base, struct cr_img *i)
 {
 	struct inet_sk_info *ii = o;
+	int ret;
 
 	ii->ie = pb_msg(base, InetSkEntry);
+	
+	ret = run_plugins(UPDATE_INETSK, ii->ie->family, ii->ie->state, ii->ie->src_addr, ii->ie->dst_addr);
```
avagin, line 715: ii->ie->src_addr, ii->ie->dst_addr are pointers. Is this code compilable?

avagin:
Do we need to handle bound/listening sockets as well?

rst0git:
>Do we need to handle bound/listening sockets as well?

Yes, we have use-cases where users want to remap the ports of listening sockets during restore:
https://github.com/checkpoint-restore/criu-image-streamer/commit/fd83e8521af57af6f171a16205516c6ded63db22

criu/include/criu-plugin.h
Rowen-Ye: src_ip and dst_ip are passed as pointers rather than fixed-size values. The actual address length is determined by the family parameter, so the hook can handle both IPv4 and IPv6. For IPv4, the pointer refers to a 4-byte address, and for IPv6, it refers to a 16-byte address.

criu/sk-inet.c
Rowen-Ye: Yes, this is compilable. In addition to src_addr and dst_addr being pointers, the family and state parameters together provide enough information to identify the socket type and its current state.

avagin: https://app.circleci.com/pipelines/github/checkpoint-restore/criu/3636/workflows/a11cdf06-1308-45fe-9d81-7a37da610531/jobs/7270
```
In file included from criu/sk-inet.c:38:
criu/sk-inet.c: In function ‘collect_one_inetsk’:
criu/sk-inet.c:715:79: error: passing argument 3 of ‘(int (*)(uint32_t,  uint32_t,  uint32_t,  uint32_t))this->d->hooks[15]’ makes integer from pointer without a cast [-Werror=int-conversion]
  715 |         ret = run_plugins(UPDATE_INETSK, ii->ie->family, ii->ie->state, ii->ie->src_addr, ii->ie->dst_addr);
      |                                                                         ~~~~~~^~~~~~~~~~
      |                                                                               |
      |                                                                               uint32_t * {aka unsigned int *}
criu/include/plugin.h:39:33: note: in definition of macro ‘run_plugins’
   39 |                                 __VA_ARGS__);                                                               \
      |                                 ^~~~~~~~~~~
criu/sk-inet.c:715:79: note: expected ‘uint32_t’ {aka ‘unsigned int’} but argument is of type ‘uint32_t *’ {aka ‘unsigned int *’}
  715 |         ret = run_plugins(UPDATE_INETSK, ii->ie->family, ii->ie->state, ii->ie->src_addr, ii->ie->dst_addr);
      |                                                                         ~~~~~~^~~~~~~~~~
criu/include/plugin.h:39:33: note: in definition of macro ‘run_plugins’
   39 |                                 __VA_ARGS__);                                                               \
      |                                 ^~~~~~~~~~~
criu/sk-inet.c:715:97: error: passing argument 4 of ‘(int (*)(uint32_t,  uint32_t,  uint32_t,  uint32_t))this->d->hooks[15]’ makes integer from pointer without a cast [-Werror=int-conversion]
  715 |         ret = run_plugins(UPDATE_INETSK, ii->ie->family, ii->ie->state, ii->ie->src_addr, ii->ie->dst_addr);
      |                                                                                           ~~~~~~^~~~~~~~~~
      |                                                                                                 |
      |                                                                                                 uint32_t * {aka unsigned int *}
criu/include/plugin.h:39:33: note: in definition of macro ‘run_plugins’
   39 |                                 __VA_ARGS__);                                                               \
      |                                 ^~~~~~~~~~~
criu/sk-inet.c:715:97: note: expected ‘uint32_t’ {aka ‘unsigned int’} but argument is of type ‘uint32_t *’ {aka ‘unsigned int *’}
  715 |         ret = run_plugins(UPDATE_INETSK, ii->ie->family, ii->ie->state, ii->ie->src_addr, ii->ie->dst_addr);
      |                                                                                           ~~~~~~^~~~~~~~~~
criu/include/plugin.h:39:33: note: in definition of macro ‘run_plugins’
   39 |                                 __VA_ARGS__);                                                               \
      |                                 ^~~~~~~~~~~
```

criu/include/criu-plugin.h
avagin: `uint32_t src_ip` doesn't look like a pointer;)

Rowen-Ye: Apologies, this was an error in my previous commit. 
I’ve corrected the hook arguments to use pointers and updated the commit accordingly.
Thank you for pointing this out.

PR #2850, opened 2025-12-16
Check the return value of sys_rseq and report errors in advance

leeq2016:
I encountered a SIGSEGV issue and after debugging for a long time, I found that it was caused by a failure of sys_rseq.  Adding a check on the return value and announcing the error beforehand would greatly facilitate problem detection. Please consider this. :)

### DIFF
--- a/criu/pie/restorer.c
+++ b/criu/pie/restorer.c
@@ -1363,13 +1363,19 @@ __visible void __export_unmap(void)
 	sys_munmap(bootstrap_start, bootstrap_len - vdso_rt_size);
 }
 
-static void unregister_libc_rseq(struct rst_rseq_param *rseq)
+static int unregister_libc_rseq(struct rst_rseq_param *rseq)
 {
-	if (!rseq->rseq_abi_pointer)
-		return;
+	long ret;
 
-	/* can't fail if rseq is registered */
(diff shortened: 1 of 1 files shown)


adrianreber:
Sounds reasonable. We have seen many people confused about this error.

leeq2016:
> Sounds reasonable. We have seen many people confused about this error.

Thanks for reply.
Because the rseq syscall might not have unmapped rseq and returned an error code, but this wasn't checked here, and later `unmap_old_vmas` directly unmapped the relevant addresses, if the rseq-related processing is accessed in the subsequent user-mode return of the syscall, a segmentation fault will be triggered.

avagin:
LGTM. Thanks!

avagin:
@leeq2016 btw did you figure out why it failed for your workload?

leeq2016:
> @leeq2016 btw did you figure out why it failed for your workload?

I was using an older version of CRIU. In my case, the rseq_abi_size was 20, while the kernel required it to be 32. 
Later, I discovered that the following commit resolved my issue:
https://github.com/checkpoint-restore/criu/commit/089345f77a34d1bc7ef146d650636afcd3cdda21
:-)

avagin:
Merged. I slightly changed the commit message to make it more formal.

mihalicyn (commented):
LGTM. Thanks!

PR #2855, opened 2025-12-26
util: Make close_safe() reset fd to -1 even on close() failure

Snorch:
The "man 2 close":"Dealing with error returns from close()" says:

  "Retrying the close() after a failure return is the wrong thing to do"

We should not leave the fd there, attempting to close it again on next close()/close_safe() may lead to accidentally closing something else.

It confirms with the kernel code where sys_close() removes fd from fdtable in this stack:

```
  +-> sys_close
    +-> file_close_fd
      +-> file_close_fd_locked
        +-> rcu_assign_pointer(fdt->fd[fd], NULL)
```

If there was an fd this stack is always reached and fd is always removed.

Let's replace the fd with -1 after close no matter what.

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -222,10 +222,9 @@ int close_safe(int *fd)
 
 	if (*fd > -1) {
 		ret = close(*fd);
-		if (!ret)
-			*fd = -1;
-		else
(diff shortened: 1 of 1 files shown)


rst0git (approved):
LGTM

PR #2725, opened 2025-08-31
Aarch64/gcs

svilenkov:
This PR adds initial support for the Arm64 Guarded Control Stack (GCS). This work is part of the Google Summer of Code 2025 projects.

Checklist before merge:
- [x] retest with randomization off
- [x] retest with randomization on
- [x] retest on arm64 system without GCS

PR #2857, opened 2025-12-29
ci: fix build errors on Alpine Linux

rst0git:
The  arm64 linux-headers package in Alpine provides a version of `/usr/include/asm/ptrace.h` that includes a definition for `struct user_gcs`. This causes the build to fail as mentioned in https://github.com/checkpoint-restore/criu/pull/2725#issuecomment-3664519298. In addition, the latest Alpine linux container provides gcc version 15.2.0, which reports an error when uninitialized variable is passed as a pointer argument.

### DIFF
--- a/compel/arch/aarch64/src/lib/include/uapi/asm/infect-types.h
+++ b/compel/arch/aarch64/src/lib/include/uapi/asm/infect-types.h
@@ -20,8 +20,11 @@ typedef struct user_pt_regs user_regs_struct_t;
 
 /*
  * GCS (Guarded Control Stack)
+ *
+ * This mirrors the kernel definition but renamed to cr_user_gcs
+ * to avoid conflict with kernel headers (/usr/include/asm/ptrace.h).
  */
-struct user_gcs {
+struct cr_user_gcs {
 	__u64 features_enabled;
 	__u64 features_locked;
 	__u64 gcspr_el0;
@@ -29,7 +32,7 @@ struct user_gcs {
 
 struct user_fpregs_struct {
 	struct user_fpsimd_state fpstate;
-	struct user_gcs gcs;
+	struct cr_user_gcs gcs;
 };
 typedef struct user_fpregs_struct user_fpregs_struct_t;
 
--- a/compel/arch/arm/plugins/std/syscalls/syscall.def
+++ b/compel/arch/arm/plugins/std/syscalls/syscall.def
@@ -85,7 +85,7 @@ timer_settime			110	258	(kernel_timer_t timer_id, int flags, const struct itimer
 timer_gettime			108	259	(int timer_id, const struct itimerspec *setting)
 timer_getoverrun		109	260	(int timer_id)
 timer_delete			111	261	(kernel_timer_t timer_id)
-clock_gettime			113	263	(const clockid_t which_clock, const struct timespec *tp)
+clock_gettime			113	263	(clockid_t which_clock, struct timespec *tp)
 exit_group			94	248	(int error_code)
 set_robust_list			99	338	(struct robust_list_head *head, size_t len)
 get_robust_list			100	339	(int pid, struct robust_list_head **head_ptr, size_t *len_ptr)
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -2283,7 +2283,7 @@ void *parallel_restore_bo_contents(void *_thread_data)
 			continue;
 
 		entry = &restore_cmd->entries[i];
-		fseeko64(bo_contents_fp, entry->read_offset + offset, SEEK_SET);
+		fseeko(bo_contents_fp, entry->read_offset + offset, SEEK_SET);
 		ret = sdma_copy_bo(restore_cmd->fds_write[entry->write_id], entry->size, bo_contents_fp,
 				   buffer, buffer_size, h_dev,
 				   max_copy_size, SDMA_OP_VRAM_WRITE, false);
--- a/test/zdtm/static/tempfs_subns.c
+++ b/test/zdtm/static/tempfs_subns.c
@@ -20,7 +20,7 @@ int main(int argc, char **argv)
 {
 	int fds[2], i;
 	pid_t pid;
-	int fd, status;
+	int status, fd = -1;
 
 	test_init(argc, argv);
 
--- a/compel/arch/aarch64/src/lib/infect.c
+++ b/compel/arch/aarch64/src/lib/infect.c
@@ -41,7 +41,7 @@ bool __compel_host_supports_gcs(void)
 	return (hwcap & HWCAP_GCS) != 0;
 }
 
-static bool __compel_gcs_enabled(struct user_gcs *gcs)
+static bool __compel_gcs_enabled(struct cr_user_gcs *gcs)
 {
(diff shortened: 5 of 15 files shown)


adrianreber:
There is one more failure in the Fedora Rawhide Test:
```
 + make -C test/others/libcriu run
make: Leaving directory '/criu/test/others/criu-coredump'
make: Entering directory '/criu/test/others/libcriu'
gcc -c test_sub.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_sub.o -Werror
gcc -c lib.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o lib.o -Werror
gcc test_sub.o lib.o -L ../../../../criu/lib/c/ -L ../../../../criu/images/ -lcriu -o test_sub
gcc -c test_self.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_self.o -Werror
gcc test_self.o lib.o -L ../../../../criu/lib/c/ -L ../../../../criu/images/ -lcriu -o test_self
gcc -c test_notify.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_notify.o -Werror
gcc test_notify.o lib.o -L ../../../../criu/lib/c/ -L ../../../../criu/images/ -lcriu -o test_notify
gcc -c test_rpc_config.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_rpc_config.o -Werror
gcc test_rpc_config.o lib.o -L ../../../../criu/lib/c/ -L ../../../../criu/images/ -lcriu -o test_rpc_config
gcc -c test_iters.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_iters.o -Werror
gcc test_iters.o lib.o -L ../../../../criu/lib/c/ -L ../../../../criu/images/ -lcriu -o test_iters
gcc -c test_errno.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_errno.o -Werror
gcc test_errno.o lib.o -L ../../../../criu/lib/c/ -L ../../../../criu/images/ -lcriu -o test_errno
gcc -c test_join_ns.c -iquote ../../../../criu/criu/include -I../../../../criu/lib/c/ -I../../../../criu/images/ -o test_join_ns.o -Werror
In file included from /usr/include/bits/fcntl-linux.h:468,
                 from /usr/include/bits/fcntl.h:61,
                 from /usr/include/fcntl.h:35,
                 from test_join_ns.c:3:
../../../../criu/criu/include/linux/openat2.h:6:10: fatal error: common/config.h: No such file or directory
    6 | #include "common/config.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
make: *** [Makefile:29: test_join_ns.o] Error 1
make: Leaving directory '/criu/test/others/libcriu'
```

Just in case you have not seen it. Does not sound too difficult to fix.

compel/arch/aarch64/src/lib/include/uapi/asm/infect-types.h
```diff
@@ -18,6 +18,8 @@
 
 typedef struct user_pt_regs user_regs_struct_t;
 
+/* Define user_gcs only when the kernel headers don't provide it */
+#if !defined(HAVE_STRUCT_USER_GCS) || !HAVE_STRUCT_USER_GCS
```
avagin: it is easier to rename user_gcs to cr_user_gcs and left a comment where it is originally defined.

criu/pie/restorer.c
```diff
@@ -1219,7 +1219,7 @@ static int timerfd_arm(struct task_restore_args *args)
 		pr_debug("timerfd: arm for fd %d (%d)\n", t->fd, i);
 
 		if (t->settime_flags & TFD_TIMER_ABSTIME) {
-			struct timespec ts;
+			struct timespec ts = { .tv_sec = 0, .tv_nsec = 0 };
```
avagin: I think the proper fix is to remove const from the declaration of clock_gettime, it should not be there...

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -49,6 +49,11 @@
 #include "cr_options.h"
 #include "util.h"
 
+#if !defined(__GLIBC__)
+/* musl libc does not need 64 suffix */
+#define fseeko64 fseeko
```
avagin: why do we need to use fseeko64?

test/zdtm/static/tempfs_subns.c
```diff
@@ -20,7 +20,7 @@ int main(int argc, char **argv)
 {
 	int fds[2], i;
 	pid_t pid;
-	int fd, status;
+	int fd = -1, status = -1;
```
avagin: you don't need to initialized `status`...

plugins/amdgpu/amdgpu_plugin.c
rst0git: This is related to the problem we fixed recently in https://github.com/checkpoint-restore/criu/pull/2826

avagin: we always compile with _FILE_OFFSET_BITS=64, so even fseeko should handle 64-bit offsets...

rst0git:
> Just in case you have not seen it. Does not sound too difficult to fix.

Thanks! I missed this one. I will open separate pull request for this: https://github.com/checkpoint-restore/criu/pull/2860

criu/pie/restorer.c
rst0git: Removing `const` from the declaration of `clock_gettime` doesn't seem to fix the error. Here is a simple reproducer inside `alpine:latest` container:

```console
# cat test.c 
#include <time.h>

int sys_clock_gettime(clockid_t clk_id, const struct timespec *tp)
{
    return (int)clk_id + (int)tp->tv_nsec;
}

int main(void)
{
    struct timespec ts;

    if (sys_clock_gettime(CLOCK_REALTIME, &ts)) {
        return 1;
    }

    return 0;
}

# clang -Wall -Wextra -Werror test.c
test.c:12:44: error: variable 'ts' is uninitialized when passed as a const pointer argument here [-Werror,-Wuninitialized-const-pointer]
   12 |     if (sys_clock_gettime(CLOCK_REALTIME, &ts)) {
      |                                            ^~
1 error generated.
```

compel/arch/arm/plugins/std/syscalls/syscall.def
```diff
@@ -85,7 +85,7 @@ timer_settime			110	258	(kernel_timer_t timer_id, int flags, const struct itimer
 timer_gettime			108	259	(int timer_id, const struct itimerspec *setting)
 timer_getoverrun		109	260	(int timer_id)
 timer_delete			111	261	(kernel_timer_t timer_id)
-clock_gettime			113	263	(const clockid_t which_clock, const struct timespec *tp)
+clock_gettime			113	263	(clockid_t which_clock, const struct timespec *tp)
```
avagin: you need to remove the second const...

rst0git: Thanks!

PR #2859, opened 2026-01-03
[WIP]: added  support for file descriptor restorsation during restore using  og_fd[N] syntax for --inherit-fd option

### LINKED ISSUE #2837: Feature request: Inherit fd using original fd number
The context: I have a long-running process (that cannot easily be restarted without losing work), and I wanted to compress its output before saving it to disk. So I unwisely ran `$ ./long_running | pv | xz > out.txt.xz`, not knowing that by default, `xz` will buffer several megabytes of data before outputting anything. I want to use CRIU to create a copy of the `xz` process with a `/dev/null` input stream and a new output file, so that the copied process will dump all of its buffered data immediately. In this context, I know for a fact that it's safe to redirect the input and output streams of the `xz` process.

I've found that it is indeed possible to perform this task, with the caveat that I have to run the copied process in a new PID namespace. (In this context, I really would have loved to have #2811, since I also know that single-threaded `xz` doesn't look at or care about its own PID.) But the process is very tedious, since I have to use `stat` to collect the filename or pipe inode behind each fd of the original process:
```
lm978:~/example$ sh test.sh | xz > out.txt.xz & # simple loop program
[1] 2588497
lm978:~/example$ sleep 5 && wc -c out.txt.xz # no output due to buffering
0 out.txt.xz
lm978:~/example$ ps
    PID TTY          TIME CMD
2584379 pts/16   00:00:00 bash
2588496 pts/16   00:00:00 sh
2588497 pts/16   00:00:00 xz
2588833 pts/16   00:00:00 sleep
2588841 pts/16   00:00:00 ps
lm978:~/example$ stat /proc/2588497/fd/0
  File: /proc/2588497/fd/0 -> pipe:[121675510]
  Size: 64          Blocks: 0          IO Block: 1024   symbolic link
Device: 0,25    Inode: 121683652   Links: 1
Access: (0500/lr-x------)  Uid: ( 1000/   lm978)   Gid: ( 1000/   lm978)
Access: 2025-12-02 12:14:56.560796463 -0500
Modify: 2025-12-02 12:14:56.560796463 -0500
Change: 2025-12-02 12:14:56.560796463 -0500
 Birth: -
lm978:~/example$ stat /proc/2588497/fd/1
  File: /proc/2588497/fd/1 -> /home/lm978/example/out.txt.xz
  Size: 64          Blocks: 0          IO Block: 1024   symbolic link
Device: 0,25    Inode: 121683653   Links: 1
Access: (0300/l-wx------)  Uid: ( 1000/   lm978)   Gid: ( 1000/   lm978)
Access: 2025-12-02 12:14:56.560796463 -0500
Modify: 2025-12-02 12:14:56.560796463 -0500
Change: 2025-12-02 12:14:56.560796463 -0500
 Birth: -
lm978:~/example$ stat /proc/2588497/fd/2
  File: /proc/2588497/fd/2 -> /dev/pts/16
  Size: 64          Blocks: 0          IO Block: 1024   symbolic link
Device: 0,25    Inode: 121683654   Links: 1
Access: (0700/lrwx------)  Uid: ( 1000/   lm978)   Gid: ( 1000/   lm978)
Access: 2025-12-02 12:14:56.560796463 -0500
Modify: 2025-12-02 12:14:56.560796463 -0500
Change: 2025-12-02 12:14:56.560796463 -0500
 Birth: -
lm978:~/example$ sudo -s
root:/home/lm978/example# mkdir img
root:/home/lm978/example# criu dump --tree 2588497 --leave-running --images-dir img --shell-job
Warn  (compel/arch/x86/src/lib/infect.c:418): Will restore 2588497 with interrupted system call
root:/home/lm978/example# unshare -p -m --fork --mount-proc
root:/home/lm978/example# criu restore --images-dir img --shell-job --inherit-fd 'fd[3]:pipe:[121675510]' --inherit-fd 'fd[4]:home/lm978/example/out.txt.xz' --inherit-fd 'fd[2]:dev/pts/16' 3< /dev/null 4> out2.txt.xz
root:/home/lm978/example# exit
root:/home/lm978/example# exit
lm978:~/example$ wc -c out.txt.xz out2.txt.xz 
  0 out.txt.xz
232 out2.txt.xz
232 total
lm978:~/example$ xzcat out2.txt.xz | head
Tue Dec  2 12:14:05 PM EST 2025
Tue Dec  2 12:14:06 PM EST 2025
Tue Dec  2 12:14:07 PM EST 2025
Tue Dec  2 12:14:08 PM EST 2025
Tue Dec  2 12:14:09 PM EST 2025
Tue Dec  2 12:14:10 PM EST 2025
Tue Dec  2 12:14:11 PM EST 2025
Tue Dec  2 12:14:12 PM EST 2025
Tue Dec  2 12:14:13 PM EST 2025
Tue Dec  2 12:14:14 PM EST 2025
```

If the original `xz` process had already exited, I'm not sure how I'd even be able to extricate the pipe inode from the image files.

Thus, my **feature request** is syntax for the `--inherit-fd` option to name files and pipes by their fds in the original dumped process. This may be ambiguous in the rare case that a process creates threads without `CLONE_FILES`, but if support for that case is needed, it would only need a "tid:fd" syntax.

BhuvanB404:
Added support for og_fd[N] syntax in --inherit-fd option to allow referencing file descriptors by their original fd number in the dumped process. This eliminates the need to manually extract pipe inodes or file paths using stat commands.

The implementation stores og_fd[N] as literal strings in inh_id and passes them  during matching if found.. This keeps the inherit_fd structure unchanged gives flexibility to use original file descripters or inherited ones.

Example usage:
criu restore -D img --inherit-fd 'fd[5]:og_fd[3]' 5>output2.txt 

This redirects original fd 3 to fd 5 in the restored process.
Pipe implementation for file descriptors made easier.

Works on the example given in issue #2837. 

Fixes: #2837


BhuvanB404:
i used the existing functions and char string used to store fd to provide my implementfor  this feature.  The manual effort of pid copying should be erased now.  Is this acceptable?

BhuvanB404:
@avagin @rst0git , please review

avagin:
How should it work if multiple processes are restored?

BhuvanB404:
I was tunnel visioned when working on this and understanding the codebase, I was trying with only single process examples , sorry.

while thinking of this more and looking at global state of file_desc , I think one approach could be parsing pid for each process from the dumped img, and redirecting specific files. I am not sure how to implement this into this type of parsing because well need to store an extra field inside the file structure. so remembering the process-ids in the img.

I couldnt think of a solution using only  existing  file structs without changing how restore currently works.

While I think adding functionality for multiple processes will lead to backtracking on conveience asked for in the issue , it would still be a good feature for removing manual input for many reroutes. 

please provide  your sugestions.

BhuvanB404:
@avagin , suggestions pls

BhuvanB404:
@avagin @rst0git

rst0git:
@BhuvanB404 Sorry for the delay! The name `og_fd` is confusing to me. What about using `pipe_fd` instead?

For example, CRIU would support `--inherit-fd fd[3]:pipe:[472728]` and  `--inherit-fd fd[3]:pipe_fd:[5]`.

Or if the intention is to support different types of files, perhaps we use just `fd`, e.g., `--inherit-fd fd[3]:fd:[5]`?

BhuvanB404:
@rst0git thank you for the suggestions  :+1: 
For the generic  implementation ``` --inherit-fd fd[3]:fd:[5] ```
Like we take a generic input fd[] and parse it internally? I'll try to implement that. 

Any suggestions for direction to start looking into  for  multiprocesses implementation of this? I can't think of simpler implementation than what I suggested.

PR #2860, opened 2026-01-03
test/libcriu: fix include paths for CRIU headers

rst0git:
The libcriu tests include CRIU headers that are split across "criu/include" and "criu/criu/include". Using only "criu/criu/include" causes CRIU's "linux/openat2.h" to fail to find "common/config.h":
    
      gcc -c test_join_ns.c -iquote ../../../criu -iquote ../../../criu/include -I../../../lib/c/ -I../../../images/ -o test_join_ns.o -Werror
      In file included from /usr/include/bits/fcntl-linux.h:468,
                       from /usr/include/bits/fcntl.h:61,
                       from /usr/include/fcntl.h:35,
                       from test_join_ns.c:3:
      ../../../criu/include/linux/openat2.h:6:10: fatal error: common/config.h: No such file or directory
          6 | #include "common/config.h"
            |          ^~~~~~~~~~~~~~~~~
      compilation terminated.
    
Fixes: f8125b8bef7bf5a7bbaea5e6e1d29578c45bf53d ("Couple of fixes to build and run libcriu tests")

### DIFF
--- a/test/others/libcriu/Makefile
+++ b/test/others/libcriu/Makefile
@@ -1,4 +1,4 @@
-include ../../../../criu/Makefile.versions
+include ../../../Makefile.versions
 
 TESTS += test_sub
 TESTS += test_self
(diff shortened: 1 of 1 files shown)


avagin:
This test should not use only headers from uapi/ directories. In the error message, it is trying to include one of the system header from criu sources, this is definitely wrong.

rst0git:
> This test should not use only headers from uapi/ directories. In the error message, it is trying to include one of the system header from criu sources, this is definitely wrong.

I think you are right - we need to refactor the libcriu tests. They should not include headers from the CRIU sources.

PR #2865, opened 2026-01-16
Make arch shstk unlock use correct pid

Snorch:
In a simple case where the parent process and the child one are in one
pid namespace we can safely use vpid(item) to prace the child. But, for
the cases where the child is a pid namespace init, or the child is put
into external pid namespace, the parent and the child have different pid
namespaces and using pid vpid(item) (which e.g. for init will always be
1 here) to ptrace the child process is inorrect.

Let's use the pid reported to us from clone as it's always the right pid
of the child from the parent's point of view.


mihalicyn (approved):
Good catch! ;)

LGTM

rppt (approved):

rst0git (approved):
LGTM

PR #2867, opened 2026-01-21
fsnotify: Improve error logging and fix memory leaks

3idey:
## What

This PR improves the fsnotify subsystem with better error logging and fixes memory leaks on error paths.

## How

- Add `free_inotify_wd_entries()` and `free_fanotify_mark_entries()` helpers to fix memory leaks on error paths
- Add phase prefixes (dump/restore/collect) to log messages
- Include more context (dev, ino, mnt_id, wd, mask) in error messages
- Add null pointer checks in `get_mark_path()`

## Why

The improved error messages make debugging fsnotify checkpoint/restore failures much easier by providing full context about which inotify/fanotify watch failed and at which phase.

The memory leak fixes ensure that allocated inotify watch descriptor and fanotify mark entries are properly freed even when errors occur during pre-dump operations.

---

- [x] Signed-off-by line included
- [x] Single logical change in one commit

### DIFF
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -527,7 +527,7 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
 		*target = open_handle(s_dev, i_ino, f_handle);
 
 	if (*target < 0) {
-		pr_perror("Unable to open %s", f_handle->path);
+		pr_perror("Unable to open %s", f_handle ? f_handle->path : NULL);
 		goto err;
 	}
 
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -587,17 +587,17 @@ static int restore_one_inotify(int inotify_fd, struct fsnotify_mark_info *info)
 		wd = inotify_add_watch(inotify_fd, path, mask);
 		if (wd < 0) {
 			pr_perror("Can't add watch for 0x%x with 0x%x", inotify_fd, iwe->wd);
-			break;
+			goto err;
 		} else if (wd == iwe->wd) {
 			ret = 0;
 			break;
 		} else if (wd > iwe->wd) {
 			pr_err("Unsorted watch 0x%x found for 0x%x with 0x%x\n", wd, inotify_fd, iwe->wd);
-			break;
+			goto err;
 		}
 
 		if (kdat.has_inotify_setnextwd)
-			return -1;
+			goto err;
 
 		inotify_rm_watch(inotify_fd, wd);
 	}
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -281,7 +281,7 @@ fault:
 	pr_warn("\tHandle 0x%x:0x%lx cannot be opened\n", s_dev, i_ino);
 	irmap_path = irmap_lookup(s_dev, i_ino);
 	if (!irmap_path) {
-		pr_err("\tCan't dump that handle\n");
+		pr_err("\tCan't dump inode handle (dev %#x ino %#lx): no irmap path found\n", s_dev, i_ino);
 		return -1;
 	}
 	path = xstrdup(irmap_path);
@@ -298,16 +298,19 @@ err:
 
 static int check_one_wd(InotifyWdEntry *we)
 {
-	pr_info("wd: wd %#08x s_dev %#08x i_ino %#16" PRIx64 " mask %#08x\n", we->wd, we->s_dev, we->i_ino, we->mask);
+	pr_info("inotify wd %#08x s_dev %#08x i_ino %#16" PRIx64 " mask %#08x\n", we->wd, we->s_dev, we->i_ino, we->mask);
 	pr_info("\t[fhandle] bytes %#08x type %#08x __handle %#016" PRIx64 ":%#016" PRIx64 "\n", we->f_handle->bytes,
 		we->f_handle->type, we->f_handle->handle[0], we->f_handle->handle[1]);
 
 	if (we->mask & KERNEL_FS_EVENT_ON_CHILD)
-		pr_warn_once("\t\tDetected FS_EVENT_ON_CHILD bit "
-			     "in mask (will be ignored on restore)\n");
+		pr_warn_once("\t\tDetected FS_EVENT_ON_CHILD bit in inotify wd %#08x "
+			     "mask %#08x (will be ignored on restore)\n", we->wd, we->mask);
 
-	if (check_open_handle(we->s_dev, we->i_ino, we->f_handle))
+	if (check_open_handle(we->s_dev, we->i_ino, we->f_handle)) {
+		pr_err("Failed to check inotify wd (wd %#08x, dev %#x, ino %#016" PRIx64 ", mask %#x)\n",
+		       we->wd, we->s_dev, we->i_ino, we->mask);
 		return -1;
+	};
 
 	return 0;
 }
@@ -322,7 +325,7 @@ static int dump_one_inotify(int lfd, u32 id, const struct fd_parms *p)
 	if (ret < 0)
 		return -1;
 	else if (ret > 0)
-		pr_warn("The %#08x inotify events will be dropped\n", id);
+		pr_warn("inotify %#08x events will be dropped\n", id);
 
 	ie.id = id;
 	ie.flags = p->flags;
@@ -339,7 +342,7 @@ static int dump_one_inotify(int lfd, u32 id, const struct fd_parms *p)
 	fe.id = ie.id;
 	fe.ify = &ie;
 
-	pr_info("id %#08x flags %#08x\n", ie.id, ie.flags);
+	pr_info("inotify id %#08x flags %#08x\n", ie.id, ie.flags);
 	if (pb_write_one(img_from_set(glob_imgset, CR_FD_FILES), &fe, PB_FILE))
 		goto free;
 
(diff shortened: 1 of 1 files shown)


3idey:
@rst0git can you check this?

criu/fsnotify.c
```diff
@@ -423,52 +456,57 @@ static int dump_one_fanotify(int lfd, u32 id, const struct fd_parms *p)
 	if (ret < 0)
 		return -1;
 	else if (ret > 0)
-		pr_warn("The %#08x fanotify events will be dropped\n", id);
+		pr_warn("dump: fanotify %#08x events will be dropped\n", id);
```
rst0git: I'm not sure if it is a good idea to add `"dump:"` in every log message.
We already use `LOG_PREFIX "fsnotify: "` at the beginning of this file.

criu/fsnotify.c
```diff
@@ -145,6 +165,7 @@ static char *alloc_openable(unsigned int s_dev, unsigned long i_ino, FhEntry *f_
 		suitable_mount_found = 1;
 
 		if (read_fd_link(fd, buf, sizeof(buf)) < 0) {
+			pr_perror("Can't read fd link for dev %#x ino %#lx", s_dev, i_ino);
```
avagin: nit: pr_perror is used only when errno is set (after syscalls). In this case, we need to use pr_err.

3idey:
@avagin @rst0git I just committed the changes you asked any comments ?

3idey:
@avagin @rst0git i did what you asked, any other comments ? and thank you.

rst0git:
@3idey The following contributor guides and blog post provides some useful information on how to separate logical changes into commits, write commit messages, any why this is important:

https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md
https://chris.beams.io/git-commit

While AI chatbots can be useful, the LLM models they provide are not trained or fine-tuned on our code base, guidelines, git history, and previous discussions.

3idey:
I appreciate your efforts @rst0git I've separated each logical change into  it's commit and the commit message is descriptive about what i changed and why.

criu/fsnotify.c
```diff
@@ -445,30 +462,28 @@ static int dump_one_fanotify(int lfd, u32 id, const struct fd_parms *p)
...
-	xfree(fe.mark);
+	free_fanotify_mark_entries(fe.mark, fe.n_mark);
 	return ret;
 }
 
 static int pre_dump_one_fanotify(int pid, int lfd)
 {
 	FanotifyFileEntry fe = FANOTIFY_FILE_ENTRY__INIT;
-	int i;
+	int ret = -1, i;
```
avagin: rename ret to exit_code. It is the common pattern used in many other places.

3idey:
@avagin @rst0git anything else ?

criu/fsnotify.c
```diff
@@ -501,28 +501,44 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
...
-			path = f_handle->path + 1;
+			fpath = f_handle->path + 1;
 
-		pr_debug("\t\tRestore with path hint %d:%s\n", mnt_id, path);
-		*target = openat(mntns_root, path, O_PATH);
-	} else
+		pr_debug("\t\tRestore with path hint %d:%s\n", mnt_id, fpath);
+		*target = openat(mntns_root, fpath, O_PATH);
+	} else {
+		if (!f_handle) {
```
avagin: you can check f_handle once:
```
diff --git a/criu/fsnotify.c b/criu/fsnotify.c
index 63a0dde57..83e07479e 100644
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -517,7 +517,11 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
 
                pr_debug("\t\tRestore %s watch for %#08x:%#016lx (via %s)\n", who, s_dev, i_ino, remap->rpath);
                *target = openat(mntns_root, remap->rpath, O_PATH);
-       } else if (f_handle && f_handle->path) {
+       } else if (!f_handle) {
+               pr_err("Null f_handle for %s mark (dev %#x ino %#lx)\n",
+                      who, s_dev, i_ino);
+               goto err;
+       } else if (f_handle->path) {
                int mntns_root;
                char *fpath = ".";
                uint32_t mnt_id = f_handle->has_mnt_id ? f_handle->mnt_id : -1;
@@ -537,11 +541,6 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
                pr_debug("\t\tRestore with path hint %d:%s\n", mnt_id, fpath);
                *target = openat(mntns_root, fpath, O_PATH);
        } else {
-               if (!f_handle) {
-                       pr_err("Null f_handle for %s mark (dev %#x ino %#lx)\n",
-                              who, s_dev, i_ino);
-                       goto err;
-               }
                *target = open_handle(s_dev, i_ino, f_handle);
        }
 
```

criu/fsnotify.c
```diff
@@ -376,7 +376,7 @@ static int dump_one_inotify(int lfd, u32 id, const struct fd_parms *p)
 static int pre_dump_one_inotify(int pid, int lfd)
 {
 	InotifyFileEntry ie = INOTIFY_FILE_ENTRY__INIT;
-	int ret = -1, i;
+	int exit_code = -1, i;
```
avagin: this change should be merged with the change that introduced this ret

3idey:
all done @avagin

criu/fsnotify.c
```diff
@@ -75,6 +75,26 @@ typedef struct {
 	u64 __handle[16];
 } fh_t;
 
+/* Helper to free inotify watch descriptor entries */
+static void free_inotify_wd_entries(InotifyWdEntry **wd, int n_wd)
```
rst0git: @3idey Would you be able to split this commit into two: one commit for `pre_dump_one_inotify`, and another for `pre_dump_one_fanotify`?

This makes it easier to review the changes.

criu/fsnotify.c
```diff
@@ -501,28 +501,43 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
...
 		pr_debug("\t\tRestore %s watch for %#08x:%#016lx (via %s)\n", who, s_dev, i_ino, remap->rpath);
 		*target = openat(mntns_root, remap->rpath, O_PATH);
+	} else if (!f_handle) {
+		pr_err("Null f_handle for %s mark (dev %#x ino %#lx)\n",
+		       who, s_dev, i_ino);
+		goto err;
 	} else if (f_handle->path) {
 		int mntns_root;
-		char *path = ".";
+		char *fpath = ".";
```
rst0git, line 523: >- Check mntns_get_root_by_mnt_id() return value before using mntns_root
>- Add null check for f_handle before dereferencing
>- Rename local variable 'path' to 'fpath' to avoid shadowing
>- Improve error message for open failure to include device and inode

Could you please move each of the above changes into separate commit?

criu/fsnotify.c
```diff
@@ -592,18 +600,20 @@ static int restore_one_inotify(int inotify_fd, struct fsnotify_mark_info *info)
...
-			pr_err("Unsorted watch 0x%x found for 0x%x with 0x%x\n", wd, inotify_fd, iwe->wd);
-			break;
+			pr_err("Unsorted inotify watch (got wd %#x) while restoring expected wd %#x on fd %#x for path %s\n",
+			       wd, iwe->wd, inotify_fd, path);
+			goto err;
 		}
 
 		if (kdat.has_inotify_setnextwd)
-			return -1;
+			goto err;
```
rst0git, line 609: The majority of this patch aims to improve error messages and doesn't introduce functional changes. If changing `break;` and `return -1;` to `goto err;` is needed, could you please make this change into a separate commit.

criu/fsnotify.c
```diff
@@ -486,28 +509,43 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
...
 		int mntns_root;
-		char *path = ".";
+		char *fpath = ".";
 		uint32_t mnt_id = f_handle->has_mnt_id ? f_handle->mnt_id : -1;
 
 		/* irmap cache is collected in the root namespaces. */
 		mntns_root = mntns_get_root_by_mnt_id(mnt_id);
+		if (mntns_root < 0) {
+			pr_err("Can't get mntns root for %s path hint (mnt_id %d)\n",
+			       who, mnt_id);
```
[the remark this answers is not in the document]
avagin: mnt_id should be int in this case

criu/fsnotify.c
```diff
@@ -511,10 +516,15 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
...
-		uint32_t mnt_id = f_handle->has_mnt_id ? f_handle->mnt_id : -1;
+		int mnt_id = f_handle->has_mnt_id ? (int)f_handle->mnt_id : -1;
 
 		/* irmap cache is collected in the root namespaces. */
 		mntns_root = mntns_get_root_by_mnt_id(mnt_id);
+		if (mntns_root < 0) {
+			pr_err("Can't get mntns root for %s path hint (mnt_id %d)\n",
+			       who, mnt_id);
+			goto err;
+		int mnt_id = f_handle->has_mnt_id ? (int)
```
avagin: what is going on here?

3idey: @avagin That was a mistake in an earlier commit where `mnt_id` was declared after its use, This was fixed in commit 40357e590147348b06c373251e002abe6d119956 by changing the type to int and moving the declaration before the error check. The incomplete line was removed in the next commit (b0f23b77aa2fce4366007c9821bc34af20af3e6d). The code is now correct.

avagin: @3idey it doesn't work this way. Each patch should be self-contained and the code should be functional after each changes.

rst0git: > The incomplete line was removed in the next commit

@3idey You can use `git rebase` to edit previous commits. Andrei is referring to the following part from the contributor guidelines:

>When dividing your change into a series of commits, take special care to ensure that CRIU builds and runs properly after each commit in the series. Developers using git bisect to track down a problem can end up splitting your patch series at any point; they will not thank you if you introduce bugs in the middle.

3idey:
@avagin can you take a final look if you have time ?

PR #2869, opened 2026-01-26
criu: Support binfmt_misc sandboxing

seppd:
The Linux kernel already supports per user namespace sandboxed mounts: 21ca59b365c0 ("binfmt_misc: enable sandboxed mounts")

Add support for this functionality to criu:
 - Drop the BINFMT_MISC_VIRTUALIZED config option
 - Store the binfmt_misc data to the userns image.

The overall dump/restore logic is based on the fact that a binfmt_misc superblock is static and is allocated once per a binfmt_misc mount withing a given user namespace. So the code makes a temporary mount to read/write the entries and then removes it.

### DIFF
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -1810,6 +1810,16 @@ static int check_external_net_ns(void)
 	return 0;
 }
 
+static int check_binfmt_misc_sandboxing(void)
+{
+	if (!kdat.has_binfmt_misc_sandboxing) {
+		pr_info("binfmt_misc sandboxing isn't supported\n");
+		return -1;
+	}
+
+	return 0;
+}
+
 struct feature_list {
 	char *name;
 	int (*func)(void);
@@ -1861,6 +1871,7 @@ static struct feature_list feature_list[] = {
 	{ "overlayfs_maps", check_overlayfs_maps },
 	{ "breakpoints", check_breakpoints },
 	{ "pagemap_scan_guard_pages", check_pagemap_scan_guard_pages },
+	{ "binfmt_misc_sandboxing", check_binfmt_misc_sandboxing },
 	{ NULL, NULL },
 };
 
--- a/criu/filesystems.c
+++ b/criu/filesystems.c
@@ -79,9 +79,8 @@ static int parse_binfmt_misc_entry(struct bfd *f, BinfmtMiscEntry *bme)
 	return 0;
 }
 
-static int __maybe_unused dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
+static int dump_binfmt_misc_entry(int dfd, char *name, BinfmtMiscEntry *bme)
 {
-	BinfmtMiscEntry bme = BINFMT_MISC_ENTRY__INIT;
 	struct bfd f;
 	int ret = -1;
 
@@ -94,62 +93,232 @@ static int __maybe_unused dump_binfmt_misc_entry(int dfd, char *name, struct cr_
 	if (bfdopenr(&f))
 		return -1;
 
-	if (parse_binfmt_misc_entry(&f, &bme))
+	if (parse_binfmt_misc_entry(&f, bme))
 		goto err;
 
-	bme.name = name;
-
-	if (pb_write_one(img, &bme, PB_BINFMT_MISC))
+	bme->name = xstrdup(name);
+	if (!bme->name)
 		goto err;
+
 	ret = 0;
 err:
-	free(bme.interpreter);
-	free(bme.flags);
-	free(bme.extension);
-	free(bme.magic);
-	free(bme.mask);
 	bclose(&f);
 	return ret;
 }
 
-static int write_binfmt_misc_entry(char *mp, char *buf, BinfmtMiscEntry *bme)
+static void free_bme_fields(BinfmtMiscEntry *bme)
+{
+	xfree(bme->name);
+	xfree(bme->interpreter);
+	xfree(bme->flags);
+	xfree(bme->extension);
+	xfree(bme->magic);
+	xfree(bme->mask);
+}
+
+void free_pb_binfmt_misc_entries(BinfmtMiscEntry **bmes, int n)
+{
+	int i;
+
+	if (!bmes || n == 0)
+		return;
+
+	for (i = 0; i < n; i++)
+		free_bme_fields(bmes[i]);
+
+	xfree(bmes[0]);
+	xfree(bmes);
+}
+
+static int do_binfmt_misc_dump(int fd, BinfmtMiscEntry ***pb_bmes)
+{
+	BinfmtMiscEntry *bmes = NULL;
+	struct dirent *de;
+	DIR *fdir = NULL;
+	int i, ret = 0, len = 0, size = 0;
+
+	*pb_bmes = NULL;
+
+	fdir = fdopendir(fd);
+	if (fdir == NULL) {
+		pr_perror("Failed to open mount directory");
+		close(fd);
+		return -1;
+	}
+
+	while ((de = readdir(fdir))) {
+		BinfmtMiscEntry *bme;
+
+		if (dir_dots(de))
+			continue;
+		if (!strcmp(de->d_name, "register"))
+			continue;
+		if (!strcmp(de->d_name, "status"))
+			continue;
+
+		if (len == size) {
+			BinfmtMiscEntry *e;
+
+			size = size * 2 + 1;
+			e = xrealloc(bmes, size * sizeof(BinfmtMiscEntry));
+			if (e == NULL) {
+				pr_perror("Failed to allocate memory for BinfmtMiscEntry");
+				ret = -1;
+				goto close_dir;
+			}
+			bmes = e;
+		}
+
+		bme = &bmes[len];
+		binfmt_misc_entry__init(bme);
+		/*
+		 * Increase len now as dump_binfmt_misc_entry() might fail with
+		 * partially allocated BinfmtMiscEntry fields.
+		 */
+		len++;
+		ret = dump_binfmt_misc_entry(fd, de->d_name, bme);
+		if (ret < 0)
+			goto close_dir;
+	}
+
+	if (len == 0)
+		goto close_dir;
+
+	*pb_bmes = xmalloc(sizeof(BinfmtMiscEntry *) * len);
+	if (*pb_bmes == NULL) {
+		pr_perror("Failed to allocate memory for BinfmtMiscEntry pointers");
+		ret = -1;
+		goto close_dir;
+	}
+
+	for (i = 0; i < len; i++)
+		(*pb_bmes)[i] = &bmes[i];
+
+close_dir:
+	closedir(fdir);
+
+	if (ret >= 0)
+		return len;
+
+	for (i = 0; i < len; i++)
+		free_bme_fields(&bmes[i]);
+	xfree(bmes);
+
+	return -1;
+}
+
+struct binfmt_misc_dump_arg {
+	pid_t pid;
+	BinfmtMiscEntry ***bmes;
+	int n;
+};
+
+static int binfmt_misc_dump_from_child(void *arg)
+{
+	int exit_code = 1, fd, mnt_fd, n;
+	struct binfmt_misc_dump_arg *dump_arg = (struct binfmt_misc_dump_arg *)arg;
+	BinfmtMiscEntry ***bmes = dump_arg->bmes;
+
+	if (switch_mnt_ns(dump_arg->pid, NULL, NULL) < 0) {
+		pr_err("Failed to switch mount namespace\n");
+		return 1;
+	}
+
+	if (switch_ns(dump_arg->pid, &user_ns_desc, NULL) < 0) {
+		pr_err("Failed to switch user namespace\n");
+		return 1;
+	}
+
+	mnt_fd = mount_detached_fs("binfmt_misc");
+	if (mnt_fd < 0)
+		return 1;
+
+	fd = openat(mnt_fd, ".", O_DIRECTORY | O_RDONLY);
+	if (fd < 0) {
+		pr_perror("Failed to open mountpoint");
+		goto close_mnt_fd;
+	}
+
+	n = do_binfmt_misc_dump(fd, bmes);
+	if (n < 0) {
+		pr_err("Failed to dump binfmt_misc contents\n");
+		goto close_mnt_fd;
+	}
+
+	dump_arg->n = n;
+	exit_code = 0;
+
+close_mnt_fd:
+	close(mnt_fd);
+
+	return exit_code;
+}
+
+/*
+ * The function allocates memory for pb_bmes. It's up to the caller to free it.
+ *
+ * Returns the number of dumped entries or -1 on error.
+ */
+int binfmt_misc_dump_sandboxed(pid_t pid, BinfmtMiscEntry ***pb_bmes)
+{
+	struct binfmt_misc_dump_arg dump_arg;
+	BinfmtMiscEntry **bmes = NULL;
+
+	if (!kdat.has_binfmt_misc_sandboxing)
+		return 0;
+
+	if (!(root_ns_mask & CLONE_NEWUSER)) {
+		pr_err("PID %i is not in a sandbox\n", pid);
+		return -1;
+	}
+
+	dump_arg.pid = pid;
+	dump_arg.bmes = &bmes;
+
+	if (call_in_child_process(binfmt_misc_dump_from_child, (void *)&dump_arg) < 0) {
+		pr_err("Failed to dump binfmt_misc contents from child process\n");
+		return -1;
+	}
+
+	*pb_bmes = bmes;
+	return dump_arg.n;
+}
+
+static int write_binfmt_misc_entry(int mnt_fd, char *buf, BinfmtMiscEntry *bme)
 {
 	int fd, len, ret = -1;
-	char path[PATH_MAX + 1];
 
-	snprintf(path, PATH_MAX, "%s/register", mp);
-
-	fd = open(path, O_WRONLY);
+	fd = openat(mnt_fd, "register", O_WRONLY);
 	if (fd < 0) {
-		pr_perror("binfmt_misc: can't open %s", path);
+		pr_perror("binfmt_misc: can't open 'register'");
 		return -1;
 	}
 
 	len = strlen(buf);
 
 	if (write(fd, buf, len) != len) {
-		pr_perror("binfmt_misc: can't write to %s", path);
+		pr_perror("binfmt_misc: can't write to 'register");
 		goto close;
 	}
 
 	if (!bme->enabled) {
 		close(fd);
-		snprintf(path, PATH_MAX, "%s/%s", mp, bme->name);
 
-		fd = open(path, O_WRONLY);
+		fd = openat(mnt_fd, bme->name, O_WRONLY);
 		if (fd < 0) {
-			pr_perror("binfmt_misc: can't open %s", path);
+			pr_perror("binfmt_misc: can't open %s", bme->name);
 			goto out;
 		}
 		if (write(fd, "0", 1) != 1) {
-			pr_perror("binfmt_misc: can't write to %s", path);
+			pr_perror("binfmt_misc: can't write to %s", bme->name);
 			goto close;
 		}
 	}
 
-	ret = 0;
 close:
 	close(fd);
+
+	ret = 0;
 out:
 	return ret;
 }
@@ -192,7 +361,7 @@ static int make_bfmtm_magic_str(char *buf, BinfmtMiscEntry *bme)
 	return 1;
 }
 
-static int __maybe_unused binfmt_misc_restore_bme(struct mount_info *mi, BinfmtMiscEntry *bme, char *buf)
+static int binfmt_misc_restore_bme(int mnt_fd, BinfmtMiscEntry *bme, char *buf)
 {
 	int ret;
 
@@ -215,13 +384,60 @@ static int __maybe_unused binfmt_misc_restore_bme(struct mount_info *mi, BinfmtM
 		goto bad_dump;
 
 	pr_debug("binfmt_misc_pattern=%s\n", buf);
-	ret = write_binfmt_misc_entry(service_mountpoint(mi), buf, bme);
+	ret = write_binfmt_misc_entry(mnt_fd, buf, bme);
 
 	return ret;
 
 bad_dump:
 	pr_perror("binfmt_misc: bad dump");
 	return -1;
+}
+
+/*
+ * The function is expected to be called from the 'pid' context with namespaces
+ * already set up so no namespace switching is required here.
+ */
+int binfmt_misc_restore_sandboxed(pid_t pid, BinfmtMiscEntry **bmes, size_t n)
+{
+	int ret, mnt_fd;
+	size_t i;
+	char *buf;
+
+	if (!(root_ns_mask & CLONE_NEWUSER)) {
+		pr_err("PID %i is not in a sandbox\n", pid);
+		return -1;
+	}
+
+	if (n == 0)
+		return 0;
+
+	if (!kdat.has_binfmt_misc_sandboxing) {
+		pr_err("binfmt_misc sandboxing is not supported\n");
+		return -1;
+	}
+
+	mnt_fd = mount_detached_fs("binfmt_misc");
+	if (mnt_fd < 0)
+		return -1;
+
+	buf = xmalloc(BINFMT_MISC_STR);
+	if (!buf) {
+		ret = -1;
+		goto close_mnt_fd;
+	}
+
+	for (i = 0; i < n; i++) {
+		ret = binfmt_misc_restore_bme(mnt_fd, bmes[i], buf);
+		if (ret < 0)
+			break;
+	}
+
+	xfree(buf);
+
+close_mnt_fd:
+	close(mnt_fd);
+
+	return ret;
 }
 
 static int tmpfs_dump(struct mount_info *pm)
--- a/criu/filesystems.c
+++ b/criu/filesystems.c
@@ -32,25 +32,6 @@ static int attach_option(struct mount_info *pm, char *opt)
 	return pm->options ? 0 : -1;
 }
 
-#ifdef CONFIG_BINFMT_MISC_VIRTUALIZED
-struct binfmt_misc_info {
-	BinfmtMiscEntry *bme;
-	struct list_head list;
-};
-
-LIST_HEAD(binfmt_misc_list);
-
-static int binfmt_misc_parse_or_collect(struct mount_info *pm)
-{
-	opts.has_binfmt_misc = true;
-	return 0;
-}
-
-static int binfmt_misc_virtual(struct mount_info *pm)
-{
-	return kerndat_fs_virtualized(KERNDAT_FS_STAT_BINFMT_MISC, pm->s_dev);
-}
-
 static int parse_binfmt_misc_entry(struct bfd *f, BinfmtMiscEntry *bme)
 {
 	while (1) {
@@ -98,7 +79,7 @@ static int parse_binfmt_misc_entry(struct bfd *f, BinfmtMiscEntry *bme)
 	return 0;
 }
 
-static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
+static int __maybe_unused dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
 {
 	BinfmtMiscEntry bme = BINFMT_MISC_ENTRY__INIT;
 	struct bfd f;
@@ -128,62 +109,6 @@ err:
 	free(bme.magic);
 	free(bme.mask);
 	bclose(&f);
-	return ret;
-}
-
-static int binfmt_misc_dump(struct mount_info *pm)
-{
-	static bool dumped = false;
-	struct cr_img *img = NULL;
-	struct dirent *de;
-	DIR *fdir = NULL;
-	int fd, ret;
-
-	ret = binfmt_misc_virtual(pm);
-	if (ret <= 0)
-		return ret;
-
-	if (dumped) {
-		pr_err("Second binfmt_misc superblock\n");
-		return -1;
-	}
-	dumped = true;
-
-	fd = open_mountpoint(pm);
-	if (fd < 0)
-		return fd;
-
-	fdir = fdopendir(fd);
-	if (fdir == NULL) {
-		close(fd);
-		return -1;
-	}
-
-	ret = -1;
-	while ((de = readdir(fdir))) {
-		if (dir_dots(de))
-			continue;
-		if (!strcmp(de->d_name, "register"))
-			continue;
-		if (!strcmp(de->d_name, "status"))
-			continue;
-
-		if (!img) {
-			/* Create image only if an entry exists, i.e. here */
-			img = open_image(CR_FD_BINFMT_MISC, O_DUMP);
-			if (!img)
-				goto out;
-		}
-
-		if (dump_binfmt_misc_entry(fd, de->d_name, img))
-			goto out;
-	}
-
-	ret = 0;
-out:
-	if (img)
-		close_image(img);
-	closedir(fdir);
 	return ret;
 }
 
@@ -267,7 +192,7 @@ static int make_bfmtm_magic_str(char *buf, BinfmtMiscEntry *bme)
 	return 1;
 }
 
-static int binfmt_misc_restore_bme(struct mount_info *mi, BinfmtMiscEntry *bme, char *buf)
+static int __maybe_unused binfmt_misc_restore_bme(struct mount_info *mi, BinfmtMiscEntry *bme, char *buf)
 {
 	int ret;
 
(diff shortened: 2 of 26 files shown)


criu/filesystems.c
```diff
@@ -153,40 +236,222 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+		}
+
+		n = do_binfmt_misc_dump(fd, &bmes);
+		if (n < 0) {
+			pr_err("Failed to dump binfmt_misc contents\n");
+			ret = -1;
+			goto unmount;
+		}
+
+		if (write(sock_pair[1], &n, sizeof(n)) != sizeof(n)) {
```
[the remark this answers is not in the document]
seppd: Not correct. If the child returns an error the parrent won't try to read anything by skipping the read call via a goto jump.

criu/filesystems.c
```diff
@@ -113,31 +111,116 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+		ret = dump_binfmt_misc_entry(fd, de->d_name, bme);
+		if (ret < 0)
+			goto close_dir;
+	}
+
+	if (len == 0)
+		goto close_dir;
+
+	*pb_bmes = xmalloc(sizeof(BinfmtMiscEntry *) * len);
+	if (*pb_bmes == NULL) {
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -153,40 +236,222 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+
+close_parent_sock:
+	if (close(sock_pair[0]) < 0) {
+		pr_perror("Failed to close parents's socket");
+		ret = -1;
+	}
+
+	*pb_bmes = bmes;
+
+	return ret >= 0 ? n : -1;
```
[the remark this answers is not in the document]
seppd: Not correct. 
1. Child exit status is handled by `if (!WIFEXITED(status)) -> goto close_parent_sock;` 
2. `pb_read_one` returns 1 on success so `ret > 0` is a valid state to still retrun `n` (even if `n` itself is 0).

criu/filesystems.c
```diff
@@ -349,6 +622,66 @@ static int binfmt_misc_restore(struct mount_info *mi)
...
 
+/*
+ * The function is expected to be called from the 'pid' context with namespaces
+ * already set up so no namespace switching is required here.
+ */
+int binfmt_misc_restore_sandboxed(pid_t pid, BinfmtMiscEntry **bmes, size_t n)
+{
+	int ret;
+	size_t i;
+	struct mount_info mi;
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -153,40 +236,222 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+			if (pb_write_one(&img, bmes[i], PB_BINFMT_MISC) < 0) {
+				pr_perror("Failed to write BinfmtMiscEntry %i", i);
+				ret = -1;
+				break;
+			}
+		}
+
+free_pb_bmes:
+		for (i = 0; i < n; i++)
+			free_bme_fields(bmes[i]);
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -153,40 +236,222 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+			xfree(bmes);
+			bmes = NULL;
+			ret = -1;
+			break;
+		}
+	}
+
+close_parent_sock:
+	if (close(sock_pair[0]) < 0) {
+		pr_perror("Failed to close parents's socket");
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -113,31 +111,116 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+		if (!strcmp(de->d_name, "status"))
+			continue;
+
+		if (len == size) {
+			BinfmtMiscEntry *e;
+
+			size = size * 2 + 1;
+			e = xrealloc(bmes, size * sizeof(BinfmtMiscEntry));
+			if (e == NULL)
+				break;
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -153,40 +236,222 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+	}
+
+	if (socketpair(PF_UNIX, SOCK_STREAM, 0, sock_pair) < 0) {
+		pr_perror("Failed to create a socket pair");
+		return -1;
+	}
+
+	child_pid = fork();
+	if (child_pid == -1) {
+		pr_perror("Unable to fork a child");
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -153,40 +236,222 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
-	if (img)
-		close_image(img);
-	closedir(fdir);
+
+close_image:
+	close_image(img);
+
+free_pb_bmes:
+	for (i = 0; i < n; i++)
+		free_bme_fields(pb_bmes[i]);
```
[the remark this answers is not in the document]
seppd: Fixed.

criu/filesystems.c
```diff
@@ -153,40 +241,230 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+	child_pid = fork();
+	if (child_pid == -1) {
+		pr_perror("Unable to fork a child");
+		if (close(sock_pair[0] < 0))
+			pr_perror("Failed to close parent's socket");
+		if (close(sock_pair[1]) < 0)
+			pr_perror("Failed to close child's socket");
+		return -1;
+	}
+	if (child_pid == 0) {
```
avagin: you can consider moving a child code to a new helper

criu/filesystems.c
```diff
@@ -153,40 +241,230 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+
+		ret = switch_ns(pid, &user_ns_desc, NULL);
+		if (ret < 0) {
+			pr_err("Failed to switch user namespace\n");
+			goto close_child_sock;
+		}
+
+		if (setuid(0) < 0) {
+			pr_perror("Failed to setuid(0)");
+			goto close_child_sock;
```
avagin: It is unclear what the return code will be here. ret isn't explicitly set and it is probably zero from the previous call. We can follow our standard pattern of using `exit_code`: initialize it to -1 and only set it to 0 (or a success value) at the end of the logic.

criu/filesystems.c
```diff
@@ -153,40 +241,230 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+
+	if (!WIFEXITED(status)) {
+		pr_err("Child process did not exit normally\n");
+		ret = -1;
+		goto close_parent_sock;
+	}
+
+	ret = WEXITSTATUS(status);
+
+	if (read(sock_pair[0], &n, sizeof(n)) != sizeof(n)) {
```
avagin: You should read from the socket before calling wait. Otherwise, the child process may block on the write system call if the socket buffer becomes full.

criu/filesystems.c
```diff
@@ -153,40 +241,230 @@ static int binfmt_misc_dump(struct mount_info *pm)
...
+		if (setuid(0) < 0) {
+			pr_perror("Failed to setuid(0)");
+			goto close_child_sock;
+		}
+		if (setgid(0) < 0) {
+			pr_perror("Failed to setgid(0)");
+			goto close_child_sock;
+		}
+
+		dir = mkdtemp(template);
```
avagin: mkdtemp is called in the dumped mount namespace, isn't it? Is it really required? We can't be sure that /tmp will be writable there? Can we use a new mount api to create a detached tree? Cc: @Snorch

criu/filesystems.c
seppd: I see your point but I don't think this pattern fits well here. The cleanup code must be executed either way, it does not matter if we are in the error state or not. If we go the `initialize it to -1` way we won't have any means to figure out when to set `ret` to 0.

I think I instead should properly set `ret` to -1 in the highlighted code, I indeed missed that.

criu/filesystems.c
seppd: > mkdtemp is called in the dumped mount namespace, isn't it? Is it really required?

Yes, it is called in the dumped mount namespace. Do you mean we should probably create the target directory before switching the mount namespace?

Snorch: I think it should be possible to use anonymous (not shown in mount namespace) mounts via mount_detached_fs("binfmt_misc") for binfmt.

And yes we should try to avoid modifying dumped mount namespace and the view of the file system when possible.

Snorch:
I think we should cleanup binfmt_misc fstype in fstypes array in criu/filesystems.c, we don't need parse/collect/dump/restore per-mount dumping as binfmt now goes to userns image.

criu/filesystems.c
seppd: Done

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: I updated the code to use `mount_detached_fs()` instead.

criu/filesystems.c
avagin: > I think I instead should properly set ret to -1 in the highlighted code, I indeed missed that.

This pattern was introduced to avoid this sort of issues... I think it fits well. Please take a look at the following patch.

```
diff --git a/criu/filesystems.c b/criu/filesystems.c
index 6cf65e9a2..98b258a8f 100644
--- a/criu/filesystems.c
+++ b/criu/filesystems.c
@@ -199,63 +199,53 @@ err:
 
 static int binfmt_misc_dump_from_child(pid_t pid, int sock_pair[2])
 {
-	int ret = 0, fd, mnt_fd, i, n;
-	struct cr_img img;
+	int exit_code = -1, ret, fd, mnt_fd, i, n;
 	BinfmtMiscEntry **bmes = NULL;
+	struct cr_img img;
 
-	ret = close(sock_pair[0]);
-	if (ret < 0) {
+	if (close(sock_pair[0])) {
 		pr_perror("Failed to close parent's socket end");
 		goto close_child_sock;
 	}
 
-	ret = switch_mnt_ns(pid, NULL, NULL);
-	if (ret < 0) {
+	if (switch_mnt_ns(pid, NULL, NULL)) {
 		pr_err("Failed to switch mount namespace\n");
 		goto close_child_sock;
 	}
 
-	ret = switch_ns(pid, &user_ns_desc, NULL);
-	if (ret < 0) {
+	if (switch_ns(pid, &user_ns_desc, NULL)) {
 		pr_err("Failed to switch user namespace\n");
 		goto close_child_sock;
 	}
 
-	ret = setuid(0);
-	if (ret < 0) {
+	if (setuid(0)) {
 		pr_perror("Failed to setuid(0)");
 		goto close_child_sock;
 	}
 
-	ret = setgid(0);
-	if (ret < 0) {
+	if (setgid(0)) {
 		pr_perror("Failed to setgid(0)");
 		goto close_child_sock;
 	}
 
 	mnt_fd = mount_detached_fs("binfmt_misc");
-	if (mnt_fd < 0) {
-		ret = -1;
+	if (mnt_fd < 0)
 		goto close_child_sock;
-	}
 
 	fd = openat(mnt_fd, ".", O_DIRECTORY | O_RDONLY);
 	if (fd < 0) {
 		pr_perror("Failed to open mountpoint");
-		ret = -1;
 		goto close_mnt_fd;
 	}
 
 	n = do_binfmt_misc_dump(fd, &bmes);
 	if (n < 0) {
 		pr_err("Failed to dump binfmt_misc contents\n");
-		ret = -1;
 		goto close_mnt_fd;
 	}
 
 	if (write(sock_pair[1], &n, sizeof(n)) != sizeof(n)) {
 		pr_perror("Failed to write number of BinfmtMiscEntry");
-		ret = -1;
 		goto free_pb_bmes;
 	}
 
@@ -264,10 +254,10 @@ static int binfmt_misc_dump_from_child(pid_t pid, int sock_pair[2])
 	for (i = 0; i < n; i++) {
 		if (pb_write_one(&img, bmes[i], PB_BINFMT_MISC) < 0) {
 			pr_perror("Failed to write BinfmtMiscEntry %i", i);
-			ret = -1;
 			break;
 		}
 	}
+	exit_code = 0;
 
 free_pb_bmes:
 	for (i = 0; i < n; i++)
@@ -279,16 +269,16 @@ free_pb_bmes:
 close_mnt_fd:
 	if (close(mnt_fd) < 0) {
 		pr_perror("Failed to close file system fd");
-		ret = -1;
+		exit_code = -1;
 	}
 
 close_child_sock:
 	if (close(sock_pair[1]) < 0) {
 		pr_perror("Failed to close child's socket");
-		ret = -1;
+		exit_code = -1;
 	}
 
-	return ret;
+	return exit_code;
 }
 
 /*
```

criu/filesystems.c
avagin: do we still need to change a mount namespace?

criu/filesystems.c
```diff
@@ -113,118 +93,343 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+		pr_perror("Unable to fork a child");
+		if (close(sock_pair[0]) < 0)
+			pr_perror("Failed to close parent's socket");
+		if (close(sock_pair[1]) < 0)
+			pr_perror("Failed to close child's socket");
+		return -1;
+	}
+	if (child_pid == 0) {
+		ret = binfmt_misc_dump_from_child(pid, sock_pair);
+		exit(ret);
```
avagin: In this case, we need to be sure that `ret` isn't negative. Passing a negative code to `exit()` isn't a good practice.

criu/filesystems.c
```diff
@@ -113,118 +93,343 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+		ret = -1;
+	} else if (!WIFEXITED(status)) {
+		pr_err("Child process did not exit normally\n");
+		ret = -1;
+	} else if (ret != -1) {
+		ret = WEXITSTATUS(status) == 0 ? 0 : -1;
+	}
 
-	fd = open(path, O_WRONLY);
+	if (close(sock_pair[0]) < 0) {
```
avagin: we need to close the socket before waitpid(). Otherwise, the child can block if the parent hasn't read all data from the socket (e.g. in case of any error).

criu/filesystems.c
```diff
@@ -113,118 +93,343 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
 
-	snprintf(path, PATH_MAX, "%s/register", mp);
+	bmes = xmalloc(sizeof(BinfmtMiscEntry *) * n);
+	if (!bmes) {
+		ret = -1;
+		goto wait_child;
+	}
+
+	memset(&img, 0, sizeof(img));
+	img._x.fd = sock_pair[0];
```
avagin: nit: you can use img_from_fd here.

criu/filesystems.c
seppd: It seems so, yes. Otherwise I get `Operation not permitted` on `fsopen()`.

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Thanks, fixed

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Updated

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+	for (i = 0; i < len; i++)
+		free_bme_fields(&bmes[i]);
+	xfree(bmes);
+
+	return -1;
+}
+
+static int binfmt_misc_dump_from_child(pid_t pid, int sock_pair[2])
+{
+	int ret = -1, fd, mnt_fd, i, n;
```
avagin: Please use `exit_code` instead of `ret`. As I mentioned, exit_code is a specific variable name; anyone working on CRIU code will recognize it immediately. In this context, using ret is misleading.

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+	if (socketpair(PF_UNIX, SOCK_STREAM, 0, sock_pair) < 0) {
+		pr_perror("Failed to create a socket pair");
+		return -1;
+	}
+
+	child_pid = fork();
+	if (child_pid == -1) {
+		pr_perror("Unable to fork a child");
+		if (close(sock_pair[0]) < 0)
+			pr_perror("Failed to close parent's socket");
```
avagin: nit: We usually don't check close() return codes unless a failure could actually lead to data corruption.

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+	img = img_from_fd(sock_pair[0]);
+	if (!img) {
+		pr_err("Failed to create parent image from fd\n");
+		goto out;
+	}
+
+	for (i = 0; i < n; i++) {
+		if (pb_read_one(img, &bmes[i], PB_BINFMT_MISC) <= 0) {
+			pr_perror("Failed to read BinfmtMiscEntry %i", i);
+			break;
```
avagin: ret will be set to 0...

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+		}
+	}
+
+	xfree(img);
```
avagin: xfree should be in the `out` section. You need to initialize it to NULL, so xfree will do nothing if it hasn't been allocated.

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+	if (close(sock_pair[0]) < 0) {
+		pr_perror("Failed to close parent's socket");
+		ret = -1;
+	}
+
+	if (waitpid(child_pid, &status, 0) < 0) {
+		pr_perror("Failed to wait for child process");
+		ret = -1;
+	} else if (!WIFEXITED(status)) {
+		pr_err("Child process did not exit normally\n");
```
avagin: Let's print status in the error message:
		pr_err("Child process did not exit normally (status=0x%x)\n", status);

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+		ret = -1;
+	} else if (ret != -1) {
+		ret = WEXITSTATUS(status) == 0 ? 0 : -1;
```
avagin: nit: we may want to print a error if ret is set to -1.

criu/filesystems.c
```diff
@@ -113,118 +93,346 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+
+	img = img_from_fd(sock_pair[0]);
+	if (!img) {
+		pr_err("Failed to create parent image from fd\n");
+		goto out;
+	}
+
+	for (i = 0; i < n; i++) {
+		if (pb_read_one(img, &bmes[i], PB_BINFMT_MISC) <= 0) {
+			pr_perror("Failed to read BinfmtMiscEntry %i", i);
```
avagin: nit: pr_err should be used here. pb_read_one doesn't set errno.

Snorch:
I also did some fixes for binfmt_misc test here https://github.com/Snorch/criu/tree/binfmt-misc-test-fixes , @seppd please push them to your PR.

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Makes sense, fixed

criu/filesystems.c
seppd: Updated

criu/filesystems.c
seppd: Updated (though I haven't seen any code using `exit_code` while working on this patch)

seppd:
@Snorch I've experienced issues with your binfmt_misc test fixes likely caused by randomization so I haven't included them so far. E.g.:
```
./zdtm.py run -t zdtm/static/binfmt_misc
userns is supported
=== Run 1/1 ================ zdtm/static/binfmt_misc
====================== Run zdtm/static/binfmt_misc in uns ======================
Start test
Test is SUID
Running zdtm/static/binfmt_misc.hook(--post-start)
./binfmt_misc --pidfile=binfmt_misc.pid --outfile=binfmt_misc.out --dirname=binfmt_misc.dir.test --filename=binfmt_misc.test
The test failed (0, 1)
make: *** [Makefile:533: binfmt_misc.pid] Error 1
 Test zdtm/static/binfmt_misc FAIL at ['make', '--no-print-directory', '-C', 'zdtm/static', 'binfmt_misc.pid'] 
09:36:50.110:     4: string: :binfmt_misc.test_magic:M:43:\x96\x01\xfd\x32\x90\x7c\x17\xc5\x7a\x69::/bin/interpreter:OCP
09:36:50.110:     4: string: :binfmt_misc.test_extension:E::lqw2>        ^OCD▒o/U
                                                                                       jön▒gۓ
09:36:50.110:     4: FAIL: binfmt_misc.c:144: write binfmt_misc.test_extension (errno = 22 (Invalid argument))
09:36:50.111:     3: ERR: test.c:320: Test exited unexpectedly with code 1

Running zdtm/static/binfmt_misc.hook(--clean)
Cleaning binfmt_misc.test: --clean
```
This issue has some probabilistic nature and reproduces on roughly every other run.

seppd:
I pushed the test related commits from @Snorch but w/o the one initializing the random seed.

criu/filesystems.c
avagin: > Updated (though I haven't seen any code using exit_code while working on this patch)

```
$ git grep "exit_code" | wc -l
266
```

@seppd I really don't understand all these push-backs. If you think the exit_code pattern doesn't solve the problem, you can always describe your thoughts and we are open to discuss them.

seppd: That was not meant as any kind of push-back. I first thought you had renamed the variable in the example patch just to make it clearer. And as this var name was not really used in the code I touched (namespaces.c, filesystems.c etc) I thought that was something opinion based and felt reluctant doing extra rebases merely to rename a variable. If it's a rule - no problem.

avagin:
@Snorch please review this pr.


> I pushed the test related commits from @Snorch but w/o the one initializing the random seed.

I have no idea what it means. @Snorch this is on you too.

Snorch:
1) I would prefer if we had a kdat/check key for binfmt virtualization, so that we can detect that per-userns virtualization of binfmt_misc is available on current kernel or not, and make feature in criu and tests depend on it. -> @seppd  Please add separate patch which adds this detection to kdat and criu-check.

2) Seed:

> @Snorch please review this pr.
> 
> > I pushed the test related commits from @Snorch but w/o the one initializing the random seed.
> 
> I have no idea what it means. @Snorch this is on you too.

I think that random seed thing should be a matter of another PR, because we never really had real random there as seed was always 0, and the data generated with seed zero works fine with binfmt, but almost any other really random data breaks things () (and we would need time to figure out why and what are real restrictions on data we can write into binfmt). I tested it couple of times on my system and was "lucky" enough to miss that it breaks... Let's drop this patch for now and think of it later.

3) I see more places where we should remove old binfmt_misc code, e.g. you stop using kerndat_fs_virtualized() -> get_host_dev() function for binfmt, but it still has binftm related code inside, it would be nice to also drop it. 

upd: Also for removal: mount_cr_time_mount(), clean_cr_time_mounts(), add_cr_time_mount(), populate_mnt_ns(cr_time) argument and related code, and FD_ENTRY(BINFMT_MISC*, binfmt_misc_list.

4) I think it would be easier to review if you split the removal of old code (leaving some useful helpers to be reused later, you can add __maybe_unused attribute for them to prevent compilation errors) from adding the new userns based code. This way it would be much easier to verify that we only leave what will be reused. Else we might miss something else similar to 3).

5) Instead of using heavy fork() + sending the result through socket pair, I think it would be easier to use call_in_child_process() helper which creates a child with memory shared with parent so you can put result directly to parent's memory.

criu/filesystems.c
```diff
@@ -113,118 +93,321 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+
+	img = img_from_fd(sock_pair[1]);
+	if (!img) {
+		pr_err("Failed to create child image from fd\n");
+		exit_code = -1;
+		goto free_pb_bmes;
+	}
+
+	for (i = 0; i < n; i++) {
+		if (pb_write_one(img, bmes[i], PB_BINFMT_MISC) < 0) {
```
Snorch: This image_from_fd + pb_write_one + pb_read_one logic to pass data between criu processes does not look to be a correct design. We use protobuf to put(/get) data to(/from) images on dump(/restore), not to pass it around.

criu/filesystems.c
```diff
@@ -113,118 +93,321 @@ static int dump_binfmt_misc_entry(int dfd, char *name, struct cr_img *img)
...
+
+	img = img_from_fd(sock_pair[0]);
+	if (!img) {
+		pr_err("Failed to create parent image from fd\n");
+		goto out;
+	}
+
+	for (i = 0; i < n; i++) {
+		if (pb_read_one(img, &bmes[i], PB_BINFMT_MISC) <= 0) {
+			pr_err("Failed to read BinfmtMiscEntry %i", i);
```
Snorch: In pr_err() you need to add explicit "\n". Only helpers which implicitly print errno do not need explicit "\n". You can catch such errors with "make lint".

seppd:
I pushed an update based on the @Snorch 's comments

Snorch:
I filled the issue for improving random in binfmt_misc test https://github.com/checkpoint-restore/criu/issues/2886

seppd:
I updated the PR as per an off-github converstion with @Snorch :
1. Rebase to criu-dev
2. Use `__maybe_unused` instead of `__attribute__((unused))`
3. Some `ret` -> `exit_code` substitutions

seppd:
Updated:

1. Added `'feature': 'binfmt_misc_sandboxing'` to the test descriptor
2. Treat all `mount_detached_fs` errors as `has_binfmt_misc_sandboxing == false`

criu/kerndat.c
```diff
@@ -1848,6 +1848,82 @@ void kerndat_warn_about_madv_guards(void)
...
+	int exit_code = 0, mnt_fd;
+
+	mnt_fd = mount_detached_fs("binfmt_misc");
+	if (mnt_fd < 0) {
+		/* Most likely a permission error, so we can't use binfmt_misc */
+		return 1;
+	}
+
+	if (fstat(mnt_fd, st) < 0)
+		exit_code = -1;
```
avagin: should we log a error?

criu/cr-check.c
```diff
@@ -1810,6 +1810,16 @@ static int check_external_net_ns(void)
 	return 0;
 }
 
+static int check_binfmt_misc_sandboxing(void)
+{
+	if (!kdat.has_binfmt_misc_sandboxing) {
+		pr_warn("binfmt_misc sandboxing isn't supported\n");
```
avagin: Why is it a warning?

test/zdtm/static/binfmt_misc_ns.desc
```diff
@@ -0,0 +1 @@
+{'flavor': 'ns', 'flags': 'excl suid noauto'}
```
avagin, line 1: please add a comment why it is noauto and why it "should fail"

avagin:
> I filled the issue for improving random in binfmt_misc test #2886

I attached a patch in a comment, it probably will fix the issue. Could you test it and add to this series if it works?

test/zdtm/static/binfmt_misc_ns.desc
seppd: @Snorch

criu/cr-check.c
seppd: Because all other functions I had a chance to observe within this file report a negative check result as `warn`.

Snorch: Please make it pr_info.

In other cases with pr_warn, what we want to say is that without this feature CRIU might dump something incorrectly (partially), without binfmt_misc we just don't dump binfmt_misc cause there is nothing to dump, it should be completely safe -> hence no need for warning.

test/zdtm/static/binfmt_misc_ns.desc
Snorch: Without userns there is no binfmt_misc dumping -> binfmt_misc should not be present after restore (due to cleanup test hook which cleans it on pre-restore).

@avagin I just don't know how to add it to CI, do we, by chance, have a way to run this test and handle it's fail "Test zdtm/static/binfmt_misc_ns FAIL at result check" as a PASS? I see that "crfail" option only allows fail on c/r, but not on result check. Do we add a separate zdtm option for that or maybe we already have something?

Snorch:
> > I filled the issue for improving random in binfmt_misc test #2886
> 
> I attached a patch in a comment, it probably will fix the issue. Could you test it and add to this series if it works?

Yes that seems to do the trick. The length of extension and magic should probably be non-zero and extension characers should not exceed 256 explicitly (though sprintf probably wraps them anyway), looks good.

test/zdtm/static/binfmt_misc_ns.desc
Snorch: JFYI: I want to add this _ns failing test because previously we had binftm_misc test running in ns flavor with no error, due to missing cleanup hook before restore, this _ns test basically verifies that cleanup hook works.

seppd:
Updates:
 - Resolved comments
 - Applied https://github.com/checkpoint-restore/criu/issues/2886#issuecomment-3886024672 to fix random patterns in the `binfmt_misc` test
 - Enabled the `binfmt_misc` test in CI

test/zdtm/static/binfmt_misc_ns.desc
seppd: Updated

criu/kerndat.c
seppd: Updated

criu/kerndat.c
```diff
@@ -1853,6 +1848,92 @@ void kerndat_warn_about_madv_guards(void)
...
+		has_arg->has = false;
+		return 0;
+	}
+
+	if (has_arg->st_dev != st.st_dev)
+		has_arg->has = true;
+	else
+		has_arg->has = false;
+
+	close(mnt_fd);
```
Snorch, line 1902: I suggest using https://bitbucket.org/openvz/criu.ovz/commits/72bed0cadb4c91a3c9eb55cc6625e2f7076b2b01 cleanup_fd attribute. Like:

```
int cleanup_fd fd = -1;
```
Else you forget to close it on reurn-s above.

seppd:
Updates:
 - Drop `Makefile.orig` that has accidentally made it into a commit
 - Fetch a correct `make the random generation actually random` commit

criu/kerndat.c
Snorch: Ok I'm wrong in returns above we are ok, no leak.

seppd:
Updates:
 - `pr_warn` -> `pr_info` in `check_binfmt_misc_sandboxing()`

Snorch:
I moved kerndat related fixup misplaced in criu: Support binfmt_misc sandboxing to kerndat: add kerndat_has_binfmt_misc_sandboxing(). (no functional change)

Snorch (approved):
I added a small hunk to make binfmt_misc_ns (expected to fail) to be actually checked in CI:
```
@@ -239,6 +239,12 @@ LAZY_OPTS=(-p 2 -T "$LAZY_TESTS" "${LAZY_EXCLUDE[@]}" "${ZDTM_OPTS[@]}")
 ./test/zdtm.py run "${LAZY_OPTS[@]}" --remote-lazy-pages
 ./test/zdtm.py run "${LAZY_OPTS[@]}" --remote-lazy-pages --tls
 
+# This test is expected o fail as without uesrns binfmt misc is not c/r-ed
+if ./test/zdtm.py run -t zdtm/static/binfmt_misc_ns; then
+       echo "Unexpected pass of zdtm/static/binfmt_misc_ns"
+       exit 1
+fi
+
```
LGTM. (I also checked that test errors seem unrelated.)

avagin:
I've merged everything except [zdtm: Add binfmt_misc test without userns (should fail)](https://github.com/checkpoint-restore/criu/pull/2869/changes/55ff410467ac2a43161bce42806651d76c0fa51c).

seppd:
Closed as all of the main commits of the PR have been merged.

PR #2876, opened 2026-02-05
fsnotify: Fix file descriptor leak in restore_one_inotify()

3idey:
When restoring inotify watches, the restore_one_inotify() function opens a file descriptor via get_mark_path() and stores it in the 'target' variable. However, in two error paths, the function returns directly without properly closing this file descriptor:

1. When INOTIFY_IOC_SETNEXTWD ioctl fails (line 556)
2. When has_inotify_setnextwd is true but watch descriptor mismatch occurs (line 577)

Both cases bypass the cleanup code at the 'err' label which calls close_safe(&target), resulting in a file descriptor leak.

Fix this by using 'goto err' instead of 'return -1' in both error paths, ensuring proper cleanup of the target file descriptor.


3idey:
I'm sorry for the misunderstanding @avagin @rst0git    
I closed this pr #2867 and i will make a single pr for each change contains the related commits.
thank you.

PR #2878, opened 2026-02-06
zdtm: add test for UDP multicast socket membership restoration

### LINKED ISSUE #2873: Missing ZDTM coverage for UDP multicast sockets
Hello, so while testing and reviewing the ZDTM Test Suite for CRIU, I noticed that there is no regression tests for UDP Multicast (e.g., sockets using IP_ADD_MEMBERSHIP).

Is it since some inherent feature of CRIU doesnt support Multicast dump/restore or just not added yet?

CRIU already has extensive coverage for UDP sockets in general, but multicast group membership represents kernel-maintained socket state that may be lost across checkpoint/restore without explicit handling.

So that might mean, that in case of a process which uses Multicast sockets, post restore the multicast funtionality might be lost and we might not be able to restore the membership addresses the way we want to as per original process. 

This kind of failure might be difficult to detect without active explicit tests, or prior knowledge of the process workings.
I couldn’t find any tests or documentation for this.

As per the understanding, if something isnt being tested by ZDTM then CRIU prolly doesnt support it, please let me know if I'm missing something.

euphoricair7:
This PR adds a test case for UDP multicast socket membership restoration,
documenting that CRIU currently does not support restoring IP_ADD_MEMBERSHIP
socket options.

### Summary

While reviewing the ZDTM Test Suite, I noticed there are no tests for UDP
multicast sockets. Multicast group memberships (IP_ADD_MEMBERSHIP) are
kernel-maintained socket state that may be lost across checkpoint/restore
without explicit handling.

### Changes

1. **New test**: `test/zdtm/static/socket_udp_multicast.c`
   - Creates a UDP socket that joins a multicast group (224.0.0.250)
   - Tests communication after checkpoint/restore
   - Demonstrates that multicast membership is not preserved

2. **Test descriptor**: `test/zdtm/static/socket_udp_multicast.desc`
   - Marks the test as broken (`"broken": true`)
   - Documents the limitation in the `notes` field

3. **Makefile update**: Added build rule for the new test

### Findings

Through analysis of CRIU's source code, I confirmed that:
- No protobuf fields exist for multicast memberships in `sk-inet.proto`
- No dump/restore code handles `IP_ADD_MEMBERSHIP` in `sk-inet.c`
- The test currently fails as expected, demonstrating the gap

The test is designed to:
1. Join a multicast group before checkpoint
2. Attempt to receive multicast packets after restore
3. Time out (fail) since CRIU doesn't restore the membership

### Related Issue

Fixes: #2873 (Missing ZDTM coverage for UDP multicast sockets)

### Next Steps

This test serves as:
1. Documentation of the current limitation
2. A starting point for implementing multicast support
3. A test case to verify when support is added

To implement multicast support, the following would be needed:
- Add multicast membership fields to protobuf definitions
- Implement dump/restore of IP_ADD_MEMBERSHIP in sk-inet.c
- Potentially handle other multicast options (IP_MULTICAST_IF, etc.)

### Testing

Test can be run with:
```bash
sudo ./test/zdtm.py run -t zdtm/static/socket_udp_multicast

### CODE REVIEW

File: test/zdtm/static/socket_udp_multicast.c [hunk 1]
```diff
@@ -0,0 +1,154 @@
...
+	if (setsockopt(recv_sk, SOL_SOCKET, SO_RCVTIMEO, &tv, sizeof(tv)) < 0) {
+		pr_perror("Can't set receive timeout");
+		return 1;
+	}
+
+	test_daemon();
+	test_waitsig();
+
+	/* Small delay to ensure everything is ready */
+	usleep(100000);
```
avagin, line 112: why do we need this sleep?

test/zdtm/static/socket_udp_multicast.desc
```diff
@@ -0,0 +1,10 @@
+{
+  "comment": "Tests UDP multicast socket membership restoration. Currently expected to fail because CRIU doesn't restore IP_ADD_MEMBERSHIP socket option.",
+  "author": "Shailja Shaktawat",
+  "needs": ["ipv4"],
+  "runs": ["h"],
+  "timeout": 10,
+  "broken": True,
+  "requires": "loopback interface with multicast enabled",
+  "notes": "Multicast group memberships (IP_ADD_MEMBERSHIP) are not currently preserved across checkpoint/restore. This test documents this limitation."
+}
```
avagin: What AI agent have you used to write this file?

avagin:
This pr breaks CI tests.

test/zdtm/static/socket_udp_multicast.c
euphoricair7: i added the usleep delay such that more of for the observability while the test runs and eventually fails for now, but when the udp multicast might be added eventually in the future, just to ensure that no packets are dropped. 
But if needed we can totally remove/comment it for now, since it doens't affect the current functionality test at all.

test/zdtm/static/socket_udp_multicast.desc
euphoricair7: Well, this is my own mistake, i apologise. I had simply searched up .desc format on google and edited it as per my content. I failed to edit the data format for json correctly.  

I just now compared my .desc files with the other .desc files in the zdtm suite and realised the format mistakes, I will modify it as per the  other files.

euphoricair7:
> This pr breaks CI tests.

I will apply some changes of .desc file and try once again.

euphoricair7:
Resolved it.

euphoricair7:
@avagin all the CI tests except, Cirrus CI / Vagrant Fedora Rawhide based test, failed, is this a test something I should worry about?

euphoricair7:
> @avagin all the CI tests except, Cirrus CI / Vagrant Fedora Rawhide based test, failed, is this a test something I should worry about?

Here the test is failing on zombie00 with a segfault. This seems unrelated to my UDP changes. Could you verify?

PR #2880, opened 2026-02-07
fsnotify: Improve error messages with more context

3idey:
## What
Improve error messages in the fsnotify subsystem with additional context.

## How
Enhanced error messages to include:
- Device and inode numbers for file handles
- Watch descriptor (wd) for inotify operations  
- Mask values for fanotify marks
- Mount ID for mount-related operations

## Why
Makes debugging checkpoint/restore failures much easier by providing complete information about which watch or mark failed without needing additional debug logging.

### DIFF
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -102,7 +102,11 @@ static int open_by_handle(void *arg, int fd, int pid)
 	return syscall(__NR_open_by_handle_at, fd, arg, O_PATH);
 }
 
-enum { ERR_NO_MOUNT = -1, ERR_NO_PATH_IN_MOUNT = -2, ERR_GENERIC = -3 };
+enum {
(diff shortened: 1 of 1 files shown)


criu/fsnotify.c
```diff
@@ -454,21 +471,31 @@ static int dump_one_fanotify(int lfd, u32 id, const struct fd_parms *p)
...
+			       me->s_dev, me->ie->i_ino, me->mask);
+			goto out;
+		}
 
 		xfree(me);
 	}
+
+	ret = 0;
+out:
+	for (; i < fe.n_mark; i++)
```
avagin: this should be done in another patch. It isn't about error messages.

3idey:
Done @avagin. Removed the memory leak cleanup code from this PR - will submit it as a separate patch focused on memory management.

PR #2881, opened 2026-02-08
fsnotify: Fix memory leaks in pre_dump error paths

3idey:
## What

Fix memory leaks in `pre_dump_one_inotify` and `pre_dump_one_fanotify` error paths.

## How

- Add cleanup labels to ensure proper deallocation on error
- Free remaining entries and arrays when `irmap_queue_cache` fails

## Why

When `irmap_queue_cache` fails during pre-dump, the remaining watch descriptors/mark entries and their arrays are not freed, causing memory leaks.

## Testing

- [x]   inotify00 - Basic inotify test
- [x]   fanotify00 - Basic fanotify test  
- [x]   inotify_irmap - Tests irmap_queue_cache error path

### DIFF
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -364,7 +364,7 @@ free:
 static int pre_dump_one_inotify(int pid, int lfd)
 {
 	InotifyFileEntry ie = INOTIFY_FILE_ENTRY__INIT;
-	int i;
+	int i, ret = -1;
 
 	if (parse_fdinfo_pid(pid, lfd, FD_TYPES__INOTIFY, &ie)) {
 		pr_err("Failed to parse fdinfo for inotify (pid %d lfd %d)\n", pid, lfd);
@@ -377,13 +377,18 @@ static int pre_dump_one_inotify(int pid, int lfd)
 		if (irmap_queue_cache(we->s_dev, we->i_ino, we->f_handle)) {
 			pr_err("Failed to queue irmap cache for inotify wd %#x (dev %#x ino %#" PRIx64 ")\n",
 			       we->wd, we->s_dev, we->i_ino);
-			return -1;
+			goto err;
 		}
 
 		xfree(we);
 	}
 
-	return 0;
+	ret = 0;
+err:
+	for (; i < ie.n_wd; i++)
+		xfree(ie.wd[i]);
+	xfree(ie.wd);
+	return ret;
 }
 
 const struct fdtype_ops inotify_dump_ops = {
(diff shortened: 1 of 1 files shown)


criu/fsnotify.c
```diff
@@ -377,13 +377,18 @@ static int pre_dump_one_inotify(int pid, int lfd)
 		if (irmap_queue_cache(we->s_dev, we->i_ino, we->f_handle)) {
 			pr_err("Failed to queue irmap cache for inotify wd %#x (dev %#x ino %#" PRIx64 ")\n",
 			       we->wd, we->s_dev, we->i_ino);
-			return -1;
+			goto out;
```
rst0git: We usually use `out` for early exit and `err` in for error handling - `goto err`?

criu/fsnotify.c
```diff
@@ -377,13 +377,18 @@ static int pre_dump_one_inotify(int pid, int lfd)
...
 		}
 
 		xfree(we);
 	}
 
-	return 0;
+	ret = 0;
+out:
+	for (; i < ie.n_wd; i++)
```
rst0git, line 388: `i` is also used in the loop above. Should we set `i=0` here?

criu/fsnotify.c
3idey: It should be free then not out
Thank you.

3idey: The label name 'free' is appropriate here since the cleanup path is specifically freeing memory allocations. Both success and error paths need to free the ie.wd array, while only the error path needs to free remaining entries.

criu/fsnotify.c
3idey: thank you. i think we should not set i=0. The continuation loop `for (; i < ie.n_wd; i++)` correctly handles both cases:
- error: i is at the failing entry, so the loop frees remaining entries from i to n_wd-1 (entries 0 to i-1 were already freed in the main loop)
- success: i == n_wd, so the loop doesn't run - only the array itself is freed

Setting i=0 would cause double-free of entries 0 to i-1 that were already freed in the main loop.

criu/fsnotify.c
rst0git: `free` is the name of a standard library function and using it as a `goto` label will be confusing.

`err` is a conventional `goto` label that is used throughout our codebase. It is appropriate to be used here because we handle an error, as indicated by the error message above.

3idey: I appreciate your patient.
I've updated both commits to use `err:` instead of `free:`

3idey:
@rst0git @avagin is there anything missing or needs to be fixed here ?

PR #2883, opened 2026-02-10
fsnotify: Check mntns_get_root_by_mnt_id() return value in get_mark_path()

3idey:
When `mntns_get_root_by_mnt_id()` fails and returns a negative value, it is used directly in `openat()` calls without checking, leading to potential invalid file descriptor usage.

Add error checks after `mntns_get_root_by_mnt_id()` calls to ensure the returned file descriptor is valid before using it. Add error logging to provide diagnostic context about which mount namespace and path were being accessed when the failure occurs.

Also fix variable shadowing by renaming the inner scope path variable to `fpath`, avoiding confusion with the outer scope path variable.

Additionally, fix `mnt_id` type from `uint32_t` to int. Using `uint32_t` with -1 sentinel causes the value to wrap to `UINT32_MAX` , and the conversion to int when passed to `mntns_get_root_by_mnt_id(int)` is implementation-defined, potentially causing lookup failures and `BUG_ON()` triggers.

### DIFF
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -529,12 +529,12 @@ static char *get_mark_path(const char *who, struct file_remap *remap, FhEntry *f
 	} else if (f_handle->path) {
 		int mntns_root;
 		char *fpath = ".";
-		uint32_t mnt_id = f_handle->has_mnt_id ? f_handle->mnt_id : -1;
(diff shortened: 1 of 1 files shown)


3idey:
@avagin @rst0git could you please review this ?
Thank you.

Issue #47, opened 2015-09-23
Support undo semaphores (sysvipc)


xemul:
Pulled from criu.org/Todo

These are SysVIPC objects created with semctl() and SEM_UNDO flag. Shame on us, we don't even detect these are created. Fortunately they are not in active use. Need to do it -- dump and restore. Requires modifications from both sides — criu and kernel.

PR #2889, opened 2026-02-12
Fix lease type corruption in restore_breaking_file_lease() due to fd_types misuse

WHOIM1205:
# Fix: Incorrect Lease Type Restoration in `restore_breaking_file_lease()`

## Overview

This PR fixes a critical lease restoration bug in CRIU that could cause:

- Restore failures
- Silent state corruption
- Incorrect kernel I/O blocking behavior

The issue was caused by passing a `fd_types` enum value where file open flags were expected during lease restoration.

---

## Problem

In `criu/file-lock.c`, the function:

### DIFF
--- a/test/zdtm/static/file_lease05.c
+++ b/test/zdtm/static/file_lease05.c
@@ -1,0 +1,89 @@
+#include <fcntl.h>
+#include <limits.h>
+#include <signal.h>
+#include <stdio.h>
+#include <sys/wait.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check broken read-lease is restored with correct type";
+const char *test_author = "CRIU";
+
+char *filename;
+TEST_OPTION(filename, string, "file name", 1);
+
+static void break_sigaction(int signo, siginfo_t *info, void *ctx)
+{
+}
+
+int main(int argc, char **argv)
+{
+	struct sigaction act = {};
+	int fd;
+
+	test_init(argc, argv);
+
+	act.sa_sigaction = break_sigaction;
+	act.sa_flags = SA_SIGINFO | SA_RESTART;
+	if (sigaction(SIGIO, &act, NULL)) {
+		pr_perror("Can't set signal action");
+		return 1;
+	}
+
+	fd = open(filename, O_RDONLY | O_CREAT, 0666);
+	if (fd < 0) {
+		pr_perror("Can't open file");
+		return 1;
+	}
+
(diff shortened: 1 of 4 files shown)


WHOIM1205:
@rst0git @avagin 
This PR fixes a lease restoration bug in `restore_breaking_file_lease()` where a `fd_types` enum value was incorrectly passed as file open flags.

Because `FD_TYPES__REG` numerically matches `O_WRONLY`, broken read leases (`F_RDLCK`) were always restored as write leases (`F_WRLCK`). This could cause restore failures (`EAGAIN` / `EBUSY`) or silent lease type corruption.

The fix replaces the enum with the actual stored open flags (`fdle->fe->flags`), ensuring the correct lease type is restored.

A ZDTM test (`file_lease05.c`) has been added to verify the fix.

test/zdtm/static/file_lease05.c
```diff
@@ -0,0 +1,121 @@
+#include <fcntl.h>
```
avagin: ERROR: DOS line endings
#47: FILE: test/zdtm/static/file_lease05.c:1:
+#include <fcntl.h>^M$

avagin:
> Because `FD_TYPES__REG` numerically matches `O_WRONLY`, broken read leases (`F_RDLCK`) were always restored as write leases (`F_WRLCK`). This could cause restore failures (`EAGAIN` / `EBUSY`) or silent lease type corruption.
> 
> The fix replaces the enum with the actual stored open flags (`fdle->fe->flags`), ensuring the correct lease type is restored.

You need to update the commit message with all this information.

> 
> A ZDTM test (`file_lease05.c`) has been added to verify the fix.

The new test should be introduced in a separate commit. Please try to run you test with and without the fix. It has to fail without the fix and pass with the fix.

WHOIM1205:
Thanks for the review @avagin 

I have addressed the feedback:

 Split the changes into two commits (fix and test separately).
 Updated the fix commit message with detailed root cause and impact explanation.
 Converted the test files to LF line endings.
 Verified that the new test fails without the fix and passes with the fix.

Please let me know if anything else needs adjustment.

avagin:
@WHOIM1205 please look at ci test failures

avagin:
> Verified that the new test fails without the fix and passes with the fix.

This test passes without the fix in my environment. How do you run it?

avagin:
@WHOIM1205 I've slightly changed your patches. Could you please review them and let me know what you think?

PR #2890, opened 2026-02-13
compel/x86: probe kernel task_size at runtime

### LINKED ISSUE #2877: VDSO and VVAR are not properly unmapped before remapping during task restore
**Description**: During the restore operation, the unmap_old_vmas() function is responsible for clearing the process address space. However, it relies on kdat.task_size as the upper bound for unmapping.

As captured in the restorer logs, on certain environments (e.g., specific kernel configurations or 5-level paging systems), **the vDSO and vVAR segments are mapped at addresses exceeding the standard task_size boundary** (e.g., mapped at 0x8000... while task_size is 0x7ffffffff000). Consequently, these segments are not unmapped during the cleanup phase and remain in the address space, causing conflicts when map_vdso is subsequently called.

**Evidence** (Restorer Logs): The log shows that unmap_old_vmas only covers ranges up to 0x7ffffffff000, but vvar and vdso persist at higher addresses:

```(00.101841) pie: 11: Unmapping range: 0x7fffe3787000 - 0x7ffffffff000

(00.101893) pie: 11: ========== /proc/self/maps (after unmap_old_vmas) ==========

...

8000000c1000-8000000c5000 r--p 00000000 00:00 0 [vvar]

8000000c5000-8000000c7000 r-xp 00000000 00:00 0 [vdso]
```

TusharMohapatra07:
Change compel_task_size() to probe the actual kernel address space
limit at runtime via munmap(), matching the approach used by aarch64
architecture.

On systems where the CPU supports 5-level paging but the kernel
doesn't have it enabled, we need to detect the actual task size the
kernel is using.

The function compel_task_size() now iterates from TASK_SIZE (2 ^ 47)
upward, doubling each time, until munmap() fails, indicating the
kernel doesn't support address that high.

Fixes #2877

### DIFF
--- a/compel/arch/x86/src/lib/infect.c
+++ b/compel/arch/x86/src/lib/infect.c
@@ -17,6 +17,7 @@
 #include <compel/plugins/std/syscall-codes.h>
 #include <compel/plugins/std/syscall.h>
 #include "common/err.h"
+#include "common/page.h"
 #include "asm/infect-types.h"
 #include "ptrace.h"
 #include "infect.h"
@@ -738,7 +739,9 @@ int ptrace_set_regs(pid_t pid, user_regs_struct_t *regs)
 	return ptrace(PTRACE_SETREGSET, pid, NT_PRSTATUS, &iov);
 }
 
-#define TASK_SIZE ((1UL << 47) - PAGE_SIZE)
+#define TASK_SIZE_47 ((1UL << 47) - PAGE_SIZE)
+#define TASK_SIZE_56 ((1UL << 56) - PAGE_SIZE)
+
 /*
  * Task size may be limited to 3G but we need a
  * higher limit, because it's backward compatible.
@@ -747,7 +750,23 @@ int ptrace_set_regs(pid_t pid, user_regs_struct_t *regs)
 
 unsigned long compel_task_size(void)
 {
-	return TASK_SIZE;
+	void *addr;
+
+	/*
+	 * On x86-64, task_size is either (1UL << 47) for 4-level paging
+	 * or (1UL << 56) for 5-level paging.
+	 */
+	addr = mmap((void *)(TASK_SIZE_47), page_size(), PROT_NONE,
+		    MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED_NOREPLACE,
+		    -1, 0);
+	if (addr != MAP_FAILED) {
+		munmap(addr, page_size());
+		return TASK_SIZE_56;
+	}
+	if (errno == EEXIST)
+		return TASK_SIZE_56;
+
+	return TASK_SIZE_47;
 }
 
 bool __compel_shstk_enabled(user_fpregs_struct_t *ext_regs)


TusharMohapatra07:
Hey Maintainers! 
I have tested the changes by emulating a 5-level paging system using qemu. The zdtm tests related to vdso (vdso00, vdso01 etc.)  all passed successfully.

criu/pie/restorer.c
```diff
@@ -1816,6 +1852,14 @@ __visible long __export_restore_task(struct task_restore_args *args)
 			   args->task_size))
 		goto core_restore_end;
 
+	/*
+	 * Unmap runtime vdso/vvar if they exist above task_size.
```
avagin: If they exist above task_size, so task_size is incorrect....

TusharMohapatra07: Thanks for the review! 
I’m not sure I fully understand your point. Would you prefer that I reword the comment, or should i resize the `task_size` ?

TusharMohapatra07: i used this approach as the issue reporter mentioned that changing task_size didn't work.  My thought behind the approach was to explicitly unmap vdso/var before the restorer starts mapping vdso that isn't parked (for ref: https://github.com/checkpoint-restore/criu/issues/2877#issuecomment-3894598137)

TusharMohapatra07: also, the task size is wrong because `compel_task_size()` is hardcoded to `(1 << 47) - PAGE_SIZE`. So an alternative way of approaching the issue would be to define task_size for 5 level systems and return that [here](https://github.com/checkpoint-restore/criu/blob/b5d531eb4f92ddec796065631d8948c55d69ec9f/compel/arch/x86/src/lib/infect.c#L750) instead. 
Do you suggest this approach over the above ?

avagin: @TusharMohapatra07 We don't make changes without fully understanding exactly what is going on. In this case, we need to set a proper task_size when 5-level paging is enabled. If the issue is still reproducible after that, we will need to investigate and find the root cause.

TusharMohapatra07: Gotcha! will change the implementation. One final question - Do you have any suggestions for testing ? I previously used qemu to emulate a 5-level paging system and ran the zdtm tests as my current workstation supports 4-level only.

avagin: We can try to ask the reporter to test a change when it will be ready.

TusharMohapatra07: cool! Thanks :)

compel/arch/x86/src/lib/infect.c
```diff
@@ -747,7 +750,14 @@ int ptrace_set_regs(pid_t pid, user_regs_struct_t *regs)
 
 unsigned long compel_task_size(void)
 {
-	return TASK_SIZE;
+	unsigned long task_size;
+
+	for (task_size = TASK_SIZE; task_size < TASK_SIZE_MAX;
+	     task_size <<= 1)
+		if (munmap((void *)task_size, page_size()))
```
avagin: This is dangerous. It can unmap a valid mapping

avagin:
> Signed-off-by: TusharMohapatra07 <tusharmohapatra.gig@gmail.com>

Please use a real name.

TusharMohapatra07:
@avagin I've introduced guardrails using mmap(), now we only  unmap when mmap succeds. Please take a look and let me know what you think about the approach

question: I also considered parsing `/proc/self/maps` to calculate the max address and determine `task_size`. Which approach do you think would be better ?

compel/arch/x86/src/lib/infect.c
```diff
@@ -747,7 +750,25 @@ int ptrace_set_regs(pid_t pid, user_regs_struct_t *regs)
 
 unsigned long compel_task_size(void)
 {
-	return TASK_SIZE;
+	unsigned long task_size;
+
+	for (task_size = TASK_SIZE; task_size < TASK_SIZE_MAX;
+	     task_size <<= 1) {
```
avagin: task_size can be just (1UL << 47) or (1UL << 56).

TusharMohapatra07: refactored to remove the loop:  https://github.com/checkpoint-restore/criu/commit/92925d59290ce0bb7db45de98b8a876e1184b838

compel/arch/x86/src/lib/infect.c
```diff
@@ -747,7 +750,23 @@ int ptrace_set_regs(pid_t pid, user_regs_struct_t *regs)
...
 unsigned long compel_task_size(void)
 {
-	return TASK_SIZE;
+	void *addr;
+
+	/*
+	 * On x86-64, task_size is either (1UL << 47) for 4-level paging
+	 * or (1UL << 56) for 5-level paging.
+	 */
+	addr = mmap((void *)(TASK_SIZE_47), page_size(), PROT_NONE,
```
TusharMohapatra07, line 759: Hey @avagin! I had actually thought this through in my previous commit, please correct me if i'm wrong, but `TASK_SIZE_47` defines the last valid page in the user address space so mmap might always succeed and we might incorrectly return `TASK_SIZE_MAX`. In order to probe at the 4-level boundary, i had explicitly set it to (1UL << 47). 

For reference: https://github.com/checkpoint-restore/criu/commit/92925d59290ce0bb7db45de98b8a876e1184b838#diff-e09d70f38ddfdaa51646acb31723ec7ad582fa8f282c28c41a952a4d3b500e68R759

Please let me know what you think :)

avagin: > Hey @avagin! I had actually thought this through in my previous commit, please correct me if i'm wrong, but `TASK_SIZE_47` defines the last valid page in the user address space 

It should not be right. TASK_SIZE_47 is the size of a task virtual address space. Basically, it is the first virtual address that can't be mapped.

TusharMohapatra07: Gotcha! Thanks for clarifying :)

PR #2892, opened 2026-02-14
dump: Fix uninitialized thread leader vtid breaking posix timer dump

### LINKED ISSUE #2887: Thread leader has invalid vtid leading to failing timer restore
I'm investigating an error during restore where criu fails to restore a timer in some hard to reproduce cases. I think I narrowed it down to the `notify_thread_id` being wrongly stored on dump but only if the timer was created by the thread leader. For debugging I added this to print out the `vtid` of all threads before dumping the timers:

```diff
diff --git a/criu/cr-dump.c b/criu/cr-dump.c
index 60b8e793c..d8ac53aa8 100644
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -1553,6 +1553,15 @@ err_cure:
        goto err_free;
 }

+static void print_virts(struct pstree_item *item)
+{
+       int i;
+       pr_info("print vtdis for %d threads\n", item->nr_threads);
+       for (i = 0; i < item->nr_threads; i++) {
+               pr_info("thread %d vtid: %d\n", i, item->threads[i].ns[0].virt);
+       }
+}
+
 static int dump_one_task(struct pstree_item *item, InventoryEntry *parent_ie)
 {
        pid_t pid = item->pid->real;
@@ -1726,6 +1735,7 @@ static int dump_one_task(struct pstree_item *item, InventoryEntry *parent_ie)
                goto err_cure;
        }

+       print_virts(item);
        ret = parasite_dump_posix_timers_seized(&proc_args, parasite_ctl, item);
        if (ret) {
                pr_err("Can't dump posix timers (pid: %d)\n", pid);
```

(on top of tag `v4.2`)

This gives me the following output when dumping my affected program:

```
(00.020882) print vtdis for 8 threads
(00.020884) thread 0 vtid: -282857328
(00.020885) thread 1 vtid: 7
(00.020886) thread 2 vtid: 8
(00.020886) thread 3 vtid: 9
(00.020887) thread 4 vtid: 10
(00.020888) thread 5 vtid: 11
(00.020889) thread 6 vtid: 12
(00.020890) thread 7 vtid: 130
```

The thread 0 vtid is clearly wrong, I guess unitialized? This basically works with any process I dump (runc containers to be more specific), the thread 0 always has an invalid vtid. If this thread has a timer (this is where I struggle a bit more to reproduce it), it will result in the timer getting that id so in my case this ends up in the image:
```
$ cat /var/lib/zeropod/i/0226c640a15bc40ce4f8fce83b4849bced6ea235bd04331a1f0e1a855fe602d3/snapshot/core-129.img | crit decode | jq .
{
  "magic": "CORE",
  "entries": [
    {
      "mtype": "AARCH64",
      "tc": {
        "task_state": 1,
        "exit_code": 0,
        "personality": 0,
        "flags": 4194560,
        "blk_sigset": 0,
        "comm": "php",
        "timers": {
          "real": {
            "isec": 0,
            "iusec": 0,
            "vsec": 0,
            "vusec": 0
          },
          "virt": {
            "isec": 0,
            "iusec": 0,
            "vsec": 0,
            "vusec": 0
          },
          "prof": {
            "isec": 0,
            "iusec": 0,
            "vsec": 0,
            "vusec": 0
          },
          "posix": [
            {
              "it_id": 0,
              "clock_id": 7,
              "si_signo": 34,
              "it_sigev_notify": 4,
              "sival_ptr": 187651547268680,
              "overrun": 0,
              "isec": 0,
              "insec": 0,
              "vsec": 0,
              "vnsec": 0,
              "notify_thread_id": -282844592
            }
          ]
        },
[...]
```

When restoring this, I get:

```
(00.119789) pie: 129: Error (criu/pie/restorer.c:1293): Can't create posix timer - 0: -22
(00.119792) pie: 129: Error (criu/pie/restorer.c:2201): Can't restore posix timers -1
(00.119800) pie: 129: Error (criu/pie/restorer.c:2332): Restorer fail 129
```

I think the root cause is somewhere here https://github.com/checkpoint-restore/criu/blob/b5d531eb4f92ddec796065631d8948c55d69ec9f/criu/seize.c#L904 and here https://github.com/checkpoint-restore/criu/blob/b5d531eb4f92ddec796065631d8948c55d69ec9f/criu/seize.c#L856
The first thread will not have the `vtid` set as it happens to the other threads. I would open a PR but I'm not sure how to even get the correct `vtid` for the thread leader there.

3idey:
Fixes #2887

## Problem

When a multithreaded process has a POSIX timer created with SIGEV_THREAD_ID notification targeting the thread leader, CRIU dump fails with:

```
Error (criu/timer.c:329): Unable to convert the notify thread id <pid>
Can't dump posix timers (pid: <pid>)
Dumping FAILED.
```

If the dump somehow completed (e.g. in older versions where the vtid happened to be zero and didn't match), the garbage notify_thread_id value (e.g. -282857328) gets written into the image and timer_create() fails with -EINVAL on restore.

## Root Cause

There is a vtid initialization gap for the thread leader in the dump path:

1. collect_threads() in seize.c initializes threads[0] for the thread leader but only sets .real and .item — it never sets .ns[0].virt:

```c
if (item->nr_threads == 0) {
    item->threads[0].real = item->pid->real;
    item->nr_threads = 1;
    item->threads[0].item = NULL;
    /* .ns[0].virt left uninitialized! */
}
```

Non-leader threads correctly get their vtid set:

```c
item->threads[item->nr_threads].ns[0].virt = t_creds.s.vpid;
```

2. dump_one_task() sets item->pid->ns[0].virt = misc.pid but does not propagate this to threads[0].ns[0].virt.

3. parasite_dump_posix_timers_seized() is called next, which calls encode_notify_thread_id(). This function iterates item->threads[] looking for a matching .real pid and reads its .ns[0].virt. For the thread leader, it finds threads[0] but reads garbage because the vtid was never initialized.

4. dump_task_threads() finally sets threads[0].ns[0].virt = vpid(item) — but this runs after the timer dump, so it is too late.

## Fix

Initialize threads[0].ns[0].virt immediately after item->pid->ns[0].virt is set, before the posix timer dump:

```diff
 item->pid->ns[0].virt = misc.pid;
+item->threads[0].ns[0].virt = misc.pid;
 pstree_insert_pid(item->pid);
```

## New Test

Added zdtm/static/leader_posix_timer — creates a multithreaded process where the thread leader sets up a SIGEV_THREAD_ID timer targeting itself. After C/R, verifies the timer still fires and has the correct interval.

## Test Output — Before Fix (FAIL)

```
=== Run 1/1 ================ zdtm/static/leader_posix_timer
=================== Run zdtm/static/leader_posix_timer in ns ===================
Start test
./leader_posix_timer --pidfile=leader_posix_timer.pid --outfile=leader_posix_timer.out
Run criu dump
=[log]=> dump/zdtm/static/leader_posix_timer/77/1/dump.log
b'(00.037771) Error (criu/timer.c:329): Unable to convert the notify thread id 80'
b"(00.037801) Error (criu/cr-dump.c:1731): Can't dump posix timers (pid: 80)"
b'(00.093711) Error (criu/cr-dump.c:2128): Dumping FAILED.'
############ Test zdtm/static/leader_posix_timer FAIL at CRIU dump #############
```

## Test Output — After Fix (PASS)

```
=== Run 1/1 ================ zdtm/static/leader_posix_timer
================== Run zdtm/static/leader_posix_timer in uns ===================
Start test
./leader_posix_timer --pidfile=leader_posix_timer.pid --outfile=leader_posix_timer.out
Run criu dump
Run criu restore
=================== Test zdtm/static/leader_posix_timer PASS ===================
=================== Run zdtm/static/leader_posix_timer in ns ===================
Start test
./leader_posix_timer --pidfile=leader_posix_timer.pid --outfile=leader_posix_timer.out
Run criu dump
Run criu restore
=================== Test zdtm/static/leader_posix_timer PASS ===================
=================== Run zdtm/static/leader_posix_timer in h ====================
Start test
./leader_posix_timer --pidfile=leader_posix_timer.pid --outfile=leader_posix_timer.out
Run criu dump
Run criu restore
=================== Test zdtm/static/leader_posix_timer PASS ===================
```

## Full Timer Regression Suite — All Pass

| Test              | uns  | ns   | h    |
|-------------------|------|------|------|
| posix_timers      | PASS | PASS | PASS |
| pthread_timers    | PASS | PASS | PASS |
| pthread_timers_h  | —    | —    | PASS |
| timers            | PASS | PASS | PASS |
| timers01          | PASS | PASS | PASS |
| leader_posix_timer| PASS | PASS | PASS |

### DIFF
--- a/criu/cr-dump.c
+++ b/criu/cr-dump.c
@@ -1671,6 +1671,7 @@ static int dump_one_task(struct pstree_item *item, InventoryEntry *parent_ie)
 	}
 
 	item->pid->ns[0].virt = misc.pid;
+	item->threads[0].ns[0].virt = misc.pid;
 	pstree_insert_pid(item->pid);
 	item->sid = misc.sid;
 	item->pgid = misc.pgid;
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -73,6 +73,7 @@ TST_NOFILE	:=				\
 		pthread02			\
 		pthread_timers			\
 		pthread_timers_h		\
+		leader_posix_timer		\
 		rseq00				\
 		membarrier			\
 		vdso00				\
@@ -612,6 +613,7 @@ pthread01:		LDLIBS += -pthread
 pthread02:		LDLIBS += -pthread
 pthread_timers:		LDLIBS += -lrt -pthread
 pthread_timers_h:	LDLIBS += -lrt -pthread
+leader_posix_timer:	LDLIBS += -lrt -pthread
 different_creds:	LDLIBS += -pthread
 sigpending:		LDLIBS += -pthread
 sigaltstack:		LDLIBS += -pthread
--- a/test/zdtm/static/leader_posix_timer.c
+++ b/test/zdtm/static/leader_posix_timer.c
@@ -1,0 +1,122 @@
+#include <errno.h>
+#include <stdlib.h>
+#include <string.h>
+#include <signal.h>
+#include <time.h>
+#include <pthread.h>
+#include <unistd.h>
+#include <sys/syscall.h>
+
+#include "zdtmtst.h"
+
+#ifndef SIGEV_THREAD_ID
+#define SIGEV_THREAD_ID 4
+#endif
+
+const char *test_doc = "Check SIGEV_THREAD_ID timer on the thread leader";
+const char *test_author = "CRIU developers";
+
+static volatile int sigcnt;
+
+static void timer_handler(int sig, siginfo_t *si, void *uc)
+{
+	sigcnt++;
+}
+
+static void *worker(void *arg)
+{
+	volatile int *stop = (volatile int *)arg;
+
+	while (!*stop)
+		usleep(100000);
+
+	return NULL;
+}
+
+int main(int argc, char **argv)
+{
+	struct sigevent evp = {};
+	struct itimerspec its = {};
+	struct sigaction sa = {};
+	timer_t timerid;
+	pthread_t thr;
+	int stop = 0;
+	pid_t tid;
+
+	test_init(argc, argv);
+
+	/*
+	 * Spawn a worker thread so the process is multithreaded.
+	 * The bug only manifests when collect_threads() runs and
+	 * skips vtid initialization for threads[0] (the leader).
+	 */
+	if (pthread_create(&thr, NULL, worker, &stop)) {
+		pr_perror("pthread_create");
+		return 1;
+	}
+
+	sa.sa_sigaction = timer_handler;
+	sa.sa_flags = SA_SIGINFO;
+	sigemptyset(&sa.sa_mask);
+	if (sigaction(SIGRTMIN, &sa, NULL)) {
+		pr_perror("sigaction");
+		return 1;
+	}
+
+	tid = syscall(SYS_gettid);
+
+	evp.sigev_notify = SIGEV_THREAD_ID;
+#ifdef __GLIBC__
+	evp._sigev_un._tid = tid;
+#else
+	evp.sigev_notify_thread_id = tid;
+#endif
+	evp.sigev_signo = SIGRTMIN;
+
+	if (timer_create(CLOCK_MONOTONIC, &evp, &timerid)) {
+		pr_perror("timer_create");
+		return 1;
+	}
+
+	its.it_interval.tv_nsec = 10000000; /* 10 ms */
+	its.it_value.tv_nsec = 10000000;
+
+	if (timer_settime(timerid, 0, &its, NULL)) {
+		pr_perror("timer_settime");
+		return 1;
+	}
+
+	/* Let at least one signal arrive */
+	while (!sigcnt)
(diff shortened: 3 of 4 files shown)


avagin:
Overall, the pr looks good to me. You need to fix ci failures:
```
#12 10.73 leader_posix_timer.c:65:6: error: no member named '_sigev_un' in 'struct sigevent'
1192
#12 10.73    65 |         evp._sigev_un._tid = tid;
1193
#12 10.73       |         ~~~ ^
```

3idey:
@avagin Thanks for the review! Fixed — the _sigev_un._tid field is glibc-internal and doesn't exist on musl. I added a #ifdef __GLIBC__ guard with sigev_notify_thread_id as the fallback, same pattern used in criu/pie/restorer.c. Also added a #ifndef SIGEV_THREAD_ID define since musl doesn't provide it (matching criu/proc_parse.c).

PR #2895, opened 2026-02-18
Add support for memory pages compression

rst0git:
This pull request extends CRIU with support for LZ4 compression applied during `dump` / `pre-dump` and reversed on `restore`. The goal is to reduce the size of data stored in `pages-*.img` files by compressing memory pages before they are written to the image and decompressing them after they are read back. This approach minimizes the amount of data written to and read from storage for local image-based dump and restore.

This functionality supports two compression modes:

- *Per-page compression* (`-c`, `--compress`): each 4 KiB page is compressed as its own LZ4 block.
- *Region compression* (`--compress-region SIZE`):  consecutive pages are grouped into regions of SIZE bytes, and each region is compressed as a single LZ4 block (default 256K, max 4M).

In addition, the `--compress-acceleration N` option exposes an LZ4 parameter providing a trade-off between speed and compression ratio. It controls how often LZ4 looks for repeated data while compressing. A value of 1 checks the input most thoroughly and usually produces the best compression ratio. Higher values cause LZ4 to skip more locations between searches and reduce the amount of work it does. This makes compression faster but increases the chance of missing repeated patterns. Decompression speed is not affected by this parameter.

Each compressed block stores its size in `pagemap_entry.compressed_size[]`, with three cases:
- `0`: all-zero block (no payload written)
- `block_bytes`: stored raw (incompressible, or below the 7/8  threshold where compressing isn't worth the restore cost)
- *any other value*: an LZ4 block of that size.

In addition, `total_compressed_size` lets the reader size a single `pread()`, and `region_pages` distinguishes the two modes.  The compression mode is saved in the inventory image for automatic detection on local image-based restore.

The following benchmark (`contrib/compression-benchmark/`) results show the saved pages image size when running with 1 GB workload (median of 5 runs) on Intel Core Ultra 9 275HX:

  | Workload | per-page | region (256K) |
  |---|---:|---:|
  | all-zero | 100% | 100% |
  | mixed (50% zero / 25% repeat / 25% random) | 75% | 75% |
  | JSON-shaped heap | 24% | 29% |
  | concatenated ELF binaries | 27% | 31% |
  | random (incompressible) | 0% | 0% |

Incompressible pages fall back to raw storage (7/8 threshold), so the worst case is break-even rather than a regression. Region mode improves the ratio on realistic data by sharing one LZ4 block across consecutive pages. The benchmark verifies memory integrity with SHA-256 after every restore.

```
CRIU Compression Benchmark
  Kernel : 6.8.0-106-generic
  CPU    : AMD Ryzen Threadripper PRO 7975WX 32-Cores
  Memory : 257235 MB
  CRIU   : Version: 4.2
  Config : 256 MB, 3+1 iterations, modes=none,per-page,region-64K,region-256K,region-1024K
                                        
========================================================================================
  MIXED (256 MB, n=3)
========================================================================================
  Memory integrity: PASS

  STORAGE                 pages         total       ratio     saved
  --------------------------------------------------------------------------------------
  none                266.2 MB      266.2 MB     1.000x       0%
  per-page             71.9 MB       72.1 MB     0.270x      73%
  region-64K           71.6 MB       71.6 MB     0.269x      73%
  region-256K          71.6 MB       71.6 MB     0.269x      73%
  region-1024K         71.6 MB       71.6 MB     0.269x      73%

  LATENCY (median)
  metric         |         none |     per-page |   region-64K |  region-256K | region-1024K
  --------------------------------------------------------------------------------------
  Frozen time    |     458.8 ms |     436.6 ms |     385.6 ms |     389.9 ms |     416.9 ms
  Memory dump    |     148.8 ms |     129.9 ms |      75.6 ms |      79.4 ms |     105.9 ms
  Page write     |     144.9 ms |     126.8 ms |      72.7 ms |      76.5 ms |     102.9 ms
  Dump total     |     765.5 ms |     746.9 ms |     694.3 ms |     699.2 ms |     728.5 ms
  Restore total  |     142.5 ms |      44.5 ms |      43.3 ms |      42.9 ms |      44.5 ms

  THROUGHPUT (per 266.2 MB of pages dumped)
  metric         |         none |     per-page |   region-64K |  region-256K | region-1024K
  --------------------------------------------------------------------------------------
  Dump (write)   |    1837 MB/s |    2100 MB/s |    3661 MB/s |    3481 MB/s |    2588 MB/s
  Restore        |    1918 MB/s |    6493 MB/s |    6701 MB/s |    6760 MB/s |    6498 MB/s
```

The benchmark was run on an ext4 filesystem backed by a 3.6 TB Corsair NVMe SSD (ROTA=0). For this mixed workload, compression consistently reduces checkpoint size and improves both dump and restore time compared uncompressed checkpoints. This is because less data has to be written and read back. However, these results are hardware-dependent. While compression reliably reduces the checkpoint size, the dump/restore time depends on whether the system is more I/O-bound or CPU-bound.

The following are evaluation results for SGLang:
```
Podman SGLang Checkpoint/Restore Benchmark
  Kernel : 6.8.0-106-generic
  CPU    : AMD Ryzen Threadripper PRO 7975WX 32-Cores
  Memory : 257235 MB
  GPU    : NVIDIA GeForce RTX 4090, NVIDIA GeForce RTX 4090, NVIDIA GeForce RTX 4090, NVIDIA GeForce RTX 4090, NVIDIA GeForce RTX 4090, NVIDIA GeForce RTX 4090
  Podman : podman version 4.6.2
  CRIU   : Version: 4.2
GitID: v4.2-306-g6c9c2da1b
  Plugin : default CRIU plugin path
  Runc conf: /etc/criu/runc.conf
  Archive: podman --compress=none
  Config : 3+1 iterations, modes=none,per-page,region-64K,region-256K,region-1024K
  Restore: decompress-threads=serial
...
========================================================================================
  PODMAN SGLANG (n=3)
========================================================================================
  Inference validation: PASS

  STORAGE               archive       ratio     saved
  --------------------------------------------------------------------------------------
  none                 7.73 GB     1.000x       0%
  per-page             2.77 GB     0.358x      64%
  region-64K           2.50 GB     0.324x      68%
  region-256K          2.46 GB     0.318x      68%
  region-1024K         2.45 GB     0.316x      68%

  LATENCY (median)
  metric         |         none |     per-page |   region-64K |  region-256K | region-1024K
  --------------------------------------------------------------------------------------
  Checkpoint     |     29.413 s |     25.445 s |     23.298 s |     22.847 s |     24.165 s
  Restore        |     16.318 s |      9.192 s |      8.734 s |      8.875 s |      8.703 s
  Request after  |      26.0 ms |      27.4 ms |      27.4 ms |      27.4 ms |      27.2 ms
```

The following presentation and research paper provide the motivation and a detailed evaluation of this work in the context of low-latency elastic inference serving:
- *[Optimizing Checkpoints with Built-in Memory Page Compression](https://lpc.events/event/19/contributions/2236/)*
- *[Towards On-the-Fly Snapshot Memory Compression for Low-Latency Elastic Inference Serving Systems](https://dl.acm.org/doi/10.1145/3805621.3807612)*

This pull request also extends CRIT with two offline image-conversion commands that mirror the built-in CRIU compression:

- `crit compress <dir>`: Compress the memory pages of a checkpoint directory in place.
- `crit decompress <dir>`: Decompress the memory pages of a checkpoint directory back to their uncompressed form

Both commands default to writing `.bak` backups of the files they rewrite, and support `--in-place` option that skips creating these backups.


criu/compression.c
```diff
@@ -0,0 +1,195 @@
...
+	if (ret <= 0) {
+		pr_err("Failed to decompress data: %d\n", ret);
+		return -1;
+	}
+
+	return 0;
+}
+
+
+int start_vma_io_pipe_daemon(int pages_img_fd, int pipe_fds[2][2])
```
avagin, line 264: why do we need a separate daemon?

criu/compression.c
```diff
@@ -0,0 +1,195 @@
...
+		for (int i = 0; i < nr; i++) {
+			ret = read(pipe_read_fd, &iov_len, sizeof(size_t));
+			if (ret < -1) {
+				pr_perror("Failed reading iov_len");
+				exit(1);
+			}
+
+			/* process_vm_writev() would fail with EINVAL if the
+			 * sum of the iov_len values overflows a ssize_t value */
+			if ((iov_len + total_len) > SSIZE_MAX) {
```
avagin: process_vm_writev will not write more than MAX_RW_COUNT

criu/compression.c
```diff
@@ -0,0 +1,195 @@
...
+				if (decompress_data(local_iovs[i].iov_base + j, compressed_size[i], PAGE_SIZE, remote_iovs[i].iov_base + j)) {
+					pr_perror("Decompression failed\n");
+					exit(1);
+				}
+			}
+		}
+
+		/* Write decrypted data to remote process address space */
+		ret = process_vm_writev(remote_pid, local_iovs, nr, remote_iovs, nr, 0);
+		if (ret < 0) {
```
avagin: you need to handle partial writes...

criu/compression.c
rst0git: It solves the same problem we had with encryption [^1]. Since the restorer code is PIE-compiled and uses [`preadv()`](https://github.com/checkpoint-restore/criu/blob/cff99dbcc2ea3e54574bc2812d8b14abe2d17c94/criu/pie/restorer.c#L1601) to read memory pages directly from the image file at a specific offset, this "daemon" acts as a pipeline that handles the read calls from the restorer, calls `preadv()`, and performs decompression. The idea is to be able to extend this pipeline also with support for decryption.

[^1]: See Figure 4 in [Towards Efficient End-to-End Encryption for Container Checkpointing Systems](https://dl.acm.org/doi/10.1145/3678015.3680477)

avagin: please add a comment to explain this in the code

images/pagemap.proto
```diff
@@ -14,4 +14,12 @@ message pagemap_entry {
 	optional bool	in_parent	= 3;
 	optional uint32	flags		= 4 [(criu).flags = "pmap.flags" ];
 	optional uint64 nr_pages = 5;
+	repeated uint32 compressed_size		= 6;
+	optional uint64	total_compressed_size	= 7;
```
avagin, line 45: what is a difference between these two fields? Please put a comment here.

criu/pie/restorer.c
```diff
@@ -1760,6 +1773,117 @@ static int reap_aio_events(struct task_restore_args *args, aio_context_t aio_ctx
 	return 0;
 }
 
+/*
+ * Wire protocol header between the PIE restorer and the helper daemon.
+ * Must match struct pipe_hdr in compression.c.
```
avagin: why is it not in a header?

criu/pie/restorer.c
```diff
@@ -2261,11 +2386,84 @@ __visible long __export_restore_task(struct task_restore_args *args)
 	 * Both helpers consume args->vma_ios and close args->vma_ios_fd on exit.
 	 */
 	if (args->vma_ios_n > 0 && args->vma_ios_fd != -1) {
-		int rc = args->vma_ios_use_direct
-			? restore_vma_aio(args)
-			: restore_vma_preadv(args);
-		if (rc < 0)
-			goto core_restore_end;
+		if (args->compress_mode) {
+			/*
```
avagin: it should be better to shuffle code to a few helpers here.

criu/include/restorer.h
```diff
@@ -248,6 +262,16 @@ struct task_restore_args {
 	 */
 	struct rst_rseq_param libc_rseq;
 
+	/*
+	 * enum compress_mode value. 0 means no compression — pipes/daemon
+	 * are unused. PIE callers only need the predicate, so we keep this
+	 * as a plain int rather than pulling the enum into PIE headers.
+	 */
+	int compress_mode;
+	int page_pipe_fd_r;
```
avagin: why do we need two pipes? If this is just for a two-direction communication channel, you probably need to use a socketpair.

avagin: If this channel is used just in case of compression, please rename it to indicate that.

criu/pie/restorer.c
```diff
@@ -1760,6 +1773,117 @@ static int reap_aio_events(struct task_restore_args *args, aio_context_t aio_ctx
...
+ * compressed_size arrays or iovec arrays.
+ */
+static int pipe_write_full(int fd, const void *buf, size_t size)
+{
+	size_t done = 0;
+
+	while (done < size) {
+		ssize_t ret = sys_write(fd, (const char *)buf + done,
+					size - done);
+		if (ret < 0)
```
avagin: should we print a error here while we have the error code?

criu/compression.c
```diff
@@ -136,3 +136,545 @@ int decompress_region(const char *src, int compressed_size,
...
+			local[iov_off].iov_len -= ret;
+			remote[iov_off].iov_base += ret;
+			remote[iov_off].iov_len -= ret;
+		}
+	}
+
+	return 0;
+}
+
+int start_vma_io_pipe_daemon(int pages_img_fd, int pipe_fds[2][2])
```
avagin: please consider using asyncd that was introduced recently.

criu/compression.c
```diff
@@ -136,3 +136,545 @@ int decompress_region(const char *src, int compressed_size,
...
+	size_t cs_cap = 0;
+	uint32_t *page_cs_shadow = NULL;
+	size_t pcs_cap = 0;
+	uint16_t *block_pages = NULL;
+	size_t bp_cap = 0;
+	struct iovec *remote_iovs = NULL, *local_iovs = NULL;
+	struct iovec *remote_iovs_wr = NULL;  /* write-side remote iovecs */
+	size_t iovs_cap = 0;
+	char *compressed_buf = NULL, *decompressed_buf = NULL;
+	size_t comp_cap = 0, decomp_cap = 0;
```
avagin: nit: please resort these declarations to alligne them by the line length, usually it makes the code more readable.

criu/compression.c
```diff
@@ -136,3 +136,545 @@ int decompress_region(const char *src, int compressed_size,
...
+
+	if (child_pid > 0) {
+		return child_pid;
+	}
+
+	close(pipe_fds[1][0]);
+	close(pipe_fds[0][1]);
+
+	/* Hint the kernel for sequential readahead on the pages image */
+	posix_fadvise(pages_img_fd, 0, 0, POSIX_FADV_SEQUENTIAL);
```
avagin: why do we need to do that from the daemon?

images/pagemap.proto
rst0git: Thanks! I've updated the proto file with comments to explain the purpose of these fields.

criu/pie/restorer.c
rst0git: I've updated the patch - this is now in `criu/include/compression.h` as `struct vma_io_compress_hdr`.

criu/pie/restorer.c
rst0git: I've  split this into helpers: `pipe_write_full()`, `pipe_read_full()`, `compressed_preadv()`, and `restore_vma_compressed()`.

criu/include/restorer.h
rst0git: This was removed. Compressed restore now uses asyncd.

criu/pie/restorer.c
rst0git: I've updated the read/write helpers and `restore_vma_compressed()` to print errors.

criu/compression.c
rst0git: I've updated the patch to replace the separate helper daemon with asyncd. The PIE restorer sends compressed restore requests to `start_vma_io_asyncd()`, and the asyncd service performs decompression and `process_vm_writev()`. This approach allows us to introduce also support for  encryption/decryption in the future.

criu/compression.c
rst0git: I've added a comment  - the PIE restorer cannot link against LZ4, so decompression is delegated to asyncd. The service reads compressed data from pages.img, decompresses it, and writes restored pages with `process_vm_writev()`.

criu/pagemap.c
```diff
@@ -404,6 +821,340 @@ static int maybe_read_page_local(struct page_read *pr, unsigned long vaddr, unsi
...
+				       idx, cs, this_region);
+				goto out;
+			}
+		} else {
+			/* Partial slice: decompress into scratch, then memcpy. */
+			char *region_dec = xmalloc(this_bytes);
+
+			if (!region_dec)
+				goto out;
+			if (pread_full(fd, scratch, cs, pr->pi_off)) {
```
avagin: I think you may want to consider caching the last partially read region, as there is a chance it will be read again on the next attempt:
https://github.com/avagin/criu/commit/2f43b5a0be44cd59b03d08b30f1195d2ad490474
```
root@criu-next-2:/home/avagin/git/criu# python3 test/zdtm.py run -t zdtm/static/pre-dump00 --compress-region=1M --pre 1 -f h -k always
mem_dirty_track is supported
userns is supported
=== Run 1/1 ================ zdtm/static/pre-dump00
======================= Run zdtm/static/pre-dump00 in h ========================
 DEP       pre-dump00.d
 CC        pre-dump00.o
 LINK      pre-dump00
Start test
./pre-dump00 --pidfile=pre-dump00.pid --outfile=pre-dump00.out
Run criu pre-dump
Execution time: 0.214077 seconds
Send pre-dump notify to 89
Wait pre-dump notify reply
Completed pre-dump notify with 0
Run criu dump
Execution time: 0.714707 seconds
=[log]=> dump/zdtm/static/pre-dump00/89/2/dump.log
b'Error (criu/page-xfer.c:800): page-xfer: No parent image found, though parent directory is set: No such file or directory'
Run criu restore
==== ALARM ====
Execution time: 20.994700 seconds
Send the 15 signal to  89
Wait for zdtm/static/pre-dump00(89) to die for 0.100000
======================= Test zdtm/static/pre-dump00 PASS =======================
root@criu-next-2:/home/avagin/git/criu# python3 test/zdtm.py run -t zdtm/static/pre-dump00 --pre 1 -f h -k always
mem_dirty_track is supported
userns is supported
=== Run 1/1 ================ zdtm/static/pre-dump00
======================= Run zdtm/static/pre-dump00 in h ========================
Start test
./pre-dump00 --pidfile=pre-dump00.pid --outfile=pre-dump00.out
Run criu pre-dump
Execution time: 0.213991 seconds
Send pre-dump notify to 66
Wait pre-dump notify reply
Completed pre-dump notify with 0
Run criu dump
Execution time: 0.264085 seconds
=[log]=> dump/zdtm/static/pre-dump00/66/2/dump.log
b'Error (criu/page-xfer.c:800): page-xfer: No parent image found, though parent directory is set: No such file or directory'
Run criu restore
Execution time: 0.314092 seconds
Send the 15 signal to  66
Wait for zdtm/static/pre-dump00(66) to die for 0.100000
======================= Test zdtm/static/pre-dump00 PASS =======================
```

rst0git: Thanks! I've added a patch that implements caching.

rst0git:
I rebased the branch on criu-dev (to include the changes from https://github.com/checkpoint-restore/criu/pull/3066 and resolve the merge conflicts). I also applied a few additional performance optimizations suggested by @avagin.

test/zdtm/static/compress_pages00.c
```diff
@@ -0,0 +1,202 @@
...
+}
+
+static int setup_file_zero_mapping(char **file_zero_region, char *filename)
+{
+	char buf[PAGE_SIZE];
+	int fd;
+
+	memset(buf, 0x7f, sizeof(buf));
+
+	fd = mkstemp(filename);
```
avagin: why do you want to create a temporary file rather then using the zdtm file test pattern?

contrib/compression-benchmark/main.py
```diff
@@ -0,0 +1,489 @@
+#!/usr/bin/env python3
```
avagin, line 1: why is it in the contrib/ directory? Should it be in test/? Plus, if anything that isn't executed in CI is already broken or going to be broken soon ;).

criu/include/pagemap.h
```diff
@@ -1,6 +1,9 @@
 #ifndef __CR_PAGE_READ_H__
```
avagin, line 1: I think Co-authored-by should be replaced with Suggested-by. I haven't written any code in this cl.

criu/include/pagemap.h
```diff
@@ -114,6 +117,15 @@ struct page_read {
 	 */
 	unsigned int region_block_offset;
 
+	/* Last decompressed region block for repeated partial region reads. */
+	char *region_cache;
+	size_t region_cache_cap;
```
avagin: what is region_cache_cap for?

avagin: why do we need more than cached_region_vaddr and cached_region_size?

criu/include/compression.h
```diff
@@ -115,16 +129,12 @@ int decompress_region(const char *src, int compressed_size,
...
- * runs as a regular process, reads compressed pages from the image
- * file, decompresses them, and writes the result into the target
- * address space via process_vm_writev(). Communication with the
- * restorer happens over a pair of pipes.
+ * Start a dedicated decompression daemon for PIE VMA I/O. The returned
+ * socket is passed to PIE as task_restore_args.page_asyncd_fd, and the
+ * returned pid is reaped by PIE after it closes the socket.
  */
-int start_vma_io_pipe_daemon(int pages_img_fd, int pipe_fds[2][2]);
+int start_vma_io_compress_daemon(int pages_img_fd, unsigned int decompress_threads,
```
avagin: why do we need a dedicated decompression daemon? Why asyncd can't be used for that?

criu/include/compression.h
```diff
@@ -1,6 +1,9 @@
 #ifndef __CR_COMPRESSION_H__
```
avagin, line 1: What mappings can't be restored before switching to the pie code? I remember MAP_HUGETLB mappings didn't support mremap, but I think it was fixed:
```
commit 550a7d60bd5e35a56942dba6d8a26752beb26c9f
Author: Mina Almasry <almasrymina@google.com>
Date:   Fri Nov 5 13:41:40 2021 -0700

    mm, hugepages: add mremap() support for hugepage backed vma
```

is there anything else? I don't like this helper daemon and I am trying to think whether it is really required or not.

Cc: @mihalicyn

test/zdtm/static/compress_pages00.c
rst0git: Thanks! I've updated the test to use the zdtm file test pattern.

contrib/compression-benchmark/main.py
rst0git: I've added tests in `test/others` to run all benchmarks and kept the benchmarks in `contrib/` as they can be used with different parameters for performance evaluation.

criu/include/pagemap.h
rst0git: Sure, I've changed it to "Suggested-by:".

criu/include/pagemap.h
rst0git: You are right - I have updated the patch to simplify the cache. I've also updated the commit message and added a benchmark script.

criu/include/compression.h
rst0git: The daemon is not fundamentally required - it was introduced because `pr-pieok` leaves unshared private VMAs to be restored at PIE-time and the PIE code can't be linked with liblz4. asyncd is not a replacement in the current flow because it is drained before CR_STATE_RESTORE. When I tried to implement this, keeping a callback alive until the PIE code also resulted in a deadlock and encounters the overhead `process_vm_writev()` copy. As you suggested, a better solution is to restore the VMAs containing the LZ4 blocks before PIE. This removes the daemon.

criu/include/compression.h
rst0git: As you mentioned, unshared private VMAs can be restored before PIE. The exceptions seems to be MAP_HUGETLB and VMA_EXT_PLUGIN. We can use `mremap()` for hugetlb, and the mappings handled by plugins are skipped because the plugin may replace their FD or offset later during UPDATE_VMA_MAP.

I've updated the patches so that VMAs with LZ4 blocks are premapped and decompressed before the PIE code, while raw-only pages use the fast path, reading them directly.

avagin:
@rst0git you can address CI failures and merge this pr. All father development will be done on top of criu-dev. @mihalicyn please try to find time to review this series. All Alex's findings can be addressed after merging this pr.

criu/include/compression.h
mihalicyn: yeah, `mremap()` for HUGETLB is supported now. I quickly looked in the kernel and it seems that all mappings we need support `mremap()` these days.

criu/pagemap.c
```diff
@@ -575,6 +1477,225 @@ static int process_async_reads(struct page_read *pr)
 
 		pr_debug("Read piov iovs %d, from %ju, len %ju, first %p:%zu\n", piov->nr, piov->from,
 			 piov->end - piov->from, piov->to->iov_base, piov->to->iov_len);
+
+		if (piov->n_compressed_size) {
```
mihalicyn: nit: I wonder if we can somehow refactor this `list_for_each_entry_safe(piov, n, &pr->async, l)` so this quite a big chunk of code that we are adding for a compressed case will go to a separate function.

Another argument to the same direction is that we use `goto next;` to skip the entire logic for non-compressed case. Wouldn't it be simpler to, instead of doing `goto next;` just put non-compressed logic in the `else` block for `if (piov->n_compressed_size)`?

P.S. Not a call to action or rework, but probably a good point for a future refactoring and improvement.

criu/pie/restorer.c
```diff
@@ -150,8 +151,6 @@ static void sigchld_handler(int signal, siginfo_t *siginfo, void *data)
 	char *r;
 	int i;
 
-	/* We can ignore helpers that die, we expect them to after
```
mihalicyn, line 153: Hmm.. why this chunk is here?

criu/pie/restorer.c
```diff
@@ -2257,15 +2421,24 @@ __visible long __export_restore_task(struct task_restore_args *args)
...
+	 * delayed file-backed range is raw and page-aligned; encoded ranges are
+	 * premapped and decoded before PIE.
 	 */
-	if (args->vma_ios_n > 0 && args->vma_ios_fd != -1) {
-		int rc = args->vma_ios_use_direct
-			? restore_vma_aio(args)
-			: restore_vma_preadv(args);
-		if (rc < 0)
-			goto core_restore_end;
+	if (args->vma_ios_n > 0) {
```
mihalicyn: Is this intended change? Looks like you wanted a simple `restore_vma_preadv` -> `restore_vma_preadv_mixed` rename here.

test/zdtm/static/compress_pages00.c
```diff
@@ -0,0 +1,261 @@
...
+#include "zdtmtst.h"
+
+const char *test_doc = "Check memory page content integrity across checkpoint/restore with compression";
+const char *test_author = "Radostin Stoyanov <rstoyanov@fedoraproject.org>";
+
+char *filename;
+TEST_OPTION(filename, string, "file name", 1);
+
+/*
+ * Allocate memory regions with different content patterns that
```
mihalicyn: We already have a plenty of tests like `mapsXYZ` and `cowXYZ` too, I wonder how these tests are different with what we already have and if so, then why we only run these tests with `--compress` option. Shouldn't we also add a corresponding `desc` file + symlink to exercise no-compresion case too with these tests?

Please, understand me right - more tests is good, but I believe we should run the same tests against with/no-compression scenarios. ;-)

criu/page-xfer.c
```diff
@@ -1532,21 +1599,24 @@ struct page_xfer_job {
...
+	if (cxfer.dst_id != ~0ULL) {
+		/* Make repeated/error-path cleanup harmless before closing resources. */
+		cxfer.dst_id = ~0ULL;
 		cxfer.loc_xfer.close(&cxfer.loc_xfer);
-	if (pipe_read_dest.sink_fd != -1) {
-		close(pipe_read_dest.sink_fd);
-		close(pipe_read_dest.p[0]);
-		close(pipe_read_dest.p[1]);
 	}
+	close_safe(&pipe_read_dest.sink_fd);
```
mihalicyn, line 1575: nit: shouldn't this be a separate patch?

criu/page-xfer.c
```diff
@@ -1870,10 +1960,22 @@ static int page_server_serve(int sk)
 			return -1;
 		}
 
-		cxfer.pipe_size = fcntl(cxfer.p[0], F_GETPIPE_SZ, 0);
+		pipe_size = fcntl(cxfer.p[0], F_GETPIPE_SZ, 0);
```
mihalicyn, line 1911: nit: looks like this should be in a separate patch

criu/mem.c
```diff
@@ -1390,11 +1525,16 @@ int prepare_mappings(struct pstree_item *t)
 {
 	int ret = 0;
 	void *addr;
+	void *new_premmapped_addr = NULL;
```
mihalicyn: please, can you explain why all these changes to `prepare_mappings` are needed? For me all this looks like some refactoring or attempt to fix some leaks, but if so -> it should be in a separate commit and properly explained.

Sorry for being a bit picky here, but every single line of code in `mem.c` has a big history going back to our great predecessors (I mean CRIU developers who wrote all this originally ;-) ), so if we change something we must be 10 billion times sure that this change is for good ;-)

criu/compression.c
```diff
@@ -193,3 +411,471 @@ int decompress_region(const char *src, int compressed_size,
...
+			end = queue->nr_jobs;
+
+		for (i = first; i < end; i++) {
+			struct decompress_job *job = &queue->jobs[i];
+
+			if (decompress_job_run(job)) {
+				int failed_block = __atomic_load_n(
+					&queue->failed_block, __ATOMIC_RELAXED);
+
+				while (job->block_index < failed_block &&
```
mihalicyn: This chunk is really hard to read. AFAIU, idea here is to make sure that `queue->failed_block` holds a first failed block. But still, I think good explanatory comment is needed here.

criu/compression.c
```diff
@@ -193,3 +411,471 @@ int decompress_region(const char *src, int compressed_size,
...
+				       !__atomic_compare_exchange_n(
+					       &queue->failed_block,
+					       &failed_block, job->block_index,
+					       false, __ATOMIC_RELAXED,
+					       __ATOMIC_RELAXED))
+					;
+				__atomic_store_n(&queue->failed, 1,
```
mihalicyn: Why do we use compiler builtins instead of `atomic_t`?

mihalicyn (commented):
I spend a few hours today reading all this, and wasn't able to reach the tests part, but went through the most of key places in the mem, page xfer and other critical pieces.

Amazing work, Radostin!

I have a feeling that some chunks of code were probably accidentally touched by AI Agent and this looks really scary to me. We should try to make changes as small as possible, especially when it comes to mem-related code.

Just to be clear: given all the work you've done and extensive tests I see no problem merging this. But if you can take a quick look on my comments regarding chunks with unrelated changes or bug fixes as a part of main logic code and split all these somehow that would be awesome.

criu/pagemap.c
rst0git: Thanks! I've updated this patch to use helper functions.

criu/pie/restorer.c
rst0git: Thanks for pointing this out! I've updated the patches to drop these.

criu/page-xfer.c
rst0git: Thanks! I moved it in a separate patch.

criu/page-xfer.c
rst0git: Thanks! I moved it in a separate patch.

criu/pie/restorer.c
rst0git: Yes, the inner read loop becomes `restore_vma_preadv_one(args, rio, allow_dedup)` and `restore_vma_preadv_mixed()` dispatches each entry by storage type (i.e., uncompressed, packed-raw, and zero entries). I've updated the patch to improve this.

criu/compression.c
rst0git: This is a shared work queue for the decompression threads - each thread updates `next_job` to claim a different group of blocks, so the update must be atomic. `next_job` is a `size_t`, while CRIU's `atomic_t` only supports `int`. We use the compiler builtins as they work with `size_t`.

test/zdtm/static/compress_pages00.c
rst0git: The main difference from `mapsXYZ` and `cowXYZ` is the page content and layout. These tests create zero, compressible, raw-fallback, and mixed blocks. I've updated the patches so that uncompressed tests are included in default ZDTM runs.

criu/mem.c
rst0git: You are right - the changes are just cleanup of the existing failure paths by closing the page reader, releasing a failed premap reservation, preserving the old mapping until the new one is ready, and propagating `munmap()` failures correctly. I will drop them from this patch as they are unrelated.

criu/compression.c
rst0git: Thanks! I've updated this file to improve readability.

contrib/compression-benchmark/main.py
avagin: It might be better to create test/benchmarks/ and put all benchmarks there.

test/others/compression/incremental/image_tool.py
```diff
@@ -0,0 +1,126 @@
...
+        ]
+        if not present:
+            print("FAIL: no present mapping available to modify in %s" % path)
+            return 1
+        entry = max(present, key=lambda item: int(item.get("nr_pages", 0)))
+        address = int(entry["vaddr"])
+        with open(address_file, "w") as saved:
+            saved.write("%x\n" % address)
+
+    with open("/proc/%s/mem" % pid, "r+b", buffering=0) as memory:
```
avagin, line 85: zdtm tests can handle pre-dump-s and do specified actions. I sent you an example a few weeks ago. Why don't we use this functionality?

rst0git: The purpose of this test is slightly different from what we do with ZDTM tests. It aims to cover incremental checkpoints mixing uncompressed images with per-page or region-compressed images in both directions:
- uncompressed parent images -> compressed final dump
- compressed parent images -> uncompressed final dump

It also verifies each image version, memory contents, and confirms that restore is successful.

PR #2899, opened 2026-02-19
Handle truncated pages on restore

fntlnz:
I noticed that when restoring a dump containing a pages file that was truncated (likely caused by a filesystem error) the restore process got stuck while performing `preadv` on the pages file because of the unexpected EOF it would get.

Also this bug is present in two different cases:

- when pages are read directly during a restore
- when restoring premapped VMAs (either because of a predump or because of things like memory COW in a fork like the test here). I decided to do the test with COW because it was simpler.

### DIFF
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -559,6 +559,12 @@ static int process_async_reads(struct page_read *pr)
 			return -1;
 		}
 
+		if (ret == 0 && piov->end != piov->from) {
+			pr_err("Unexpected EOF reading pages: expected %ju more bytes at offset %ju\n",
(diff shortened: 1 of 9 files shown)


avagin:
could you please move tests to a separate commit.

fntlnz:
@avagin makes sense, done

avagin:
I’ve merged both fixes, but I’m still unsure about the tests. Regression tests rarely catch anything useful, and we are already facing resource constraints for running all test configurations. Thank you for contributing to the project.

fntlnz:
@avagin sounds fair, I contributed the tests mainly to show you how the issue manifests itself.

PR #2900, opened 2026-02-20
bfd: Support dynamic expansion for massively long text lines

### LINKED ISSUE #2898: Cannot dump from user added in too many groups
Hi all,

We are using CRIU to migrate processes from one machine to another. However, we are encountering failures when checkpointing processes for users who have a large size of group file (user has read access to alot of teams and subteams).

Even a simple sleep process dump for such users is failing.

For example, when attempting to dump:
`sudo /usr/local/bin/remoteCheckpoint dump -t 675488 -D /tmp/ --shell-job --tcp-close --file-locks`

We are seeing the following errors:
`Error (criu/bfd.c:199): bfd: The bfd buffer is too small `
`Error (criu/proc_parse.c:1216): Error parsing proc status file`
`Error (compel/src/lib/infect.c:418): Unable to detach from 675488: No such process`
`Error (criu/cr-dump.c:2111): Dumping FAILED. `
Is there any fix to this?

DongSunchao:
Fixes #2898 

**Motivation**
Currently, `breadchr()` aborts with "The bfd buffer is too small" when encountering a process with a massive number of supplementary groups, because the `Groups:` line in `/proc/<pid>/status` exceeds the hardcoded `BUFSIZE` (4096 bytes). 

While increasing `BUFSIZE` globally (e.g., to `PAGE_SIZE * 16`) works as a quick workaround, it is not optimal. Due to Linux's demand paging, while this only reserves virtual memory initially, recycling a dirtied, globally enlarged buffer back into the standard pool via `list_add_tail()` causes severe internal fragmentation (RSS bloat). Those physical pages would remain resident in memory indefinitely even when reused for tiny files.

**Implementation Details**
This PR implements a strict dynamic expansion mechanism in `bfd.c`:
- Keeps the default zero-copy batched `mmap` pool (`BUFSIZE`) intact for 99% of normal routine operations.
- When `breadchr()` hits the buffer boundary, it dynamically maps an independent, doubled VMA to handle the long text line (capped at 2MB to prevent OOM).
- Explicitly intercepts and safely unmaps (`munmap`) these oversized custom buffers in both `buf_put()` and during consecutive re-expansions in `breadchr()`, strictly preventing them from polluting the fixed-size 4KB memory pool.
- Included a new ZDTM test (`massive_groups`) using 900 groups with 10-digit GIDs (e.g., `1000000000`). This generates a ~10KB `Groups:` line, intentionally forcing **multiple consecutive dynamic expansions** (4K -> 8K -> 16K) to thoroughly validate the expansion and safe-discard logic, while safely staying below the `PARASITE_MAX_GROUPS` limit.

Tested locally with `make zdtm` and successfully passed the `massive_groups` test in the host namespace.


criu/bfd.c
```diff
@@ -77,6 +83,14 @@ static int buf_get(struct xbuf *xb)
 
 static void buf_put(struct xbuf *xb)
 {
+	struct bfd_buf *b = xb->buf;
+
+	if (b->mem != b->orig_mem) {
+		munmap(b->mem, b->capacity);
+		b->mem = b->orig_mem;
+		b->capacity = BUFSIZE;
+	}
```
Snorch, line 92: Thank for your patch!

I think we can be more frugal, by not having "orig_mem" hanging there allocated but unused.

We can probably add a size field to bfd_buf, and reallocate ->mem directly to the desired size (effectively freeing old ->mem).

This way we should only reuse (put back to bufs list) the buf if it's size is standard, all non-standard size buffers should be freed immediately after put.

Does it make sense?

Snorch: Alternatively we can put the bfd_buf to list immediately if we see that data does not fit, and create a custom size buffer to replace it.

DongSunchao: Thank you for the review and the brilliant suggestion!

Hanging allocated but unused memory is wasteful. I'll update the PR with these changes shortly.

criu/bfd.c
```diff
@@ -195,9 +213,49 @@ char *breadchr(struct bfd *f, char c)
...
+			new_mem = mmap(NULL, new_cap, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
+			if (new_mem == MAP_FAILED) {
+				xfree(new_buf);
+				pr_perror("Unable to mmap a buffer");
+				return ERR_PTR(-ENOMEM);
+			}
+
+			memcpy(new_mem, b->data, b->sz);
+
+			if (b_buf) {
```
Snorch: nit: Please remove unnecesarry brackets around single line if branch.

criu/bfd.c
```diff
@@ -195,9 +213,49 @@ char *breadchr(struct bfd *f, char c)
...
+
+			b_buf = new_buf;
+			b->buf = b_buf;
+
+			new_buf->mem = new_mem;
+			new_buf->size = new_cap;
+
+			b->mem = new_mem;
+			b->data = new_mem;
+
```
Snorch, line 275: `ss = b->sz;` might be a good idea here.

test/zdtm/static/massive_groups.c
```diff
@@ -0,0 +1,75 @@
...
+		fail("Failed to allocate memory for group IDs");
+		return 1;
+	}
+
+	/*
+	 * Fill the group array with unique group IDs to reach the maximum limit.
+	 * This will trigger the dynamic expansion of the BFD buffer when dumping
+	 * the process.
+	 */
+	for (i = 0; i < ngroups; i++) {
```
Snorch: nit: again don't need {} here.

Snorch (approved):
I added small improvement comments, please take a look. But generally this looks good to me.

criu/bfd.c
```diff
@@ -195,9 +213,49 @@ char *breadchr(struct bfd *f, char c)
...
+			if (new_mem == MAP_FAILED) {
+				xfree(new_buf);
+				pr_perror("Unable to mmap a buffer");
+				return ERR_PTR(-ENOMEM);
+			}
+
+			memcpy(new_mem, b->data, b->sz);
+
+			if (b_buf) {
+				list_add_tail(&b_buf->l, &bufs);
```
avagin: it can be a non-"standard" buffer, can't it?

criu/bfd.c
DongSunchao: I totally missed that. Without `ss = b->sz`;, strnchr would redundantly re-scan the entire previously matched chunk after every expansion.

Now I have fixed it.

criu/bfd.c
DongSunchao: You're right. I have changed the logic to check for this. And also change the test file to multiple expansions while staying within the `PARASITE_MAX_GROUPS` limit.

Thanks again for the sharp eye! The updated commits have been pushed.

criu/bfd.c
```diff
@@ -77,11 +81,20 @@ static int buf_get(struct xbuf *xb)
...
+	if (b->size > BUFSIZE) {
+		/* This buffer was remapped to fit a long line, unmap it back */
+		munmap(b->mem, b->size);
+		xfree(b);
+	} else {
+		/*
+		 * Don't unmap standard buffer back, it will get reused
+		 * by next bfdopen call
+		 */
+		list_add_tail(&b->l, &bufs);
```
avagin: Do you understand why using list_add_tail here will "unnecessarily inflates the memory footprint  for all routine checkpoint operations", but  increasing `BUFSIZE` probably will not do that?

criu/bfd.c
```diff
@@ -195,9 +213,60 @@ char *breadchr(struct bfd *f, char c)
...
+			new_buf = xmalloc(sizeof(*new_buf));
+			if (!new_buf) {
+				pr_perror("Unable to allocate buffer for bfd");
+				return ERR_PTR(-ENOMEM);
+			}
+
+			new_mem = mmap(NULL, new_cap, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
+			if (new_mem == MAP_FAILED) {
+				xfree(new_buf);
+				pr_perror("Unable to mmap a buffer");
```
avagin: pr_perror prints errno and should be used right after a system call...

test/zdtm/static/massive_groups.desc
```diff
@@ -0,0 +1 @@
+{'flavor': 'h', 'flags': 'suid'}
```
avagin: why is it running only in the host namespace?

test/zdtm/static/massive_groups.c
```diff
@@ -0,0 +1,74 @@
...
+	/*
+	 * Fill the group array with large and unique group IDs to reach the maximum limit.
+	 * This will trigger the multiple dynamic expansion of the BFD buffer when dumping
+	 * the process.
+	 */
+	for (i = 0; i < ngroups; i++)
+		group[i] = i + 1000000000;
+
+	if (setgroups(ngroups, group) < 0) {
+		fail("Failed to set supplementary groups");
```
avagin: fail() should be used only when you are checking results. For other errors, you can use pr_err/pr_perror.

test/zdtm/static/massive_groups.c
```diff
@@ -0,0 +1,74 @@
...
+	getgroups(restored_ngroups, restored_groups);
+
+	for (i = 0; i < ngroups; i++) {
+		if (restored_groups[i] != group[i])
+			fail("Restored group ID at index %d (%d) does not match expected (%d)", i, restored_groups[i], group[i]);
+	}
+
+	free(restored_groups);
+	free(group);
+	pass();
```
avagin, line 94: pass() should not be called if the test fails...

criu/bfd.c
DongSunchao: Got it！Using `list_add_tail` after reading a massive line will cause **internal Fragmentation** when the buffer is reused for a smaller file later. And thank to **Demand Paging**, increasing the global `BUFSIZE` only reserves virtual memory, and physical memory is only allocated when it's actually touched.

test/zdtm/static/massive_groups.desc
DongSunchao: The main reason is that this test will call `setgroups()` with 900, extremely large `GIDs`.
If we run it without `'flavor': 'h'` (in a user namespace), it requires a massive and complex `gid_map` setup for all those custom groups. The ZDTM sandbox initialization actually fails under these conditions (I encountered a `mount(/proc) failed` error when I tried running it in `ns`).

criu/bfd.c
```diff
@@ -195,9 +213,60 @@ char *breadchr(struct bfd *f, char c)
 		if (!b->sz)
 			return NULL;
 
-		if (b->sz == BUFSIZE) {
-			pr_err("The bfd buffer is too small\n");
-			return ERR_PTR(-EIO);
+		cap = b_buf ? b_buf->size : BUFSIZE;
+		if (b->sz == cap) {
+			struct bfd_buf *new_buf;
```
avagin: The buffer grow logic should be moved to the helper.

criu/bfd.c
```diff
@@ -251,15 +325,17 @@ static int bflush(struct bfd *bfd)
 static int __bwrite(struct bfd *bfd, const void *buf, int size)
 {
 	struct xbuf *b = &bfd->b;
+	struct bfd_buf *b_buf = b->buf;
+	size_t cap = b_buf ? b_buf->size : BUFSIZE;
```
avagin: why don't we initialize b_buf->size for "standard" buffers?

avagin:
@DongSunchao what do you think about this draft? 
```
diff --git a/criu/bfd.c b/criu/bfd.c
index 2c5f5b64a..2e44acd03 100644
--- a/criu/bfd.c
+++ b/criu/bfd.c
@@ -70,6 +70,7 @@ static int buf_get(struct xbuf *xb)
 
 	xb->mem = b->mem;
 	xb->data = xb->mem;
+	xb->bsize = BUFSIZE;
 	xb->sz = 0;
 	xb->buf = b;
 	return 0;
@@ -77,11 +78,15 @@ static int buf_get(struct xbuf *xb)
 
 static void buf_put(struct xbuf *xb)
 {
-	/*
-	 * Don't unmap buffer back, it will get reused
-	 * by next bfdopen call
-	 */
-	list_add(&xb->buf->l, &bufs);
+	if (xb->buf) {
+		/*
+		 * Don't unmap buffer back, it will get reused
+		 * by next bfdopen call
+		 */
+		list_add(&xb->buf->l, &bufs);
+	} else {
+		munmap(xb->mem, xb->bsize);
+	}
 	xb->buf = NULL;
 	xb->mem = NULL;
 	xb->data = NULL;
@@ -144,7 +149,7 @@ static int brefill(struct bfd *f)
 	memmove(b->mem, b->data, b->sz);
 	b->data = b->mem;
 
-	ret = read_all(f->fd, b->mem + b->sz, BUFSIZE - b->sz);
+	ret = read_all(f->fd, b->mem + b->sz, b->bsize - b->sz);
 	if (ret < 0) {
 		pr_perror("Error reading file");
 		return -1;
@@ -172,6 +177,30 @@ char *breadline(struct bfd *f)
 	return breadchr(f, '\n');
 }
 
+static int bextend(struct bfd *f)
+{
+	struct xbuf *b = &f->b;
+	void *newbuf;
+	long newsize = b->bsize * 2;
+
+	if (b->buf) {
+		newbuf = mmap(NULL, newsize, PROT_READ | PROT_WRITE, MAP_ANONYMOUS | MAP_PRIVATE, 0, 0);
+		if (newbuf == MAP_FAILED)
+			return -1;
+		memcpy(newbuf, b->mem, b->sz);
+		list_add(&b->buf->l, &bufs);
+		b->buf = NULL;
+	} else {
+		newbuf = mremap(b->mem, b->bsize, newsize, MREMAP_MAYMOVE);
+		if (newbuf == MAP_FAILED)
+			return -1;
+	}
+	b->mem = newbuf;
+	b->data = newbuf;
+	b->bsize = newsize;
+	return 0;
+}
+
 char *breadchr(struct bfd *f, char c)
 {
 	struct xbuf *b = &f->b;
@@ -195,9 +224,12 @@ again:
 		if (!b->sz)
 			return NULL;
 
-		if (b->sz == BUFSIZE) {
-			pr_err("The bfd buffer is too small\n");
-			return ERR_PTR(-EIO);
+		if (b->sz == b->bsize) {
+			if (bextend(f)) {
+				pr_err("The bfd buffer is too small\n");
+				return ERR_PTR(-EIO);
+			}
+			goto refill;
 		}
 		/*
 		 * Last bytes may lack the \n at the
@@ -216,6 +248,7 @@ again:
 		return b->data;
 	}
 
+refill:
 	/*
 	 * small optimization -- we've scanned b->sz
 	 * symbols already, no need to re-scan them after
@@ -252,14 +285,14 @@ static int __bwrite(struct bfd *bfd, const void *buf, int size)
 {
 	struct xbuf *b = &bfd->b;
 
-	if (b->sz + size > BUFSIZE) {
+	if (b->sz + size > b->bsize) {
 		int ret;
 		ret = bflush(bfd);
 		if (ret < 0)
 			return ret;
 	}
 
-	if (size > BUFSIZE)
+	if (size > b->bsize)
 		return write_all(bfd->fd, buf, size);
 
 	memcpy(b->data + b->sz, buf, size);
diff --git a/criu/include/bfd.h b/criu/include/bfd.h
index 2846ec628..050158327 100644
--- a/criu/include/bfd.h
+++ b/criu/include/bfd.h
@@ -8,6 +8,7 @@ struct xbuf {
 	char *mem;	 /* buffer */
 	char *data;	 /* position we see bytes at */
 	unsigned int sz; /* bytes sitting after b->pos */
+	unsigned int bsize;
 	struct bfd_buf *buf;
 };
 
```

criu/bfd.c
DongSunchao: Good point. I have moved the capacity tracking into `xbuf` as a `bsize` field. I also used `mremap` for subsquent extensions rather than `mmap + memcpy`.

test/zdtm/static/massive_groups.desc
avagin: >  The ZDTM sandbox initialization actually fails under these conditions (I encountered a mount(/proc) failed error when I tried running it in ns).

Have you tried to investigate why it fails?

DongSunchao: The issue I encountered was caused by a previous aborted run that left a `.constructed` marker file in the `test/` directory, though I'm not sure why the `run/` directory was lost. The script only checks for the `.constructed` to assume the sandbox is ready. I cleaned up my environment and the issue was resolved. I'm not sure if this edge case warrants a separate PR to improve the check.

By the way, `'flavor': 'h'` is still necessary. This is because the test explicitly relies on `setgroups()` to attach hundreds of arbitrary GIDs.

avagin: As minimum, it should work with the ns flavor. For uns, you will need to extend uid/gid mappings.

DongSunchao: I have updated the PR with a new commit to fully support the uns flavor.

To achieve this cleanly without polluting the global environment or breaking other tests, I added a new flag (`ext-uid-map`) in the `.desc file`. The Python framework detects this flag and injects a massive ID mapping (up to 4 billion) into the native test environment (`ZDTM_UID/GID_MAP`), which the C wrapper safely applies during namespace initialization.

Since the test now runs perfectly across all namespaces, I have removed the `flavor` restriction from the `.desc` file.

PR #2901, opened 2026-02-20
unix: fix dangling pointers in icon hash on error path

3idey:
Fixes a use-after-free risk in CRIU’s `AF_UNIX` socket collection by ensuring partially collected “in-flight connection” icons can’t leave dangling references in the global `unix_listen_icons` hash when allocation fails.

Approach:
Instead of traversing `unix_listen_icons` to remove icons on error, this patch pre-allocates all `unix_sk_listen_icon` nodes before inserting any into `unix_listen_icons`. If allocation fails, the temporary list is freed and the function aborts before mutating the global hash. This guarantees that the hash is only updated when all allocations succeed, and avoids the need for hash cleanup on error.

### DIFF
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -785,6 +785,7 @@ static int unix_collect_one(const struct unix_diag_msg *m, struct nlattr **tb, s
 
 	if (tb[UNIX_DIAG_ICONS]) {
 		unsigned int len = nla_len(tb[UNIX_DIAG_ICONS]);
+		struct unix_sk_listen_icon *icon_head = NULL;
 		unsigned int i;
 
 		d->icons = xmalloc(len);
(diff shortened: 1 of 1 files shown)


3idey:
@avagin could you take a look on this ?
thank you.

criu/sk-unix.c
```diff
@@ -832,6 +832,23 @@ static int unix_collect_one(const struct unix_diag_msg *m, struct nlattr **tb, s
 
 	return 0;
 err:
+	int j;
+	if (d->icons) {
+		for (j = 0; j < d->nr_icons; j++) {
```
avagin: I think we can consider to do something like this:
```
index 1d5764427..3d4cd39f2 100644
--- a/criu/sk-unix.c
+++ b/criu/sk-unix.c
@@ -785,6 +785,7 @@ static int unix_collect_one(const struct unix_diag_msg *m, struct nlattr **tb, s
 
        if (tb[UNIX_DIAG_ICONS]) {
                unsigned int len = nla_len(tb[UNIX_DIAG_ICONS]);
+               struct unix_sk_listen_icon *icon_head = NULL;
                unsigned int i;
 
                d->icons = xmalloc(len);
@@ -799,12 +800,27 @@ static int unix_collect_one(const struct unix_diag_msg *m, struct nlattr **tb, s
                 * to fix up in-flight sockets peers.
                 */
                for (i = 0; i < d->nr_icons; i++) {
-                       struct unix_sk_listen_icon *e, **chain;
-                       unsigned int n;
+                       struct unix_sk_listen_icon *e;
 
                        e = xzalloc(sizeof(*e));
-                       if (!e)
+                       if (!e) {
+                               while (icon_head) {
+                                       e = icon_head;
+                                       icon_head = e->next;
+                                       xfree(e);
+                               }
                                goto err;
+                       }
+                       e->next = icon_head;
+                       icon_head = e;
+
+               }
+               for (i = 0; i < d->nr_icons; i++) {
+                       struct unix_sk_listen_icon *e, **chain;
+                       unsigned int n;
+
+                       e = icon_head;
+                       icon_head = e->next;
 
                        n = d->icons[i];
                        chain = &unix_listen_icons[n % SK_HASH_SIZE];
```

imho it is simpler and more readable...

avagin:
> Signed-off-by: 3idey <elaidya225@gmail.com>

could you please use the real name

PR #2906, opened 2026-02-25
sk-queue: Add MSG_CTRUNC check and fix memory leaks

3idey:
This patch series fixes two categories of bugs in sk-queue.c:

**Commit 1: Add missing MSG_CTRUNC check in dump_sk_queue**

After recvmsg() with MSG_PEEK, dump_sk_queue() checks for MSG_TRUNC but not
MSG_CTRUNC. When the control message buffer is too small, the kernel silently
discards SCM data (e.g. passed file descriptors). This adds an explicit check
so dump fails instead of producing an incomplete image.

**Commit 2: Fix memory leaks in error paths**

- `collect_one_packet()`: Free `pkt->data` when `n_scm > 1` check fails
- `collect_one_packet()`: Move `list_add_tail()` after `read_img_buf()` so a packet with freed data is never left on the list
- `dump_scm_rights()`: Free `scme` when `dump_my_file()` fails
- `send_one_pkt()`: Free `pkt->scm` after `sendmsg()`

Tested with zdtm/static/scm00 through scm04 — all pass.

### DIFF
--- a/criu/sk-queue.c
+++ b/criu/sk-queue.c
@@ -244,6 +244,17 @@ int dump_sk_queue(int sock_fd, int sock_id)
 			ret = -E2BIG;
 			goto err_set_sock;
 		}
+		if (msg.msg_flags & MSG_CTRUNC) {
+			/*
+			 * Control data truncated. This means the cmsg
+			 * buffer was too small and SCM data (such as
+			 * passed file descriptors) has been silently
+			 * discarded by the kernel.
+			 */
+			pr_err("sys_recvmsg failed: control data truncated\n");
+			ret = -E2BIG;
+			goto err_set_sock;
+		}
 
 		if (dump_packet_cmsg(&msg, &pe))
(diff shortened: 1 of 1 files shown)


3idey:
I previously used my GitHub username but was asked by @avagin to use my real name in the Signed-off-by on a prior PR #2901, so the author and Signed-off-by now match.

criu/sk-queue.c
avagin, line 113: should we call xfree(scme) here as well?

avagin:
> I previously used my GitHub username but was asked by @avagin to use my real name in the Signed-off-by on a prior PR #2901, so the author and Signed-off-by now match.

Signed-off-by should match the author name and both should be the real author name.

criu/sk-queue.c
3idey: I added this thanks.

3idey:
@avagin Could you take a final look?

PR #2913, opened 2026-03-02
criu: save and restore timer_slack_ns per thread

3idey:
## Problem

Linux has a per-thread `timer_slack_ns` value that controls wakeup
coalescing for `select()`, `poll()`, `nanosleep()`, and similar calls.
It defaults to 50000 ns and is set via `prctl(PR_SET_TIMERSLACK)`.

CRIU does not save or restore this value, so any process that
customises its timer slack (e.g. latency-sensitive applications that
set it to 0, or battery-saving daemons that increase it) silently
reverts to the default after checkpoint/restore.

## Solution

Collect and restore `timer_slack_ns` per thread, following the same
pattern used for `pdeath_sig`:

1. **Image schema** — add `optional uint64 timerslack_ns = 17` to
   `thread_core_entry` in `images/core.proto`.

2. **Dump** — read the value via `prctl(PR_GET_TIMERSLACK)` inside
   the parasite and write it into the protobuf image during dump.

3. **Restore** — read the value from the image, pass it via
   `struct thread_restore_args`, and call `prctl(PR_SET_TIMERSLACK)`
   in the restorer for each thread.

4. **Test** — add a ZDTM test (`zdtm/static/timerslack_ns`) that sets
   a non-default value, checkpoints, and verifies it is preserved.

## Before (criu-dev)

```
$ sudo ./test/zdtm.py run -t zdtm/static/timerslack_ns --ignore-taint
zdtm/static/timerslack_ns                       FAIL
```

Test binary does not exist; `timer_slack_ns` is silently lost on C/R.

## After (this PR)

```
$ sudo ./test/zdtm.py run -t zdtm/static/timerslack_ns --ignore-taint
Test zdtm/static/timerslack_ns PASS (h)
Test zdtm/static/timerslack_ns PASS (ns)
Test zdtm/static/timerslack_ns PASS (uns)
```

All three flavours pass — `timer_slack_ns` is correctly preserved.

## Commits

Each commit builds and runs independently (bisect-safe):

1. `images:` add schema field
2. `criu:` dump — parasite collection + `cr-dump.c` wiring
3. `criu:` restore — restorer function + `cr-restore.c` wiring
4. `zdtm:` test

### DIFF
--- a/criu/pie/parasite.c
+++ b/criu/pie/parasite.c
@@ -41,6 +41,10 @@ static struct parasite_dump_pages_args *mprotect_args = NULL;
 #define PR_GET_PDEATHSIG 2
 #endif
 
+#ifndef PR_GET_TIMERSLACK
+#define PR_GET_TIMERSLACK 30
+#endif
+
 #ifndef PR_GET_CHILD_SUBREAPER
 #define PR_GET_CHILD_SUBREAPER 37
 #endif
@@ -192,6 +196,16 @@ static int dump_thread_common(struct parasite_dump_thread *ti)
 	if (ret) {
 		pr_err("Unable to get the parent death signal: %d\n", ret);
 		goto out;
+	}
+
+	{
+		long slack = sys_prctl(PR_GET_TIMERSLACK, 0, 0, 0, 0);
+		if (slack < 0) {
+			pr_err("Unable to get timer slack: %ld\n", slack);
+			ret = (int)slack;
+			goto out;
+		}
+		ti->timerslack_ns = (unsigned long)slack;
 	}
 
 	ret = sys_prctl(PR_GET_NAME, (unsigned long)&ti->comm, 0, 0, 0);
--- a/test/zdtm/static/timerslack_ns.c
+++ b/test/zdtm/static/timerslack_ns.c
@@ -1,0 +1,90 @@
+#include <stdlib.h>
+#include <stdint.h>
+#include <pthread.h>
+#include <sys/prctl.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check that timer_slack_ns is preserved across C/R for multiple threads";
+const char *test_author = "Ahmed Elaidy <elaidya225@gmail.com>";
+
+#define MAIN_SLACK_NS	 123456UL
+#define THREAD_SLACK_NS	 654321UL
+
+static void *thread_fn(void *arg)
+{
+	long slack;
+
+	(void)arg;
+
+	if (prctl(PR_SET_TIMERSLACK, THREAD_SLACK_NS, 0, 0, 0)) {
+		pr_perror("thread: can't set timer_slack_ns");
+		return (void *)(uintptr_t)1;
+	}
+
+	test_waitsig();
+
+	slack = prctl(PR_GET_TIMERSLACK, 0, 0, 0, 0);
+	if (slack < 0) {
+		pr_perror("thread: can't get timer_slack_ns");
+		return (void *)(uintptr_t)1;
+	}
+
+	if (slack != THREAD_SLACK_NS) {
+		pr_err("thread: timer_slack_ns changed: expected %lu, got %ld\n",
+		       THREAD_SLACK_NS, slack);
+		return (void *)(uintptr_t)1;
+	}
+
+	return NULL;
+}
+
+int main(int argc, char **argv)
+{
+	pthread_t thread;
+	void *thread_ret;
+	long slack;
+	int ret;
+
+	test_init(argc, argv);
+
+	if (prctl(PR_SET_TIMERSLACK, MAIN_SLACK_NS, 0, 0, 0)) {
+		pr_perror("can't set timer_slack_ns");
+		return 1;
+	}
+
+	ret = pthread_create(&thread, NULL, thread_fn, NULL);
+	if (ret) {
+		pr_err("pthread_create: %d\n", ret);
+		return 1;
+	}
+
+	test_daemon();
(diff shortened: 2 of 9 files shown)


criu/pie/parasite.c
```diff
@@ -194,6 +198,14 @@ static int dump_thread_common(struct parasite_dump_thread *ti)
 		goto out;
 	}
 
+	ret = sys_prctl(PR_GET_TIMERSLACK, 0, 0, 0, 0);
+	if (ret < 0) {
+		pr_err("Unable to get timer slack: %d\n", ret);
+		goto out;
+	}
+	ti->timerslack_ns = (unsigned long)ret;
+	ret = 0;
```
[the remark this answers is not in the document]
3idey: `PR_GET_TIMERSLACK` has been available since Linux 2.6.28, and CRIU already requires kernel 3.11+ (due to `PTRACE_SEIZE`, `kcmp`, etc.).  so i think this will be misleading because this is an impossible case.

test/zdtm/static/timerslack_ns.c
```diff
@@ -0,0 +1,46 @@
...
+
+	slack = prctl(PR_GET_TIMERSLACK, 0, 0, 0, 0);
+	if (slack < 0) {
+		pr_perror("Can't get timer_slack_ns");
+		exit(1);
+	}
+
+	if ((unsigned long)slack != TEST_SLACK_NS) {
+		fail("timer_slack_ns changed: expected %lu, got %lu",
+		     TEST_SLACK_NS, (unsigned long)slack);
```
avagin: return 1;

then the next else will be unneeded...

test/zdtm/static/timerslack_ns.c
```diff
@@ -0,0 +1,46 @@
...
+
+#define TEST_SLACK_NS 123456789UL
+
+int main(int argc, char **argv)
+{
+	long slack;
+
+	test_init(argc, argv);
+
+	if (prctl(PR_SET_TIMERSLACK, TEST_SLACK_NS, 0, 0, 0)) {
```
avagin: you need to create a second thread and verify that a timerslack value will be restored for it too.

test/zdtm/static/timerslack_ns.c
```diff
@@ -0,0 +1,46 @@
...
+	test_daemon();
+	test_waitsig();
+
+	slack = prctl(PR_GET_TIMERSLACK, 0, 0, 0, 0);
+	if (slack < 0) {
+		pr_perror("Can't get timer_slack_ns");
+		exit(1);
+	}
+
+	if ((unsigned long)slack != TEST_SLACK_NS) {
```
avagin: why do we need this type conversion?

test/zdtm/static/timerslack_ns.c
```diff
@@ -0,0 +1,46 @@
+#include <stdlib.h>
+#include <sys/prctl.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check that timer_slack_ns is preserved across C/R";
+const char *test_author = "Ahmed Elaidy <elaidya225@gmail.com>";
+
+#ifndef PR_SET_TIMERSLACK
+#define PR_SET_TIMERSLACK 29
```
avagin: I think PR_SET_TIMERSLACK was added a long time ago, are you sure they can be undefined?

test/zdtm/static/timerslack_ns.c
3idey: Fixed. I removed the unnecessary `else` branch and made the failure path return directly.

test/zdtm/static/timerslack_ns.c
3idey: Done. I updated `timerslack_ns` to create a secondary thread, set a distinct `timer_slack_ns` in that thread, checkpoint/restore, and verify both:
- main thread value (`MAIN_SLACK_NS`)
- secondary thread value (`THREAD_SLACK_NS`)

I also synchronized the check so the secondary thread verifies its value after restore.

test/zdtm/static/timerslack_ns.c
3idey: Yes we don’t need it. I now compare directly using `long` from `prctl(PR_GET_TIMERSLACK)`.

test/zdtm/static/timerslack_ns.c
```diff
@@ -0,0 +1,98 @@
...
+	(void)arg;
+
+	if (prctl(PR_SET_TIMERSLACK, THREAD_SLACK_NS, 0, 0, 0)) {
+		pr_perror("thread: can't set timer_slack_ns");
+		thread_result = -1;
+		return NULL;
+	}
+
+	while (!__atomic_load_n(&thread_can_check, __ATOMIC_ACQUIRE))
+		usleep(1000);
```
avagin: Why you don't use test_waitsig here?

test/zdtm/static/timerslack_ns.c
```diff
@@ -0,0 +1,98 @@
...
+	if (slack < 0) {
+		pr_perror("thread: can't get timer_slack_ns");
+		thread_result = -1;
+		return NULL;
+	}
+
+	if (slack != THREAD_SLACK_NS) {
+		pr_err("thread: timer_slack_ns changed: expected %lu, got %ld\n",
+		       THREAD_SLACK_NS, slack);
+		thread_result = -1;
```
avagin: why do we need a global variable for the return code?

3idey:
All comments addressed @avagin 
Could u take a look ?

Issue #2911, opened 2026-03-02
too many ci tests are not stable


avagin:
https://github.com/checkpoint-restore/criu/actions/workflows/archlinux-test.yml
https://github.com/checkpoint-restore/criu/actions/workflows/gcov-test.yml

PR #2914, opened 2026-03-02
ci: consolidate workflows into two-stage pipeline

### LINKED ISSUE #2912: Multi-stage CI
Some resources, like our ARM runners, are limited, and we often exhaust our quotas too quickly. I think we should rework our GitHub CI jobs into stages: the first stage should contain basic, absolutely stable tests and checks, while all other tasks should move to a second stage that only runs if the first stage succeeds.

3idey:
Fixes: #2912


3idey:
The Stage 1 lint failure is not caused by this PR — it is a pre-existing
issue on criu-dev introduced by commit 555b257c5e77 ("plugin/amdgpu: Catch
error for failure to open drm device"), which passes \n to pr_perror. That
macro appends its own newline, so the explicit \n violates the lint rule and
causes CI to fail on any PR currently branched off criu-dev, not just this one.

I have submitted a separate one-line fix in PR #2917.

Ironically, this failure demonstrates that the two-stage pipeline is working
as intended — lint caught the error in Stage 1 and all ARM runners (Stage 2)
were never started, which is exactly the quota-saving behaviour this PR aims
to achieve.

Once the lint fix is merged into criu-dev I will rebase this branch and the
full CI run should go green end-to-end.
@avagin

3idey:
@avagin ,The failing checks are all pre-existing and unrelated to this PR:

- Arch Linux and Coverage Tests are acknowledged as flaky in #2911 (i will make another pr once this is merged)  
- mips64el cross-compile is a known toolchain issue
- Fedora Rawhide / Vagrant Rawhide are infrastructure flakiness

All Stage 1 jobs passed (Lint, X86_64 GCC/CLANG, gate, check-commits).
Stage 2 ran correctly with the majority passing, including all aarch64
ARM jobs, Alpine, Docker, ASAN, Java, Podman, Nftables, and LoongArch64.

avagin:
@3idey I haven't seen this pr and pushed a similar one https://github.com/checkpoint-restore/criu/pull/2919. I think all lint-like jobs should be standalone. The jobs running tests should be gated by one which is stable and relatively fast.

3idey:
Thanks @avagin for the feedback and for working on #2919! Agreed — Alpine is a better gate than x86 GCC/CLANG since it's stable, fast, and actually runs the test suite. I also see the point about lint-like jobs being standalone rather than part of the gate; tying check-commits into the gate complicates the skipped-on-push edge case unnecessarily. Happy to close #2914 in favor of #2919 if that works for you.
and if you want I will also update this to match your design if you could take a look.

rst0git:
@3idey Thank you for your help! I went ahead and merged https://github.com/checkpoint-restore/criu/pull/2919

PR #2917, opened 2026-03-02
plugin/amdgpu: fix pr_perror trailing newline

3idey:
pr_perror already appends a newline; the explicit \n in the format string violates the lint rule and breaks CI.

Fixes: 555b257c5e77 ("plugin/amdgpu: Catch error for failure to open drm device")


plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1909,7 +1908,7 @@ int amdgpu_plugin_restore_file(int id, bool *retry_needed)
 			fd = ret;
 			goto fail;
 		}
-	fail:
+fail:
```
avagin: please remove all unrelated changes from this patch.

3idey: Done.

PR #2919, opened 2026-03-03
ci: Consolidate test workflows and gate them by Alpine Test

avagin:
Rework GitHub workflows to consolidate all test-related jobs into a single ci.yml file. This ensures that the alpine-test job runs first, and all other tests are executed only if it passes.

This reduces redundant triggers and provides a clear dependency path for CI runs.


Snorch (approved):
Looks good.

rst0git (approved):
LGTM

PR #2920, opened 2026-03-03
rseq: use kernel rseq.h when glibc detects it

adrianreber:
Compilation fails with the latest glibc-devel in Fedora rawhide. Recent glibc conditionally includes the kernel's linux/rseq.h when __GLIBC_HAVE_KERNEL_RSEQ is defined, but CRIU's own copy of the rseq definitions does not account for this and conflicts with the kernel header which has additional information.

Check if __GLIBC_HAVE_KERNEL_RSEQ is defined and if so include the kernel header directly using angle brackets (which bypasses -iquote and finds the real kernel header instead of CRIU's copy).  Otherwise fall back to CRIU's own definitions.

### DIFF
--- a/criu/include/linux/rseq.h
+++ b/criu/include/linux/rseq.h
@@ -14,6 +14,16 @@
 
 #include "common/config.h"
 
+/*
+ * If glibc detected that the kernel rseq.h header exists, include
+ * it directly using angle brackets (which bypasses -iquote and finds
+ * the real kernel header).  Otherwise fall back to our own copy of
+ * the definitions.
+ */
+#ifdef __GLIBC_HAVE_KERNEL_RSEQ
+#include <linux/rseq.h>
+#else /* !__GLIBC_HAVE_KERNEL_RSEQ */
+
 #ifdef CONFIG_HAS_NO_LIBC_RSEQ_DEFS
 /*
  * linux/rseq.h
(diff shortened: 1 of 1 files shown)


adrianreber:
CC: @mihalicyn

avagin:
I don't see any recent changes there. Did I miss something?
```
95e114a0919 (Florian Weimer  2021-12-09 09:49:32 +0100)|#ifdef __GLIBC_HAVE_KERNEL_RSEQ
95e114a0919 (Florian Weimer  2021-12-09 09:49:32 +0100)|/* We use the structures declarations from the kernel headers.  */
95e114a0919 (Florian Weimer  2021-12-09 09:49:32 +0100)|# include <linux/rseq.h>
95e114a0919 (Florian Weimer  2021-12-09 09:49:32 +0100)|#else /* __GLIBC_HAVE_KERNEL_RSEQ */
```

avagin:
We have the feature check https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/feature-tests.mak#L161. Why doesn't it work?

avagin:
btw: https://github.com/avagin/criu/commit/ae53b51bab6c1c8bd721265306455b6ff6010f86 seems fixing the issue, but I still don't understand why...

avagin:
I think gcc doesn't report a warning/error if there are two identical enum-s.  rseq.h was recently changed by https://github.com/torvalds/linux/commit/d7a5da7a0f7fa7ff081140c4f6f971db98882703#diff-d11bc1d6b40e9430965f4918c7fd81dd619972806ef5dbc43056162ceb4a9966. rseq_cpu_id_state hasn't been changed and this is why our feature check doesn't work as expected. If I am right, we need to do something like https://github.com/avagin/criu/commit/3db1cb11f7a476a3d731d231161eb3f6dfbe7a70.

PR #2923, opened 2026-03-03
restore: make a runc poc

everzakov:
The Pull request is a part of runc PoC (link) to create devices for user namespace containers. It adds if conditions to dump / restore /dev in the initial user namespace. Also, in the usernsd we need to find info about mount id that's why usernsd is created after loading mount info (we are stricted to the data length transfer to the usernsd). 

Current PoC limitations are:
1) The devices major/minor should be the same for checkpoint/restore.
2) The user namespace info (host id, container id, length) should be the same for checkpoint/restore.
3) Only simple mount scenarios are checked (e.g. not checked if user will mount container dev to some path).

If I understand correctly if we want to make it more general solution then we need to understand that the mount is created in the initial user namespace not in container user namespace. There is a syscall statmount to retieve idmappings https://lwn.net/Articles/1007019/ However, the /dev is not idmapped that's why it will return the empty list and will be equal to any mount created in the container user namespace. I think that it is possible to create the similiar syscall statmount to retrive mount user namespace mapping.


criu/filesystems.c
```diff
@@ -480,6 +480,11 @@ static int tmpfs_dump(struct mount_info *pm)
 	if (root_ns_mask & CLONE_NEWUSER)
 		userns_pid = root_item->pid->real;
 
+	// We need to dump tmpfs in initial user namespace to save device owner mappings.
+	if (strcmp(pm->mountpoint, "./dev") == 0) {
+		userns_pid = -1;
```
Snorch, line 485: What exact error do we get if we don't have this hunk? I just wonder what if instead we can restore devtmpfs from the correct user namespace in tmpfs_restore (upd: as far as I can see we already do before patch)?

note: We can probably improve filtering for devtmpfs by checking `pm->fstype.code == FSTYPE__DEVTMPFS`.

Snorch: If this tmpfs was originally created inside the container user namespace there should not be any problem with id mapping right?

So likely you create /dev outside container, this may mean that it should be treated as external mount. E.g.: on dump --external mnt[/dev]:DEV, on restore --external mnt[DEV]:/path/to/it/on/host .

PR #2925, opened 2026-03-03
scripts: rseq: fix detection of rseq_cpu_id_state

avagin:
gcc doesn't report any warnings if there are two identical enums with the same name. Actually, we wanted to detect if a specific enum is already exists. By using a different member name in the test, we ensure that a conflict occurs if the enum is already defined in the system headers.


adrianreber (approved):

mihalicyn (commented):
LGTM, thanks!

PR #2928, opened 2026-03-03
criu: fix double-open of userns image in --stream mode

### LINKED ISSUE #2924: `--stream` mode broken since binfmt_misc sandboxing support (e6510a338)
Commit e6510a338 ("criu: Support binfmt_misc sandboxing") breaks restore when using --stream mode (criu-image-streamer).                                                                                                                    
                                                                                                                                                                                                                                              
  The new restore_userns_binfmt_misc() function opens userns-<id>.img a second time to read binfmt_misc entries, but prepare_userns() has already opened and consumed the same file. With regular file-based images this is fine, but         
  criu-image-streamer removes files from memory after the first transfer and rejects duplicate requests:                                                                                                                                      
                                                                                                                                                                                                                                              
  criu-image-streamer Error: CRIU is requesting the image file `userns-15.img`                                                                                                                                                                
  multiple times. This is not allowed to keep the memory usage low                                                                                                                                                                            
                                                                                                                                                                                                                                              
  Reproducer:                                                                                                                                                                                                                                 
  ./test/zdtm.py run --stream -t zdtm/static/env00 --ignore-taint                                                                                                                                                                             
                                                                                                                                                                                                                                              
  Fails in the uns (user namespace) mode. The test passes on the parent commit ecab9e357.

3idey:
Fix `--stream` mode restore broken by e6510a338 ("criu: Support
binfmt_misc sandboxing").

`restore_userns_binfmt_misc()` was opening the userns image a second
time, but `criu-image-streamer` rejects duplicate file requests:

  criu-image-streamer Error: CRIU is requesting the image file
  userns-15.img multiple times. This is not allowed to keep the
  memory usage low

Fix by pre-reading the image in a new `read_user_ns_img()` helper,
called from `prepare_namespace_before_tasks()` which runs before the
first `clone()`. The `UsernsEntry` is cached in a module-level static
pointer (`userns_restore_info`). Both the CRIU coordinator
(`prepare_userns`) and the forked root task
(`restore_userns_binfmt_misc`) access the cache via COW semantics ---
no second `open_image()` call is needed.

Tested with:
  sudo ./test/zdtm.py run --stream -t zdtm/static/env00 --ignore-taint
  sudo ./test/zdtm.py run --stream -t zdtm/static/env00 --ignore-taint -f uns

Both pass with the fix applied. Both failed before it (the `uns` variant
with the streamer duplicate-file error from the issue).

Closes #2924

### DIFF
--- a/criu/include/namespaces.h
+++ b/criu/include/namespaces.h
@@ -7,6 +7,7 @@
 #include "files.h"
 #include "common/list.h"
 #include "images/netdev.pb-c.h"
+#include "images/userns.pb-c.h"
 
 #ifndef CLONE_NEWNS
 #define CLONE_NEWNS 0x00020000
@@ -138,6 +139,10 @@ struct ns_id {
 			struct list_head links;
 			NetnsEntry *netns;
 		} net;
+
+		struct {
+			UsernsEntry *e;
+		} user;
 	};
 };
 extern struct ns_id *ns_ids;
--- a/criu/namespaces.c
+++ b/criu/namespaces.c
@@ -1581,19 +1581,49 @@ int stop_usernsd(void)
 	return ret;
 }
 
-int prepare_userns(struct pstree_item *item)
+static int read_user_ns_img(void)
 {
+	struct ns_id *ns;
 	struct cr_img *img;
-	UsernsEntry *e;
 	int ret;
 
-	img = open_image(CR_FD_USERNS, O_RSTR, item->ids->user_ns_id);
+	if (!(root_ns_mask & CLONE_NEWUSER))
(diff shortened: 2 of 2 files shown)


avagin:
we usually save per-ns things on ns_id structures. Why should we not do that in this case too?

3idey:
> we usually save per-ns things on ns_id structures. Why should we not do that in this case too?

Good point @avagin .I updated it to store `UsernsEntry *e` on the `ns_id` union, following the same pattern as `net.netns`. The module-level static is gone.

Snorch:
Looks good, thank you!

Issue #2007, opened 2022-12-01
support of different IP socket options


Snorch:
**Description**
There are multiple ip socket options which are not dumped by criu, e.g.:

```
#define IP_TOS		1
#define IP_TTL		2
#define IP_HDRINCL	3
#define IP_OPTIONS	4
#define IP_ROUTER_ALERT	5
#define IP_RECVOPTS	6
#define IP_RETOPTS	7

#define IP_PKTOPTIONS	9
#define IP_MTU_DISCOVER	10
#define IP_RECVERR	11
#define IP_RECVTTL	12
#define	IP_RECVTOS	13
#define IP_MTU		14

#define IP_IPSEC_POLICY	16
#define IP_XFRM_POLICY	17
#define IP_PASSSEC	18
#define IP_TRANSPARENT	19

#define IP_ORIGDSTADDR       20

#define IP_MINTTL       21
#define IP_NODEFRAG     22
#define IP_CHECKSUM	23
#define IP_BIND_ADDRESS_NO_PORT	24
#define IP_RECVFRAGSIZE	25
#define IP_RECVERR_RFC4884	26
```

See full list in kernel source: https://github.com/torvalds/linux/blob/4bbf3422df78029f03161640dcb1e9d1ed64d1ea/include/uapi/linux/in.h#L97

Currently we only dump IP_FREEBIND and plan to dump IP_PKTINFO https://github.com/checkpoint-restore/criu/pull/2006. We need to investigate other options and also support them in CRIU if they are needed.

rahulk789:
Hey! I would love to work on this. Are there any updates that you'd like to share other than the ones mentioned? Thank you!

Snorch:
Hi, note that in this issue important part is to understand those options: what do they do? why they are needed? can they break something if we just get/set them as is?

Another thing to remember is that always need a good regression test in zdtm for each option we support.

You can use my PR as an example of how one option can be supported here https://github.com/checkpoint-restore/criu/pull/2006 Though it is not yet merged.

rahulk789:
Thank you for the information!

PR #2931, opened 2026-03-03
support IPV6_TCLASS, IPV6_UNICAST_HOPS and IP_CHECKSUM options

TusharMohapatra07:
IPv6 sockets have TOS and TTL equivalents (`IPV6_TCLASS` and `IPV6_UNICAST_HOPS`) that were not being dumped or restored. Add support for both options in the `AF_INET6` branch of `dump_ip_opts()` and `restore_ip_opts()`, reusing the existing `ip_opts_entry.tos` and `ip_opts_entry.ttl` proto fields which are already used for the IPv4 `IP_TOS` and `IP_TTL` counterparts.

Additionally, `ip_opts_raw_entry` already had a `checksum` field defined in the protobuf schema for `IP_CHECKSUM`, but the corresponding dump and restore calls were never implemented. Wire up `IP_CHECKSUM` in `dump_ip_raw_opts()` and `restore_ip_opts()` for `AF_INET` raw sockets.

Part of #2007


TusharMohapatra07:
cc @Snorch @avagin

PR #2932, opened 2026-03-04
scripts: improve layer caching of container builds

### LINKED ISSUE #2926: Container images are fully rebuilt after any change in the repo
https://github.com/checkpoint-restore/criu/commit/25f8be0f6016bd6ef0e0a1222cd3fbfca8e0b6fd introduced a serious issue. Now all images are rebuilt after any changes in the repo. `COPY ./criu` should be done at the very end and definitely after installing all packages

rst0git:
The scripts installing dependencies are now copied and executed before copying the full source tree. This allows to cache the dependency installation layers as long as these scripts remain unchanged. Previously, copying the entire repository invalidated the cache on every source change. While our CI intentionally doesn't preserve cached layers across independent runs, the changes in this patch help with local testing when investigating issues.

Fixes: #2926

### DIFF
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -1,4 +1,4 @@
-FROM alpine
+FROM docker.io/library/alpine:latest
 ARG CC=gcc
 
 COPY contrib/dependencies/apk-packages.sh /criu/contrib/dependencies/apk-packages.sh
--- a/scripts/build/Dockerfile.alpine
+++ b/scripts/build/Dockerfile.alpine
@@ -1,10 +1,11 @@
 FROM alpine
 ARG CC=gcc
 
+COPY contrib/dependencies/apk-packages.sh /criu/contrib/dependencies/apk-packages.sh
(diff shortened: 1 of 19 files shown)


scripts/build/Dockerfile.alpine
```diff
@@ -1,14 +1,13 @@
...
+
 COPY . /criu
 WORKDIR /criu
 
-RUN apk add --no-cache "$CC" && /criu/contrib/dependencies/apk-packages.sh
+RUN make mrproper && make -j $(nproc) CC="$CC"
 
-RUN make mrproper && date && make -j $(nproc) CC="$CC" && date
-
-# The rpc test cases are running as user #1000, let's add the user
```
avagin, line 11: nit: do we really want to delete this comment?

rst0git: Thank you for pointing it out! I somehow missed this.

avagin:
@rst0git could you look at ci failures

rst0git:
> @rst0git could you look at ci failures

The previous failures might have been unrelated to the changes in this pull request. I rebased on the `criu-dev` branch.

PR #2934, opened 2026-03-04
Contrib/void linux dependencies

Abdullah-Badawy1:
- Void Linux was not covered in `contrib/dependencies/`. This PR adds `contrib/dependencies/xbps-packages.sh` following the same structure as the existing scripts, and updates `CONTRIBUTING.md` to reference to existing Arch script because it widely used by programmers, adding void installations makes the CONTRIBUTING file too long.
- Because Arch is similar to Void, I tried to run and install all packages, but many failed because the xbps repos do not have them all with the same name, so I did some research and added the appropriate ones.

**Notes on Void Linux package naming:**
- Development headers are in separate `-devel` packages
- `util-linux` is split on Void — only `libuuid-devel` is needed
- `nftables` is split on Void — only `libnftables-devel` is needed
- `protobuf` must be installed alongside `protobuf-devel` to get the `protoc` compiler binary

Void Linux required special attention due to package splitting:
- `libuuid-devel` instead of `util-linux-devel`
- `libnftables-devel` instead of `nftables-devel`
- `protobuf` must be installed alongside `protobuf-devel` to get the `protoc` compiler binary

**Testing:** Tested on my current Void machine and in a Docker container.
- All packages installed successfully
- All libraries verified via `pkg-config`
- `./criu/criu check` reports `Looks good.`
image
image
image

Signed-off-by: Abdullah Albadawy <abdullahalbadawy1@gmail.com>

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -58,6 +58,14 @@ First, you need to install the required build dependencies. We provide scripts t
 nix develop
 ```
 
+##### Using Arch:
+
+```
+./contrib/dependencies/pacman-packages.sh
+```
+```
+
+
 #### Compiling CRIU
 
 Once the dependencies are installed, you can compile CRIU by running the `make` command from the root of the source directory:
--- a/contrib/dependencies/xbps-packages.sh
+++ b/contrib/dependencies/xbps-packages.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+
+sudo xbps-install -Sy \
+    gcc \
+    make \
+    git \
+    pkg-config \
+    binutils \
+    protobuf \
+    protobuf-devel \
+    protobuf-c-devel \
+    python3 \
+    python3-protobuf \
+    python3-setuptools \
+    python3-pip \
+    python3-yaml \
+    libnet-devel \
+    libnl3-devel \
+    libcap-devel \
+    libaio-devel \
+    gnutls-devel \
+    libbsd-devel \
+    libuuid-devel \
+    elfutils-devel \
+    libbpf-devel \
+    libdrm-devel \
+    libnftables-devel \
+    libselinux-devel \
+    iproute2 \
+    libtraceevent-devel \
+    libtracefs-devel \
+    xmlto \
+    asciidoc


contrib/dependencies/xbps-packages.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+
+sudo xbps-install -Sy \
```
rst0git, line 3: @Abdullah-Badawy1 We usually add these install scripts so we can enable testing for specific Linux distribution in our CI:
- https://github.com/checkpoint-restore/criu/tree/criu-dev/scripts/build
- https://github.com/checkpoint-restore/criu/blob/criu-dev/.github/workflows/ci.yml

@avagin Would it make sense to enable CI testing with Void Linux?

Abdullah-Badawy1: Thank you @rst0git  for the review! I looked at scripts/build and ci.yml.
I am happy to add Void Linux CI support alongside the dependency
script. Would you be able to confirm if adding Void Linux to the
CI makes sense before I proceed? I want to make sure the effort
is worthwhile before expanding the scope of this PR.

rst0git: >I want to make sure the effort is worthwhile before expanding the scope of this PR.

If we don't add tests, the installation script could break in the future without us noticing :)

avagin: I don't think that we need to test all existing distros. We tests ubuntu (most popular one), fedora (cutting edge), alpine (musl), rhel/centos (enterprise/legacy ;)). If someone think we need to test on something else, they need to provide strong reasons for that.

CONTRIBUTING.md
```diff
@@ -58,6 +58,14 @@ First, you need to install the required build dependencies. We provide scripts t
 nix develop
 ```
 
+##### Using Arch:
+
+```
+./contrib/dependencies/pacman-packages.sh
+```
+```
```
rst0git, line 66: It looks like there is one extra ``` here

Abdullah-Badawy1: Thank you for the feedback.

Issue #2938, opened 2026-03-05
restore: user ns image is not read


everzakov:
In runc i have tried to create simple https://github.com/opencontainers/runc/issues/5121 to checkpoint / restore container with user namespace and host id not equal to 0. However i have got the restore error while trying to [read](https://github.com/checkpoint-restore/criu/blob/criu-dev/criu/namespaces.c#L1593) the user ns image.

```bash
not ok 11 checkpoint userns container
# (in test file tests/integration/userns.bats, line 302)
#   `[ "$status" -eq 0 ]' failed
# runc spec (status=0)
#
# skipping file /home: cannot remap user 65534:65534 -> -1:-1
# runc run -d --console-socket /tmp/bats-run-4jW2dq/runc.Sb1bS2/tty/sock test_busybox (status=0)
#
# runc state test_busybox (status=0)
# {
#   "ociVersion": "1.3.0",
#   "id": "test_busybox",
#   "pid": 7072,
#   "status": "running",
#   "bundle": "/tmp/bats-run-4jW2dq/runc.Sb1bS2/bundle",
#   "rootfs": "/tmp/bats-run-4jW2dq/runc.Sb1bS2/bundle/rootfs",
#   "created": "2026-03-05T18:40:23.511068887Z",
#   "owner": ""
# }
# runc checkpoint --work-path ./work-dir test_busybox (status=0)
#
# runc state test_busybox (status=1)
# time="2026-03-05T18:40:24Z" level=error msg="container does not exist"
# runc restore -d --work-path ./work-dir --console-socket /tmp/bats-run-4jW2dq/runc.Sb1bS2/tty/sock test_busybox (status=1)
# time="2026-03-05T18:40:24Z" level=warning msg="--- Quoting \"work-dir/restore.log\""
# time="2026-03-05T18:40:24Z" level=warning msg="282:(00.009265) mnt-v2: Inspecting sharing on 365 shared_id 0 master_id 2 (@/proc/timer_list)"
# time="2026-03-05T18:40:24Z" level=warning msg="283:(00.009267) mnt-v2: Detected external slavery for shared group (0, 2) with source /dev/null"
# time="2026-03-05T18:40:24Z" level=warning msg="284:(00.009269) mnt: Mountpoint 483 (@/) moved to the root yard"
# time="2026-03-05T18:40:24Z" level=warning msg="285:(00.009277) No netns-10.img image"
[...]

PR #2940, opened 2026-03-05
pstree: restore all namespace types during pstree image reading

TusharMohapatra07:
While exploring issue #2938 which was later solved by https://github.com/checkpoint-restore/criu/commit/09134c8ca8ea1889e7d45f8fc439a9c6812687f4, I found out that while dumping, CRIU saves a total of 8 namespace IDs (ref: https://github.com/checkpoint-restore/criu/blob/09134c8ca8ea1889e7d45f8fc439a9c6812687f4/criu/namespaces.c#L694)  while during restore, CRIU only reads 4 (after the fix) out of these 8 namespace IDs.

This PR adds handling for the 4 missing namespaces types in `read_pstree_ids()`:
- `uts_ns_id`: UTS namespace
- `ipc_ns_id`: IPC namespace
- `cgroup_ns_id`: Cgroup namespace
- `time_ns_id`: Time namespace

This creates symmetry between dump and restore paths, ensuring all namespace IDs saved during checkpoint are properly restored.

### DIFF
--- a/criu/pstree.c
+++ b/criu/pstree.c
@@ -15,6 +15,10 @@
 #include "dump.h"
 #include "util.h"
 #include "net.h"
+#include "uts_ns.h"
+#include "ipc_ns.h"
+#include "cgroup.h"
+#include "timens.h"
 
 #include "protobuf.h"
 #include "images/pstree.pb-c.h"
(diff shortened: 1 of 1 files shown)


TusharMohapatra07:
cc: @avagin @rst0git

rst0git:
>This creates symmetry between dump and restore paths, ensuring all namespace IDs saved during checkpoint are properly restored.

@TusharMohapatra07 I am trying to understand why this change is needed - are there any tests or use-cases where namespace IDs are not properly restored?

TusharMohapatra07:
Not really, to be honest. I just noticed the asymmetry while exploring issue 2938 because that helps me understand the codebase better and thought this PR would help. But if there's no usecase for this, I'll close the PR. Thanks a lot for reviewing :)

Snorch:
I don't think we really need this, at least until we have multiple uts/ipc/cgroup/time namespaces supported in dumped container, for now if we do so, all tasks would have same ids to add =) . I can imagine how it can be helpful to detect image inconsistency cases where we have a task with namespace id, but we don't have namespace for this id or something like that.

TusharMohapatra07:
Makes total sense. I initially thought of discussing the idea on Gitter, but since the changes were very simple, I decided to open a PR instead so the discussion could be more streamlined. Once again, Thanks for the review @rst0git @Snorch  :)

PR #2942, opened 2026-03-06
cr-service: generate uuid only once

rst0git:
When checkpointing containers, the function `util_init()` appears to be called twice, once in `crtools.c` then again in `cr-service.c`:

```
(00.000000) Parsing config file /etc/criu/default.conf
(00.000000) Unable to get $HOME directory, local configuration file will not be used.
(00.000000) CRIU run id = 1c9f31c3-aa70-4e8b-9b0c-19410efd92fe
(00.000000) CRIU run id = f19c528e-593c-4f29-b8e5-a2ff9c6b4d73
(00.000000) Would overwrite RPC settings with values from /etc/criu/runc.conf
(00.000000) Parsing config file /etc/criu/runc.conf
(00.000082) Version: 4.2 (gitid v4.2-93-gd691ed685)
```

This appears to be unnecessary as `cr_service()` is always called in `crtools.c` after `util_init()`. Here, `util_init()` is used to initialize global state (`criu_run_id`, `compel_run_id`) and not intended to be called multiple times in the same CRIU process.

Fixes: 622b433 ("criu: Initialize util before service worker starts")

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -1358,14 +1358,6 @@ int cr_service_work(int sk)
 	int ret = -1;
 	CriuReq *msg = 0;
 
-	/*
-	 * util_init initializes criu_run_id and compel_run_id so that sockets
-	 * are generated with an unique name identifying the specific process
-	 * even in cases where multiple processes with the same pid in
-	 * different pid namespaces are sharing the same network namespace.
-	 */
-	util_init();
-
 more:
 	opts.mode = CR_SWRK;
 


criu/cr-service.c
```diff
@@ -1358,14 +1358,6 @@ int cr_service_work(int sk)
 	int ret = -1;
 	CriuReq *msg = 0;
 
-	/*
-	 * util_init initializes criu_run_id and compel_run_id so that sockets
-	 * are generated with an unique name identifying the specific process
-	 * even in cases where multiple processes with the same pid in
-	 * different pid namespaces are sharing the same network namespace.
-	 */
-	util_init();
```
avagin, line 1367: If I remember right, one service daemon can be used to do dump operations more than once, but we need to regenerate uuid for each dump operation. Do I miss something?

rst0git: You are correct - I tested this locally with [libcriu-client](https://github.com/rst0git/criu/tree/libcriu-client/contrib/libcriu-client), where a single service daemon performs multiple dump operations at the same time. The uuid here is generated for each operation.

PR #2945, opened 2026-03-07
readme: update reference to consolidated workflows

rst0git:
Use the new consolidated GitHub Actions workflow (`ci.yml`) instead of individual workflow badges.


rst0git:
@avagin This change follows up on [https://github.com/checkpoint-restore/criu/pull/2919](https://github.com/checkpoint-restore/criu/pull/2919). After consolidating the CI workflows, we now have a single README badge for the unified workflow.

PR #2947, opened 2026-03-09
criu: introduce LUO image backend

Thesola10:
This patch adds Live Update Orchestrator as a storage backend for CRIU images.

## What this PR does

It adds two flags, `--luo-session` and `--luo-session-fd`, which specify a LUO session to save CRIU images into.

If CRIU owns the session file descriptor, that is the flag `--luo-session` is used, CRIU will remain open in order for the session file descriptor to be preserved when `kexec` is called.

## How it works

A new module, `luo.c`, has been introduced to manage LUO session file descriptors as a storage medium, by preserving `memfd`s as is recommended in the kernel documentation.

**Dump side**:
- If `--luo-session` is specified, a new LUO session is allocated with the name given as argument.
- For each image, a new `memfd` is opened and returned as the target file descriptor
- Once the image is closed, a token is generated from the image type and its numerical parameter (usually the PID or a small index) for subsequent retrieval.
- Again, if `--luo-session` is specified, CRIU hangs on a no-op in order to keep the session FD open until `kexec`

**Restore side**:
- If `--luo-session` is specified, CRIU will query `/dev/liveupdate` for a session with that name.
- When opening an image, the token is derived from the requested file name as above, and the corresponding file descriptor is returned.
- If `--luo-session` is specified, the LUO session is finalized and its resources are freed.

## Implementation notes

- The UAPI header for Live Update Orchestrator has been added into `criu/include/linux/liveupdate.h`
- `--luo-session-fd` is intended for an externally managed file descriptor (ideally `systemd`?) and CRIU will not attempt to close or preserve it.

## TODO:

- Short out the current file descriptor serializer in favor of natively preserving file descriptors as-is in the LUO session
- Save memory pages with zero-copy into the target `memfd`
- Introduce an end-to-end test for proper image persistence on LUO-enabled systems


avagin:
Is it better than using a dax file system? (https://www.kernel.org/doc/Documentation/filesystems/dax.txt)

Thesola10:
The idea of using LUO as an explicit backend is that we can then add LUO-specific optimisations, such as adding open FDs to the session instead of serialising them, or writing memory pages zero-copy into the memfd images.

avagin:
> For each image, a new memfd is opened and returned as the target file descriptor

I mean why we may want to do that? Where are you going to store memfd luo ids?

LongWei-27:
> The idea of using LUO as an explicit backend is that we can then add LUO-specific optimisations, such as adding open FDs to the session instead of serialising them, or writing memory pages zero-copy into the memfd images.

Currently, the LUO feature only supports backup and recovery for shmem memory. Will support for backup and recovery of anonymous memory be added in the future?

avagin:
The LUO integration will be implemented based on https://github.com/checkpoint-restore/criu/pull/3119.

PR #2952, opened 2026-03-11
plugin/amdgpu: Don't print error when restoring drm file

fdavid-amd:
When the amdgpu plugin is called for restore of a file img, we don't initially know if it's a kfd or drm file. So we first try to open a kfd file, and if that fails, we assume it's a drm file. That failure leads to a lot of error messages like

Error amdgpu_plugin: amdgpu-kfd-141.img: Failed to open for read

For this normal behaviour.

Add an option on open_img_file to suppress this error when we aren't sure if the file we're trying to open exists.

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -1141,6 +1141,9 @@ int amdgpu_restore_init(void)
 			while ((dir = readdir(d)) != NULL) {
 				if (strncmp("amdgpu-kfd-", dir->d_name, strlen("amdgpu-kfd-")) == 0) {
 					img_fp = open_img_file(dir->d_name, false, &img_size);
+					if (!img_fp)
+						return -EINVAL;
+
 					buf = xmalloc(img_size);
 					if (!buf) {
 						fclose(img_fp);
@@ -1163,6 +1166,9 @@ int amdgpu_restore_init(void)
 				}
 				if (strncmp("amdgpu-renderD-", dir->d_name, strlen("amdgpu-renderD-")) == 0) {
 					img_fp = open_img_file(dir->d_name, false, &img_size);
+					if (!img_fp)
+						return -EINVAL;
+
 					buf = xmalloc(img_size);
 					if (!buf) {
 						fclose(img_fp);
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -849,7 +849,7 @@ void *dump_bo_contents(void *_thread_data)
 	}
 
 	snprintf(img_path, sizeof(img_path), IMG_KFD_PAGES_FILE, thread_data->id, thread_data->gpu_id);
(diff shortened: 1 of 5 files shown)


plugins/amdgpu/amdgpu_plugin_drm.c
```diff
@@ -378,7 +378,7 @@ int amdgpu_plugin_drm_dump_file(int fd, int id, struct stat *drm)
 		drmPrimeHandleToFD(device_fd, boinfo->handle, 0, &dmabuf_fd);
 
 		snprintf(img_path, sizeof(img_path), IMG_DRM_PAGES_FILE, rd->id, rd->drm_render_minor, i);
-		bo_contents_fp = open_img_file(img_path, true, &image_size);
+		bo_contents_fp = open_img_file(img_path, true, &image_size, true);
```
rst0git, line 385: Do we need to check for error here and in the line above (`bo_contents_fp == NULL`)?

plugins/amdgpu/amdgpu_plugin_util.c
```diff
@@ -241,7 +242,8 @@ FILE *open_img_file(char *path, bool write, size_t *size)
 		fd = openat(criu_get_image_dir(), path, write ? (O_WRONLY | O_CREAT) : O_RDONLY, 0600);
 
 	if (fd < 0) {
-		pr_err("%s: Failed to open for %s\n", path, write ? "write" : "read");
+		if (expect_present)
+			pr_err("%s: Failed to open for %s\n", path, write ? "write" : "read");
```
rst0git, line 246: Would it make sense to remove the error message here and show it when appropriate?

plugins/amdgpu/amdgpu_plugin.c
```diff
@@ -1162,7 +1162,7 @@ int amdgpu_restore_init(void)
 					xfree(buf);
 				}
 				if (strncmp("amdgpu-renderD-", dir->d_name, strlen("amdgpu-renderD-")) == 0) {
-					img_fp = open_img_file(dir->d_name, false, &img_size);
+					img_fp = open_img_file(dir->d_name, false, &img_size, true);
```
rst0git, line 1168: Should we check for `img_fp == NULL`?

plugins/amdgpu/amdgpu_plugin_util.c
fdavid-amd: Given that it's almost always correct to show it, except for the one place where it isn't, I think it's worth the complication here.

plugins/amdgpu/amdgpu_plugin_drm.c
fdavid-amd: Yes we should. I'll add that in a separate patch before this one

plugins/amdgpu/amdgpu_plugin.c
fdavid-amd: Added in separate patch, thanks

PR #2956, opened 2026-03-12
ci: remove mips64el-stable-cross and mips64el-unstable-cross

avagin:
These jobs are failing for a long time. Debian is in the process of removing mips64el:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1105972.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -85,7 +85,6 @@ jobs:
           armv7-stable-cross,
           aarch64-stable-cross,
           ppc64-stable-cross,
-          mips64el-stable-cross,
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -85,7 +85,6 @@ jobs:
           armv7-stable-cross,
           aarch64-stable-cross,
           ppc64-stable-cross,
-          mips64el-stable-cross,
```
rst0git, line 88: Should we also remove `scripts/build/Dockerfile.mips64el-stable-cross.*`?

rst0git (approved):
LGTM

Issue #2285, opened 2023-10-13
C/R Support for POSIX message queues


tobse80:
**Description**

It seems that checkpoint/restore for [POSIX message queues](https://man7.org/linux/man-pages/man7/mq_overview.7.html) is (still) not supported. Checkpointing applications using such queues fail:

```log
(00.149697) Error (criu/files-reg.c:1710): Can't lookup mount=413 for fd=3 path=/mq0
(00.149706) Error (criu/cr-dump.c:1635): Dump files (pid: 13273) failed with -1(00.186459) Error (criu/cr-dump.c:2053): Dumping FAILED.
(00.186459) Error (criu/cr-dump.c:2053): Dumping FAILED.
```

**Steps to reproduce the issue:**

1. Create test application `mq0.py`
   
   ```python
   import posix_ipc as ipc
   from time import sleep
   
   mq = ipc.MessageQueue('/mq0', ipc.O_CREAT)
   
   while True:
      sleep(0.5)
   ```
   
   *[Note that Python was chosen for writing a compact test program. One could use C instead.]*

2. Prepare podman (or docker) container file with Python3 support and `posix_ipc` module:
   
   ```dockerfile
   FROM python:3-slim-bookworm
   RUN apt-get update && apt-get install -y --no-install-recommends \
         python3 python3-pip python3-dev gcc \
      && rm -rf /var/lib/apt/lists/*
   
   RUN pip install --break-system-packages --no-cache-dir \
         posix-ipc
   
   COPY mq0.py /
   ```

3. Build test container
   
   ```shell
   sudo podman build -t test container
   ```

4. Run container
   
   ```shell
   sudo podman run -dt test python3 /mqtest.py
   ```

5. Checkpoint container
   
   ```shell
   sudo podman container checkpoint -l
   ```

**Describe the results you received:**

Checkpointing the container fails:

```shell
2023-10-13T14:04:48.324540Z: CRIU checkpointing failed -52.  Please check CRIU logfile /var/lib/containers/storage/overlay-containers/8b54dfcfd617302f2f48a8a16dbd75e4b9a539591ab3dd5c2d7625b0eeaf802f/userdata/dump.log
[...]

PR #2957, opened 2026-03-12
criu: cr-check: add POSIX mqueue kernel support check

Abdullah-Badawy1:
Dumping a process with open mqueue descriptors fails  and no useful guidance for the user.

Add check_posix_mqueue() to cr-check.c that probes
/proc/sys/fs/mqueue, which only exists when
CONFIG_POSIX_MQUEUE=y. Placed in the Category 2 block
alongside check_posix_timers(). Exposed via feature_list[]
so it can be queried directly:

  sudo criu check --feature posix_mqueue

This is the first step toward full POSIX mqueue support.

Test : 
image

Link: https://github.com/checkpoint-restore/criu/issues/2285
Signed-off-by: Abdullah Albadawy abdullahalbadawy1@gmail.com

### DIFF
--- a/criu/cr-check.c
+++ b/criu/cr-check.c
@@ -839,6 +839,17 @@ static int check_posix_timers(void)
 	return -1;
 }
 
+static int check_posix_mqueue(void)
+{
+	if (access("/proc/sys/fs/mqueue", F_OK)) {
+		pr_err("POSIX mqueue sysctl interface is missing.\n"
+		       "Kernel must be built with CONFIG_POSIX_MQUEUE.\n");
+		return -1;
+	}
+
+	return 0;
+}
+
 static unsigned long get_ring_len(unsigned long addr)
 {
 	FILE *maps;
@@ -1688,6 +1699,7 @@ int cr_check(void)
 		ret |= check_ptrace_dump_seccomp_filters();
 		ret |= check_mem_dirty_track();
 		ret |= check_posix_timers();
+		ret |= check_posix_mqueue();
 		ret |= check_tun_cr(0);
 		ret |= check_timerfd();
 		ret |= check_mnt_id();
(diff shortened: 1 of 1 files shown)


Abdullah-Badawy1:
Hi, @rst0git  @avagin , @Snorch, @adrianreber  I've been reading and understading things i could do about the POSIX message queues checkpointing project and before I write a proposal or touch any code, I wanted to reach out because I have some honest questions about how to approach this the right way.
My understanding of what GSoC actually means at least how I've come to see it  is that the goal isn't producing code. It's building something that you understand deeply enough that you can defend every line, discuss every decision, and keep contributing long after the program ends. Code that nobody fully understands is a liability, not a contribution. That's the framing I'm bringing to this.
On PRs before applying: I could open pull requests right now. But if I submit code I don't fully own mentally, I'm just creating review work for you without giving anything real back. So my question is: is a PR only worth making if I can stand behind every line and explain every decision? Or is there a smaller, more honest contribution  a question on the issue, a test case, something exploratory  that would be more useful at this stage?
On the proposal: If I write a proposal that shows I genuinely understand the problem  the kernel-side gap, why there's no non-destructive read interface today, what CRIU needs from a new syscall or interface, and what a realistic implementation path looks like  is that depth of insight enough to open a real conversation with you? I'm not asking for a guaranteed acceptance. I'm asking whether a proposal with honest depth is the right place to put my energy right now.
On why I'm here: GSoC is an incentive and a door, but it's not the actual goal. What I actually want is to be around people who understand kernel internals and systems programming at a deep level  and to earn the right to contribute and discuss alongside them seriously over time. The discussion itself, with people who really know this, is the valuable thing. GSoC just makes it more likely to happen.
I'd really appreciate any direction you can give me.
Signed-off-by: Abdullah Albadawy <abdullahalbadawy1@gmail.com>

PR #2959, opened 2026-03-13
cr-service: fix address handling of requests

rst0git:
The page server address from RPC requests was ignored when lazy_pages was enabled. The code that copies the address into `opts.addr` only ran inside the `!opts.lazy_pages` branch, so with lazy pages the page server would bind to `0.0.0.0` instead of the specified address.
    
This breaks live migration when the container runtime needs the page server to listen on a specific address. The [zeropod](https://github.com/ctrox/zeropod/blob/0ee8461eccd1ec3d2671c93e701e975f3a2669ee/criu/rpc-ps-address.patch) project worked around this with a downstream patch that adds a raw pointer assignment (`opts.addr = req->ps->address`), but that causes a use-after-free when the protobuf message is freed.

Fix this by moving `SET_CHAR_OPTS` (which copies via `strdup`) and the `NULL` fallback outside the `!opts.lazy_pages` block so they apply unconditionally.

Reproducer (run as root, requires pycriu):
    
```bash
sleep 99999 &
criu dump -t $! -D /tmp/imgs --shell-job -v0
criu service --address /tmp/criu.sock -v0 &

python3 -c "
import os, socket
from pycriu import rpc_pb2 as rpc
s = socket.socket(socket.AF_UNIX, socket.SOCK_SEQPACKET)
s.connect('/tmp/criu.sock')
req = rpc.criu_req()
req.type = rpc.PAGE_SERVER
req.opts.images_dir_fd = os.open('/tmp/imgs', os.O_DIRECTORY)
req.opts.lazy_pages = True
req.opts.ps.port = 27
req.opts.ps.address = '127.0.0.1'
s.send(req.SerializeToString())
"

ss -tlnp sport = :27  # Bug: shows 0.0.0.0:27 instead of 127.0.0.1:27
```

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -554,12 +554,15 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 	if (req->ps) {
 		opts.port = (short)req->ps->port;
 
+		if (req->ps->address) {
+			SET_CHAR_OPTS(addr, req->ps->address);
+		} else {
+			xfree(opts.addr);
+			opts.addr = NULL;
+		}
+
 		if (!opts.lazy_pages) {
 			opts.use_page_server = true;
-			if (req->ps->address)
-				SET_CHAR_OPTS(addr, req->ps->address);
-			else
-				opts.addr = NULL;
 
 			if (req->ps->has_fd) {
 				if (!opts.swrk_restore)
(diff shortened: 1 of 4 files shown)


rst0git:
cc: @ctrox

ctrox:
Oof, thanks for flagging and fixing!

avagin:
Acked-by: Andrei Vagin

PR #2961, opened 2026-03-14
ci: port Vagrant Fedora Rawhide test to GitHub Actions

adrianreber:
Replace the Cirrus CI Vagrant-based Fedora Rawhide test with a Lima VM-based equivalent in GitHub Actions. This follows the same pattern used by the runc project (lima-vm/lima-actions).

The new vagrant-fedora-rawhide-test job:
- Starts a Lima Fedora VM with KVM acceleration
- Installs CRIU build dependencies and the latest vanilla kernel
- Reboots the VM to activate the new kernel
- Runs the fedora-rawhide CI target inside a podman container

The existing fedora-rawhide-test job is kept unchanged as it tests against rawhide userspace on the host kernel, while the new job tests against both rawhide userspace and a vanilla Fedora kernel.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -49,26 +49,6 @@ task:
     make -C scripts/ci local SKIP_CI_PREP=1 CC=gcc CD_TO_TOP=1 ZDTM_OPTS="-x zdtm/static/socket-raw"
 
 task:
-  name: Vagrant Fedora Rawhide based test
-  environment:
-    HOME: "/root"
-    CIRRUS_WORKING_DIR: "/tmp/criu"
-
-  compute_engine_instance:
-    image_project: cirrus-images
-    image: family/docker-kvm
-    platform: linux
-    cpu: 4
-    memory: 16G
-    nested_virtualization: true
-
-  setup_script: |
-    contrib/apt-install make gcc pkg-config git perl-modules iproute2 kmod wget cpu-checker
-    sudo kvm-ok
-  build_script: |
-    make -C scripts/ci vagrant-fedora-rawhide
-
-task:
   name: Vagrant Fedora based test (non-root)
   environment:
     HOME: "/root"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,41 @@ jobs:
       # FIXME: https://github.com/containers/podman/issues/14920
       run: sudo -E XDG_RUNTIME_DIR= make -C scripts/ci fedora-rawhide CONTAINER_RUNTIME=podman BUILD_OPTIONS="--security-opt seccomp=unconfined"
 
+  vagrant-fedora-rawhide-test:
+    name: Vagrant Fedora Rawhide based test
+    needs: [alpine-test]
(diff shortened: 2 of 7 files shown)


scripts/ci/lima-fedora-rawhide.sh
```diff
@@ -0,0 +1,46 @@
...
+test() {
+	#
+	# Workaround the problem:
+	# error running container: error from /usr/bin/crun creating container
+	# for [...]: sd-bus call: Transport endpoint is not connected
+	# Let's just use runc instead of crun
+	# see also https://github.com/kata-containers/tests/issues/4283
+	#
+	dnf remove -y crun || true
+	dnf install -y podman runc
```
rst0git: I believe `crun` will be installed on the second dnf line as a dependency for Podman:
```
$ sudo dnf repoquery --installed --whatrequires crun
containers-common-extra-5:0.67.0-1.fc43.noarch
$ sudo dnf repoquery --installed --whatrequires containers-common-extra
buildah-2:1.43.0-1.fc43.x86_64
podman-5:5.8.0-1.fc43.x86_64
```
and we also need to change the default runtime in the Podman config:
https://github.com/checkpoint-restore/criu/blob/f81e46b7311110f88328e7a0f61721e4a8eda06d/scripts/ci/podman-test.sh#L18

adrianreber: It works with the defaults. I removed the whole runc part.

avagin:
All patches look good to me except 'zdtm: fix socket-tcpbuf hang after restore.' The logic in that last one doesn't sound right; it's like saying let's ignore if we don't hear a heartbeat. Sorry if I misunderstood anything.

adrianreber:
> All patches look good to me except 'zdtm: fix socket-tcpbuf hang after restore.' The logic in that last one doesn't sound right; it's like saying let's ignore if we don't hear a heartbeat. Sorry if I misunderstood anything.

The thread-bomb test and the socket-tcpbuf where flaky in CI and I tried to fix them. I removed the socket-tcpbuf patch for now. Something is not right with that test as it is not always working and I am not sure what the problem is exactly. The change was something which made the problem go away, but if that broke the test it is not really useful.

adrianreber:
Now it fails again:
```
################### 1 TEST(S) FAILED (TOTAL 486/SKIPPED 46) ####################
 * zdtm/static/socket-tcpbuf-local(ns)
```

avagin:
> Now it fails again:
> 
> ```
> ################### 1 TEST(S) FAILED (TOTAL 486/SKIPPED 46) ####################
>  * zdtm/static/socket-tcpbuf-local(ns)
> ##################################### FAIL #####################################
> ```

@adrianreber You can file an issue and exclude it for now. Someone needs to investigate what is really going on with it.

avagin (approved):

adrianreber:
Created a tracker bug: https://github.com/checkpoint-restore/criu/issues/2987 and mentioned it in the test script where the test is marked to be skipped.

rst0git (approved):
LGTM

PR #2962, opened 2026-03-14
 file-lock: fix type confusion in broken lease restore

avagin:
When restoring a breaking lease, CRIU needs to first restore the pre-breaking state of the lease and then trigger the break again.
    
Previously, restore_lease_prebreaking_state() was incorrectly using the  file descriptor type (e.g., FD_TYPES__REG) instead of the lease type to  determine which lease to set. This caused it to use O_RDONLY/O_WRONLY logic on a value that wasn't an open flag.


rst0git (approved):
LGTM

PR #2964, opened 2026-03-15
A few fixes here and there

avagin:


criu/cr-restore.c
```diff
@@ -3516,6 +3520,10 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 
 	/* No longer need it */
 	core_entry__free_unpacked(core, NULL);
+	for (i = 0; i < current->nr_threads; i++) {
+		if (current->core[i] && current->core[i] != core)
```
rst0git, line 3526: It looks like here `core_entry__free_unpacked(core, NULL)` frees `core` then we are comparing against a freed pointer. 

We need to move `core_entry__free_unpacked(core, NULL)` after the loop.

avagin: bzzzzz. core isn't dereferenced here, why is it the problem?

rst0git (approved):
LGTM

PR #2966, opened 2026-03-15
ci: shard alpine-test into parallel jobs to reduce CI time

adrianreber:
The alpine-test CI job runs all ~483 zdtm tests sequentially three times (normal, mntns-compat-mode, criu-config), followed by many non-shardable tests. This dominates overall CI wait time. With only 2 jobs running in parallel (GCC and CLANG) the alpine tests take around 30 minutes.

Use the existing --test-shard-index and --test-shard-count flags already built into test/zdtm.py to split the zdtm test suite across four parallel runners (shards 0-3). A fifth shard runs all non-shardable tests (lazy pages, fault injection, test/others/*, rootless, compel, plugins, etc.) independently and in parallel with the zdtm shards. This increases parallelism from 2 to 10 jobs and reduces the alpine test wall-clock time from ~30 to ~10 minutes.

Changes:
- run-ci-tests.sh: Build SHARD_OPTS from ZDTM_SHARD_INDEX/COUNT env vars and pass them to zdtm.py. Extract all non-shardable tests into a run_non_shardable_tests() function. Dispatch based on shard index: 0-3 run zdtm slices, 4 runs non-shardable tests, unset runs everything sequentially (preserving existing behavior).
- Makefile: Pass ZDTM_SHARD_INDEX and ZDTM_SHARD_COUNT into the container when set.
- ci.yml: Add shard: [0, 1, 2, 3, 4] to the alpine-test matrix, producing 10 jobs (2 compilers x 5 shards). Job labels now show descriptive shard names (e.g. "zdtm 1/4", "non-zdtm") instead of raw indices.

When sharding is not configured the script behaves identically to before, so other CI jobs (aarch64, compat, gcov, etc.) are unaffected.

### DIFF
--- a/scripts/ci/Makefile
+++ b/scripts/ci/Makefile
@@ -46,10 +46,19 @@ ifeq ($(CONTAINER_RUNTIME),podman)
 endif
 
 export ZDTM_OPTS
+export ZDTM_SHARD_INDEX
+export ZDTM_SHARD_COUNT
 
 $(TARGETS):
 	$(MAKE) -C ../build $@$(target-suffix)
-	$(CONTAINER_RUNTIME) run --env-file docker.env -v `pwd`/../../:/criu $(if $(ZDTM_OPTS),-e ZDTM_OPTS) $(CONTAINER_OPTS) criu-$@ scripts/ci/run-ci-tests.sh
+	$(CONTAINER_RUNTIME) run \
+		--env-file docker.env \
+		-v `pwd`/../../:/criu \
+		$(if $(ZDTM_OPTS),-e ZDTM_OPTS) \
+		$(if $(ZDTM_SHARD_INDEX),-e ZDTM_SHARD_INDEX) \
+		$(if $(ZDTM_SHARD_COUNT),-e ZDTM_SHARD_COUNT) \
+		$(CONTAINER_OPTS) \
+		criu-$@ scripts/ci/run-ci-tests.sh
 
 fedora-asan:
 	$(MAKE) -C ../build $@$(target-suffix)
--- a/scripts/ci/run-ci-tests.sh
+++ b/scripts/ci/run-ci-tests.sh
@@ -8,6 +8,16 @@ X86_64_PKGS=(gcc-multilib)
 # Convert from string to array.
 IFS=" " read -r -a ZDTM_OPTS <<< "$ZDTM_OPTS"
 
+SHARD_OPTS=()
+if [ -n "$ZDTM_SHARD_COUNT" ] && [ "$ZDTM_SHARD_COUNT" -gt 0 ]; then
+	if [ -z "$ZDTM_SHARD_INDEX" ]; then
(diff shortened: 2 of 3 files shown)


scripts/ci/run-ci-tests.sh
```diff
@@ -8,6 +8,12 @@ X86_64_PKGS=(gcc-multilib)
 # Convert from string to array.
 IFS=" " read -r -a ZDTM_OPTS <<< "$ZDTM_OPTS"
 
+SHARD_OPTS=()
+if [ -n "$ZDTM_SHARD_COUNT" ] && [ "$ZDTM_SHARD_COUNT" -gt 0 ]; then
+	SHARD_OPTS=(--test-shard-index "$ZDTM_SHARD_INDEX" \
```
rst0git, line 17: It might be good to check if `ZDTM_SHARD_INDEX` is set as well:
```bash
if [ -z "$ZDTM_SHARD_INDEX" ]; then                                                                                                                                  
        echo "ERROR: ZDTM_SHARD_COUNT set but ZDTM_SHARD_INDEX is not"                                                                                         
        exit 1                                                                                                                                                       
fi
```

scripts/ci/Makefile
```diff
@@ -45,10 +45,12 @@ ifeq ($(CONTAINER_RUNTIME),podman)
 endif
 
 export ZDTM_OPTS
+export ZDTM_SHARD_INDEX
+export ZDTM_SHARD_COUNT
 
 $(TARGETS):
 	$(MAKE) -C ../build $@$(target-suffix)
-	$(CONTAINER_RUNTIME) run --env-file docker.env -v `pwd`/../../:/criu $(if $(ZDTM_OPTS),-e ZDTM_OPTS) $(CONTAINER_OPTS) criu-$@ scripts/ci/run-ci-tests.sh
+	$(CONTAINER_RUNTIME) run --env-file docker.env -v `pwd`/../../:/criu $(if $(ZDTM_OPTS),-e ZDTM_OPTS) $(if $(ZDTM_SHARD_INDEX),-e ZDTM_SHARD_INDEX) $(if $(ZDTM_SHARD_COUNT),-e ZDTM_SHARD_COUNT) $(CONTAINER_OPTS) criu-$@ scripts/ci/run-ci-tests.sh
```
rst0git: This line is very long. Would it be possible to split this into multiple lines?

scripts/ci/run-ci-tests.sh
adrianreber: done

scripts/ci/Makefile
adrianreber: done

avagin:
LGTM, thanks.

An alternative solution would be to splice run-ci-tests.sh. We don't need to run all these tests in one job. It can be separate jobs to run zdtm tests, pre-dump tests, fault tests, etc.

PR #2967, opened 2026-03-15
ci: port CentOS Stream test to GitHub Actions using Lima

adrianreber:
Move the CentOS Stream 9 based test from Cirrus CI to GitHub Actions using Lima VMs. Expand coverage to a matrix of CentOS Stream 9 and 10 on both x86_64 and aarch64.

The new job follows the same Lima-based pattern introduced for the Fedora Rawhide test but runs the local CI target directly in the VM instead of inside a container.

Depends on #2961

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -19,36 +19,6 @@ task:
     make -C scripts/ci vagrant-fedora-no-vdso
 
 task:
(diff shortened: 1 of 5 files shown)


adrianreber:
See #3033

Issue #2426, opened 2024-06-28
Using CRIU with nesting LXC containers


alexfrolov:
Hi!

I am considering using CRIU for checkpoint/restore of nested LXC containers. Do I understand it right that in this case CRIU should be called inside parent container? 

For example, as far as I understood CRIU is using mnt namespace of the CRIU process (`/proc/self/ns/mnt`) to resolve external mountpoints for target process, so this means that CRIU wont be able to work from the host level in case target process is running in nested container. Is that correct?

Thanks,
   Alex

avagin:
For LXC, the auto mode for external mounts will not work, you need to enumerate them manually. But you still need to call CRIU from a parent container to dump a target pid namespace properly. Otherwise, it will look like two nested pid namespaces.

I don't recommend to use CRIU directly to dump/restore LXC containers. It should be easier to use lxc tools for that.

alexfrolov:
Just to clarify about auto-detection of external mounts.

When I am running non-nested LXC container, it uses external mp at least for rootfs (besides, ` /proc` stuff which extensivly uses `fuse.lxcfs`). For example, for Xenial-based container `/proc/1/mountinfo` looks like this:

```
root@u3:/home/ubuntu# cat /proc/1/mountinfo | grep master
690 619 253:1 /u3/rootfs / rw,relatime master:322 - ext4 /dev/mapper/dummy--vg-dummy--lv rw
696 697 0:34 / /sys/fs/fuse/connections rw,nosuid,nodev,noexec,relatime master:18 - fusectl fusectl rw
700 692 0:42 /proc/cpuinfo /proc/cpuinfo rw,nosuid,nodev,relatime master:154 - fuse.lxcfs lxcfs rw,user_id=0,group_id=0,allow_other
701 692 0:42 /proc/diskstats /proc/diskstats rw,nosuid,nodev,relatime master:154 - fuse.lxcfs lxcfs rw,user_id=0,group_id=0,allow_other
702 692 0:42 /proc/loadavg /proc/loadavg rw,nosuid,nodev,relatime master:154 - fuse.lxcfs lxcfs rw,user_id=0,group_id=0,allow_other
703 692 0:42 /proc/meminfo /proc/meminfo rw,nosuid,nodev,relatime master:154 - fuse.lxcfs lxcfs rw,user_id=0,group_id=0,allow_other
[...]

Issue #2086, opened 2023-02-20
Can criu be used to c/r the ’containerd-shim’ process?


Carol-code2222:
I'm making an attempt:

Start a Docker container:
[root@localhost ~]# docker run -d --name looper test /bin/sh -c 'i=0; while true; do echo (expr $i + 1); sleep 1; done'

I'm trying to use criu to dump the process 'containerd-shim'

[root@localhost tmp]# ps -ef | grep while
root      399002  **398984**  0 Feb17 ?        00:00:00 /bin/sh -c i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done
root     1452408    2334  0 17:04 pts/0    00:00:00 grep --color=auto while

[root@localhost tmp]# ps -ef | grep 398984
root      **398984**    2667  0 Feb17 ?        00:00:07 containerd-shim -namespace moby -workdir /var/lib/containerd/io.containerd.runtime.v1.linux/moby/2e1deb44e00349759690efa26cf3df0a7f5db243cc1713abc57e203999c91686 -address /run/containerd/containerd.sock -containerd-binary /usr/bin/containerd -runtime-root /var/run/docker/runtime-runc
root      399002  **398984**  0 Feb17 ?        00:00:00 /bin/sh -c i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done

[root@localhost tmp]# criu dump -D imgdir/ -j -t **398984** -v4 -o dump.log

An error occurred: Error (criu/namespaces.c:463): Can't dump nested pid namespace for **398984**

Is this error normal? Is there a solution?

adrianreber:
> Is this error normal?  Is there a solution?

No, nested namespaces is something CRIU currently cannot handle.

> I'm trying to use criu to dump the process 'containerd-shim'

Why are you trying to do that?

Carol-code2222:
> > Is this error normal?  Is there a solution?
> 
> No, nested namespaces is something CRIU currently cannot handle.
> 
> > I'm trying to use criu to dump the process 'containerd-shim'
> 
> Why are you trying to do that?

I want to dump all processes in the container by dumping the parent process.

Currently, there are many types of containers, such as lxc, podman, docker, and apptainer. Each type of container uses different commands to invoke the criu. The criu needs to adapt to each type of container, which is troublesome.

If all containers comply with the OCI runtime (runc) standard, can the criu dump and restore the processes in the container in a unified manner from outside container?

[...]

Issue #1462, opened 2021-04-30
per pidns pid_max


brauner:
Hey folks,

I picked up pid_max namespacing again because we have use-cases for it. Is this something you're interested in too and have use-cases for?
https://gitlab.com/brauner/linux/-/tree/pid_max_namespacing

The test-cases give a pretty good indication how this is supposed to work. Assume nested pid namespaces:

```
pidns_1: echo 400 > /proc/sys/kernel/pid_max
\
 pidns_2: echo 500 > /proc/sys/kernel/pid_max
 ```
 
 If there are 399 processes in pidns_1 then pidns_2 can't create anymore processes.
 In pidns_2 pids wrap at 500 whereas they wrap at 400 in pidns_1.
 
 So pid_max still acts as a global limit but gives leeway to nested pid namespaces in when their pid numbers wrap.

mihalicyn:
Hi, Christian,

looks interesting!
The changes that you've suggested will be very useful also when we port nested PID namespaces support from Virtuozzo to upstream CRIU. (In the VZ kernel, we even have pid_max virtualization at now.)

[off]
We getting stuck with that porting because we need to have some support from the upstream kernel to handle complex mounts configurations (https://lore.kernel.org/patchwork/patch/1402167/). Because without extending mount namespaces it makes no sense to have nested PID namespaces configurations C/R.
[/off]

Snorch:
@brauner Hi!

> Is this something you're interested in too and have use-cases for?

Yes. Usecase for criu is migration between nodes where source node has higher pid_max then source. If pid_max is not virtualized and we want to restore process with pid > destination node pid_max criu would fail without changing global value. If pid_max is virtualized we are free to restore any pids we want.

PR #2968, opened 2026-03-16
pidns: support single-level nested PID namespace dump/restore

nidhishgajjar:
## Summary

This PR adds support for dumping and restoring process trees that contain a **single-level child PID namespace**. This was identified as an [unresolved problem at Linux Plumbers Conference 2024](https://lpc.events/event/18/contributions/1811/) and has been requested in multiple issues (#2426, #232).

The change is minimal (2 code files + 1 test) and enables sandbox/container use cases where processes run in isolated PID namespaces — the most common deployment pattern for Kubernetes pods, Podman containers, and agent runtimes.

## Changes

**`criu/include/namespaces.h`** — Add `CLONE_NEWPID` to `CLONE_SUBNS`:
```c
-#define CLONE_SUBNS (CLONE_NEWNS | CLONE_NEWNET)
+#define CLONE_SUBNS (CLONE_NEWNS | CLONE_NEWNET | CLONE_NEWPID)
```

**`criu/pstree.c`** — Fix restore validation for nested namespace types:
```c
-} else if (cflags & ~(root_ns_mask & CLONE_SUBNS)) {
+} else if (cflags & ~(root_ns_mask | CLONE_SUBNS)) {
```

The old check only allowed sub-task namespaces that were both in `CLONE_SUBNS` AND in `root_ns_mask`. This meant a child could only use a PID namespace if the root task also used one. The fix allows any `CLONE_SUBNS` type to be created by sub-tasks, even if root doesn't use it — matching how mount and network namespaces already work, and enabling the sandboxing pattern where root spawns children in isolated PID namespaces.

**`test/zdtm/static/pidns_nested.c`** — New ZDTM test:
- Parent calls `unshare(CLONE_NEWPID)` + `fork()`
- Child is PID 1 in new namespace, calls `setsid()` (session leader must be visible in child pidns for dump)
- Dump/restore with pre-dump
- Verifies `getpid() == 1` preserved after restore

## Who benefits

- **Kubernetes** — KEP-2008 (checkpoint/restore of pods) hits this limitation since pods run in PID namespaces. Currently requires workarounds to checkpoint from the host pidns.
- **Podman / CRI-O** — Container checkpoint/restore. Same workaround pattern.
- **LXC / LXD** — Issue #2426 is from LXC users who can't checkpoint nested containers.
- **Agent runtimes / sandboxes** — Any system that spawns processes in isolated PID namespaces for security (our use case — bare-metal agent runtime with CRIU-based sleep/wake).

## Scope and limitations

This PR intentionally covers the **single-level case** (one child PID namespace) which handles ~90% of real-world use cases. It does NOT attempt arbitrary N-level nesting, which would require:
- Per-namespace PID trees (current tree uses `ns[0].virt` as the unique key)
- Multi-level PID storage in the protobuf image format
- `clone3()` with `set_tid_size > 1` for multi-level PID restore

The single-level implementation is additive — it doesn't complicate future N-level support.

**Known limitation:** When CRIU itself runs inside a PID namespace where the root process is also PID 1, and a child pidns also has PID 1, there's a collision in the pid red-black tree. This only affects nested-container scenarios (CRIU-in-a-container), not the common case of CRIU dumping from the host.

## Testing

- ZDTM `pidns_nested`: PASS (dump + restore + pre-dump)
- Existing tests (pid00, pidfd_child, pidfd_self, timens_nested, maps00, mprotect00, pipe00, sigpending): all PASS, no regressions
- Real-world validation: tested with multi-process agent workloads (Node.js process trees) in sandboxed PID namespaces with checkpoint/restore cycles

Signed-off-by: nidhishgajjar <hacker@scigic.com>


adrianreber:
> Kubernetes — KEP-2008 (checkpoint/restore of pods) hits this limitation since pods run in PID namespaces. Currently requires workarounds to checkpoint from the host pidns.

Can you elaborate a bit more where the problems are here.

nidhishgajjar:
Good question, let me elaborate.

Kubernetes currently works around this. runc passes `--external pid[<inode>]:extRootPidNS` on dump and `--inherit-fd` on restore, so CRIU never actually dumps or restores the PID namespace, it just records a reference. The container runtime is responsible for recreating the namespace on restore. This works, but the runtime has to manage PID namespace recreation, PID preservation requires `ns_last_pid` manipulation, and you can't dump from outside the container's namespace hierarchy. Native support in CRIU would let container runtimes simplify that path, the same way mount and network namespaces are already handled natively via `CLONE_SUBNS`.

Our use case is simpler. We run AI agent workloads on bare metal, each in its own PID namespace for process isolation. CRIU dumps from the host looking into the agent's namespace. The `--external pid[]` workaround doesn't apply here since there's no container runtime managing namespace recreation.

For context, this has come up before:
- In #2086 you noted that nested namespaces is something CRIU currently cannot handle
- In #2426 avagin confirmed the limitation for LXC users
- In #1462 mihalicyn mentioned Virtuozzo has an internal implementation but upstream porting was blocked on kernel mount namespace changes

This PR takes the minimal approach for the single-level case. It adds `CLONE_NEWPID` to `CLONE_SUBNS` and fixes the restore validation. It doesn't attempt the full N-level nesting which would need per-namespace PID trees and image format changes.

Happy to update the PR description to be more precise about the Kubernetes angle.

test/zdtm/static/pidns_nested.c
```diff
@@ -0,0 +1,161 @@
...
+	}
+	close(pipe_result[0]);
+
+	ret = waitpid(child, &status, 0);
+	if (ret < 0 && errno == ECHILD) {
+		/*
+		 * After restore, the host PID may have changed.
+		 * Wait for any child instead.
+		 */
+		ret = waitpid(-1, &status, 0);
```
avagin, line 139: This is why the feature shouldn't be labeled as 'support for single-level nested PID namespace dump/restore.' It feels more like a workaround for a specific use case. At a minimum, you should introduce an option to enable this feature; it should not be active by default. We probably need to do something to be sure that we use/mount proper /proc mounts...

rst0git:
> Our use case is simpler. We run AI agent workloads on bare metal, each in its own PID namespace for process isolation. CRIU dumps from the host looking into the agent's namespace. The --external pid[] workaround doesn't apply here since there's no container runtime managing namespace recreation.

@nidhishgajjar Have you tried using [`criu-ns`](https://github.com/checkpoint-restore/criu/blob/criu-dev/scripts/criu-ns)? It handles the PID namespace recreation during restore.

PR #2970, opened 2026-03-17
build: gate amdgpu plugin on amdgpu dev support

### LINKED ISSUE #1900: Build failure if libdrm installed without amdgpu support
**Description**

criu-3.17 fails to build (with the needed backports: cd0ed7e5491c7c827b887b2d80cf35e87819278b and 2b3763fe6a39033d4e3d3d80e28911d1737ac23b) if libdrm is installed without amdgpu support.

Originally reported downstream in [Gentoo](https://bugs.gentoo.org/846446).

**Steps to reproduce the issue:**
1. Install libdrm without amdgpu support.
2. Try to build criu-3.17 (with above patches from git applied)

**Describe the results you received:**

Build failure:
```
amdgpu_plugin.c:20:10: fatal error: libdrm/amdgpu.h: No such file or directory
   20 | #include <libdrm/amdgpu.h>
      |          ^~~~~~~~~~~~~~~~~
```

**Describe the results you expected:**

Successful build.

**Additional information you deem important (e.g. issue happens only occasionally):**

The check at https://github.com/checkpoint-restore/criu/blob/criu-dev/Makefile.config#L24 has two issues:
1. it only checks for the presence of libdrm, not libdrm's amdgpu support;
2. it's [automagic](https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies) (i.e. building criu will use libraries available on the system without giving a way to disable them, which can be an issue if e.g. building binaries for another system).

**Additional environment details:**

Gentoo.

cheese-cakee:
### Problem
CONFIG_AMDGPU was enabled whenever libdrm was detected, even on systems where libdrm is present without amdgpu development support (libdrm/amdgpu.h / -ldrm_amdgpu).

This causes build failures in plugins/amdgpu/amdgpu_plugin.c.

### Change
- Add FEATURE_TEST_LIBDRM_AMDGPU in scripts/feature-tests.mak to verify both headers and linkability for amdgpu (-ldrm -ldrm_amdgpu).
- Gate CONFIG_AMDGPU in Makefile.config on that feature test instead of generic pkg-config-check,libdrm.
- Keep behavior minimal: amdgpu plugin is built only when amdgpu dev support is available.

### Notes
This is a surgical build-detection fix with no runtime logic changes.

Fixes: #1900


rst0git:
@cheese-cakee To build the AMD GPU plugin, we currently require `libdrm-devel` / `libdrm-dev` as a build dependency, and without this package the plugin will be skipped.

> CONFIG_AMDGPU was enabled whenever libdrm was detected, even on systems where libdrm is present without amdgpu development support (libdrm/amdgpu.h / -ldrm_amdgpu). This causes build failures in plugins/amdgpu/amdgpu_plugin.c.

Would you be able to provide more information on how to reproduce this error?

cheese-cakee:
Thanks for the question @rst0git . Repro is the one documented in #1900 : `libdrm` installed but AMDGPU development bits absent (`libdrm/amdgpu.h` and `-ldrm_amdgpu`), while generic `libdrm` detection still enabled `CONFIG_AMDGPU`; build then fails in `plugins/amdgpu/amdgpu_plugin.c`. This patch makes detection explicit (header + link), so the plugin is built only when AMDGPU dev support is present. If this case is no longer possible in current CI/package constraints, I'm happy to close as obsolete.

rst0git:
> Repro is the one documented in https://github.com/checkpoint-restore/criu/issues/1900

@cheese-cakee This issue was opened in 2022, and the code has changed significantly since then. Were you able to reproduce the error yourself?

cheese-cakee:
I reproduced this on current code in a clean Ubuntu 24.04 container.

What I tested:
- Installed `build-essential pkg-config libdrm-dev`
- Simulated the problematic package state (`libdrm` present, AMDGPU dev bits absent) by removing:
  - `/usr/include/libdrm/amdgpu.h`
  - `/usr/lib/x86_64-linux-gnu/libdrm_amdgpu.so`

Observed results:
- On current `criu-dev` logic:
  - `make -pn | grep CONFIG_AMDGPU` -> `CONFIG_AMDGPU = y`
  - plugin compile fails:
    - `plugins/amdgpu/amdgpu_plugin.c:22:10: fatal error: libdrm/amdgpu.h: No such file or directory`
- On this PR branch:
  - `make -pn | grep CONFIG_AMDGPU` -> no `CONFIG_AMDGPU` (disabled as expected)

So with libdrm detected but missing AMDGPU dev support, current logic still enables AMDGPU, while this patch gates it correctly.

cheese-cakee:
for clarity:

```bash
docker run --rm -it \
  -v "$PWD:/src:ro" ubuntu:24.04 bash -lc '
set -eux
apt-get update -qq
DEBIAN_FRONTEND=noninteractive apt-get install -y -qq build-essential pkg-config libdrm-dev
cp -a /src /work

# Simulate: libdrm present, AMDGPU dev bits missing
mv /usr/include/libdrm/amdgpu.h /usr/include/libdrm/amdgpu.h.bak || true
mv /usr/lib/x86_64-linux-gnu/libdrm_amdgpu.so /usr/lib/x86_64-linux-gnu/libdrm_amdgpu.so.bak || true

# Current criu-dev behavior: generic libdrm check still enables AMDGPU
(cd /work && make -pn 2>/dev/null | grep -m1 CONFIG_AMDGPU || true)

# This then fails in amdgpu plugin compilation
(gcc $(pkg-config --cflags libdrm) -c /work/plugins/amdgpu/amdgpu_plugin.c -o /tmp/amdgpu.o) || true
'
```

rst0git:
```
# Simulate: libdrm present, AMDGPU dev bits missing
mv /usr/include/libdrm/amdgpu.h /usr/include/libdrm/amdgpu.h.bak || true
mv /usr/lib/x86_64-linux-gnu/libdrm_amdgpu.so /usr/lib/x86_64-linux-gnu/libdrm_amdgpu.so.bak || true
```
These lines look like intentionally breaking parts of the AMD GPU userspace stack :)

cheese-cakee:
@rst0git  I reran this with a package-native setup (no manual deletion of distro files):

- Built and installed `libdrm` from source with AMDGPU disabled (`-Damdgpu=disabled`) into an isolated prefix.
- Verified in that prefix:
  - `pkg-config --exists libdrm` -> yes
  - `include/libdrm/amdgpu.h` -> missing
  - `libdrm_amdgpu.so` -> missing

Results:
- Old logic equivalent (`pkg-config-check,libdrm`) would enable `CONFIG_AMDGPU`.
- New feature-test logic fails as expected:
  - `fatal error: libdrm/amdgpu.h: No such file or directory`
  - therefore `CONFIG_AMDGPU` stays disabled.

So this reproduces the exact mismatch using a real package build configuration, without mutating installed distro package contents.

cheese-cakee:
Adding exact code context that this PR changes:

**Old gate (current `criu-dev`):**
```make
ifeq ($(call pkg-config-check,libdrm),y)
export CONFIG_AMDGPU := y
endif
```
This only checks generic `libdrm` presence.

**New gate (`Makefile.config`):**
```make
LIBDRM_CFLAGS := $(shell $(PKG_CONFIG) --cflags libdrm 2>/dev/null)
ifeq ($(call try-cc,$(FEATURE_TEST_LIBDRM_AMDGPU),$(LIBDRM_CFLAGS) -ldrm -ldrm_amdgpu),true)
export CONFIG_AMDGPU := y
endif
```

**Feature test (`scripts/feature-tests.mak`):**
```c
#include <xf86drm.h>
#include <libdrm/amdgpu.h>
...
return amdgpu_device_initialize(-1, &major, &minor, &dev);
```
So we now require both headers and linkability to AMDGPU userspace before enabling the plugin.

cheese-cakee:
Hi @fdavid-amd,  following up on the review request from @avagin. The change is a surgical build-detection fix: adds a compile-time check for libdrm/amdgpu.h and ldrm_amdgpu before enabling CONFIG_AMDGPU. 

Happy to answer any questions.

cheese-cakee:
shall I change this PR to draft for now ? @rst0git

PR #2971, opened 2026-03-17
zdtm: fix TOCTOU race creating criu.tree directory

adrianreber:
When running tests in parallel (-p 2) with GCOV=1, multiple forked processes race to create ../criu.tree. The check-then- create pattern (os.access + os.mkdir) allows two processes to both see the directory as missing and then both attempt mkdir, causing FileExistsError in the second process.

Replace with os.makedirs(exist_ok=True) which handles the race atomically.

### DIFF
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -2018,9 +2018,7 @@ def do_run_test(tname, tdesc, flavs, opts):
         # to it.  Without this, those mounts propagate into the source and
         # become MNT_LOCKED inside the user namespace, causing
         # open_tree(OPEN_TREE_CLONE) to return EINVAL on restore.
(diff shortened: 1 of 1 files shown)


avagin:
@adrianreber the patch looks good to me. But can I ask you to clean up the commit message. You can use any tool to create patches, but let's avoid mentioning them in the commit message. Another option is to follow the linux kernel guidance: https://docs.kernel.org/process/coding-assistants.html

adrianreber:
> he patch looks good to me. But can I ask you to clean up the commit message. You can use any tool to create patches, but let's avoid mentioning them in the commit message. Another option is to follow the linux kernel guidance: https://docs.kernel.org/process/coding-assistants.html

Sure, done.

PR #2975, opened 2026-03-19
bfd: fix partial write handling in bwritev()

3idey:
When bfd is unbuffered, bwritev() used writev() directly, which can return partial writes and leave checkpoint images incomplete.

This patch fixes this by ensuring writev() is retried until all iovecs are fully written or an error occurs. 

Changes:
* Removes `const` from the `iov` parameter in `bwritev()` (and updates the header) to allow direct pointer manipulation.
* Handles partial writes by temporarily adjusting the `iov_base` and `iov_len` of the current iovec and advancing the `iov` pointer and `cnt` as data is successfully written.
* Avoids zero-progress loops and extra memory allocations by tracking the offset inline.

### DIFF
--- a/criu/bfd.c
+++ b/criu/bfd.c
@@ -275,16 +275,51 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
 	return __bwrite(bfd, buf, size);
 }
 
-int bwritev(struct bfd *bfd, const struct iovec *iov, int cnt)
+int bwritev(struct bfd *bfd, struct iovec *iov, int cnt)
 {
 	int i, written = 0;
 
 	if (!bfd_buffered(bfd)) {
-		/*
-		 * FIXME writev() should be called again if writev() writes
-		 * less bytes than requested.
-		 */
-		return writev(bfd->fd, iov, cnt);
+		size_t off = 0;
+
+		while (cnt) {
+			ssize_t ret;
+
+			/*
+			 * Temporarily modify the first iovec to skip already-written
+			 * bytes from previous partial writes, then restore it before
+			 * the next iteration.
+			 */
+			iov[0].iov_base += off;
+			iov[0].iov_len -= off;
+			ret = writev(bfd->fd, iov, cnt);
+			iov[0].iov_base -= off;
+			iov[0].iov_len += off;
+			if (ret < 0) {
+				pr_perror("writev failed for fd %d", bfd->fd);
+				return ret;
+			}
+			if (ret == 0) {
+				errno = EIO;
+				pr_perror("writev made no progress (fd %d)", bfd->fd);
+				return -1;
+			}
+
+			written += ret;
+			ret += off;
+			while (ret) {
+				if (iov[0].iov_len > ret) {
+					off = ret;
+					break;
+				}
+				ret -= iov[0].iov_len;
+				iov++;
+				cnt--;
+				off = 0;
+			}
+
+		}
+		return written;
 	}
 
 	for (i = 0; i < cnt; i++) {
--- a/criu/include/bfd.h
+++ b/criu/include/bfd.h
@@ -34,7 +34,7 @@ char *breadline(struct bfd *f);
 char *breadchr(struct bfd *f, char c);
 int bwrite(struct bfd *f, const void *buf, int sz);
 struct iovec;
-int bwritev(struct bfd *f, const struct iovec *iov, int cnt);
+int bwritev(struct bfd *f, struct iovec *iov, int cnt);
 int bread(struct bfd *f, void *buf, int sz);
 int bfd_flush_images(void);
 #endif


3idey:
@avagin Could you please take a look at this?

criu/bfd.c
```diff
@@ -275,16 +275,75 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+{
+	ssize_t n = 0;
+
+	while (cnt > 0) {
+		ssize_t ret = writev(fd, iov, cnt);
+		if (ret == -1) {
+			if (errno == EINTR)
+				continue;
+			if (errno == EAGAIN || errno == EWOULDBLOCK)
+				errno = EINVAL;
```
avagin: why?

criu/bfd.c
```diff
@@ -275,16 +275,75 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+ * partial writes. Use only with blocking I/O.
+ */
+static ssize_t bfd_writev_all(int fd, struct iovec *iov, int cnt)
+{
+	ssize_t n = 0;
+
+	while (cnt > 0) {
+		ssize_t ret = writev(fd, iov, cnt);
+		if (ret == -1) {
+			if (errno == EINTR)
```
avagin: we install all signal handlers with SA_RESTART, so we don't expect to get EINTR here

criu/bfd.c
```diff
@@ -275,16 +275,75 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+		}
+
 		/*
-		 * FIXME writev() should be called again if writev() writes
-		 * less bytes than requested.
+		 * We need to make a copy of iov because bfd_writev_all()
+		 * modifies it to track progress on partial writes.
 		 */
-		return writev(bfd->fd, iov, cnt);
+		iov_copy = xmemdup(iov, cnt * sizeof(struct iovec));
```
avagin: it isn't required

avagin: I mean we can avoid this xmemdup

criu/bfd.c
3idey: You are right, there is no good reason to rewrite EAGAIN/EWOULDBLOCK to EINVAL here. I removed errno remapping and now return writev errors as-is.

criu/bfd.c
3idey: Removed xmemdup as suggested.
Now bfd_writev_all handles partial writev by advancing iov and cnt in place until everything is written or an error occurs, so there is no extra allocation.
Current bwritev usage does not need the original iov after the call.
Is this what you had in mind?

3idey: Pushed an update. Dropped xmemdup and kept the iov array const by tracking partial progress with local variables (idx and off). This also fixes the CI test failures which were previously crashing on expected 0-length iovecs.

criu/bfd.c
```diff
@@ -275,16 +275,78 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+		if (iov[idx].iov_len <= off) {
+			idx++;
+			off = 0;
+			continue;
+		}
+
+		if (!off) {
+			ret = writev(fd, &iov[idx], cnt - idx);
+		} else {
+			struct iovec cur = {
```
avagin: I think it should be okay to temporary change iov when it is required...

3idey: Got it! I reverted and just advanced iov and cnt directly. It makes it much simpler. I kept off to track the partial writes and skip empty iovecs so we don't trip up the tests.

avagin: should we do something like:
```
diff --git a/criu/bfd.c b/criu/bfd.c
index 2c5f5b64a..23059c9b8 100644
--- a/criu/bfd.c
+++ b/criu/bfd.c
@@ -275,16 +275,40 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
        return __bwrite(bfd, buf, size);
 }
 
-int bwritev(struct bfd *bfd, const struct iovec *iov, int cnt)
+int bwritev(struct bfd *bfd, struct iovec *iov, int cnt)
 {
        int i, written = 0;
 
        if (!bfd_buffered(bfd)) {
-               /*
-                * FIXME writev() should be called again if writev() writes
-                * less bytes than requested.
-                */
-               return writev(bfd->fd, iov, cnt);
+               size_t off = 0;
+
+               while (cnt) {
+                       int ret;
+
+                       iov[0].iov_base += off;
+                       iov[0].iov_len -= off;
+                       ret = writev(bfd->fd, iov, cnt);
+                       iov[0].iov_base -= off;
+                       iov[0].iov_len += off;
+                       if (ret < 0) {
+                               pr_perror("writev");
+                               return ret;
+                       }
+
+                       written += ret;
+                       while (ret) {
+                               if (iov[0].iov_len > ret) {
+                                       off += ret;
+                                       break;
+                               }
+                               ret -= iov[0].iov_len;
+                               iov++;
+                               cnt--;
+                               off = 0;
+                       }
+
+               }
+               return written;
        }
 
        for (i = 0; i < cnt; i++) {
diff --git a/criu/include/bfd.h b/criu/include/bfd.h
index ca983d7b4..057fad2fe 100644
--- a/criu/include/bfd.h
+++ b/criu/include/bfd.h
@@ -34,7 +34,7 @@ char *breadline(struct bfd *f);
 char *breadchr(struct bfd *f, char c);
 int bwrite(struct bfd *f, const void *buf, int sz);
 struct iovec;
-int bwritev(struct bfd *f, const struct iovec *iov, int cnt);
+int bwritev(struct bfd *f, struct iovec *iov, int cnt);
 int bread(struct bfd *f, void *buf, int sz);
 int bfd_flush_images(void);
 #endif
```

3idey: Done - removed the helper function, changed the signature to non-const, and simplified it to just temporarily modify the first iovec to track partial writes. Much cleaner code now.

3idey:
@avagin updated! let me know if this looks good to you.

criu/bfd.c
```diff
@@ -275,16 +275,45 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+
+		while (cnt) {
+			int ret;
+
+			/*
+			 * Temporarily modify the first iovec to skip already-written
+			 * bytes from previous partial writes, then restore it before
+			 * the next iteration.
+			 */
+			iov[0].iov_base = (char *)iov[0].iov_base + off;
```
avagin: why do you need to convert it to (char *)?

criu/bfd.c
```diff
@@ -275,16 +275,45 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+			iov[0].iov_base = (char *)iov[0].iov_base - off;
+			iov[0].iov_len += off;
+			if (ret < 0) {
+				pr_perror("writev");
+				return ret;
+			}
+
+			written += ret;
+			while (ret) {
+				if (iov[0].iov_len > ret) {
```
avagin, line 311: I think here is a bug. we adjusted iov_len after the write syscall. We probably need to do ret += off before this loop.

criu/bfd.c
3idey: yeah u right, we dont need it. dropped the cast.

3idey:
@avagin,I think we should ignore the copilot comments, it's suggesting things we already decided against (like adding EINTR checks and keeping iovec constness). let me know if we're good to merge!
Thank you.

criu/bfd.c
```diff
@@ -275,16 +275,48 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+			 */
+			iov[0].iov_base += off;
+			iov[0].iov_len -= off;
+			ret = writev(bfd->fd, iov, cnt);
+			iov[0].iov_base -= off;
+			iov[0].iov_len += off;
+			if (ret < 0) {
+				pr_perror("writev");
+				return ret;
+			}
```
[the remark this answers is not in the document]
avagin: irrelevant

criu/bfd.c
```diff
@@ -275,16 +275,48 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
-		/*
-		 * FIXME writev() should be called again if writev() writes
-		 * less bytes than requested.
-		 */
-		return writev(bfd->fd, iov, cnt);
+		size_t off = 0;
+
+		while (cnt) {
+			int ret;
+
```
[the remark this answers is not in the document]
avagin: please fix

criu/bfd.c
```diff
@@ -321,16 +275,48 @@ int bwrite(struct bfd *bfd, const void *buf, int size)
...
+			 * bytes from previous partial writes, then restore it before
+			 * the next iteration.
+			 */
+			iov[0].iov_base += off;
+			iov[0].iov_len -= off;
+			ret = writev(bfd->fd, iov, cnt);
+			iov[0].iov_base -= off;
+			iov[0].iov_len += off;
+			if (ret < 0) {
+				pr_perror("writev");
```
avagin: the error should be a bit more descriptive

avagin:
> @avagin,I think we should ignore the copilot comments, it's suggesting things we already decided against (like adding EINTR checks and keeping iovec constness). let me know if we're good to merge! Thank you.

A few suggestions look valid to me.

PR #2985, opened 2026-03-23
pipes: restore pipe ownership to fix /proc/self/fd access

### LINKED ISSUE #2984: Pipe Permissions on Restore: owned by root, not readable by process
I am attempting to checkpoint a program that uses unix pipes for communication within its own process.  In particular, it attempts to open `/proc/%d/fd/%d`.  CRIU's restore completes successfully but during restored program execution, the open() call on `/proc/%d/fd/%d` fails with invalid permission.

I believe criu launched as root creates the pipe with `rw-------` mode, then attaches the file descriptors.  Later when application is restored it is no longer running as root, and the open fails.

Here is `ls -l` and `stat -L` on the fd when criu is run as root:
```
lr-x------ 1 lrobison lrobison 64 Mar 23 19:11 /proc/432216/fd/3 -> 'pipe:[20980593]'
  File: /proc/432216/fd/3
  Size: 0               Blocks: 0          IO Block: 4096   fifo
Device: 0,14    Inode: 20980593    Links: 1
Access: (0600/prw-------)  Uid: (    0/    root)   Gid: (    0/    root)
```

Here is the same if restoring with --unprivileged and running criu as my user:
```
lr-x------ 1 lrobison lrobison 64 Mar 23 19:18 /proc/432612/fd/3 -> 'pipe:[21002808]'
  File: /proc/432612/fd/3
  Size: 0               Blocks: 0          IO Block: 4096   fifo
Device: 0,14    Inode: 21002808    Links: 1
Access: (0600/prw-------)  Uid: ( 1050/lrobison)   Gid: ( 1051/lrobison)
```
And in this case the application can continue as expected.

However, for other reasons I need restore to run as root.  Given that CRIU should know it will hand off the pipe to the unprivileged user, can CRIU change the pipe's ownership to that user?

<details>
<summary>Small reproducer</summary>

```
/*
 * pipetest.c - Minimal CRIU reproducer: open a pipe, checkpoint, then
 * re-open the read end via /proc/self/fd/<n> after restore.
 *
 * Steps to reproduce:
 *   1. Build:   gcc -o pipetest pipetest.c
 *   2. Run:     ./pipetest &
 *   3. In another terminal, checkpoint:
 *        criu dump -t $(pgrep pipetest) -D /tmp/pipetest-img --shell-job
 *   4. Restore:
 *        criu restore -D /tmp/pipetest-img --shell-job
 *
 * Expected: open(/proc/self/fd/<read_end>) succeeds after restore.
 * Observed: open() fails with Permission denied.
 */

#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <unistd.h>

int main(void)
{
    int fds[2];

    if (pipe(fds) < 0) {
        perror("pipe");
        return 1;
    }

    printf("pipe read_fd=%d write_fd=%d\n", fds[0], fds[1]);
    printf("sleeping 20s — checkpoint pid %d now\n", getpid());
    fflush(stdout);

    sleep(20);
    printf("Continuing.  5 seconds of sleep for debug.\n");
    sleep(5);

    /* After restore, attempt to open the read end via /proc/self/fd */
    char path[64];
    snprintf(path, sizeof(path), "/proc/%d/fd/%d", getpid(), fds[0]);

    printf("re-opening pipe read end via %s\n", path);
    fflush(stdout);

    int reopen_fd = open(path, O_RDONLY);
    if (reopen_fd < 0) {
        fprintf(stderr, "open(%s) failed: %s\n", path, strerror(errno));
        return 1;
    }

    printf("open succeeded: new_fd=%d\n", reopen_fd);

    /* Quick sanity check: write to write end, read from re-opened fd */
    const char msg[] = "hello through pipe\n";
    if (write(fds[1], msg, sizeof(msg) - 1) < 0) {
        perror("write");
        return 1;
    }

    char buf[64] = {0};
    ssize_t n = read(reopen_fd, buf, sizeof(buf) - 1);
    if (n < 0) {
        perror("read");
        return 1;
    }

    printf("read %zd bytes: %s", n, buf);

    close(reopen_fd);
    close(fds[0]);
    close(fds[1]);
    return 0;
}
```

</details>

3idey:
## Summary

When CRIU runs as root and restores a pipe, the pipe inode is created with root ownership. If the restored process runs as a non-root user and tries to reopen the pipe via `/proc/self/fd/<n>`, the `open()` fails with `EACCES`.

## Solution

- Add `uid` and `gid` fields to `pipe_entry` in `images/pipe.proto`
- Save pipe ownership during dump from `stat()`
- Restore ownership with `fchown()` after `pipe()` creation
- Add a test `pipe_owenr` that verifies pipe ownership (uid/gid) is preserved across
checkpoint/restore.

This matches how memfds and TTYs already handle ownership restoration.

## Test plan

- `zdtm/static/pipe00` passes
- `zdtm/static/pipe01` passes
- `zdtm/static/pipe02` passes
- `zdtm/static/pipe_owner` passes after the fix and fails when we roll it back

Fixes: #2984

### DIFF
--- a/criu/pipes.c
+++ b/criu/pipes.c
@@ -305,6 +305,13 @@ int open_pipe(struct file_desc *d, int *new_fd)
 		return -1;
 	}
 
+	if (pi->pe->has_uid && pi->pe->has_gid) {
+		if (cr_fchown(pfd[0], pi->pe->uid, pi->pe->gid) < 0) {
+			pr_perror("Can't change pipe ownership");
+			return -1;
+		}
+	}
+
 	ret = restore_pipe_data(CR_FD_PIPES_DATA, pfd[1], pi->pe->pipe_id, pd_hash_pipes);
 	if (ret)
 		return -1;
@@ -495,6 +502,10 @@ static int dump_one_pipe(int lfd, u32 id, const struct fd_parms *p)
 	pe.pipe_id = pipe_id(p);
 	pe.flags = p->flags & ~O_DIRECT;
 	pe.fown = (FownEntry *)&p->fown;
+	pe.uid = userns_uid(p->stat.st_uid);
+	pe.gid = userns_gid(p->stat.st_gid);
+	pe.has_uid = true;
+	pe.has_gid = true;
 
 	fe.type = FD_TYPES__PIPE;
 	fe.id = pe.id;
--- a/images/pipe.proto
+++ b/images/pipe.proto
@@ -10,4 +10,6 @@ message pipe_entry {
 	required uint32		pipe_id		= 2;
 	required uint32		flags		= 3 [(criu).hex = true];
 	required fown_entry	fown		= 4;
+	optional uint32		uid		= 5;
+	optional uint32		gid		= 6;
 }
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -67,6 +67,7 @@ TST_NOFILE	:=				\
 		pipe00				\
 		pipe01				\
 		pipe02				\
+		pipe_owner			\
 		pthread00			\
 		pthread00-pac			\
 		pthread01			\
--- a/test/zdtm/static/pipe_owner.c
+++ b/test/zdtm/static/pipe_owner.c
@@ -1,0 +1,114 @@
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+#include <fcntl.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <linux/limits.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check pipe ownership is preserved after C/R";
+const char *test_author = "Ahmed Elaidy <elaidya225@gmail.com>";
+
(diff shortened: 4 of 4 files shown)


avagin:
could you please introduce a zdtm test for this change?

3idey:
@avagin Done.

criu/pipes.c
```diff
@@ -305,6 +305,17 @@ int open_pipe(struct file_desc *d, int *new_fd)
 		return -1;
 	}
 
+	if (pi->pe->has_uid && pi->pe->has_gid) {
+		if (cr_fchown(pfd[0], pi->pe->uid, pi->pe->gid) < 0) {
+			pr_perror("Can't change pipe ownership");
+			return -1;
+		}
+		if (cr_fchown(pfd[1], pi->pe->uid, pi->pe->gid) < 0) {
```
avagin: I think it is actually not required to set uid/gid on both ends. Each pipe has just one associated inode.

3idey: Done. `fchown` is called on one pipe end

Issue #2951, opened 2026-03-10
Inherited File Descriptor Issue When Checkpointing Processes in Wayland/Hyprland Sessions


Abdullah-Badawy1:
Hi
I've been exploring CRIU and ran into an interesting real-world problem while trying to checkpoint a simple Python HTTP server running inside a Hyprland (Wayland compositor) session. I'd like to describe the issue and discuss whether it represents an opportunity for improvement in CRIU.

**What I Was Doing**

I started a Python HTTP server and attempted to checkpoint and restore it using CRIU:

```bash
python3 -m http.server 8080 &
SERVER_PID=$!

sudo criu dump \
    -t $SERVER_PID \
    -D /tmp/webserver-ckpt \
    --shell-job \
    --display-stats

sudo criu restore \
    -D /tmp/webserver-ckpt \
    --shell-job \
    -d
```

**The Error I Got**

The restore failed with:

```
504149: Error (criu/files-reg.c:2175): File run/user/1000/hypr/
4b07770b9ef1cceb2e6f56d33538aaffb9186b9c_1773096523_1849152812/
hyprland.log has bad size 5230917 (expect 5222646)
504149: Error (criu/files.c:1221): Unable to open fd=8 id=0x24
Error (criu/cr-restore.c:2324): Restoring FAILED.
```

**Root Cause Analysis**

After investigating, I traced the problem to Linux's file descriptor inheritance model. When the Python server was launched from a terminal inside a Hyprland session, it silently inherited open file descriptors from its parent processes, including Hyprland's active runtime log file at `/run/user/1000/hypr/.../hyprland.log`.

The process tree looked like this:

```
Hyprland (compositor) — has hyprland.log open for writing
    └── terminal emulator
            └── shell
                    └── python3 -m http.server  ← what we dumped
                            └── fd 8 → hyprland.log (inherited)
```

CRIU correctly recorded the file size at dump time (5,222,646 bytes). However, because Hyprland continued actively writing to its log between the dump and restore operations, the file had grown to 5,230,917 bytes by restore time — a difference of about 8KB. CRIU's file validation caught this mismatch and hard-failed the restore.

**Why This Matters**

[...]

PR #2992, opened 2026-03-29
external: normalize unambiguous file id lookup

cheese-cakee:
## Why this PR is necessary

external_lookup_id() used strict string comparison, so logically identical file[mnt_id:inode] IDs could fail to
match when represented differently.

## What this PR changes

- Adds numeric normalization for file[mnt_id:inode] lookup in criu/external.c, with strict compatibility guards.
 - Keeps exact strcmp() match as the first path.
 - Applies normalization only when both components are unambiguous:
  - 0x... => hex
  - contains a-f/A-F => hex
  - digits-only => ambiguous, no normalization
 - Prevents ambiguous cross-matching (e.g. 11 vs 17) for backward compatibility.

## Scope

- Single logical change in criu/external.c only.

## Testing

- `git diff --check HEAD~1..HEAD`
- `python3 -m py_compile test/zdtm.py test/zdtm/criu_config.py`


cheese-cakee:
please review when possible @avagin

avagin:
@cheese-cakee please update the commit message with more details.

criu/external.c
```diff
@@ -39,10 +94,18 @@ int add_external(char *key)
...
+	unsigned int id_mnt_id, ext_mnt_id;
+	uint64_t id_inode, ext_inode;
+	bool normalized_id;
+
+	normalized_id = parse_external_file_id(id, &id_mnt_id, &id_inode);
 
 	list_for_each_entry(ext, &opts.external, node)
 		if (!strcmp(ext->id, id))
 			return true;
+		else if (normalized_id && parse_external_file_id(ext->id, &ext_mnt_id, &ext_inode) &&
```
avagin, line 112: If we have file[a:a] and file[10:10], after this change a wrong external entry can be looked up... can't it?

avagin:
I get the goal here, but we need to be careful. Since the file[hex:hex] format lacks the 0x prefix, any hex value consisting only of numbers could be parsed as a decimal either...

cheese-cakee:
> I get the goal here, but we need to be careful. Since the file[hex:hex] format lacks the 0x prefix, any hex value consisting only of numbers could be parsed as a decimal either...

I addressed this by making CRIU emit the 0x prefix (changed the snprintf in dump_one_reg_file from %x to 0x%x), so the format is unambiguous going forward. The parser now requires 0x for hex

criu/external.c
cheese-cakee: I've changed the approach: CRIU now emits file[0x%x:0x%PRIx64] so the output carries an explicit 0x prefix, and the parser only interprets 0x-prefixed values as hex. Bare numbers are always decimal

avagin:
> > I get the goal here, but we need to be careful. Since the file[hex:hex] format lacks the 0x prefix, any hex value consisting only of numbers could be parsed as a decimal either...
> 
> I addressed this by making CRIU emit the 0x prefix (changed the snprintf in dump_one_reg_file from %x to 0x%x), so the format is unambiguous going forward. The parser now requires 0x for hex

I don't understand how it will help us. We can't break existing users, so we need to proper handle the old format...

rst0git:
Adding a note that I believe this is also related to the comments in https://github.com/checkpoint-restore/criu/issues/2951#issuecomment-4076295635 and https://github.com/checkpoint-restore/criu/pull/2973

criu/external.c
```diff
@@ -8,6 +13,57 @@
 
 #include "net.h"
 
+static bool parse_external_id_component(const char *id, char delim, char **end, unsigned long long *val)
+{
+	char *tmp;
+	int base;
+
+	/*
+	 * 0x prefix means hexadecimal; everything else is decimal.
```
rst0git: I believe this is the breaking change that Andrei is referring to. There is no reliable way for us to parse decimal numbers here. For example, `0x11 == 17`. When `0x` is removed, we have no way of knowing if the input is `11` or `17`.

cheese-cakee: I'll drop the emission change and instead make the parser try decimal first (matching user expectation for --inherit-fd arguments) and fall back to hex only when the decimal parse fails

cheese-cakee:
> > > I get the goal here, but we need to be careful. Since the file[hex:hex] format lacks the 0x prefix, any hex value consisting only of numbers could be parsed as a decimal either...
> > 
> > 
> > I addressed this by making CRIU emit the 0x prefix (changed the snprintf in dump_one_reg_file from %x to 0x%x), so the format is unambiguous going forward. The parser now requires 0x for hex
> 
> I don't understand how it will help us. We can't break existing users, so we need to proper handle the old format...

I am reverting that and instead use a decimal-first fallback parser that correctly handles both the existing %x output and user-provided decimal IDs

criu/external.c
avagin: again, it can break backward compatibility. Please explain in the commit message in details why existing users will be not affected by this change. Try to think on a case when there are two external resources with ids '11' and '17'.

cheese-cakee: Now normalization is applied only when both components (mnt_id and inode) are unambiguous:

   - 0x... => unambiguous hex
   - contains a-f/A-F => unambiguous hex
   - digits-only => ambiguous, no normalization

  So file[11:17] and file[17:23] keep literal/exact semantics and won’t be cross-matched by numeric normalization.
  external_lookup_id() still does exact strcmp() first, then numeric compare only for unambiguous IDs

rst0git: @cheese-cakee Were you able to test these changes? It sounds overcomplicated, and I am not sure the logic is correct. Note that AI chatbots can make mistakes.

cheese-cakee: I have tested the new changes and I have tried to simplify the implementation : 
Normalization only when both mnt_id AND inode are unambiguous:
   - 0x... or contains a-f → hex
   - digits-only (like 11) → treated as decimal, no normalization 
   
please review if its alright

cheese-cakee:
please review when possible @avagin @rst0git

PR #2994, opened 2026-03-31
Add scripts to inject container mount points during the CRIU restore

wangixt:
When managing inference containers (e.g. vLLM) on Kubernetes, restore flows often need extra mount points for model weights so that model switching can happen without rebuilding container images.

This commit introduces:
- scripts/criu-move-mount.c:
  - a CRIU action helper that parses CRIU_ADD_MOUNTS and performs bind/rbind mounts into the restored container mount namespace during pre-resume.
- scripts/runc-action-add-mounts.sh:
  - an action-script wrapper that compiles criu-move-mount.c at runtime and executes it during CRIU restore.

The solution enables best-effort mount injection at restore time, reducing operational friction for dynamic model weight switching.


rst0git:
@wangixt Why do you want these scripts to be in the CRIU repository?

wangixt:
> @wangixt Why do you want these scripts to be in the CRIU repository?

Thanks for raising this — that’s a fair question.

Our motivation for proposing these scripts in the CRIU repo is that they solve a generic CRIU restore-time workflow, not a vLLM-specific one:

1. During restore, users sometimes need to inject additional bind mounts in pre-resume (e.g., model/data paths that are decided at runtime).
2. The scripts use CRIU’s standard --action-script mechanism and CRTOOLS_* env vars, so they are tightly coupled to CRIU restore semantics.
3. We think having a minimal reference implementation in-tree lowers adoption friction and avoids each runtime/operator re-implementing similar logic.

That said, we understand scope concerns.
If this is considered too runtime-specific for core CRIU, we’re happy to:
- move it under contrib/ or scripts/examples/, or
- maintain it in a separate repository and only keep documentation pointers here.

We’re flexible on placement; the main goal is to provide a reusable restore-time mount injection pattern for CRIU users. @rst0git

rst0git:
@wangixt Have you considered using the "prestart" OCI runtime hook or just modifying the container config to include the additional mounts?
- https://github.com/opencontainers/runtime-spec/blob/main/config.md#prestart
- https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle

Using a CRIU action-script to inject mounts doesn't sound like a reliable approach for doing this.

wangixt:
> @wangixt Have you considered using the "prestart" OCI runtime hook or just modifying the container config to include the additional mounts?
> 
> * https://github.com/opencontainers/runtime-spec/blob/main/config.md#prestart
> * https://github.com/opencontainers/runtime-spec/blob/main/runtime.md#lifecycle
> 
> Using a CRIU action-script to inject mounts doesn't sound like a reliable approach for doing this.

@rst0git  Thanks, good suggestion. We’ll first validate config/prestart-hook based mount injection in our restore workflow.

PR #2996, opened 2026-04-06
dump: skip POSIX timers targeting dead threads

pojntfx:
**What**:

When a POSIX timer uses SIGEV_THREAD_ID and the target thread has exited, dump fails in PID namespaces (`Unable to convert the notify thread id`) and restore fails without PID namespaces (`EINVAL` from timer_create()). We noticed this when an application switched from Java 8 to Java 21.

**How**:

To fix this, this PR detects dead notify threads during dump and skips the timer instead of failing. The kernel already silently drops signals from these timers, so skipping them during dump is safe.

**Tests**:

- `zdtm/static/dead_thread_posix_timer`
~~- `zdtm/static/dead_thread_posix_timer_h`~~

~~I added `dead_thread_posix_timer_h` since I wanted to specifically trigger the restore failure case, which needs host mode. I followed the convention that `pthread_timers.c` and `pthread_timers_h.c` uses.~~


test/zdtm/static/Makefile
```diff
@@ -77,6 +77,7 @@ TST_NOFILE	:=				\
 		pthread_timers_h		\
 		leader_posix_timer		\
 		dead_thread_posix_timer		\
+		dead_thread_posix_timer_nons		\
```
avagin, line 80: It isn't required. The previous test will be executed in non-ns, ns and uns envs.

rst0git:
@pojntfx Would it be possible to add commit messages?
- https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes
- https://chris.beams.io/posts/git-commit/

criu/timer.c
```diff
@@ -387,13 +394,21 @@ int parasite_dump_posix_timers_seized(struct proc_posix_timers_stat *proc_args,
 		goto end_posix;
 
 	i = 0;
+	j = 0;
```
rst0git, line 397: Would it be possible to add a comment to explain what we do here?
```
  /*
   * i advances every iteration (indexes parasite-filled args->timer[]);                                                                                                                                            
   * j advances only for timers we keep, both the next output slot                                                                                                                                            
   * and the final count. Timers whose notify thread has exited are                                                                                                                                                 
   * skipped (i.e., j <= i).                                                                                                                                                                                            
   */ 
```

pojntfx: Good idea; resolved via 0b360035e8cc52fdf395db52d3fe41d340273132

test/zdtm/static/Makefile
pojntfx: I think this was a review on an older commit pushed as part of this PR? From the PR description:

> I added dead_thread_posix_timer_h since I wanted to specifically trigger the restore failure case, which needs host mode. I followed the convention that pthread_timers.c and pthread_timers_h.c uses.

I'll remove it regardless, it was mostly used as a reproducer initially. Resolved via c2459602c7844c2d98de1b2dc9063928286e5f75

pojntfx:
@rst0git @avagin Thank you for your reviews. I collapsed it into a single commit and rebased on the latest `criu-dev`. I already added commit messages to the original commits:

Screenshot showing the commit messages

When collapsing them into a single new commit I also added a body based on the PR description. Let me know if I should make any further changes :)

3idey:
>  When collapsing them into a single new commit I also added a body based on the PR description. Let me know if I should make any further changes :)

@pojntfx Thank you!

Could you please separate this into two commits? The PR currently squashes the core fix and the ZDTM test into a single commit. According to the [Contribution Guidelines](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#separate-your-changes), each logical change should be separated into a distinct patch.

pojntfx:
@3idey Done!

avagin:
> To fix this, this PR detects dead notify threads during dump and skips the timer instead of failing. The kernel already silently drops signals from these timers, so skipping them during dump is safe.

These timers can be still monitored using timer_gettime. Applications can use that. Should we restore them as SIGEV_NONE (https://man7.org/linux/man-pages/man2/timer_create.2.html)?

pojntfx:
@avagin Didn't know about SIGEV_NONE. I guess that would allow us to still monitor them while also not notifying any dead threads anymore? Should this best be done as a follow-up PR?

avagin:
> @avagin Didn't know about SIGEV_NONE. I guess that would allow us to still monitor them while also not notifying any dead threads anymore? Should this best be done as a follow-up PR?

I think it’s easier to do that in this PR. Basically, instead of skipping timers, you just need to modify them to use SIGEV_NONE.

aidancully:
> > @avagin Didn't know about SIGEV_NONE. I guess that would allow us to still monitor them while also not notifying any dead threads anymore? Should this best be done as a follow-up PR?
> 
> I think it’s easier to do that in this PR. Basically, instead of skipping timers, you just need to modify them to use SIGEV_NONE.

Hi @avagin, @pojntfx asked me to look at this, and I've implemented the change you requested in 3730eb7ced6421e4f8ce6987d75bb1096620b61a. Let me know if you'd like anything else!

avagin:
@aidancully please update this pr or create a new one. Could you update the test to check the new behaviour? Thanks.

aidancully:
> @aidancully please update this pr or create a new one. Could you update the test to check the new behaviour? Thanks.

Hi @avagin, I've updated the test in 3ef16f1fdf3daa0f13ae85e0d3c5956189740dbc. Let me know if you need anything else, I assume I'll need to fix the commit history prior to merge. Thanks!

Issue #2995, opened 2026-04-01
criu-dev fails with "Can't dump page with parasite" in runc ci


kolyshkin:
We're testing criu-dev as part of runc ci. It appears to fail now.

Here's criu version output:
```
Version: 4.2
GitID: 78a7120
```

Here's the failure (from https://github.com/opencontainers/runc/actions/runs/23871278688/job/69603647307?pr=5212) runc CI job:

```
=== RUN   TestCheckpoint
time="2026-04-01T21:18:58Z" level=warning msg="--- Quoting \"/tmp/TestCheckpoint3042148804/003/criu/dump.log\""
time="2026-04-01T21:18:58Z" level=warning msg="718:(00.016747) page-xfer: \tbuf 2/1"
time="2026-04-01T21:18:58Z" level=warning msg="719:(00.016748) page-xfer: \tp 0x7faf03233000 - 0x7faf03235000"
time="2026-04-01T21:18:58Z" level=warning msg="720:(00.016755) page-xfer: \tbuf 3/1"
time="2026-04-01T21:18:58Z" level=warning msg="721:(00.016758) page-xfer: \th 0x7faf0323b000 [2]"
time="2026-04-01T21:18:58Z" level=warning msg="722:(00.016759) page-xfer: Checking 0x7faf0323b000 - 0x7faf0323d000 hole"
time="2026-04-01T21:18:58Z" level=warning msg="723:(00.016761) Error (criu/page-xfer.c:299): page-xfer: Missing 7faf0323b000 in parent pagemap"
time="2026-04-01T21:18:58Z" level=warning msg="724:(00.016764) Error (criu/page-xfer.c:344): page-xfer: Hole 0x7faf0323b000 - 0x7faf0323d000 not found in parent"
time="2026-04-01T21:18:58Z" level=warning msg="725:(00.016777) page-pipe: Killing page pipe"
time="2026-04-01T21:18:58Z" level=warning msg="726:(00.016798) ----------------------------------------"
time="2026-04-01T21:18:58Z" level=warning msg="727:(00.016799) Error (criu/mem.c:718): Can't dump page with parasite"
time="2026-04-01T21:18:58Z" level=warning msg=...
time="2026-04-01T21:18:58Z" level=warning msg="768:(00.017086) net: Unlock network"
time="2026-04-01T21:18:58Z" level=warning msg="769:(00.017089) Running network-unlock scripts"
[...]

PR #2997, opened 2026-04-06
page-xfer: write zero pages when parent hole is missing

rst0git:
During incremental dump with pre-dumps, pages allocated between pre-dumps may have their soft-dirty bit clear (the kernel zero-fills new mmap'd pages without setting it). CRIU classifies these as PE_PARENT holes, but since the page is newly allocated, no parent has it.

This causes restore to fail with `"Missing in parent pagemap"` (https://github.com/checkpoint-restore/criu/issues/2995) because `write_pagemap_loc()` wrote a broken `PE_PARENT` entry referencing data that does not exist in any parent image.
    
To fix this, write zero-filled PE_PRESENT pages instead, which is correct because new mmap'd pages with a clear soft-dirty bit are guaranteed to be zero-filled by the kernel.

### DIFF
--- a/test/zdtm/transition/maps_newpage.c
+++ b/test/zdtm/transition/maps_newpage.c
@@ -1,0 +1,139 @@
+#include <fcntl.h>
+#include <stdlib.h>
+#include <string.h>
+#include <unistd.h>
+#include <sys/mman.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Test pages with cleared soft-dirty after pre-dump";
+const char *test_author = "Radostin Stoyanov <rstoyanov@fedoraproject.org>";
+
+/*
+ * After a pre-dump, soft-dirty bits are cleared. If new pages are
+ * then written and soft-dirty is cleared again (e.g. by a device
+ * driver or by writing to /proc/self/clear_refs), CRIU sees
+ * "present, not dirty" and classifies them as PE_PARENT. But the
+ * parent has no data for these pages, causing restore to fail with
+ * "Hole not found in parent".
+ *
+ * This test simulates the scenario by:
+ * 1. Mapping and writing init pages (will be in pre-dump)
+ * 2. Letting the pre-dump run
+ * 3. Mapping new pages in a separate VMA and writing a pattern
+ * 4. Clearing soft-dirty via /proc/self/clear_refs
+ * 5. Re-dirtying init pages so they are dumped normally
+ * 6. Letting the final dump run — the new pages appear "clean"
+ *    but have no parent data
+ */
+
+#define INIT_PAGES	4
+#define NEW_PAGES	4
+
+static int clear_soft_dirty(void)
+{
+	int fd;
+	ssize_t ret;
+
+	fd = open("/proc/self/clear_refs", O_WRONLY);
+	if (fd < 0) {
+		pr_perror("open clear_refs");
+		return -1;
+	}
+	/* 4 = clear soft-dirty bits */
+	ret = write(fd, "4", 1);
+	close(fd);
+	if (ret != 1) {
+		pr_perror("write clear_refs");
+		return -1;
+	}
+	return 0;
+}
+
+int main(int argc, char **argv)
+{
+	unsigned char *init_mem, *new_mem = NULL;
+	size_t init_size = INIT_PAGES * PAGE_SIZE;
+	size_t new_size = NEW_PAGES * PAGE_SIZE;
+	int i;
+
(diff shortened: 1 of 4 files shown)


avagin:
> the kernel zero-fills new mmap'd pages without setting it

We should detect zero pages... Can you be more precise when it can happen? How to reproduce this case from user-space?

test/zdtm/transition/maps_newpage.c
```diff
@@ -0,0 +1,139 @@
...
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Test pages with cleared soft-dirty after pre-dump";
+const char *test_author = "Radostin Stoyanov <rstoyanov@fedoraproject.org>";
+
+/*
+ * After a pre-dump, soft-dirty bits are cleared. If new pages are
+ * then written and soft-dirty is cleared again (e.g. by a device
+ * driver or by writing to /proc/self/clear_refs), CRIU sees
```
avagin, line 15: > soft-dirty is cleared again

Only CRIU has to clear soft-dirty bits, otherwise we can't guarantee consistent memory snapshot. I don't like the idea of using /proc/self/clear_refs in the test.

avagin:
Can it be a kernel bug like: https://github.com/torvalds/linux/commit/29f40594a28114b9a9bc87f6cf7bbee9609628f2 or https://github.com/torvalds/linux/commit/6707915e030a3258868355f989b80140c1a45bbe

avagin:
> Can it be a kernel bug like: [torvalds/linux@29f4059](https://github.com/torvalds/linux/commit/29f40594a28114b9a9bc87f6cf7bbee9609628f2) or [torvalds/linux@6707915](https://github.com/torvalds/linux/commit/6707915e030a3258868355f989b80140c1a45bbe)

I think you triggered the second issue. Here is a test to reproduce it: https://github.com/avagin/criu/commit/daf5e4d6fe9fbe6f3db219449ff9a163226b353e. I think we need to send the kernel fix to stable trees.

avagin:
> > Can it be a kernel bug like: [torvalds/linux@29f4059](https://github.com/torvalds/linux/commit/29f40594a28114b9a9bc87f6cf7bbee9609628f2) or [torvalds/linux@6707915](https://github.com/torvalds/linux/commit/6707915e030a3258868355f989b80140c1a45bbe)
> 
> I think you triggered the second issue. Here is a test to reproduce it: [avagin@daf5e4d](https://github.com/avagin/criu/commit/daf5e4d6fe9fbe6f3db219449ff9a163226b353e). I think we need to send the kernel fix to stable trees.

@3idey I know you want to do something with the linux kernel. could you try to back port https://github.com/torvalds/linux/commit/6707915e030a3258868355f989b80140c1a45bbe to stable branches?

3idey:
> @3idey I know you want to do something with the linux kernel. could you try to back port [torvalds/linux@6707915](https://github.com/torvalds/linux/commit/6707915e030a3258868355f989b80140c1a45bbe) to stable branches?

@avagin Okay, I'll take a look at the commit and try to backport it to the stable trees. Is there a specific stable branch you'd like me to target first?

avagin:
> > @3idey I know you want to do something with the linux kernel. could you try to back port [torvalds/linux@6707915](https://github.com/torvalds/linux/commit/6707915e030a3258868355f989b80140c1a45bbe) to stable branches?
> 
> @avagin Okay, I'll take a look at the commit and try to backport it to the stable trees. Is there a specific stable branch you'd like me to target first?

You can start with 6.18 which is the lts release: https://www.kernel.org/category/releases.html

3idey:
@avagin I've submitted the backport to the stable mailing list. Since 6.18.y doesn't have the VM_STICKY infrastructure yet, I cherry-picked the full dependency chain (9 patches). The series is here: https://lore.kernel.org/stable/20260424211315.1072123-1-elaidya225@gmail.com/T/#t
Let me know if you'd prefer a minimalist standalone fix instead

avagin:
@3idey this isn't quite what I expected. I thought you were going to show us this patch first. I suspect they will reject the backport of the entire series.

3idey:
@avagin, the 6.18.y backport is in the stable queue. Still-maintained LTS trees that could need it are 6.12, 6.6, 6.1, 5.15, and 5.10. The older ones don't have the VM_STICKY infrastructure, so each would need a minimal standalone fix instead of the full chain. What do you think we should do now?

avagin:
@3idey Good job! Thanks a lot for working on this. I am not sure if we need to backport the fix into these older releases.

PR #2998, opened 2026-04-10
criu: optimize mntinfo_add_list to O(1)

3idey:
currently, `mntinfo_add_list()` appends elements to the `mntinfo` linked list using an O(N) iterative loop to find the tail. 

this patch introduces a static `mntinfo_tail` pointer to track the end of the list, making appending an O(1) operation. it also ensures `mntinfo_tail` correctly handles appending entire sub-lists by advancing to the actual end, and keeps the tail synced whenever `mntinfo` is directly reassigned in the codebase.

fixes the `FIXME -- make O(1)` comment in `mntinfo_add_list()`.

### DIFF
--- a/criu/mount.c
+++ b/criu/mount.c
@@ -110,19 +110,38 @@ static char *ext_mount_lookup(char *key)
  * Single linked list of mount points get from proc/images
  */
 struct mount_info *mntinfo;
+static struct mount_info *mntinfo_tail;
+
+void mntinfo_set(struct mount_info *new_mntinfo)
+{
+	mntinfo = new_mntinfo;
+	if (mntinfo) {
+		mntinfo_tail = mntinfo;
+		while (mntinfo_tail->next)
+			mntinfo_tail = mntinfo_tail->next;
+	} else {
+		mntinfo_tail = NULL;
+	}
+}
 
 static void mntinfo_add_list(struct mount_info *new)
 {
+	struct mount_info *tail = new;
+
+	/* nothing to append */
+	if (!new)
+		return;
+
+	/* find the actual end of the appended sequence */
+	while (tail->next)
+		tail = tail->next;
+
 	if (!mntinfo)
 		mntinfo = new;
-	else {
-		struct mount_info *pm;
+	else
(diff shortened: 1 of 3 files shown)


criu/mount.c
```diff
@@ -110,19 +110,37 @@ static char *ext_mount_lookup(char *key)
...
+	if (!new)
+		return;
 
-		/* Add to the tail. (FIXME -- make O(1) ) */
-		for (pm = mntinfo; pm->next != NULL; pm = pm->next)
-			;
-		pm->next = new;
+	if (!mntinfo) {
+		mntinfo = new;
+		mntinfo_tail = new;
```
avagin: new isn't just one entry, it is a chain of entries...

3idey: @avagin yeah u right. `new` can be a chain of entries. i had a loop at the end to fix the tail but it looked super confusing.

i refactored it to find the tail of the `new` chain first before adding it so its much cleaner now.

criu/mount.c
```diff
@@ -110,19 +110,38 @@ static char *ext_mount_lookup(char *key)
...
 {
+	struct mount_info *tail = new;
+
+	/* nothing to append */
+	if (!new)
+		return;
+
+	/* find the actual end of the appended sequence */
+	while (tail->next)
+		tail = tail->next;
```
avagin, line 137: so it isn't O(1)...

avagin: honestly, I don't like this approach. We can use a double linked list, it will require to add one extra pointer on mount_info and some dancing with enumerating per-namespace mounts.  Another option is to remove the global list completely and always enumerate all mount namespaces and then all per-namespace mounts... Actually, I don't think this is the problem that we need to solve right now. Usually, there are just a few mounts...

3idey: @avagin fair enough! if there are usually only a few mounts, then optimizing this isn't really worth the extra complexity. 

i can either close this PR or explore one of your suggestions (like removing the global list entirely) if you think it's worth doing eventually. let me know if you want me to drop it for now!

avagin:
Let's close it for now.

PR #3007, opened 2026-04-21
zdtm: Fix rseq01 test for kernel 7.0 rseq changes

### LINKED ISSUE #3005: cr_rseq01 fails in CI on a 7.0.0 kernel
```
2026-04-20T20:26:38.3030013Z === Run 479/487 ===============- zdtm/transition/rseq01
2026-04-20T20:26:38.3030601Z Skipping zdtm/transition/rseq02 (manual run only)
2026-04-20T20:26:38.7605998Z ======================= Run zdtm/transition/rseq01 in h ========================
2026-04-20T20:26:38.7606621Z Start test
2026-04-20T20:26:38.7607008Z ./rseq01 --pidfile=rseq01.pid --outfile=rseq01.out
2026-04-20T20:26:38.7607819Z make: *** [Makefile:52: rseq01.pid] Error 1
2026-04-20T20:26:38.7608330Z  Test zdtm/transition/rseq01 FAIL at ['make', '--no-print-directory', '-C', 'zdtm/transition', 'rseq01.pid'] 
2026-04-20T20:26:38.7608835Z Test output: ================================
2026-04-20T20:26:38.7609245Z 20:26:38.683:    50: FAIL: rseq01.c:106: Failed to register rseq (errno = 22 (Invalid argument))
2026-04-20T20:26:38.7609724Z 20:26:38.683:    49: ERR: test.c:320: Test exited unexpectedly with code 1
```

@mihalicyn

adrianreber:
Starting with Linux 7.0, the rseq feature size has grown to 33 bytes and AT_RSEQ_ALIGN has increased to 64.  This causes sizeof(struct rseq) from the uapi header to be 64 (33 bytes padded to aligned(32)), while the actual registration size used by glibc (__rseq_size) is 33.

The rseq01 test was using sizeof(struct rseq) as the registration size and a test-local __rseq_abi variable (with only 32-byte alignment from the uapi header) as the rseq area.  Both are wrong on kernel 7.0:

 - The kernel now checks alignment against __alignof__(struct rseq) which is 64 internally; the 32-byte-aligned test variable may not satisfy this.

 - sizeof(struct rseq) = 64 does not match the feature size of 33 that glibc registered with, so re-registration with a different size fails.

Fix by:

 - Using __rseq_size (the feature size reported by glibc) as the registration size instead of sizeof(struct rseq).

 - Using the glibc-provided rseq area (at thread_pointer() + __rseq_offset) which is allocated with proper AT_RSEQ_ALIGN alignment, instead of a test-local variable.

Both fall back to the previous behavior when glibc rseq support is not available.

This mirrors the glibc fix:

    https://sourceware.org/cgit/glibc/commit/?id=67f303b47dc584f204e3f2441b9832082415eebc

Fixes: #3005

@mihalicyn

### DIFF
--- a/test/zdtm/transition/rseq01.c
+++ b/test/zdtm/transition/rseq01.c
@@ -20,9 +20,19 @@
 #endif
 #endif
 
+/*
+ * HAVE_GLIBC_RSEQ is set when the C library provides rseq support
+ * (__rseq_size, __rseq_offset).  glibc >= 2.35 defines RSEQ_SIG in
+ * <sys/rseq.h>; musl does not, so the test falls back to its own
+ * definitions below.
+ */
+#if defined(RSEQ_SIG)
+#define HAVE_GLIBC_RSEQ 1
+#endif
+
 #if defined(__x86_64__)
 
-#if defined(__x86_64__) && defined(RSEQ_SIG)
+#ifdef HAVE_GLIBC_RSEQ
 static inline void *thread_pointer(void)
 {
 	void *result;
@@ -38,11 +48,11 @@ static inline void unregister_old_rseq(void)
 		size = 32;
 	syscall(__NR_rseq, (void *)((char *)thread_pointer() + __rseq_offset), size, 1, RSEQ_SIG);
 }
-#else
+#else /* !HAVE_GLIBC_RSEQ */
 static inline void unregister_old_rseq(void)
 {
 }
-#endif
+#endif /* HAVE_GLIBC_RSEQ */
 
 const char *test_doc = "rseq() transition test";
 const char *test_author = "Alexander Mikhalitsyn <alexander.mikhalitsyn@virtuozzo.com>";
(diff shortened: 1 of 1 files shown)


adrianreber:
Alpine needs some additional changes.

test/zdtm/transition/rseq01.c
```diff
@@ -97,11 +107,26 @@ static int sys_rseq(volatile struct rseq *rseq_abi, uint32_t rseq_len, int flags
...
+ * Return the rseq registration size.  Starting with Linux 7.0,
+ * AT_RSEQ_ALIGN is 64 but the feature size is 33, so sizeof(struct rseq)
+ * (padded to alignment) no longer matches the registration size the kernel
+ * expects.  Use __rseq_size when available.
+ */
+static uint32_t rseq_reg_size(void)
+{
+#ifdef HAVE_GLIBC_RSEQ
+	if (__rseq_size)
+		return __rseq_size;
```
mihalicyn, line 120: I think we should do something like this:
```
return (__rseq_size < 32) ? 32 : __rseq_size;
```
like glibc folks here https://github.com/bminor/glibc/blob/04e750e75b73957cf1c791535a3f4319534a52fc/sysdeps/unix/sysv/linux/rseq-internal.h#L109

Otherwise, we break this stuff on older kernels (like here https://github.com/checkpoint-restore/criu/actions/runs/24719996760/job/72310777939?pr=3007 )

mihalicyn (changes requested):
Hi Adrian,

thank you for looking into this stuff! ;-)

test/zdtm/transition/rseq01.c
adrianreber: done, changed as suggested

mihalicyn (approved):
LGTM

Thanks, Adrian!

avagin:
Have you seen https://lore.kernel.org/criu/87bjf4yuym.ffs@tglx/T/#t? Is it related to this issue?

PR #3010, opened 2026-04-23
mem: don't PROT_WRITE on reservation mmaps

bsach64:
Consider the following program:

```
#include <stdio.h>
#include <stdlib.h>
#include <sys/mman.h>
#include <unistd.h>

#define SIZE (64ULL * 1024 * 1024 * 1024)

int main(void)
{
    void *addr = mmap(NULL, SIZE, PROT_NONE,
                      MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);

    if (addr == MAP_FAILED) {
        perror("mmap");
        return EXIT_FAILURE;
    }

    printf("Mapped 64 GB at %p\n", addr);

    while (1) {
        sleep(1);
    }

    return 0;
}
```

while doing a restore using criu-image-streamer we get the following error:
```
64334: Error (criu/mem.c:964): Unable to map ANON_VMA: Cannot allocate memory
Error (criu/cr-restore.c:2369): Restoring FAILED.
Error (criu/cr-restore.c:1268): 64334 killed by signal 9: Killed
Error (criu/cgroup.c:1998): cg: cgroupd: recv req error: No such file or directory
```

full restore log here: https://gist.github.com/bsach64/d0eac6c4db420ddb574c5ad8b98d447c

This PR is a potential fix for the problem, although i am not sure if it is the right approach.

This issue does not occur normal checkpoint restore because we can read the pages multiple times and
verify if there are any pages for a vma and remove `PROT_WRITE` which this commit implements:
https://github.com/checkpoint-restore/criu/commit/c9194500b

### DIFF
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -951,15 +951,25 @@ static int premap_private_vma(struct pstree_item *t, struct vma_area *vma, void 
 		}
 
 		/*
-		 * All mappings here get PROT_WRITE regardless of whether we
-		 * put any data into it or not, because this area will get
-		 * mremap()-ed (branch below) so we MIGHT need to have WRITE
-		 * bits there. Ideally we'd check for the whole COW-chain
-		 * having any data in.
+		 * For VMAs that have PROT_NONE and are not accountable
+		 * (did not have the "ac" flag in /proc/pid/smaps), we
+		 * can safely mmap them with PROT_NONE because we know
+		 * we will never need to write any bits to them.
 		 */
-		addr = mmap(*tgt_addr, size, vma->e->prot | PROT_WRITE, vma->e->flags | MAP_FIXED | flag, vma->e->fd,
-			    vma->e->pgoff);
-
+		if (vma->e->prot == PROT_NONE && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
+			addr = mmap(*tgt_addr, size, PROT_NONE, vma->e->flags | MAP_FIXED | flag, vma->e->fd,
+				    vma->e->pgoff);
+		} else {
+			/*
+			 * All mappings here get PROT_WRITE regardless of whether we
+			 * put any data into it or not, because this area will get
+			 * mremap()-ed (branch below) so we MIGHT need to have WRITE
+			 * bits there. Ideally we'd check for the whole COW-chain
+			 * having any data in.
+			 */
+			addr = mmap(*tgt_addr, size, vma->e->prot | PROT_WRITE, vma->e->flags | MAP_FIXED | flag, vma->e->fd,
+				    vma->e->pgoff);
+		}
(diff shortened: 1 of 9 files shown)


criu/mem.c
```diff
@@ -951,15 +951,25 @@ static int premap_private_vma(struct pstree_item *t, struct vma_area *vma, void
...
-		 * having any data in.
+		 * For VMAs done for reserving virtual address space
+		 * with PROT_NONE which will never be written to,
+		 * we don't want to mmap them with PROT_WRITE because
+		 * they can be arbitrarily large.
 		 */
-		addr = mmap(*tgt_addr, size, vma->e->prot | PROT_WRITE, vma->e->flags | MAP_FIXED | flag, vma->e->fd,
-			    vma->e->pgoff);
-
+		if (vma->e->prot == PROT_NONE && (vma->e->flags & MAP_ANONYMOUS)) {
```
avagin: this doesn't sound right. Here is a test that fails with this change:
```
#include <sys/mman.h>

#include "zdtmtst.h"

#define MEM_SIZE   (4UL * (1UL << 20)) /* 4MB */


{
        char *m1;

        test_init(argc, argv);
        m1 = mmap(NULL, MEM_SIZE, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
        if (m1 == MAP_FAILED) {
                pr_perror("Failed to mmap %lu Mb anonymous shared memory", MEM_SIZE >> 20);
                return 1;
        }
        mmap(NULL, MEM_SIZE, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);


        m1[0] = 5;
        mprotect(m1, MEM_SIZE, PROT_NONE);
        if (fork() == 0) {
                test_waitsig();
                return 0;
        }
        test_daemon();
        test_waitsig();
        mprotect(m1, MEM_SIZE, PROT_READ);

        if (m1[0] != 5) {
                fail("Data mismatch");
                return 1;
        }

        pass();

        return 0;
}
```

avagin: The pure PROT_NONE mapping will not have the ac flag in VmFlags (/proc/pid/smaps). There might be a better way to find out if a mapping is just a reservation.

bsach64: @avagin I have updated the patch based on the above suggestion, we now check for a reservation by checking for `PROT_NONE` and seeing that the vma did not have ac flag set during dump.

avagin:
@bsach64 please introduce a test

bsach64:
@avagin added a test!

bsach64:
@avagin any thoughts?

avagin:
LGTM. Please fix clang-format findings.

bsach64:
@avagin I have fixed the formatting issues!

PR #3014, opened 2026-05-04
plugins/amdgpu: Cleanup env variable handling

tursulin:
This is not saving a lot of lines, just two, but perhaps the pattern of:

```
  kfd_fw_version_check = getenv_bool("KFD_FW_VER_CHECK", true);
```

Is somewhat cleaner than:

```
  /* Default Values */
  kfd_fw_version_check = true;
...
  getenv_bool("KFD_FW_VER_CHECK", &kfd_fw_version_check); 
```

I also make the helpers static and demote parsing errors to warnings since the execution does continue.

I am locally toying with the idea of adding a new env variable to ignore gpu id hence I spotted this cleanup opportunity.

cc @fdavid-amd

### DIFF
--- a/plugins/amdgpu/amdgpu_plugin.c
+++ b/plugins/amdgpu/amdgpu_plugin.c
@@ -308,33 +308,36 @@ int devinfo_to_topology(KfdDeviceEntry *devinfos[], uint32_t num_devices, struct
 	return 0;
 }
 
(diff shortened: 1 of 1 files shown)


avagin:
LGTM

avagin:
@tursulin I have merged this pr by mistake. please review copilot comments.

PR #3021, opened 2026-05-17
restore: parallelize memfd content restore via async fill daemon

dfeigin-nv:
memfd content restore is the slowest part of memfd reconstruction, and nothing else in the restore pipeline depends on it. This offloads just the content fill to an async daemon and keeps memfd creation in the restoring task.

The restoring task does `memfd_create` + `ftruncate`, hands the fd to a `usernsd`-style daemon (`criu/async.c`) over SCM_RIGHTS, and the daemon's worker threads run the mmap + `pread` fill. `start_asyncd()`/`stop_asyncd()` bracket `CR_STATE_RESTORE`. Creating the memfd in the task rather than the coordinator keeps ownership correct in the task's userns.

### DIFF
--- a/criu/asyncd.c
+++ b/criu/asyncd.c
@@ -1,0 +1,211 @@
+#include <unistd.h>
+#include <sys/wait.h>
+#include <sys/stat.h>
+#include <errno.h>
+#include <pthread.h>
+#include <sys/socket.h>
+#include <sys/syscall.h>
+#include <sched.h>
+
+#include "namespaces.h"
+#include "fdstore.h"
+#include "log.h"
+#include "util.h"
+#include "rst_info.h"
+#include "asyncd.h"
+#include "common/compiler.h"
+
+static void *asyncd_thread(void *arg)
+{
+	int sk = (long)arg;
+
+	while (1) {
+		struct unsc_msg um;
+		char msg[MAX_UNSFD_MSG_SIZE];
+		uns_call_t call;
+		int flags, fd, ret;
+		pid_t pid;
+
+		unsc_msg_init_nocreds(&um, &call, &flags, msg, sizeof(msg), 0);
+		ret = recvmsg(sk, &um.h, 0);
+		if (ret == 0)
+			break;
+		if (ret < 0) {
+			pr_perror("async: recv req error");
+			syscall(SYS_exit_group, 1);
+		}
+
+		unsc_msg_pid_fd(&um, &pid, &fd);
+		pr_debug("async: daemon calls %p (%d, %d, %x)\n", call, pid, fd, flags);
+
+		/*
+		 * Caller has sent us bare address of the routine it
+		 * wants to call. Since the caller is fork()-ed from the
+		 * same process as the daemon is, the latter has exactly
+		 * the same code at exactly the same address as the
+		 * former guy has. So go ahead and just call one!
+		 */
+
+		ret = call(msg, fd, pid);
+
+		if (fd >= 0)
+			close(fd);
+
+		if (ret < 0) {
+			pr_err("async: Async call failed. Exiting\n");
+			syscall(SYS_exit_group, 1);
+		}
+	}
+	return (void *)0;
+}
+
+/*
+ * The daemon is forked before the number of restore jobs is known, so it
+ * cannot size its fill pool from the job count. Size it to the number of
+ * available CPUs instead, capped at ASYNC_THREAD_NR_MAX. Content fill is the
+ * bottleneck, so matching the pool to the available cores is a good default.
+ */
+#define ASYNC_THREAD_NR_MAX 16
+static int get_avail_cpus(void)
+{
+	cpu_set_t cpuset;
+
+	CPU_ZERO(&cpuset);
+	if (sched_getaffinity(0, sizeof(cpuset), &cpuset) == 0)
+		return CPU_COUNT(&cpuset);
+
+	return sysconf(_SC_NPROCESSORS_ONLN);
+}
+
+static int asyncd(int sk)
+{
+	pthread_t threads[ASYNC_THREAD_NR_MAX];
+	int nr_threads = min_t(int, get_avail_cpus(), ASYNC_THREAD_NR_MAX);
+	int i;
+
+	if (nr_threads < 1)
+		nr_threads = 1;
+
+	for (i = 0; i < nr_threads; i++) {
+		if (pthread_create(&threads[i], NULL, asyncd_thread, (void *)(long)sk)) {
+			pr_perror("async: pthread_create");
+			return 1;
+		}
+	}
+
+	pr_info("async: Daemon started with %d threads\n", nr_threads);
+	for (i = 0; i < nr_threads; i++) {
+		if (pthread_join(threads[i], NULL)) {
+			pr_perror("async: pthread_join");
+			return 1;
+		}
+	}
+
+	return 0;
+}
+
+int __async_call(const char *func_name, uns_call_t call, int flags, void *arg, size_t arg_size, int fd)
+{
+	int ret, sk = -1;
+	struct unsc_msg um;
+
+	if (unlikely(arg_size > MAX_UNSFD_MSG_SIZE)) {
+		pr_err("async: message size exceeded\n");
+		return -1;
+	}
+
+	sk = fdstore_get(task_entries->asyncd_sk_id);
+	if (sk < 0) {
+		pr_err("async: cannot get ASYNCD_SK fd\n");
+		return -1;
+	}
+	pr_debug("async: calling %s (%d, %x)\n", func_name, fd, flags);
+
+	unsc_msg_init_nocreds(&um, &call, &flags, arg, arg_size, fd);
+	ret = sendmsg(sk, &um.h, 0);
+	if (ret <= 0) {
+		pr_perror("async: send req error");
+		ret = -1;
+		goto out;
+	}
+
+	ret = 0;
+out:
+	close_safe(&sk);
+	return ret;
+}
+
+static int asyncd_pid;
+int start_asyncd(void)
+{
+	int sk, id;
+
+	sk = start_unix_cred_daemon(&asyncd_pid, asyncd, false);
+	if (sk < 0) {
+		pr_err("failed to start asyncd\n");
+		return -1;
+	}
+
+	id = fdstore_add(sk);
+	close(sk);
+	if (id < 0) {
+		if (waitpid(asyncd_pid, NULL, 0) < 0)
+			pr_perror("async: waitpid");
+		asyncd_pid = 0;
+
+		return -1;
+	}
+
+	task_entries->asyncd_sk_id = id;
+
+	return 0;
+}
+
+int stop_asyncd(void)
+{
+	int exit_code = -1;
+	int sk = -1, status = -1;
+	sigset_t blockmask, oldmask;
+
+	/* No daemon was started (e.g. stream restore, or already stopped). */
+	if (asyncd_pid == 0)
+		return 0;
+
+	/*
+	 * Don't let the sigchld_handler() mess with us
+	 * calling waitpid() on the exited daemon. The
+	 * same is done in cr_system().
+	 */
+
+	sigemptyset(&blockmask);
+	sigaddset(&blockmask, SIGCHLD);
+	sigprocmask(SIG_BLOCK, &blockmask, &oldmask);
+
+	sk = fdstore_get(task_entries->asyncd_sk_id);
+	if (sk < 0) {
+		pr_err("async: cannot get ASYNCD_SK fd\n");
+		goto out;
+	}
+	if (shutdown(sk, SHUT_WR)) {
+		pr_perror("async: shutdown");
+		goto out;
+	}
+	if (waitpid(asyncd_pid, &status, 0) < 0) {
+		pr_perror("async: waitpid");
+		goto out;
+	}
+
+	if (!WIFEXITED(status) || WEXITSTATUS(status) != 0) {
+		pr_err("async: daemon exited abnormally (status %#x)\n", status);
+		goto out;
+	}
+
+	pr_info("async: daemon stopped\n");
+	exit_code = 0;
+out:
+	close_safe(&sk);
+	asyncd_pid = 0;
+	sigprocmask(SIG_SETMASK, &oldmask, NULL);
+
+	return exit_code;
+}
--- a/criu/bfd.c
+++ b/criu/bfd.c
@@ -6,6 +6,7 @@
 #include <fcntl.h>
 #include <sys/uio.h>
 #include <errno.h>
+#include <pthread.h>
 
 #include "int.h"
 #include "log.h"
@@ -33,6 +34,7 @@ struct bfd_buf {
 };
 
 static LIST_HEAD(bufs);
+static pthread_mutex_t bufs_lock = PTHREAD_MUTEX_INITIALIZER;
 
 #define BUFBATCH (16)
 
@@ -40,12 +42,15 @@ static int buf_get(struct xbuf *xb)
 {
 	struct bfd_buf *b;
 
+	pthread_mutex_lock(&bufs_lock);
+
 	if (list_empty(&bufs)) {
 		void *mem;
 		int i;
 
 		mem = mmap(NULL, BUFBATCH * BUFSIZE, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, 0, 0);
 		if (mem == MAP_FAILED) {
+			pthread_mutex_unlock(&bufs_lock);
 			pr_perror("No buf");
 			return -1;
 		}
@@ -54,6 +59,7 @@ static int buf_get(struct xbuf *xb)
 			b = xmalloc(sizeof(*b));
 			if (!b) {
 				if (i == 0) {
+					pthread_mutex_unlock(&bufs_lock);
 					pr_err("No buffer for bfd\n");
 					return -1;
 				}
@@ -75,6 +81,7 @@ static int buf_get(struct xbuf *xb)
 	xb->bsize = BUFSIZE;
 	xb->sz = 0;
 	xb->buf = b;
+	pthread_mutex_unlock(&bufs_lock);
 	return 0;
 }
 
@@ -85,7 +92,9 @@ static void buf_put(struct xbuf *xb)
 		 * Don't unmap standard buffer back, it will get reused
 		 * by next bfdopen call
 		 */
+		pthread_mutex_lock(&bufs_lock);
 		list_add(&xb->buf->l, &bufs);
+		pthread_mutex_unlock(&bufs_lock);
 	} else {
 		/* This buffer was dynamically extended, unmap it */
 		munmap(xb->mem, xb->bsize);
@@ -199,7 +208,9 @@ static int bextend(struct bfd *f)
 			return -1;
 		}
 		memcpy(newbuf, b->mem, b->sz);
+		pthread_mutex_lock(&bufs_lock);
 		list_add(&b->buf->l, &bufs);
+		pthread_mutex_unlock(&bufs_lock);
 		b->buf = NULL;
 	} else {
 		newbuf = mremap(b->mem, b->bsize, newsize, MREMAP_MAYMOVE);
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -49,6 +49,7 @@
 #include "crtools.h"
 #include "uffd.h"
 #include "namespaces.h"
+#include "asyncd.h"
 #include "mem.h"
 #include "mount.h"
 #include "fsnotify.h"
@@ -143,6 +144,8 @@ static inline int stage_participants(int next_stage)
 		return 1;
 	case CR_STATE_FORKING:
 		return task_entries->nr_tasks + task_entries->nr_helpers;
+	case CR_STATE_PRE_RESTORER:
+		return task_entries->nr_tasks + task_entries->nr_helpers;
 	case CR_STATE_RESTORE:
 		return task_entries->nr_threads + task_entries->nr_helpers;
 	case CR_STATE_RESTORE_SIGCHLD:
@@ -158,16 +161,8 @@ static inline int stage_current_participants(int next_stage)
 {
 	switch (next_stage) {
 	case CR_STATE_FORKING:
+	case CR_STATE_PRE_RESTORER:
 		return 1;
-	case CR_STATE_RESTORE:
-		/*
-		 * Each thread has to be reported about this stage,
-		 * so if we want to wait all other tasks, we have to
-		 * exclude all threads of the current process.
-		 * It is supposed that we will wait other tasks,
-		 * before creating threads of the current task.
-		 */
-		return current->nr_threads;
 	}
 
 	BUG();
@@ -801,6 +796,7 @@ static int restore_one_zombie(CoreEntry *core)
 
 	prctl(PR_SET_NAME, (long)(void *)core->tc->comm, 0, 0, 0);
 
+	restore_finish_stage(task_entries, CR_STATE_PRE_RESTORER);
 	if (task_entries != NULL) {
 		wait_exiting_children();
 		zombie_prepare_signals();
@@ -978,6 +974,7 @@ static int restore_one_helper(void)
 	if (prepare_fds(current))
 		return -1;
 
+	restore_finish_stage(task_entries, CR_STATE_PRE_RESTORER);
 	if (wait_exiting_children())
 		return -1;
 
@@ -1670,8 +1667,6 @@ static int __restore_task_with_children(void *_arg)
 	if (populate_pid_proc())
 		goto err;
 
-	sfds_protected = true;
-
 	if (unmap_guard_pages(current))
 		goto err;
 
@@ -1694,11 +1689,18 @@ static int __restore_task_with_children(void *_arg)
 		if (restore_wait_other_tasks())
 			goto err;
 		fini_restore_mntns();
-		__restore_switch_stage(CR_STATE_RESTORE);
+
+		/* streamer serves each image once, one at a time; asyncd reads in parallel. */
+		if (!opts.stream && start_asyncd())
+			goto err;
+
+		__restore_switch_stage(CR_STATE_PRE_RESTORER);
 	} else {
 		if (restore_finish_stage(task_entries, CR_STATE_FORKING) < 0)
 			goto err;
 	}
+
+	sfds_protected = true;
 
 	if (restore_one_task(vpid(current), ca->core))
 		goto err;
@@ -1706,6 +1708,13 @@ static int __restore_task_with_children(void *_arg)
 	return 0;
 
 err:
+	/*
+	 * Reap the async daemon before waking the coordinator: once the
+	 * abort is signalled the coordinator tears down the task tree and
+	 * may kill us mid-reap. stop_asyncd() is a no-op if no daemon was
+	 * started (asyncd_pid == 0).
+	 */
+	stop_asyncd();
 	if (current->parent == NULL)
 		futex_abort_and_wake(&task_entries->nr_in_progress);
 	exit(1);
@@ -2184,7 +2193,7 @@ skip_ns_bouncing:
 		goto out_kill;
 
 	/*
-	 * Zombies die after CR_STATE_RESTORE which is switched
+	 * Zombies die after CR_STATE_PRE_RESTORER which is switched
 	 * by root task, not by us. See comment before CR_STATE_FORKING
 	 * in the header for details.
 	 */
@@ -3241,6 +3250,12 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 			goto err_nv;
 		if (root_ns_mask & CLONE_NEWNS && remount_readonly_mounts())
 			goto err_nv;
+		if (stop_asyncd())
+			goto err_nv;
+		__restore_switch_stage(CR_STATE_RESTORE);
+	} else {
+		if (restore_finish_stage(task_entries, CR_STATE_PRE_RESTORER) < 0)
+			goto err;
 	}
 
 	/*
@@ -3603,6 +3618,8 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 err:
 	free_mappings(&self_vmas);
 err_nv:
+	/* Reap the async daemon if it is still running (no-op otherwise). */
+	stop_asyncd();
 	/* Just to be sure */
 	exit(1);
 	return -1;
--- a/criu/namespaces.c
+++ b/criu/namespaces.c
@@ -1221,7 +1221,8 @@ static int write_id_map(pid_t pid, UidGidExtent **extents, int n, char *id_map)
 
 static int usernsd_pid;
 
-inline void unsc_msg_init(struct unsc_msg *m, uns_call_t *c, int *x, void *arg, size_t asize, int fd, pid_t *pid)
+static void unsc_msg_init_flags(struct unsc_msg *m, uns_call_t *c, int *x, void *arg, size_t asize, int fd, pid_t *pid,
+				bool send_creds)
 {
 	struct cmsghdr *ch;
 	struct ucred *ucred;
@@ -1251,24 +1252,31 @@ inline void unsc_msg_init(struct unsc_msg *m, uns_call_t *c, int *x, void *arg, 
 	 */
 	memzero(&m->c, sizeof(m->c));
 
-	m->h.msg_controllen = CMSG_SPACE(sizeof(struct ucred));
+	m->h.msg_controllen = 0;
+	ch = NULL;
 
-	ch = CMSG_FIRSTHDR(&m->h);
-	ch->cmsg_len = CMSG_LEN(sizeof(struct ucred));
-	ch->cmsg_level = SOL_SOCKET;
-	ch->cmsg_type = SCM_CREDENTIALS;
+	if (send_creds) {
+		m->h.msg_controllen = CMSG_SPACE(sizeof(struct ucred));
+		ch = CMSG_FIRSTHDR(&m->h);
+		ch->cmsg_len = CMSG_LEN(sizeof(struct ucred));
+		ch->cmsg_level = SOL_SOCKET;
+		ch->cmsg_type = SCM_CREDENTIALS;
 
-	ucred = (struct ucred *)CMSG_DATA(ch);
-	if (pid)
-		ucred->pid = *pid;
-	else
-		ucred->pid = getpid();
-	ucred->uid = getuid();
-	ucred->gid = getgid();
+		ucred = (struct ucred *)CMSG_DATA(ch);
+		if (pid)
+			ucred->pid = *pid;
+		else
+			ucred->pid = getpid();
+		ucred->uid = getuid();
+		ucred->gid = getgid();
+	}
 
 	if (fd >= 0) {
 		m->h.msg_controllen += CMSG_SPACE(sizeof(int));
-		ch = CMSG_NXTHDR(&m->h, ch);
+		if (ch)
+			ch = CMSG_NXTHDR(&m->h, ch);
+		else
+			ch = CMSG_FIRSTHDR(&m->h);
 		BUG_ON(!ch);
 		ch->cmsg_len = CMSG_LEN(sizeof(int));
 		ch->cmsg_level = SOL_SOCKET;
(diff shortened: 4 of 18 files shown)


adrianreber:
Some better PR title would be nice. "Upstream parallel memfd" is not very descriptive.

So far the only user of pthread seems to be the AMD plugin. I see there and here that the logging is not protected against writes at the same time from multiple workers. Have you seen any garbled log output during your tests?

criu/mem.c
```diff
@@ -1458,38 +1450,377 @@ int unmap_guard_pages(struct pstree_item *t)
...
+			continue;
+
+		if (u->is_memfd) {
+			if (!inherited_fd(u->vmfd, &fd))
+				fd = memfd_open(u->vmfd, &u->flags, true);
+		} else
+			fd = open_file_for_vma(u->rep_vma, u->flags);
+
+		if (fd < 0) {
+			for (int j = 0; j < nr_unique; j++)
```
adrianreber: This seems to be the only place where the loop variable is not defined outside of the loop. Just mentioning because of consistency.  Not really important.

dfeigin-nv:
> Some better PR title would be nice. "Upstream parallel memfd" is not very descriptive.
> 
> So far the only user of pthread seems to be the AMD plugin. I see there and here that the logging is not protected against writes at the same time from multiple workers. Have you seen any garbled log output during your tests?

Updated title. I haven't ran into garbled logs, but I will check for consistency.

criu/mem.c
```diff
@@ -1458,38 +1450,377 @@ int unmap_guard_pages(struct pstree_item *t)
...
+	nr_jobs = memfd_open_jobs_collect(unique, nr_unique, &jobs);
+	if (nr_jobs < 0)
+		return -1;
+	if (nr_jobs <= 1) {
+		int ret = 0;
+
+		if (nr_jobs == 1) {
+			int fd;
+
+			if (!inherited_fd(jobs[0]->vmfd, &fd))
```
adrianreber: Any specific reason you are checking with `inherited_fd()` again? `jobs` with `inherited_fd()` returning `true` are already filtered out earlier in `memfd_open_jobs_collect()`, right? I am a bit confused here with the threading. Maybe some comments would be helpful to make it a bit more clear.

adrianreber: `inherited_fd()` should always return false at this point, so it will never change `fd` and could be removed.

adrianreber:
Do you have any performance numbers you can share? The PR kind of suggests that this might be faster, but I have not seen any numbers in the PR description at least.

dfeigin-nv:
> > Some better PR title would be nice. "Upstream parallel memfd" is not very descriptive.
> > So far the only user of pthread seems to be the AMD plugin. I see there and here that the logging is not protected against writes at the same time from multiple workers. Have you seen any garbled log output during your tests?
> 
> Updated title. I haven't ran into garbled logs, but I will check for consistency.

Good catch @adrianreber  - there is a little garbling, found 70 lines in a 49k-line restore log which are a off.

It's the `static buffer[]` in `criu/log.c::vprint_on_level()` - two threads `vsnprintf `into the same staging area before the `write()`. AMD plugin has the same race.

The PIE side (`compel/plugins/std/log.c)` already builds the line in a stack-local `simple_buf `per call. Mirroring that on the daemon side drops the statics and fixes both. WDYT? I can put it in as a separate commit in this PR.

adrianreber:
> Good catch @adrianreber - there is a little garbling, found 70 lines in a 49k-line restore log which are a off.

Thanks for checking.

> The PIE side (compel/plugins/std/log.c) already builds the line in a stack-local simple_buf per call. Mirroring that on the daemon side drops the statics and fixes both. WDYT? I can put it in as a separate commit in this PR.

Not familiar with that code, so I cannot really say. Maybe make it a separate PR which makes it easier to review. Let's see if anybody else chimes in.

Besides that, there are always a couple CI tests that are flaky, but this one looks a bit strange (especially as you are changing how memory pages are written):


```
################### 1 TEST(S) FAILED (TOTAL 487/SKIPPED 55) ####################
 * zdtm/transition/maps007(uns)
```

This is the relevant log: https://github.com/checkpoint-restore/criu/actions/runs/25984069415/job/76406749374?pr=3021

dfeigin-nv:
@adrianreber I edited the description to be clearer, and include some benchmarks for vLLM loaded with different models

dfeigin-nv:
Hi Adrian,
Is there a way to retrigger failed CI jobs?

adrianreber:
> Hi Adrian, Is there a way to retrigger failed CI jobs?

For you probably close and re-open the PR.

adrianreber:
FYI: The CI failures do not seem related to this change: nftables-test:
```
################### 4 TEST(S) FAILED (TOTAL 487/SKIPPED 37) ####################
 * zdtm/static/socket-tcpbuf(h)
 * zdtm/static/socket-tcpbuf-local(h)
 * zdtm/static/socket-tcpbuf6(h)
 * zdtm/static/socket-tcpbuf6-local(uns)
```
Also the other failed test is and I have also seen this before:
```
################### 1 TEST(S) FAILED (TOTAL 489/SKIPPED 47) ####################
 * zdtm/transition/rseq01(h)
```

criu/mem.c
dfeigin-nv: Still fixed.

criu/mem.c
dfeigin-nv: added a bit more comments. and yes you're right some of those checks aren't really adding anything, so I removed them

dfeigin-nv:
@adrianreber I pushed an amend commit that addresses your comments. The safe-logging fix I'll do as a separate PR as discussed.
CI failures look like dump-side flakes. WDYT?

dfeigin-nv:
@adrianreber Could you please take another look when you have a moment?

criu/mem.c
```diff
@@ -1475,38 +1477,380 @@ int unmap_guard_pages(struct pstree_item *t)
...
 
+	/* Phase 1: non-filemap VMAs (shmem, socket) + plugin; collect filemap + memfd for parallel open */
 	list_for_each_entry(vma, &vmas->h, list) {
 		if (!vma_area_is(vma, VMA_AREA_REGULAR) || !vma->vm_open)
 			continue;
 
 		pr_info("Opening %#016" PRIx64 "-%#016" PRIx64 " %#016" PRIx64 " (%x) vma\n", vma->e->start,
 			vma->e->end, vma->e->pgoff, vma->e->status);
 
-		if (vma->vm_open(pid, vma)) {
```
avagin, line 1502: memfds are files, so parallel restore should happen at the file level. While reviewing this PR, I realized that the slowest part of the process is restoring the memfd data. Because other restore stages don't depend on it, we can do this asynchronously. The memfd restore sequence could roughly look like: memfd_create, ftruncate, add_async_work(memfd_restore_data, ...). The async work engine could be generic, as there are likely several other places we can optimize in a similar way.

oOraph:
Both https://github.com/checkpoint-restore/criu/pull/3021 and https://github.com/checkpoint-restore/criu/pull/3022 optims seem quite promising :+1: @dfeigin-nv !

I've read this interesting related blog page here
https://developer.nvidia.com/blog/nvidia-dynamo-snapshot-fast-startup-for-inference-workloads-on-kubernetes
and it mentions both this pr and 
https://github.com/checkpoint-restore/criu/pull/3022 if I'm not mistaken

I need some insight though, as I'm having a hard time observing the mentionned perf improvements. Do you have more details about the hardware you ran your tests on ?

some more details about the benchmarks I ran so far
https://github.com/oOraph/bench-criu/tree/main/results

(1 really basic, not real use case, just a random tensor dump/restore, and another one closer to the one you did it seems, with real models for inference + vllm or not depending on the model),

I'm pretty sure I'm missing sth and my bottleneck is not where I expect it to be, but in the meantime I'm also confident I'm far from saturating the raid throughput on my test machine

criu/mem.c
dfeigin-nv: That's a good design. One caveat though - if the threads use locks (or malloc locks, and they do) and a fork occurs while lock, the forked child may get deadlocked if tries to allocate something. If we move the the launch of thread to after the last fork of the master process, there is no real async gain from it (no much happens after the thread launch, we immediately start draining). However the design is better and other files may benefit from it, so I'll finish updating the code according to this design and push

dfeigin-nv: Hi @avagin 
I pushed the code that implements this redesign. Please take a loop when you have the time :)

dfeigin-nv:
Hi @oOraph , 
Thank you for your interest in this work.

Several questions:
1) Are you using sleep/wake-up mechanisms in you benchmarks? Can you please run you benchmarks on a sleeping vllm too?
2) Can you please provide restore.log (-v4 -o restore.log)?
3) On your checkpoints - can you please run `crit show memfd.img`?

criu/mem.c
dfeigin-nv: After running some tests, I think the coordinator approach may be somewhat problematic.

The file-level/async idea itself is fine — the problem is what it forces: to share the memfd across tasks, the coordinator has to create it. And a memfd's owner is fixed at creation, by whoever creates it, so a coordinator-created memfd is owned by the coordinator's uid. On a restore into a user namespace the original owner doesn't map there, so when we restore the memfd's ownership (`cr_fchpermat`/`cr_fchperm`) it can't be reconciled and fails — CI's `vdso-proxy` catches it. So what's at odds with userns ownership is the coordinator-side creation, not the file-level/async approach itself.

It can be solved (e.g. creating the memfd in the task and sending the fd to the coordinator), but I don't want to add another layer of complexity to this PR. To make CI pass, I reverted to the original design. I think the redesign is better left to a follow-up PR. Please let me know what you think.

oOraph:
> 
> Several questions:
> 
>     1. Are you using sleep/wake-up mechanisms in you benchmarks? Can you please run you benchmarks on a sleeping vllm too?
> 
>     2. Can you please provide restore.log (-v4 -o restore.log)?
> 
>     3. On your checkpoints - can you please run `crit show memfd.img`?

Hi @dfeigin-nv, thanks for your answer!

## 1. Sleep/wake-up benchmarks

Initially our benchmarks did not include any app-side cooperative sleep prior to checkpointing (no `/sleep?level=1` call before the CRIU dump, because we did not only benchmark vLLM). Following your suggestion, we ran a cooperative sleep benchmark with vllm only. Here are the results for Qwen/Qwen3-8B, including both the standalone restore and the full sleep → checkpoint → restore → wake_up cycle:

**Standalone restore (no sleep call before checkpoint)**

| variant | dump (s) | restore (s) |
|---------|----------|-------------|
| baseline | 19.3 | 14.4 |
| upstream-optimized | 19.4 | 14.5 |

**Sleep → checkpoint → restore → wake_up**

| variant | sleep (s) | dump (s) | restore (s) | wake_up (s) | restore + wake_up (s) |
|---------|-----------|----------|-------------|-------------|----------------------|
| baseline | 12.8 | 16.1 | 22.6 | ~2.0 | **~24.5** |
| upstream-optimized | 12.8 | 16.0 | 15.8 | ~1.7 | **~17.8** |

From what we can reproduce so far, PRs #3021 and #3022 provide a significant speedup **in the sleep case** — roughly -27% on restore time (22.6s → 15.8s). In the standalone/non cooperative restore scenario both variants converge to ~14.4–14.5s, faster than the cooperative checkpoint/restore, and the PRs make no measurable difference, unless we're missing something in our benchmark :).

Branch references:
- **baseline**: criu-dev, commit 4d76d1acdcd69afcedf1335c810bef5f8e18d95e
- **upstream-optimized**: https://github.com/oOraph/criu/tree/optim1 -> criu-dev, commit 5141fbcca85e3fd85371ac9adb76872f78b0b967 + this pr

---

## 2. restore.log

restore.log (`-v4`) for a sleep checkpoint of Qwen/Qwen3-8B (restore time ~18.9s):
https://github.com/oOraph/bench-criu/blob/main/results/vllm_sleep_awake/checkpoint_example/work/restore.log

---

## 3. `crit show memfd.img`

Full output (same checkpoint, Qwen/Qwen3-8B, after sleep):
https://github.com/oOraph/bench-criu/blob/main/results/vllm_sleep_awake/checkpoint_example/memfd_show.out

Summary: 156 entries, all named `/dev/zero`, ~20.94 GB total.

Size distribution:
```
  2.0 MB:    7 entries
  2.7 MB:    1 entry
  16.0 MB:   1 entry
  32.0 MB:  37 entries
  64.0 MB:  36 entries
  128.0 MB: 36 entries
  256.0 MB: 36 entries
  2048.0 MB: 2 entries
```

dfeigin-nv:
@oOraph Thank you for the response. If that's alright with you, I'd like to suggest to take this offline, to make messaging easier and more streamlined than github. Can we please continue the conversation via email (dfeigin@nvidia.com)? If you're in Kubernetes Slack we can message there too.

criu/mem.c
avagin: @dfeigin-nv It isn't about adding another layer of complexity. As I mentioned, the 'coordinator' don't need to create the memfd, since creation is fast. The heaviest operation is restoring the memfd data, and this part can be done asynchronously. Please take a look at this poc: https://github.com/avagin/criu/commits/async-memfd/. It is a quick draft version, so probably it can be optimized here and there.

dfeigin-nv: Thanks @avagin - I adopted your async-memfd POC here. I added another commit with a few adjustments on top of it.

rst0git: @dfeigin-nv Would it be possible to edit Andrei's patch above to add a commit message?
https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes

> I added another commit with a few adjustments on top of it.

If it makes sense, you can combine both commits into one and add `Co-authored-by:`?

We generally don't use `cherry picked from commit ...`

dfeigin-nv: Done.

dfeigin-nv:
@rst0git @avagin I think the failing tests are flakes (or suites that run flakes). Am I missing something?

adrianreber:
> @rst0git @avagin I think the failing tests are flakes (or suites that run flakes). Am I missing something?

I had a look at the CI failuers and I think they are indeed unrelated to your change:

On " aarch64-test (ubuntu-24.04-arm, CLANG=1)", "aarch64-test (ubuntu-24.04-arm, GCC=1)", "Arch Linux Test (zdtm 2/4)", "gcov-test", "nftables-test" we see some `socket-tcpbuf*` failures which happen much too often unfortunately, but they are not yet understood. Seems to depend on the kernel of the GitHub Actions runners.

There is something new on "CentOS Stream 9", "CentOS Stream 10", "fedora-asan-test", "VM Fedora No VDSO based test", "VM Fedora Stable based test", "x86_64 Fedora Rawhide":
```
################### 1 TEST(S) FAILED (TOTAL 490/SKIPPED 48) ####################
 * zdtm/static/futex(uns)
```
Never seen that before. Something with restartable sequences and user namespaces. It is all on CentOS and Fedora systems, which is interesting. Especially as it seems to happen on Fedora and CentOS at the same time. Not sure if that is related to your change or not.

And then there is the failure on "VM Fedora Next based test":
```
The kernel is tainted: '512'
  File "/criu/./test/zdtm.py", line 2970, in <module>
    opts['action'](opts)
    ~~~~~~~~~~~~~~^^^^^^
  File "/criu/./test/zdtm.py", line 2504, in run_tests
    launcher = Launcher(opts, len(torun))
  File "/criu/./test/zdtm.py", line 2121, in __init__
    self.__report_kernel_taint("The kernel is tainted: %r" % self.__taint)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/criu/./test/zdtm.py", line 2132, in __report_kernel_taint
    raise Exception(msg)
Exception: The kernel is tainted: '512'
```

Which is tracked upstream at https://lore.kernel.org/all/20260601-btrfs-fix-wq-warning-qgroup-rescan-v1-1-aff9a1128f27@kernel.org/

Not sure about the state of the futex test error. Haven't seen that before.

dfeigin-nv:
@adrianreber the futex error was my bad, not a flake. The new daemon shares the restored pidns and was stopped only after restore, so its worker threads collided with the restored threads on their tids. Fixed by forking the daemon from the coordinator (host pidns) before the task tree, like usernsd. I pushed and closed and reopened the PR for a new CI run.

avagin:
> @adrianreber the futex error was my bad, not a flake. The new daemon shares the restored pidns and was stopped only after restore, so its worker threads collided with the restored threads on their tids. Fixed by forking the daemon from the coordinator (host pidns) before the task tree, like usernsd. I pushed and closed and reopened the PR for a new CI run.

I intentionally forked them from the root task. The main criu process is outside of restored cgroups and a restored shared memory will be charged to a wrong cgroup.

dfeigin-nv:
> I intentionally forked them from the root task. The main criu process is outside of restored cgroups and a restored shared memory will be charged to a wrong cgroup

@avagin  I see. It collided with the restored TIDs. 
I can think of other options, not sure what's the best approach:
- Writing the daemon PID into the restored cgroup, but that needs to happen after the initial fork I think.
- Reserving very high TIDs (with clone3). It isn't readily available in pthread_create, so I don't think that's a good option.

Let me know what you think.

avagin:
@dfeigin-nv The idea was to stop the asyncd process before restoring the threads. My previous draft was incomplete. I think it should look something like this: https://github.com/avagin/criu/commits/wip/async

dfeigin-nv:
@avagin Thank you. I thought about adding a new stage but I thought it may be too surgical for that purpose. I incorporated that change, so now everything is spawned from the root task, and the daemon is stopped before we need the TIDs. Restarted CI.

avagin:
> @avagin Thank you. I thought about adding a new stage but I thought it may be too surgical for that purpose. I incorporated that change, so now everything is spawned from the root task, and the daemon is stopped before we need the TIDs. Restarted CI.

Adding a new stage isn't ideal, I wanted to avoid that, but could not find any valuable solution yet.

dfeigin-nv:
I think that the failures are in flaky tests/dump side/weird linux kernel runner hiccups. WDYT? @avagin @adrianreber

adrianreber:
One of the errors looks like it is related to your change. Running as non-root seems to break.
```
unshare --map-auto -fpm --mount-proc bash -c './test/zdtm.py run -t zdtm/static/maps00 -f h --rootless && true'
userns is supported
=== Run 1/1 ================ zdtm/static/maps00
Start test
Run criu dump
=[log]=> dump/zdtm/static/maps00/84/1/dump.log
b'(00.000000) CRIU run id = d3480aee-8350-4971-a50e-f0a35bcacf91'
b'(00.000077) Version: 4.2 (gitid bdc8b85)'
b'(00.000092) Running on lima-default Linux 6.19.10-300.fc44.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 25 18:23:49 UTC 2026 x86_64'
b'(00.000098) Warn  (criu/kerndat.c:1195): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.012923) cpu: fpu: xfeatures_mask 0x5 xsave_size 832 xsave_size_max 832 xsaves_size 832'
b'(00.012986) cpu: fpu: x87 floating point registers     xstate_offsets      0 / 0      xstate_sizes    160 / 160'
b'(00.012991) cpu: fpu: AVX registers                    xstate_offsets    576 / 576    xstate_sizes    256 / 256'
b'(00.013238) Effective capability 12 missing'
b'(00.014089) Error (criu/util.c:1620): Unable to open the binfmt_misc file system: Operation not permitted'
b'(00.014101) binfmt_misc sandboxing is not supported'
b'(00.014180) Found mmap_min_addr 0x10000'
b'(00.014193) files stat: fs/nr_open 2147483584'
b'(00.014198) Warn  (criu/kerndat.c:1195): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
Run criu restore
=[log]=> dump/zdtm/static/maps00/84/1/restore.log
b'(00.000000) CRIU run id = c6136b8d-554c-43f2-bdbd-43e8d94eceee'
b'(00.000071) Version: 4.2 (gitid bdc8b85)'
b'(00.000085) Running on lima-default Linux 6.19.10-300.fc44.x86_64 #1 SMP PREEMPT_DYNAMIC Wed Mar 25 18:23:49 UTC 2026 x86_64'
b'(00.000091) Warn  (criu/kerndat.c:1195): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.013619) cpu: fpu: xfeatures_mask 0x5 xsave_size 832 xsave_size_max 832 xsaves_size 832'
b'(00.013682) cpu: fpu: x87 floating point registers     xstate_offsets      0 / 0      xstate_sizes    160 / 160'
b'(00.013687) cpu: fpu: AVX registers                    xstate_offsets    576 / 576    xstate_sizes    256 / 256'
b'(00.013972) Effective capability 12 missing'
b'(00.014850) Error (criu/util.c:1620): Unable to open the binfmt_misc file system: Operation not permitted'
b'(00.014861) binfmt_misc sandboxing is not supported'
b'(00.014989) Found mmap_min_addr 0x10000'
b'(00.015004) files stat: fs/nr_open 2147483584'
b'(00.015010) Warn  (criu/kerndat.c:1195): $XDG_RUNTIME_DIR not set. Cannot find location for kerndat file'
b'(00.019488)     84: Opening 0x0000000040a000-0x0000000040b000 0x00000000009000 (40041) vma'
b'(00.019494)     84: Opening 0x007fddb0ace000-0x007fddb0ad0000 0000000000000000 (40101) vma'
b'(00.019500)     84: Search for 0x007fddb0ace000 shmem 0x2f 0x7f704df07c98/84'
b'(00.019540)     84: async: calling async_restore_shmem_content (13, 0)'
b'(00.019568)     84: Error (criu/asyncd.c:114): async: send req error: Operation not permitted'
b"(00.019577)     84: Error (criu/shmem.c:677): Can't offload shmem restore"
b"(00.019592)     84: Error (criu/mem.c:1503): `- Can't open vma"
b'(00.019595)     84: async: Daemon started with 4 threads'
b'(00.019711) Error (criu/cr-restore.c:2370): Restoring FAILED.'
b'(00.020053) Error (criu/cr-restore.c:1265): 84 killed by signal 9: Killed'
################# Test zdtm/static/maps00 FAIL at CRIU restore #################
14:24:55.739:    84: map: ptr 0x7fddb0ceb000 flag        2 prot        0
14:24:55.740:    84: map: ptr 0x7fddb0ce9000 flag        1 prot        0
14:24:55.740:    84: map: ptr 0x7fddb0ae8000 flag       22 prot        0
14:24:55.740:    84: map: ptr 0x7fddb0ae6000 flag       21 prot        0
14:24:55.740:    84: map: ptr 0x7fddb0ae4000 flag        2 prot        1
14:24:55.740:    84: map: ptr 0x7fddb0ae2000 flag        1 prot        1
14:24:55.740:    84: map: ptr 0x7fddb0ae0000 flag       22 prot        1
14:24:55.740:    84: map: ptr 0x7fddb0ade000 flag       21 prot        1
14:24:55.740:    84: map: ptr 0x7fddb0adc000 flag        2 prot        3
14:24:55.741:    84: map: ptr 0x7fddb0ada000 flag        1 prot        3
14:24:55.741:    84: map: ptr 0x7fddb0ad8000 flag       22 prot        3
14:24:55.741:    84: map: ptr 0x7fddb0ad6000 flag       21 prot        3
14:24:55.741:    84: map: ptr 0x7fddb0ad4000 flag        2 prot        7
14:24:55.741:    84: map: ptr 0x7fddb0ad2000 flag        1 prot        7
14:24:55.741:    84: map: ptr 0x7fddb0ad0000 flag       22 prot        7
14:24:55.742:    84: map: ptr 0x7fddb0ace000 flag       21 prot        7

```

All the other errors seems to be well known.

avagin:
> b'(00.019568)     84: Error (criu/asyncd.c:114): async: send req error: Operation not permitted'

https://github.com/avagin/criu/commit/7c01780c89608999db04e624533ed92ca07a9325 should fix this issue.

dfeigin-nv:
> > b'(00.019568) 84: Error (criu/asyncd.c:114): async: send req error: Operation not permitted'

> https://github.com/avagin/criu/commit/7c01780c89608999db04e624533ed92ca07a9325 should fix this issue.

Incorporated the fix @rst0git @avagin :)

criu/cr-restore.c
```diff
@@ -1694,12 +1689,18 @@ static int __restore_task_with_children(void *_arg)
 		if (restore_wait_other_tasks())
 			goto err;
 		fini_restore_mntns();
-		__restore_switch_stage(CR_STATE_RESTORE);
+
+		if (!opts.stream && start_asyncd())
```
rst0git, line 1694: It would be good to add a comment also here why criu-image-streamer is incompatibile, e.g., serves images in a single sequential pass.

criu/namespaces.c
```diff
@@ -1221,7 +1221,7 @@ static int write_id_map(pid_t pid, UidGidExtent **extents, int n, char *id_map)
 
 static int usernsd_pid;
 
-inline void unsc_msg_init(struct unsc_msg *m, uns_call_t *c, int *x, void *arg, size_t asize, int fd, pid_t *pid)
+void unsc_msg_init_flags(struct unsc_msg *m, uns_call_t *c, int *x, void *arg, size_t asize, int fd, pid_t *pid, bool send_creds)
```
rst0git: `unsc_msg_init_flags` should be static?

criu/asyncd.c
```diff
@@ -0,0 +1,193 @@
...
+		ret = recvmsg(sk, &um.h, 0);
+		if (ret == 0)
+			break;
+		if (ret < 0) {
+			pr_perror("async: recv req error");
+			syscall(__NR_exit_group, 1);
+		}
+
+		unsc_msg_pid_fd(&um, &pid, &fd);
+		pr_debug("async: daemon calls %p (%d, %d, %x)\n", call, pid, fd, flags);
```
rst0git, line 39: I believe these logs run concurrently from the worker pool, but CRIU's logger formats into a single process-global buffer without locking (see `vprint_on_level()` in `criu/log.c`)

criu/namespaces.c
dfeigin-nv: Yes. Done

criu/cr-restore.c
dfeigin-nv: Done

criu/asyncd.c
dfeigin-nv: You are correct. See the discussion above with Adrian. https://github.com/checkpoint-restore/criu/pull/3021#issuecomment-4475580165
WDYT?

rst0git: > there is a little garbling, found 70 lines in a 49k-line restore log which are a off.

> It's the static buffer[] in criu/log.c::vprint_on_level() - two threads vsnprintf into the same staging area before the write(). AMD plugin has the same race.

> The PIE side (compel/plugins/std/log.c) already builds the line in a stack-local simple_buf per call. Mirroring that on the daemon side drops the statics and fixes both. WDYT? I can put it in as a separate commit in this PR.

Sure, I think it makes sense to be a separate commit in this PR.

rst0git:
LGTM

@dfeigin-nv @avagin I will rebase the branch for https://github.com/checkpoint-restore/criu/pull/2895 to include the patches here to use asyncd for decompression (https://github.com/checkpoint-restore/criu/pull/2895#discussion_r3429845032).

criu/shmem.c
```diff
@@ -544,13 +545,33 @@ int restore_memfd_shmem_content(int fd, unsigned long shmid, unsigned long size)
 	}
 
 	ret = 0;
-
 out:
 	if (addr)
 		munmap(addr, size);
 	return ret;
```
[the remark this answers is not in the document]
dfeigin-nv: Added

criu/namespaces.c
```diff
@@ -1470,14 +1488,16 @@ int start_unix_cred_daemon(pid_t *pid, int (*daemon_func)(int sk))
...
 	}
 
-	if (setsockopt(sk[0], SOL_SOCKET, SO_PASSCRED, &one, sizeof(one)) < 0) {
-		pr_perror("failed to setsockopt");
-		return -1;
-	}
+	if (passcred) {
+		if (setsockopt(sk[0], SOL_SOCKET, SO_PASSCRED, &one, sizeof(one)) < 0) {
+			pr_perror("failed to setsockopt");
+			return -1;
```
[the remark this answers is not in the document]
dfeigin-nv: Done

criu/asyncd.c
```diff
@@ -0,0 +1,193 @@
...
+		 */
+
+		ret = call(msg, fd, pid);
+
+		if (fd >= 0)
+			close(fd);
+
+		if (ret < 0) {
+			pr_err("async: Async call failed. Exiting\n");
+			syscall(__NR_exit_group, 1);
```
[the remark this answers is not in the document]
dfeigin-nv: Done

criu/asyncd.c
```diff
@@ -0,0 +1,193 @@
...
+		int flags, fd, ret;
+		pid_t pid;
+
+		unsc_msg_init_nocreds(&um, &call, &flags, msg, sizeof(msg), 0);
+		ret = recvmsg(sk, &um.h, 0);
+		if (ret == 0)
+			break;
+		if (ret < 0) {
+			pr_perror("async: recv req error");
+			syscall(__NR_exit_group, 1);
```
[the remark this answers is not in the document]
dfeigin-nv: Done

criu/asyncd.c
```diff
@@ -0,0 +1,193 @@
+#include <unistd.h>
+#include <sys/wait.h>
+#include <sys/stat.h>
+#include <errno.h>
+#include <pthread.h>
+#include <sys/socket.h>
```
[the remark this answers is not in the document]
dfeigin-nv: Done

criu/asyncd.c
dfeigin-nv: Done.

dfeigin-nv:
@rst0git Thanks. @avagin Do you think it's ready to merge? (Besides flakes/github-runner kernel version issues)

criu/cr-restore.c
```diff
@@ -3241,6 +3243,12 @@ static int sigreturn_restore(pid_t pid, struct task_restore_args *task_args, uns
 			goto err_nv;
 		if (root_ns_mask & CLONE_NEWNS && remount_readonly_mounts())
 			goto err_nv;
+		if (stop_asyncd())
+			goto err_nv;
```
rst0git, line 3254: Do we need to call `stop_asyncd()` also in the error path?
e.g., `err_nv` in `sigreturn_restore` and `err` in `__restore_task_with_children`

dfeigin-nv: Probably a good idea. Added

test/zdtm/static/Makefile
```diff
@@ -706,6 +705,7 @@ socket-tcp4v6-last-ack:	CFLAGS += -D ZDTM_TCP_LAST_ACK -D ZDTM_IPV4V6
 socket-tcp4v6-closing:	CFLAGS += -D ZDTM_IPV4V6
 memfd02-hugetlb:	CFLAGS += -D ZDTM_HUGETLB
 memfd05:		CFLAGS += -D ZDTM_MEMFD05
+memfd_threads:		LDLIBS += -pthread
```
rst0git: Should we add `memfd_threads` in `test/zdtm/static/.gitignore`?

criu/bfd.c
```diff
@@ -33,19 +34,23 @@ struct bfd_buf {
 };
 
 static LIST_HEAD(bufs);
+static pthread_mutex_t bufs_lock = PTHREAD_MUTEX_INITIALIZER;
```
rst0git, line 37: It looks like `static unsigned ids` in `open_page_read_at()` (`criu/pagemap.c`) is another shared variable on the parallel path where `pr->id = ids++` is a non-atomic read-modify-write.

```diff
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -8,6 +8,7 @@
 #include <limits.h>
 
 #include "types.h"
+#include "atomic.h"
 #include "image.h"
 #include "cr_options.h"
 #include "servicefd.h"
@@ -857,7 +858,8 @@ int probe_pages_o_direct(int fd)
 int open_page_read_at(int dfd, unsigned long img_id, struct page_read *pr, int pr_flags)
 {
        int flags, i_typ;
-       static unsigned ids = 1;
+       /* Shared across asyncd fill-daemon workers, which open page-reads concurrently. */
+       static atomic_t ids = ATOMIC_INIT(0);
        bool remote = pr_flags & PR_REMOTE;
 
        /*
@@ -943,7 +945,7 @@ int open_page_read_at(int dfd, unsigned long img_id, struct page_read *pr, int p
        pr->seek_pagemap = seek_pagemap;
        pr->reset = reset_pagemap;
        pr->io_complete = NULL; /* set up by the client if needed */
-       pr->id = ids++;
+       pr->id = atomic_inc_return(&ids);
        pr->img_id = img_id;
```

test/zdtm/static/Makefile
dfeigin-nv: I think that file is autogenerated.

criu/bfd.c
dfeigin-nv: Yes you're right. I missed that. I'll apply this diff.

test/zdtm/static/Makefile
rst0git: Yes, it is (https://github.com/checkpoint-restore/criu/commit/fe45e3626aba5a86939bd6cfebe8508e2b679c52) - I had to git-clean & update it locally.

rst0git (approved):
LGTM

test/zdtm/static/memfd_threads.c
```diff
@@ -0,0 +1,130 @@
...
+#include <stdlib.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <sys/types.h>
+#include <unistd.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "memfd content restore with many parked threads";
```
avagin: I think we may consider adding threads to one of the memfd tests.

test/zdtm/static/memfd_threads.c
```diff
@@ -0,0 +1,130 @@
...
+	})
+
+static task_waiter_t tw[NR_THREADS];
+
+static int _memfd_create(const char *name, unsigned int flags)
+{
+	return syscall(SYS_memfd_create, name, flags);
+}
+
+static void fill_pattern(char *buf, int variant)
```
avagin: we already have datagen/datachk. You can look at test/zdtm/static/maps04.c as an example.

test/zdtm/static/memfd_threads.c
```diff
@@ -0,0 +1,130 @@
...
+	}
+}
+
+static void *parked_thread(void *arg)
+{
+	long idx = (long)arg;
+
+	/* Signal "alive", then park until the main task wakes us post-restore. */
+	task_waiter_complete(&tw[idx], 1);
+	task_waiter_wait4(&tw[idx], 2);
```
avagin: nit: if I am not mistaken, task_waiter-s were added to synchronize multiple processes. In case of threads, It may be easier to use pthread_cond_signal / pthread_cond_wait.

test/zdtm/static/memfd_threads.c
```diff
@@ -0,0 +1,130 @@
...
+		err(1, "Can't mmap memfd");
+
+	for (i = 0; i < NR_THREADS; i++) {
+		if (pthread_create(&threads[i], NULL, parked_thread, (void *)i))
+			err(1, "Can't pthread_create");
+	}
+
+	/* Make sure every thread is alive and parked before C/R. */
+	for (i = 0; i < NR_THREADS; i++)
+		task_waiter_wait4(&tw[i], 1);
```
avagin: threads don't check anything. pthread_create guarantees a thread creation. I think this sync point isn't required.

test/zdtm/static/memfd_threads.c
```diff
@@ -0,0 +1,130 @@
...
+ * thread at its original TID via clone3(set_tid=...), the clone fails with
+ * EEXIST ("Unable to create a thread: -17"). The fix drains and reaps the
+ * daemon during CR_STATE_PRE_RESTORER, before any application thread is
+ * cloned. memfd06 (no threads) cannot catch this.
+ */
+
+#define NR_THREADS 64
+#define LEN	   PAGE_SIZE
+
+#define err(exitcode, msg, ...)                \
```
avagin: nit: I would prefer avoiding this sort of helpers. It might be convenient in a moment, but for contributors who work with random tests, it is much better when all tests use the same style.

test/zdtm/static/memfd_threads.c
dfeigin-nv: I fixed the test according to the other comments. I didn't quite catch this. You would like to consolidate the test to another of the memfd tests? Or keep this as a standalone test file?

test/zdtm/static/memfd_threads.c
dfeigin-nv: Removed

test/zdtm/static/memfd_threads.c
dfeigin-nv: It is easier, thanks

test/zdtm/static/memfd_threads.c
dfeigin-nv: Cool, done.

test/zdtm/static/memfd_threads.c
dfeigin-nv: Reverted to the same style.

criu/bfd.c
```diff
@@ -54,6 +59,7 @@ static int buf_get(struct xbuf *xb)
 			b = xmalloc(sizeof(*b));
 			if (!b) {
 				if (i == 0) {
+					pthread_mutex_unlock(&bufs_lock);
 					pr_err("No buffer for bfd\n");
 					return -1;
```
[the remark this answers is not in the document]
dfeigin-nv: Done.

dfeigin-nv:
@avagin the log thread-safety fix got dropped in your latest push; asyncd workers still log concurrently. Was this by mistake?
d9739422b0722fee56b0013e7abfefdca23c07df

avagin:
Merged with a few changes. @dfeigin-nv thanks a lot for working on this.

PR #3022, opened 2026-05-17
pie: Reduce restore time by replacing synchronous preadv with native AIO for private VMA restore

dfeigin-nv:
### Issue

Restore time on large checkpoints (inference engines, e.g. **vLLM**) is dominated by synchronous per-chunk reads against `pages.img` in the PIE restorer.  Each `pread()` blocks the restorer thread until the kernel returns.

On a 134 GB checkpoint the current `criu-dev` spends 116 s in the CRIU restore phase; this PR by itself removes 67 s of that.

### Idea

Replace the synchronous per-chunk read loop with native Linux AIO for VMA page reads.  Submissions for one VMA are batched into a single submit; we wait on the batch with one event call before moving on.

Commits:

* `compel: add io_destroy syscall` — the PIE restorer cannot link
  libc, so any syscall it issues has to be in compel's table.
* `pagemap/mem: probe O_DIRECT for page reads` — startup probe of
  the backing fs; falls back to buffered reads when O_DIRECT is
  unavailable or alignment cannot be satisfied.
* `shmem: batch page reads with PR_ASYNC` — switches shmem page
  restore to the existing PR_ASYNC pagemap flag so it defers waits
  and shares the AIO submit path.
* `pie: restore VMA pages with native AIO` — main change in the PIE
  restorer.

### Benchmarks

CRIU restore timing, for vLLM loaded with an AI model, 3 cold runs each, mean ± stddev.  The "combined" column applies this PR on top of the parallel memfd-open PR:

| checkpoint model |   size  | base criu-dev | this PR     | diff | combined w/ parallel memfd (https://github.com/checkpoint-restore/criu/pull/3021)|
|--------------|-------:|---------------|-------------|------|----------------------------|
| Qwen3-0.6B   | 7.8 GB | 9.3±0.3 s     | 4.0±0.2 s   | -57% | 2.9±0.7 s (-68%)           |
| Qwen3-8B     |  28 GB | 26.1±1.0 s    | 11.1±0.4 s  | -57% | 5.3±0.1 s (-80%)           |
| Qwen3-14B    |  49 GB | 39.0±6.3 s    | 19.1±0.2 s  | -51% | 8.6±1.2 s (-78%)           |
| gpt-oss-120B | 134 GB | 116.1±1.5 s   | 49.4±0.5 s  | -57% | 18.6±0.6 s (-84%)          |


adrianreber:
Compilation breaks very early on Alpine. Please take a look.

criu/mem.c
```diff
@@ -1519,6 +1510,56 @@ static int prepare_vma_ios(struct pstree_item *t, struct task_restore_args *ta)
...
 	ta->vma_ios_fd = img_raw_fd(pages);
+	if (ta->vma_ios_fd >= 0) {
+		int fl = fcntl(ta->vma_ios_fd, F_GETFL);
+		if (fl >= 0) {
+			int ret = fcntl(ta->vma_ios_fd, F_SETFL, fl | O_DIRECT);
+			if (ret < 0) {
+				pr_warn("Failed to set O_DIRECT on pages fd: %s\n", strerror(errno));
+			} else {
+				/*
+				 * Probe that O_DIRECT reads actually work. Some
```
adrianreber: This comment and the whole hunk look very much like the changes to pagemap.c (around line 918). Is this something that can be a helper?

criu/pagemap.c
```diff
@@ -823,6 +915,65 @@ int open_page_read_at(int dfd, unsigned long img_id, struct page_read *pr, int p
...
+
+						pr_info("O_DIRECT rejected at read time on pages fd %d, using buffered I/O\n",
+							pfd);
+						if (fcntl(pfd, F_SETFL, fl) < 0) {
+							pr_perror("Failed to clear O_DIRECT on pages fd %d", pfd);
+							close_page_read(pr);
+							return -1;
+						}
+
+						/* Hint the kernel that fallback buffered reads will mostly
```
adrianreber: This is not the comment style CRIU prefers:
```c
/*
 * ...
```

like you did above.

criu/pagemap.c
```diff
@@ -527,49 +551,108 @@ static void advance_piov(struct page_read_iov *piov, ssize_t len)
...
 		struct iovec *iovs = piov->to;
+		bool io_failed = false;
 
 		pr_debug("Read piov iovs %d, from %ju, len %ju, first %p:%zu\n", piov->nr, piov->from,
 			 piov->end - piov->from, piov->to->iov_base, piov->to->iov_len);
 	more:
-		ret = preadv(fd, piov->to, piov->nr, piov->from);
-		if (fault_injected(FI_PARTIAL_PAGES)) {
-			/*
+			bytes = preadv(fd, piov->to, piov->nr, piov->from);
```
adrianreber: This indentation looks wrong. One level too much.

criu/pie/Makefile
```diff
@@ -29,6 +29,8 @@ endif
 
 LDS		:= compel/arch/$(ARCH)/scripts/compel-pack.lds.S
 
+ccflags-y	+= -iquote $(obj)
```
adrianreber: What is this for?

adrianreber:
The commit messages should have line breaks. In CONTRIBUTING.md we link to https://chris.beams.io/posts/git-commit/ which says:

    Limit the subject line to 50 characters
    ...
    Wrap the body at 72 characters

Something in that direction would be nice.

avagin:
> The commit messages should have line breaks. In CONTRIBUTING.md we link to https://chris.beams.io/posts/git-commit/ which says:
> 
> ```
> Limit the subject line to 50 characters
> ...
> Wrap the body at 72 characters
> ```
> 
> Something in that direction would be nice.

This is a wrong guide, we don't follow this style. The proper one is described here: https://www.kernel.org/doc/html/v4.17/process/submitting-patches.html#the-canonical-patch-format

criu/pie/restorer.c
```diff
@@ -1927,67 +1942,172 @@ __visible long __export_restore_task(struct task_restore_args *args)
 	}
 
 	/*
-	 * Now read the contents (if any)
+	 * Now read the contents (if any).
+	 * Use Native AIO (io_submit + O_DIRECT) for high-throughput page reads.
+	 * Handles short reads (MAX_RW_COUNT 0x7FFFF000) by advancing and resubmitting.
```
avagin: I think it is better to introduce a helper to restore private vma data:
https://docs.kernel.org/process/coding-style.html#indentation

criu/pie/restorer.c
```diff
@@ -1927,67 +1942,172 @@
...
+			rio = (void *)rio + RIO_SIZE(rio->nr_iovs);
+		}
 
-			if (r == 0) {
-				pr_err("Unexpected EOF reading pages data at offset %ld (%d iovs remaining)\n",
-				       (long)rio->off, nr);
-				goto core_restore_end;
+		/* Submit and reap in batches */
+		while (submitted < n || completed < n) {
+			while (submitted < n && (submitted - completed) < AIO_BATCH) {
```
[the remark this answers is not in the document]
dfeigin-nv: Fixed

criu/pie/Makefile
dfeigin-nv: Unused - no file under `criu/pie/` includes a header from `criu/pie/`. Dropped.

criu/pie/restorer.c
dfeigin-nv: Indeed. Done.

criu/pagemap.c
dfeigin-nv: Fixed

criu/pagemap.c
dfeigin-nv: Fixed.

criu/mem.c
dfeigin-nv: Yes, and done.

criu/pie/restorer.c
dfeigin-nv: I created several helper functions to make the code nicer and more readable

criu/pagemap.c
```diff
@@ -249,18 +253,38 @@ static int read_local_page(struct page_read *pr, unsigned long vaddr, unsigned l
 	if (pr->sync(pr))
 		return -1;
 
+	if (pr->use_direct) {
+		if (((unsigned long)buf & (PAGE_SIZE - 1)) || (len & (PAGE_SIZE - 1)) ||
```
avagin: please put a comment here to explain what is going on here. We probably can just fail if buf or len are not aligned... wdyt?

criu/pagemap.c
```diff
@@ -762,6 +786,60 @@ static int init_pagemaps(struct page_read *pr)
...
+	 * probe buffer is allocated via posix_memalign() instead of a
+	 * stack array with __attribute__((aligned)).
+	 */
+	memerr = posix_memalign(&probe, PAGE_SIZE, PAGE_SIZE);
+	if (memerr) {
+		pr_err("O_DIRECT probe alloc failed on pages fd %d: %s\n", fd, strerror(memerr));
+		return -1;
+	}
+
+	probe_ret = pread(fd, probe, PAGE_SIZE, 0);
```
avagin, line 830: Have you seen in practice that fcntl(fd, F_SETFL, fl | O_DIRECT) succeeded, but read from fd failed?

criu/pie/restorer.c
```diff
@@ -1607,36 +1610,267 @@ static int fd_poll(int inotify_fd)
...
+			pr_err("AIO sub-page short read: %lld of %llu\n", ev->res, ev->data);
+			return -1;
+		}
+	}
 
-	ret = sys_preadv(fd, iovs, nr, offs);
-	if (saved_last_iov_len)
-		iovs[nr - 1].iov_len = saved_last_iov_len;
+	if (args->auto_dedup)
+		sys_fallocate(fd, FALLOC_FL_KEEP_SIZE | FALLOC_FL_PUNCH_HOLE,
```
avagin:
```
				int fr = sys_fallocate(args->vma_ios_fd, FALLOC_FL_KEEP_SIZE | FALLOC_FL_PUNCH_HOLE,
1963		-
						       rio->off, r);
1964		-
				if (fr < 0) {
1965		-
					pr_debug("Failed to punch holes with fallocate: %d\n", fr);
1966		-
				}
```

I don't think that we really need to remove this debug message...

avagin:
@dfeigin-nv Overall, the series looks pretty good. Thank you for working on this. The Linux AIO interface is a bit clunky. Out of curiosity, have you tried implementing this using io_uring?

dfeigin-nv:
Thank you @avagin

Tried io_uring earlier in this work. 2 main issues arose I didn't solve:

1. **Worker threads.** io_uring spawns `iou-wrk-*` workers (`PF_IO_WORKER`) into the calling task's thread group, visible in `/proc/<pid>/task/`. Since the PIE restorer becomes the restored task, those workers would serve as extra TIDs the dump image never captured, diverging from `nr_threads_image`. Not unsolvable, but felt like a lot of effort compared to the speedup AIO already gave with much less work.

2. **Container seccomp.** Both Docker's default profile and CRI-O's (the one OpenShift inherits) default-deny and don't allowlist `io_uring_setup` / `_enter` / `_register`; the four AIO syscalls are explicitly allowed in both. Since CRIU restore may run inside workload pods, I didn't want to force every deployment to ship a custom seccomp profile just to support the restorer.

criu/shmem.c
```diff
@@ -468,81 +469,107 @@ static int shmem_wait_and_open(struct shmem_info *si, VmaEntry *vi)
...
 
-	/*
-	 * do_restore_shmem_content needs size to be page aligned.
-	 */
-	if (do_restore_shmem_content(addr, round_up(size, PAGE_SIZE), shmid) < 0) {
-		pr_err("Can't restore shmem content\n");
-		goto out;
-	}
+	ret = shmem_restore_async(&pr, addr, aligned_size);
+	pr.close(&pr);
```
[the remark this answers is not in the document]
dfeigin-nv: Tested with reproducer, no SIGBUS. Linux only signals on whole pages past EOF.

criu/pie/restorer.c
dfeigin-nv: Right - lost when rewriting the preadv path into AIO. Adding the pr_debug back.

criu/pagemap.c
dfeigin-nv: Yes - I tested it on AKS with Azure Files NFS. The filesystem accepts fcntl(F_SETFL, O_DIRECT) but rejects the first preadv with EINVAL, so w/o the fallback the restore would fail mid-run without the probe.

criu/pagemap.c
dfeigin-nv: I'll add a comment. Regarding failing -  if we just fail here, restore breaks in a real path: the COW-compare loop in `mem.c::restore_priv_vma_content` declares unsigned `char buf[PAGE_SIZE]` on the stack and uses it to read each child page, so we can memcmp it against the parent - that's how we decide whether the page is still shared via copy-on-write or needs a real write. The stack only guarantees 16-byte alignment (I think that depends on CPU arch).
So, every one of those reads would be refused while O_DIRECT is on. That's why the bounce buffer was added. I can also change the stack allocation code in mem.c to call `posix_memalign`, or just add the comment. Whichever you prefer.

criu/mem.c
```diff
@@ -1509,6 +1510,12 @@ static int prepare_vma_ios(struct pstree_item *t, struct task_restore_args *ta)
 		return -1;
 
 	ta->vma_ios_fd = img_raw_fd(pages);
+	if (ta->vma_ios_fd >= 0) {
+		if (probe_pages_o_direct(ta->vma_ios_fd) < 0) {
+			close_image(pages);
+			return -1;
+		}
+	}
```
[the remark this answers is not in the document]
dfeigin-nv: Fixed

criu/pagemap.c
```diff
@@ -527,48 +562,107 @@ static void advance_piov(struct page_read_iov *piov, ssize_t len)
...
+			} else {
+				if (piov->from < first_off)
+					first_off = piov->from;
+				if (piov->end > last_end)
+					last_end = piov->end;
+			}
+		}
+		if (have_range && last_end > first_off) {
+			if (posix_fadvise(fd, first_off, (off_t)(last_end - first_off), POSIX_FADV_WILLNEED) != 0)
+				pr_debug("posix_fadvise(WILLNEED) failed for async range\n");
```
bsach64: this chunk can be rewritten as
```
diff --git a/criu/pagemap.c b/criu/pagemap.c
index c5eec8922..6f6aa8381 100644
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -28,6 +28,9 @@

 #define MAX_BUNCH_SIZE 256

+#define OFF_MAX (sizeof(off_t) == sizeof(long long) ? LLONG_MAX : sizeof(off_t) == sizeof(int) ? INT_MAX : -999999)
+#define OFF_MIN (sizeof(off_t) == sizeof(long long) ? LLONG_MIN : sizeof(off_t) == sizeof(int) ? INT_MIN : -999999)
+
 /*
  * One "job" for the preadv() syscall in pagemap.c
  */
@@ -584,24 +587,15 @@ static int process_async_reads(struct page_read *pr)
 {
        int fd, ret = 0;
        struct page_read_iov *piov, *n;
-       off_t first_off = 0, last_end = 0;
-       bool have_range = false;
+       off_t first_off = OFF_MAX, last_end = OFF_MIN;

        fd = img_raw_fd(pr->pi);
        if (!pr->use_direct) {
                list_for_each_entry(piov, &pr->async, l) {
-                       if (!have_range) {
-                               first_off = piov->from;
-                               last_end = piov->end;
-                               have_range = true;
-                       } else {
-                               if (piov->from < first_off)
-                                       first_off = piov->from;
-                               if (piov->end > last_end)
-                                       last_end = piov->end;
-                       }
+                       first_off = min(piov->from, first_off);
+                       last_end = max(piov->end, last_end);
                }
-               if (have_range && last_end > first_off) {
+               if (last_end > first_off) {
                        if (posix_fadvise(fd, first_off, (off_t)(last_end - first_off), POSIX_FADV_WILLNEED) != 0)
                                pr_debug("posix_fadvise(WILLNEED) failed for async range\n");
                }
```

criu/pagemap.c
```diff
@@ -578,10 +672,18 @@ static int process_async_reads(struct page_read *pr)
...
 		}
 
+		if (io_failed) {
+			list_del(&piov->l);
+			xfree(iovs);
+			xfree(piov);
+			drain_async_queue(pr);
+			return -1;
+		}
+
```
bsach64: this chunk can be rewritten as
```
diff --git a/criu/pagemap.c b/criu/pagemap.c
index c5eec8922..7a55836da 100644
--- a/criu/pagemap.c
+++ b/criu/pagemap.c
@@ -610,7 +610,6 @@ static int process_async_reads(struct page_read *pr)
 	list_for_each_entry_safe(piov, n, &pr->async, l) {
 		ssize_t bytes;
 		struct iovec *iovs = piov->to;
-		bool io_failed = false;

 		pr_debug("Read piov iovs %d, from %ju, len %ju, first %p:%zu\n", piov->nr, piov->from,
 			 piov->end - piov->from, piov->to->iov_base, piov->to->iov_len);
@@ -633,11 +632,13 @@ static int process_async_reads(struct page_read *pr)
 		if (bytes < 0) {
 			pr_err("Can't read async pr bytes (%zd / %ju read, %ju off, %d iovs)\n", bytes,
 			       piov->end - piov->from, piov->from, piov->nr);
-			io_failed = true;
+			ret = -1;
+			goto cleanup;
 		} else if (bytes == 0) {
 			pr_err("Unexpected EOF in async page read (%ju bytes remaining at off %ju, %d iovs)\n",
 			       piov->end - piov->from, piov->from, piov->nr);
-			io_failed = true;
+			ret = -1;
+			goto cleanup;
 		} else {
 			/*
 			 * O_DIRECT requires page-aligned retry. Mask
@@ -651,18 +652,20 @@ static int process_async_reads(struct page_read *pr)

 				if (aligned == 0) {
 					pr_err("Sub-page short read on O_DIRECT fd: %zd bytes\n", bytes);
-					io_failed = true;
+					ret = -1;
+					goto cleanup;
 				} else {
 					bytes = aligned;
 				}
 			}

-			if (!io_failed && opts.auto_dedup &&
-			    punch_hole(pr, piov->from, bytes, false))
-				io_failed = true;
+			if (opts.auto_dedup && punch_hole(pr, piov->from, bytes, false)) {
+				ret = -1;
+				goto cleanup;
+			}
 		}

-		if (!io_failed && bytes != piov->end - piov->from) {
+		if (bytes != piov->end - piov->from) {
 			/*
 			 * The preadv() can return less than requested. It's
 			 * valid and doesn't mean error or EOF. We should advance
@@ -676,19 +679,15 @@ static int process_async_reads(struct page_read *pr)
 			goto more;
 		}

-		if (io_failed) {
-			list_del(&piov->l);
-			xfree(iovs);
-			xfree(piov);
-			drain_async_queue(pr);
-			return -1;
-		}
-
 		BUG_ON(pr->io_complete); /* FIXME -- implement once needed */
-
+cleanup:
 		list_del(&piov->l);
 		xfree(iovs);
 		xfree(piov);
+		if (ret < 0) {
+			drain_async_queue(pr);
+			return -1;
+		}
 	}

 	if (pr->parent)
```

criu/shmem.c
```diff
@@ -468,81 +469,107 @@ static int shmem_wait_and_open(struct shmem_info *si, VmaEntry *vi)
...
+		}
 	}
 
-	pr.close(&pr);
-	return ret;
-}
+	if (ret < 0) {
+		pr->sync(pr); /* drain async queue before close */
+		return -1;
+	}
```
bsach64: this can be rewritten as
```
index 5ce024f56..fe9ef4589 100644
--- a/criu/shmem.c
+++ b/criu/shmem.c
@@ -478,7 +478,9 @@ static int shmem_restore_async(struct page_read *pr, void *addr, unsigned long s
 		unsigned nr_pages;

 		ret = pr->advance(pr);
-		if (ret <= 0)
+		if (ret < 0)
+			goto err;
+		if (ret == 0)
 			break;

 		vaddr = (unsigned long)decode_pointer(pr->pe->vaddr);
@@ -486,23 +488,18 @@ static int shmem_restore_async(struct page_read *pr, void *addr, unsigned long s

 		if (vaddr + nr_pages * PAGE_SIZE > size) {
 			pr_err("Shmem read out of bounds: %lx + %lu > %lx\n", vaddr, (unsigned long)nr_pages * PAGE_SIZE, size);
-			pr->sync(pr); /* drain async queue before close */
-			return -1;
+			goto err;
 		}

 		ret = pr->read_pages(pr, vaddr, nr_pages, addr + vaddr, PR_ASYNC);
-		if (ret < 0) {
-			pr->sync(pr); /* drain async queue before close */
-			return -1;
-		}
-	}
-
-	if (ret < 0) {
-		pr->sync(pr); /* drain async queue before close */
-		return -1;
+		if (ret < 0)
+			goto err;
 	}

 	return pr->sync(pr);
+err:
+	pr->sync(pr); /* drain async queue before close */
+	return -1;
 }

 static int restore_shmem_content_from_pages(void *addr, unsigned long size, unsigned long shmid)

```

bsach64 (commented):
All of these are refactors that make it easier to reason about the code, feel free to not apply them if you disagree.

criu/pagemap.c
avagin: if you commit these changes and post a link here to the commit here, it will be much easier to read the proposed code.

avagin:
> * **Worker threads.** io_uring spawns `iou-wrk-*` workers (`PF_IO_WORKER`) into the calling task's thread group, visible in `/proc/<pid>/task/`. Since the PIE restorer becomes the restored task, those workers would serve as extra TIDs the dump image never captured, diverging from `nr_threads_image`. Not unsolvable, but felt like a lot of effort compared to the speedup AIO already gave with much less work.

io_uring_queue_exit() destroys all associated helper threads. The native Linux AIO is highly limited; it only works asynchronously with file systems that support O_DIRECT. Without O_DIRECT, it silently falls back to synchronous execution.

I think the best path forward is to introduce an I/O abstraction layer to seamlessly swap between our current synchronous read implementation, Linux AIO, and any future asynchronous engines (such as io_uring) via configuration/feature probes.

> * **Container seccomp.** Both Docker's default profile and CRI-O's (the one OpenShift inherits) default-deny and don't allowlist `io_uring_setup` / `_enter` / `_register`; the four AIO syscalls are explicitly allowed in both. Since CRIU restore may run inside workload pods, I didn't want to force every deployment to ship a custom seccomp profile just to support the restorer.

Does criu work with the standard seccomp profile now?

criu/pagemap.c
bsach64: please look at this: https://github.com/bsach64/criu/commit/4a6e31906a877f52a7b976d4cd8d17ee45200c58

criu/pagemap.c
bsach64: please look at: https://github.com/bsach64/criu/commit/1a6ada6c15538750c09c78e9b92c74af6205c0cb

criu/shmem.c
bsach64: please look at: https://github.com/bsach64/criu/commit/8c002cd6c2b944cc12ae1e1c71cf8d7b18232b56

dfeigin-nv:
Thanks @avagin. Quick context on why this PR landed on AIO: the existing restorer path uses synchronous `sys_preadv` for VMA reads. The goal here is to add parallelism without changing the thread set CRIU's restore depends on. io_uring was the obvious candidate, but when we tried it inside the restorer, reads that couldn't complete inline got handed off to kernel helper threads (`iou-wrk-*`) inside the dumpee's thread group — and those threads stuck around past `close(ring_fd)` long enough to break CRIU's own thread-count check. `io_submit` doesn't spawn those workers, so this PR replaces the restorer's `preadv` with `io_submit` and probes the page-image fd for `O_DIRECT` at restore time. When the filesystem supports `O_DIRECT`, the AIO path is genuinely async. When it doesn't, the fd stays buffered and `io_submit` runs synchronously inside the kernel — same synchronous/blocking behavior as the `preadv` it replaces, and no worker threads either way.

On `io_uring_queue_exit()` specifically: it does *start* the worker teardown, but `close(ring_fd)` returns to userspace before the workers are gone from `/proc/<pid>/task/`. The teardown is dispatched onto a workqueue — `io_ring_ctx_wait_and_kill` calls `queue_work(iou_wq, &ctx->exit_work)` and returns; the actual reaping (signaling workers to exit, cancelling in-flight requests, waiting for ctx refs to drop) happens later inside `io_ring_exit_work`. Workers that are mid-IO when we close still have to finish their current syscall before they can exit, and the userspace-visible window between `close()` returning and the worker TIDs disappearing isn't tightly bounded.

That gap is what bit us. Right after the PIE restorer signals it's done (`restore_finish_stage(CR_STATE_RESTORE)`), the master process walks `/proc/<pid>/task/` via `parse_threads` and aborts if the live thread count doesn't match the dump's `nr_threads_image`. With io_uring active, we consistently saw leftover `iou-wrk-*` worker threads still in the restored task's `/proc/task/` at that walk, and short post-close sleeps weren't reliable enough to close the gap.

`io_submit` sidesteps this by keeping submission in-context — no kernel threads appear in the dumpee's thread group, so the restorer exits cleanly w.r.t. the thread-count check. That's the architectural reason this PR uses AIO rather than a preference for the interface itself.

I'd be glad to do an I/O backend abstraction as a separate follow-up PR. That's the natural place to introduce alternative engines (including io_uring) once there's a way to integrate them without the thread-count regression. This PR is the concrete perf improvement we can ship today inside the existing contract.

**standard seccomp** — yes, CRIU works on the default profile unchanged. The syscalls this PR relies on (`io_setup` / `io_submit` / `io_getevents` / `io_destroy`) are all in the default Docker/Moby seccomp allowlist and in the container-runtime defaults pulled from `containers/common`. `io_uring_*` is not — both projects deliberately removed it from defaults for security ([moby#46762](https://github.com/moby/moby/pull/46762), [containerd#9048](https://github.com/containerd/containerd/issues/9048)).

avagin:
Thank you for the explanation regarding  io_uring . Native AIO has some downsides as well:
 * It can be disabled in the kernel configuration ( CONFIG_AIO ).
 * It is subject to a global limit ( fs.aio-max-nr ).
 
 @rst0git is working on memory image compression, which will require us to introduce another code path here anyway.
 
> I'd be glad to do an I/O backend abstraction as a separate follow-up PR. That's the natural place to introduce alternative engines (including io_uring) once there's a way to integrate them without the thread-count regression. This PR is the concrete perf improvement we can ship today inside the existing contract.

@dfeigin-nv: Calling it an "I/O backend abstraction" might make it sound more complicated than it actually is. In reality, I mean a very simple approach: we just move the current  preadv  implementation into a helper function and then choose the engine (AIO vs. sync)  dynamically based on  O_DIRECT  support. This is exactly what we have done in these reworked patches to keep the implementation simple and clean.

criu/pagemap.c
dfeigin-nv: Thank you, applied this.

criu/shmem.c
dfeigin-nv: Thank you. Applied this

criu/pagemap.c
dfeigin-nv: Thank you. Applied this.

dfeigin-nv:
Thanks @avagin , took your suggestion and pushed something along that idea. preadv is back and the call site picks it or AIO based on the O_DIRECT probe. The probe result lands in `task_restore_args.vma_ios_use_direct` so the PIE side just reads a bool.

Also picked up @bsach64's 3 readability commits. Thanks Bhavik.
And rebased over criu-dev.

Let me know if there is anything else or is this good to go? :)

avagin:
@dfeigin-nv could you please take a look at https://github.com/avagin/criu/commits/aio/? I think with these changes the code looks a bit cleaner.

dfeigin-nv:
Thanks @avagin - went through your repo and folded all of it in, I agree it reads cleaner. I squashed these into the original four commits rather than stacking them on top, so the series doesn't add-then-remove changes. I hope that's ok.
Good to go from my side - anything else? :)

criu/pie/restorer.c
```diff
@@ -1645,6 +1818,178 @@
...
+					continue;
+				}
+
+				iovs->iov_base += r;
+				iovs->iov_len -= r;
+				break;
+			} while (nr > 0);
+		}
+
+		rio = ((void *)rio) + RIO_SIZE(rio->nr_iovs);
```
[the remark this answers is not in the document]
dfeigin-nv: Fixed

criu/pie/restorer.c
```diff
@@ -1645,6 +1818,178 @@
...
+			if (r == 0) {
+				pr_err("Unexpected EOF reading pages data at offset %ld (%d iovs remaining)\n",
+				       (long)rio->off, nr);
+				return -1;
+			}
+
+			pr_debug("`- returned %ld\n", (long)r);
+			/* If the file is open for writing, then it means we should punch holes
+			 * in it. */
+			if (r > 0 && args->auto_dedup) {
```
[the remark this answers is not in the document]
dfeigin-nv: Fixed

bsach64:
Hey @dfeigin-nv, can you provide some more details for reproducing the benchmark results that you show in the PR description, things like the GPU you used, how you ran criu (which options did you pass) and how you circumvented the [know limitations](https://github.com/checkpoint-restore/criu/tree/criu-dev/plugins/cuda#known-limitations) of the plugin?

rst0git:
@bsach64 You can find more information in the following paper and pull request:
- [CRIUgpu: Transparent Checkpointing of GPU-Accelerated Workloads](https://radostin.io/files/criugpu-2025.pdf)
- https://github.com/ai-dynamo/dynamo/pull/4978

dfeigin-nv:
Thanks @bsach64 — full writeup of the approach here: https://developer.nvidia.com/blog/nvidia-dynamo-snapshot-fast-startup-for-inference-workloads-on-kubernetes/ . Details below.

Up front, since it reframes your third question: we don't use CRIU's CUDA plugin at all. CRIU runs with no plugin dir configured (`libdir` intentionally empty — external `cuda-checkpoint` only), so the plugin never loads. GPU state is handled out of band by NVIDIA's `cuda-checkpoint`: `lock` → `checkpoint`, then `criu dump`; on the way back, `criu restore`, then `cuda-checkpoint restore` → `unlock`. CRIU only handles the host-side process tree.

**GPU:** one NVIDIA B200, single GPU, tensor-parallel size 1.

**criu options:** go-criu RPC with `tcp-close`, `ext-unix-sk`, `file-locks`, `link-remap`, `shell-job`, `manage-cgroups` (soft), `evasive-devices`, `force-irmap`, `rst-sibling`, an inherited net namespace and stdio pipes; `criu.conf` adds `allow-uprobes`, `skip-in-flight`, empty `libdir`. Assembled in [`restore.go`](https://github.com/ai-dynamo/dynamo/blob/main/deploy/snapshot/internal/criu/restore.go) and [`util.go`](https://github.com/ai-dynamo/dynamo/blob/main/deploy/snapshot/internal/criu/util.go).

**On the linked plugin limitations** — plugin's out of the loop, so:
- *GPU mem → host RAM:* we checkpoint vLLM asleep. `/sleep` (level 1) before the dump frees the KV cache and offloads weights to RAM, so the GPU is nearly empty at checkpoint; refilled only on `/wake_up` after restore.
- *cuInit / PAUSE_DEVICES race:* avoided — we only checkpoint an idle, asleep server, never mid-init.
- *NVML / fork `/dev/nvidia*` refs:* we don't special-case these; they don't block checkpointing here (single vLLM process, asleep) — either `cuda-checkpoint`'s detach clears them or vLLM doesn't leave them on this path.
- *Same GPU type/count:* on restore the agent passes `cuda-checkpoint` a device map (`--device-map`, old→new UUID) so a checkpoint can come back on a different physical GPU of the same type; count must match. Single-GPU here.

dfeigin-nv:
@avagin, @rst0git The socket-tcpbuf family (socket-tcpbuf, socket-tcpbuf6, socket-tcpbuf6-local) is failing in CI, but I don't think it's my changes.

It fails at restore in the TCP window path: `Restoring TCP 1 queue data …` → `Error (soccr/soccr.c:750): Unable to set window parameters` → `Error (criu/cr-restore.c:2369): Restoring FAILED`. That's **soccr**, not the private-VMA page reads this PR touches. Looks like a runner-kernel change in TCP window handling or something like that.

Let me know how you think is best to proceed.

criu/pagemap.c
```diff
@@ -28,6 +28,9 @@
 
 #define MAX_BUNCH_SIZE 256
 
+#define OFF_MAX (sizeof(off_t) == sizeof(long long) ? LLONG_MAX : sizeof(off_t) == sizeof(int) ? INT_MAX : -999999)
```
avagin, line 31: nit: It took me sometime to figure out what this -999999 is for here. I think you can consider to use this macros:
tools/testing/selftests/nolibc/nolibc-test.c:#define SINT_MAX_OF_TYPE(type) (((type)1 << (sizeof(type) * 8 - 2)) - (type)1 + ((type)1 << (sizeof(type) * 8 - 2)))
tools/testing/selftests/nolibc/nolibc-test.c:#define SINT_MIN_OF_TYPE(type) (-SINT_MAX_OF_TYPE(type) - 1)

dfeigin-nv: Thanks, I'll put it in the next PR

avagin:
This pr introduced a major performance regression. I haven't look at it closely to figure out what exactly goes wrong:
```
root@criu-next-2:/home/avagin/git/criu# python3 test/zdtm.py run -t zdtm/static/maps04 -f h
userns is supported
=== Run 1/1 ================ zdtm/static/maps04
========================= Run zdtm/static/maps04 in h ==========================
 DEP       maps04.d
 CC        maps04.o
 LINK      maps04
Start test
./maps04 --pidfile=maps04.pid --outfile=maps04.out
Run criu dump
Execution time: 4.971509 seconds
Run criu restore
==== ALARM ====
Execution time: 56.193118 seconds
Send the 15 signal to  84
Wait for zdtm/static/maps04(84) to die for 0.100000
Wait for zdtm/static/maps04(84) to die for 0.200000
Removing dump/zdtm/static/maps04/84
========================= Test zdtm/static/maps04 PASS =========================
root@criu-next-2:/home/avagin/git/criu# git diff
index 6ffc2617807a..05256ef331e1 100755
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1297,8 +1297,12 @@ class criu:
 
         preload_libfault = self.__preload_libfault and action in ['dump', 'pre-dump', 'restore']
 
+        start_time = time.perf_counter()
         ret = self.__criu.run(action, s_args, self.__criu_bin, self.__fault,
                               strace, preexec, preload_libfault, nowait)
+        end_time = time.perf_counter()
+        execution_time = end_time - start_time
+        print(f"Execution time: {execution_time:.6f} seconds")
 
         if nowait:
             os.close(status_fds[1])
diff --git a/test/zdtm/static/maps04.c b/test/zdtm/static/maps04.c
index dbbb5c644029..ec58091d5dd3 100644
--- a/test/zdtm/static/maps04.c
+++ b/test/zdtm/static/maps04.c
@@ -11,7 +11,7 @@
 #include <linux/limits.h>
 #include "zdtmtst.h"
 
-#define MEM_SIZE (1L << 29)
+#define MEM_SIZE (8UL << 30)
 
 const char *test_doc = "Test big mappings";
 const char *test_author = "Andrew Vagin <avagin@openvz.org";
@@ -19,8 +19,7 @@ const char *test_author = "Andrew Vagin <avagin@openvz.org";
 {
        void *m;
-       uint32_t crc;
-       int i;
+       long i;
 
        test_init(argc, argv);
 
@@ -30,27 +29,15 @@ int main(int argc, char **argv)
                fail();
                return 1;
        }
+       madvise(m ,MEM_SIZE, MADV_POPULATE_WRITE);
 
-       crc = ~0;
-       datagen(m, MEM_SIZE, &crc);
-
-       for (i = 0; i < MEM_SIZE / (1 << 20); i++)
-               if (mprotect(m + (lrand48() * PAGE_SIZE % MEM_SIZE), PAGE_SIZE, PROT_NONE)) {
-                       pr_perror("mprotect");
-                       return 1;
-               }
+       for (i = 0; i < MEM_SIZE; i += 4096)
+               *(char*)(m + i) = i;
 
        test_daemon();
        test_waitsig();
 
-       if (mprotect(m, MEM_SIZE, PROT_READ))
-               pr_perror("mprotect");
-
-       crc = ~0;
-       if (datachk(m, MEM_SIZE, &crc))
-               fail("Mem corrupted");
-       else
-               pass();
+       pass();
 
        return 0;
 }
root@criu-next-2:/home/avagin/git/criu# python3 test/zdtm.py run -t zdtm/static/maps04 -f h
userns is supported
=== Run 1/1 ================ zdtm/static/maps04
========================= Run zdtm/static/maps04 in h ==========================
Start test
./maps04 --pidfile=maps04.pid --outfile=maps04.out
Run criu dump
Execution time: 4.970804 seconds
Run criu restore
==== ALARM ====
Execution time: 55.752694 seconds
Send the 15 signal to  66
Wait for zdtm/static/maps04(66) to die for 0.100000
Wait for zdtm/static/maps04(66) to die for 0.200000
Removing dump/zdtm/static/maps04/66
========================= Test zdtm/static/maps04 PASS =========================
```

The same test on the master branch:
```
root@criu-next-2:/home/avagin/git/criu# python3 test/zdtm.py run -t zdtm/static/maps04 -f h
userns is supported
=== Run 1/1 ================ zdtm/static/maps04
========================= Run zdtm/static/maps04 in h ==========================
 DEP       maps04.d
 CC        maps04.o
 LINK      maps04
Start test
./maps04 --pidfile=maps04.pid --outfile=maps04.out
Run criu dump
Execution time: 5.021624 seconds
Run criu restore
==== ALARM ====
Execution time: 3.668980 seconds
Send the 15 signal to  100
Wait for zdtm/static/maps04(100) to die for 0.100000
Wait for zdtm/static/maps04(100) to die for 0.200000
Removing dump/zdtm/static/maps04/100
========================= Test zdtm/static/maps04 PASS =========================
```

PR #3024, opened 2026-05-20
Envvar handling fixups

tursulin:
Follow up fixes/cleanups after 1961c5e0de7e ("plugins/amdgpu: Cleanup env variable handling")

+ @avagin


avagin:
Merged. Thanks a lot.

PR #3026, opened 2026-05-25
sk-queue: port SCM_CREDENTIALS checkpoint/restore support from OpenVZ

3idey:
Unix domain sockets can carry SCM_CREDENTIALS ancillary messages via
sendmsg(). Previously CRIU could not checkpoint/restore processes that had
such messages queued in their socket buffers. This series ports the
SCM_CREDENTIALS support from the OpenVZ CRIU fork to upstream.

The commits are cherry-picked one by one from OpenVZ (with `git cherry-pick
-x`, so the original commit hashes are preserved in the messages for easy
comparison). Original authorship and Signed-off-by tags are kept intact, and
my Signed-off-by is added on top.

Tests:

  * zdtm/static/sk-unix-cred: verifies SCM_CREDENTIALS are saved and
    restored.
  * zdtm/static/scm07: SCM_RIGHTS & SCM_CREDENTIALS in one message.
  * zdtm/static/scm08: several SCM_RIGHTS and SCM_CREDENTIALS messages.
  * zdtm/static/scm10: scm sent over a unix socket with a closed sender.

This is part of Google Summer of Code 2026 work on adding SCM_CREDENTIALS
and SCM_PIDFD checkpoint/restore support to CRIU.

Source: https://bitbucket.org/openvz/criu.ovz/


mihalicyn (changes requested):
Hi @3idey,

when you porting somebody else's code you can't just drop their authorship and Signed-off-by tags ;-)

You should use `git cherry-pick` to make a perfect copy of the original commit or `git format-patch + git am`.

What you are expected to do here:
- identify needed commits from OpenVZ CRIU fork
- cherry-pick them one by one
- add your Signed-off-by tag in addition to the original ones

to make it even better, you can use `git cherry-pick -x` to keep original commit hash so we can compare it with OpenVZ fork easier.

3idey:
Hi @mihalicyn, 
there is still some work left to port, including the tests and a few bug fixes from these commits. I will continue working on this as soon as I am free. :)

3idey:
All commits is cherry-picked with the `-x` is there anything else i should do ?
and all tests passes but scm10 is failing, will look into it.

avagin:
please clean up openvz specific parts from commit messages (subject prefixes, jira links and etc)

criu/include/sk-queue.h
```diff
@@ -2,7 +2,12 @@
 #define __CR_SK_QUEUE_H__
 
 extern struct collect_image_info sk_queues_cinfo;
-extern int dump_sk_queue(int sock_fd, int sock_id);
+
+#define SK_QUEUE_REAL_PID      0x1 /* scm creds contains a real pid */
+#define SK_QUEUE_DUMP_ADDR 0x2 /* save a sender address for messages */
+#define SK_QUEUE_TRACK_ENOBUFS 0x4 /* don't fail at ENOBUFS, fill *val if it is present */
```
mihalicyn: `SK_QUEUE_TRACK_ENOBUFS` likely comes from "netlink: save and generate ENOBUFS at restore" (commit hash 19f90ad44eeb918d9758d8724c0a2ab7ce53e98c) and it requires a custom kernel extension. As this code is not a part of upstream kernel, we can't really port it. I would just drop everything related to that commit. As far as I see you bring that extra code as a part of "netlink: dump and restore ucred-s for netlink messages" so you can just do `git rebase -i HEAD~18`, then `edit` on that commit and then carefully remove everything related to "netlink: save and generate ENOBUFS at restore".

criu/include/sk-queue.h
```diff
@@ -2,7 +2,12 @@
 #define __CR_SK_QUEUE_H__
 
 extern struct collect_image_info sk_queues_cinfo;
-extern int dump_sk_queue(int sock_fd, int sock_id);
+
+#define SK_QUEUE_REAL_PID      0x1 /* scm creds contains a real pid */
+#define SK_QUEUE_DUMP_ADDR 0x2 /* save a sender address for messages */
```
mihalicyn: `SK_QUEUE_DUMP_ADDR` seems to be a part of "sk-queue: use flags in dump_sk_queue" (commit hash 8b83405614703211fee27f90f08450cba080a44c). Instead of just glue-ing it together with "netlink: dump and restore ucred-s for netlink messages" you should just backport cherry-pick that commit first if you want to have this chunk.

criu/sk-queue.c
```diff
@@ -155,29 +382,139 @@ static int dump_packet_cmsg(struct msghdr *mh, SkPacketEntry *pe)
...
+
+				ret |= dump_sk_creds(ucred, pe, flags);
+				if (ret < 0)
+					return -1;
+				continue;
+			} else if (ch->cmsg_type == SCM_TIMESTAMP ||
+				   ch->cmsg_type == SCM_TIMESTAMPNS ||
+				   ch->cmsg_type == SCM_TIMESTAMPING) {
+				/*
+				 * Allow to receive timestamps from the kernel.
```
mihalicyn, line 388: this chunk originally was a part of "dump_packet_cmsg -- Allow to receive timestamp data" (commit hash 1ecd757bf23831d1574e1dca0566dbe26e0fcb49) you have it as a part of "sk-queue: Add support of ucreds in unix sockets". You can consider it as a fix and just glue, but then you need to add this to the commit description. Something like this:

```
sk-queue: Add support of ucreds in unix sockets
Checkpoint and restore ucreds on unix sockets are not
<...CUT..>

(cherry picked from commit 8f3061c360381894a97edca2ada85466f95e1d88)
Ahmed Elaidy:
- cherry-picked
- combined with 1ecd757bf238 ("dump_packet_cmsg -- Allow to receive timestamp data")
Signed-off-by: Ahmed Elaidy <elaidya225@gmail.com>
```

so this will be visible.

mihalicyn: IMHO, it is simpler to just port "dump_packet_cmsg -- Allow to receive timestamp data" as a fix on top... you have `git rebase -i ..` and it is a pretty powerful tool.

mihalicyn (changes requested):
Looks almost finished. I left a few comments about what we have to fix.

3idey:
Hi @mihalicyn,

I've reworked the bottom of the series to
address all three points. The branch is now 20 commits (was 18).

1. ENOBUFS dropped
   SK_QUEUE_TRACK_ENOBUFS and everything tied to it (the int *val
   out-parameter, the `errno == ENOBUFS` handling, the header flag) is gone.
   It originated from 19f90ad44 ("netlink: save and generate ENOBUFS at
   restore"), which needs a non-upstream kernel extension, so it isn't ported.
   The removal touches the netlink-ucred commit and is also handled when
   replaying the later "sk-queue: fix dumping queue packets" refactor. Each
   affected commit notes the drop.

2. SK_QUEUE_DUMP_ADDR split out
   It now comes from its own commit, cherry-picked (-x) from 8b834056
   ("sk-queue: use flags in dump_sk_queue"), placed first.

   One adaptation worth flagging: upstream has no netlink queue dump, so
   upstream dump_sk_queue() is the 2-arg form and there was no `bool dump_addr`
   argument to convert into `int flags`. I therefore introduced `int flags` on
   the 2-arg base and updated the only caller (unix sockets); there's no
   sk-netlink.c hunk for the same reason. I documented this under the
   "Ahmed Elaidy:" note in that commit. If you'd rather I keep the chunk 1:1
   with the fork instead, that would mean pulling in the netlink queue-dump
   prerequisites too - happy to go either way, just let me know.

3. Timestamp cmsg handling split out
   The SCM_TIMESTAMP / SCM_TIMESTAMPNS / SCM_TIMESTAMPING pass-through in
   dump_packet_cmsg() is now its own commit, cherry-picked (-x) from 1ecd757b
   ("dump_packet_cmsg -- Allow to receive timestamp data"), and removed from
   the unix-ucreds commit. Splitting cleanly was feasible, so I went with the
   preferred option rather than the documented-combine fallback.

New commit layout (bottom -> top of the series):

  sk-queue: use flags in dump_sk_queue                   (Pavel, -x 8b834056)
  netlink: dump and restore ucred-s for netlink messages (Andrei, -x 66f8f7b2a)
  dump_packet_cmsg -- Allow to receive timestamp data    (Cyrill, -x 1ecd757b)
  sk-queue: Add support of ucreds in unix sockets        (Cyrill, -x 8f3061c3)
  ... tests + the remaining fix commits, unchanged ...

Verification:
- the net code change versus the previous version is exactly the ENOBUFS
  removal (DUMP_ADDR is preserved, the timestamp code is only relocated);
- build is clean and every commit builds on its own.
- sk-unix-cred, scm07, scm08 and scm10 pass locally.

Please take another look, happy to adjust the DUMP_ADDR handling if you'd
prefer the strict backport.

images/sk-packet.proto
```diff
@@ -7,12 +7,16 @@ message scm_entry {
...
+	required uint32		gid		= 2;
+	required uint32		pid		= 3;
+}
+
 message sk_packet_entry {
 	required uint32		id_for		= 1;
 	required uint32		length		= 2;
-	// Reserved for message address
-	// optional bytes		addr	= 3;
+	optional bytes		addr		= 3;
```
mihalicyn: This field is a part of netlink socket packet queue checkpoint/restore feature (see https://bitbucket.org/openvz/criu.ovz/commits/e9acd2df772751ce243cea797fed63a652e7b715 and https://bitbucket.org/openvz/criu.ovz/commits/895dfb5e21e679544f63f2d8c6cbdd107850c06d). As far as I understand we don't port this feature, because upstream Linux kernel does not support `setsockopt(sk, SOL_NETLINK, NETLINK_REPAIR, ...)`.

criu/sk-queue.c
```diff
@@ -132,10 +155,214 @@ static int dump_scm_rights(struct cmsghdr *ch, SkPacketEntry *pe)
...
+	sk_packet_entry__init(pe);
+	sk_ucred_entry__init(ue);
+
+	pkt->entry = pe;
+	pkt->data_off = p - (void *)pkt;
+
+	pe->id_for = entry->id_for;
+	pe->length = entry->length;
+	pe->has_addr = entry->has_addr;
+	if (entry->has_addr) {
```
mihalicyn: If we remove `sk_packet_entry.addr` field this should go away, right?

criu/sk-queue.c
```diff
@@ -155,29 +382,133 @@ static int dump_packet_cmsg(struct msghdr *mh, SkPacketEntry *pe)
...
+		 */
+		pr_err("recvmsg failed: truncated\n");
+		return -1;
+	}
+
+	ret = dump_packet_cmsg(&msg, &pe, flags);
+	if (ret < 0)
+		goto cleanup_packet;
+
+	if (flags & SK_QUEUE_DUMP_ADDR && msg.msg_namelen) {
```
mihalicyn: This is a dead code, because again, it is for netlink sockets and we don't support queue C/R for netlink sockets.

criu/sk-queue.c
```diff
@@ -330,6 +629,38 @@ static int send_one_pkt(int fd, struct sk_packet *pkt)
 	 * boundaries messages should be saved.
 	 */
 
+	if (entry->has_addr) {
```
mihalicyn: dead code

mihalicyn (changes requested):

criu/sk-queue.c
```diff
@@ -336,9 +459,28 @@ int dump_sk_queue(int sock_fd, int sock_id, int flags)
...
+			pe.addr.data = addr;
+			pe.addr.len = msg.msg_namelen;
+		}
+
+		if (ret > 0) {
+			if (ret == 1) {
+				if (queue_packet_entry(&pe, data, pe.length))
+					goto err_set_sock;
+			}
+			if (pe.scm)
```
mihalicyn: Hmm, I'm comparing this with the original commit https://bitbucket.org/openvz/criu.ovz/commits/8f3061c360381894a97edca2ada85466f95e1d88. Why this chunk of code around `queue_packet_entry(..)` is so different? It is not necessarily wrong, but I'm just trying to understand where this comes from.

mihalicyn (changes requested):

criu/sk-queue.c
3idey: This chunk is different because it is not a straight copy of the original OpenVZ logic. I split it so SCM_CREDENTIALS packets with a real pid can be deferred until all tasks are collected and the pid can be rewritten to the virtual pid. That is why `queue_packet_entry()` stores the packet in memory and `sk_queue_post_actions()` writes it later. The plain queue path is still the same; only the deferred SCM_CREDENTIALS path is new.

images/sk-packet.proto
3idey: Yes, i will rebase and apply all of these and remove any remaining dead code related to them.

3idey:
Hi @mihalicyn,

Pushed an update addressing the netlink-dead-code feedback, and while I was in
there I cleaned up the history a bit more:

- Removed the last of the netlink queue-C/R dead code. On top of the
  sk_packet_entry.addr field, the SOL_NETLINK / NETLINK_PKTINFO handling and
  the two <linux/socket.h> / <linux/netlink.h> includes that came with it are
  gone. Since upstream doesn't support queue C/R for netlink sockets, none of
  that ever ran on the unix path.

- Dropped the "netlink: skip pktinfo scm if a sender group is less than 32"
  commit entirely - it's purely netlink-queue and shouldn't be part of this
  series. That also means the two commits it was tangled up with ("netlink:
  dump and restore ucred-s" and "sk-queue: fix dumping queue packets") are
  faithful cherry-picks again, so a git range-diff against the fork should
  line up cleanly now. The series is 20 commits.

- Rebased the whole branch onto current criu-dev to pick up the CI/infra fixes
  (the old base was ~130 commits behind;) ).

And yeah - untangling the netlink bits back out of a fork port was not the fun
work indeed :) but the history is clean now. Ready for another look.

criu/sk-queue.c
3idey: Correction to my earlier reply - the deferral design (queue_packet_entry() +
sk_queue_post_actions()) is not mine, it's from the original 8f3061c360.

What i did is:

- leak cleanup (release_cmsg()/xfree(pe.ucred)) on the deferred path, added
  while fixing the CodeQL findings - the original leaks there until
  6832ccd49, so this front-loads part of that later fix;
- has_addr/_K_SS_MAXSIZE removed, per your netlink dead-code review;
- cosmetic (void *) -> (char *) cast.

Happy to restructure if you'd prefer a strict copy with the leak
fix moved to where the fork does it; otherwise I'll document the deltas in
the commit message:)

mihalicyn:
@3idey [969a2aba0331ff940501205077766859453b9a35](https://bitbucket.org/openvz/criu.ovz/commits/969a2aba0331ff940501205077766859453b9a35) was missing and this caused more conflicts for you during rebase ;-)

I have added it and rebased everything properly. Made one force-push to see how what CI thinks about all this.

3idey:
Thanks for adding the missing commit and rebasing everything properly, @mihalicyn!

Could you re-trigger the compat-test (CLANG) job? other failing jobs seems unrelated to me:)

mihalicyn:
```
2026-07-14T08:14:19.1300224Z Removing dump/zdtm/static/fpu00/322
2026-07-14T08:14:19.1300343Z ========================= Test zdtm/static/fpu00 PASS ==========================
2026-07-14T08:14:19.1300351Z 
2026-07-14T08:14:19.1300453Z === Run 93/495 ===------------- zdtm/static/fpu01
2026-07-14T08:14:19.1300543Z === Run 94/495 ===------------- zdtm/static/fpu02
2026-07-14T08:14:19.1300668Z ========================== Run zdtm/static/fpu01 in h ==========================
2026-07-14T08:14:19.1300731Z Start test
2026-07-14T08:14:19.1300842Z ./fpu01 --pidfile=fpu01.pid --outfile=fpu01.out
2026-07-14T08:14:19.1300906Z Run criu dump
2026-07-14T08:14:19.1300977Z Run criu restore
2026-07-14T08:14:19.1301054Z Send the 15 signal to  66
2026-07-14T08:14:19.1301160Z Wait for zdtm/static/fpu01(66) to die for 0.100000
2026-07-14T08:14:19.1301295Z ################# Test zdtm/static/fpu01 FAIL at result check ##################
2026-07-14T08:14:19.1301395Z Test output: ================================
2026-07-14T08:14:19.1301797Z 08:14:11.771:    66: Data mismatch ('code9Ee5sohphie1ae1kaeMahngoh5oe' 'Tacuthahhien9Fi7aGhaa5toGh6vi7Ch' 'code9Ee5sohphie1' 'Tacuthahhien9Fi7')
2026-07-14T08:14:19.1302009Z 08:14:11.771:    66: FAIL: fpu01.c:91:  (errno = 11 (Resource temporarily unavailable))
2026-07-14T08:14:19.1302014Z 
2026-07-14T08:14:19.1302110Z  <<< ================================
2026-07-14T08:14:19.1302113Z 
2026-07-14T08:14:19.1302210Z === Run 95/495 ===------------- zdtm/static/fpu03
2026-07-14T08:14:19.1302348Z ========================== Run zdtm/static/fpu02 in h ==========================
2026-07-14T08:14:19.1302421Z Start test
```

unrelated failure of [CI / nftables-test (pull_request)](https://github.com/checkpoint-restore/criu/actions/runs/29316453423/job/87033728774?pr=3026), but worth looking into it separately

mihalicyn (approved):
Good job, @3idey 

I did a few tiny changes:
- added one more commit [sk-unix: allocate FdinfoEntry with shmalloc in get_fle_for_task](https://github.com/checkpoint-restore/criu/pull/3026/changes/9f44d6648f0ee53590f60d45f79824d569a065b3)
- restored the original order of the patches

LGTM

Issue #2720, opened 2025-08-22
CRIU hang at the final steps of restoration on aarch64 architecture


juranvrsk:
**Problem Description**
During restoration CRIU stopped in infinite waiting at the first `wait4` of `compel_stop_on_syscall` function. The processes and threads states:
```
    TID    PPID S COMMAND         CMD                                                                                                                                                                                                                                       
 105409  105403 t payload         /payload                                                                                                                                                                                                                                  
 105436  105403 t m_1             /payload                                                                                                                                                                                                                                  
 105437  105403 t m_2             /payload                                                                                                                                                                                                                                  
 105438  105403 S m_3             /payload                                                                                                                                                                                                                                  
 105427  105409 t payload         /payload                                                                                                                                                                                                                                  
 105429  105409 t b_1             /payload                                                                                                                                                                                                                                  
 105430  105409 t b_2             /payload                                                                                                                                                                                                                                  
 105431  105409 t b_3             /payload                                                                                                                                                                                                                                  
 105428  105409 t payload         /payload                                                                                                                                                                                                                                  
 105432  105409 t a_1             /payload                                                                                                                                                                                                                                  
 105433  105409 t a_2             /payload                                                                                                                                                                                                                                  
 105434  105409 t a_3             /payload 
```
Some tasks (it can be any kind of task), are not interrupted by `ptrace`, it’s state is “S”. The CRIU restore log:
```
(00.079924) 105434 was trapped                                                                                                                                                                                                                                              
(00.079966) 105433 was trapped                                                                                                                                                                                                                                              
(00.079998) 105434 was trapped                                                                                                                                                                                                                                              
(00.080005) 105434 (native) is going to execute the syscall 139, required is 139                                                                                                                                                                                            
(00.080034) 105434 was stopped                                                                                                                                                                                                                                              
(00.080039) 105433 was trapped                                                                                                                                                                                                                                              
(00.080044) 105433 (native) is going to execute the syscall 139, required is 139                                                                                                                                                                                            
(00.080084) 105433 was stopped                                                                                                                                                                                                                                              
(00.080109) 105432 was trapped                                                                                                                                                                                                                                              
(00.080141) 105431 was trapped                                                                                                                                                                                                                                              
(00.080170) 105432 was trapped                                                                                                                                                                                                                                              
(00.080177) 105432 (native) is going to execute the syscall 139, required is 139                                                                                                                                                                                            
[...]

Issue #2658, opened 2025-05-02
Pre-dump stuck on AMD SEV-SNP (TEE environment)


micmeier:
Hi,

I'm trying to run a pre-dump on a Trusted Execution Environment (TEE) with AMD SEV-SNP enabled. The pre-dump command gets stuck without showing any errors. Running the exact same setup and commands works fine outside the SEV-SNP environment.

**Setup**

- AMD server with SEV-SNP enabled
- Target process is a simple Docker container running nginx
- Default CRIU configuration with empty default.conf

**Command**
`sudo criu pre-dump -t 4456 --images-dir /tmp/criu-dump/pre-dump/ -v5 -o /tmp/criu-dump/pre-dump/pre-dump2.log`

**Logs**

> (00.014500) ----------------------------------------
> (00.014523) Sent msg to daemon 65 0 0
> (00.014528) Wait for ack 65 on daemon socket
> pie: 1: __fetched msg: 65 0 0
> pie: 1: __sent ack msg: 65 65 0
> pie: 1: Daemon waits for command
> (00.014621) Fetched ack: 65 65 0
> (00.014646) Waiting for 4456 to trap
> (00.014690) Daemon 4456 exited trapping
> (00.014709) Sent msg to daemon 3 0 0
> pie: 1: __fetched msg: 3 0 0
> pie: 1: 1: new_sp=0x7222b017cac8 ip 0x7222afa5b2d3

No other log is shown after this. Full log file is attached.

**Questions**
- Has CRIU been tested on TEE environments (specifically AMD SEV-SNP)? If yes, is there any special configuration or workaround needed
- Any known reasons why CRIU pre-dump might have this behaviour?

Thanks!

[pre-dump2.log](https://github.com/user-attachments/files/20016009/pre-dump2.log)

mihalicyn:
Hi!

> Trusted Execution Environment (TEE) with AMD SEV-SNP enabled

So, it's a VM, right? So you basically running a confidential VM with docker container inside, and when you are trying to checkpoint that container criu get stuck, right?

>Any known reasons why CRIU pre-dump might have this behaviour?

Is it a problem only with pre-dump? Does `dump` work?

Have you tried to build CRIU from source code and run some tests?
I would do something like:
```
git clone https://github.com/checkpoint-restore/criu.git
[...]

PR #3028, opened 2026-05-26
github: add linux-next test workflow

mihalicyn:
```
#
# This workflow is to test CRIU on linux-next tree.
# It involves external GitHub Actions runners (EC2 instances on AWS).
# To run it requires necessary setup on AWS side (auth via GitHub OIDC to
# manage EC2 instances and S3 bucket), and GH_RUNNERS_PAT_TOKEN
# (GitHub Personal Access Token to manage GitHub Actions Runners for a repo).
#
# Logic is simple:
# 1. Build linux kernel on GitHub-hosted runner and upload .deb-packages
#    to S3 bucket
# 2. Create EC2 instance and register it as GH self-hosted runner
#    (we use machulav/ec2-github-runner for this)
# 3. Schedule a job on newly created runner, install a new kernel
# 4. Reboot EC2 instance
# 5. Schedule CRIU test job on EC2 runner
# 6. Destroy EC2 instance and unregister it from GitHub.
#
# I use explicit commit hashes for actions steps which are not GitHub-provided
# (for example machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef).
# This is to be on a safe side and ensure that any update is done manually.
#
```

Example of result: https://github.com/checkpoint-restore/criu/actions/runs/26462306037

### DIFF
--- a/.github/workflows/linux-next.yml
+++ b/.github/workflows/linux-next.yml
@@ -1,0 +1,401 @@
+name: linux-next-tests
+on:
+  schedule:
+    - cron: "20 8 * * 0"
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+#
+# This workflow is to test CRIU on linux-next tree.
+# It involves external GitHub Actions runners (EC2 instances on AWS).
+# To run it requires necessary setup on AWS side (auth via GitHub OIDC to
+# manage EC2 instances and S3 bucket), and GH_RUNNERS_PAT_TOKEN
+# (GitHub Personal Access Token to manage GitHub Actions Runners for a repo).
+#
+# Logic is simple:
+# 1. Build linux kernel on GitHub-hosted runner and upload .deb-packages
+#    to S3 bucket
+# 2. Create EC2 instance and register it as GH self-hosted runner
+#    (we use machulav/ec2-github-runner for this)
+# 3. Schedule a job on newly created runner, install a new kernel
+# 4. Reboot EC2 instance
+# 5. Schedule CRIU test job on EC2 runner
+# 6. Destroy EC2 instance and unregister it from GitHub.
+#
+# I use explicit commit hashes for actions steps which are not GitHub-provided
+# (for example machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef).
+# This is to be on a safe side and ensure that any update is done manually.
+#
+
+permissions:
+  id-token: write
+
+jobs:
+  kernel-build:
+    env:
+      BUILDDIR: "/home/runner/kernel-build/tmp"
+    name: Kernel build
+    runs-on: ubuntu-24.04
+    outputs:
+      date: ${{ steps.get-date.outputs.date }}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Prepare directory for build
+        run: |
+          set -eux
+
+          echo $BUILDDIR
+          mkdir -p $BUILDDIR
+          # Not enough RAM for kernel builds anymore
+          # sudo mount -t tmpfs tmpfs $BUILDDIR
+
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+
+      - name: Restore linux-next build artifacts from cache
+        id: build-artifacts-restore
+        uses: actions/cache@v5
+        with:
+          path: ${{ env.BUILDDIR }}/artifacts
+          key: linux-next-build-${{ steps.get-date.outputs.date }}
+
+      - name: Install dependencies
+        if: steps.build-artifacts-restore.outputs.cache-hit != 'true'
+        run: |
+          set -eux
+
+          sudo apt-get update
+
+          sudo apt-get install --no-install-recommends -y \
+            curl \
+            git \
+            build-essential \
+            libssl-dev \
+            libelf-dev \
+            libdw-dev \
+            bc \
+            bison \
+            cpio \
+            flex \
+            debhelper-compat
+
+      # We don't want to put too much stress to git.kernel.org so we
+      # use a cache (valid for 1 day) for git clone copy of the repo.
+      - name: Restore linux-next git cache
+        if: steps.build-artifacts-restore.outputs.cache-hit != 'true'
+        id: linux-next-git-restore
+        uses: actions/cache@v5
+        with:
+          path: ${{ env.BUILDDIR }}/linux
+          key: linux-next-git-cache-${{ steps.get-date.outputs.date }}
+
+      - name: Checkout Linux kernel
+        if: ${{ steps.build-artifacts-restore.outputs.cache-hit != 'true' &&
+                steps.linux-next-git-restore.outputs.cache-hit != 'true' }}
+        run: |
+          set -eux
+
+          cd "$BUILDDIR"
+
+          git clone --depth=1 --branch master --single-branch \
+                    https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git linux
+
+          # Record exactly which linux-next state we built, so a CI failure
+          # can be correlated with a specific tree (and its daily next-* tag).
+          git -C linux --no-pager log -1 --format='linux-next HEAD: %H %cs %s'
+
+      - name: Save linux-next git cache
+        if: ${{ steps.build-artifacts-restore.outputs.cache-hit != 'true' &&
+                steps.linux-next-git-restore.outputs.cache-hit != 'true' }}
+        uses: actions/cache/save@v5
+        with:
+          path: |
+            ${{ env.BUILDDIR }}/linux
+          key: linux-next-git-cache-${{ steps.get-date.outputs.date }}
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@254c19bd240aabef8777f48595e9d2d7b972184b # 6.2.1
+        with:
+          role-to-assume: ${{ vars.AWS_IAM_ROLE }}
+          aws-region: ${{ vars.AWS_REGION }}
+
+      - name: Build Linux kernel
+        if: steps.build-artifacts-restore.outputs.cache-hit != 'true'
+        run: |
+          set -eux
+
+          cd "$BUILDDIR/linux"
+
+          # kernel config was generated with make localmodconfig on EC2 instance
+          # after full CRIU tests run (to trigger all modules load)
+          aws s3 cp \
+            s3://criu-linux-next-ci-634567146514-eu-central-1-an/t3.xlarge-kernel-config \
+            .config
+
+          scripts/config --disable SYSTEM_TRUSTED_KEYS
+          scripts/config --disable SYSTEM_REVOCATION_KEYS
+
+          make olddefconfig
+
+          time make -j$(nproc) bindeb-pkg LOCALVERSION="-criu-ci"
+
+          ls -la .
+          ls -la ..
+
+          cd "$BUILDDIR"
+          mkdir artifacts/
+          mv linux-*.deb artifacts/
+
+          echo "artifactPath=${BUILDDIR}/artifacts" >> $GITHUB_ENV
+
+      - name: Save linux-next build results cache
+        if: steps.build-artifacts-restore.outputs.cache-hit != 'true'
+        uses: actions/cache/save@v5
+        with:
+          path: ${{ env.artifactPath }}
+          key: linux-next-build-${{ steps.get-date.outputs.date }}
+
+      - name: Upload kernel packages to S3
+        run: |
+          set -eux
+
+          cd "$BUILDDIR"
+
+          BUILD_DATE="${{ steps.get-date.outputs.date }}"
+
+          aws s3 cp \
+            artifacts \
+            s3://criu-linux-next-ci-634567146514-eu-central-1-an/artifacts/$BUILD_DATE/ \
+            --recursive
+
+      - uses: actions/upload-artifact@v7
+        with:
+          name: linux-kernel-build
+          path: ${{ env.BUILDDIR }}/artifacts/linux-*.deb
+          if-no-files-found: error
+          retention-days: 14
+
+  start-runner:
+    name: Start self-hosted EC2 runner
+    needs:
+      - kernel-build
+    runs-on: ubuntu-latest
+    outputs:
+      label: ${{ steps.start-ec2-runner.outputs.label }}
+      ec2-instance-id: ${{ steps.start-ec2-runner.outputs.ec2-instance-id }}
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@254c19bd240aabef8777f48595e9d2d7b972184b # 6.2.1
+        with:
+          role-to-assume: ${{ vars.AWS_IAM_ROLE }}
+          aws-region: ${{ vars.AWS_REGION }}
+
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef # 2.6.1
+        with:
+          mode: start
+          startup-timeout-minutes: 10
+          github-token: ${{ secrets.GH_RUNNERS_PAT_TOKEN }}
+          ec2-image-id: ami-0596cf3199908321b # Ubuntu 24.04 LTS image id on AWS
+          ec2-instance-type: t3.xlarge
+          subnet-id: subnet-0ddb356c3fc41e51a
+          security-group-id: sg-054aa948984162822
+          packages: '["git", "docker.io"]'
+          runner-debug: true
+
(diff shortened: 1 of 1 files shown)


adrianreber:
We already have Linux next testing with one of the Fedora runners.

Not sure how the self hosted runners are financed, but you can easily start a VM on GitHub Actions thanks to nested virtualization. Sounds a bit simpler than using a self hosted runner.

Just as a point of view what is possible. Nothing from my side against having a self hosted runner.

mihalicyn:
> Not sure how the self hosted runners are financed, but you can easily start a VM on GitHub Actions thanks to nested virtualization. Sounds a bit simpler than using a self hosted runner.

I've got https://aws.amazon.com/blogs/opensource/aws-cloud-credits-for-open-source-projects-affirming-our-commitment/ for CRIU organization

> We already have Linux next testing with one of the Fedora runners.

That's good, it is something I missed, sorry.

adrianreber:
> > Not sure how the self hosted runners are financed, but you can easily start a VM on GitHub Actions thanks to nested virtualization. Sounds a bit simpler than using a self hosted runner.
> 
> I've got https://aws.amazon.com/blogs/opensource/aws-cloud-credits-for-open-source-projects-affirming-our-commitment/ for CRIU organization

Cool. Great. Thanks.


> > We already have Linux next testing with one of the Fedora runners.
> 
> That's good, it is something I missed, sorry.

It is not as up to date as checking out latest git, but reasonably up to date. Seems like almost every rc release gets a build.

mihalicyn:
> It is not as up to date as checking out latest git, but reasonably up to date. Seems like almost every rc release gets a build.

Yeah, that's super cool to have it on native runners!

I believe we can then consider having some other stuff like extra arm64 tests on AWS (there were a few issues in the past with CRIU on AWS arm64 machines, like https://github.com/checkpoint-restore/criu/issues/2720 for example. Also, get some tests on confidential VMs too as we've seen issues like https://github.com/checkpoint-restore/criu/issues/2658.

adrianreber:
arm runners sound like a good idea. The arm VMs on GitHub Actions do not support nested virtualization, so it is much more limited there.

Currently the Linux next test is using following kernel 

```
[    0.000000] Linux version 7.1.0-0.0.next.20260526.235.vanilla.fc43.x86_64 (mockbuild@778fdeeddab748928d9aab7cdde77e70) (gcc (GCC) 15.2.1 20260123 (Red Hat 15.2.1-7), GNU ld version 2.45.1-4.fc43) #1 SMP PREEMPT_DYNAMIC Tue May 26 16:37:47 UTC 2026
```

But I definitely don't want to block this PR. Just wanted to highlight that we have something similar.

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,357 @@
...
+
+      - name: Start EC2 runner
+        id: start-ec2-runner
+        uses: machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef # 2.6.1
+        with:
+          mode: start
+          startup-timeout-minutes: 10
+          github-token: ${{ secrets.GH_RUNNERS_PAT_TOKEN }}
+          ec2-image-id: ami-0596cf3199908321b # Ubuntu 24.04 LTS image id on AWS
+          ec2-instance-type: t2.xlarge
```
rst0git: @mihalicyn Would it be possible to specify an instance with NVIDIA GPU and install the latest GPU driver?
- https://docs.nvidia.com/datacenter/tesla/driver-installation-guide/ubuntu.html
- https://www.nvidia.com/en-gb/drivers/unix/

I'm testing the CUDA Plugin locally, but it would be great if we have CI tests for it as well.

mihalicyn: > Would it be possible to specify an instance with NVIDIA GPU and install the latest GPU driver?

good question, I don't know for sure. But if AWS has EC2 instances with NVIDIA GPUs then yes. Basically, you can specify any ec2 instance type in there.

mihalicyn: > I'm testing the CUDA Plugin locally, but it would be great if we have CI tests for it as well.

I can give you an access to AWS if you want ;-)

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,357 @@
...
+
+permissions:
+  id-token: write
+
+jobs:
+  kernel-build:
+    env:
+      TMPFSDIR: "/home/runner/kernel-build/tmpfs"
+    name: Kernel build
+    runs-on: ubuntu-24.04
```
avagin, line 41: nit: should we use the last lts release?

avagin:
overall, it looks good. Thank you for working on that. I think we can to do a few extra thing:
* save serial console logs for each run. In ideal case, we need to get them with aws tools in case of unexpected kernel panic or something like that. 
* check that the right kernel has been booted. If the target kernel has failed to boot, a default kernel can be booted...
* we can do a few other sanity checks after running tests: check /proc/sys/kernel/tainted, chec that x86_fpu_xstate_check_failed hasn't been triggered, etc
* a kernel can be build with debug options and lockdep.

.github/workflows/linux-next.yml
mihalicyn: There is no Ubuntu 26.04 runners yet (https://github.com/actions/runner-images#available-images)

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,357 @@
...
+# 5. Schedule CRIU test job on EC2 runner
+# 6. Destroy EC2 instance and unregister it from GitHub.
+#
+# I use explicit commit hashes for actions steps which are not GitHub-provided
+# (for example machulav/ec2-github-runner@343a1b2ae682e681c3cec9a235d882da17ff04ef).
+# This is to be on a safe side and ensure that any update is done manually.
+#
+
+permissions:
+  id-token: write
```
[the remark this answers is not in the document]
mihalicyn: As a matter of fact, it doesn't break anything. This workflow actually works :) [ see https://github.com/checkpoint-restore/criu/actions/runs/26462306037 ]

mihalicyn:
> * save serial console logs for each run. In ideal case, we need to get them with aws tools in case of unexpected kernel panic or something like that.

Have fixed this. It turns our that not all EC2 instance types support serial console. AWS Nitro ones do.
So I switched `t2.xlarge` (Xen) to `t3.xlarge` (Nitro). Pricing is roughly the same.

> * check that the right kernel has been booted. If the target kernel has failed to boot, a default kernel can be booted...

this was done initially by `uname -a | grep next | grep "criu-ci"` check

> * we can do a few other sanity checks after running tests: check /proc/sys/kernel/tainted, chec that x86_fpu_xstate_check_failed hasn't been triggered, etc

have added this logic too, thanks for suggestion!

> * a kernel can be build with debug options and lockdep.

`CONFIG_LOCKDEP_SUPPORT` is enabled already. I'm not sure if we want to enable ASAN or something else.

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Import my ssh key (debug)
+        run: |
+          set -eux
+
+          id
+          ssh-import-id lp:mihalicyn
```
rst0git: Can we avoid importing an SSH key?
There are other ways to debug CI actions (e.g., https://github.com/mxschmitt/action-tmate)

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+  install-kernel:
+    name: Install kernel on the EC2 runner
+    needs:
+      - start-runner # required to start the main job when the runner is ready
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
+    steps:
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
```
rst0git: We can compute the build date once in `kernel-build` and exposing it as a job output for `install-kernel`:
```diff
diff --git a/.github/workflows/linux-next.yml b/.github/workflows/linux-next.yml
index aa3a7efb3..ccbe2091a 100644
--- a/.github/workflows/linux-next.yml
+++ b/.github/workflows/linux-next.yml
@@ -35,6 +35,8 @@ jobs:
        TMPFSDIR: "/home/runner/kernel-build/tmpfs"
      name: Kernel build
      runs-on: ubuntu-24.04
+    outputs:
+      date: ${{ steps.get-date.outputs.date }}

      steps:
        - name: Checkout
@@ -223,15 +225,10 @@ jobs:
    install-kernel:
      name: Install kernel on the EC2 runner
      needs:
+      - kernel-build # required to consume the build date output
        - start-runner # required to start the main job when the runner is ready
      runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
      steps:
-      - name: Get Date
-        id: get-date
-        run: |
-          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
-        shell: bash
-
        - name: Import my ssh key (debug)
          run: |
            set -eux
@@ -260,7 +257,7 @@ jobs:
          run: |
            set -eux

-          BUILD_DATE="${{ steps.get-date.outputs.date }}"
+          BUILD_DATE="${{ needs.kernel-build.outputs.date }}"

            aws s3 cp \
              s3://criu-linux-next-ci-634567146514-eu-central-1-an/artifacts/$BUILD_DATE/ \
```

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+          INSTANCE_ID="${{ needs.start-runner.outputs.ec2-instance-id }}"
+          aws ec2 reboot-instances --instance-ids "$INSTANCE_ID"
+          aws ec2 wait instance-status-ok --instance-ids "$INSTANCE_ID"
+
+  do-the-job:
+    name: Do the job on the runner
+    needs:
+      - start-runner
+      - reboot-runner # required to start the main job when the runner is ready
+    runs-on: ${{ needs.start-runner.outputs.label }} # run the job on the newly created runner
```
rst0git, line 295: Could we add a `timeout-minutes` here (e.g., [`timeout-minutes: 60`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax))? This job runs the full CRIU suite on a `t3.xlarge` self-hosted runner, and if a test ever hangs the EC2 instance keeps billing until GitHub's default 6-hour job timeout kicks in before `stop-runner` reaps it.

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
+name: linux-next-tests
+on:
+  schedule:
+    - cron: "20 8 * * 0"
+  workflow_dispatch:
```
rst0git, line 5: Could we add a concurrency group to the workflow file?

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+          time make -j$(nproc) bindeb-pkg LOCALVERSION="-criu-ci"
+
+          ls -la .
+          ls -la ..
+
+          cd $TMPFSDIR
+          mkdir artifacts/
+          mv linux-*.deb artifacts/
+
+          echo "artifactPath=${TMPFSDIR}/artifacts" >> $GITHUB_ENV
```
rst0git: It seems like `artifactPath` is not used.

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+        uses: actions/cache/save@v5
+        with:
+          path: |
+            ${{ env.TMPFSDIR }}/linux
+          key: linux-next-git-cache-${{ steps.get-date.outputs.date }}
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # 6.2.0
+        with:
+          role-to-assume: arn:aws:iam::634567146514:role/criutest
```
rst0git: Instead of hardcoding these values, can we use use [GitHub Actions variables](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables) (e.g., `vars.AWS_ACCOUNT_ID`) and reference them here?

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+        if: ${{ steps.build-artifacts-restore.outputs.cache-hit != 'true' &&
+                steps.linux-next-git-restore.outputs.cache-hit != 'true' }}
+        run: |
+          set -eux
+
+          GIT_REFSPEC="master"
+
+          cd $TMPFSDIR
+
+          git clone --depth=1 https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git linux
```
rst0git:
```diff
-          GIT_REFSPEC="master"
-
            cd $TMPFSDIR

-          git clone --depth=1 https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git linux
-          cd linux
-          git fetch --depth=1 origin ${GIT_REFSPEC}
-          git checkout ${GIT_REFSPEC}
+          git clone --depth=1 --branch master --single-branch \
+            https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git linux
+
+          # Record exactly which linux-next state we built, so a CI failure
+          # can be correlated with a specific tree (and its daily next-* tag).
+          git -C linux --no-pager log -1 --format='linux-next HEAD: %H %cs %s'
```

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+    if: ${{ always() }} # required to get logs even if the error happened in the previous jobs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@e7f100cf4c008499ea8adda475de1042d6975c7b # 6.2.0
+        with:
+          role-to-assume: arn:aws:iam::634567146514:role/criutest
+          aws-region: eu-central-1
+
+      - name: Install AWS CLI
```
rst0git: For `reboot-runner` and `get-serial-console`, AWS CLI appears to be installed by default:
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md#cli-tools

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+
+          echo $TMPFSDIR
+          mkdir -p $TMPFSDIR
+          sudo mount -t tmpfs tmpfs $TMPFSDIR
+
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y%m%d")" >> $GITHUB_OUTPUT
+        shell: bash
```
rst0git: `bash` is already the default shell on Linux runners

.github/workflows/linux-next.yml
```diff
@@ -0,0 +1,421 @@
...
+
+      - name: Checkout Linux kernel
+        if: ${{ steps.build-artifacts-restore.outputs.cache-hit != 'true' &&
+                steps.linux-next-git-restore.outputs.cache-hit != 'true' }}
+        run: |
+          set -eux
+
+          GIT_REFSPEC="master"
+
+          cd $TMPFSDIR
```
rst0git: `cd "$TMPFSDIR"`

.github/workflows/linux-next.yml
mihalicyn: Good idea, thanks! I'll add this in the next version.

Also, I have created an AWS Lambda that removes all GitHub-created instances older than 4 hours old.

.github/workflows/linux-next.yml
mihalicyn: Sure, no problem. I'll remove this in the final version. It was there mostly for debugging.

I know about tmate, but in general, once we have a full access to EC2 instance it is always more convenient to have full SSH access rather than using tricky ways with ssh reverse tunnels (this is what tmate does, I guess).

.github/workflows/linux-next.yml
mihalicyn: yeah, for some reason, it didn't work for me before. Maybe it is installed in `sbin`?.. Let me try.

.github/workflows/linux-next.yml
mihalicyn: Thanks, will be fixed.

.github/workflows/linux-next.yml
mihalicyn: Thanks, will be fixed.

.github/workflows/linux-next.yml
mihalicyn: Will be fixed ;)

.github/workflows/linux-next.yml
mihalicyn: Will fix this, thanks ;)

.github/workflows/linux-next.yml
mihalicyn: Thanks, fixed!

.github/workflows/linux-next.yml
mihalicyn: Fixed, thanks!

avagin:
LGTM. Thanks!

PR #3031, opened 2026-06-05
zdtm: add test for SO_SNDTIMEO and SO_RCVTIMEO

rocker-zhang:
CRIU dumps and restores the SO_SNDTIMEO and SO_RCVTIMEO socket timeout options in `dump_socket_opts()`/`restore_socket_opts()` (`criu/sockets.c`), but ZDTM has no dedicated coverage for them. SO_SNDTIMEO is not exercised by any test, and SO_RCVTIMEO is only used incidentally in `socket_icmp`, never asserted across checkpoint/restore.

This adds `socket-timeo`, which sets distinct send and receive timeouts, performs checkpoint/restore, and verifies both values survive.

Whole-second timeouts are used because the kernel stores these options with jiffy granularity: setsockopt() may round a sub-second value to the nearest tick depending on HZ, so a literal sub-second expectation could fail to match the stored value regardless of checkpoint/restore. The send and receive values differ so a mix-up between the two options is caught.

Tested on x86_64 (kernel 5.15) and aarch64 (kernel 6.17): passes in the `h`, `uns`, and `ns` flavors.

### DIFF
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -137,6 +137,7 @@ TST_NOFILE	:=				\
 		socket-tcp-skip-in-flight	\
 		socket-tcp-keepalive		\
 		socket-linger			\
+		socket-timeo			\
 		sock_opts00			\
 		sock_opts01			\
 		sock_opts02			\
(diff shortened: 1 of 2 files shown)


avagin:
LGTM. Thanks

rst0git (approved):
LGTM

PR #2754, opened 2025-09-23
Files on Unmounted Bind Mounts

bsach64:
This PR adds initial support for files on detached bind mounts.
Detached mounts are mounts which are unmounted using the `MNT_DETACH` flag. This commit adds support for files opened on detached mounts which are a bind mount of a existing mount (which is not detached).

This PR requires the kernel patches in this github branch to be applied: 
~https://github.com/bsach64/linux/tree/umount-mnt-ns-plus-statx~
~https://github.com/bsach64/linux/tree/statmount-fd-v2~
~https://github.com/bsach64/linux/tree/statmount-fd-v3~
~https://github.com/bsach64/linux/tree/statmount-fd-v4~
https://github.com/bsach64/linux/tree/statmount-fd-v5

This PR is pretty draft, we are still to support a lot of things. This work is part of Google Summer of Code 2025.

PR #3036, opened 2026-06-10
cgroup: preserve cgroup2 superblock options on dump

### LINKED ISSUE #3029: CRIU checkpoint mutates host cgroup v2 mount options in default cgroup mode
## Description

When checkpointing a runc container on a cgroup v2 host, CRIU appears to mutate the host/global `/sys/fs/cgroup` mount options.

This was noticed while investigating opencontainers/runc#5267:
https://github.com/opencontainers/runc/pull/5267#discussion_r3274845060

Before checkpoint, the host cgroup v2 mount includes options such as `nsdelegate` and `memory_recursiveprot`. After `runc checkpoint` using the default cgroup management mode, those options disappear from `/proc/self/mountinfo`.

Example:

```console
BEFORE: ... /sys/fs/cgroup rw,... - cgroup2 cgroup2 rw,seclabel,nsdelegate,memory_recursiveprot,memory_hugetlb_accounting
AFTER:  ... /sys/fs/cgroup rw,... - cgroup2 cgroup2 rw,seclabel
```
A checkpoint/restore operation should not change the host cgroup v2 mount options.

## Reproducer

Prerequisites:

- root
- unified cgroup v2
- runc
- criu
- jq
- docker, only to create a simple BusyBox rootfs

```bash
#!/usr/bin/env bash
set -euo pipefail

: "${RUNC:=runc}"
: "${IMAGE:=busybox:latest}"
: "${MODE:=default}"

id="criu-cgv2-mountopts-repro-$$"
work="$(mktemp -d)"
cid=""

cleanup() {
      set +e
      cd "$work" 2>/dev/null || true
      "$RUNC" kill "$id" KILL >/dev/null 2>&1 || true
      "$RUNC" delete -f "$id" >/dev/null 2>&1 || true
      [ -n "$cid" ] && docker rm "$cid" >/dev/null 2>&1 || true
      rm -rf "$work"
}
trap cleanup EXIT

if [ "$(id -u)" -ne 0 ]; then
      echo "run as root"
      exit 1
fi

test -f /sys/fs/cgroup/cgroup.controllers || {
      echo "requires unified cgroup v2"
      exit 1
}

command -v "$RUNC" >/dev/null
command -v criu >/dev/null
command -v jq >/dev/null
command -v docker >/dev/null

cgline() {
      awk '$5 == "/sys/fs/cgroup" {print; exit}' /proc/self/mountinfo
}

cd "$work"
mkdir rootfs image-dir work-dir

docker pull "$IMAGE" >/dev/null
cid="$(docker create "$IMAGE" true)"
docker export "$cid" | tar -C rootfs -xf -
docker rm "$cid"
cid=""

"$RUNC" spec
jq '
      .process.terminal = false
      | .process.args = ["sh", "-c", "exec </dev/null >/dev/null 2>&1; while true; do sleep 1; done"]
' config.json > config.new
mv config.new config.json

before="$(cgline)"
printf 'BEFORE: %s\n' "$before"

"$RUNC" run -d "$id"
sleep 1

checkpoint_args=(checkpoint --work-path ./work-dir --image-path ./image-dir)
if [ "$MODE" != "default" ]; then
      checkpoint_args+=(--manage-cgroups-mode "$MODE")
fi

set +e
"$RUNC" "${checkpoint_args[@]}" "$id"
checkpoint_rc=$?
set -e

after="$(cgline)"
printf 'AFTER:  %s\n' "$after"
printf 'checkpoint exit status: %s\n' "$checkpoint_rc"

if [ "$before" != "$after" ]; then
      echo "BUG: host cgroup v2 mountinfo changed"
      exit 2
fi

echo "No host cgroup v2 mountinfo change observed"
```

Run:

sudo ./repro.sh

Observed result:

- runc checkpoint with default cgroup mode changes the /sys/fs/cgroup mountinfo line.
- In my environment, nsdelegate and memory_recursiveprot disappear.

Workaround check:

sudo MODE=ignore ./repro.sh

When --manage-cgroups-mode ignore is used, the mutation does not happen in my testing. Run this on a fresh VM or after restoring the host cgroup mount options, because the default-mode run already mutates the host state.

## Expected result

CRIU should not mutate host/global cgroup v2 mount options during checkpoint/restore, regardless of cgroup management mode.

## Actual result

With default cgroup management mode, checkpointing causes /sys/fs/cgroup mount options on the host to change.

## Investigation clue

The suspicious path seems to be CRIU cgroup management when manage-cgroups-mode is the default/soft mode and no cgroup_yard is provided.

From reading the CRIU cgroup code, the path to investigate is around:

- prepare_cgroup_namespace
- cgroup yard setup
- open_cgroupfs
- __new_open_cgroupfs

On kernels with the new mount API, __new_open_cgroupfs uses fsopen("cgroup2"), fsconfig(FSCONFIG_CMD_CREATE), and fsmount() for cgroup v2. Since this is done without preserving the existing cgroup v2 mount options, it appears to
create or reconfigure the cgroup2 superblock with default options, which drops options such as nsdelegate and memory_recursiveprot from the host-visible mount.

Using --manage-cgroups-mode ignore avoids this path and avoids the mutation.

Related runc discussion:
https://github.com/opencontainers/runc/pull/5267#discussion_r3284874536

rocker-zhang:
criu reconfigures the host cgroup2 mount during checkpoint, dropping its superblock options.

When collecting cgroups for dump, `__new_open_cgroupfs()` opens the unified cgroup2 hierarchy with `fsopen("cgroup2")` and then `cr_fsconfig(FSCONFIG_CMD_CREATE)` (or, without fsopen support, `mount("none", ..., "cgroup2", 0, NULL)`), neither of which carries any superblock option. cgroup2 has a single superblock shared by every mount, so this reconfigures the live host mount and drops options criu did not set, such as `nsdelegate`, `memory_recursiveprot` and `memory_hugetlb_accounting`. The detached-mount teardown only removes criu's own mount instance, so the host cgroup2 is left altered after a checkpoint (#3029).

The fix reads the options of the existing cgroup2 mount from `/proc/self/mountinfo` and replays the known superblock flags before `FSCONFIG_CMD_CREATE` (and passes them as mount data in the fsopen-less path), leaving the host mount intact.

I reproduced this in a throwaway VM (its own kernel, so the mutated superblock is the VM's): the current sequence strips `nsdelegate,memory_recursiveprot` from an existing cgroup2 mount, and the replay preserves them. Builds on x86_64 and aarch64.

Two things I would like direction on. The fix replays an allowlist of known superblock flags; an alternative is to clone the existing mount with `open_tree(OPEN_TREE_CLONE)` and avoid reconfiguring the superblock at all, which is a larger change. Separately, preserving these options across checkpoint/restore (rather than only avoiding the host mutation) would need a new field in `cg_controller_entry`, which is an image-format change I have left out of this PR.

Fixes: #3029


avagin:
@Snorch could you please review this pr?

Snorch:
The problem reproduces:
```
cat /proc/self/mountinfo  | grep cgroup2
48 46 0:30 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:7 - cgroup2 cgroup2 rw,seclabel,nsdelegate,memory_recursiveprot,memory_hugetlb_accounting

python3 test/zdtm.py run -t zdtm/static/cgroup00 --ignore-taint

cat /proc/self/mountinfo  | grep cgroup2
48 46 0:30 / /sys/fs/cgroup rw,nosuid,nodev,noexec,relatime shared:7 - cgroup2 cgroup2 rw,seclabel
```

Snorch:
Parsing mountinfo is generally slow (especially host's mountinfo). Preferably we want to avoid doing this one more time (in ideal world we should be switching to listmount+statmount everywhere).

What about alternative approach:
```
int fd = open("/sys/fs/cgroup", O_PATH);
struct mnt_id_req req = { .size = sizeof(req), .mnt_fd = fd,
                          .param = STATMOUNT_MNT_OPTS };
statmount(&req, sm, bufsz, STATMOUNT_BY_FD);
```
We will have sb opts in `sm->mnt_opts`.

Note: Need to verify that it's cgroup2 mount and not something else from other stamount fields.

Yes it only works when cgroup-v2 is mounted into `/sys/fs/cgroup` on host (which is pretty much always), overwise we have no other option except searching for it which should probably be done via listmount/statmount.

rocker-zhang:
> Parsing mountinfo is generally slow (especially host's mountinfo). [...] What about alternative approach: `statmount(&req, sm, bufsz, STATMOUNT_BY_FD)` [...] We will have sb opts in `sm->mnt_opts`.

Thanks for reproducing it and for the statmount suggestion. Done in the latest revision.

I went with the portable form rather than `STATMOUNT_BY_FD`: that flag and the `mnt_id_req.mnt_fd` field aren't in a released kernel yet (they're in your `vfs-6.20.namespace` "statmount: accept fd as a parameter" series), so as a baseline I get the unique mount id from `statx(STATX_MNT_ID_UNIQUE)` on `/sys/fs/cgroup` and pass it as `mnt_id_req.mnt_id` with `flags=0`. statmount(2)/the unique id are 6.8 and `STATMOUNT_MNT_OPTS` is 6.11, so on older kernels (or if the kernel doesn't return the bit) it falls back to the previous `/proc/self/mountinfo` parse. The mount is verified as cgroup2 via `sb_magic` (`STATMOUNT_SB_BASIC`) before the options are used.

Once `STATMOUNT_BY_FD` lands it'd be a clean follow-up to drop the statx step and pass the O_PATH fd directly.

The struct/flag definitions are CRIU-local (`cr_` prefixed) because libc `<sys/mount.h>` (glibc 2.39) already declares `struct statmount` with the pre-6.11 layout that lacks `mnt_opts`, so reusing it isn't an option; field offsets are checked with `BUILD_BUG_ON`.

Verified locally on 6.17: statmount returns the cgroup2 `mnt_opts` and, after the allowlist filter, the same set of options the mountinfo parse produced.

criu/cgroup.c
```diff
@@ -574,6 +576,134 @@ static int add_freezer_state(struct cg_controller *controller)
...
+ * cgroup2 mount so they can be replayed, leaving the host mount intact (#3029).
+ */
+static const char *const cgroup2_sb_opts[] = {
+	"nsdelegate",
+	"memory_recursiveprot",
+	"memory_localevents",
+	"memory_hugetlb_accounting",
+	"favordynmods",
+	"pids_localevents",
+};
```
Snorch, line 595: Please order them in the same order they appear in kernel:
```
static const struct fs_parameter_spec cgroup2_fs_parameters[] = {
        fsparam_flag("nsdelegate",              Opt_nsdelegate),
        fsparam_flag("favordynmods",            Opt_favordynmods),
        fsparam_flag("memory_localevents",      Opt_memory_localevents),
        fsparam_flag("memory_recursiveprot",    Opt_memory_recursiveprot),
        fsparam_flag("memory_hugetlb_accounting", Opt_memory_hugetlb_accounting),
        fsparam_flag("pids_localevents",        Opt_pids_localevents),
        {}
};
```
So that it's easier to match them.

criu/cgroup.c
```diff
@@ -574,6 +576,134 @@ static int add_freezer_state(struct cg_controller *controller)
...
+	return 0;
+}
+
+/* Fallback: scan /proc/self/mountinfo for the cgroup2 superblock options. */
+static void cgroup2_collect_sb_opts_mountinfo(char *opts, size_t size)
+{
+	FILE *f;
+	char line[1024];
+
+	f = fopen("/proc/self/mountinfo", "r");
```
Snorch: Let's use our helper for this `fopen_proc(getpid(), "mountinfo")`.

criu/cgroup.c
```diff
@@ -574,6 +576,134 @@ static int add_freezer_state(struct cg_controller *controller)
...
+
+	fclose(f);
+}
+
+static void cgroup2_collect_sb_opts(char *opts, size_t size)
+{
+	opts[0] = '\0';
+
+	if (cgroup2_collect_sb_opts_statmount(opts, size) == 0)
+		return;
```
Snorch, line 703: In ideal world we should try to reuse https://github.com/checkpoint-restore/criu/pull/2754/changes/d67f966c30f717bad6f4094117648fa7d930027e from this PR https://github.com/checkpoint-restore/criu/pull/2754

cc @bsach64

The kerndat solution is more native to CRIU, where we check support of statmount only once on a system and cache it to kerndat file so we don't need to retry syscall on every criu run.

bsach64: happy to split off introduction of `statmount` out of https://github.com/checkpoint-restore/criu/pull/2754
so that this PR and that one can rebase on top of the new one which will introduce statmount

how does that sound @Snorch?

rocker-zhang:
Thanks for the review. I've reordered the allowlist to match the kernel's `cgroup2_fs_parameters[]` and switched the fallback to `fopen_proc(PROC_SELF, "mountinfo")`.

On statmount: agreed, a shared introduction with kerndat-cached detection is cleaner than the inline structs and per-run probe I have here. @bsach64 splitting it out of #2754 sounds good to me — I'll drop the vendored statmount bits from this PR and rebase on top of that once it lands, using the shared helper. Until then I can keep the mountinfo path as-is or hold this PR for the rebase, whichever you prefer.

criu/cgroup.c
Snorch: @bsach64 Sounds good to me, please do!

bsach64:
> I went with the portable form rather than `STATMOUNT_BY_FD`: that flag and the `mnt_id_req.mnt_fd` field aren't in a released kernel yet (they're in your `vfs-6.20.namespace` "statmount: accept fd as a parameter" series), so as a baseline I get the unique mount id from `statx(STATX_MNT_ID_UNIQUE)` on `/sys/fs/cgroup` and pass it as `mnt_id_req.mnt_id` with `flags=0`. statmount(2)/the unique id are 6.8 and `STATMOUNT_MNT_OPTS` is 6.11, so on older kernels (or if the kernel doesn't return the bit) it falls back to the previous `/proc/self/mountinfo` parse. The mount is verified as cgroup2 via `sb_magic` (`STATMOUNT_SB_BASIC`) before the options are used.

`STATMOUNT_BY_FD` is in the 7.0 released kernel (ubuntu 26.04 LTS ships with it).
There is an argument to be that it is too new, but in order use the `statx()` approach with CRIU we would have to properly add kerndat checks for:
- `statx`
- `STATX_MNT_ID_UNIQUE`
- `STATMOUNT_MNT_OPTS`

We could instead use the fallback of parsing `mountinfo` for kernels < 7.0 and use `STATMOUNT_BY_FD` for anything newer (`STATMOUNT_BY_FD` would automatically imply support for `STATMOUNT_MNT_OPTS`)

But, maybe a large enough user base is between `6.11` and `7.0` that adding proper support for `statx` is worth it.

In any case, I have raised https://github.com/checkpoint-restore/criu/pull/3054 which adds `statmount` and `STATMOUNT_BY_FD`  (I will take a little bit more time to clean it up)

rocker-zhang:
> `STATMOUNT_BY_FD` is in the 7.0 released kernel (ubuntu 26.04 LTS ships with it).

Good to know — I'll drop the "not in a released kernel" caveat.

> maybe a large enough user base is between `6.11` and `7.0` that adding proper support for `statx` is worth it.

I'd lean against it. Over the mountinfo fallback the `statx` path only buys skipping the mountinfo parse on 6.11–7.0, and that window only shrinks from here, while it costs three new kerndat probes (`statx`, `STATX_MNT_ID_UNIQUE`, `STATMOUNT_MNT_OPTS`) plus the vendored `cr_` structs we want to drop. The fallback is already correct there, just slower, which is fine for a once-per-dump sb read.

So my preference is your second option: use `STATMOUNT_BY_FD` when #3054's kerndat reports it available (which implies `STATMOUNT_MNT_OPTS`), and fall back to mountinfo parsing otherwise. If a real 6.11–7.0 need turns up later, the `statx` path can be a self-contained follow-up.

> I have raised #3054 which adds `statmount` and `STATMOUNT_BY_FD` (I will take a little bit more time to clean it up)

Either way #3036 should sit on top of #3054, so I'll rebase onto the shared helper and drop the vendored bits once it lands. Ping me when it's ready. I'll defer to you and @Snorch on the 6.11–7.0 call.

criu/cgroup.c
bsach64: @rocker-zhang @Snorch https://github.com/checkpoint-restore/criu/pull/3054 is good for you to take a look

avagin: #3054 was merged. @rocker-zhang could you please update this pr. Thanks.

kolyshkin:
> Either way #3036 should sit on top of #3054, so I'll rebase onto the shared helper and drop the vendored bits once it lands. Ping me when it's ready. I'll defer to you and @Snorch on the 6.11–7.0 call.

@rocker-zhang PTAL; #3054 is merged; also see comments from @Snorch above.

rocker-zhang:
Updated. Rebased onto current `criu-dev` now that #3054 has merged, and refactored to reuse that statmount infrastructure instead of the private definitions this PR used to carry:

- Dropped the vendored `cr_mnt_id_req` / `cr_statmount` / `cr_statx` / `STATMOUNT_*` / `__NR_statmount` defs (−92 lines from `criu/include/linux/mount.h`).
- Now uses `sys_statmount()`, `MNT_ID_REQ_SIZE_VER1`, the shared structs/flags from `mount-v2.h`, and `struct cr_statx` from `util.h`.
- Gates the fast path on `kdat.has_statmount` and reuses the same statx-by-`O_PATH`-fd idiom as `kerndat_has_statmount()`.
- Only `CGROUP2_SUPER_MAGIC` stays local.

Builds clean; `zdtm/static/cgroup00` and `cgroup02` pass C/R, and the statmount fast path reads back the host cgroup2 options (`nsdelegate,memory_recursiveprot`) on a 6.11+ kernel. @avagin @kolyshkin PTAL.

criu/cgroup.c
```diff
@@ -574,6 +577,144 @@ static int add_freezer_state(struct cg_controller *controller)
...
+		.param = STATMOUNT_SB_BASIC | STATMOUNT_MNT_OPTS,
+	};
+	/* statmount(2) copies the mnt_opts string into the space past the
+	 * fixed struct (cr_statmount ends in a flexible str[] member), so the
+	 * buffer must be larger than the struct itself. */
+	union {
+		struct cr_statmount sm;
+		char buf[sizeof(struct cr_statmount) + 4096];
+	} u = {};
+	int fd, ret = -1;
```
avagin, line 645: we use exit_code instead of ret in such case. It is a well-known pattern in criu.

criu/cgroup.c
```diff
@@ -574,6 +577,144 @@ static int add_freezer_state(struct cg_controller *controller)
...
+
+	if (!kdat.has_statmount)
+		return -1;
+
+	fd = open("/sys/fs/cgroup", O_PATH);
+	if (fd < 0)
+		return -1;
```
avagin, line 652: If something goes wrong, we want to know what actually fails. Please log syscall errors here and in other cases.

criu/cgroup.c
```diff
@@ -574,6 +577,144 @@ static int add_freezer_state(struct cg_controller *controller)
...
+
+	if (syscall(SYS_statx, fd, "", AT_EMPTY_PATH, STATX_MNT_ID_UNIQUE, &stx) < 0)
+		goto close;
```
avagin, line 655: nit: close sounds more like err

criu/cgroup.c
```diff
@@ -603,6 +744,16 @@ static int __new_open_cgroupfs(struct cg_ctl *cc)
 			}
 			name = strtok_r(NULL, ",", &saveptr);
 		}
+	} else { /* cgroup2: replay the host superblock options (#3029) */
+		char opts[256], *tok, *saveptr;
+
+		cgroup2_collect_sb_opts(opts, sizeof(opts));
```
avagin, line 750: cgroup2_collect_sb_opts can fail, can't it? We need to handle its errors. Can __new_open_cgroupfs be called multiple times? If it can, we can consider caching sb opts.

rocker-zhang:
Updated. All feedback addressed:

- **Complete mountinfo records**: replaced `fgets()` with `bfd`/`breadline()` (matching `parse_mountinfo()` in `proc_parse.c`), so long cgroup2 lines are no longer silently truncated.
- **Explicit error propagation**: both `cgroup2_collect_sb_opts_statmount()` and `cgroup2_collect_sb_opts_mountinfo()` now return `int` with a CRIU-style `exit_code` pattern. A found mount with zero allowlisted flags returns 0 (success); open/read/parse/search failures return −1. Both callers in `__new_open_cgroupfs()` and `open_cgroupfs()` check the return value and abort the mount before `FSCONFIG_CMD_CREATE`/`mount()`.
- **Fail-closed mount**: if any option-replay `FSCONFIG_SET_FLAG` call fails, the code now goes to `err:` and closes the fs context before `FSCONFIG_CMD_CREATE`, matching the cgroup v1 error path above.
- **Successful-result cache**: `collect_cgroups()` can be invoked for multiple cgroup sets during dump, so the first successful collection is cached and replayed for subsequent opens; failures are not cached, so later attempts can retry.
- **Old-kernel fallback kept**: when `statmount(2)` is unsupported or does not return `STATMOUNT_MNT_OPTS`, the code falls back to mountinfo. Fallback-only diagnostics use `pr_debug()` so they do not emit false "Error" lines on normal fallback.
- **Logging**: actual syscall failures now use `pr_perror` so the cause is visible.
- **Allowlist order**: matches kernel `cgroup2_fs_parameters[]` as requested.

PR #3037, opened 2026-06-10
mem: keep COW root VMAs writable when premapping

rst0git:
Applications sometimes call mmap with `PROT_NONE` to reserve a large amount of the virtual address space. When CRIU premaps private VMAs during restore, it used to map them with `PROT_WRITE`  so it can fill in their pages. However, the kernel charges the full size of writable private mapping against the overcommit limit, so premapping a large reservation as writable can fail with `ENOMEM`. 

Commit 7daaa11a (`"mem: don't PROT_WRITE on reservation mmaps"`) added an optimization to premap such a VMA as `PROT_NONE` instead of writable when a region has no pages to be restored. However, this assumption does not hold for a COW root.

When CRIU looks for COW-shared VMAs between a parent and child, it matches them by start/end and flags but ignores their protection bits (see `check_cow_vmas()`). So a parent's `PROT_NONE`, not-accountable reservation can be matched as the COW root of a child VMA that does have pages (the child made the same range writable and wrote to it after `fork()`).

On restore:
1. `premap_private_vma()` maps the COW root as `PROT_NONE`.
2. the inherited child re-positions that same mapping with `mremap()`, so the child's premapped area is `PROT_NONE` too.
3. `restore_priv_vma_content()` then restores the child's pages. It `memcmp()`s them against the premapped area and copies them in.

 Because the mapping is PROT_NONE, the memcmp() and copy in the previous step are accessing memory with no read or write permission, which faults and kills the restored task with SIGSEGV:

```
$ sudo ./zdtm.py run -t zdtm/static/cow02 -f h --ignore-taint
userns is supported
The kernel is tainted: '12288'
=== Run 1/1 ================ zdtm/static/cow02
========================== Run zdtm/static/cow02 in h ==========================
Start test
./cow02 --pidfile=cow02.pid --outfile=cow02.out
Run criu dump
Run criu restore
=[log]=> dump/zdtm/static/cow02/70/1/restore.log
b'(00.011907)     71: pr71-2 Read 409000 1 pages'
b'(00.011909)     71: \tpr71-2 Read page from self 409000/2000'
b'(00.011925)     71: pr71-2 Read 7f0dc1762000 1 pages'
b'(00.011927)     71: \tpr71-2 Read page from self 7f0dc1762000/3000'
b'(00.053817)     70: Error (criu/cr-restore.c:1268): 71 killed by signal 11: Segmentation fault'
b'(00.053835) Error (criu/cr-restore.c:2369): Restoring FAILED.'
b'(00.054002) Error (criu/cr-restore.c:1268): 70 killed by signal 9: Killed'
################# Test zdtm/static/cow02 FAIL at CRIU restore ##################
```

Fixes: 7daaa11aa039 ("mem: don't PROT_WRITE on reservation mmaps")

### DIFF
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -958,8 +958,17 @@ static int premap_private_vma(struct pstree_item *t, struct vma_area *vma, void 
 		 * (did not have the "ac" flag in /proc/pid/smaps), we
 		 * can safely mmap them with PROT_NONE because we know
 		 * we will never need to write any bits to them.
+		 *
+		 * This only holds for a standalone VMA (vma->pvma == NULL).
+		 * A COW root (vma->pvma == VMA_COW_ROOT) shares this very
+		 * mapping with inherited children via the mremap() in the
+		 * branch below; those children may have pages to restore,
+		 * and restoring content into a PROT_NONE mapping faults
+		 * (e.g. the memcmp()/copy in restore_priv_vma_content()).
+		 * So a COW root must stay writable even when it is itself
+		 * PROT_NONE and not accountable.
 		 */
-		if (vma->e->prot == PROT_NONE && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
+		if (vma->e->prot == PROT_NONE && vma->pvma == NULL && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
 			addr = mmap(*tgt_addr, size, PROT_NONE, vma->e->flags | MAP_FIXED | flag, vma->e->fd,
 				    vma->e->pgoff);
 		} else {
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -192,6 +192,7 @@ TST_NOFILE	:=				\
 		socket-ext			\
 		unhashed_proc			\
 		cow00				\
+		cow02				\
 		child_opened_proc		\
 		posix_timers			\
 		sigpending			\
--- a/test/zdtm/static/cow02.c
+++ b/test/zdtm/static/cow02.c
@@ -1,0 +1,100 @@
+#include <sys/mman.h>
+#include <unistd.h>
+#include <sys/wait.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Restore a COW child whose COW-root VMA is a PROT_NONE reservation";
+const char *test_author = "Radostin Stoyanov <rstoyanov@fedoraproject.org>";
+
+/*
+ * A parent reserves an address range with PROT_NONE and never touches it.
+ * Such a mapping has no "ac" flag in /proc/pid/smaps, so CRIU marks it
+ * VMA_AREA_NOT_ACCOUNTABLE. After fork() the child turns the very same range
+ * into a writable mapping and fills it with data.
+ *
+ * CRIU pairs VMAs that coincide by start/end and flags regardless of their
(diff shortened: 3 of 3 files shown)


adrianreber:
Looks reasonable to me. Do you know why CI is not running?

rst0git:
> Do you know why CI is not running?

One of the `CI / Alpine Test` jobs failed due to a flaky test and cancelled all other CI tests. I restarted the failed test, and it passed, but the cancelled tests didn't restart automatically. I've opened the following pull request to fix this: https://github.com/checkpoint-restore/criu/pull/3038

bsach64:
lgtm @rst0git for what it's worth, thank you for fixing this!

adrianreber (approved):
Seen this error a couple of times with a gdb backtrace. Good to have a zdtm test for this. Thanks.

criu/mem.c
```diff
@@ -958,8 +958,17 @@ static int premap_private_vma(struct pstree_item *t, struct vma_area *vma, void
...
+		 * A COW root (vma->pvma == VMA_COW_ROOT) shares this very
+		 * mapping with inherited children via the mremap() in the
+		 * branch below; those children may have pages to restore,
+		 * and restoring content into a PROT_NONE mapping faults
+		 * (e.g. the memcmp()/copy in restore_priv_vma_content()).
+		 * So a COW root must stay writable even when it is itself
+		 * PROT_NONE and not accountable.
 		 */
-		if (vma->e->prot == PROT_NONE && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
+		if (vma->e->prot == PROT_NONE && vma->pvma == NULL && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
```
avagin, line 971: Do you understand that a non-accountable mapping can't hold any pages? I think the proper patch should look like:
```
index 69ccf05ca0a4..0ac00f3d400d 100644
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -838,6 +838,9 @@ static inline bool check_cow_vmas(struct vma_area *vma, struct vma_area *pvma)
        /* ... belong to the same file if being filemap */
        if (!(vma->e->flags & MAP_ANONYMOUS) && vma->e->shmid != pvma->e->shmid)
                return false;
+       if (vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE) ||
+           vma_area_is(pvma, VMA_AREA_NOT_ACCOUNTABLE))
+               return false;
 
        pr_debug("Found two COW VMAs @0x%" PRIx64 "-0x%" PRIx64 "\n", vma->e->start, pvma->e->end);
        return true;
@@ -968,7 +971,7 @@ static int premap_private_vma(struct pstree_item *t, struct vma_area *vma, void
                 * So a COW root must stay writable even when it is itself
                 * PROT_NONE and not accountable.
                 */
-               if (vma->e->prot == PROT_NONE && vma->pvma == NULL && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
+               if (vma->e->prot == PROT_NONE && vma_area_is(vma, VMA_AREA_NOT_ACCOUNTABLE)) {
                        addr = mmap(*tgt_addr, size, PROT_NONE, vma->e->flags | MAP_FIXED | flag, vma->e->fd,
                                    vma->e->pgoff);
                } else {
```

rst0git: Thanks Andrei! We've applied this patch in https://github.com/checkpoint-restore/criu/pull/3069

PR #3038, opened 2026-06-10
ci: disable fail-fast for CI matrix jobs

rst0git:
By default GitHub Actions cancels all in-progress and queued matrix legs as soon as one leg fails (`fail-fast: true`). Re-running only the failed job afterwards does not bring back the cancelled siblings, so recovering from a single failure requires re-running the whole job.

Cancelling the siblings also hides information: whether the other shards or environments would have passed or failed is exactly what we want to know from a CI run, not something to discard on the first failure. Set fail-fast: false on the matrix jobs that were still inheriting the default, so every leg runs to completion independently.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,7 @@ jobs:
   alpine-test:
     name: Alpine Test (${{ matrix.target }}, ${{ matrix.shard_name }})
     strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


adrianreber:
All CI failures are known failures (mainly tcp socketbuf flakes). LGTM.

adrianreber (approved):

PR #3039, opened 2026-06-11
ci: install gdb on dnf based distributions

adrianreber:
The custom core dump handler tries to call gdb which makes it much easier to understand why something crashed especially in CI. For dnf based distribution this currently does not work because gdb is missing. This installs gdb on dnf based CI runs.


rst0git (approved):
LGTM

Issue #2854, opened 2025-12-24
Error (criu/fsnotify.c:284): fsnotify:      Can't dump that handle


qiansheng1990:
hi, guys

I'm using criu 4.2 with k8s and containerd + runc, have a pod of ubuntu 22.04 + xfce4 running. 
After curl <kubelet_ip>/checkpoint/<namespace>/<pod>/<container> to trigger a checkpoint of that container, it failed to dump, part of log:

(00.067959) irmap: Scanning /. hint
(00.067960) irmap: Refresh stat for /.
(00.067962) irmap: Scanning /no-such-path hint
(00.067964) irmap: Refresh stat for /no-such-path
(00.067968) Warn  (criu/irmap.c:105): irmap: Can't stat /no-such-path: No such file or directory
(00.067971) Error (criu/fsnotify.c:284): fsnotify:      Can't dump that handle
(00.067979) ----------------------------------------
(00.067990) Error (criu/cr-dump.c:1698): Dump files (pid: 1327079) failed with -1

host system: ubuntu 22.04
kernel: 5.15.0-151-generic 
k8s: 1.31.2
criu: 4.2
full log: 
[criu.20251224.log](https://github.com/user-attachments/files/24329360/criu.20251224.log)

Please help to see how to solve? Thanks.

adrianreber:
Your container is running a lot of different processes. One of the processes seems to use a resource CRIU cannot handle. Maybe just run the one process you need and not a complete desktop environment. What is your container supposed to do?

qiansheng1990:
> Your container is running a lot of different processes. One of the processes seems to use a resource CRIU cannot handle. Maybe just run the one process you need and not a complete desktop environment. What is your container supposed to do?

hi, i'm traing agent to perform specific task in contaienr with desktop. thinking using criu to checkpoint env, restore env for evaluating different strategy. 

looking into detailed log, the handle criu can not dump seems to to be: 
a process sets up inotify watches on files located in the lowerdir of the container’s overlay filesystem.

adrianreber:
Yes, seems like that will not work.

BuildWithNitish:
Hi,
I’m interested in this issue and would like to work on it to better understand CRIU.

I’m still learning CRIU internals, but I’d like to explore the fsnotify part, reproduce the problem locally, and check whether this is a known limitation or something that could be improved or documented more clearly.

If that’s okay, could you please assign this issue to me?
[...]

PR #3043, opened 2026-06-11
fsnotify: handle inotify watches on overlayfs mounts 

adrianreber:
When a process has inotify or fanotify watches on files inside an overlayfs mount, criu dump fails because open_by_handle_at() does not support overlay file handles (type OVL_FILEID_V1 = 0xf8). The dump log shows:                       

```                                                                                                                                                                                                                 
  fsnotify: wd 0x000001 s_dev 0x00002d i_ino 0x101ad2 mask 0x00033a  
  fsnotify: [fhandle] bytes 0x000020 type 0x0000f8 ...    
  fsnotify: Handle 0x2d:0x101ad2 cannot be opened     
  Error (criu/fsnotify.c:284): fsnotify: Can't dump that handle
```
                                                                                                                                                                                                                                        
To reproduce use something like (also take a look at ` test/others/overlayfs`) :          
                                                                                                                                                                                                                     
```                                                                                                                                                                                                                                              
  mkdir -p /tmp/{lower,upper,work,merged}   
  touch /tmp/lower/testfile    
  mount -t overlay overlay \ 
    -o lowerdir=/tmp/lower,upperdir=/tmp/upper,workdir=/tmp/work \     
    /tmp/merged              
  inotifywait -m /tmp/merged/testfile &    
  PID=$!  
  criu dump -t $PID -D /tmp/imgs --shell-job --ext-unix-sk   
```       
                                                                                                                                                                                                                                       
This adds a directory-walk fallback in alloc_openable() (criu/fsnotify.c): when open_by_handle_at() fails and the mount is FSTYPE__OVERLAYFS, the code walks the overlay mount's directory tree using opendir/readdir/fstatat to find the file matching the (s_dev, i_ino) pair from the handle. The discovered absolute path is stored in f_handle->path so that get_mark_path() can use the path-based openat() strategy on restore, which works on overlay mounts.                 
                                                                                                                                                                                                                                              
The fallback only triggers when both conditions are met (handle open failure and overlay fstype), so there is no impact on other filesystems. The walk is capped at a depth of 64 and emits a warning after 10,000 entries.                 
                                                                                                                                                                                                                                              
  - A ZDTM test (inotify_overlayfs) and a test/others/overlayfs integration test were added to cover inotify C/R on overlayfs.

### DIFF
--- a/criu/fsnotify.c
+++ b/criu/fsnotify.c
@@ -7,6 +7,7 @@
 #include <string.h>
 #include <utime.h>
 #include <limits.h>
+#include <dirent.h>
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/inotify.h>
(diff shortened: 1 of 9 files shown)


rst0git (approved):
LGTM

avagin:
Have you reproduced the issue with a specific workload? Have you explored any other ways to solve it?

Honestly, I don't like the fact that a directory-walk fallback is executed for each entry separately. Maybe there is a way to collect all such entries and then run the fallback once for all of them?

Another idea we could try here is to look for this (dev, ino) pair in the opened file descriptors. In many cases, a process that sets up a notification on a file or directory will have it open.

adrianreber:
> Have you reproduced the issue with a specific workload?

It was a .NET application that creates this error in a container and it was possible to reproduce it locally with inotify on overlayfs.

> Have you explored any other ways to solve it?

No.

> Honestly, I don't like the fact that a directory-walk fallback is executed for each entry separately. Maybe there is a way to collect all such entries and then run the fallback once for all of them?
> 
> Another idea we could try here is to look for this (dev, ino) pair in the opened file descriptors. In many cases, a process that sets up a notification on a file or directory will have it open.

I will take a look and post a PR.

adrianreber:
#2854 could also  be related

PR #3045, opened 2026-06-12
test: fix unused variable warnings reported by CodeQL

adrianreber:
Fix four "unused local/global variable" alerts reported by CodeQL:

- zdtm.py: drop unused assignment to print_next in the tail-printing loop at the end of grep_errors(). The return value of print_error() is not needed here since no further iterations follow.

- crit-recode.py: remove duplicate pycriu.images.dumps(pb) call inside the except handler. The same call just raised the exception, so re-executing it would raise again and prevent the error message from being printed.

- exhaustive/unix.py: drop unused msg variable. The recv() call is used for its blocking side effect to synchronize with the parent, the received data is not needed.

- others/mounts/mounts.py: rename unused loop variable i to _ in the range(10) iteration.

### DIFF
--- a/test/crit-recode.py
+++ b/test/crit-recode.py
@@ -24,7 +24,6 @@ def recode_and_check(imgf, o_img, pretty):
     try:
         r_img = pycriu.images.dumps(pb)
     except Exception as e:
-        r_img = pycriu.images.dumps(pb)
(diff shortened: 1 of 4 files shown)


rst0git (approved):
LGTM

PR #3046, opened 2026-06-13
ci: cover the O_DIRECT page-read path on restore

rocker-zhang:
On restore criu probes the pages image with `probe_pages_o_direct()` (criu/pagemap.c) and, when the image lives on an O_DIRECT-capable filesystem, reads pages with O_DIRECT and selects the io_submit-based `restore_vma_aio()` engine for the async vma_io path (criu/pie/restorer.c). The CI image directory does not support O_DIRECT, so the probe always falls back to buffered I/O and this selection is never exercised in CI.

This adds `test_odirect()` to the non-shardable tests. It puts the zdtm images on a dedicated O_DIRECT-capable ext4 loop device, runs an iterative memory test (`zdtm/transition/maps007`), and asserts criu logged `O_DIRECT enabled on pages fd` during restore. The check skips rather than fails where loop mounts or O_DIRECT are unavailable, so it stays inert on runners that cannot provide them.

I verified it locally: with the images on the loop ext4 the restore logs `O_DIRECT enabled on pages fd`, and with the images on tmpfs the same test logs `Failed to set O_DIRECT on pages fd ... Invalid argument` and the asserted marker is absent, so the check distinguishes the two.

The marker proves O_DIRECT was selected for the page reads. The AIO restorer engine (`restore_vma_aio()`) additionally runs only when the restored VMAs take the async vma_io path rather than being premapped; `maps007` exercises that path in practice, though it is not structurally guaranteed across kernels, so the assertion is scoped to the O_DIRECT selection it can prove reliably.

One open question: is `run_non_shardable_tests` the right home for this, or would you prefer it behind its own opt-in flag like the streaming test?


adrianreber:
Reading this I was thinking maybe link the 007 test to 008 and have a custom hook which does the steps before and after the test. This way it would not add that many lines to the main ci script.

rocker-zhang:
> Reading this I was thinking maybe link the 007 test to 008 and have a custom hook which does the steps before and after the test. This way it would not add that many lines to the main ci script.

Thanks, that's a nicer shape — keeping the setup next to the test instead of in the CI script.

Before I rework it, one mechanical point I'd like to get right. What we're covering is the restore-time page-read probe (`probe_pages_o_direct()`), so the coverage needs the **image files** themselves to live on an O_DIRECT-capable fs, not just some scratch mount during the test. With a hook that means putting the image directory on a loop ext4 from dump time onward — e.g. `mkfs.ext4` a loop device and mount it at the image dir on `--pre-dump`, then umount + detach on `--clean` (capability-gated skip when loop/`mkfs.ext4` aren't available, like the current patch).

Two questions so I build the right thing for `maps008` + `maps008.hook`:
1. Does that match what you had in mind, i.e. the hook owns the loop-device/ext4 lifecycle around the image dir?
2. Are you fine with a per-test hook creating a loop device in CI, or would you prefer it reuse an existing O_DIRECT-capable mount when one is available and skip otherwise?

adrianreber:
My personal opinion is that a loop device sounds acceptable. No guarantee, however, that everybody sees it that way. But if you look at different hooks a lot of different things are set up there, so mounting a loop device sounds acceptable.

rocker-zhang:
An update before I push the hook rework, because checking what filesystem the image directory actually has in CI changes the framing here.

The image directory is O_DIRECT-capable on every job that runs the memory tests:
- the container jobs (alpine/archlinux/fedora) bind-mount the repo from the runner (`-v $(pwd)/../../:/criu`), so it is the runner's ext4;
- the aarch64 and other non-container jobs run in the workspace on the runner's ext4;
- the Lima VM jobs `limactl copy` the tree into the VM's own disk (not a virtiofs/sshfs share), so it is the VM root fs (ext4/xfs/btrfs).

So the restore-time probe already runs and selects O_DIRECT on existing memory tests such as maps007 — my original "the default image directory does not support O_DIRECT" premise is wrong. (On kernels >= 6.6 even tmpfs serves O_DIRECT.)

What is missing is not the path running but anything checking it: nothing greps for the marker, so a silent regression in the O_DIRECT selection would leave every job green. So I see two honest options, and I would defer to your preference:

1. Keep maps009 as a deterministic regression guard: pin the image directory to an ext4 loop device and assert the "O_DIRECT enabled on pages fd" marker (a hard assert, not fail-soft). Coverage then does not silently depend on the runner filesystem.
2. Drop the loop device and just assert the marker after one existing maps007 run in run-ci-tests.sh. Simpler, but it would fail rather than skip if a future runner filesystem stopped serving O_DIRECT.

Either way the marker only proves O_DIRECT was selected for the page reads, not that the AIO restorer engine ran. I am happy to go whichever way you prefer, or to close this if the assertion is not worth a dedicated test.

adrianreber:
So the challenge is, how can you verify that O_DIRECT has been used. Currently it seems this is how you do it:
```
if [ "$rc" -eq 0 ] && ! grep -qr "O_DIRECT enabled on pages fd" "$mnt/dump"; then
```

So maybe it would be better to move your code to a separate test in `others/`. I see there are a some tests calling `zdtm.py`, so maybe that would be the way to go. Instead of embedding it in the main CI script, put it in a separate script and just call that one.

rocker-zhang:
Reworked into a standalone test under `test/others/odirect/` as you suggested, so the CI script just calls `make -C test/others/odirect/ run` (next to the other `others/` tests). The test pins the zdtm image directory to a dedicated ext4 loop device, runs `zdtm/transition/maps007` (host flavor, which takes the async vma_io path that reaches the probe), and checks for the `O_DIRECT enabled on pages fd` marker.

To answer "how do you verify O_DIRECT was used" without making it flaky: it skips (not fails) when the loop device, mkfs.ext4 or root aren't available, and also when criu logs that the kernel rejected O_DIRECT for the pages fd at runtime ("Failed to set O_DIRECT" / "O_DIRECT rejected at read time"). It only fails if the marker is absent with no such rejection logged, i.e. an actual regression in the selection. As before, the marker proves O_DIRECT was selected for the page reads, not that the AIO restorer engine ran.

test/others/odirect/run.sh
```diff
@@ -0,0 +1,92 @@
...
+# image directory to a dedicated ext4 loop device and assert the marker, so a
+# regression in the O_DIRECT selection is caught. The marker proves O_DIRECT
+# was selected for the page reads; it does not assert that the AIO restorer
+# engine itself ran.
+#
+# Needs root, the loop module, mkfs.ext4 and a filesystem that actually serves
+# O_DIRECT; skips (exit 0) where any of that is unavailable. zdtm/transition/
+# maps007 (host flavor) takes the async vma_io path that reaches the probe.
+
+set -u
```
adrianreber: For a CI script please also set `-x` and `-e`.

test/others/odirect/run.sh
```diff
@@ -0,0 +1,92 @@
...
+	exit 0
+}
+
+[ "$(id -u)" = 0 ] || skip "needs root"
+command -v mkfs.ext4 >/dev/null 2>&1 || skip "mkfs.ext4 not available"
+modprobe loop 2>/dev/null
+
+img=$(mktemp ./odirect.img.XXXXXX) || skip "cannot create image file"
+mnt=$(mktemp -d ./odirect.mnt.XXXXXX) || skip "cannot create mountpoint"
+truncate -s 1G "$img" || skip "cannot size image file"
```
adrianreber: 1GB sounds pretty large. Is that necessary?

test/others/odirect/run.sh
```diff
@@ -0,0 +1,92 @@
...
+python3 ../../zdtm.py run -t zdtm/transition/maps007 -f h --keep-img always || ret=$?
+if [ "$ret" -ne 0 ]; then
+	echo "FAIL others/odirect: maps007 checkpoint/restore failed (rc=$ret)"
+	exit 1
+fi
+
+# Marker source: criu/pagemap.c "O_DIRECT enabled on pages fd". This is not a
+# streaming restore and the filesystem serves O_DIRECT (probed above), so the
+# marker is expected.
+if grep -qr "O_DIRECT enabled on pages fd" "$mnt/dump"; then
```
adrianreber: What about just looking at the logs and not all files?

test/others/odirect/run.sh
```diff
@@ -0,0 +1,92 @@
...
+if ! dd if=/dev/zero of="$mnt/probe" bs=4096 count=1 status=none 2>/dev/null ||
+	! dd if="$mnt/probe" of=/dev/null bs=4096 count=1 iflag=direct status=none 2>/dev/null; then
+	skip "filesystem does not serve O_DIRECT"
+fi
+rm -f "$mnt/probe"
+
+# Point the zdtm image directory at the O_DIRECT filesystem.
+mkdir -p "$mnt/dump"
+rm -rf "$dumplink"
+ln -sfn "$(readlink -f "$mnt")/dump" "$dumplink"
```
adrianreber: First `rm` and then `ln -f`. Not sure that is necessary.

adrianreber:
Make sure your new script is running through shellcheck when calling `make lint`.

rocker-zhang:
> Make sure your new script is running through shellcheck when calling `make lint`. (and the inline notes on `set -e -x`, the 1G image, grepping all files, and the rm + ln)

Thanks, all addressed:

- `set -e -x` at the top, and the script is now in the `make lint` shellcheck list (`shellcheck -x test/others/odirect/*.sh`); it passes clean.
- The loop image is 256M now (maps007's image is ~33M, so that's comfortable headroom) instead of 1G.
- The marker check greps only the restore logs (`--include='*.log'`) rather than every file.
- Dropped the rm + symlink: it now mounts the loop device directly over the image directory (`test/dump`) and unmounts it on exit, so the directory's original contents are left untouched.

adrianreber:
Thanks. Unfortunately one of our test setups uses busybox (on alpine);

```
+ grep -qrE '--include=*.log' 'Failed to set O_DIRECT|O_DIRECT rejected' ../../dump
grep: unrecognized option: include=*.log
BusyBox v1.37.0 (2026-01-10 15:38:28 UTC) multi-call binary.
```

You need to handle it differently.

rocker-zhang:
> Unfortunately one of our test setups uses busybox (on alpine) [...] You need to handle it differently.

Good catch, thanks — `grep --include` is a GNU extension that busybox lacks. Fixed by dropping `--include` and instead collecting the logs with `find "$dumpdir" -name '*.log'` and feeding each one to a plain `grep -qE` (a small `log_has()` helper), which keeps the "search only the log files" behaviour without relying on the GNU-only flag.

Verified locally: shellcheck is clean, the helper matches the marker under busybox `grep`, and a full `make -C test/others/odirect/ run` passes (the `O_DIRECT enabled on pages fd` marker is produced on restore). Pushed as a fixup to the same commit.

test/others/odirect/run.sh
```diff
@@ -0,0 +1,101 @@
...
+# off). bs=4096 is a valid O_DIRECT alignment on every architecture (criu
+# aligns to the runtime PAGE_SIZE, which is >= 4096).
+if ! dd if=/dev/zero of="$dumpdir/probe" bs=4096 count=1 status=none 2>/dev/null ||
+	! dd if="$dumpdir/probe" of=/dev/null bs=4096 count=1 iflag=direct status=none 2>/dev/null; then
+	skip "filesystem does not serve O_DIRECT"
+fi
+rm -f "$dumpdir/probe"
+
+ret=0
+python3 ../../zdtm.py run -t zdtm/transition/maps007 -f h --keep-img always || ret=$?
```
avagin: please run zdtm.py directly (without python3).

test/others/odirect/run.sh
```diff
@@ -0,0 +1,101 @@
...
+# Some kernels reject O_DIRECT even on ext4 (for example when CONFIG_AIO is
+# off). bs=4096 is a valid O_DIRECT alignment on every architecture (criu
+# aligns to the runtime PAGE_SIZE, which is >= 4096).
+if ! dd if=/dev/zero of="$dumpdir/probe" bs=4096 count=1 status=none 2>/dev/null ||
+	! dd if="$dumpdir/probe" of=/dev/null bs=4096 count=1 iflag=direct status=none 2>/dev/null; then
+	skip "filesystem does not serve O_DIRECT"
+fi
+rm -f "$dumpdir/probe"
+
+ret=0
```
avagin, line 69: ../../zdtm.py run -t zdtm/transition/maps007 -f h --keep-img always
ret=$?
if [ "$ret" ...

test/others/odirect/run.sh
```diff
@@ -0,0 +1,101 @@
...
+# selection would go unnoticed. Mount a dedicated ext4 loop device over the
+# zdtm image directory and assert the marker. The marker proves O_DIRECT was
+# selected for the page reads; it does not assert that the AIO restorer engine
+# itself ran.
+#
+# Needs root, the loop module, mkfs.ext4 and a filesystem that serves O_DIRECT;
+# skips (exit 0) where any of that is unavailable. zdtm/transition/maps007
+# (host flavor) takes the async vma_io path that reaches the probe.
+
+set -e -x
```
avagin: set -e -x -o pipefail

test/others/odirect/Makefile
```diff
@@ -0,0 +1,4 @@
+run:
+	./run.sh
```
avagin: let's run it in a new mount namespace: unshare -m --propagation slave ./run.sh

test/others/odirect/run.sh
```diff
@@ -0,0 +1,101 @@
...
+		umount "$dumpdir" 2>/dev/null || umount -l "$dumpdir" 2>/dev/null || :
+	fi
+	[ -n "$img" ] && rm -f "$img"
+	return 0
+}
+trap cleanup EXIT
+
+[ "$(id -u)" = 0 ] || skip "needs root"
+command -v mkfs.ext4 >/dev/null 2>&1 || skip "mkfs.ext4 not available"
+modprobe loop 2>/dev/null || :
```
avagin: I don't think it is actually required...

rocker-zhang: Removed — `mount -o loop` sets up the loop device itself.

test/others/odirect/run.sh
rocker-zhang: Added `-o pipefail`.

test/others/odirect/Makefile
rocker-zhang: Done — the Makefile now runs `unshare -m --propagation slave ./run.sh`; the loop mount is released on exit, so `cleanup()` only removes the backing image.

test/others/odirect/run.sh
rocker-zhang: Done — invoking `../../zdtm.py` directly now. (Kept the `|| ret=$?` capture so the explicit FAIL message still fires under `set -e`.)

test/others/odirect/Makefile
```diff
@@ -0,0 +1,4 @@
+run:
+	unshare -m --propagation slave ./run.sh
```
dfeigin-nv, line 2: if someone runs ./run.sh w/o the makefile, the ../../dump may be left in a weird state. Maybe add `umount "$dumpdir" 2>/dev/null || true` in cleanup()

avagin: If someone runs it w/o the makefile, it is their responsibility to do a proper clean up. plus, In the script, we have the atexit hook to umount ../../dump.

test/others/odirect/run.sh
```diff
@@ -0,0 +1,101 @@
...
+
+# shellcheck disable=SC2329  # invoked via trap below
+cleanup() {
+	if mountpoint -q "$dumpdir"; then
+		umount "$dumpdir" 2>/dev/null || umount -l "$dumpdir" 2>/dev/null || :
+	fi
+	[ -n "$img" ] && rm -f "$img"
+	return 0
+}
+trap cleanup EXIT
```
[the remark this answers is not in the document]
adrianreber: `mountpoint` seems to exist on alpine which is the minimal image in CI.

Issue #3047, opened 2026-06-14
CLONE dump: process live-migration 


asafpamzn:
## Summary

We'd like to upstream a new CRIU mode that **clones a running process from one machine to another with minimal downtime on the source**, no intermediate disk image required. 

A high level design is at: https://asafpamzn.github.io/criu/clone-dump-design.html. 
This issue gives the short version.

See PR: https://github.com/checkpoint-restore/criu/pull/3049

## Goals

1. **Reduce downtime at the source.** The source process must be responsive almost the entire migration.
2. **Minimize total migration time.** The wall-clock from "start migrating" to "process running on target" should be dominated by physical resources, that is, network bandwidth or CPU, not by design limitations.
3. **No disk needed.** Pages stream directly from source memory to target memory; no intermediate image file is written for pages.

## How it works

CLONE dump introduce a way to clone a process from one machine to the other with a three-phase flow that uses userfaultfd write-protect and `PAGEMAP_SCAN` to track writes while the process keeps running. Conceptually similar to the existing `pre-dump` + `dump` split, but the running phase tracks dirty pages via `UFFD_FEATURE_WP_ASYNC` rather than via soft-dirty/pagemap snapshots.

Note: `pre-dump` did not meet the requirements as it required going over all of the dirty pages during the freeze, which caused a long freeze time for our use case.

### Phase 1 — short freeze, set up tracking
Seize the process, enumerate VMAs, register them with userfaultfd in `UFFDIO_REGISTER_MODE_WP | UFFD_FEATURE_WP_ASYNC` mode, apply `UFFDIO_WRITEPROTECT` to mark the address space write-protected (async), then **unfreeze**. After Phase 1, every page the process writes leaves a soft-dirty trail in pagemap; reads run normally.

### Phase 2 — process runs; pages stream
Multiple scanner threads call `PAGEMAP_SCAN` to find newly-dirty regions. They publish dirty regions to a shared MPMC queue. P3 sender threads pull from the queue, `process_vm_readv` the pages out of the running process, LZ4-compress, and send them over a TLS-protected socket to the target. The target's `clone-receive` daemon receives, decompresses, and buffers the pages in a per-chunk hash table keyed by virtual address. The scanners iterate until the dirty-page count converges below a threshold.

### Phase 3 — final short freeze
[...]

PR #3049, opened 2026-06-14
CLONE: introduce phased live-migration dump (--clone-dump)

asafpamzn:
Issue: https://github.com/checkpoint-restore/criu/issues/3047
Design: https://asafpamzn.github.io/criu/clone-dump-design.html

This is a draft PR, lets continue the discussion at the issue, giving a short description here

Adds a new CRIU mode that clones a running process from one host to another with minimal source downtime. Pages stream over per-thread TCP sockets while the source keeps running; only two short freezes bracket the online transfer.

The change ships with system tests under test/zdtm/static/clone_dump_* that exercise end-to-end dump-and-restore for basic memory, large processes, multi-threaded workloads, file-backed mappings, mmap/munmap churn during dump, mremap, write storms, and several VMA-lifecycle edge cases. Unit tests under test/unit/clone/ cover the internal building blocks: page pool, lock-free SPSC/SPMC/MPSC queues, atomic and batch bitmaps, the unmapped-range tracker, and the optional page-state and hung-page debug trackers.

Assisted-by: claude-code:opus-4-7


adrianreber:
You should rebase your branch instead of introducing merge commits from upstream.

I see there are also unit tests, CRIU already has a very minimal unittest setup. Maybe try to include your unittests there.

asafpamzn:
> You should rebase your branch instead of introducing merge commits from upstream.
> 
> I see there are also unit tests, CRIU already has a very minimal unittest setup. Maybe try to include your unittests there.

sure will do

PR #3052, opened 2026-06-17
[RFC] fsnotify: optimize overlay path resolution during dump

adrianreber:
Optimize fsnotify overlay path resolution from O(watches x directory_size) to O(mounts x directory_size) via two complementary approaches:

1. **File cache fast path** -- During dump, each FD's (kdev, ino) -> path mapping is already resolved by dump_one_file(). Cache these in a hash table so overlay fsnotify resolution can look up watched inodes in O(1) without any directory walk. This also handles cross-process matches (a file watched by process A but only open in process B).

2. **Batched overlay walks** -- For watches that miss the cache, instead of walking the overlay tree once per watch, defer unresolved entries to a pending list grouped by mount. A second pass walks each overlay mount exactly once, matching all pending inodes in a single recursive traversal. This reduces overlay walks from O(watches) to O(mounts), which is typically 1.

3. **Test coverage** -- Extend the inotify_overlayfs ZDTM test with batch file watches (no open FDs) to exercise the batched overlay directory walk code path and verify watches survive C/R.

Follow-up to #3043.


adrianreber:
#2854 could also be related

criu/fsnotify.c
```diff
@@ -266,7 +266,82 @@ static char *find_path_on_overlay(struct mount_info *m,
...
+	dir = opendir_proc(pid, "fd");
+	if (!dir)
+		return NULL;
+
+	while (1) {
+		int fd;
+		struct stat st;
+		char link[PATH_MAX];
+
+		errno = 0;
```
rst0git: Why do we set `errno = 0;` here?

rst0git: If I understand correctly (`man 3 readdir`), the convention is to use `errno` to distinguishing between end-of-directory from an error?
```
de = readdir(dir);
if (!de && errno)
    pr_perror("overlay walk: readdir failed on %s", base);
```

criu/fsnotify.c
```diff
@@ -386,13 +403,30 @@ static char *alloc_openable(pid_t pid, unsigned int s_dev,
...
+				 * walk handle it later.
+				 */
+				if (ovl_mnt_out) {
+					*ovl_mnt_out = m;
+					return ERR_PTR(ERR_NEEDS_OVL_WALK);
 				}
+
+				pr_debug("\t\tFD lookup missed,"
+					 " trying dir walk"
+					 " for %lx\n", i_ino);
```
rst0git: Splitting these messages into multiple lines breaks the ability to grep for them

criu/fsnotify.c
```diff
@@ -564,29 +621,313 @@ int check_open_handle(unsigned int s_dev, unsigned long i_ino,
...
+			if (pe->s_dev == kdev && pe->i_ino == st.st_ino) {
+				pe->f_handle->path = xstrdup(base);
+				if (pe->f_handle->path) {
+					if (root_ns_mask & CLONE_NEWNS) {
+						pe->f_handle->has_mnt_id = true;
+						pe->f_handle->mnt_id =
+							m->mnt_id;
+					}
+					list_del(&pe->list);
+					xfree(pe);
```
rst0git: Should the entry be removed on `xstrdup(base);` failure as well?

criu/fsnotify.c
adrianreber: I tried to add some comments to make it hopefully clearer.

criu/fsnotify.c
adrianreber: Good point. All messages should be on one line.

criu/fsnotify.c
adrianreber: Reworked this slightly.

criu/fsnotify.c
avagin: but you don't check errno, don't handle errors. Do I miss something?

adrianreber: Right. Copy - paste error. There is a similar flow three times in this PR. I added an error message also here.

avagin:
> @avagin Is this the right direction?

This isn't exactly what I mean. In ideal case, we need to collect all unopenable handle/files for all processes and resolve them all at once. Instead of directly parsing /proc/pid/fd, we can iterate through all dumped files.

adrianreber:
> > @avagin Is this the right direction?
> 
> This isn't exactly what I mean. In ideal case, we need to collect all unopenable handle/files for all processes and resolve them all at once. Instead of directly parsing /proc/pid/fd, we can iterate through all dumped files.

How about now.

PR #3054, opened 2026-06-19
criu: introduce statmount

bsach64:
This pull request is in preparation for https://github.com/checkpoint-restore/criu/pull/3036 and https://github.com/checkpoint-restore/criu/pull/2754 which both plan on using the `statmount()` syscall with the `STATMOUNT_BY_FD` flag.

This pull request adds `statmount()` support to CRIU along with the necessary kerndat checks.


criu/kerndat.c
```diff
@@ -1736,6 +1736,94 @@ static int kerndat_has_timer_cr_ids(void)
...
+						STATMOUNT_MNT_POINT |
+						STATMOUNT_FS_TYPE;
+		return 0;
+	}
+
+	if (statmnt) {
+		pr_err("statmount() call unexpectedly succeeded during kerndat check\n");
+		free(statmnt);
+	}
+	return -1;
```
rocker-zhang: Should an unexpected errno here degrade instead of failing kerndat init? Right now anything other than `ENOSYS`/`EINVAL`/`ENOENT` falls through to `return -1`, and `kerndat_init()` turns that into a hard failure — so a seccomp sandbox that returns `EPERM` for the not-yet-allowlisted `statmount(457)` would stop criu from starting, even though the feature is just unavailable.

`kerndat_has_uffd()` already handles this exact case (`EPERM` → `has_uffd = false`, `return 0`), and `kerndat_has_openat2()` treats any unknown errno as "feature off". Since `check_statmount_by_fd()` only `pr_warn`s, the absence looks non-fatal by design, so maybe it could degrade the same way:

```suggestion
	if (statmnt) {
		pr_err("statmount() call unexpectedly succeeded during kerndat check\n");
		free(statmnt);
		return -1;
	}

	pr_info("statmount() is not usable (errno %d), assuming unsupported\n", errno);
	kdat.has_statmount = false;
	kdat.statmount_supported_mask = 0;
	return 0;
```

That keeps the genuinely-impossible "succeeded on an invalid mnt_id" case fatal and degrades everything else. `kerndat_has_statmount_by_fd()` just below (the `ret = -1` around line 1814) has the same shape, though it's less exposed since `statmount()` has already succeeded by the time it runs. Happy to be wrong if there's a reason to keep it strict.

rocker-zhang (commented):
> https://github.com/checkpoint-restore/criu/pull/3054 is good for you to take a look

Thanks for the ping. I checked the statmount syscall numbers across the arch tables, the `STATMOUNT_BY_FD` value and its argument namespace (it shares `0x1` with `STATMOUNT_SB_BASIC` but lives in the separate flags arg, so that's fine), and the `struct statmount` layout against mainline — those all look right to me. One robustness question on the kerndat probe, inline.

criu/include/mount-v2.h
```diff
@@ -57,7 +57,129 @@ static inline long sys_openat2(int dirfd, const char *pathname, struct open_how
...
+	union {
+		__u32 mnt_ns_fd;
+		__u32 mnt_fd;
+	};
+	__u64 mnt_id;
+	__u64 param;
+	__u64 mnt_ns_id;
+};
+
+struct cr_statmount {
```
kolyshkin, line 160: Since you're copying this from the kernel, it makes sense to copy the comment as well. The current one is this:

```c
  /*
   * Structure for getting mount/superblock/filesystem info with statmount(2).
   *
   * The interface is similar to statx(2): individual fields or groups can be
   * selected with the @mask argument of statmount().  Kernel will set the @mask
   * field according to the supported fields.
   * 
   * If string fields are selected, then the caller needs to pass a buffer that
   * has space after the fixed part of the structure.  Nul terminated strings are
   * copied there and offsets relative to @str are stored in the relevant fields.
   * If the buffer is too small, then EOVERFLOW is returned.  The actually used   
   * size is returned in @size.
   */
```

criu/include/mount-v2.h
```diff
@@ -57,7 +57,129 @@ static inline long sys_openat2(int dirfd, const char *pathname, struct open_how
...
+
+#ifndef STATMOUNT_BY_FD
+#define STATMOUNT_BY_FD		0x0000001U /* want mountinfo for given fd */
+#endif
+
+#ifndef MNT_ID_REQ_SIZE_VER1
+#define MNT_ID_REQ_SIZE_VER1	32 /* sizeof second published struct */
+#endif
+
+struct cr_mnt_id_req {
```
kolyshkin, line 136: Similar to https://github.com/checkpoint-restore/criu/pull/3054/changes/db6fe8702d07ef56dbfa0dd88f5c56f5ac93a353#r3575746761

```c
  /* 
   * Structure for passing mount ID and miscellaneous parameters to statmount(2) 
   * and listmount(2). 
   * 
   * For statmount(2) @param represents the request mask. 
   * For listmount(2) @param represents the last listed mount id (or zero). 
   */
```

criu/kerndat.c
```diff
@@ -1736,6 +1736,94 @@ static int kerndat_has_timer_cr_ids(void)
...
+		kdat.statmount_supported_mask = STATMOUNT_SB_BASIC |
+						STATMOUNT_MNT_BASIC |
+						STATMOUNT_PROPAGATE_FROM |
+						STATMOUNT_MNT_ROOT |
+						STATMOUNT_MNT_POINT |
+						STATMOUNT_FS_TYPE;
+		return 0;
+	}
+
+	if (statmnt) {
```
kolyshkin: You can move this check to right after `do_statmount` call and simplify the further checks.

Also, handle the unexpected error codes (and maybe use `switch`):

I.e.
```c
	struct cr_statmount *statmnt = do_statmount(&req, 0);
	if (statmnt) {
		pr_err("statmount() call unexpectedly succeeded during kerndat check\n");
		return -1;
	}

	switch (errno) {
    case ENOSYS:
		...

		return 0;
	case EINVAL:
	case ENOENT:
		...
		return 0;
	default:
		pr_err("statmount() call unexpectedly returned errno %d\n", errno);
		return -1;
	}
}
```

criu/kerndat.c
```diff
@@ -1736,6 +1736,94 @@ static int kerndat_has_timer_cr_ids(void)
...
+	/*
+	 * passing in a invalid mount, the first possible
+	 * valid mount is (1ULL << 31) + 1. So this call
+	 * should always either return EINVAL or ENOENT
+	 * if statmount is supported.
+	 */
+	struct cr_mnt_id_req req = {
+		.size = MNT_ID_REQ_SIZE_VER1,
+		.mnt_id = 1,
+		.param = STATMOUNT_MNT_BASIC
```
kolyshkin: nit: please add `,` at the end of line

criu/mount-v2.c
```diff
@@ -45,6 +45,36 @@ int check_mount_v2(void)
...
+		if (errno != EOVERFLOW)
+			goto out;
+
+		bufsize <<= 1;
+		if (bufsize >= UINT_MAX / 2)
+			goto out;
+	}
+
+out:
+	free(stmnt);
```
kolyshkin: Should we print an error here telling that statmount failed?

bsach64: I would rather the caller log the error
and when `do_statmount()` is called by kerndat (which is expected to fail) we would log an unnecessary error

criu/kerndat.c
bsach64: done

criu/kerndat.c
bsach64: modified both the statmount and STATMOUNT_BY_FD to use `switch` instead

criu/include/mount-v2.h
bsach64: done

criu/include/mount-v2.h
bsach64: done

criu/kerndat.c
bsach64: addressed this and changed to disabling statmount in some cases, instead of failing kerndat

criu/kerndat.c
```diff
@@ -1736,6 +1736,114 @@ static int kerndat_has_timer_cr_ids(void)
...
+		 */
+		kdat.statmount_supported_mask = STATMOUNT_SB_BASIC |
+						STATMOUNT_MNT_BASIC |
+						STATMOUNT_PROPAGATE_FROM |
+						STATMOUNT_MNT_ROOT |
+						STATMOUNT_MNT_POINT |
+						STATMOUNT_FS_TYPE;
+		return 0;
+	default:
+		pr_perror("statmount() call returned unexpected errno, assuming statmount() is not supported");
```
kolyshkin: nit: I guess it makes sense to print the errno value here.

kolyshkin: My bad; perror does just that!

kolyshkin (commented):
LGTM

@rst0git @adrianreber @mihalicyn PTAL

kolyshkin (approved):

criu/kerndat.c
```diff
@@ -1736,6 +1736,114 @@ static int kerndat_has_timer_cr_ids(void)
 	return 0;
 }
 
+static int kerndat_has_statmount(void)
+{
+	/*
+	 * passing in a invalid mount, the first possible
```
rst0git: typo: "an invalid mount"

criu/kerndat.c
```diff
@@ -1736,6 +1736,114 @@ static int kerndat_has_timer_cr_ids(void)
...
+	if (statmnt) {
+		pr_err("statmount() call unexpectedly succeeded during kerndat check\n");
+		free(statmnt);
+		return -1;
+	}
+
+	switch (errno) {
+	case ENOSYS:
+		kdat.has_statmount = false;
+		kdat.statmount_supported_mask = 0;
```
rst0git: Can we move these two lines before `switch (errno)` to avoid duplicating them for `case ENOSYS` and `default`?

criu/kerndat.c
bsach64: fixed

criu/mount-v2.c
```diff
@@ -45,6 +45,36 @@ int check_mount_v2(void)
...
+struct cr_statmount *do_statmount(struct cr_mnt_id_req *req, unsigned int flags)
+{
+	size_t bufsize = 1 << 15;
+	struct cr_statmount *stmnt = NULL, *tmp = NULL;
+	int ret;
+
+	for (;;) {
+		tmp = xrealloc(stmnt, bufsize);
+		if (!tmp)
+			goto out;
```
rst0git: When the loop exits here, `kerndat_has_statmount()` would still use `errno` and show "statmount not supported" via the `default`.

criu/kerndat.c
bsach64: done

criu/mount-v2.c
bsach64: nice catch! added `ENOMEM` handling to the switch case in `kerndat_has_statmount()`, which causes the check to fail

criu/include/mount-v2.h
```diff
@@ -57,7 +57,149 @@ static inline long sys_openat2(int dirfd, const char *pathname, struct open_how
...
+#ifndef STATMOUNT_MNT_UIDMAP
+#define STATMOUNT_MNT_UIDMAP		0x00002000U	/* Want/got uidmap... */
+#endif
+
+#ifndef STATMOUNT_MNT_GIDMAP
+#define STATMOUNT_MNT_GIDMAP		0x00004000U	/* Want/got gidmap... */
+#endif
+
+#ifndef STATMOUNT_BY_FD
+#define STATMOUNT_BY_FD		0x0000001U /* want mountinfo for given fd */
```
rst0git: It looks like this value in the kernel has one more `0`?
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/include/uapi/linux/mount.h#n243

compel/arch/riscv64/plugins/std/syscalls/syscall.def
```diff
@@ -124,3 +124,4 @@ move_mount          		429	429	(int from_dfd, const char *from_pathname, int to_d
 open_tree			428	428	(int dirfd, const char *pathname, unsigned int flags)
 openat2				437	437	(int dirfd, char *pathname, struct open_how *how, size_t size)
 membarrier 			283	283	(int cmd, unsigned int flags, int cpu_id)
+statmount 			457     457     (struct mnt_id_req *req, struct statmount *smbuf, size_t bufsize, unsigned long flags)
```
rst0git: Whitespace inconsistencies between both `457`

compel/arch/x86/plugins/std/syscalls/syscall_64.tbl
```diff
@@ -121,3 +121,4 @@ __NR_pidfd_getfd		438		sys_pidfd_getfd		(int pidfd, int targetfd, unsigned int f
 __NR_rseq       		334		sys_rseq		(void *rseq, uint32_t rseq_len, int flags, uint32_t sig)
 __NR_membarrier 		324		sys_membarrier		(int cmd, unsigned int flags, int cpu_id)
 __NR_map_shadow_stack		453		sys_map_shadow_stack	(unsigned long addr, unsigned long size, unsigned int flags)
+__NR_statmount			457		sys_statmount		(struct mnt_id_req *req, struct statmount *smbuf, size_t bufsize, unsigned long flags)
```
rst0git: It looks like the kernel is using `unsigned int flags`

bsach64: fixed, the man page incorrectly mentions `unsigned long` (https://man7.org/linux/man-pages/man2/statmount.2.html) (i'll send a patch later on to fix it)

compel/arch/riscv64/plugins/std/syscalls/syscall.def
bsach64: fixed

criu/include/mount-v2.h
bsach64: fixed

criu/mount-v2.c
avagin: in this case, this function should always return a reasonable errno.

criu/kerndat.c
```diff
@@ -1736,6 +1736,115 @@ static int kerndat_has_timer_cr_ids(void)
...
+	 * valid mount is (1ULL << 31) + 1. So this call
+	 * should always either return EINVAL or ENOENT
+	 * if statmount is supported.
+	 */
+	struct cr_mnt_id_req req = {
+		.size = MNT_ID_REQ_SIZE_VER1,
+		.mnt_id = 1,
+		.param = STATMOUNT_MNT_BASIC,
+	};
+	struct cr_statmount *statmnt = do_statmount(&req, 0);
```
avagin: I think we can pass 0xffffffff flags and expect EINVAL. Flags should be always validate as the first step.

criu/kerndat.c
```diff
@@ -1736,6 +1736,115 @@ static int kerndat_has_timer_cr_ids(void)
...
+		.param = STATMOUNT_SUPPORTED_MASK,
+	};
+	char tempfile[PATH_MAX] = "/tmp/.criu.kerndat.XXXXXX";
+
+	if (!kdat.has_statmount) {
+		kdat.has_statmount_by_fd = false;
+		return 0;
+	}
+
+	tempfd = mkstemp(tempfile);
```
avagin: We don't need to create file here. Let's just trigger EBADF.

criu/kerndat.c
bsach64: done, this is cleaner

criu/kerndat.c
bsach64: changed to triggering EBADF

criu/mount-v2.c
```diff
@@ -45,6 +45,36 @@ int check_mount_v2(void)
 	return 0;
 }
 
+struct cr_statmount *do_statmount(struct cr_mnt_id_req *req, unsigned int flags)
+{
+	size_t bufsize = 1 << 15;
```
avagin: why do we use this buf size by default? Is iti really required to use a dynamic buffer here?

bsach64: we only need a dynamic buffer if we want to get string fields, so i have removed the `do_statmount()` helper and am now passing an object on the stack for the check.

avagin:
Why don't we use statmount_supported_mask reported by statmount?

bsach64:
> Why don't we use statmount_supported_mask reported by statmount?

Need to call `statmount()` with a valid `mnt_id` or `mnt_fd` in order to use `STATMOUNT_SUPPORTED_MASK`.
Also `STATMOUNT_SUPPORTED_MASK` was added later on in 6.15, so we can have kernels where statmount is supported but STATMOUNT_SUPPORTED_MASK is not.

Before trying to trigger EBADF, when the code was using a valid fd, I did use `STATMOUNT_SUPPORTED_MASK` to get supported mask in the `has_statmount_by_fd` check, but i figured no new mask was added since `STATMOUNT_BY_FD` so I thought it's okay to trigger EBADF.

avagin:
> > Why don't we use statmount_supported_mask reported by statmount?
> 
> Need to call `statmount()` with a valid `mnt_id` or `mnt_fd` in order to use `STATMOUNT_SUPPORTED_MASK`. Also `STATMOUNT_SUPPORTED_MASK` was added later on in 6.15, so we can have kernels where statmount is supported but STATMOUNT_SUPPORTED_MASK is not.

I think it should be ok. If STATMOUNT_SUPPORTED_MASK isn't supported, we can set a default mask.

> 
> Before trying to trigger EBADF, when the code was using a valid fd, I did use `STATMOUNT_SUPPORTED_MASK` to get supported mask in the `has_statmount_by_fd` check, but i figured no new mask was added since `STATMOUNT_BY_FD` so I thought it's okay to trigger EBADF.

If STATMOUNT_SUPPORTED_MASK is available, we can use it and we don't need an explicit check for STATMOUNT_BY_FD, do we? STATMOUNT_BY_FD has been added after STATMOUNT_SUPPORTED_MASK, hasn't it?

bsach64:
> I think it should be ok. If STATMOUNT_SUPPORTED_MASK isn't supported, we can set a default mask.

@avagin I don't fully understand what you mean here and my previous explanation was poor. We need a valid mnt_id_unique which can be obtained either from `statx()` or `listmount()`, these are different from regular `mnt_id`. We don't have support for `statx()` or `listmount()` in CRIU, so we can't do this without `STATMOUNT_BY_FD` anyways.

> If STATMOUNT_SUPPORTED_MASK is available, we can use it and we don't need an explicit check for STATMOUNT_BY_FD, do we? STATMOUNT_BY_FD has been added after STATMOUNT_SUPPORTED_MASK, hasn't it?

No, `STATMOUNT_SUPPORTED_MASK` tells what bits can be set in `mnt_id_req.param`.
`STATMOUNT_BY_FD` is a _flag_ and `STATMOUNT_SUPPORTED_MASK` will not report for its support.
Please see man page here: https://man7.org/linux/man-pages/man2/statmount.2.html

This discussion has made me realize it is useless to check for statmount() separately, since both of the PRs will rely on `STATMOUNT_BY_FD` we should directly check for support with that flag. I pass a valid fd to it now so that we get `smbuf.supported_mask`

avagin:
> > I think it should be ok. If STATMOUNT_SUPPORTED_MASK isn't supported, we can set a default mask.
> 
> @avagin I don't fully understand what you mean here and my previous explanation was poor. We need a valid mnt_id_unique which can be obtained either from `statx()` or `listmount()`, these are different from regular `mnt_id`. We don't have support for `statx()` or `listmount()` in CRIU, so we can't do this without `STATMOUNT_BY_FD` anyways.

It should be ok to call statx from criu. We don't need any special support for it. It is just a system call. As far as I remember it can't be disabled by a kernel config option. Basically, it startx returns ENOSYS, we can desclare that statemount isn't supported. Does it make sense?

> 
> > If STATMOUNT_SUPPORTED_MASK is available, we can use it and we don't need an explicit check for STATMOUNT_BY_FD, do we? STATMOUNT_BY_FD has been added after STATMOUNT_SUPPORTED_MASK, hasn't it?
> 
> No, `STATMOUNT_SUPPORTED_MASK` tells what bits can be set in `mnt_id_req.param`. `STATMOUNT_BY_FD` is a _flag_ and `STATMOUNT_SUPPORTED_MASK` will not report for its support. Please see man page here: https://man7.org/linux/man-pages/man2/statmount.2.html

My bad. You are right. However, we don't need to run statmount with STATMOUNT_BY_FD to get supported_mask.

> 
> This discussion has made me realize it is useless to check for statmount() separately, since both of the PRs will rely on `STATMOUNT_BY_FD` we should directly check for support with that flag. I pass a valid fd to it now so that we get `smbuf.supported_mask`

I think we may want to  fall back to statemount by mntid to get cgroupfs options, if STATMOUNT_BY_FD isn't supported.

avagin:
LGTM. Thanks!

PR #3057, opened 2026-06-22
criu: support rootless container C/R in non-initial user namespaces

arcceus:
This draft collects the CRIU-side changes needed for checkpoint/restore when the
restored process tree belongs to a user namespace created and managed outside
CRIU.

The original `--rootless-container` switch was dropped. The current approach is
resource-local: CRIU uses `--join-ns user`, namespace capability/probe results,
non-initial userns checks, unprivileged mode, and syscall failures to decide
which restore path is needed.

## Architecture

1. Userns-aware namespace helpers

   Some restore operations need credentials from the target user namespace before
   they can enter or operate on another namespace. CRIU now has small broker
   helpers for those cases. The broker enters the target user namespace first,
   enters the resource namespace when needed, performs the operation there, and
   passes file descriptors or results back to the main restore flow.

   This is used for network namespace socket handling, network locking, mount
   operations, and UTS restore.

2. Privilege-aware restore paths

   Restore paths that can fail in a non-initial user namespace now either use
   the target namespace credentials or tolerate the failure when CRIU cannot
   safely manage host-owned state.

   In particular:

   - mount `uid=`/`gid=` options are translated through the restored userns maps
   - denied restore mounts can be retried through the mount namespace broker
   - host/runtime-owned binfmt_misc and cgroup state can be skipped in
     unprivileged userns contexts
   - pipe resize `EPERM` is tolerated in userns-limited contexts
   - UTS hostname/domainname restore can run with target userns credentials
   - `RLIMIT_NOFILE` restore avoids treating an unprivileged raise failure as
     fatal when the current limit is already sufficient

3. Seccomp BPF through RPC

   The kernel may deny `PTRACE_SECCOMP_GET_FILTER` across user namespace
   boundaries. For restore, CRIU can accept an externally supplied seccomp BPF
   program over RPC and restore that filter instead of depending on ptrace access
   to the dumped task's filter bytes.

4. Runtime-provided user namespace restore

   CRIU can restore into a user namespace created by the caller using
   `--join-ns user`. This keeps namespace lifecycle ownership with the container
   tooling while allowing CRIU to restore the process tree inside that namespace.

## Test coverage

CRIU-owned coverage added or reused in this series:

- `zdtm/static/mapfile_priv`
- `zdtm/static/rootless_mount_opts_id`
- `zdtm/static/rootless_runtime_userns`
- `zdtm/static/rootless_runtime_userns_broker`
- `zdtm/static/rootless_sysv_shmem`
- `zdtm/static/rootless_sysv_shmem_map_files`
- `zdtm/static/rootless_uts_rlimit`
- `test/others/libcriu/test_rpc_seccomp_bpf`
- `test/others/libcriu/test_rpc_seccomp_bpf_restore`
- `test/others/libcriu/test_rpc_join_ns_errors`

## Current boundaries

CRIU restores the process tree and its kernel resources inside the supplied user
namespace, but it does not take ownership of the higher-level container lifecycle
around that namespace.

In particular:

- Cgroup placement remains owned by the caller/container manager. CRIU may skip
  cgroupfs descriptors that refer to host- or runtime-owned cgroup state in an
  unprivileged user namespace.

- Network namespace lifecycle remains owned by the caller/container manager.
  CRIU can restore sockets and network state inside the target namespace, but
  external plumbing such as port-forwarding helpers, host-side sockets, and
  teardown/recreation of the namespace wrapper are outside CRIU's ownership.

- Seccomp filters installed only inside the application at runtime may be
  unreadable across user namespace boundaries. CRIU can restore a filter supplied
  by the caller over RPC, but it cannot reconstruct filter bytecode that the
  kernel refuses to expose.

## Notes

This PR is still draft while the individual pieces are reviewed and split into
smaller upstreamable changes.


compel/src/lib/infect.c
```diff
@@ -307,8 +307,12 @@ int compel_wait_task(int pid, int ppid, int (*get_status)(int pid, struct seize_
 		return -1;
 	}
 
-	if (ss->seccomp_mode != SECCOMP_MODE_DISABLED && ptrace_suspend_seccomp(pid) < 0)
-		goto err;
+	if (ss->seccomp_mode != SECCOMP_MODE_DISABLED) {
+		if (ptrace_suspend_seccomp(pid) < 0) {
+			pr_warn("Unable to suspend seccomp for %d, memory will be read via process_vm_readv\n", pid);
```
avagin: it isn't just about memory. We can a few other syscalls that can be blocked by seccomp. As for process_vm_read instead of splice, it should be customized via arguments.  We probably will need to consider switching to process_vm_read for all workloads: https://lore.kernel.org/all/20260531010107.1953702-1-safinaskar@gmail.com/#t

arcceus: Yeah I originially form the start had podman's default seccomp profile in mind and only vmsplice was not in the allowlist will add proper error logging for the rest of the rpc call sites, if they fail. Also will add argument like --force-process-vm-readv or something.

criu/filesystems.c
```diff
@@ -268,6 +269,11 @@ int binfmt_misc_dump_sandboxed(pid_t pid, BinfmtMiscEntry ***pb_bmes)
 		return 0;
 
 	if (!(root_ns_mask & CLONE_NEWUSER)) {
+		if (opts.rootless_container) {
```
arcceus: I had the wrong idea here, binfmt_misc registrations are host state, not containers removing from limitations

criu/mount.c
```diff
@@ -2158,6 +2274,52 @@ int mount_root(void *args, int fd, pid_t pid)
...
+static int remount_bind_flags(struct mount_info *mi, unsigned long mflags)
+{
+	if (!mflags)
+		return 0;
+
+	if (!mount(NULL, service_mountpoint(mi), NULL, MS_BIND | MS_REMOUNT | mflags, NULL))
+		return 0;
+
+	if ((opts.unprivileged || opts.rootless_container) && opts.mode == CR_RESTORE && (errno == EPERM || errno == EACCES)) {
+		pr_warn("mnt: re-mount EPERM at %s in unprivileged restore, per-mount flags dropped (nosuid,noexec,nodev,etc.)\n",
```
arcceus: per mount flags survive c/r of rootless podman tmpfs host binds preserve flags so do other volumes dropping this from limitations aswell

criu/seccomp.c
```diff
@@ -149,6 +149,18 @@ static int collect_filter(struct seccomp_entry *entry)
...
+				 * userns boundary. The filter count
+				 * is available via /proc/pid/status
+				 * (Seccomp_filters), but the filter
+				 * program bytes themselves are
+				 * unreachable without ptrace. On
+				 * restore, only the OCI-spec filter
+				 * will be installed; any additional
+				 * layers from sidecars or admission
+				 * controllers are silently dropped.
+				 */
```
arcceus, line 166: hey @avagin is there existing lkml work on migrating seccomp_get_filter to use checkpoint_restore_ns_capable instead of capable(CAP_SYS_ADMIN)? currently the only userspace recovery paths i see are the oci spec or user explicitly passing --security-opt seccomp= in podman which flows into the same path any runtime installed filters are just not recoverable rootless

criu/config.c
```diff
@@ -1168,13 +1168,8 @@ int check_options(void)
...
 	}
 
-	if (opts.rootless_container && opts.mode != CR_DUMP && opts.mode != CR_RESTORE) {
-		pr_err("--rootless-container is only valid for dump and restore\n");
-		return 1;
-	}
-
-	if (opts.rootless_container && opts.swrk_restore) {
-		pr_err("--rootless-container is incompatible with swrk restore\n");
+	if (opts.rootless_container && opts.mode != CR_DUMP && opts.mode != CR_PRE_DUMP && opts.mode != CR_RESTORE) {
```
rst0git: @arcceus Could you try to squash the fixed into the previous commits?
(e.g., using `git rebase` & `git push -f`)

criu/config.c
```diff
@@ -1064,6 +1096,15 @@ int parse_options(int argc, char **argv, bool *usage_error, bool *has_exec_cmd,
...
 	}
 
+	/*
+	 * Rootless CRIU cannot directly switch_ns(CLONE_NEWNET) because the broker
+	 * handles userns-first netns entry. Default to nftables, which is what
+	 * modern rootless container networking (netavark) uses. The user can
+	 * still override via --network-lock iptables|nftables|skip.
+	 */
+	if (opts.rootless_container && !has_network_lock_opt)
+		opts.network_lock_method = NETWORK_LOCK_NFTABLES;
```
adrianreber: I am unsure about this change. The network locking is something which relates to the system and not the use case. On an nftables based system it should be set to `NETWORK_LOCK_NFTABLES`. It doesn't make much sense to me that CRIU would be built for iptables based locking but for Podman it should use nftables. I would say this can be dropped.

arcceus: Dropped it. My reasoning was since netavark uses nftable by default auto-selecting it on --rootless-container made sense.

criu/config.c
arcceus: Done squashed the fixes to their original commits.

compel/src/lib/infect.c
avagin: let's add --memory-dump-mode or something like that. It should be a generic variant for --pre-dump-mode. Under the hood, it can be just an alias to --pre-dump-mode. Please create another pr for this feature.

arcceus: Done once #3073 gets merged will drop the --force-process-vm-readv commit from this branch.

criu/config.c
```diff
@@ -704,6 +704,8 @@ int parse_options(int argc, char **argv, bool *usage_error, bool *has_exec_cmd,
 		BOOL_OPT("mntns-compat-mode", &opts.mntns_compat_mode),
 		BOOL_OPT("unprivileged", &opts.unprivileged),
 		BOOL_OPT("rootless-container", &opts.rootless_container),
+		{ "userns-uid-map", required_argument, 0, 1101 },
```
avagin: have you consider to use an external user namespace? A container runtime creates a user namespace and provide it to criu as an external resource.

criu/include/cr_options.h
```diff
@@ -251,6 +251,18 @@ struct cr_options {
...
+	/*
+	 * User-provided UID/GID map overrides for rootless container
+	 * restore. Format: first:lower_first:count.
+	 * If empty, auto-detect from /etc/subuid and /etc/subgid.
+	 */
+	struct {
+		unsigned int first;
+		unsigned int lower_first;
+		unsigned int count;
+	} uid_map_extents[8], gid_map_extents[8];
```
avagin: 8?

criu/seccomp.c
avagin: No, there is not. It is interesting why seccomp_get_filter is guarded with a capability check.

criu/include/cr_options.h
arcceus: Arbitrary, It's enough for rootless maps(1-2 extents) doesn't really reflects to any kernel limits or anything.

criu/config.c
arcceus: I haven't actually let me take a look.

arcceus: Can be done OCI spec does have uid/gid, and crun does apply them already on creation but its just moving all this work to every runtime that supports rootless c/r. I don't really see any strong benefit of moving this to runtime. It does look cleaner thoo. Do you have anything specific in mind regarding this or is it more of a long term direction?

adrianreber: We do this is also with network namespace with Podman and the Kubernetes use case. For the network namespace it has proofed to be the better approach if the container runtime/engine sets up the network namespace and CRIU just joins it. Network namespaces can probably be much more complicated the user namespaces, but the container tooling knows how to set it up for a new container so that joining it seems much simpler.

avagin: Podman probably will want to create a network namespace in a container user namespace, so it might be another reason why we may want to create a user namespace in the container runtime tool.

arcceus: I looked into this a bit further and worked on --userns=auto side of things it was kind of broken. Addressing the netns ownership problem, podman already had some work on post network namespace configuration it just wasnt wired up properly for rootless restore. With default podman create, netns is created under podman's userns and with --userns=auto flag runtime creates a different userns according to the oci mapping. It already handles what @avagin is saying the netns is owned by the container userns and the PostConfigureNetNS function correctly deals with configuring the netns from podman side. Im starting the work on the external userns part because it is infact cleaner alternative. Thank you for suggesting it!

adrianreber:
I was thinking maybe it would make sense to split this PR. In the beginning maybe focus less on the Podman integration. Just try to enhance the non root support with tests. Would that work?

rst0git:
> In the beginning maybe focus less on the Podman integration. Just try to enhance the non root support with tests.

This sounds like a good idea - Thanks Adrian!

arcceus:
> I was thinking maybe it would make sense to split this PR. In the beginning maybe focus less on the Podman integration. Just try to enhance the non root support with tests. Would that work?

@adrianreber Should I include #3073 in this new PR aswell?

adrianreber:
> > I was thinking maybe it would make sense to split this PR. In the beginning maybe focus less on the Podman integration. Just try to enhance the non root support with tests. Would that work?
> 
> @adrianreber Should I include #3073 in this new PR aswell?

Not sure. Small PRs are easier to review so if it is not totally necessary and makes sense make it independent.

arcceus:
@adrianreber I have opened these 2 PR's for now, are stacked on #3073 and #3100. Both split from here.

PR #3058, opened 2026-06-23
docs: Clarify memory consistency in pre-dump documentation

avagin:
Correct the documentation regarding `vmsplice` and `SPLICE_F_GIFT` behavior.  `SPLICE_F_GIFT` does not enforce Copy-on-Write (COW) when the user-space process modifies the gifted memory. Instead, it is a zero-copy mechanism where post-resume modifications can lead to inconsistent intermediate dumps.

Explain how CRIU handles this inconsistency via its iterative design, relying on the soft-dirty tracker to catch post-resume modifications and re-dump them in subsequent iterations, ensuring final consistency.

Also emphasize that the `splice` scheme relies heavily on `vmsplice` performance to minimize freeze time, making the migration almost invisible to the process.


rst0git (approved):
LGTM

PR #3063, opened 2026-06-26
github: bump up the ubuntu version for arm jobs

avagin:


scripts/ci/run-ci-tests.sh
```diff
@@ -28,6 +28,15 @@ if [ "$UNAME_M" != "x86_64" ]; then
...
 fi
 
+chmod_parents_o_x() {
+	local d="$(pwd -P)"
+
+	while [ "$d" != "/" ]; do
+		chmod o+x "$d" || return 1
+		d=$(dirname "$d")
+	done
+}
```
adrianreber, line 40: I see copilot already complained about this. I guess you closed this comment because `run-ci-tests.sh` is not supposed to be run on non-CI systems, right?

avagin: You are right. It is never supposed to be run on non-ci systems.

adrianreber:
ShellCheck is not happy with one of your changes and the Copilot comments do not look wrong. Besides that this PR would be good to have as the Ubuntu 26.04 kernel seems to fix the `socket-tcpbuf*` errors we have been seeing.

adrianreber:
It would be nice to fix the typo "bupm" in the first commit. Any maybe add that commit as the last to first have the commits that fix the errors introduce by bumping it to 26.04 and then the actual bump.

scripts/ci/run-ci-tests.sh
```diff
@@ -65,6 +76,8 @@ ci_prep () {
 	contrib/dependencies/apt-packages.sh
 	contrib/apt-install "${CI_PKGS[@]}"
 	chmod a+x "$HOME"
+	chmod_parents_o_x
```
rst0git, line 79: Out of curiosity, were any tests failing before this change?

avagin: For example, unix tests want to bind sockets by absolute paths. These tests fail without this change.

Issue #3053, opened 2026-06-18
#3022 introduced a performance regression


avagin:
This pr introduced a major performance regression. I haven't look at it closely to figure out what exactly goes wrong:
```
root@criu-next-2:/home/avagin/git/criu# python3 test/zdtm.py run -t zdtm/static/maps04 -f h
userns is supported
=== Run 1/1 ================ zdtm/static/maps04
========================= Run zdtm/static/maps04 in h ==========================
 DEP       maps04.d
 CC        maps04.o
 LINK      maps04
Start test
./maps04 --pidfile=maps04.pid --outfile=maps04.out
Run criu dump
Execution time: 4.971509 seconds
Run criu restore
==== ALARM ====
Execution time: 56.193118 seconds
Send the 15 signal to  84
Wait for zdtm/static/maps04(84) to die for 0.100000
Wait for zdtm/static/maps04(84) to die for 0.200000
Removing dump/zdtm/static/maps04/84
========================= Test zdtm/static/maps04 PASS =========================
root@criu-next-2:/home/avagin/git/criu# git diff
diff --git a/test/zdtm.py b/test/zdtm.py
index 6ffc2617807a..05256ef331e1 100755
--- a/test/zdtm.py
+++ b/test/zdtm.py
@@ -1297,8 +1297,12 @@ class criu:
 
         preload_libfault = self.__preload_libfault and action in ['dump', 'pre-dump', 'restore']
 
+        start_time = time.perf_counter()
         ret = self.__criu.run(action, s_args, self.__criu_bin, self.__fault,
                               strace, preexec, preload_libfault, nowait)
+        end_time = time.perf_counter()
+        execution_time = end_time - start_time
+        print(f"Execution time: {execution_time:.6f} seconds")
 
         if nowait:
             os.close(status_fds[1])
diff --git a/test/zdtm/static/maps04.c b/test/zdtm/static/maps04.c
index dbbb5c644029..ec58091d5dd3 100644
--- a/test/zdtm/static/maps04.c
+++ b/test/zdtm/static/maps04.c
@@ -11,7 +11,7 @@
 #include <linux/limits.h>
 #include "zdtmtst.h"
 
-#define MEM_SIZE (1L << 29)
+#define MEM_SIZE (8UL << 30)
 
 const char *test_doc = "Test big mappings";
 const char *test_author = "Andrew Vagin <avagin@openvz.org";
[...]

PR #3066, opened 2026-06-28
criu: add an --image-io-mode option to control O_DIRECT for the pages image

dfeigin-nv:
PR #3022 switched restore to read the pages image with O_DIRECT unconditionally. O_DIRECT helps a cold or cross-host restore, but on a same-host restore it bypasses the still-warm page cache and reads from disk, which regressed maps04 ( #3053 ).

This PR adds an `--image-io-mode` option to choose the pages-image I/O method:
```
  --image-io-mode writeback  (default) buffered I/O; the pre-#3022 behaviour, so the regression is gone by default.
  --image-io-mode direct     O_DIRECT; bypasses the page cache.
```

It can be set on the command line or over RPC, and is chosen independently at dump and restore time — the image bytes are identical either way.

- **Restore** reads the pages image with O_DIRECT only in `direct` mode (gated at the read-probe call sites in `open_page_read_at()` and `mem.c`); it falls back to buffered I/O if the filesystem rejects O_DIRECT.
- **Dump** sets O_DIRECT on the pages-image fd and lets the existing `splice()` in `write_pages_loc()` do the write. `splice()` to a file goes through `iter_file_splice_write()` and honours O_DIRECT on filesystems that support it (ext4, xfs); where it does not, `splice()` fails with `EFAULT`/`EINVAL`, and `write_pages_loc()` clears O_DIRECT and retries the remainder buffered. No bounce buffer.

Commits:
  1. add the `--image-io-mode` option and gate the restore-side O_DIRECT read (CLI / RPC / libcriu / man).
  2. write the pages image with O_DIRECT on dump, via `splice()` + buffered fallback.
  3. exercise `--image-io-mode=direct` in the zdtm suite (maps04, maps00 `--rpc`).

Tested on maps04/maps00 (aarch64, ext4): default writeback restores at the pre-#3022 speed; `--image-io-mode=direct` restores correctly; dumping in one mode and restoring in the other works, so the mode is an I/O choice, not a property of the image. Verified that the dump-side O_DIRECT actually bypasses the page cache (pages image 0% resident after a `direct` dump vs 100% under writeback), and that the `splice()` → buffered fallback produces a valid, restorable image when the filesystem rejects O_DIRECT. zdtm.py gained an `--image-io-mode` passthrough, and run-ci-tests.sh runs maps04 and maps00 (incl. `--rpc`) with `--image-io-mode=direct`, so the O_DIRECT paths and the RPC handler are exercised in CI.


adrianreber:
I am not sure `--cache` is the most descriptive name. Using it because qemu uses it is a valid approach, but maybe having a name that is better to understand as this is more about direct IO or not. Just adding this here as an open question. Not a blocker.

criu/config.c
```diff
@@ -593,6 +594,20 @@
 	return -1;
 }
 
+static int parse_cache_mode(struct cr_options *opts, const char *optarg)
```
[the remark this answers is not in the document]
adrianreber: This seems to be the default in this file. All functions do that here.

images/rpc.proto
```diff
@@ -60,6 +60,11 @@ enum criu_pre_dump_mode {
 	VM_READ =	2;
 };
 
+enum criu_cache_mode {
```
adrianreber: This could also be called just cache_mode. I see some parameters are prefixed with `criu_` and some aren't. Doesn't seem necessary.

Documentation/criu.txt
```diff
@@ -713,6 +728,21 @@ The 'mode' may be one of the following:
                 build-ID cannot be obtained, 'chksm-first' method will be
                 used. This is the default if mode is unspecified.
 
+*--cache* ['mode']::
+    Set the cache mode used when writing and reading the memory pages
+    image. It controls whether the host page cache is used. The same mode
+    should be given to both *dump* and *restore*.
```
rst0git: It would be better to save the cache_mode value in the inventory image during dump, and load it during restore instead.

See `network_lock_method` as an example: https://github.com/checkpoint-restore/criu/commit/ca3e3c50bec92390a6d57f827f9685b48ec6200c

adrianreber: I think it would be important that it still can be changed from the command line or via RPC.

It is slightly different than network locking I think.

rst0git: > The same mode should be given to both *dump* and *restore*.

My understanding was that the cache mode cannot be changed during restore.

adrianreber: Ah, okay. My understanding was it can be changed anytime.

avagin: > My understanding was that the cache mode cannot be changed during restore.

This isn't true. It only affects how images are written to or read from storage. There is no reason to save it in the inventory. I think we can rename this option to --image-io-mode, since `cache` sounds too generic.

images/rpc.proto
dfeigin-nv: it seems that the enums are always prefixed with `criu_`.

adrianreber: Right. In `rpc.proto` they are all prefixed. Probably better to be consistent.

criu/pagemap.c
```diff
@@ -807,6 +807,13 @@ int probe_pages_o_direct_read(int fd)
 	void *probe = NULL;
 	ssize_t probe_ret;
 
+	/*
+	 * In writeback mode the pages fd is never switched to O_DIRECT,
+	 * so callers keep reading through the host page cache.
+	 */
+	if (opts.image_io_mode != IMAGE_IO_DIRECT)
```
avagin: I think this check should be in open_page_read_at(). probe_pages_o_direct_read probably will need to print a warning if the direct mode isn't supported (as it is explicitly requested by a user).

criu/pagemap.c
```diff
@@ -862,6 +862,74 @@ int probe_pages_o_direct_read(int fd)
...
+			pr_perror("Failed to rewind pages fd %d after O_DIRECT probe", fd);
+			return -1;
+		}
+		pr_debug("O_DIRECT enabled on pages write fd %d\n", fd);
+		return 1;
+	}
+
+	/* A partial write (probe_ret >= 0) or a non-EINVAL error means O_DIRECT failed. */
+	if (probe_ret >= 0 || errno != EINVAL) {
+		pr_perror("O_DIRECT write probe failed on pages fd %d", fd);
```
avagin: probe_ret >= 0 should be handled separatly, errno isn't set in this case.

criu/page-xfer.c
```diff
@@ -252,15 +253,75 @@ static int open_page_server_xfer(struct page_xfer *xfer, int fd_type, unsigned l
...
 /* local xfer */
+/*
+ * Bounce-buffer size for O_DIRECT page writes. 1 MiB is a multiple of any
+ * PAGE_SIZE, so each write stays block-aligned.
+ */
+#define O_DIRECT_XFER_CHUNK (1 << 20)
+
+/*
+ * splice() only routes writes through the O_DIRECT path on recent kernels; on
+ * older kernels (still supported by criu) it does not bypass the page cache.
```
avagin: In which kernel version was it added? On older kernels, we can fall back to non-direct writes. I don't like the idea of using a bounce buffer. With one, I don't think a direct write will ever be faster than a normal write followed by an fdatasync before closing the file descriptor.

criu/pagemap.c
dfeigin-nv: Indeed

criu/pagemap.c
dfeigin-nv: Done

criu/page-xfer.c
dfeigin-nv: 6.5. And sure I'll make it so if splice returns an error we fallback to buffered.

avagin:
LGTM. Thanks.

rst0git (approved):
LGTM

PR #3069, opened 2026-06-28
mem: properly check for COW vmas in case of no ac

bsach64:
This pull request ports the suggestion mentioned here: https://github.com/checkpoint-restore/criu/pull/3037#discussion_r3437235630

I honestly should've been more careful providing a review to https://github.com/checkpoint-restore/criu/pull/3037 considering I am not very familiar with that part of the codebase and memory management in general

In any case, have tried to properly understand and document this fix from @avagin.


rst0git:
@bsach64 Thanks! I've updated the patch with a few minor changes.

PR #3071, opened 2026-06-29
cgroup: fix stack buffer overflow in ctrl_dir_and_opt()

adrianreber:
ctrl_dir_and_opt() builds comma-separated controller directory and mount option strings by repeatedly appending via snprintf() into stack-allocated buffers.  The offsets (doff, ooff) are advanced by the snprintf() return value, but snprintf() returns the number of bytes that *would* have been written on truncation, not the number actually written.  When a controller name is long enough to cause truncation, the offset advances past the buffer end.  On the next iteration, the remaining-length calculation (e.g. ds - doff) goes negative and is implicitly cast to a large size_t, disabling the truncation guard entirely.  Subsequent snprintf() calls and the trailing-comma removal then write out of bounds on the stack (CWE-121: Stack-based Buffer Overflow).

A crafted cgroup.img with oversized cnames entries triggers this during restore in any code path that calls ctrl_dir_and_opt(): prepare_cgns(), move_in_cgroup(), prepare_cgroup_properties(), prepare_cgroup_sfd(), and cgroupd().

Fix this by checking every snprintf() call for truncation before advancing the offset.  Each append now verifies that remaining space is positive and that snprintf() did not return a value greater than or equal to the remaining space.  On any overflow the function returns -1 and all five callers now propagate this error.  Also change the type of `off` in prepare_cgroup_properties() from unsigned int to int so the -1 return is not silently converted to a large positive value.

This vulnerability was identified and analyzed by an AI security analysis tool.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1058,10 +1058,24 @@ err:
 	return ret;
 }
 
+/*
+ * Build the comma-separated controller directory name and mount option
+ * strings from a CgControllerEntry.  Returns the length written into
+ * @dir (without the trailing comma, which is replaced by '\0'), or -1
+ * if the output would overflow the supplied buffers.
+ *
+ * Every snprintf() call is checked for truncation: snprintf() returns
+ * the number of bytes that *would* have been written, so blindly
+ * adding that to the offset can move it past the buffer end.  The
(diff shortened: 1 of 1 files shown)


avagin:
LGTM. Thanks.

PR #3072, opened 2026-06-30
Refactor imports to avoid namespace pollution and clean up code

adrianreber:
This tries to fix two complaints from codeql.


rst0git (approved):
LGTM

PR #3073, opened 2026-07-01
criu: add --memory-dump-mode, a generic alias for --pre-dump-mode

arcceus:
--memory-dump-mode shares the same opt code and parser as --pre-dump-mode (splice|read). Remove the mdc->pre_dump guard from PRE_DUMP_READ paths in mem.c so the process_vm_readv-based read mode works for full dumps too, not just pre-dumps.

Suggested by @avagin in https://github.com/checkpoint-restore/criu/pull/3057#discussion_r3501553625

### DIFF
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -554,8 +554,9 @@ usage:
 	       "                        pages images of previous dump\n"
 	       "                        when used on restore, as soon as page is restored, it\n"
 	       "                        will be punched from the image\n"
-	       "  --pre-dump-mode       splice - parasite based pre-dumping (default)\n"
-	       "                        read   - process_vm_readv syscall based pre-dumping\n"
+	       "  --memory-dump-mode    splice - parasite based memory dumping (default)\n"
+	       "                        read   - process_vm_readv syscall based memory dumping\n"
+	       "                        (alias for --pre-dump-mode, applies to dump as well)\n"
 #ifdef CONFIG_LZ4
 	       "  -c|--compress         enable LZ4 per-page compression of memory pages\n"
 	       "  --compress-region size\n"
--- a/criu/crtools.c
+++ b/criu/crtools.c
@@ -556,7 +556,8 @@ usage:
 	       "                        will be punched from the image\n"
 	       "  --memory-dump-mode    splice - parasite based memory dumping (default)\n"
 	       "                        read   - process_vm_readv syscall based memory dumping\n"
-	       "                        (alias for --pre-dump-mode, applies to dump as well)\n"
+	       "                        (applies to dump as well)\n"
+	       "  --pre-dump-mode       (alias for --memory-dump-mode)\n"
 #ifdef CONFIG_LZ4
 	       "  -c|--compress         enable LZ4 per-page compression of memory pages\n"
 	       "  --compress-region size\n"
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -387,16 +387,23 @@ static int drain_pages(struct page_pipe *pp, struct parasite_ctl *ctl, struct pa
 	return 0;
 }
 
-static int xfer_pages(struct page_pipe *pp, struct page_xfer *xfer)
+static int xfer_pages(struct page_pipe *pp, struct page_xfer *xfer, int pid, bool external_read)
 {
 	int ret;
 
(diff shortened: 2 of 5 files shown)


criu/mem.c
```diff
@@ -359,16 +359,23 @@ static int drain_pages(struct page_pipe *pp, struct parasite_ctl *ctl, struct pa
...
 	 *           pre-dump action (see pre_dump_one_task)
+	 *
+	 * In READ mode no parasite vmsplice ever happened, so the pages
+	 * are pulled straight from the target's address space here via
+	 * process_vm_readv() instead of drained from the parasite pipe.
 	 */
 	timing_start(TIME_MEMWRITE);
-	ret = page_xfer_dump_pages(xfer, pp);
+	if (external_read)
+		ret = page_xfer_predump_pages(pid, xfer, pp);
```
avagin, line 404: This doesn't look right... why do we use page_xfer_predump_pages in case of the full dump?

arcceus: Just reused it because it uses process_vm_readv, page_xfer_dump_pages drains pages that are already vmspliced, i can add a comment explaining this or add a read mode in page_xfer_dump_pages if extra iov handling is a problem.

avagin: Ok, it seems just bad function names.

criu/page-xfer.c
```diff
@@ -859,7 +859,7 @@ int page_xfer_predump_pages(int pid, struct page_xfer *xfer, struct page_pipe *p
 			if (xfer->write_pagemap(xfer, &iov, flags))
 				goto err;
 
-			if (xfer->write_pages(xfer, ppb->p[0], iov.iov_len))
+			if ((flags & PE_PRESENT) && xfer->write_pages(xfer, ppb->p[0], iov.iov_len))
```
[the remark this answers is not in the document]
arcceus: Skipping drain is intentional that data has to stay in the pipe. Same pattern is used in Splice + --lazy-pages, same check is there in page_xfer_dump_pages() aswell on write_pages.

criu/crtools.c
```diff
@@ -554,8 +554,10 @@ int main(int argc, char *argv[], char *envp[])
 	       "                        pages images of previous dump\n"
 	       "                        when used on restore, as soon as page is restored, it\n"
 	       "                        will be punched from the image\n"
-	       "  --pre-dump-mode       splice - parasite based pre-dumping (default)\n"
-	       "                        read   - process_vm_readv syscall based pre-dumping\n"
+	       "  --memory-dump-mode    splice - parasite based memory dumping (default)\n"
+	       "                        read   - process_vm_readv syscall based memory dumping\n"
+	       "                        (applies to dump as well)\n"
+	       "  --pre-dump-mode       (alias for --memory-dump-mode)\n"
```
rst0git, line 560: @avagin How do we plan to use `--memory-dump-mode` in the future? If it is just an alias, do we need to keep `--pre-dump-mode` in the output from `--help`?

arcceus: @rst0git I think --pre-dump-mode is a bit misleading in this contex, which is why we are adding --memory-dump-mode with --pre-dump-mode kept as an alias. The memory can now also be read using process_vm_readv during a regular dump. We should keep --pre-dump-mode in --help for a while for backward compatibility before eventually deprecating it.

rst0git: @arcceus `--pre-dump-mode` is not misleading. It was introduced in another GSoC project optimizing the pre-dump functionality: https://dubeyabhishek.github.io/GSoC19CRIU/

When `--pre-dump-mode=read` is specified (`process_vm_readv` is used), memory regions that do not have `PROT_READ` are skipped. In the inventory image we save `pre_dump_mode` as it affects whether non-readable mappings may be absent.

arcceus: @rst0git My bad, I was trying to say that the same read path is now available for normal dump aswell, so --memory-dump-mode is intended as a more general name for that use case.

arcceus:
@avagin bumping this up ^

avagin:
@arcceus I am busy with some other things. Regarding GSoC projects, I really like the idea of getting mentor approval first before pinging me. Cc: @rst0git @adrianreber

PR #3076, opened 2026-07-02
plugins/cuda: switch to libcuda checkpoint API

rst0git:
This pull request adds support for CUDA plugin backend that allows to switch from invoking `cuda-checkpoint` to using the CUDA Driver API (`libcuda.so.1`) directly. The main benefit of using the driver API is that it makes the CUDA plugin for CRIU self-contained and avoids the need to install the `cuda-checkpoint` binary on every Kubernetes node.

In addition, compared to invoking the `cuda-checkpoint` tool, using the driver APIs allows us to avoid the overhead of executing an external CLI tool and parsing its output. This overhead occurs for each PID we check to determine whether it has a CUDA restore thread, as well as when performing the lock, checkpoint, restore, and unlock operations.

For example, the following tables show the cost of checking the CPU-only processes of a container (i.e., PIDs that don't have CUDA state) for the Qwen3.6-27B model served with SGLang on H200:

| Phase | Calls per backend | Driver API (ms) | cuda-checkpoint (ms) |
|---|---:|---:|---:|
| `pause_devices` | 17 | 0.630 | 448.880 |
| `checkpoint_devices` | 17 | 0.516 | 450.244 |
| `resume_devices_late` | 17 | 0.586 | 447.972 |

For the same model, we can observe similar performance improvements in the other phases of the CUDA plugin:

| Phase | Driver API (ms) | cuda-checkpoint (ms) |
|---|---:|---:|
| `checkpoint/init` | 11.717 | 2.046 |
| `checkpoint/pause_devices` | 4.672 | 773.338 |
| `checkpoint/checkpoint_devices` | 3517.400 | 4164.799 |
| `checkpoint/dump_devices_late` | 499.380 | 818.827 |
| `checkpoint/fini` | 10.106 | 0.013 |
| `restore/init` | 11.752 | 2.341 |
| `restore/restore_init` | 0.020 | 0.019 |
| `restore/resume_devices_late` | 7916.480 | 8793.203 |
| `restore/fini` | 22.247 | 0.011 |


jesus-ramos:
Forwarded this over to Steven as well to take a look as well as some of the other cuda checkpoint devs. 4th of July is a holiday so might not get responses there for a couple days just as a heads up.

The value redefinitions of the structs may be a bit brittle in that you may have to carry forward some backwards compatibility that the cuda-checkpoint utility hides away by doing some manual cuda version checks for what features are available in which versions. So instead of checking whether cuda-checkpoint reports a certain flag is supported you'd have to check the cuda version and see which fields of the structs you can use as new features are added and the reserved area  carved away. If you don't mind doing that handling here as new features are added that's fine, the API will return CUDA_ERROR_INVALID_VALUE if the reserved fields in that version of CUDA are non-zero if for example you try to use a newly added feature via CRIU in a slightly older driver that doesn't have the feature.

One thing to point out though is that while CUDA does have minor version compatibility a major CUDA version number may have revised definitions for the API's or structs that could throw a wrench in things so that's something to keep in mind.

I'll take a more thorough pass on the patch though and see if anything else sticks out and let Steven and others weigh in too.

adrianreber:
I have a question around the license.

CRIU is GPLv2, libcriu LGPL and anything in images/ is MIT licensed.

Using `dlopen()`  sounds like it should be no problem whatever license libcuda is, but how do you know the names of the functions to call and the parameters and in which order? At what documentation and header files did you look? What license where those parts?

Also the libcuda mock library looks like nothing CRIU has in the code base. Where was that taken from?

rst0git:
@jesus-ramos Thank you so much Jesus! Steven actually suggested that we do this some time ago. These changes also make it easier to deploy the CUDA plugin for CRIU in Kubernetes clusters. We presented some preliminary results for these changes, along with a PoC improvements for memory page compression and incremental GPU checkpointing at KubeCon earlier this year:

- [_Optimizing Error Recovery for Cost-Efficient Distributed AI Model Training with Kubernetes_](https://sched.co/2CW7Z)

>Using dlopen() sounds like it should be no problem whatever license libcuda is, but how do you know the names of the functions to call and the parameters and in which order? At what documentation and header files did you look? What license where those parts?

The following is a link to the CUDA documentation:
https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CHECKPOINT.html

I will add a comment that we intentionally avoid including NVIDIA's proprietary headers and the plugin resolves all symbols dynamically with `dlopen()`.

> Also the libcuda mock library looks like nothing CRIU has in the code base. Where was that taken from?

It is a minimal CUDA Driver API declarations reconstructed from the published CUDA Driver API specification.

adrianreber:
> The following is a link to the CUDA documentation:
> https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CHECKPOINT.html

How is that licensed? Can it be included in GPLv2 code?

rst0git:
> > The following is a link to the CUDA documentation:
> > https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__CHECKPOINT.html
> 
> How is that licensed? Can it be included in GPLv2 code?

My understanding is that using `dlopen("libcuda.so")` / `dlopen("libcuda.so.1")` with independently written minimal declarations using the public CUDA Driver API documentation should be permissible in GPLv2 code as the code is not incorporating NVIDIA SDK material and is not linked against libcuda at build time. I found the following precedents:

Slurm uses `dlopen()` to load GPU vendor libraries such as NVIDIA NVML (`libnvidia-ml.so` / `libnvidia-ml.so.1`) at runtime. This file is licensed GPLv2-or-later.
- https://github.com/SchedMD/slurm/blob/master/src/interfaces/gpu.c

Blender's CUEW loader dynamically loads CUDA (`libcuda.so` / `libcuda.so.1`) and resolves CUDA symbols at runtime. This file is licensed Apache-2.0, even though Blender as a whole is GPL.
- https://github.com/blender/blender/blob/main/extern/cuew/src/cuew.c

DPDK's CUDA GPU driver loads `libcuda.so` at runtime with `dlopen()` and resolves all CUDA API symbols dynamically rather than linking libcuda at build time. The DPDK core libraries and drivers are BSD-3-Clause.
- https://github.com/DPDK/dpdk/blob/main/drivers/gpu/cuda/cuda.c

adrianreber:
Sounds like the examples you have given should allow CRIU to adopt it. Although I am not the right person for this decision. Slurm is interesting as it is owned by nvidia now.

What about the tests you added, can that be redistributed as GPLv2?

avagin:
> libcuda

As far as you don't distribute nvidia binary code, you should be fine.

avagin:
@rst0git maybe we need to introduce a new plugin, release it and then deprecate the old one in the next release? If something doesn't work, users will be able to switch back to the old plugin? wdyt?

plugins/cuda/cuda_plugin.c
```diff
@@ -10,21 +10,79 @@
...
+typedef struct {
+	unsigned long long reserved[8];
+} CUcheckpointCheckpointArgs;
+
+typedef struct {
+	unsigned char oldUuid[16];
+	unsigned char newUuid[16];
+} CUcheckpointGpuPair;
+
+typedef struct {
```
dfeigin-nv: Why do we redefine the structs here? I'm sure there's a reason I'm just curious.

fals:
We already know that Nvidia and Amd are here and will not go anywhere for inference and training for a while. Can we instead have this build-in in CRIU instead of dynamic loading? Dynamic loading is terrible for static compilation and does not help with distribution. Now is very hard to release criu for each single cloud provider with different versions of kernel, glibc and OS.

adrianreber:
@fals not sure what you are asking. Can you try to elaborate.

adrianreber:
@fals Not sure if I understood your comment correctly, but I think what you are asking for is not possible because of licensing reasons.

fals:
> @fals Not sure if I understood your comment correctly, but I think what you are asking for is not possible because of licensing reasons.

My ask is simple, embedded this plugins in the binary instead of doing dlopen, keeping it as a external plugins makes no sense anymore and does not bring to much in my option but difficulties on building criu statically. Ideal world we should embedded everything that is possible in criu itself to help to release a static version for any architecture, that is my ask.

fals:
> @fals Not sure if I understood your comment correctly, but I think what you are asking for is not possible because of licensing reasons.

if is not possible I will try to come up with something, because main issue at the moment is this packing and distribution to multi cloud and architectures, is quite not viable building for all OS versions right now.

rst0git:
@fals I have patches that build static CRIU binary. It makes it easier to install and update CRIU in Kubernetes clusters. I will open a pull request for this soon.

fals:
> @fals I have patches that build static CRIU binary. It makes it easier to install and update CRIU in Kubernetes clusters. I will open a pull request for this soon.

We do build static but you will see the problem when you mix this dynamic load plugins for gpu, unless you do something I couldn't figure out, but doesn't work well on cloud with different versions of glibc.

rst0git:
> maybe we need to introduce a new plugin, release it and then deprecate the old one in the next release? If something doesn't work, users will be able to switch back to the old plugin? wdyt?

@avagin I've updated the pull request to preserve the old plugin as v1 and introduce the new plugin as v2.

> Why do we redefine the structs here? I'm sure there's a reason I'm just curious.

@dfeigin-nv I've updated the patch to fix this.

criu/plugin.c
```diff
@@ -180,26 +203,77 @@ static int cr_lib_load(int stage, char *path)
...
+	if (plugin->d->version == CRIU_PLUGIN_VERSION_OLD)
+		xfree(plugin->d);
+	dlclose(plugin->dlhandle);
+	xfree(plugin);
+}
+
+static void cr_plugin_select_versions(void)
+{
+	plugin_desc_t *plugin, *tmp, *latest;
+
```
dfeigin-nv: IIUC, cuda_plugin_v2.so and cuda_plugin_v1.so are both loaded, then unloads old impls.
What happens if a cuda_plugin_v2.so is not supported (Driver API requires r570, so older than that? The call to cr_plugin_activate will fail, and the dump/restore will fail, because there is no fallback to 1.

rst0git: Thanks Dan - this is a good point! The problem is not specific to the CUDA plugin, I've updated the patches with the following solution.

First, CRIU finds the plugin with highest `implementation_version` and calls `init()`. If `init()` returns `-ENOTSUP`, CRIU unloads only that plugin and tries the next version. If we have a plugin with multiple versions, alternative versions are unloaded and the hooks registered only after one candidate returns success.

In our case, if v2 fails to resolve the checkpoint Driver API, it would return `-ENOTSUP`. Then, v1 checks if the `cuda-checkpoint --action` CLI is available.

avagin:
>     plugin: select supported implementation versions

please don't do that. We don't want to choose a plugin version dynamically. Users have to install just one version.

avagin:
> > @fals I have patches that build static CRIU binary. It makes it easier to install and update CRIU in Kubernetes clusters. I will open a pull request for this soon.
> 
> We do build static but you will see the problem when you mix this dynamic load plugins for gpu, unless you do something I couldn't figure out, but doesn't work well on cloud with different versions of glibc.

NVIDIA doesn't provide static libraries as far as I know. The issue here isn't the plugin mechanism, but the NVIDIA driver libraries. Even with static libraries, we'd run into compatibility problems whenever CRIU is linked against one version while containers use another.

rst0git:
> We don't want to choose a plugin version dynamically. Users have to install just one version.

@avagin This patch aims to address the earlier comment here: https://github.com/checkpoint-restore/criu/pull/3076#discussion_r3675319512

The plugin that uses the `cuda-checkpoint` tool supports older NVIDIA driver versions (e.g., 565, 570), while the new plugin supports newer driver versions (e.g., 580, 595, and 610). I expect future driver versions might introduce new features that may require corresponding updates to the plugin.

The intended behavior is for CRIU to automatically detect whether the new plugin is supported by the installed driver. If it is supported, CRIU uses the new plugin, otherwise, it falls back to the `cuda-checkpoint`-based plugin. This allows the same deployment to work across nodes running different NVIDIA driver versions without requiring users to select or configure a plugin manually.

Users typically do not install CRIU or either plugin themselves. They are deployed automatically to each Kubernetes node, for example, through a DaemonSet. The deployment is an OCI image that contains the CRIU binary and the required libraries and plugins.

avagin:
> > We don't want to choose a plugin version dynamically. Users have to install just one version.
> 
> @avagin This patch aims to address the earlier comment here: [#3076 (comment)](https://github.com/checkpoint-restore/criu/pull/3076#discussion_r3675319512)
> 
> The plugin that uses the `cuda-checkpoint` tool supports older NVIDIA driver versions (e.g., 565, 570), while the new plugin supports newer driver versions (e.g., 580, 595, and 610). I expect future driver versions might introduce new features that may require corresponding updates to the plugin.
> 
> The intended behavior is for CRIU to automatically detect whether the new plugin is supported by the installed driver. If it is supported, CRIU uses the new plugin, otherwise, it falls back to the `cuda-checkpoint`-based plugin. This allows the same deployment to work across nodes running different NVIDIA driver versions without requiring users to select or configure a plugin manually.
> 
> Users typically do not install CRIU or either plugin themselves. They are deployed automatically to each Kubernetes node, for example, through a DaemonSet. The deployment is an OCI image that contains the CRIU binary and the required libraries and plugins.

I don't like this over-complicated logic in the criu plugin engine. Maybe it will be easier if we will have one nvidia plugin that will support both modes and the plugin will decide what mode to use.

rst0git:
@avagin I rebased this branch on top of the changes in https://github.com/checkpoint-restore/criu/pull/3138

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1057 @@
...
+	} while (waited < 0 && errno == EINTR);
+	if (!ret && waited < 0)
+		ret = -errno;
+	if (!ret && (!WIFEXITED(status) || WEXITSTATUS(status)))
+		ret = -EIO;
+	if (!ret)
+		ret = cuda_validate_inventory(&list->inventory);
+
+restore_mask:
+	if (sigprocmask(SIG_SETMASK, &saved_mask, NULL) && !ret)
```
avagin: please log a error here

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1057 @@
...
+		ret = cuda_img_at_eof(pipefd[0]);
+	close(pipefd[0]);
+
+	if (ret)
+		kill(child, SIGKILL);
+	do {
+		waited = waitpid(child, &status, 0);
+	} while (waited < 0 && errno == EINTR);
+	if (!ret && waited < 0)
+		ret = -errno;
```
avagin: please log a error here

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1057 @@
...
+	 * Collect only UUIDs in a child; all image and streamer I/O stays here.
+	 */
+	child = fork();
+	if (child == 0) {
+		close(pipefd[0]);
+		reply.error = cuda_collect_gpus(list);
+		if (!reply.error)
+			reply.count = (unsigned int)list->inventory.n_gpus;
+		ret = cuda_img_write(pipefd[1], &reply, sizeof(reply));
+		if (!ret && !reply.error)
```
avagin, line 357: please log a error here

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1065 @@
...
+	if (opts.stream) {
+		fd = img_streamer_open(CUDA_GPU_INVENTORY_IMAGE,
+				       write_image ? O_DUMP : O_RSTR);
+		if (fd < 0)
+			return fd == -1 ? -EIO : fd;
+	} else {
+		flags = write_image ? O_WRONLY | O_CREAT | O_TRUNC : O_RDONLY;
+		fd = openat(criu_get_image_dir(), CUDA_GPU_INVENTORY_IMAGE, flags, 0600);
+		if (fd < 0)
+			return -errno;
```
avagin: print an error message here and in all other cases...

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1065 @@
...
+
+static int cuda_img_read(int fd, void *buf, size_t len)
+{
+	char *ptr = buf;
+
+	while (len) {
+		ssize_t ret = read(fd, ptr, len);
+
+		if (ret < 0) {
+			if (errno == EINTR || errno == EAGAIN)
```
avagin: why do we need to handle them here?

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1065 @@
...
+	if (child == 0) {
+		close(pipefd[0]);
+		reply.error = cuda_collect_gpus(list);
+		if (!reply.error)
+			reply.count = (unsigned int)list->inventory.n_gpus;
+		ret = cuda_img_write(pipefd[1], &reply, sizeof(reply));
+		if (!ret && !reply.error)
+			ret = cuda_img_write(pipefd[1], list->uuids,
+					     reply.count * sizeof(*list->uuids));
+		if (ret)
```
avagin, line 360: if reply.error is non-zero, cuda_img_write isn't called, but the process can exit with zero code.

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1065 @@
...
+	 * Collect only UUIDs in a child; all image and streamer I/O stays here.
+	 */
+	child = fork();
+	if (child == 0) {
+		close(pipefd[0]);
+		reply.error = cuda_collect_gpus(list);
+		if (!reply.error)
+			reply.count = (unsigned int)list->inventory.n_gpus;
+		ret = cuda_img_write(pipefd[1], &reply, sizeof(reply));
+		if (!ret && !reply.error)
```
avagin, line 357: I think we need to log reply.error if it isn't zero

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1065 @@
...
+		ret = cuda_img_read(pipefd[0], list->uuids, reply.count * sizeof(*list->uuids));
+	if (!ret)
+		ret = cuda_img_at_eof(pipefd[0]);
+	close(pipefd[0]);
+
+	if (ret)
+		kill(child, SIGKILL);
+	do {
+		waited = waitpid(child, &status, 0);
+	} while (waited < 0 && errno == EINTR);
```
avagin: again, if there is no real reason when we expect EINTR, we don't need to handle it

plugins/cuda/cuda_device_map.c
```diff
@@ -0,0 +1,1065 @@
...
+		goto restore_mask;
+	}
+
+	ret = cuda_img_read(pipefd[0], &reply, sizeof(reply));
+	if (!ret && reply.error)
+		ret = reply.error < 0 ? reply.error : -EIO;
+	if (!ret && (!reply.count || reply.count > CUDA_GPU_INVENTORY_MAX_GPUS))
+		ret = -EINVAL;
+	if (!ret)
+		ret = cuda_gpu_list_init(list, reply.count);
```
avagin: imho, this `if (!ret)` pattern doesn't work in this case. The code doesn't look good and in case of any error it will be impossible to find out what is actually failed.

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -392,53 +404,140 @@ static int resume_restore_thread(int restore_tid, k_rtsigset_t *save_sigset)
...
+			return COMPEL_TASK_STOPPED;
+		}
+		/* Consume our delivery stop without delivering SIGSTOP to the task. */
+		if (info.si_signo == SIGSTOP && info.si_code == SI_QUEUE && info.si_value.sival_ptr == op)
+			return COMPEL_TASK_ALIVE;
+		if (info.si_code > 0 && (info.si_code >> 8) == PTRACE_EVENT_STOP) {
+			pr_err("CUDA restore thread %d entered a group-stop\n", op->restore_tid);
+			return COMPEL_TASK_STOPPED;
+		}
+		if (ptrace(PTRACE_CONT, op->restore_tid, NULL, (unsigned long)info.si_signo)) {
```
avagin: we don't expect any signals to the restore thread, it should fail in case of any unexpected signals.

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -392,53 +404,140 @@ static int resume_restore_thread(int restore_tid, k_rtsigset_t *save_sigset)
...
+}
+
+static int run_cuda_operation(int pid, int restore_tid, int (*run)(void *arg), void *arg)
+{
+	struct cuda_operation op = { .pid = pid, .restore_tid = restore_tid, .run = run, .arg = arg };
+	k_rtsigset_t restore_sigset;
+	sigset_t blocked, saved_mask;
+	pthread_t worker;
+	void *result;
+	int ret = -1, err, state;
```
avagin: s/ret/exit_code/

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -392,53 +404,140 @@ static int resume_restore_thread(int restore_tid, k_rtsigset_t *save_sigset)
...
+		pr_err("Cannot block SIGCHLD for CUDA operation: %s\n", strerror(err));
 		return -1;
 	}
+	if (resume_restore_thread(restore_tid, &restore_sigset))
+		goto restore_mask;
 
-	if (task_info->initial_task_state == CUDA_TASK_CHECKPOINTED) {
-		pr_info("pid %d already in a checkpointed state\n", pid);
-		return 0;
+	err = pthread_create(&worker, NULL, cuda_operation_thread, &op);
```
avagin: I think the thread approach here isn't ideal too. We are working around a rare NVIDIA bug by adding threading and ptrace overhead (pthread_create/join + rt_tgsigqueueinfo SIGSTOP) to the normal path for all users.
  
Also, forwarding fatal signals to unblock libcuda will likely only work in single-threaded test cases (where restore_tid == pid). In real multi-threaded CUDA processes, all other threads are stopped by ptrace, so killing/stopping one thread will not close a cuda rpc fd and libcuda (and pthread_join) will still hang.
  
Since we block signals on restore_tid before resuming it, we don't expect any signals during a healthy CUDA call, any ptrace stop or exit of restore_tid means something went wrong. Could we instead run libcuda directly, catch unexpected ptrace stops in a   SIGCHLD handler, and use siglongjmp() to abort the libcuda call?
  
    Pros:
    - Zero overhead on the normal path (no threads, no extra ptrace stops/SIGSTOP injection).
    - Works reliably for multi-threaded processes without needing libcuda to unblock itself or killing the target process.
    - Simpler code (no -lpthread, signal forwarding loop, or sival_ptr matching).
  
    Cons:
    - A theoretical risk of a libc deadlock if SIGCHLD interrupts libcuda while it holds a libc lock (e.g., malloc/free) and we longjmp out. However, this can only happen when restore_tid has already crashed (a fatal error where CRIU is aborting anyway) and while libcuda is  almost certainly blocked in a kernel syscall waiting for restore_tid rather than in userspace libc code.

rst0git: Thanks! I've updated the patches.

plugins/cuda/cuda_plugin_driver.c
rst0git: I've updated the pull request with this change.

criu/cr-dump.c
```diff
@@ -1960,7 +1961,18 @@ static int cr_dump_finish(int ret)
...
 		return -1;
 
-	cr_plugin_fini(CR_PLUGIN_STAGE__DUMP, ret);
+	/*
+	 * A pre-dump script can fail before cr_plugin_init() initializes the
+	 * hook chains. The plugin list is initialized statically, so it is safe
+	 * to check here: if no plugins are loaded, there are no hooks to run.
+	 */
+	if (!list_empty(&cr_plugin_ctl.head)) {
+		plugin_ret = run_plugins_all(DUMP_FINISH, ret ?: post_dump_ret);
```
avagin, line 1970: we need to log all ENOTSUP errors here.

criu/cr-dump.c
```diff
@@ -1960,7 +1961,18 @@ static int cr_dump_finish(int ret)
 	if (arch_set_thread_regs(root_item, true) < 0)
 		return -1;
 
-	cr_plugin_fini(CR_PLUGIN_STAGE__DUMP, ret);
+	/*
+	 * A pre-dump script can fail before cr_plugin_init() initializes the
+	 * hook chains. The plugin list is initialized statically, so it is safe
+	 * to check here: if no plugins are loaded, there are no hooks to run.
+	 */
+	if (!list_empty(&cr_plugin_ctl.head)) {
```
avagin, line 1969: I think it would be better if cr_plugin_ctl is proper initialized when a criu process is started. You don't need to load plugins, just initialize lists. You can introduce a function with __attribute__((constructor)) for that.

avagin:
```
commit 66fe7a1ddaa11d08dea15b7addffbbaae687eb7d
Author: Radostin Stoyanov <rstoyanov@fedoraproject.org>
Date:   Sun Aug 30 16:09:43 2026 +0100

    plugins: run all global device hooks
    
    DUMP_DEVICES_LATE, RESTORE_INIT, and POST_FORKING coordinate global
    state for each device plugin. Stopping after the first handler prevents
    other installed plugins from completing their cleanup or restore
    preparation.
    
    Add a broadcast dispatcher for all three global device hooks. It calls
    every registered handler and returns the first error after all handlers
    have run. Treat only zero and -ENOTSUP as nonfatal results. Cover success,
    unsupported, and multiple-error ordering in the unit suite.
    
    Plugin finalizers cannot report errors because they return void. Add a
    DUMP_FINISH broadcast hook and move CUDA rollback into it so dump -R
    fails when restoring device state fails. Attempt rollback for every
    tracked task and preserve any earlier dump error. Pass post-dump script
    failures to the hook and finalizers so device rollback follows the
    decision to resume the source tasks.
    
    Inject restore and unlock failures in both CUDA backends and check that
    dump -R reports failure while releasing the source tasks. Cover
    successful rollback and verify that one task failure does not skip later
    tasks. Check that a post-dump script failure triggers rollback without
    -R, including when CUDA restore also fails.
    
    Assisted-by: Codex:GPT-6
    Signed-off-by: Radostin Stoyanov <rstoyanov@fedoraproject.org>

 criu/cr-dump.c                                   |  20 ++++++++++++++++++--
 criu/cr-restore.c                                |   4 ++--
 criu/include/criu-plugin.h                       |   2 ++
 criu/include/plugin.h                            |  24 ++++++++++++++++++++++++
 criu/plugin.c                                    |   4 ++--
 criu/unittest/unit.c                             | 131 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 plugins/cuda/cuda_plugin.c                       |   9 +++++++++
 plugins/cuda/cuda_plugin.h                       |   1 +
 plugins/cuda/cuda_plugin_cli.c                   |  29 +++++++++++++++++++----------
 plugins/cuda/cuda_plugin_driver.c                |  29 +++++++++++++++++++----------
 test/cuda-checkpoint/backend-errors.py           |  70 ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++--
 test/cuda-checkpoint/cuda-checkpoint.c           |  62 +++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++-
 test/cuda-checkpoint/libcuda.c                   |   6 ++++++
 test/others/action-script/check_early_failure.py |  43 +++++++++++++++++++++++++++++++++++++++++++
 test/others/action-script/run.sh                 |   2 ++
 15 files changed, 407 insertions(+), 29 deletions(-)
```

This patch definitely contains more than one logical change... Let's split this change into patches properly and move it to a separate pr.

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -398,20 +470,163 @@ static int resume_restore_thread(int restore_tid, k_rtsigset_t *save_sigset)
...
+}
+
+static int run_cuda_operation(int pid, int restore_tid, int (*run)(void *arg), void *arg)
+{
+	struct sigaction action = { .sa_handler = cuda_sigchld, .sa_flags = SA_RESTART };
+	struct sigaction saved_action;
 	k_rtsigset_t restore_sigset;
-	int exit_code, int_ret;
+	sigset_t saved_mask;
+	volatile int exit_code = -1;
```
avagin, line 544: does exit_code changed outside of this function? why is it volatile?

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -331,25 +376,52 @@ static int restore_thread_settings(int restore_tid, k_rtsigset_t *restore_sigset
...
-	if (compel_wait_task(restore_tid, -1, parse_pid_status, NULL,
-			     &creds.s, NULL) != COMPEL_TASK_ALIVE) {
-		pr_err("compel_wait_task failed after interrupt\n");
-		return -1;
+	for (;;) {
+		pid = waitpid(restore_tid, &status, __WALL | WNOHANG);
+		if (pid == restore_tid)
+			break;
+		if (pid < 0) {
+			if (errno == EINTR)
```
avagin, line 406: waitpid will not return EINTR with WNOHANG. We don't set any signal handlers without SA_RESTART. We almost never need to handle EINTR in criu code.

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -398,20 +470,163 @@ static int resume_restore_thread(int restore_tid, k_rtsigset_t *save_sigset)
...
+		cuda_guard.status = status;
+	return true;
+}
+
+static void cuda_sigchld(int signal)
 {
+	int saved_errno = errno;
+	int owner = __atomic_load_n(&cuda_guard.owner_tid, __ATOMIC_ACQUIRE);
+
+	(void)signal;
```
avagin, line 496: ai garbage

plugins/cuda/cuda_plugin_driver.c
```diff
@@ -12,11 +13,48 @@
...
+static bool driver_call_aborted;
+
+/* Only the tracing thread accesses the jump buffer and wait status. Other
+ * threads can receive process-directed SIGCHLD and only notify this thread.
+ * Keep this storage alive when a late handler outlives an operation.
+ */
+static struct {
+	sigjmp_buf env;
+	sigset_t blocked;
+	int owner_tid;
```
avagin, line 36: Using this function in a multithreaded context can be dangerous. If someone attempts to make CUDA calls from multiple threads concurrently, I don't think the current implementation will handle it properly. Masking signals only affects the calling thread, and calling sigaction a second time will return an incorrect previous handler. Honestly, I don't think we should try to work around a CUDA library bug this way. If you think the issue is critical for end users, I would suggest marking this new plugin mode as experimental and keeping the old mode as the default.

PR #3079, opened 2026-07-06
python: fix file handle leaks flagged by CodeQL

adrianreber:
Fix "File is not always closed" warnings reported by CodeQL across 7  Python files. The changes ensure file handles are properly closed on both normal and exceptional code paths, using 'with' statements,   context managers, and try/finally blocks as appropriate.                                                                                                                                                                                    
                                                                                                                                                                                                                                              
The bulk of the changes are in crit/__main__.py, where inf(), outf(), and dinf() are converted to `@contextmanager` generators so all call sites can use 'with' statements while still avoiding closing  stdin/stdout. The remaining files are mechanical conversions from manual open/close to 'with' statements.                                                                                                                                                                                                     
                                                                                                                                                                                                                                              
Skipped test/inhfd/fifo.py — the open FDs there are intentionally returned to the caller by design.


coredump/criu_coredump/coredump.py
```diff
@@ -888,70 +888,70 @@ def _gen_mem_chunk(self, pid, vma, size):
 
             f.seek(off)
```
rst0git: Should we move `f.seek(off)` also inside the `try` / `finally` block or switch to `with open(...) as f:`?

adrianreber: Yes, `f.seek()` should be moved. 

The way it is opened with `f = None` at the beginning an then sometimes checking for if `f is not None` didn't easily work with `with open()`. Let me try something different.

adrianreber: Okay, reworked.

coredump/criu_coredump/coredump.py
```diff
@@ -872,86 +873,86 @@
...
+                # Open file before iterating vma pages
+                shmid = vma["shmid"]
+                off = vma["pgoff"]
+
+                files = self.reg_files
+                fname = next(
+                    filter(lambda x: x["id"] == shmid, files))["name"]
+
+                try:
+                    f = stack.enter_context(open(fname, 'rb'))
```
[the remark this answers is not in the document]
adrianreber: Using ExitStack `__exit__()` of the file object will be called.

rst0git (approved):
LGTM

PR #3080, opened 2026-07-07
util: Ignore chown errors on restore in unprivileged mode

### LINKED ISSUE #2762: UNIX socket restore in non-root container fails with "Unable to change ownership to (0, 0): operation not permitted"
We are mainly running into this issue when checkpointing and restoring Java processes that have had `jstack` run on them.

To reproduce: inside a container running as a **non-root** user (e.g. `nobody`):
1. Start a Java process in the background. Grab its PID `p`.
2. Run jstack against `p`. This also creates a UNIX socket `/tmp/.java_pid${p}`.
3. Grab a checkpoint of `p` with `criu dump`.
4. Restore `p` from that same checkpoint. Restore fails with the error shown below.

Error output tail:
```
(00.015929)      7: unix: ghost: bind id 0x18 ino 47314041 addr /tmp/.java_pid7.tmp
(00.015936)      7: unix: ghost: socket id 0x18 ino 47314041 name /tmp/.java_pid7.tmp detected F_OK /tmp
(00.015939)      7: unix: ghost: socket id 0x18 ino 47314041 name /tmp/.java_pid7.tmp creating /tmp
(00.016098)      7: Error (criu/util.c:1028): Unable to change [-100]//tmp/.java_pid7.tmp ownership (65534, 65534) to (0, 0): Operation not permitted
(00.016168)      7: Error (criu/sk-unix.c:1651): unix: Can't bind id 0x18 ino 47314041 addr /tmp/.java_pid7.tmp: Operation not permitted
(00.016197)      7: Error (criu/files.c:1221): Unable to open fd=5 id=0x18
(00.016752) Error (criu/cr-restore.c:1258): 7 exited, status=1
(00.016788) Error (criu/cr-restore.c:2324): Restoring FAILED.
```

I have a [simple setup](https://github.com/user-attachments/files/22565949/socket-enoperm.zip) that reproduces this. Configurations I've been able to reproduce this on:
* Debian `bookworm` + CRIU 3.19
* Debian `trixie` + CRIU 3.19
* Debian `trixie` + CRIU 4.1.1
* Debian `trixie` + `criu-dev` HEAD

ApurveKaranwal:
# Description

### What does this PR do?
This PR resolves an issue where restoring UNIX domain sockets (and other files such as pipes or ghost files) fails with an `EPERM` error when running CRIU in unprivileged mode (e.g., inside a user namespace or a rootless container). 

### How does it do it?
In unprivileged mode, the restoring process does not possess `CAP_CHOWN`. Thus, any attempt to restore the socket's owner back to the original `uid`/`gid` via `fchown`/`fchownat` will fail. Since the restored process runs under the unprivileged user's UID anyway, it will still have access to the socket, making this ownership mismatch non-fatal.

This PR modifies `cr_fchown` and `cr_fchpermat` in `criu/util.c` to:
1. Log a warning using `pr_warn` when `fchown` or `fchownat` fails with `EPERM` due to insufficient privileges.
2. Gracefully bypass the ownership change failure and return `0` (success) when `opts.unprivileged` is set.
3. Added a new ZDTM static test case `sk_unix_owner` that validates UNIX domain socket ownership is correctly handled during checkpoint and restore under both root (`h` and `ns` flavors with SUID) and unprivileged/rootless (`uns` flavor) environments.

Fixes: #2762 

### Automated ZDTM Tests
A new test `sk_unix_owner` was added to verify both SUID preservation (as root) and graceful fallback (under user namespaces):
- `sudo ./test/zdtm.py run -t zdtm/static/sk_unix_owner` (verifies `chown` ownership is preserved across C/R)
- `sudo ./test/zdtm.py run -t zdtm/static/sk_unix_owner -f uns` (verifies that ownership failure is ignored gracefully and socket communication remains fully functional)

### DIFF
--- a/criu/util.c
+++ b/criu/util.c
@@ -994,8 +994,14 @@ int cr_fchown(int fd, uid_t new_uid, gid_t new_gid)
 	}
 	pr_debug("fstat(%d): uid %u gid %u\n", fd, st.st_uid, st.st_gid);
 
-	if (new_uid != st.st_uid || new_gid != st.st_gid)
+	if (new_uid != st.st_uid || new_gid != st.st_gid) {
+		if (opts.unprivileged) {
+			pr_warn("Unable to change fd %d ownership (%d, %d) to (%d, %d): Operation not permitted (ignored in unprivileged mode)\n",
+				fd, st.st_uid, st.st_gid, new_uid, new_gid);
+			return 0;
+		}
 		goto out_eperm;
+	}
 
 	return 0;
 out_eperm:
@@ -1024,11 +1030,16 @@ int cr_fchpermat(int dirfd, const char *path, uid_t new_uid, gid_t new_gid, mode
 	}
 
 	if (new_uid != st.st_uid || new_gid != st.st_gid) {
-		errno = EPERM;
-		pr_perror("Unable to change [%d]/%s ownership (%d, %d) to (%d, %d)",
-			  dirfd, path, st.st_uid, st.st_gid, new_uid, new_gid);
-		errno = EPERM;
-		return -1;
+		if (opts.unprivileged) {
+			pr_warn("Unable to change [%d]/%s ownership (%d, %d) to (%d, %d): Operation not permitted (ignored in unprivileged mode)\n",
+				dirfd, path, st.st_uid, st.st_gid, new_uid, new_gid);
+		} else {
+			errno = EPERM;
+			pr_perror("Unable to change [%d]/%s ownership (%d, %d) to (%d, %d)",
+				  dirfd, path, st.st_uid, st.st_gid, new_uid, new_gid);
+			errno = EPERM;
+			return -1;
+		}
 	}
 
 	if (new_mode == st.st_mode)
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -471,6 +471,7 @@ TST_DIR		=				\
 		del_standalone_un_seqpacket	\
 		sk-unix-mntns			\
 		sk-unix01			\
+		sk_unix_owner			\
 		sk-unix01-seqpacket		\
 		sk-unix-dgram-ghost		\
 		unsupported_children_collision  \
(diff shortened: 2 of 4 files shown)


adrianreber:
Can you give an example when this happens? I am not sure I understand running in unprivileged mode correctly. Do you mean running CRIU with `--unprivileged`.

ApurveKaranwal:
hey @adrianreber sir.
Yes, this refers to running CRIU with the `--unprivileged` flag (which sets `opts.unprivileged = true`), typically when checkpointing/restoring rootless containers (e.g., using rootless Podman, rootless Docker, or LXD).

### Example Scenario
Consider a container where a process creates a UNIX domain socket (or a pipe/ghost file) that ends up owned by a specific user, say `uid 1000` (or a service user like `uid 999` for database/web servers).

1. **On Dump**: The socket's ownership (`1000:1000`) is captured in the CRIU image files.
2. **On Restore** (using `criu restore --unprivileged`):
   - The restoring CRIU process runs as a rootless/unprivileged user on the host.
   - Even if running inside a user namespace where the restoring user maps to container root, they may not have `uid 1000` mapped in the user namespace (`/proc/self/uid_map`), or they may lack host-level `CAP_CHOWN`.
   - When CRIU attempts to recreate the socket file and calls `fchownat(..., 1000, 1000)`, the Linux kernel rejects the call and returns `EPERM` because the process lacks the necessary privileges/mappings to change ownership to that UID/GID.
   - Without this PR, this `EPERM` causes the entire restore operation to abort and fail.

### Why this fix is safe:
In an unprivileged/rootless restore, all restored processes will run under the unprivileged user's UID anyway. They will still have access to the restored UNIX socket or pipe, so failing the entire restore over the ownership restoration mismatch is unnecessary.

With this change, CRIU logs a warning and proceeds, allowing rootless container restores to succeed.

adrianreber:
>  typically when checkpointing/restoring rootless containers (e.g., using rootless Podman, rootless Docker, or LXD)

Is this already supported in any of the mentioned tools?

I am sorry, but I am still really confused about the use case and where you see this error happening.  Let's try to remove containers and user namespaces for a first step if possible. To make it easier to understand for me. How is the restored process supposed to work if the `chown` is ignored? I guess my main question is why is it okay to continue the restore if the `chown` fails?

> In an unprivileged/rootless restore, all restored processes will run under the unprivileged user's UID anyway. They will still have access to the restored UNIX socket or pipe

Why will they have access to the socket?

PR #3081, opened 2026-07-07
python: remove redundant variable definitions flagged by CodeQL

adrianreber:
Remove "Variable defined multiple times" warnings reported by CodeQL.  Most are C-style sentinel initializations (e.g. `x = -1`) that are   unnecessary in Python because the variable is always assigned before  use or an exception is raised. Also removes a no-op  `p1.stdout.read()` call which was unused and overwritten.


rst0git (approved):
LGTM

PR #3082, opened 2026-07-08
Fix soccr tests and run them in CI

adrianreber:
The python code in `soccr/test` claimed to be Python 3 code according to the shebang but it was not. The C code also didn't compile any more. This fixes the C code, ports the Python code to 3, runs it through the linters and also runs the soccr tests now in CI.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,6 +315,7 @@ clean-cuda_plugin:
 
 clean-top:
 	$(Q) $(MAKE) -C Documentation clean
+	$(Q) $(MAKE) -C soccr/test clean
 	$(Q) $(MAKE) $(build)=test/compel clean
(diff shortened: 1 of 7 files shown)


adrianreber:
This needs some additional changes to make it work. Will look into it.

adrianreber:
Looks like this might also need one of the chmods from #3063

adrianreber:
Now it seems to work correctly.

rst0git:
@adrianreber Would it be possible to rebase this branch on criu-dev?

adrianreber:
> @adrianreber Would it be possible to rebase this branch on criu-dev?

Rebased.

rst0git (approved):
LGTM

PR #3085, opened 2026-07-09
ci: improve cross-compilation Dockerfile maintainability and resilience

adrianreber:
- Deduplicate the riscv64 cross-compilation Dockerfile template by moving the riscv64-specific APT setup into its header file, making the template a symlink to the shared one like all other architectures                                                                                                                                                                                                         
  - Print an error message if the cross compilation fails because of package installation problems.


rst0git (approved):
LGTM

PR #3087, opened 2026-07-09
Last few Python fixes to make CodeQL happy

adrianreber:
This removes the last CodeQL warnings in Python. Some warnings have been dismissed as false positives but most of them are now fixed. This is the last batch.


rst0git (approved):
LGTM

PR #3088, opened 2026-07-09
Move dmesg out of central ci script

adrianreber:
In `scripts/ci/run-ci-tests.sh` there was a cleanup function running `dmesg`. This is helpful especially if something fails. One the problems with this approach is that if something files there are potentially hundreds of lines from `dmesg` before the actual error is visible.

This PR moves the run of `dmesg` into a separate step in the different CI setups. This way the content of `dmesg` is still there and can be investigated if necessary but it does not pollute the main CI result output.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -20,6 +20,10 @@ jobs:
             sudo -E make -C scripts/ci local << parameters.compiler >>=1
             ZDTM_SHARD_INDEX=<< parameters.shard_index >>
             ZDTM_SHARD_COUNT=<< parameters.shard_count >>
+      - run:
+          name: Print dmesg
+          when: always
+          command: sudo dmesg
 
 workflows:
   version: 2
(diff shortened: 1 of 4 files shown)


adrianreber:
Depends on #3086

rst0git:
> Depends on #3086

@adrianreber The pull request above is now merged. Would it be possible to update this PR?

adrianreber:
> > Depends on #3086
> 
> @adrianreber The pull request above is now merged. Would it be possible to update this PR?

Rebased.

rst0git:
@adrianreber Would it make sense to update also `linux-next.yml`? It has similar `Print dmesg` but without `if: always()`?

adrianreber:
I saw that it is running dmesg in its own step, but adding always probably is a good idea. Should it happen in this PR?

rst0git:
> Should it happen in this PR?

I think it makes sense to be as a separate commit in this PR, just to keep the CI tests consistent.
For example, the workflow has failed here and the "Print dmesg" step was skipped:
https://github.com/checkpoint-restore/criu/actions/runs/29186839629/job/86636161790

rst0git (approved):
LGTM

PR #3092, opened 2026-07-15
lib, criu: remove unnecessary NULL checks before free()

adrianreber:
free(NULL) is a no-op per the C standard (C99 7.20.3.2), so guarding free() calls with NULL checks is unnecessary.


rst0git (approved):
LGTM

PR #3093, opened 2026-07-16
sk-queue: add SCM_PIDFD / SO_PASSPIDFD checkpoint/restore support

3idey:
SCM_PIDFD checkpoint/restore, including senders that are already dead

A receiver with SO_PASSPIDFD mints a pidfd from the struct pid the kernel
attached to each queued skb. CRIU did not checkpoint those at all, and for
pidfd files whose task was already reaped it restored the fact of death but
not how the process died. This series covers both.

On dump we inspect the pidfd the peek already hands us instead of closing it
blindly, and record the sender's wait(2) status when the kernel can tell us.
On restore, stale packets are re-sent with spoofed SCM_CREDENTIALS naming a
stand-in process that reproduces that status, so the receiver mints a pidfd
that goes stale exactly as the original was.

Stand-ins are keyed on the pidfs inode, not on exit status. Dying the same
way is not being the same process, so two senders that exited alike keep
distinct identities across C/R. One dead process comes back as one struct
pid however many tasks refer to it.

Image format: new pidfs_attr_entry (images/pidfs.proto), plus an optional
ino_hi on pidfd_entry for 32-bit kernels. Wire compatible, existing images
stay readable.

Kernels: SO_PASSPIDFD 6.5, pidfs inodes 6.9, PIDFD_GET_INFO 6.13,
PIDFD_INFO_EXIT 6.15, pidfd of a reaped sender 6.17. Each is probed, and
missing ones degrade rather than fail. criu check --extra-features gains
pidfd_get_info.

Tests: six new zdtm tests (sk-unix-dgram-pidfd, -cred, -stale, -ino, -xtask,
-reuse) plus an exit-status check in pidfd_dead. Each probes the kernel
before the dump and asserts the ground truth it relies on, so a failure is
unambiguously ours.

Known limits: two tasks in different pid namespaces referring to one dead
process still get a stand-in each. A restored dead pid never reports
PIDFD_INFO_COREDUMP.

### DIFF
--- a/criu/sk-queue.c
+++ b/criu/sk-queue.c
@@ -11,6 +11,7 @@
 
 #include "common/list.h"
 #include "imgset.h"
+#include "kerndat.h"
 #include "image.h"
 #include "servicefd.h"
 #include "cr_options.h"
@@ -388,6 +389,20 @@ static int dump_packet_cmsg(struct msghdr *mh, SkPacketEntry *pe, int flags)
 				if (ret < 0)
 					return -1;
 				continue;
+			} else if (ch->cmsg_len == CMSG_LEN(sizeof(int)) &&
+				   ch->cmsg_type == SCM_PIDFD) {
+				int pidfd = *(int *)CMSG_DATA(ch);
+
+				/*
+				 * The skb pid the pidfd refers to is dumped
+				 * from the SCM_CREDENTIALS cmsg (SO_PASSCRED
+				 * is enabled for the peek when the socket has
+				 * SO_PASSPIDFD), so just close the fd that
+				 * recvmsg installed.
+				 */
+				pr_debug("Closing pidfd %d received on queue peek\n", pidfd);
+				close(pidfd);
+				continue;
 			} else if (ch->cmsg_type == SCM_TIMESTAMP ||
 				   ch->cmsg_type == SCM_TIMESTAMPNS ||
 				   ch->cmsg_type == SCM_TIMESTAMPING) {
@@ -482,6 +497,7 @@ int dump_sk_queue(int sock_fd, int sock_id, int flags)
 {
 	int ret, size, orig_peek_off;
 	int exit_code = -1;
+	int passcred_set = 0;
 	void *data;
 	socklen_t tmp;
 
@@ -525,6 +541,37 @@ int dump_sk_queue(int sock_fd, int sock_id, int flags)
 		goto err_free;
 	}
 
+	/*
+	 * If the socket has SO_PASSPIDFD but not SO_PASSCRED, peeked packets
+	 * carry an SCM_PIDFD cmsg only, which loses the skb uid/gid and needs
+	 * a pidfd -> pid conversion. Both PASS* options are just recvmsg-time
+	 * views of the same pid/creds the kernel attached to the skb at send
+	 * time, so temporarily enable SO_PASSCRED to make every packet yield
+	 * a full SCM_CREDENTIALS cmsg instead.
+	 *
+	 * There is no family check here because dump_sk_queue() is only ever
+	 * called for unix sockets. Should that change, gate this on AF_UNIX
+	 * the way dump_socket_opts() does: since Linux 6.16 get/setsockopt of
+	 * SO_PASSPIDFD fails with EOPNOTSUPP on anything else.
+	 */
+	if (kdat.has_so_passpidfd) {
+		int passpidfd = 0, passcred = 0;
+
+		if (dump_opt(sock_fd, SOL_SOCKET, SO_PASSPIDFD, &passpidfd))
+			goto err_set_sock;
+
+		if (dump_opt(sock_fd, SOL_SOCKET, SO_PASSCRED, &passcred))
+			goto err_set_sock;
+
+		if (passpidfd && !passcred) {
+			int one = 1;
+
+			if (restore_opt(sock_fd, SOL_SOCKET, SO_PASSCRED, &one))
+				goto err_set_sock;
+			passcred_set = 1;
+		}
+	}
+
 	while (1) {
 		ret = dump_sk_queue_packet(sock_fd, sock_id, data, size, flags);
 		if (ret == 1)
@@ -535,6 +582,12 @@ int dump_sk_queue(int sock_fd, int sock_id, int flags)
 	exit_code = 0;
 
 err_set_sock:
+	if (passcred_set) {
+		int zero = 0;
+
+		if (restore_opt(sock_fd, SOL_SOCKET, SO_PASSCRED, &zero))
+			exit_code = -1;
+	}
 	/*
 	 * Restore original peek offset.
 	 */
--- a/criu/sk-queue.c
+++ b/criu/sk-queue.c
@@ -172,22 +172,29 @@ int sk_queue_post_actions(void)
 			struct pstree_item *item, *found = NULL;
 			SkUcredEntry *ue = pkt->entry->ucred;
 
-			for_each_pstree_item(item) {
-				if (item->pid->real == ue->pid) {
-					found = item;
-					break;
+			/*
+			 * A zero pid means the kernel attached no struct pid
+			 * to the skb (credless sender), there is nothing to
+			 * fix up, write the packet as is.
+			 */
+			if (ue->pid != 0) {
+				for_each_pstree_item(item) {
+					if (item->pid->real == ue->pid) {
+						found = item;
+						break;
+					}
 				}
-			}
 
-			if (!found) {
-				pr_warn("ucred: Can't find process with pid %d, ignoring packet\n",
-					ue->pid);
-				goto next;
-			}
+				if (!found) {
+					pr_warn("ucred: Can't find process with pid %d, ignoring packet\n",
+						ue->pid);
+					goto next;
+				}
 
-			pr_debug("ucred: Fixup ucred pids %d -> %d\n",
-				 ue->pid, vpid(item));
-			ue->pid = vpid(item);
+				pr_debug("ucred: Fixup ucred pids %d -> %d\n",
+					 ue->pid, vpid(item));
+				ue->pid = vpid(item);
+			}
 
 			ret = pb_write_one(img, pkt->entry, PB_SK_QUEUES);
 			if (ret < 0) {
--- a/criu/sk-queue.c
+++ b/criu/sk-queue.c
@@ -24,12 +24,25 @@
 #include "util.h"
 
 #include "sk-queue.h"
+#include "pidfd.h"
 #include "files.h"
 #include "protobuf.h"
 #include "images/sk-packet.pb-c.h"
 
 #undef LOG_PREFIX
 #define LOG_PREFIX "skqueue: "
+
+/*
+ * Result of inspecting the SCM_PIDFD cmsg the kernel hands us for a peeked
+ * packet. When the kernel gives a definitive answer about the sender we use
+ * it instead of the pstree-lookup-miss heuristic.
+ */
+struct pidfd_probe {
+	bool probed;   /* SCM_PIDFD gave a definitive liveness answer */
+	bool stale;    /* ... and the sender is dead */
+	bool has_exit; /* ... and we know its wait(2) exit status */
+	int exit_code;
+};
 
 struct sk_packet {
 	struct list_head list;
@@ -40,6 +53,14 @@ struct sk_packet {
 	};
 	unsigned scm_len;
 	int *scm;
+	bool pidfd_probed; /* SCM_PIDFD gave a definitive answer for this pkt */
+	bool pidfd_stale;  /* ... and the sender is dead */
+	/*
+	 * Dump side: pidfs state of the sender's struct pid, preallocated
+	 * along with the packet and attached to the entry once we know the
+	 * sender is dead (see sk_queue_post_actions()).
+	 */
+	PidfsAttrEntry *dead_pid;
 };
 
 static LIST_HEAD(packets_list);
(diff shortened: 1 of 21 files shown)


3idey:
The `sk-unix-dgram-pidfd-stale` CI failure was a wrong assumption in the test, not in the C/R logic: only kernels >= 6.17 can mint a pidfd for a reaped sender ("af_unix: enable handing out pidfds for reaped tasks in SCM_PIDFD"). Older kernels (CI runs ~6.8) put the `pidfd_prepare()` error into the cmsg payload the `-22` CI saw. Pre-dump recvmsg behaves the same way there, so C/R was already faithful on both kinds of kernels.

Force-pushed: the test now probes the kernel's behavior with a throwaway socketpair before the dump and asserts post-restore behavior matches full stale checks (ESRCH, fdinfo `Pid: -1`) on >= 6.17, negative cmsg payload on older kernels.

Two known limits of the stale-packet handling worth noting, besides the one in the commit message:

1. Stale detection infers death from a pstree lookup miss, so a recycled pid number could match the wrong process. Follow-up idea: read fdinfo of the pidfd we already receive (and close) during the dump peek to detect staleness definitively.
2. On >= 6.17, `PIDFD_GET_INFO`/`PIDFD_INFO_EXIT` on the restored stale pidfd reports the helper's SIGKILL instead of the original exit status. Could be fixed by saving the exit info from the peeked pidfd at dump time and having the helper exit the same way.

Happy to do these as follow-ups or fold them into this series, whichever you prefer.

adrianreber:
What about this error:

```
26-07-16T19:13:47.3649648Z ======================= Run zdtm/static/sk-netlink in h ========================
2026-07-16T19:13:47.3649713Z Start test
2026-07-16T19:13:47.3649774Z Test is SUID
2026-07-16T19:13:47.3649921Z ./sk-netlink --pidfile=sk-netlink.pid --outfile=sk-netlink.out
2026-07-16T19:13:47.3649984Z Run criu dump
2026-07-16T19:13:47.3650086Z =[log]=> dump/zdtm/static/sk-netlink/42/1/dump.log
2026-07-16T19:13:47.3650202Z ------------------------ grep Error ------------------------
2026-07-16T19:13:47.3650350Z b'(00.005072) 42 fdinfo 3: pos:                0 flags:           400000/0'
2026-07-16T19:13:47.3650561Z b'(00.005076) Dumping path for 3 fd via self 18 [/home/runner/work/criu/criu/test/zdtm/static]'
2026-07-16T19:13:47.3651128Z b'(00.005085) 42 fdinfo 4: pos:                0 flags:                2/0'
2026-07-16T19:13:47.3651282Z b'(00.005090) sockets: Searching for socket 0x90ea7 family 16'
2026-07-16T19:13:47.3651508Z b"(00.005098) Error (criu/sockets.c:644): sockets: Can't get 1:76 opt: Operation not supported"
2026-07-16T19:13:47.3651595Z b'(00.005102) sockets: No filter for socket'
2026-07-16T19:13:47.3651698Z b'(00.005105) ----------------------------------------'
2026-07-16T19:13:47.3651866Z b'(00.005116) Error (criu/cr-dump.c:1703): Dump files (pid: 42) failed with -1'
2026-07-16T19:13:47.3651939Z b'(00.005435) 42 was stopped'
2026-07-16T19:13:47.3652014Z b'(00.005550) net: Unlock network'
2026-07-16T19:13:47.3652099Z b'(00.005552) Unfreezing tasks into 1'
2026-07-16T19:13:47.3652171Z b'(00.005553) \tUnseizing 42 into 1'
2026-07-16T19:13:47.3652291Z b'(00.005564) Error (criu/cr-dump.c:2135): Dumping FAILED.'
2026-07-16T19:13:47.3652398Z ------------------------ ERROR OVER ------------------------
2026-07-16T19:13:47.3652561Z ################ Test zdtm/static/sk-netlink FAIL at CRIU dump #################
2026-07-16T19:13:47.3652650Z Test output: ================================
2026-07-16T19:13:47.3652654Z 
2026-07-16T19:13:47.3652720Z  <<< ================================
2026-07-16T19:13:47.3652795Z Send the 9 signal to  42
2026-07-16T19:13:47.3652911Z Wait for zdtm/static/sk-netlink(42) to die for 0.100000
```

I have not seen that one before.

3idey:
> What about this error:
> 
> ```
> 26-07-16T19:13:47.3649648Z ======================= Run zdtm/static/sk-netlink in h ========================
> 2026-07-16T19:13:47.3649713Z Start test
> 2026-07-16T19:13:47.3649774Z Test is SUID
> 2026-07-16T19:13:47.3649921Z ./sk-netlink --pidfile=sk-netlink.pid --outfile=sk-netlink.out
> 2026-07-16T19:13:47.3649984Z Run criu dump
> 2026-07-16T19:13:47.3650086Z =[log]=> dump/zdtm/static/sk-netlink/42/1/dump.log
> 2026-07-16T19:13:47.3650202Z ------------------------ grep Error ------------------------
> 2026-07-16T19:13:47.3650350Z b'(00.005072) 42 fdinfo 3: pos:                0 flags:           400000/0'
> 2026-07-16T19:13:47.3650561Z b'(00.005076) Dumping path for 3 fd via self 18 [/home/runner/work/criu/criu/test/zdtm/static]'
> 2026-07-16T19:13:47.3651128Z b'(00.005085) 42 fdinfo 4: pos:                0 flags:                2/0'
> 2026-07-16T19:13:47.3651282Z b'(00.005090) sockets: Searching for socket 0x90ea7 family 16'
> 2026-07-16T19:13:47.3651508Z b"(00.005098) Error (criu/sockets.c:644): sockets: Can't get 1:76 opt: Operation not supported"
> 2026-07-16T19:13:47.3651595Z b'(00.005102) sockets: No filter for socket'
> 2026-07-16T19:13:47.3651698Z b'(00.005105) ----------------------------------------'
> 2026-07-16T19:13:47.3651866Z b'(00.005116) Error (criu/cr-dump.c:1703): Dump files (pid: 42) failed with -1'
> 2026-07-16T19:13:47.3651939Z b'(00.005435) 42 was stopped'
> 2026-07-16T19:13:47.3652014Z b'(00.005550) net: Unlock network'
> 2026-07-16T19:13:47.3652099Z b'(00.005552) Unfreezing tasks into 1'
> 2026-07-16T19:13:47.3652171Z b'(00.005553) \tUnseizing 42 into 1'
> 2026-07-16T19:13:47.3652291Z b'(00.005564) Error (criu/cr-dump.c:2135): Dumping FAILED.'
> 2026-07-16T19:13:47.3652398Z ------------------------ ERROR OVER ------------------------
> 2026-07-16T19:13:47.3652561Z ################ Test zdtm/static/sk-netlink FAIL at CRIU dump #################
> 2026-07-16T19:13:47.3652650Z Test output: ================================
> 2026-07-16T19:13:47.3652654Z 
> 2026-07-16T19:13:47.3652720Z  <<< ================================
> 2026-07-16T19:13:47.3652795Z Send the 9 signal to  42
> 2026-07-16T19:13:47.3652911Z Wait for zdtm/static/sk-netlink(42) to die for 0.100000
> ```
> 
> I have not seen that one before.

Yes, that one is on me, since kernel 6.16 (commit 7d8d93fdde50) SO_PASSPIDFD is restricted to AF_UNIX and returns EOPNOTSUPP on netlink sockets, but I was probing it in the shared unix/netlink branch of dump_socket_opts(). That's why this job was failing. I will make it for AF_UNIX only.

3idey:
@mihalicyn, i think you force pushed by accident:)

criu/sk-queue.c
```diff
@@ -381,6 +455,41 @@ static int dump_packet_cmsg(struct msghdr *mh, SkPacketEntry *pe, int flags)
...
+					 * for an already reaped sender and put a
+					 * negative errno in the payload: the
+					 * sender is definitively dead and there
+					 * is no fd to close.
+					 */
+					probe->stale = true;
+				} else {
+					int exit_code;
+
+					if (pidfd_query_exit(pidfd, &exit_code) > 0) {
```
mihalicyn: If `pidfd_query_exit()` fails it does this silently for us. Should we make use of `kdat.has_pidfd_get_info`?

mihalicyn:
> i think you force pushed by accident:)

no, it's just because `criu-dev` branch was force-pushed because of a minor CRIU release

criu/sk-queue.c
3idey: I'll add it. That makes sense, as you said.

images/sk-packet.proto
```diff
@@ -11,6 +11,12 @@ message sk_ucred_entry {
 	required uint32		uid		= 1;
 	required uint32		gid		= 2;
 	required uint32		pid		= 3;
+	// The sender was dead at dump time. Its pid is not restored but
+	// replaced with a short-lived helper's pid, so a receiver still
+	// mints a pidfd (and reads SCM_CREDENTIALS) of a dead process.
+	optional bool		stale		= 4;
```
mihalicyn: This doesn't look right to me, because see:
- `message scm_entry` mirrors kernel's `struct scm_cookie`
- `message sk_ucred_entry` mirrors kernel's `struct scm_creds` + `(struct scm_cookie).pid` field
(all this if a unix socket subsystem)

at the same time, process's exit code information lays in completely different subsystem of the kernel - pidfd:
- `struct pidfs_anon_attr`
- `struct pidfs_attr`

How they are connected? Right, via `struct pid`. (-;

Why this is important? Because this is not just about data structures organization or code formalities, I also see that you are creating a helper processes from sk-queue management code and this means that we do a layering violation - pidfd-related features are implemented for SCM_PIDFD case, but completely out of scope for generic pidfd (don't forget that we might deal with dead pidfd not only in the context of SCM_PIDFD!).

mihalicyn (changes requested):

criu/sk-queue.c
```diff
@@ -649,13 +858,22 @@ int restore_sk_queue(int fd, unsigned int peer_id)
 
 	list_for_each_entry_safe(pkt, tmp, &packets_list, list) {
 		SkPacketEntry *entry = pkt->entry;
+		pid_t dead_pid = 0;
 
 		if (entry->id_for != peer_id)
 			continue;
 
+		if (entry->dead_pid) {
+			dead_pid = dead_pid_get(&dead_pids, entry->dead_pid);
```
mihalicyn: Idea of this code is clear, but not fully correct. You can't lookup dead_pid just from attributes, because you can have two different processes with the same exit code, for instance. And what happens here is that user at the end will get the same pidfd at the end.

A correct way to identify `struct pid` these days is to rely on inode number (`ino`), see for instance https://github.com/torvalds/linux/blob/cee9395acd8043be0644b25c34bfa86623f2b935/fs/pidfs.c#L897.

In fact, we already do this in `lookup_dead_pidfd` function. You can try to do the same and hold a global hashtable using `ino` as an index, for example.

But please, before changing your code and reworking this, try to come up with an extra zdtm test reproducing the issue I'm describing here theoretically ;-)

mihalicyn (changes requested):

criu/sk-queue.c
3idey: I Wrote the test, it fails exactly as you described: https://github.com/3idey/criu/tree/scm-pidfd-ino-repro
Three senders all exit with 42, plus a held pidfd on the first one. After restore the held pidfd and its own 
queued packet come back as two different processes:

FAIL: sk-unix-dgram-pidfd-ino.c:443: restored pidfd 47242 and its queued packet 47243 are no longer the same process.

I pushed it to a different branch first if you wanted to take a look i will continue myself changing the code and rework it:).

3idey: what if two tasks holding queued packets from the same dead sender still get a stand-in each, so the receivers see two inos. I have a test for it in https://github.com/3idey/criu/tree/scm-pidfd-xtask-repro.
you can take a look:)

3idey: I guess I'm on my way to figure it out:)

3idey: so what changed is: stand-in processes for dead pids are now keyed on the pidfs inode and kept in one table shared by the whole restore, instead of a table private to each task. So if two tasks hold packets from the same dead sender, they both get the same stand-in and the receivers see one inode, not two. The root task forks them after the tree is up and reaps them once everyone has restored their files.

A task restoring outside the root pid namespace still gets private stand-ins, since a pid from another namespace is not useful to it. Same behaviour as before for that case.

while working I fixed this: a pidfd with no inode recorded no longer shares a stand-in with other such entries, and we now reject an exit status a stand-in cannot actually reproduce instead of silently restoring it as exit(1).

New tests: sk-unix-dgram-pidfd-ino (two senders that died the same way stay distinct) and sk-unix-dgram-pidfd-xtask (one dead sender across two tasks stays one inode).

3idey:
@mihalicyn, The CI is keep failing but I don't think this is my pr doing though:)

what i pushed yesterday was stuff i missed but appeared when i took another look:

 - The pidfs cookie is 64 bits but i_ino is a long, so on a 32-bit kernel the high half lives in i_generation. I now read it back with FS_IOC_GETVERSION and fold it in, otherwise two struct pids 2^32 apart would look like one. On 64-bit the generation is 0 so nothing changes there.
 - I also read the ino from the fd itself instead of the ino: line in fdinfo, so a pidfd file and a queued packet are keyed the exact same way and land on the same stand-in.

Plus a fallback for when there is no ino at all, which is every dead sender on a kernel that mints no pidfd for a reaped one. I save the sender's dump host pid and use it only to match ino-less entries against each other. It is not a real identity since pid numbers get reused, but without it every packet of one dead sender would fork its own stand-in. An ino-less entry is never matched against one that has an ino.

If two references claim different exit codes for the same dead pid it warns and keeps the first, since the second guess is no better.

The rest is comments and a cleanup of the pid branches in sk_queue_post_actions(), no behaviour change there.

images/pidfd.proto
```diff
@@ -3,11 +3,32 @@
...
+	 *
+	 * Read with pidfd_query_ino(), which yields the whole 64-bit cookie:
+	 * fstat() gives i_ino and, on a 32-bit kernel where that is only its
+	 * low half, FS_IOC_GETVERSION gives back the high half that
+	 * pidfs_init_inode() put in i_generation. A queued packet's sender is
+	 * identified the same way, so a pidfd file and a packet of one dead
+	 * process agree on the key. The ino: line of /proc/pid/fdinfo/N, which
+	 * prints i_ino alone, is the fallback if the fd can't be stat()ed.
+	 */
+	required uint64		ino		= 2;
```
mihalicyn: I don't think that we can just change a type of a **required** field. This will break compatibility.
We need to think about adding another `uint32` `optional` field and then build `ino` number from two fields in a compatible way (with an appropriate fallback codepath ofc).

test/zdtm/static/sk-unix-dgram-pidfd-ino.c
```diff
@@ -0,0 +1,474 @@
...
+#endif
+
+/*
+ * PIDFD_GET_INFO (pidfs ioctl, Linux 6.13) reports the exit status pidfs kept
+ * for a process that has already been reaped. The struct layout is versioned
+ * by size and encoded in the ioctl number, so use our own copy of the first
+ * published 64-byte version -- exit_code is its last field -- rather than
+ * whatever <linux/pidfd.h> the build host happens to have.
+ */
+struct zdtm_pidfd_info {
```
mihalicyn: I see that we have to define `struct zdtm_pidfd_info` 4 times.
And then a helper function that uses `ZDTM_PIDFD_GET_INFO` also defined a few times.

I think we can try to put all these common helpers into a separate file somewhere in `test/zdtm/lib`.

mihalicyn (changes requested):

3idey:
@mihalicyn, Found one more thing while testing on older kernels: before 6.9 there is no pidfs, so every pidfd sits on the same anon inode and has the same ino. That made all dead pidfds of a dump restore onto a single stand-in.

pidfd_query_ino() now checks that the pidfd is on pidfs, and records ino 0 (no identity) otherwise. I checked it on a 6.8 VM: two dead processes now get two stand-ins instead of one. zdtm can't catch this since pidfd_dead skips without pidfs.

Also dropped the proc_parse.c change, since the ino now always comes from the fd itself. ino_hi stays as you suggested.:)

PR #3094, opened 2026-07-17
test/zdtm: fix intermittent fpu01 failure due to VZEROUPPER

adrianreber:
The fpu01 test loads 256-bit YMM registers, calls test_daemon() and test_waitsig() as separate C function calls, then reads YMM back after checkpoint/restore.  With separate asm blocks and C calls between them the compiler is free to insert VZEROUPPER at function call boundaries, which zeros the upper 128 bits of all YMM registers before CRIU has a chance to dump them.

The symptom is that only the lower 128 bits (XMM portion) survive:
  ymm1='code9Ee5sohphie1ae1kaeMahngoh5oe'
  ymm3='code9Ee5sohphie1'              (upper half zeroed)

Fix this by combining the vmovapd loads, function calls, and vmovapd stores into a single asm volatile block with indirect calls through function pointers.  Inside an asm block the compiler cannot insert any instructions, so VZEROUPPER is avoided.  The compiler places the function pointers in callee-saved registers (rbx/rbp) so they survive across calls.

Use arch-specific clobber lists so the test also compiles correctly in compat mode (-m32) where CRIU tests checkpoint/restore of 32-bit applications on x86_64: the 32-bit build clobbers eax/ecx/edx and xmm0-xmm7, while the 64-bit build clobbers rax/rcx/rdx/rsi/rdi/r8-r11 and xmm0-xmm15.


adrianreber:
@mihalicyn You mentioned this error somewhere. I also saw fpu01 failing a couple of times now.

adrianreber:
This is not ready, yet.

avagin:
What kernel is there? Does it have https://lkml.org/lkml/2026/4/29/87?

adrianreber:
Ah, thanks for that pointer. I was still trying to understand when it happens and why.

avagin:
If it is the same issue, it should be triggered on newer Intel cpus with the AMX support.

avagin:
@adrianreber if it is a compiler behavior, the test would fail each time for the specific test job. If it fails from time to time, it is probably something else.  In this case, symptoms match the kernel issue that I mentioned before.

adrianreber:
Closing. Hard to reproduce and probably kernel related as mentioned.

PR #3096, opened 2026-07-18
criu: force link-remap ON for nvproxy/vLLM container C/R

inferx-net:
runc/crun invoke criu via the swrk RPC mode, which neither passes --link-remap on the command line nor reads the criu config file. Python/vLLM creates deleted POSIX semaphores in /dev/shm (e.g. /dev/shm/sem.* (deleted)) that criu cannot dump without link-remap. Force opts.link_remap_ok = 1 so container checkpoint of vLLM under nvproxy succeeds.

### DIFF
--- a/criu/cr-service.c
+++ b/criu/cr-service.c
@@ -506,6 +506,15 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 
 	if (req->has_link_remap)
 		opts.link_remap_ok = req->link_remap;
+	/*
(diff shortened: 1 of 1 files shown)


adrianreber:
Just hardcoding seems pretty unlikely to be accepted. Also, why do you say that the configuration file is not read? That is not true.

I think this needs to be handled differently. Either be the container runtime or engine.

inferx-net:
Open by mistake.

PR #3099, opened 2026-07-19
tree-wide: Makefile: use -MP CFLAG to produce .d files

mihalicyn:
Let's use -MP gcc flag when generating dependency (.d) files. According to the manual:
```
       -MP This option instructs CPP to add a phony target for each
           dependency other than the main file, causing each to depend on
           nothing.  These dummy rules work around errors make gives if
           you remove header files without updating the Makefile to
           match.
```
This fixes incremental rebuild failures like:
```
make[2]: 'compel/compel-host-bin' is up to date.
  DEP      criu/arch/x86/crtools.d
  CC       criu/arch/x86/crtools.o
  LINK     criu/arch/x86/crtools.built-in.o
make[3]: Nothing to be done for 'all'.
make[3]: *** No rule to make target 'criu/include/linux/rseq.h', needed by 'criu/pie/parasite.o'.  Stop.
make[2]: *** [criu/Makefile:59: pie] Error 2
```

Usually, after header files renames or source directory tree reorganization.

### DIFF
--- a/images/Makefile
+++ b/images/Makefile
@@ -118,7 +118,7 @@ ifeq ($(PROTOUFIX),y)
 endif
 $(obj)/$(1).pb-c.d: $(obj)/$(1).pb-c.c $(addsuffix .pb-c.d,$(addprefix $(obj)/,$(2))) $(makefile-deps)
 	$$(E) "  DEP     " $$@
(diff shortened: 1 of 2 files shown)


rst0git (approved):
LGTM

PR #3100, opened 2026-07-21
criu: add dump fallbacks for rootless containers

arcceus:
Follow-up to #3073, split out from #3057.

This PR contains the CRIU-side dump changes needed when seccomp suspend is not
available during dump.

Main pieces:

- fall back to `process_vm_readv` when `PTRACE_O_SUSPEND_SECCOMP` is denied
- propagate the seccomp-suspend failure from seize into memory dumping
- add diagnostics for seccomp-related parasite failures

### DIFF
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -526,7 +526,7 @@ static int generate_vma_iovs(struct pstree_item *item, struct vma_area *vma, str
 	 */
 
 	if (!(vma->e->prot & PROT_READ)) {
-		if (mem_read_externally())
+		if (mem_read_externally() && pre_dump)
 			return 0;
 		if ((parent_predump_mode == PRE_DUMP_READ && opts.pre_dump_mode == PRE_DUMP_SPLICE) || !pre_dump)
 			has_parent = false;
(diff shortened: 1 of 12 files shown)


rst0git:
@arcceus I converted this PR to draft as it depends on https://github.com/checkpoint-restore/criu/pull/3073

rst0git:
@arcceus Would it be possible to drop `(cherry picked from commit ...)` from the commit messages?

criu/mem.c
```diff
@@ -356,6 +356,15 @@ prep_dump_pages_args(struct parasite_ctl *ctl, struct vm_area_list *vma_area_lis
 	return args;
 }
 
+static bool mem_read_externally(void)
+{
+	if (opts.seccomp_suspend_failed)
+		return true;
```
rst0git, line 362: @arcceus Would it make sense to always use `process_vm_readv` when `--unprivileged` is used?

arcceus: On faliure seems better to me, if they still want to use process_vm_readv anyhow they can use --memory-dump-mode read. This one is seccomp suspension failed so the vmsplice page dump path may be blocked by the target seccomp filter by default (which it is in most of the cases) unless explicitly allowed.

PR #3000, opened 2026-04-14
Amdgpu plugin cleanups and fixes

tursulin:
Continuing after v2 from mailing list:

https://lore.kernel.org/criu/CANaxB-ybBhOAZKRUGFN4hbzhkJZmH9kkHeprfihT6LG=UThYrg@mail.gmail.com/T/#m86b6183e681a617626eab6e7cafcf7fef4be8dfd

v3:
 * Two new patches to fix to new pre-existing issues flagged by copilot.
    * Wrong shared allocation size and missing error handling.
    * Check image unpack worked during probe before dereferencing. 
 * Improved seek error handling.
 * Added O_TRUNC when opening image file for write.
 * Handle short reads/writes.

PR #3101, opened 2026-07-21
criu: add --rootless-container mode

arcceus:
Follow-up to #3100, split out from #3057.

This PR adds the CRIU-side base flag for rootless container checkpoint/restore. The Podman/crun integration is being kept separate.

Main pieces:
- add --rootless-container
- add the initial user namespace preflight for rootless container targets
- use criu_is_unprivileged_or_rootless() for shared EPERM tolerance checks

--rootless-container means the target is a rootless container. It is separate from --unprivileged, which describes how CRIU itself is running.

### DIFF
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -185,10 +185,13 @@ In addition, *page-server* options may be specified.
     not passed the memory tracker get turned on implicitly.
 
 *--pre-dump-mode*='mode'::
-    There are two 'mode' to operate pre-dump algorithm. The 'splice' mode
-    is parasite based, whereas 'read' mode is based on process_vm_readv
-    syscall. The 'read' mode incurs reduced frozen time and reduced
-    memory pressure as compared to 'splice' mode. Default is 'splice' mode.
+*--memory-dump-mode*='mode'::
+    Alias for *--pre-dump-mode*. Applies to both dump and pre-dump
+    operations. There are two 'mode' to operate memory dumping algorithm.
+    The 'splice' mode is parasite based, whereas 'read' mode is based on
+    process_vm_readv syscall. The 'read' mode incurs reduced frozen
+    time and reduced memory pressure as compared to 'splice' mode.
+    Default is 'splice' mode.
 
 *dump*
 ~~~~~~
--- a/Documentation/criu.txt
+++ b/Documentation/criu.txt
@@ -184,14 +184,14 @@ In addition, *page-server* options may be specified.
     Turn on memory changes tracker in the kernel. If the option is
     not passed the memory tracker get turned on implicitly.
 
-*--pre-dump-mode*='mode'::
 *--memory-dump-mode*='mode'::
-    Alias for *--pre-dump-mode*. Applies to both dump and pre-dump
-    operations. There are two 'mode' to operate memory dumping algorithm.
-    The 'splice' mode is parasite based, whereas 'read' mode is based on
+*--pre-dump-mode*='mode'::
+    There are two modes to operate the memory dumping algorithm. The
+    'splice' mode is parasite based, whereas 'read' mode is based on
     process_vm_readv syscall. The 'read' mode incurs reduced frozen
     time and reduced memory pressure as compared to 'splice' mode.
-    Default is 'splice' mode.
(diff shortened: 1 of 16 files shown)


adrianreber:
Thanks for the split, but do we really need the flag `--rootless-container`? Can this be autodetected somehow?

From this PR which is just one commit (minus the changes from #3000) I cannot see what it does. It just introduces the command-line option.

Can't you recreate the rootless container namespace setup using `unshare` and introduce a test how your change supports this.

Can you describe it in more detail why this is needed and how it connects to rootless containers?

arcceus:
I will open the next PR aswell with the unshare test, where this flag is firstly used for namespace work.

Later patches use this to tolerate map_files access, the helper and broker process for namespace operations that must happen form inside the container's userns, avoid messing with host and runtime owned state (cgroups, binfmtmsc etc).

The runtime knows it is checkpointing/restoring a rootless container, so passing this to CRIU explicitly seems better than trying to autodetect it.

adrianreber:
> The runtime knows it is checkpointing/restoring a rootless container, so passing this to CRIU explicitly seems better than trying to autodetect it.

I am not totally convinced. If there is no way to automatically handle it, it makes sense. But it feels like it should be doable.

arcceus:
> I am not totally convinced. If there is no way to automatically handle it, it makes sense. But it feels like it should be doable.

In my current patches:
1) crun creates the external userns from the OCI mappings and pass it to CRIU with join-ns.
2) Podman handles cgroup placement after restore.
3) Podman completes netns setup after restore for PostConfigureNetNS.
4) crun provides seccomp BPF generated from the OCI spec.
5) crun stages runtime managed mounts so CRIU can resolve them after joining the container userns.

I think some uses of the flag can probably be changed to auto detect. The netns broker in #3103 is one of them.

I can change the broker paths to autodetect where the condition is clear. For the cases where CRIU would otherwise just be guessing from uid maps/caps/EPERM, I think we still need some explicit runtime context.

Could you let me know which part exactly should be autodetectable?

adrianreber:
This is just my opinion and I might be wrong, but it seems to be the wrong approach. Sorry for being so hard on you. Also, at this point you are the expert about what the container engine and runtime needs for rootless checkpoint restore.

Looking at CRIU the concept of a container is not very often mentioned which might be related to the fact that a container is also not something Linux knows about. In fact one of the hard problems is that nobody knows what a container is. It is combination of namespace, cgroups, seccomp and other things. For me it seems the right approach would be to less focus on the point of view from the container runtime. Focus on what things CRIU must support to enable a container runtime to checkpoint and restore a container as non root. For each resource that requires special handling add that to CRIU independently if possible.

Concerning you question:

> Could you let me know which part exactly should be autodetectable?

I don't know, but if it is doable do not introduce new command-line switches.

I would recommend figure out which resources need special handling from CRIU's side when running as non-root. All those resources are not necessarily part of a container. You mention namespace, cgroups and BPF. Handle those separately. Create those resources for a test without a container runtime (if possible) and make it work as non-root. Once one thing works add a second and extend the tests for each resource and maybe combined. You never know what a user wants to do and one flag to handle all different use cases will probably not work.

Again, this is just my point of view and I might be wrong, but `--rootless-container` feels like the wrong approach. CRIU has options to influence how container related resources are handled and to me extending them to handle rootless containers in the end seems the right approach.

arcceus:
> This is just my opinion and I might be wrong, but it seems to be the wrong approach. Sorry for being so hard on you. Also, at this point you are the expert about what the container engine and runtime needs for rootless checkpoint restore.

No No, I am honestly very thankful for you pushback. By no means I am even close to the level of experience you have so I look forward to all your opinions and feedbacks. 

> Looking at CRIU the concept of a container is not very often mentioned which might be related to the fact that a container is also not something Linux knows about. In fact one of the hard problems is that nobody knows what a container is. It is combination of namespace, cgroups, seccomp and other things. For me it seems the right approach would be to less focus on the point of view from the container runtime. Focus on what things CRIU must support to enable a container runtime to checkpoint and restore a container as non root. For each resource that requires special handling add that to CRIU independently if possible.

This actually made alot of sense thank you for explaining it that way! I have stripped down all the flag related changes

PR #3102, opened 2026-07-22
vma: dump/restore content of MAP_SHARED tmpfs mappings

### LINKED ISSUE #3095: Robust pthread_mutex in /dev/shm causes EOWNERDEAD after CRIU checkpoint/restore
Dear CRIU community,

We observed an unexpected `EOWNERDEAD` when checkpointing a container that uses a `PTHREAD_MUTEX_ROBUST` + `PTHREAD_PROCESS_SHARED` mutex stored in `/dev/shm`: any thread blocked on the mutex at checkpoint time wakes up with `EOWNERDEAD` after restore, as if the owner had crashed.

The `/dev/shm` contents are not saved by CRIU itself, but by the container runtime (Podman), which snapshots and restores the shm files independently.

At the end of the post you can find the c code and sh script to reproduce this issue.

Our system:
- CRIU 4.1.1
- Kernel: 6.12.17
- Architecture: aarch64

### On checkpoint

```mermaid
sequenceDiagram
   autonumber
   participant Criu as CRIU
   participant Owner as Main thread<br/>(PID1 in container)
   participant Mutex as futex word<br/>(in shm)
   participant Worker as Worker thread
   participant Kernel as Kernel

   Owner->>Mutex: pthread_mutex_lock
   Note left of Mutex: futex word:<br/>0x00000001
   Worker->>Mutex: pthread_mutex_lock (blocks)
   Note right of Mutex: futex word:<br/>0x80000001

   Note over Criu,Kernel: CRIU DUMP

   destroy Owner
   Criu->>Owner: kill
   Kernel->>Mutex: stamp OWNER_DIED
   Note right of Mutex: futex word:<br/>0xc0000000

   destroy Worker
   Criu->>Worker: kill

   Note over Criu,Kernel: Podman saves poisoned shm
```

### On restore

```mermaid
sequenceDiagram
   autonumber
   participant Criu as CRIU
   participant Mutex as futex word<br/>(in shm)

   Note over Criu,Mutex: Podman restores poisoned shm
   Note right of Mutex: futex word:<br/>0xc0000000

   Note over Criu,Mutex: CRIU RESTORE
   create participant Owner as Main thread<br/>(PID1 in container)
   Criu->>Owner: restore
   create participant Worker as Worker thread
   Criu->>Worker: restore

   Mutex-->>Worker: acquired with EOWNERDEAD
   Note right of Mutex: futex word:<br/>0xc0000000
```

## Mitigation

A solution that works for our case is to just remove the robust_list from the threads before they are killed by CRIU.
That way, when the threads are killed, the kernel won't mark the content of the shm as OWNERDEAD.

The syscall to clear the list would be:

`sys_set_robust_list(NULL, sizeof(struct robust_list_head));`

Unfortunately, it has to be called in the context of the thread which list has to be removed... And to do this, a new
parasite_cmd is needed.

I tested this approach and the problem disappears. But this approach is only one way: if the checkpoint fails, the
process shouldn't be kept alive because the robust lists of its threads will be gone. This means, to make it mainline-ready,
there should be a way to restore the lists again... But this would probably mean to parasite the threads again after
checkpoint failed.

**Has anyone seen this before, or have ideas on how to handle it properly?**

## Program to reproduce

```c
// SPDX-License-Identifier: CC0-1.0
//
// gcc -std=c11 main.c -o main

#define _GNU_SOURCE
#include <errno.h>
#include <fcntl.h>
#include <pthread.h>
#include <signal.h>
#include <stdio.h>
#include <sys/mman.h>
#include <unistd.h>
#include <stdatomic.h>

pthread_mutex_t *g_mutex = NULL;
_Atomic(int) g_workerLockResult = -1; // -1 = still blocked

void *workerFn(void *)
{
    atomic_store(&g_workerLockResult, pthread_mutex_lock(g_mutex));
    pthread_mutex_unlock(g_mutex);
    return NULL;
}

int main(void)
{
    // We will wait on SIGUSR1...
    sigset_t set;
    sigemptyset(&set);
    sigaddset(&set, SIGUSR1);
    pthread_sigmask(SIG_BLOCK, &set, NULL);

    const char *const shmName = "/shm_with_mutex_inside";
    shm_unlink(shmName);
    const int fd = shm_open(shmName, O_RDWR | O_CREAT | O_EXCL, 0600);
    if (fd < 0)
    {
        perror("shm_open");
        return 1;
    }
    ftruncate(fd, sizeof(pthread_mutex_t));
    g_mutex = (pthread_mutex_t *)
        mmap(NULL, sizeof(pthread_mutex_t),
             PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
    close(fd);

    pthread_mutexattr_t attr;
    pthread_mutexattr_init(&attr);
    pthread_mutexattr_setrobust(&attr, PTHREAD_MUTEX_ROBUST);
    pthread_mutexattr_setpshared(&attr, PTHREAD_PROCESS_SHARED);

    // Create mutex in shm
    pthread_mutex_init(g_mutex, &attr);
    pthread_mutexattr_destroy(&attr);

    // 1. Main takes the mutex
    pthread_mutex_lock(g_mutex);

    // 2. Worker thread will now try to take the mutex.
    pthread_t worker;
    pthread_create(&worker, NULL, workerFn, NULL);
    usleep(200000);

    // At this point the worker thread should be blocked waiting for the mutex

    printf("===C/R Window===\n");
    printf("To continue program afterwards: kill -USR1 %d\n", getpid());
    fflush(stdout);

    int sig = 0;
    sigwait(&set, &sig);

    pthread_mutex_unlock(g_mutex); // Returns error if worker thread locked/unlocked in OWNERDEAD case.
    pthread_join(worker, NULL);

    const int rc = atomic_load(&g_workerLockResult);
    if (rc == 0)
        printf("TEST PASSED: worker acquired normally\n");
    else if (rc == EOWNERDEAD)
        printf("TEST FAILED: worker woke with EOWNERDEAD (%d)\n", rc);
    else
        printf("UNEXPECTED: rc=%d\n", rc);
    fflush(stdout);

    pthread_mutex_destroy(g_mutex);
    munmap(g_mutex, sizeof(pthread_mutex_t));
    shm_unlink(shmName);
    return rc;
}
```

```sh
#!/bin/sh
# SPDX-License-Identifier: CC0-1.0

peek_mutex() {
    od -An -tx4 -N4 | awk '{printf "futex word: 0x%s\n", $1}'
}

sudo podman run --network=host --name criu-test -d --replace \
    -v "$(pwd)/main:/tmp/main" debian:trixie /tmp/main

USERDATA=$(sudo podman container inspect --format "{{.StaticDir}}" criu-test)

echo "=== before checkpoint ==="
sudo podman exec criu-test cat /dev/shm/shm_with_mutex_inside | peek_mutex

sudo podman container checkpoint --keep criu-test

echo "=== in checkpoint image ==="
sudo tar -xOf "$USERDATA/devshm-checkpoint.tar" ./shm_with_mutex_inside | peek_mutex

sudo podman container restore --keep criu-test

sudo podman exec criu-test sh -c "kill -USR1 1"
EXIT=$(sudo podman wait criu-test)
if [ "$EXIT" -eq 0 ]; then
    echo "TEST PASSED"
elif [ "$EXIT" -eq 130 ]; then
    echo "TEST FAILED: worker woke with EOWNERDEAD"
else
    echo "UNEXPECTED exit code: $EXIT"
fi
```

## Result
```
a07d43e6d766f76906e832abbe35213dd5f46109760883886193d55ce6306bc6
=== before checkpoint ===
futex word: 0x80000001
criu-test
=== in checkpoint image ===
futex word: 0xc0000000
criu-test
TEST FAILED: worker woke with EOWNERDEAD

```
Regards,
specht478

rst0git:
CRIU currently treats a MAP_SHARED mapping of a regular file as external state. It reopens the backing file on restore but never touches its content, assuming whatever wrote the file stays responsible for it. This assumption breaks for shared memory in `/dev/shm`,  where the file's bytes are process-owned state, not something external.

This pull request adds `VMA_FILE_SHARED_TMPFS` for `MAP_SHARED` mappings whose backing file lives on tmpfs. Using the same per-task page-pipe path used for private memory, the content of such files is saved (before the task is killed). On restore, the content written back to the file before it is `mmap()`'d. This approach covers both file reopen and the ghost-file remap path.

Fixes: #3095

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -47,6 +47,8 @@
 #include "fs-magic.h"
 #include "namespaces.h"
 #include "proc_parse.h"
+#include "pagemap.h"
+#include "page.h"
 #include "pstree.h"
 #include "string.h"
 #include "fault-injection.h"
@@ -2464,6 +2466,81 @@ void filemap_ctx_fini(void)
 	}
 }
 
+/*
+ * A VMA_FILE_SHARED_TMPFS mapping is dumped like shared memory (see
+ * generate_vma_iovs() in mem.c), because its tmpfs-backed file can be
+ * mutated as a side effect of the dumped task dying (e.g. the kernel's
+ * exit_robust_list() rewriting a held PTHREAD_MUTEX_ROBUST futex word in
+ * place). Write the dumped content back into the just-reopened file so the
+ * live file matches the pre-dump snapshot before anyone -- including the
+ * restored task itself -- can observe it.
+ */
+static int restore_tmpfs_shared_vma_content(int pid, struct vma_area *vma, int fd)
+{
+	struct page_read pr;
+	unsigned long vstart = vma->e->start;
+	unsigned long vend = vma->e->end;
+	unsigned long va;
+	void *buf;
+	int ret;
+
+	ret = open_page_read(pid, &pr, PR_TASK);
+	if (ret <= 0)
+		return ret < 0 ? -1 : 0;
+
+	/*
+	 * seek_pagemap() leaves pr positioned exactly on the entry covering
+	 * @vstart (pr->cvaddr == vstart) -- unlike advance(), it must not be
+	 * followed by another advance() before the first read.
+	 */
+	ret = pr.seek_pagemap(&pr, vstart);
+	if (ret <= 0) {
+		ret = ret < 0 ? -1 : 0;
+		goto out_close;
+	}
+
+	buf = xmalloc(PAGE_SIZE);
+	if (!buf) {
+		ret = -1;
+		goto out_close;
+	}
+
+	va = vstart;
+	while (va < vend) {
+		unsigned long entry_end = (unsigned long)decode_pointer(pr.pe->vaddr) + pr.pe->nr_pages * PAGE_SIZE;
+		unsigned long chunk_end = entry_end < vend ? entry_end : vend;
+
+		while (va < chunk_end) {
+			off_t foff = vma->e->pgoff + (va - vstart);
+
+			ret = pr.read_pages(&pr, va, 1, buf, 0);
+			if (ret < 0)
+				goto out_free;
+
+			if (pwrite(fd, buf, PAGE_SIZE, foff) != PAGE_SIZE) {
+				pr_perror("Can't write back tmpfs shared content for %#lx", va);
+				ret = -1;
+				goto out_free;
+			}
+
+			va += PAGE_SIZE;
+		}
+
+		if (va >= vend)
+			break;
+
+		ret = pr.advance(&pr);
+		if (ret <= 0)
+			break;
+	}
+	ret = pr.sync(&pr);
+out_free:
+	xfree(buf);
+out_close:
+	pr.close(&pr);
+	return ret;
+}
+
 static int open_filemap(int pid, struct vma_area *vma)
 {
 	u32 flags;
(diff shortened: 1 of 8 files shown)


avagin:
Nack. It is a wrong approach. /dev/shm can be shared with other processes that haven't been dumped. Plus, #3095 isn't just about /dev/shm. https://github.com/avagin/criu/commit/77a28fafcb632577077bcbbc57975f7b595582bd is a reproducer that doesn't use tmpfs files.

PR #3103, opened 2026-07-22
criu: broker netns sockets through target userns

arcceus:
Follow-up to #3100.

This adds a userns broker helper and uses it when CRIU cannot directly enter and return from the target network namespace.

In that case, CRIU forks a helper that enters the target userns first, then the target netns, creates the netlink/parasite sockets there, and passes them back with SCM_RIGHTS.

### DIFF
--- a/criu/Makefile.crtools
+++ b/criu/Makefile.crtools
@@ -48,6 +48,7 @@ obj-y			+= namespaces.o
 obj-y			+= asyncd.o
 obj-y			+= netfilter.o
 obj-y			+= net.o
+obj-y			+= netns-broker.o
 obj-y			+= userns-broker.o
 obj-y			+= pagemap-cache.o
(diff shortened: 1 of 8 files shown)


adrianreber:
Is the last commit here something that could be done standalone. This looks like something that is a simple change with a clear goal. Can this be the first PR with a test? Or does this rely on the changes to the memory handling?

arcceus:
~Does not depend on the memory handling changes. Only depends on #3101 for the --rootless-container. The broker itself is only for dump-time netns socket setup.~

The test is in the runtime-userns follow-up branch, because this only changes dump-time socket preparation. A test here would be dump-only/noauto, so I kept the full test with the first restore-side userns branch.

https://github.com/arcceus/criu/tree/split/rootless-test-coverage

arcceus:
Reworking the split will reopen later.

PR #3105, opened 2026-07-24
seize, cr-restore: add sentinel notification for batched device plugins

AlanFokCo:
## Problem

The `CHECKPOINT_DEVICES` and `RESUME_DEVICES_LATE` plugin hooks are called once per task in a loop. Plugins that must act only after *all* device PIDs have been iterated — for example to run a batched/parallel checkpoint across several GPUs — have no reliable way to detect the end of iteration.

## Solution

Add a sentinel invocation with `pid = -1` after each per-PID loop:

- `seize.c` (`checkpoint_devices`): `run_plugins(CHECKPOINT_DEVICES, -1)`
- `cr-restore.c`: `run_plugins(RESUME_DEVICES_LATE, -1)`

Plugins that don't care return `-ENOTSUP`, which is already ignored, so existing plugins (e.g. amdgpu_plugin) are unaffected. No new hook point and no plugin API version bump.

## Backward compatibility

Behaviour only changes if a plugin explicitly handles `pid == -1`. With no device plugin loaded, both sentinel calls return `-ENOTSUP` and are ignored.

## Testing

- `make criu -j$(nproc)` builds cleanly with `-Werror`
- zdtm basic suite unaffected (sentinel is a no-op without a device plugin)

Design discussed in: https://github.com/checkpoint-restore/criu/discussions/3104

### DIFF
--- a/criu/seize.c
+++ b/criu/seize.c
@@ -1143,6 +1143,14 @@ int checkpoint_devices(void)
 			goto err;
 	}
 
+	/*
+	 * Sentinel: every device PID has been iterated; let plugins run
+	 * any batched work that must happen once per checkpoint.
+	 */
+	ret = run_plugins(CHECKPOINT_DEVICES, -1);
+	if (ret < 0 && ret != -ENOTSUP)
+		goto err;
+
 	exit_code = 0;
 err:
 	return exit_code;
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -2331,6 +2331,19 @@ skip_ns_bouncing:
 			pr_debug("restore late stage hook for external plugin failed\n");
 	}
 
+	/*
+	 * Sentinel: all RESUME_DEVICES_LATE per-PID hooks have completed;
+	 * let plugins run any batched/parallel finalization that must
+	 * happen once, after every device PID has been iterated.
+	 *
+	 * Handle errors as leniently as the per-PID loop above: -ENOTSUP is
+	 * ignored and any other error is logged but tolerated, so we do not
+	 * regress plugins whose hook may return a real error for pid = -1.
+	 */
+	ret = run_plugins(RESUME_DEVICES_LATE, -1);
+	if (ret < 0 && ret != -ENOTSUP)
+		pr_warn("restore late stage sentinel hook failed\n");
+
 	ret = run_scripts(ACT_PRE_RESUME);
 	if (ret)
 		pr_err("Pre-resume script ret code %d\n", ret);


AlanFokCo:
Good catch — fixed in b4a3658. The restore-side sentinel now matches the per-PID loop: -ENOTSUP is ignored and any other error is logged (pr_warn) but tolerated, so plugins like amdgpu_plugin whose kernel-mode pid check may error on pid = -1 won't abort restore. The checkpoint-side sentinel keeps goto err, consistent with the checkpoint per-PID loop.

rst0git:
> Plugins that must act only after all device PIDs have been iterated — for example to run a batched/parallel checkpoint across several GPUs — have no reliable way to detect the end of iteration.

@AlanFokCo Could you please add more information about the use-case (problem you are trying to solve)? What does  "end of iteration" mean in this context?

AlanFokCo:
@rst0git Sure, happy to expand.

## Use case

We maintain a device plugin for checkpoint/restore of multi-GPU LLM inference servers (tensor-parallel jobs that run one process per GPU, e.g. 4 or 8 processes). For each GPU process the plugin has to drive the vendor checkpoint API: lock the process, checkpoint its GPU state, then unlock.

Today `CHECKPOINT_DEVICES` is invoked once per task:

```c
for_each_pstree_item(item)
    run_plugins(CHECKPOINT_DEVICES, item->pid->real);
```

If the plugin does the actual checkpoint work *inside* each per-PID call, the GPUs are necessarily checkpointed one after another, so total time is roughly `N_gpus × per_gpu_time`. For an 8-GPU job that is a large, avoidable serialization — the GPU state saves are independent and could run concurrently.

What we want instead:

1. During the per-PID calls, just record/lock each GPU process.
2. Once every device PID has been handed to the plugin, checkpoint them all in parallel (one child per GPU), then wait for all to finish.

Step 2 can only run *after* the plugin has seen the last PID it will be given. That is the missing signal this RFC adds. The restore side is symmetric: collect the PIDs during `RESUME_DEVICES_LATE`, then resume all GPUs in parallel once the iteration is done.

## What "end of iteration" means

criu walks the process tree with `for_each_pstree_item()` and calls the hook once per task. "End of iteration" is simply the point right after that loop finishes — i.e. the plugin has now been presented with every PID it will get for this phase, and no further per-PID calls are coming.

There is currently no notification for that moment, so a batching plugin has to guess (e.g. track a PID count it obtained by other means, or use a timer), both of which are fragile. The proposal is to make it explicit with a sentinel call after the loop:

```c
run_plugins(CHECKPOINT_DEVICES, -1);   /* per-PID iteration finished */
```

`-1` is never a valid task PID, so a plugin can unambiguously distinguish "here is another PID" from "iteration is done, you may run batched work". Plugins that don't care return `-ENOTSUP`, which is already ignored, so existing plugins are unaffected.

rst0git:
> For each GPU process the plugin has to drive the vendor checkpoint API: lock the process, checkpoint its GPU state, then unlock.

How is this different from the CUDA plugin for CRIU?

> What we want instead: During the per-PID calls, just record/lock each GPU process.

The lock action is called via the `PAUSE_DEVICES` hook and this happens before `CHECKPOINT_DEVICES`. Am I missing something?

AlanFokCo:
@rst0git You're right on both points — sorry for being imprecise.

> How is this different from the CUDA plugin for CRIU?

It's not a different plugin — the flow is exactly the CUDA plugin's flow. The difference is only in *how* `CHECKPOINT_DEVICES` is executed. Today the plugin performs the checkpoint action synchronously inside each per-PID hook call, so when a job has several CUDA processes (e.g. a tensor-parallel inference server with one process per GPU), the device checkpoints run strictly one after another. For large models the checkpoint step (device memory → host) dominates the dump time, so N processes cost roughly N× that.

The sentinel enables a plugin-side optimization: during the per-PID `CHECKPOINT_DEVICES` calls the plugin only records the PID (the process is already locked at that point), and on the `pid = -1` call — when it knows no more PIDs are coming — it checkpoints all recorded processes concurrently and waits for them to finish. We run a variant of the CUDA plugin with this scheme in a production environment; with 4 GPU processes the device-checkpoint phase is roughly 4x faster than the sequential version. We'd be happy to contribute the parallel logic as a follow-up to the plugin — it should apply equally on top of the libcuda-based rewrite in #3076.

> The lock action is called via the `PAUSE_DEVICES` hook and this happens before `CHECKPOINT_DEVICES`. Am I missing something?

No, you're right — locking happens in `PAUSE_DEVICES`, before the tree is frozen, and that part is unchanged in what we do. My earlier "record/lock" wording was sloppy, sorry. To restate precisely: `PAUSE_DEVICES` locks per-PID as today; the per-PID `CHECKPOINT_DEVICES` calls only collect PIDs; the actual checkpoint action is deferred to the sentinel so it can run in parallel. The restore side mirrors this: per-PID `RESUME_DEVICES_LATE` collects, and the sentinel restores/unlocks concurrently.

criu/seize.c
```diff
@@ -1143,6 +1143,14 @@ int checkpoint_devices(void)
 			goto err;
 	}
 
+	/*
+	 * Sentinel: every device PID has been iterated; let plugins run
+	 * any batched work that must happen once per checkpoint.
+	 */
+	ret = run_plugins(CHECKPOINT_DEVICES, -1);
```
avagin, line 1150: Reusing CHECKPOINT_DEVICES this way can break other plugins. I think it is better to introduce a new hook type.

rst0git:
@avagin Fergus (@fergusfinn) also implemented similar optimisation in https://github.com/doublewordai/criu/pull/1 and described the results [here](https://blog.doubleword.ai/fast-sglang-starts). From my experiments, parallel checkpoint/restore is most helpful for multi-GPU workloads where we have multiple CUDA processes, for example multiple NCCL ranks, because CRIU can checkpoint/restore those processes concurrently. Checkpoint/restore support for NCCL workloads was recently enabled through the [NCCL checkpoint shim](https://github.com/NVIDIA/nccl/tree/master/contrib/nccl_checkpoint). However, the integration with CRIU and container runtimes is not very clear.

criu/seize.c
rst0git: As an example, @fergusfinn used `RESUME_DEVICES_LATE_FLUSH` in https://github.com/doublewordai/criu/commit/2f30122dbfa22d32021f2209f8c9ac5458694d80

PR #3107, opened 2026-07-25
Add support for plugin-specific options

rst0git:
This pull request has been rebased on top of https://github.com/checkpoint-restore/criu/pull/3076

Some CRIU plugins need to define additional CLI/RPC options that allow to configure the checkpoint/restore operation. An example of such option for the CUDA plugin is "--device-map". However, plugins are loaded after CRIU's command-line and RPC parser, which prevents them from defining additional options without adding plugin-specific fields in the core protobuf definitions.
    
To support such plugin-specific options, we use the plugin name (e.g. "cuda") as a namespace and expose a plugin options API so each plugin can validate and interpret its own options and their values.

Example:
```bash
i=0
for uuid in $(nvidia-smi --list-gpus | grep -oP 'UUID: \K[^)]+'); do
    export GPU_$i=$uuid
    i=$((i+1))
done

criu restore ... \
    --plugin-option \
   "cuda.device-map=$GPU_0=$GPU_1,$GPU_1=$GPU_0,$GPU_2=$GPU_2,$GPU_3=$GPU_3"
```


criu/plugin-options.c
```diff
@@ -0,0 +1,109 @@
...
+
+void cr_plugin_options_clear(void)
+{
+	struct cr_plugin_option *option, *tmp;
+
+	list_for_each_entry_safe(option, tmp, &opts.plugin_options, node)
+		cr_plugin_option_free(option);
+}
+
+int cr_plugin_option_add(const char *plugin, const char *name, const char *value)
```
avagin, line 27: Do we really want this (plugin, name, value) triplet? Would it be better to build a argv-like string array, allowing plugins to use a standard command-line parsers (e.g., getopt_long) to parse these options and skip unknown ones?

rst0git:
I've added these patches in https://github.com/checkpoint-restore/criu/pull/3076

PR #3109, opened 2026-07-27
fifo: don't reopen a writer-less non-blocking reader as a closed pipe

### LINKED ISSUE #3108: CPU pinned to 100% after restoring LXC container using OpenRC
When performing checkpoint/restore on an LXC container using the lxc-checkpoint command, the cpu will be pinned to 100% in the case of using a container image with OpenRC.

The main culprit is a FIFO used by OpenRC, which is located at `/run/openrc/init.ctl` on an Artix Linux image (artixlinux/artixlinux:base-openrc).

What happens is that after restoring, PID 1 (init) starts using 100% of the cpu. Running `strace -f -p $(lxc-info -pH artix)` shows that epoll is firing constantly on `/run/openrc/init.ctl`.

This issue doesn't happen on gentoo images, since gentoo by default still uses sysvinit, which then invokes OpenRC, as stated [here](https://wiki.gentoo.org/wiki/OpenRC)

Steps to reproduce (assumes all permissions are met or root user):

1. Create the artix container
```bash
docker pull artixlinux/artixlinux:base-openrc

cid=$(docker create artixlinux/artixlinux:base-openrc)

mkdir -p /var/lib/lxc/artix/rootfs
docker export "$cid" | tar -C /var/lib/lxc/artix/rootfs -xf -

docker rm "$cid"

cat > /var/lib/lxc/artix/config <<'EOF'
lxc.uts.name = artix
lxc.rootfs.path = dir:/var/lib/lxc/artix/rootfs
lxc.net.0.type = empty
lxc.init.cmd = /sbin/init
lxc.signal.halt = SIGPWR
EOF
```

2. Start the container
```bash
lxc-start -n artix
```

3. Checkpoint the container
```bash
lxc-checkpoint -n artix -D /tmp/artix -s
```

4. Restore the container
```bash
lxc-checkpoint -n artix -D /tmp/artix -r
```

5. Observe the cpu usage using `top`, inspect the init process of container
```bash
strace -f -p $(lxc-info -pH artix)
```

fyi @mihalicyn (This is regarding the question I had on the lxc/lxc PR)

emirbuljubasic:
`do_open_fifo()` opens a temporary `O_RDWR` fake writer to unblock the restore, then opens the real descriptor while that writer is still present. For an `O_RDONLY|O_NONBLOCK` reader, like OpenRC's `/run/openrc/init.ctl`, this leaves every `poll()` returning `POLLHUP` instead of blocking, so an epoll loop spins at 100% CPU.

As far as I can tell, `fifo_open()` latches `pipe->w_counter` into `filp->f_version` only when such an end is opened with no writer around, and `pipe_poll()` suppresses `POLLHUP` while the two still match. So the fake writer has to be gone before the real descriptor is opened.

Doing that unconditionally is wrong, as @avagin pointed out: an end whose writer has already been opened and closed is a closed pipe and must keep reporting `POLLHUP`. The two states seem to differ nowhere but in `poll()`, so this tells them apart on dump by polling the descriptor drained from the dumpee (it is the dumpee's own `struct file`, so `f_version` is intact) and records the answer in a new `fifo_entry.wait_writer`. Images without the field keep the old behaviour.

Per @rst0git: closing the fake writer drops the last reference to the pipe and takes the just restored pipe size with it, so the pipe parameters are restored on the final descriptor in that case.

The test covers both states. It needs two fifos, since `w_counter` is per inode and a writer opened on one end hangs up every reader of that fifo. I checked that both assertions actually catch something: with the fake writer always kept the test fails on the waiting end, and with it always dropped it fails on @avagin's closed pipe reproducer. `fifo`, `fifo_ro`, `fifo_wronly` and `fifo-ghost` still pass.

### DIFF
--- a/criu/fifo.c
+++ b/criu/fifo.c
@@ -1,4 +1,5 @@
 #include <unistd.h>
+#include <poll.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <fcntl.h>
@@ -41,6 +42,30 @@ static struct pipe_data_dump pd_fifo = {
 	.img_type = CR_FD_FIFO_DATA,
 };
 
+/*
+ * fifo_open() latches pipe->w_counter into filp->f_version for a
+ * non-blocking read-only end opened with no writer around, and
+ * pipe_poll() suppresses POLLHUP while the two still match. An end
+ * that has not seen a writer and one whose writer is gone for good
+ * differ nowhere else, so ask poll().
+ */
+static int fifo_wait_writer(int lfd)
+{
+	struct pollfd pfd = {
+		.fd = lfd,
+		.events = POLLIN,
+	};
+	int ret;
+
+	ret = poll(&pfd, 1, 0);
+	if (ret < 0) {
+		pr_perror("Can't poll fifo");
+		return -1;
+	}
+
+	return !(ret > 0 && (pfd.revents & POLLHUP));
+}
+
 static int dump_one_fifo(int lfd, u32 id, const struct fd_parms *p)
 {
 	struct cr_img *img = img_from_set(glob_imgset, CR_FD_FILES);
@@ -65,6 +90,16 @@ static int dump_one_fifo(int lfd, u32 id, const struct fd_parms *p)
 	e.has_regf_id = true;
 	e.regf_id = rf_id;
 
+	if ((p->flags & O_ACCMODE) == O_RDONLY) {
+		int wait_writer = fifo_wait_writer(lfd);
+
+		if (wait_writer < 0)
+			return -1;
+
+		e.has_wait_writer = true;
+		e.wait_writer = wait_writer;
+	}
+
 	fe.type = FD_TYPES__FIFO;
 	fe.id = e.id;
 	fe.fifo = &e;
@@ -82,37 +117,78 @@ const struct fdtype_ops fifo_dump_ops = {
 
 static struct pipe_data_rst *pd_hash_fifo[PIPE_DATA_HASH_SIZE];
 
+static bool fifo_has_data(u32 id)
+{
+	struct pipe_data_rst *pd;
+
+	for (pd = pd_hash_fifo[id & PIPE_DATA_HASH_MASK]; pd != NULL; pd = pd->next)
+		if (pd->pde->pipe_id == id)
+			return pd->pde->bytes > 0;
+
+	return false;
+}
+
 static int do_open_fifo(int ns_root_fd, struct reg_file_info *rfi, void *arg)
 {
 	struct fifo_info *info = arg;
-	int new_fifo, fake_fifo = -1;
+	int new_fifo = -1, fake_fifo = -1;
+	int flags = rfi->rfe->flags;
+
+	bool read_only = (flags & O_ACCMODE) == O_RDONLY;
+	bool nonblocking = flags & O_NONBLOCK;
+	bool pipe_empty = !fifo_has_data(info->fe->pipe_id);
+	bool wait_writer = info->fe->has_wait_writer && info->fe->wait_writer;
 
 	/*
-	 * The fifos (except read-write fifos) do wait until
-	 * another pipe-end get connected, so to be able to
-	 * proceed the restoration procedure we open a fake
-	 * fifo here.
+	 * An end which had not seen a writer has to be opened with no
+	 * writer around too, or it comes back as a closed pipe and the
+	 * application polling it spins. Such an end never blocks on open,
+	 * so it needs no fake writer at all, and opening one would bump
+	 * w_counter and hang up the ends restored before it. A fifo with
+	 * buffered data is left alone, as the fake writer is the one
+	 * holding the data.
 	 */
-	fake_fifo = openat(ns_root_fd, rfi->path, O_RDWR);
-	if (fake_fifo < 0) {
-		pr_perror("Can't open fake fifo %#x [%s]", info->fe->id, rfi->path);
-		return -1;
+	bool skip_fake_writer = read_only && nonblocking && pipe_empty && wait_writer;
+
+	/*
+	 * FIFOs (except read-write FIFOs) block until the other end is
+	 * opened. Open a temporary read-write descriptor so the restore
+	 * process can proceed.
+	 */
+	if (!skip_fake_writer) {
+		fake_fifo = openat(ns_root_fd, rfi->path, O_RDWR);
+		if (fake_fifo < 0) {
+			pr_perror("Can't open fake fifo %#x [%s]", info->fe->id, rfi->path);
+			return -1;
+		}
+
+		if (info->restore_data) {
+			if (restore_pipe_data(CR_FD_FIFO_DATA, fake_fifo, info->fe->pipe_id, pd_hash_fifo))
+				goto out;
+		}
 	}
 
-	new_fifo = openat(ns_root_fd, rfi->path, rfi->rfe->flags);
+	new_fifo = openat(ns_root_fd, rfi->path, flags);
 	if (new_fifo < 0) {
 		pr_perror("Can't open fifo %#x [%s]", info->fe->id, rfi->path);
 		goto out;
 	}
 
-	if (info->restore_data)
-		if (restore_pipe_data(CR_FD_FIFO_DATA, fake_fifo, info->fe->pipe_id, pd_hash_fifo)) {
+	/*
+	 * With no fake writer around there is nothing else holding the pipe,
+	 * so its parameters are restored on the final descriptor instead.
+	 * The fifo carries no data in this case, so this only restores the
+	 * pipe size, which works on a read-only descriptor too.
+	 */
+	if (info->restore_data && skip_fake_writer) {
+		if (restore_pipe_data(CR_FD_FIFO_DATA, new_fifo, info->fe->pipe_id, pd_hash_fifo)) {
 			close(new_fifo);
 			new_fifo = -1;
 		}
+	}
 
 out:
-	close(fake_fifo);
+	close_safe(&fake_fifo);
 	return new_fifo;
 }
 
--- a/test/zdtm/static/Makefile
+++ b/test/zdtm/static/Makefile
@@ -365,6 +365,7 @@ TST_FILE	=				\
 		fifo				\
 		fifo-ghost			\
 		fifo_ro				\
+		fifo_ro_nonblock		\
 		fifo_wronly			\
 		console				\
 		vt				\
--- a/test/zdtm/static/fifo_ro_nonblock.c
+++ b/test/zdtm/static/fifo_ro_nonblock.c
@@ -1,0 +1,191 @@
+#include <errno.h>
+#include <limits.h>
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <poll.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check that O_RDONLY|O_NONBLOCK fifos keep their POLLHUP state after restore";
+const char *test_author = "Emir Buljubasic <emir.buljubasic@bicomsystems.com>";
+
+char *filename;
+TEST_OPTION(filename, string, "file name", 1);
+
+#define FIFO_SIZE (1 << 20)
+
+static int hup(int fd, int timeout)
+{
+	struct pollfd pfd = { .fd = fd, .events = POLLIN };
+	int ret;
+
+	ret = poll(&pfd, 1, timeout);
+	if (ret < 0) {
+		pr_perror("poll() failed");
+		return -1;
+	}
+
+	return (ret > 0 && (pfd.revents & (POLLHUP | POLLERR)));
+}
+
+static int open_fifo_ro(const char *path)
+{
+	int fd;
+
+	if (mknod(path, S_IFIFO | 0600, 0)) {
+		pr_perror("can't make fifo \"%s\"", path);
+		return -1;
+	}
+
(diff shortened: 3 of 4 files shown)


adrianreber:
The CentOS Stream 10 CI failures are unrelated. Looks like the iptables based network locking is not working. On CentOS Stream 10 we should switch to nftables. Let me do that.

criu/fifo.c
```diff
@@ -82,10 +82,25 @@ const struct fdtype_ops fifo_dump_ops = {
...
+
 static int do_open_fifo(int ns_root_fd, struct reg_file_info *rfi, void *arg)
 {
 	struct fifo_info *info = arg;
-	int new_fifo, fake_fifo = -1;
+	int new_fifo = -1, fake_fifo = -1;
+	int flags = rfi->rfe->flags;
+
+	bool drop_writer_early = (flags & O_ACCMODE) == O_RDONLY && (flags & O_NONBLOCK) &&
+				 !fifo_has_data(info->fe->pipe_id);
```
rst0git: Would it be possible to split this into multiple lines so that it is easier to understand?

emirbuljubasic: of course!

rst0git: Sorry, the comment was meant to improve readability. For example, something like the following:
```C
	bool read_only = (flags & O_ACCMODE) == O_RDONLY;
	bool nonblocking = flags & O_NONBLOCK;
	bool pipe_empty = !fifo_has_data(info->fe->pipe_id);

	/*
	 * Read-only, non-blocking FIFOs with no buffered data do not need
	 * the fake writer once the FIFO contents have been restored.
	 */
	bool drop_writer_early = read_only && nonblocking && pipe_empty;

	/*
	 * FIFOs (except read-write FIFOs) block until the other end is
	 * opened. Open a temporary read-write descriptor so the restore
	 * process can proceed.
	 */
	fake_fifo = openat(ns_root_fd, rfi->path, O_RDWR);
	if (fake_fifo < 0) {
		pr_perror("Can't open fake fifo %#x [%s]", info->fe->id, rfi->path);
		return -1;
	}

	if (info->restore_data && restore_pipe_data(CR_FD_FIFO_DATA, fake_fifo, info->fe->pipe_id, pd_hash_fifo))
		goto out;

	if (drop_writer_early) {
		close(fake_fifo);
		fake_fifo = -1;
	}

	new_fifo = openat(ns_root_fd, rfi->path, flags);
	if (new_fifo < 0) {
		pr_perror("Can't open fifo %#x [%s]", info->fe->id, rfi->path);
		goto out;
	}
```

emirbuljubasic: My bad, didn't quite understand what you meant. Should be good now.

criu/fifo.c
```diff
@@ -82,37 +82,62 @@ const struct fdtype_ops fifo_dump_ops = {
...
-	if (info->restore_data)
-		if (restore_pipe_data(CR_FD_FIFO_DATA, fake_fifo, info->fe->pipe_id, pd_hash_fifo)) {
-			close(new_fifo);
-			new_fifo = -1;
-		}
-
 out:
-	close(fake_fifo);
+	if (fake_fifo >= 0)
+		close(fake_fifo);
```
avagin: nit: please use close_safe(&fake_fifo), it closes fd only if it has been initialized.

avagin:
LGTM. Usually, we put test and criu changes into separate patches.

emirbuljubasic:
> LGTM. Usually, we put test and criu changes into separate patches.

I split it into two commits.

criu/fifo.c
```diff
@@ -82,37 +82,61 @@ const struct fdtype_ops fifo_dump_ops = {
...
+	 * process can proceed.
 	 */
 	fake_fifo = openat(ns_root_fd, rfi->path, O_RDWR);
 	if (fake_fifo < 0) {
 		pr_perror("Can't open fake fifo %#x [%s]", info->fe->id, rfi->path);
 		return -1;
 	}
 
-	new_fifo = openat(ns_root_fd, rfi->path, rfi->rfe->flags);
+	if (info->restore_data)
```
rst0git: It looks like this drops the restored pipe size - `dump_one_pipe_data()` always writes a `PipeDataEntry` with `has_size = true` for a fifo dumped through a readable fd and `restore_pipe_data()` applies it to `fake_fifo` via `F_SETPIPE_SZ`. In the `drop_writer_early` path the fake writer is the only descriptor of the fifo, so closing it destroys the pipe object together with the just-restored size, and the reopened fifo comes back with the default 64K.

```
	if (info->restore_data && !drop_writer_early)
		if (restore_pipe_data(CR_FD_FIFO_DATA, fake_fifo, info->fe->pipe_id, pd_hash_fifo))
			goto out;

	/*
	 * Closing the last descriptor destroys the pipe object together
	 * with its restored parameters, so when the writer is dropped
	 * early the pipe parameters are restored on the final descriptor
	 * below instead.
	 */
	if (drop_writer_early) {
		close(fake_fifo);
		fake_fifo = -1;
	}

	new_fifo = openat(ns_root_fd, rfi->path, flags);
	if (new_fifo < 0) {
		pr_perror("Can't open fifo %#x [%s]", info->fe->id, rfi->path);
		goto out;
	}

	/*
	 * The FIFO carries no data here, so this only restores the pipe
	 * size, which works on a read-only descriptor too.
	 */
	if (info->restore_data && drop_writer_early)
		if (restore_pipe_data(CR_FD_FIFO_DATA, new_fifo, info->fe->pipe_id, pd_hash_fifo)) {
			close(new_fifo);
			new_fifo = -1;
		}
```

test/zdtm/static/fifo_ro_nonblock.c
```diff
@@ -0,0 +1,89 @@
...
+	 * instead of spinning.
+	 */
+	fd = open(filename, O_RDONLY | O_NONBLOCK);
+	if (fd < 0) {
+		pr_perror("can't open %s", filename);
+		return 1;
+	}
+
+	/* Sanity: no POLLHUP before C/R. */
+	if (hup(fd, 0)) {
```
rst0git: `hup()` returns -1 when `poll()` fails:
```
	switch (hup(fd, 0)) {
	case -1:
		pr_err("poll() failed before C/R\n");
		return 1;
	case 1:
		pr_err("fifo reports POLLHUP before C/R\n");
		return 1;
	}
```

test/zdtm/static/fifo_ro_nonblock.c
```diff
@@ -0,0 +1,89 @@
...
+#include <fcntl.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "Check an O_RDONLY|O_NONBLOCK fifo with no writer does not "
+		       "report POLLHUP (busy-loop) after restore";
+const char *test_author = "Emir Buljubasic <emir.buljubasic@bicomsystems.com>";
+
+char *filename;
+TEST_OPTION(filename, string, "file name", 1);
```
rst0git, line 16: Please extend the test to also cover the pipe size with something like the following:
```
#define FIFO_SIZE (1 << 20)

...
	fd = open(filename, O_RDONLY | O_NONBLOCK);
	if (fd < 0) {
		pr_perror("can't open %s", filename);
		return 1;
	}

	if (fcntl(fd, F_SETPIPE_SZ, FIFO_SIZE) < 0) {
		pr_perror("can't set pipe size on %s", filename);
		return 1;
	}

...

	switch (hup(fd, 100)) {
	case -1:
		fail("poll() failed after restore");
		return 1;
	case 1:
		fail("fifo reports POLLHUP after restore -- would busy-loop");
		return 1;
	}

	/*
	 * The pipe size must survive the restore even though CRIU reopens
	 * the reader against a writer-less (freshly created) pipe object.
	 */
	if (fcntl(fd, F_GETPIPE_SZ) != FIFO_SIZE) {
		fail("fifo lost its pipe size after restore");
		return 1;
	}
```

test/zdtm/static/Makefile
```diff
@@ -365,6 +365,7 @@ TST_FILE	=				\
 		fifo				\
 		fifo-ghost			\
 		fifo_ro				\
+		fifo_ro_nonblock \
```
rst0git: Could align this with tabs:
```suggestion
		fifo_ro_nonblock	\
```

avagin:
Here is one more issue introduced by this patch. We need to distinguish two cases:
* a read-only descriptor created before the write one (waiting for the write one).
* the write end has been closed and the read-only descriptor basically a closed pipe.

Here is a reproducer for the second case:
```
diff --git a/test/zdtm/static/fifo_ro_nonblock.c b/test/zdtm/static/fifo_ro_nonblock.c
index ab0a5e902faa..711421b58187 100644
--- a/test/zdtm/static/fifo_ro_nonblock.c
+++ b/test/zdtm/static/fifo_ro_nonblock.c
@@ -30,7 +30,8 @@ static int hup(int fd, int timeout)
 
 {
-       int fd;
+       int fd, fdw;
+       char buf[1024];
 
        test_init(argc, argv);
 
@@ -57,9 +58,16 @@ int main(int argc, char **argv)
                return 1;
        }
 
+       fdw = open(filename, O_WRONLY);
+       close(fdw);
+
        test_daemon();
        test_waitsig();
 
+       if (read(fd, buf, 1) != 0) {
+               fail("fifo hasn't been closed");
+               return 1;
+       }
        /*
         * After restore poll() must still not report POLLHUP. If it does,
         * CRIU reopened the reader while a (fake) writer was present, and the
@@ -69,9 +77,12 @@ int main(int argc, char **argv)
        case -1:
                fail("poll() failed after restore");
                return 1;
-       case 1:
-               fail("fifo reports POLLHUP after restore -- would busy-loop");
+       case 0:
+               fail("no hup");
                return 1;
+       case 1:
+               //fail("fifo reports POLLHUP after restore -- would busy-loop");
+               break;
        }
 
        if (close(fd) < 0) {
```

avagin:
@emirbuljubasic any update?

emirbuljubasic:
> @emirbuljubasic any update?

Sorry for the delay, I'm on vacation. Will be back on this on monday.

criu/fifo.c
emirbuljubasic: Thanks, fixed.

test/zdtm/static/fifo_ro_nonblock.c
emirbuljubasic: Thanks, added.

emirbuljubasic:
> Here is one more issue introduced by this patch. We need to distinguish two cases:
> 
>     * a read-only descriptor created before the write one (waiting for the write one).
> 
>     * the write end has been closed and the read-only descriptor basically a closed pipe.
> 
> 
> Here is a reproducer for the second case:
> 
> ```
> diff --git a/test/zdtm/static/fifo_ro_nonblock.c b/test/zdtm/static/fifo_ro_nonblock.c
> index ab0a5e902faa..711421b58187 100644
> --- a/test/zdtm/static/fifo_ro_nonblock.c
> +++ b/test/zdtm/static/fifo_ro_nonblock.c
> @@ -30,7 +30,8 @@ static int hup(int fd, int timeout)
>  
>  int main(int argc, char **argv)
>  {
> -       int fd;
> +       int fd, fdw;
> +       char buf[1024];
>  
>         test_init(argc, argv);
>  
> @@ -57,9 +58,16 @@ int main(int argc, char **argv)
>                 return 1;
>         }
>  
> +       fdw = open(filename, O_WRONLY);
> +       close(fdw);
> +
>         test_daemon();
>         test_waitsig();
>  
> +       if (read(fd, buf, 1) != 0) {
> +               fail("fifo hasn't been closed");
> +               return 1;
> +       }
>         /*
>          * After restore poll() must still not report POLLHUP. If it does,
>          * CRIU reopened the reader while a (fake) writer was present, and the
> @@ -69,9 +77,12 @@ int main(int argc, char **argv)
>         case -1:
>                 fail("poll() failed after restore");
>                 return 1;
> -       case 1:
> -               fail("fifo reports POLLHUP after restore -- would busy-loop");
> +       case 0:
> +               fail("no hup");
>                 return 1;
> +       case 1:
> +               //fail("fifo reports POLLHUP after restore -- would busy-loop");
> +               break;
>         }
>  
>         if (close(fd) < 0) {
> ```

@avagin Thanks, I completely missed that case. I've tried to handle it by telling the two apart on dump. The only difference I could find is in poll(), so I poll the descriptor drained from the dumpee (its own struct file, so f_version is still intact) and store the answer in a new `fifo_entry.wait_writer`. Restore only drops the fake writer when it's set, and images without  he field keep the old behaviour.

I wasn't sure about adding a field to the image just for this, so if there's a better place to get this from, or you'd prefer a different approach entirely, I'm happy to redo it.

Your reproducer is in the test now, as a second fifo, since w_counter is per inode and the writer would otherwise hang up the first reader too. One thing I noticed while writing it: read() seems to return 0 in both cases, so I used the poll() check as the actual assertion. Let me know if I've got that wrong.

Thank you everybody for being patient with me and for the informative comments. This is very much outside of my comfort zone.

criu/fifo.c
```diff
@@ -82,37 +117,78 @@ const struct fdtype_ops fifo_dump_ops = {
...
 	/*
-	 * The fifos (except read-write fifos) do wait until
-	 * another pipe-end get connected, so to be able to
-	 * proceed the restoration procedure we open a fake
-	 * fifo here.
+	 * FIFOs (except read-write FIFOs) block until the other end is
+	 * opened. Open a temporary read-write descriptor so the restore
+	 * process can proceed.
 	 */
 	fake_fifo = openat(ns_root_fd, rfi->path, O_RDWR);
```
avagin: why do we need to make fake_fifo if drop_writer_early is true? We can have more that one read-only file descriptors waiting for a writer.

PR #3110, opened 2026-07-27
ci: switch centos stream 10 to nftables

adrianreber:
CentOS Stream 10 defaults to nftables. Also use that in CI to avoid errors for unsupported configurations.


rst0git (approved):
LGTM

PR #3113, opened 2026-07-29
criu: restore into a caller-provided user namespace

arcceus:
This adds restore support for joining a caller-provided user namespace.

The restore path can now use `--join-ns user:<path>` when the user namespace is created and owned by the caller. This lets CRIU restore tasks into an existing user namespace instead of always creating the namespace itself.

Changes:
- support joining a user namespace during restore without moving the main CRIU process into it
- use a short-lived helper to clone the restored root inside the supplied user namespace
- skip user namespace map setup when restoring into a joined user namespace
- route image opens through usernsd where needed
- add diagnostics for joined-userns paths
- add `zdtm/static/rootless_runtime_userns` coverage
- add libcriu coverage for invalid userns join options

### DIFF
--- a/criu/cr-restore.c
+++ b/criu/cr-restore.c
@@ -1037,7 +1037,7 @@ static void maybe_clone_parent(struct pstree_item *item, struct cr_clone_arg *ca
 	 * off of 3.11, this condition can be simplified to just test the
 	 * options and not have the pdeath_sig test.
 	 */
-	if (opts.restore_sibling) {
+	if (opts.restore_sibling && !userns_join_ns_requested()) {
 		/*
 		 * This means we're called from lib's criu_restore_child().
 		 * In that case create the root task as the child one to+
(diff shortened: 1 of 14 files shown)


criu/pagemap.c
```diff
@@ -2799,8 +2819,19 @@ static int try_open_parent(int dfd, unsigned long id, struct page_read *pr, int
...
 		goto out;
 
-	if (open_parent(dfd, &pfd))
+	ret = open_pagemap_parent(dfd, &pfd);
+	if (ret < 0) {
+		int err = -ret;
+
+		if (opts.mode == CR_RESTORE && userns_join_ns_requested() &&
+		    (err == ENOENT || err == EPERM || err == EACCES)) {
+			pr_warn("Can't open parent page images in joined userns, trying current images only\n");
```
avagin: Parent images are not optional. If a parent image exists, CRIU has to use it.

criu/cr-restore.c
```diff
@@ -1190,6 +1193,23 @@ static inline int fork_with_pid(struct pstree_item *item)
 
 	pr_info("Forking task with %d pid (flags 0x%lx)\n", pid, ca.clone_flags);
 
+	if (!item->parent && userns_join_ns_requested()) {
+		if (join_user_namespace()) {
```
avagin: This call moves the main CRIU process into the target user namespace. I don't think we want to do that. Why can't an external user namespace be handled just like one restored by CRIU?

arcceus: I thought moving the main CRIU process into the target user namespace was the simplest way to handle an existing nested user namespace, since the restored root could inherit it directly. Reworking on this.

PR #3115, opened 2026-08-05
page-xfer: send compressed page-server data over TLS

rst0git:
When compression is used with the page-server, the sizes and payloads would bypass the TLS-aware transport and write directly to the socket. As a result, when TLS is used this mixes plaintext into the encrypted stream, so CRIU has to reject the combination.

To fix this, we add helpers that send and receive complete buffers through the TLS wrapper. These helpers are used for page-server commands, sizes, payloads, responses, and final status. Initialize client TLS for both newly connected and inherited page-server sockets.

### DIFF
--- a/criu/page-xfer.c
+++ b/criu/page-xfer.c
@@ -148,7 +148,36 @@ static inline int __recv(int sk, void *buf, size_t sz, int fl)
 	return opts.tls ? tls_recv(buf, sz, fl) : recv(sk, buf, sz, fl);
 }
 
-static int recv_full(int sk, void *buf, size_t size, const char *what)
+static int send_full_flags(int sk, const void *buf, size_t size, int flags,
+			   const char *what)
+{
+	size_t done = 0;
+
+	while (done < size) {
+		ssize_t ret = __send(sk, (const char *)buf + done, size - done, flags);
+
+		if (ret < 0) {
+			if (errno == EINTR)
+				continue;
+			pr_perror("Can't send %s", what);
+			return -1;
+		}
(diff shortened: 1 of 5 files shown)


criu/page-xfer.c
```diff
@@ -1876,12 +1885,7 @@ static int page_server_serve(int sk)
 		struct page_server_iov pi;
 		u32 cmd;
 
-		ret = __recv(sk, &pi, sizeof(pi), MSG_WAITALL);
-		if (!ret)
```
avagin, line 1880: recv_full fails if other side has closed the socket. Before this changes, this code exits with the success code in this case...

rst0git: Thanks Andrei! I've updated this patch and rebased the branch on criu-dev.

criu/page-xfer.c
```diff
@@ -148,7 +148,36 @@ static inline int __recv(int sk, void *buf, size_t sz, int fl)
...
+	}
+	return 0;
+}
+
+static int send_full(int sk, const void *buf, size_t size, const char *what)
+{
+	return send_full_flags(sk, buf, size, 0, what);
+}
+
+/* Return 1 for EOF before the frame, 0 for a complete frame, or -1 on error. */
```
avagin: nit: I would prefer to refactor the return values for consistency: return 0 to indicate EOF, the number of bytes read for success, and -1 on error.

rst0git: Thanks!

avagin:
LGTM

PR #3117, opened 2026-08-12
dump: skip POSIX timers targeting dead threads

avagin:
@pojntfx @aidancully please review this pr. It is #2996 with a few minor modifications.

### DIFF
--- a/criu/timer.c
+++ b/criu/timer.c
@@ -9,6 +9,10 @@
 #include "rst-malloc.h"
 #include "restorer.h"
 
+#ifndef SIGEV_NONE
+#define SIGEV_NONE 1 /* other notification: meaningless */
+#endif
+
 static inline int timeval_valid(struct timeval *tv)
 {
 	return (tv->tv_sec >= 0) && ((unsigned long)tv->tv_usec < USEC_PER_SEC);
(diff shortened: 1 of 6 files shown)


criu/timer.c
```diff
@@ -298,12 +302,22 @@ static int core_alloc_posix_timers(TaskTimersEntry *tte, int n, PosixTimerEntry
...
 {
-	pid_t vtid = 0;
+	pid_t vtid;
 	int i;
 
 	if (rtid == 0)
 		return 0;
 
+	for (i = 0; i < item->nr_threads; i++) {
+		if (item->threads[i].real == rtid)
```
rst0git, line 312: @avagin I think TID can be reused after the original thread exits, while `/proc/<pid>/timers` still reports that TID from the old struct pid. For example:
1. Thread A (TID 42) creates a timer, then exits.
2. Thread B later reuses TID 42.
3. The kernel timer still references dead thread A, but `/proc/<pid>/timers` reports 42.
4. CRIU sees live thread B with TID 42 and assumes the timer targets it.
5. After restore, the timer incorrectly delivers signals to B instead of dropping them.

avagin: You are right, but this is a separate problem that can't be solved in user-space. The kernel patch is here https://lore.kernel.org/all/20260816161216.984580-1-avagin@google.com/.

PR #3118, opened 2026-08-14
cgroup: check buffer bounds when constructing cgroup paths

rst0git:
`prepare_cgroup_dir_properties()` and `prepare_cgroup_dirs()` append nested cgroup directory names to a fixed-size path buffer with `sprintf()`, without checking the remaining space. A sufficiently deep or long cgroup hierarchy can write past the end of the buffer. To fix this, introduce a helper, `append_cgroup_dir()`, that uses `snprintf()` and verifies the result fits in the buffer, and pass the buffer size down through the recursive calls. Restore now fails with an error message when a cgroup path does not fit.

### DIFF
--- a/criu/cgroup.c
+++ b/criu/cgroup.c
@@ -1452,12 +1452,37 @@ static int restore_cgroup_subtree_control(const CgroupPropEntry *cg_prop_entry_p
 	return 0;
 }
 
+/* Append "/dir_name" to path at offset off, checking that it fits in a buffer
+ * of path_size bytes. On success, if new_off is not NULL, it is set to the
+ * offset of the resulting string's terminating null byte.
+ */
+static int append_cgroup_dir(char *path, size_t path_size, size_t off, const char *dir_name, size_t *new_off)
+{
+	int ret;
+
+	if (off >= path_size) {
+		pr_err("Cgroup path '%.*s' is too long\n", path_size > 0 ? (int)path_size - 1 : 0, path);
+		return -1;
+	}
+
+	ret = snprintf(path + off, path_size - off, "/%s", dir_name);
+	if (ret < 0 || (size_t)ret >= path_size - off) {
+		pr_err("Cgroup path %.*s/%s is too long\n", (int)off, path, dir_name);
+		return -1;
+	}
+
+	if (new_off)
+		*new_off = off + ret;
+	return 0;
+}
+
 /*
  * Note: The path string can be modified in this function,
- * the length of path string should be at least PATH_MAX.
+ * the buffer holding it must be at least path_size bytes.
  */
-static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *path, int off, bool split_lines,
-			       bool skip_fails)
+static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p,
+			       char *path, size_t path_size, size_t off,
+			       bool split_lines, bool skip_fails)
 {
 	int cg, fd, exit_code = -1, flag;
 	CgroupPerms *perms = cg_prop_entry_p->perms;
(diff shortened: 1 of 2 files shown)


criu/cgroup.c
```diff
@@ -1963,7 +1984,8 @@ static int prepare_cgroup_sfd(CgroupEntry *ce)
 	paux[off++] = '/';
 
 	for (i = 0; i < ce->n_controllers; i++) {
-		int ctl_off = off, yard_off;
+		int ctl_off = off;
+		size_t yard_off, yard_size;
```
adrianreber: Any specific reason why you are changing yard_off to an unsigned type? I think it would be good to keep it signed and check for a negative value?

adrianreber:
Do you the warning about hiding a global value? Not super important but probably good to not introduce new codeql warnings.

criu/cgroup.c
rst0git: Thanks! It looks like here `strlen()` returns `size_t`, so in `ctl_off - (strlen(cg_yard) + 1)` the `int` gets promoted to `size_t` and the subtraction is unsigned. I've updated the code to check for negative value.

adrianreber:
Lot's of different unusual CI errors. Some of them seem to be related to Fedora branching Fedora 45 but there are a lot of other errors also. Not sure where they are coming from.

adrianreber:
Looks like the apparmor_stacking error could be related to a new kernel: 7.0.0-1009-azure

adrianreber:
The new kernel actually is: 7.0.0-1011-azure

1009 is the one that works

criu/cgroup.c
```diff
@@ -1452,12 +1452,36 @@ static int restore_cgroup_subtree_control(const CgroupPropEntry *cg_prop_entry_p
...
+
 /*
  * Note: The path string can be modified in this function,
- * the length of path string should be at least PATH_MAX.
+ * the buffer holding it must be at least path_size bytes.
  */
-static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *path, int off, bool split_lines,
-			       bool skip_fails)
+static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p, char *path, size_t path_size, size_t off,
+			       bool split_lines, bool skip_fails)
```
avagin, line 1485: nit:
```
static int restore_cgroup_prop(const CgroupPropEntry *cg_prop_entry_p,
                               char *path, size_t path_size, size_t off,
                               bool split_lines, bool skip_fails)
```

criu/cgroup.c
```diff
@@ -1452,12 +1452,36 @@ static int restore_cgroup_subtree_control(const CgroupPropEntry *cg_prop_entry_p
...
+/* Append "/dir_name" to path at offset off, checking that it fits in a buffer
+ * of path_size bytes. On success, if new_off is not NULL, it is set to the
+ * offset of the resulting string's terminating null byte.
+ */
+static int append_cgroup_dir(char *path, size_t path_size, size_t off, const char *dir_name, size_t *new_off)
+{
+	int ret;
+
+	if (off >= path_size) {
+		pr_err("Cgroup path %s is too long\n", path);
```
avagin: sounds like a corrupted state? Should we do something like:
```
pr_err("Cgroup path '%.*s' is too long\n", path_size > 0 ? path_size - 1 : 0, path);
```

avagin:
please fix CI failures:
```
#12 7.969 In file included from criu/include/criu-log.h:23,
#12 7.969                  from criu/include/servicefd.h:10,
#12 7.969                  from criu/include/image.h:7,
#12 7.969                  from criu/include/cr_options.h:9,
#12 7.969                  from criu/compression.c:15:
#12 7.969 criu/compression.c: In function 'encoded_stream_read_batch':
#12 7.969 criu/include/log.h:49:52: error: format '%zu' expects argument of type 'size_t', but argument 5 has type 'long unsigned int' [-Werror=format=]
#12 7.969    49 | #define pr_err(fmt, ...) print_on_level(LOG_ERROR, "Error (%s:%d): " LOG_PREFIX fmt, __FILE__, __LINE__, ##__VA_ARGS__)
#12 7.969       |                                                    ^~~~~~~~~~~~~~~~~
#12 7.969 criu/compression.c:1533:17: note: in expansion of macro 'pr_err'
#12 7.969  1533 |                 pr_err("Unable to decompress streaming batch at page %zu\n",
#12 7.969       |                 ^~~~~~
```

Other than that, LGTM.

PR #3119, opened 2026-08-20
criu: add an external-memory provider for dump and restore

dfeigin-nv:
This adds a generic external-memory provider interface for dump and restore.

CRIU uses an inherited `extmem-provider` Unix `SOCK_SEQPACKET` FD. The
provider receives relative image names and object identities, and returns one
FD with `SCM_RIGHTS`.

On restore, a provider may supply:

- readable FDs for selected image files;
- backing FDs for selected private VMAs, including private hugetlb;
- backing FDs for selected shared-memory and memfd objects.

CRIU asks the provider while premapping each private VMA. `-ENOTSUP` is a
per-request fallback: CRIU leaves that image or memory object unchanged and
uses its normal restore path. Other provider errors fail the restore.

On dump, a provider may supply writable FDs for normal output images.
`inventory.img` and pre-dump images stay local. CRIU sends `COMMIT` after a
successful dump or restore, and `ABORT` on failure.

CRIU waits for provider readiness before applying saved memfd seals and before
the pre-resume hook.

Tests cover the request operations, image and private-VMA fallback, provider
failures, readiness, dump output, commit/abort, and a focused static restore
with a mock provider.

## 120B restore measurements

Workload: GPT-OSS-120B  
Checkpoint data: ~140 GB

“Provider I/O” is the time spent reading checkpoint data and preparing memfds before CRIU starts.
Will add a few more benchmarks in place. 

| Restore source / mode | Provider I/O | CRIU core | CUDA | Detection → complete |
|---|---:|---:|---:|---:|
| CRIU-dev, tmpfs | — | 11.468 s | 11.779 s | 23.247 s |
| CRIU-dev, one local NVMe | — | 21.802 s | 12.414 s | 34.332 s |
| Provider, prefetched memfds, one NVMe | 19.941 s | 1.466 s | 11.149 s | 33.480 s |
| Provider, prefetched memfds, four NVMe | 6.694 s | 1.312 s | 10.744 s | 19.757 s |

## S3 restore measurements

Workload: Qwen3-14B  
Checkpoint data: 45.7 GiB

| Restore path | Downloader | Download | CRIU | Download + CRIU |
|---|---|---:|---:|---:|
| S3 directly into provider memfds | Run:ai Model Streamer | 5.478 s | 0.566 s | 6.056 s |
| S3 to tmpfs, then normal CRIU | s5cmd | 6.300 s | 7.130 s | 13.430 s |

Each row is the median of five successful restores.

More information can be found in:
https://docs.google.com/document/d/1i6xlonOpnY-HkUPmsLnuXzB0RqsiM9_YjwFVS-mVQ1M/edit?tab=t.0#heading=h.qkpu5w2ghwv9


test/zdtm/static/extmem_recheckpoint.c
```diff
@@ -0,0 +1,70 @@
+#include <linux/memfd.h>
+#include <stdint.h>
+#include <string.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "zdtmtst.h"
+
+const char *test_doc = "External-memory restore can be recheckpointed without a provider";
```
rst0git, line 10: Would it be possible to move the tests in a separate commit?

Documentation/under-the-hood/index.md
```diff
@@ -23,6 +23,7 @@ This directory contains technical documentation detailing the internal implement
 * [Memory Images Deduplication](memory-images-deduplication.md): Saving space in image files.
 * [Optimizing Pre-dump Algorithm](optimizing-pre-dump-algorithm.md): Strategies for minimizing downtime.
 * [Userfaultfd](userfaultfd.md): Lazy migration and post-copy restoration.
+* [External Memory Provider](external-memory-provider.md): Restore memory from an inherited provider socket.
```
rst0git, line 26: It would be good for the docs also to be in a separate commit.

avagin:
please format commits properly.

criu/mem.c
```diff
@@ -1649,8 +1736,44 @@ int open_vmas(struct pstree_item *t)
...
+
+		list_for_each_entry(vma, &vmas->h, list) {
+			if (!vma_area_is(vma, VMA_AREA_REGULAR) || !vma_area_is(vma, VMA_ANON_PRIVATE) ||
+			    vma_area_is(vma, VMA_AREA_STACK) || vma_area_is(vma, VMA_AREA_VSYSCALL) ||
+			    vma_area_is(vma, VMA_AREA_VDSO) || vma_area_is(vma, VMA_AREA_VVAR) ||
+			    (vma->e->flags & MAP_GROWSDOWN) ||
+			    (vma->e->madv & (1ul << MADV_WIPEONFORK)))
+				continue;
+
+			vma->e->flags &= ~MAP_ANONYMOUS;
```
avagin, line 1799: Why do we need to do that here? What if a memory provider restores only a few specified VMAs or only shared VMAs?

criu/shmem.c
```diff
@@ -606,7 +609,23 @@ static int open_shmem(int pid, struct vma_area *vma)
 		memfd_flag |= MFD_HUGETLB | size_flag;
 	}
 
-	if (kdat.has_memfd && (!is_hugetlb || kdat.has_memfd_hugetlb)) {
+	provider_ret = extmem_init();
+	if (provider_ret == 0) {
+		provider_ret = extmem_get_shared(si->shmid, si->size, &f);
```
avagin: why extmem_get_shared can't call extmem_init and extmem_validate_memfd_mapping_fd?

criu/extmem.c
```diff
@@ -0,0 +1,378 @@
...
+	if (ret)
+		return ret;
+	shared.shmid = shmid;
+	shared.length = length;
+	req.op = EXTMEM_OP__EXTMEM_GET_SHARED;
+	req.get_shared = &shared;
+	return provider_request(&req, fd);
+}
+
+int extmem_validate_mapping_fd(int fd, unsigned long length)
```
avagin: why do we need validating fd? In real life, a mapping can be wider than its backing store.

avagin:
I think the first patch should be split into three patches: handling images, handling memfd/shmem files, and handling anon VMAs.

rst0git:
@dfeigin-nv Adding a few notes from the meeting earlier today: it would be good to have a reference implementation of an external-memory provider. For example, this could follow the precedent of [criu-image-streamer](https://github.com/checkpoint-restore/criu-image-streamer) - a separate repository with a more permissive license, independent code reviews, and its own release cycle; or use tighter integration, similar to the `page-server` command, allowing us to test and extend the functionality directly in CRIU.

test/zdtm/extmem_provider.py
```diff
@@ -0,0 +1,181 @@
...
+    def start(self, image_dir):
+        client, server = socket.socketpair(socket.AF_UNIX, socket.SOCK_SEQPACKET)
+        fdflags = fcntl.fcntl(client.fileno(), fcntl.F_GETFD)
+        fcntl.fcntl(client.fileno(), fcntl.F_SETFD, fdflags & ~fcntl.FD_CLOEXEC)
+        self.pid = os.fork()
+        if self.pid == 0:
+            client.close()
+            try:
+                self._serve(server, image_dir)
+            except BaseException as error:
```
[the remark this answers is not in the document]
dfeigin-nv: Already fixed in ca714a054: the child provider process catches Exception rather than BaseException.

criu/extmem.c
```diff
@@ -0,0 +1,377 @@
...
+	response.msg.msg_controllen = sizeof(response.control);
+	socket_fd = provider_socket();
+	if (socket_fd < 0)
+		return -1;
+
+	if (provider_send_recv(req->op, socket_fd, &request, packed, &response))
+		goto out;
+	ret = decode_provider_response(req->op, &response, fd);
+out:
+	if (socket_fd >= 0 && socket_fd != provider_thread_fd)
```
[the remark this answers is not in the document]
dfeigin-nv: The cleanup is intentional: provider_socket() returns the thread-local FD when it was acquired, and the inherited FD otherwise. Rewrote the condition to check provider_thread_fd < 0 directly, which makes the ownership rule explicit and avoids the CodeQL warning.

test/zdtm/static/extmem_recheckpoint.c
dfeigin-nv: The tests are separated in ca714a054 (test: add external memory provider coverage).

Documentation/under-the-hood/index.md
dfeigin-nv: The documentation is separated in 3ebcde61d (docs: document external memory provider).

criu/mem.c
dfeigin-nv: Yeah, that’s fair. A provider may only support some VMAs, or only shared memory. We’ll treat -ENOTSUP from GET_VMA as “this VMA isn’t handled” and let CRIU restore it normally. Other errors will still fail the restore.

criu/shmem.c
dfeigin-nv: Good point. I’ll make `extmem_get_shared()` handle all that.

criu/extmem.c
dfeigin-nv: Hmm, I think you're right. I was trying to be defensive but that's shouldn't be the case here. A mapping can be wider than its backing file, so I'll remove the FD validation from the provider paths.

dfeigin-nv:
> I think the first patch should be split into three patches: handling images, handling memfd/shmem files, and handling anon VMAs.
@avagin - Makes sense. I’ll split it.

dfeigin-nv:
> @dfeigin-nv Adding a few notes from the meeting earlier today: it would be good to have a reference implementation of an external-memory provider. For example, this could follow the precedent of [criu-image-streamer](https://github.com/checkpoint-restore/criu-image-streamer) - a separate repository with a more permissive license, independent code reviews, and its own release cycle; or use tighter integration, similar to the page-server command, allowing us to test and extend the functionality directly in CRIU.

@rst0git as discussed I'll open it in https://github.com/ai-dynamo/snapshot in a subdirectory so that I can start using it and you can take a look at it and see how it can look like. If you will deem it appropriate we can move it to its own repo under  checkpoint-restore. 
I'll post a comment here with a link to the directory when it is ready.

dfeigin-nv:
> I think the first patch should be split into three patches: handling images, handling memfd/shmem files, and handling anon VMAs.

You're right. I think a fourth patch for the protocol itself is also good to make it more readable. I'll restructure it.

dfeigin-nv:
@rst0git, @avagin — following up on the reference-provider discussion with @rst0git:

The plan was to implement a provider as part of NVIDIA Snapshot. The benchmarks in this PR use this API for both S3 ↔ memfd ↔ CRIU and sharded disks ↔ memfd ↔ CRIU. Most of that logic can live in a library; a provider then only needs to implement its API.

I made a reference implementation with no Snapshot- or NVIDIA-specific code:

https://github.com/ai-dynamo/snapshot/tree/criu-memory-provider-v1-clean/agent/criu-memory-provider

It is a C++ library with a C API. On restore, it turns ready checkpoint bytes into sparse FDs and serves the provider socket. On dump, it supplies writable image FDs and handles COMMIT/ABORT.

It is not merged and is not final. If this direction is useful, I can move it to a separately reviewed checkpoint-restore repository and adjust the API and implementation as needed.

criu/mem.c
```diff
@@ -1649,8 +1765,43 @@ int open_vmas(struct pstree_item *t)
...
+			vma->e->flags |= MAP_PRIVATE;
+			vma->e->status |= VMA_EXT_PROVIDER | VMA_CLOSE;
```
avagin: > Fall back per VMA on ENOTSUP, preserving normal CRIU handling for unsupported regions.

I don't understand why we need to enumerate all vma-s and shuffle flags. If we are changing existing flags, it can affect the normal CRIU handling. Why do we need setting VMA_EXT_PROVIDER is unclear too.

criu/mem.c
```diff
@@ -1128,8 +1192,27 @@ static int premap_priv_vmas(struct pstree_item *t, struct vm_area_list *vmas, vo
 
 		if (!vma_area_is_private(vma, kdat.task_size))
 			continue;
+		if ((vma->e->status & VMA_EXT_PROVIDER) && !(vma->e->flags & MAP_HUGETLB)) {
```
avagin: One of the AI findings:
```
  1. In mem.c:1195, private Hugetlb VMAs are explicitly excluded from premapping:                                                                                                                                                                                                      
    if ((vma->e->status & VMA_EXT_PROVIDER) && !(vma->e->flags & MAP_HUGETLB))                                                                                                                                                                                                         
  Thus, Hugetlb VMAs retain VMA_EXT_PROVIDER in vma->e->status without querying the provider yet.                                                                                                                                                                                      
  2. Immediately afterwards, mem.c:1393 reads the page images:                                                                                                                                                                                                                         
    if (vma->e->status & VMA_EXT_PROVIDER) {                                                                                                                                                                                                                                           
        pr->skip_pages(pr, len);                                                                                                                                                                                                                                                       
        ...                                                                                                                                                                                                                                                                            
        continue;                                                                                                                                                                                                                                                                      
    }                                                                                                                                                                                                                                                                                  
  Because VMA_EXT_PROVIDER is set, all page contents from the checkpoint image for the Hugetlb VMA are skipped.                                                                                                                                                                        
  3. Much later during task restoration, mem.c:1732 calls mem.c:1080 to request the Hugetlb VMA from the provider. If the provider returns -ENOTSUP, mem.c:1050 clears VMA_EXT_PROVIDER and marks it MAP_ANONYMOUS.                                                                    
  4. However, restore_priv_vma_content() has already executed and discarded the page stream. There is no second pass to read the image pages. The Hugetlb VMA is restored completely empty and zero-filled, losing all process state silently.   
```

criu/image.c
```diff
@@ -719,11 +720,21 @@ static int userns_openat(void *arg, int dfd, int pid)
...
+	int provider_ret;
 
 	flags = oflags & ~(O_NOBUF | O_SERVICE | O_FORCE_LOCAL);
 
-	if (opts.stream && !(oflags & O_FORCE_LOCAL)) {
+	/* The external-memory provider is restore-only for now. */
+	if (oflags & (O_FORCE_LOCAL | O_DUMP))
+		provider_ret = -ENOTSUP;
+	else
+		provider_ret = extmem_open_image(path, flags, &provider_fd);
```
avagin, line 733: nit: please add an empty line after this one to improve readability.

criu/cr-restore.c
```diff
@@ -1972,7 +1973,7 @@ static int finalize_restore_detach(void)
 				pr_perror("Restoring regs for %d failed", pid);
 				return -1;
 			}
-			if (ptrace(PTRACE_DETACH, pid, NULL, 0)) {
+			if (ptrace(PTRACE_DETACH, pid, NULL, 0) && errno != ESRCH) {
```
avagin: why is it here?

criu/cr-service.c
```diff
@@ -605,8 +605,8 @@ static int setup_opts_from_req(int sk, CriuOpts *req)
 			goto err;
 	}
 
-	if (req->n_inherit_fd && !opts.swrk_restore) {
-		pr_err("inherit_fd is not allowed in standalone service\n");
+	if (req->n_inherit_fd && !opts.swrk_restore && opts.mode != CR_DUMP) {
+		pr_err("inherit_fd is allowed for dump or restore only\n");
```
avagin: opts.swrk_restore does not mean "restore mode"; it indicates that CRIU is running as a service worker (criu swrk).

criu/extmem.c
```diff
@@ -0,0 +1,388 @@
...
+
+static int provider_send_recv(ExtmemOp op, int socket_fd, struct msghdr *request, size_t request_len,
+			      struct provider_response *response)
+{
+	ssize_t sent;
+
+	mutex_lock(provider_lock);
+	do
+		sent = sendmsg(socket_fd, request, MSG_NOSIGNAL);
+	while (sent < 0 && errno == EINTR);
```
avagin: CRIU sets all signal handlers with SA_RESTART, we don't need to handle EINTR.

criu/extmem.c
```diff
@@ -0,0 +1,388 @@
...
+	}
+	mutex_unlock(provider_lock);
+	if (sent != (ssize_t)request_len) {
+		if (sent >= 0)
+			errno = EIO;
+		pr_err("extmem send failed for op %d: %s\n", op, strerror(errno));
+		return -1;
+	}
+	if (response->len == 0)
+		errno = ECONNRESET;
```
avagin, line 170: this errno will be overwritten by the next code block.

criu/cgroup.c
```diff
@@ -2337,6 +2337,9 @@ int prepare_cgroup(void)
 	struct cr_img *img;
 	CgroupEntry *ce;
 
+	if (opts.manage_cgroups == CG_MODE_IGNORE)
```
avagin, line 2340: This chunk should be in a separate patch with a proper comment.

criu/mem.c
dfeigin-nv: I can probably make it cleaner and more correct. I’ll remove the VMA enumeration and `VMA_EXT_PROVIDER`. Before reading a private VMA’s page bytes, CRIU will ask the provider while premapping it. Only a successful reply sets `provider_backed` in the restore-only `vma_area` state. `-ENOTSUP` leaves the VMA unchanged and uses normal CRIU restore.

criu/extmem.c
```diff
@@ -0,0 +1,371 @@
...
+	return inherit_fd_lookup_id(EXTMEM_PROVIDER_KEY);
+}
+
+int extmem_acquire_provider_fd(void)
+{
+	if (provider_thread_fd >= 0)
+		return -1;
+	provider_thread_fd = inherit_fd_lookup_id(EXTMEM_PROVIDER_KEY);
+	return provider_thread_fd < 0 ? -ENOTSUP : 0;
+}
```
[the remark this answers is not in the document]
dfeigin-nv: Good catch. `extmem_acquire_provider_fd()` will return success when the current thread already holds its provider-socket duplicate. `extmem_release_provider_fd()` remains a no-op when none is held.

criu/image.c
```diff
@@ -719,11 +721,21 @@ static int userns_openat(void *arg, int dfd, int pid)
...
-	if (opts.stream && !(oflags & O_FORCE_LOCAL)) {
+	/* The provider currently handles normal dumps and restores only. */
+	if ((oflags & O_FORCE_LOCAL) || (oflags & O_DUMP && opts.mode != CR_DUMP))
+		provider_ret = -ENOTSUP;
+	else
+		provider_ret = extmem_open_image(path, flags, &provider_fd);
+	if (provider_ret == 0)
+		ret = provider_fd;
+	else if (provider_ret != -ENOTSUP)
+		return -1;
```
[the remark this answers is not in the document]
dfeigin-nv: Good catch. I’ll update the PR description and documentation to describe both dump and restore. On dump, the provider supplies writable image FDs and receives `COMMIT` or `ABORT` after CRIU finishes.

criu/mem.c
dfeigin-nv: I think the AI is correct. I’ll make private hugetlb VMAs early premap requests too. The provider will return an `MFD_HUGETLB` FD with the VMA’s saved huge-page size, or `-ENOTSUP` before CRIU reads any page-image bytes. On `-ENOTSUP`, CRIU leaves the VMA unchanged and uses its normal hugetlb restore path.

criu/image.c
dfeigin-nv: Will do.

criu/cr-restore.c
dfeigin-nv: It is not needed. The series accidentally adds this unrelated change in an earlier commit and removes it here, leaving no net change. I’ll drop both hunks when rewriting the series.

criu/cr-service.c
dfeigin-nv: Oops, `swrk_restore` confused me. I’ll keep the original worker-only guard and say `inherit_fd` is only supported with `criu swrk`.

criu/extmem.c
dfeigin-nv: Indeed, I’ll remove the retry loops.

criu/extmem.c
dfeigin-nv: Indeed. I'll fix it.

criu/cgroup.c
dfeigin-nv: Will do.

dfeigin-nv:
I hope to have addressed all the comments. Please let me know what you think.

criu/mem.c
```diff
@@ -1046,6 +1047,31 @@ static int premap_private_vma(struct pstree_item *t, struct vma_area *vma, void
...
+	int ret;
+	unsigned long flags;
+
+	ret = extmem_get_vma(vpid(t), vma_id, vma->e->start, vma_entry_len(vma->e), &fd);
+	if (ret)
+		return ret;
+
+	flags = vma->e->flags;
+	vma->pvma = NULL;
+	vma->e->flags = (flags & ~MAP_ANONYMOUS) | MAP_PRIVATE;
```
maqiao-mq: hi, @dfeigin-nv
  I'd like to make a proposal.
  That VMA restoration currently uses fd-based mmap. Could we switch to a push model instead? The external-memory provider direct write the data into the target process via `process_vm_writev`.

  There are several reasons:

  1. The current implementation seems do not support file-backed private mmaps.
  2. Anonymous private mmaps are replaced with memfd-backed private mmaps. This changes the VMA semantics. As a result, madvise(MADV_DONTNEED) may behave completely differently.
  3. A VMA can be very large. The external-memory provider may run in a sidecar container with limited memory. In that case, it cannot provide the full memfd. With process_vm_writev, the provider can use a fixed-size buffer and stream the data in.

maqiao-mq: Maybe [UFFDIO_MOVE](https://man7.org/linux/man-pages/man2/UFFDIO_MOVE.2const.html) is another solution, but it requires Linux v6.8.

avagin:
@dfeigin-nv sorry for the delay. It is still on my to-do list.

images/extmem.proto
```diff
@@ -0,0 +1,41 @@
...
+syntax = "proto2";
+
+enum extmem_op {
+	EXTMEM_INIT = 1;
+	EXTMEM_OPEN_IMAGE = 2;
+	EXTMEM_GET_VMA = 3;
+	EXTMEM_GET_SHARED = 4;
+	EXTMEM_COMMIT = 5;
+	EXTMEM_ABORT = 6;
+	EXTMEM_WAIT_READY = 7;
```
avagin: why do we need this command?

criu/extmem.c
```diff
@@ -0,0 +1,349 @@
...
+	uint8_t packet[EXTMEM_MAX_PACKET];
+	char control[CMSG_SPACE(sizeof(int))];
+	struct msghdr msg;
+	struct iovec iov;
+	ssize_t len;
+};
+
+static int decode_provider_response(ExtmemOp op, struct provider_response *response, int *fd)
+{
+	ExtmemResp *resp;
```
avagin: nit: Reverse fir tree ordering please here and in other places.

criu/extmem.c
```diff
@@ -0,0 +1,349 @@
...
+
+#include "common/lock.h"
+#include "criu-log.h"
+#include "extmem.h"
+#include "files.h"
+#include "int.h"
+#include "images/extmem.pb-c.h"
+#include "rst-malloc.h"
+
+#define EXTMEM_MAX_PACKET   (8U << 10)
```
avagin: please add a comment to explain the chosen value.

criu/cr-restore.c
```diff
@@ -2496,7 +2517,16 @@ int cr_restore_tasks(void)
...
-	cr_plugin_fini(CR_PLUGIN_STAGE__RESTORE, ret);
+	if (plugin_initialized)
+		cr_plugin_fini(CR_PLUGIN_STAGE__RESTORE, ret);
+	if (ret) {
+		if (extmem_abort())
+			pr_err("Failed to abort external memory provider session\n");
+	} else if (extmem_commit()) {
+		/* COMMIT runs after tasks have resumed, so a failure cannot roll back
+		 * the restore. */
+		pr_err("Failed to commit external memory provider session\n");
```
avagin: what does extmem_commit mean in the restore case?

criu/extmem.c
```diff
@@ -0,0 +1,349 @@
...
+static enum extmem_session_state session_state;
+static __thread int provider_thread_fd = -1;
+static mutex_t *provider_lock;
+
+bool extmem_is_active(void)
+{
+	return session_state == EXTMEM_SESSION_ACTIVE;
+}
+
+static int provider_socket(void)
```
avagin: why do we need this function? Can the init fuction set provider_thread_fd and all others calls just use it?

criu/extmem.c
```diff
@@ -0,0 +1,349 @@
...
+	if (provider_thread_fd < 0)
+		close(socket_fd);
+	return ret;
+}
+
+int extmem_init(void)
+{
+	ExtmemReq req = EXTMEM_REQ__INIT;
+	bool release_fd = false;
+	int ret = -1;
```
avagin: empty line after declarations

criu/extmem.c
```diff
@@ -0,0 +1,349 @@
...
+	if (session_state == EXTMEM_SESSION_ACTIVE)
+		return 0;
+	if (session_state == EXTMEM_SESSION_CLOSED)
+		return -EPIPE;
+	if (session_state == EXTMEM_SESSION_UNSUPPORTED)
+		return -ENOTSUP;
+	if (provider_thread_fd < 0) {
```
avagin: add empty lines to split logical blocks in functions

criu/extmem.c
```diff
@@ -0,0 +1,349 @@
...
+	return provider_request(&req, fd);
+}
+
+int extmem_get_shared(unsigned long shmid, unsigned long length, int *fd)
+{
+	ExtmemGetShared shared = EXTMEM_GET_SHARED__INIT;
+	ExtmemReq req = EXTMEM_REQ__INIT;
+	int ret;
+
+	ret = extmem_init();
```
avagin: I think it sould be better to call extmem_init explicitly rather than calling it from each call in a lazy way...

criu/mem.c
```diff
@@ -1623,11 +1685,11 @@ int unmap_guard_pages(struct pstree_item *t)
 int open_vmas(struct pstree_item *t)
 {
 	int pid = vpid(t);
+	int exit_code = 0;
```
avagin: exit_code should be set to -1 by default.

criu/mem.c
```diff
@@ -1128,6 +1157,24 @@ static int premap_priv_vmas(struct pstree_item *t, struct vm_area_list *vmas, vo
...
+			if (!provider_fd_acquired) {
+				ret = extmem_acquire_provider_fd();
+				if (ret) {
+					pr_err("Provider is unavailable for private VMA\n");
+					ret = -1;
+					break;
+				}
+				provider_fd_acquired = true;
+			}
+			ret = premap_provider_vma(t, vma, current_vma_id, at);
```
avagin: should be skip vmas with the MAP_GROWS* flags?

criu/image.c
```diff
@@ -719,11 +721,22 @@ static int userns_openat(void *arg, int dfd, int pid)
...
+	int provider_ret;
 
 	flags = oflags & ~(O_NOBUF | O_SERVICE | O_FORCE_LOCAL);
 
-	if (opts.stream && !(oflags & O_FORCE_LOCAL)) {
+	/* The provider handles normal dumps and restores only. */
+	if ((oflags & O_FORCE_LOCAL) || (oflags & O_DUMP && opts.mode != CR_DUMP))
+		provider_ret = -ENOTSUP;
+	else
+		provider_ret = extmem_open_image(path, flags, &provider_fd);
```
avagin: it might be one of parent images, should we handle this case?

criu/cr-dump.c
```diff
@@ -1957,11 +1958,27 @@ static int cr_dump_finish(int ret)
...
+	if (ret || post_dump_ret) {
+		if (extmem_abort())
+			pr_err("Failed to abort external memory provider session\n");
+	}
+
+	if (!ret && !post_dump_ret) {
+		/* Commit before releasing the dumped task. */
+		if (extmem_commit()) {
+			pr_err("Failed to commit external memory provider session\n");
+			ret = -1;
```
avagin: pstree_switch_state will resume processes in this case, but other actions like network_unlock() will not be executed. You probably need to call extmem_commit near bfd_flush_images.

PR #3122, opened 2026-08-25
criu: dump/restore inet socket uid

ctrox:
The inet sockets UID is currently lost after a privileged dump/restore and ends up belonging to 0. This writes ie->uid to the image and also restores it again if it's found in the image.

Let me know if there's a reason this is currently not being done, I mainly noticed this when checkpointing programs that make use of SO_REUSEPORT which requires all the sockets to have the same UID.


avagin:
@ctrox thanks for the contribution. At first glance, the change looks good. Could you please add a test for this?

criu/include/sk-inet.h
```diff
@@ -41,6 +41,7 @@ struct inet_sk_desc {
 	unsigned int dst_addr[4];
 	unsigned short shutdown;
 	bool cork;
+	uid_t uid;
```
rst0git, line 44: @ctrox I think we would need to also ensure that unbound sockets retain their UID across checkpoint and restore: https://github.com/rst0git/criu/commit/ac9047ee5060220e221b1314be0b3d47b96fe807

Otherwise, an unbound socket could incorrectly become owned by the user running CRIU after restore:
```
$ sudo ./test/zdtm.py run -t zdtm/static/sock_uid -f h
userns is supported
=== Run 1/1 ================ zdtm/static/sock_uid
======================== Run zdtm/static/sock_uid in h =========================
Start test
Test is SUID
./sock_uid --pidfile=sock_uid.pid --outfile=sock_uid.out
Run criu dump
Run criu restore
Send the 15 signal to  70
Wait for zdtm/static/sock_uid(70) to die for 0.100000
################ Test zdtm/static/sock_uid FAIL at result check ################
12:06:10.421:    70: FAIL: sock_uid.c:104: tcp-unbound socket uid changed: 12345 -> 0 (errno = 11 (Resource temporarily unavailable))

##################################### FAIL ####################################
```

ctrox: thanks, I cherry-picked your commit

rst0git: Thanks - it might be better to apply the change in your patch.

ctrox: sure no problem

rst0git (approved):
LGTM

criu/sk-inet.c
```diff
@@ -528,6 +529,11 @@ static int do_dump_one_inet_fd(int lfd, u32 id, const struct fd_parms *p, int fa
 	ie.backlog = sk->wqlen;
 	ie.flags = p->flags;
 
+	if (sk->uid != 0) {
```
avagin, line 532: sk->uid can be zero in a criu namespace, but this uid can be mapped to a user namespace into another uid. Should we just always dump socket uids?

ctrox: I initially did this to not waste this in the dump but agreed it's simpler to just always dump it and not have to worry about other edge cases: #3124

PR #3123, opened 2026-08-26
Random fixes

avagin:


soccr/soccr.c
```diff
@@ -750,6 +750,18 @@ int libsoccr_restore(struct libsoccr_sk *sk, struct libsoccr_sk_data *data, unsi
...
+		 * Kernel's tcp_repair_set_window() returns -EINVAL if
+		 * after(opt.snd_wl1, tp->rcv_nxt + opt.rcv_wnd).
+		 *
+		 * When rcv_wnd is zero or retracted, snd_wl1 may legitimately
+		 * exceed data->inq_seq + wopt.rcv_wnd (e.g. zero-window probe
+		 * or out-of-order segment received within previous window).
+		 * Clamp snd_wl1 to pass kernel validation.
+		 */
+		if ((int32_t)(wopt.snd_wl1 - (data->inq_seq + wopt.rcv_wnd)) > 0)
+			wopt.snd_wl1 = data->inq_seq + wopt.rcv_wnd;
```
rst0git, line 763: @avagin Can we also add a test for this? For example, something like the following: https://github.com/rst0git/criu/commit/7a1386b252bfb27e11ab6154f38c4f370b9d4051

avagin: I think you are trying to introduce tcp-repair unit tests, it is better to add such tests in the kernel tools/testing/selftests. This patch fixes socket-tcpbuf failures triggered in our ci.

rst0git: Sorry, I missed the socket-tcpbuf failures in CI. It looks like our tests can't reliably reproduce this error. Do you have any ideas how we could improve them?

rst0git (approved):

PR #3128, opened 2026-09-01
plugins/tpu: add Google TPU checkpoint/restore plugin

aishukamal:
Adds a TPU plugin modeled on the CUDA plugin. It execs the tpucheckpoint utility (from the gVisor tree) at the CHECKPOINT_DEVICES and RESUME_DEVICES_LATE hooks, selectively resuming the libtpu control thread inside the frozen process to service each action. PAUSE_DEVICES does not quiesce the device yet; the caller must ensure the workload is TPU-idle before dump (see README).

Build-tested only; not yet validated on TPU hardware.

### DIFF
--- a/plugins/tpu/README.md
+++ b/plugins/tpu/README.md
@@ -1,0 +1,50 @@
+Checkpoint and Restore for TPU applications with CRIU
+======================================================
+
+# Requirements
+The tpucheckpoint utility should be placed somewhere in your $PATH, and the
+workload's TPU runtime must support the tpu_control checkpoint/restore
+protocol. The plugin detects support by the presence of the libtpu control
+thread in the workload process.
+
+## tpucheckpoint
+The tpucheckpoint utility lives in the gVisor tree:
+https://github.com/google/gvisor/tree/master/tools/tpucheckpoint
+
+tpucheckpoint is a binary utility used to issue checkpointing commands to
+TPU applications. It speaks the libtpu control protocol
+(pkg/sentry/control/tpu_control.proto in gVisor): libtpu spawns a control
+thread named "libtpu{XXXXYYYY}" in the workload process, where the
+8-character hex suffix encodes the request/response pipe FD numbers, and
+tpucheckpoint discovers this thread via /proc and exchanges
+length-delimited protobuf messages over the pipes.
+
+# Checkpointing Procedure
+tpucheckpoint exposes 2 actions used in the checkpointing process:
+checkpoint, restore.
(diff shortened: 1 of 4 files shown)


aishukamal:
@avagin PTAL.

plugins/tpu/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+# Requirements
+The tpucheckpoint utility should be placed somewhere in your $PATH, and the
+workload's TPU runtime must support the tpu_control checkpoint/restore
+protocol. The plugin detects support by the presence of the libtpu control
+thread in the workload process.
+
+## tpucheckpoint
+The tpucheckpoint utility lives in the gVisor tree:
+https://github.com/google/gvisor/tree/master/tools/tpucheckpoint
```
rst0git, line 12: @aishukamal This file doesn't seem to exist.

plugins/tpu/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+tpucheckpoint is a binary utility used to issue checkpointing commands to
```
rst0git, line 14: @aishukamal Could we implement this within the plugin itself? For example, something similar to the approach in https://github.com/checkpoint-restore/criu/pull/3076

plugins/tpu/README.md
avagin: https://github.com/google/gvisor/pull/14455

plugins/tpu/README.md
avagin: We might consider sending libtpu checkpoint/restore control messages directly from the plugin instead of relying on the tpucheckpoint tool. With the NVIDIA plugin, we found that executing a binary for each dumped or restored process introduces measurable latency, increasing overall downtime.

trantoji: This is how being done today with gVisor.

>With the NVIDIA plugin, we found that executing a binary for each dumped or restored process introduces measurable latency, increasing overall downtime.

Do you know the biggest factor that is contributing to the latency?

rst0git: > Do you know the biggest factor that is contributing to the latency?

@trantoji We included an overview of this in the following KubeCon talk (slides 8–9): [Optimizing Error Recovery for Cost-Efficient Distributed AI Model Training with Kubernetes](https://sched.co/2CW7Z).

My observation was that most of the overhead comes from invoking the external binary and parsing its output for each PID being checkpointed.

Issue #2120, opened 2023-03-13
Codacy: condition is always true/false


mihalicyn:
https://app.codacy.com/gh/mihalicyn/criu/issues?&filters=W3siaWQiOiJMYW5ndWFnZSIsInZhbHVlcyI6W119LHsiaWQiOiJDYXRlZ29yeSIsInZhbHVlcyI6W119LHsiaWQiOiJMZXZlbCIsInZhbHVlcyI6W119LHsiaWQiOiJQYXR0ZXJuIiwidmFsdWVzIjpbIjIyMzciXX0seyJpZCI6IkF1dGhvciIsInZhbHVlcyI6W119XQ==

Daz-3ux:
I want to work on this issue, can I have a try? :)

TaeminHa:
Hi! I'm a student from UT Austin, and would be happy to take on this issue! Can you assign this to me?

mihalicyn:
Feel free to take :)

Jayanth19524:
is this issue solved??

mihalicyn:
@TaeminHa ?

TaeminHa:
Hi, this issue is resolved. I'll create a separate PR to put the changes in separate commits as discussed.

Snorch:
No this issue is NOT resolved, we have really-really draft PR #2141, which from the first glance tries to do the right thing in code. But it is not ready to review as it has bad commit structure: no signed-offs, no rebase to latest criu-dev, empty commits, bad commit subject and description format/wording, fixup commits.

Read https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md it contains all the information about how the right PR should look like.

vishisht-dubey:
Hii @Snorch i would like to contibute to this issue can u assign this to me?

Min2who:
i have seen the issue and i guess i can solve the issue can you assign me this issue

Snorch:
No real need to assign, comment that you are working on this is enough, @vishisht-dubey @Min2who please communicate with each other not to do the same thing twice =)

rohitsinghnaruka:
Hii , @Snorch  & @mihalicyn 
i'm Working on this .

AtriChatterjee1:
HI ! @Snorch and @mihalicyn ! I am a new contributor to criu, and would love to work on this

AtriChatterjee1:
Hey @Snorch Can't find the issue on codacy for some reason, mind sharing it again

[...]

PR #3130, opened 2026-09-01
kerndat: remove redundant ret check in root_only_init()

Goyamjain06:
## What problem does this solve?

Codacy static analysis (pattern 2237, "condition is always true/false") flags a redundant condition in `root_only_init()` in `criu/kerndat.c`, tracked in #2120.

## What changed

The first check in `root_only_init()` was:

    if (!ret && kerndat_loginuid()) {

Since `ret` is freshly initialized to `0` immediately before this check, and no code path between the initialization and this check can change it, `!ret` is always true here — making the check dead/redundant.

Changed to:

    if (kerndat_loginuid()) {

All subsequent `!ret && kerndat_*()` checks later in the same function are intentionally left untouched, since they correctly guard against proceeding after an earlier failure (short-circuit "stop on first error" pattern).

## How to test

1. Build CRIU: `make` — compiles cleanly with `-Werror`, no warnings.
2. Run `criu check` — behaves identically before/after this change (verified in a Docker-based Ubuntu 22.04 environment).
3. This is a pure control-flow simplification with no behavioral change, so no test suite specifically covers this function. `root_only_init()` is exercised indirectly any time CRIU runs kerndat initialization as root.

## Related issue

Relates to #2120 (Codacy pattern 2237). This is one of several redundant conditions flagged by that issue; a previous attempt to fix several of them (#2141) went stale and was closed unmerged due to messy/unrebased commit history. This PR addresses just the kerndat.c instance cleanly, as a first, minimal, reviewable step.

### DIFF
--- a/criu/kerndat.c
+++ b/criu/kerndat.c
@@ -2003,7 +2003,7 @@ static int root_only_init(void)
 	if (opts.unprivileged)
 		return 0;
 
-	if (!ret && kerndat_loginuid()) {
+	if (kerndat_loginuid()) {
 		pr_err("kerndat_loginuid failed when initializing kerndat.\n");
 		ret = -1;
 	}


avagin:
I actually like the current form, because it looks like other checks in this function, so it is easier to read this code.

Goyamjain06:
Totally fair — appreciate the quick review! Makes sense to prioritize consistency with the surrounding checks over the strict redundancy. I'll leave this as-is and look at the other patterns flagged under #2120 with that same lens before proposing anything else. Thanks for the feedback!

PR #3132, opened 2026-09-02
netfilter: fix silent truncation of the nftables table name

### LINKED ISSUE #3131: netfilter: nftables table name silently truncated (32 byte buffer, 47 needed)
### Description

`nftables_get_table()` in `criu/netfilter.c` builds the per-run nftables table name as `"inet CRIU-" + criu_run_id`, but every caller passes a 32 byte buffer while the name needs 47 bytes. The name is silently truncated and the function still reports success.

`criu_run_id` became a 36 character UUID string in commit 54795f174 ("criu: use libuuid for criu_run_id generation"):

```
"inet CRIU-"        10 chars
criu_run_id (UUID)  36 chars
NUL                  1
                  = 47 bytes required
```

All four call sites declare `char table[32]`:

- `criu/netfilter.c:161` `nftables_init_connection_lock()`
- `criu/netfilter.c:244` `nftables_lock_connection_raw()`
- `criu/net.c:3365` `nftables_lock_network_internal()`
- `criu/net.c:3472` `nftables_network_unlock()`

### Why the existing check does not catch it

```c
ret = snprintf(table, n, "inet CRIU-%s", criu_run_id);
...
if (ret < 0) {
        pr_err("Cannot generate CRIU is nftables table name\n");
        return -1;
}
return 0;
```

`snprintf()` returns the length it *would* have written, not what it wrote, so on truncation it returns 46 and the `ret < 0` test passes. The compiler cannot warn either, because the `snprintf()` sits inside `nftables_get_table()` where the destination size is only known through the `n` parameter.

### Reproducer

```c
#include <stdio.h>
#include <string.h>
#include <uuid/uuid.h>

int main(void)
{
        uuid_t uuid;
        char run_id[37], table[32];
        int ret;

        uuid_generate(uuid);
        uuid_unparse(uuid, run_id);

        ret = snprintf(table, sizeof(table), "inet CRIU-%s", run_id);

        printf("ret       = %d (>= 0, so CRIU treats this as success)\n", ret);
        printf("wanted    = inet CRIU-%s\n", run_id);
        printf("written   = %s (%zu chars)\n", table, strlen(table));
        return 0;
}
```

```
$ gcc -o repro repro.c -luuid && ./repro
ret       = 46 (>= 0, so CRIU treats this as success)
wanted    = inet CRIU-059cf9cb-3ff4-4591-a9dd-67402190e10b
written   = inet CRIU-059cf9cb-3ff4-4591-a9 (31 chars)
```

Building the same call with a visible destination makes GCC report it directly:

```
warning: output may be truncated writing up to 36 bytes into a region of size 22
note: output 47 bytes into a destination of size 32
```

### Impact

All four call sites truncate identically, so dump and restore agree on the shortened name and network locking still functions. The effect is that 15 of the 36 run id characters are discarded, weakening the unique table naming that commit 9838d34de ("criu: use unique table names for nftables based locking") introduced, and the error path that was written to catch a failed name generation cannot actually fire.

### Version

```
Version: 4.2.1
GitID: v4.2.1-433-g90dd8aea2
```

I have a fix ready and will open a PR against `criu-dev`.

Utkal059:
Fixes #3131

`nftables_get_table()` builds the table name as `"inet CRIU-" + criu_run_id`. Since commit 54795f174 ("criu: use libuuid for criu_run_id generation") the run id is a 36 character UUID string, so the name needs 47 bytes, but all four call sites pass a 32 byte buffer:

```
wanted:  inet CRIU-059cf9cb-3ff4-4591-a9dd-67402190e10b
written: inet CRIU-059cf9cb-3ff4-4591-a9
```

`snprintf()` returns the length it would have written, so the existing `ret < 0` check cannot detect the truncation and the function reports success.

This adds `NFTABLES_TABLE_NAME_LEN`, derived from `RUN_ID_HASH_LENGTH` so it stays correct if the run id format changes again, uses it for the four `table` buffers, and makes `nftables_get_table()` treat truncation as an error.

### Testing

- `make -j8` builds clean with no new warnings, with `libnftables` present so the `CONFIG_HAS_NFTABLES_LIB_API_*` paths are compiled.
- Verified against the real header that `NFTABLES_TABLE_NAME_LEN` evaluates to 47 and that the full 46 character name is preserved.

I was not able to exercise the nftables locking paths end to end locally: the test kernel here is a WSL2 6.6 build without `CONFIG_INET_DIAG_DESTROY`, so the TCP established tests that drive network locking cannot run. Happy to add a test or rework the approach if you would prefer a different fix.

### DIFF
--- a/criu/include/netfilter.h
+++ b/criu/include/netfilter.h
@@ -1,6 +1,16 @@
 #ifndef __CR_NETFILTER_H__
 #define __CR_NETFILTER_H__
 
+#include "compel/infect-util.h"
+
+/*
+ * The nftables table name is "inet CRIU-" followed by the CRIU run id,
+ * which is a UUID string. RUN_ID_HASH_LENGTH already accounts for the
+ * terminating NUL.
+ */
+#define NFTABLES_TABLE_PREFIX	"inet CRIU-"
+#define NFTABLES_TABLE_NAME_LEN (sizeof(NFTABLES_TABLE_PREFIX) - 1 + RUN_ID_HASH_LENGTH)
+
 struct inet_sk_desc;
 extern int iptables_lock_connection(struct inet_sk_desc *);
 extern int iptables_unlock_connection(struct inet_sk_desc *);
--- a/criu/net.c
+++ b/criu/net.c
@@ -3362,7 +3362,7 @@ static inline int nftables_lock_network_internal(bool restore)
 	cleanup_file FILE *fp = NULL;
 	struct nft_ctx *nft;
 	int ret = 0, exit_code = -1;
-	char table[32];
+	char table[NFTABLES_TABLE_NAME_LEN];
 	char buf[128];
 
 	if (nftables_get_table(table, sizeof(table)))
@@ -3469,7 +3469,7 @@ static inline int nftables_network_unlock(void)
 	int ret = 0;
 	cleanup_file FILE *fp = NULL;
 	struct nft_ctx *nft;
-	char table[32];
+	char table[NFTABLES_TABLE_NAME_LEN];
 	char buf[128];
 
 	if (nftables_get_table(table, sizeof(table)))
--- a/criu/netfilter.c
+++ b/criu/netfilter.c
@@ -158,7 +158,7 @@ int nftables_init_connection_lock(void)
 #if defined(CONFIG_HAS_NFTABLES_LIB_API_0) || defined(CONFIG_HAS_NFTABLES_LIB_API_1)
 	struct nft_ctx *nft;
 	int ret = 0;
-	char table[32];
+	char table[NFTABLES_TABLE_NAME_LEN];
 
 	if (nftables_get_table(table, sizeof(table)))
 		return -1;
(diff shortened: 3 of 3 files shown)


criu/include/netfilter.h
```diff
@@ -1,6 +1,15 @@
...
 #define __CR_NETFILTER_H__
 
+#include "compel/infect-util.h"
+
+/*
+ * The nftables table name is "inet CRIU-" followed by the CRIU run id,
+ * which is a UUID string. RUN_ID_HASH_LENGTH already accounts for the
+ * terminating NUL.
+ */
+#define NFTABLES_TABLE_NAME_LEN (sizeof("inet CRIU-") - 1 + RUN_ID_HASH_LENGTH)
```
avagin: please define a constant for "inet CRIU-"

Utkal059: Done, thanks.
Added NFTABLES_TABLE_PREFIX and derived NFTABLES_TABLE_NAME_LEN from it so the length and the format strings cannot drift apart.

PR #3137, opened 2026-09-06
proc_parse: use breadline() to handle long mountinfo entries

### LINKED ISSUE #3129: criu: handle mountinfo lines larger than 4 KiB
CRIU reads `/proc/<pid>/mountinfo` with a fixed 4 KiB buffer.

That breaks on valid containerd OverlayFS mounts with a long `lowerdir=` list. We saw this mountinfo entry at 6,131 bytes.

`fgets()` returns the first 4 KiB as one entry, then CRIU reads the remaining part of the same entry and tries to parse it as another mount. That second read starts in the middle of `lowerdir=...`, not with a mount ID, so parsing fails.

We probably should use `getline()` in `criu/proc_parse.c` for the mountinfo reads in:

- `parse_mountinfo()`
- `get_mountinfo_sdev_from_mntid()`

This gives the existing parser a complete mountinfo entry instead of a fixed-size chunk.

I have a small fix for this if you think this should go in.

RohanOnKeys:
fgets() with a fixed 4096-byte buffer silently truncates mountinfo entries
longer than BUF_SIZE. The truncated remainder gets passed to the parser as
a new entry on the next call, which fails because it starts mid-field
instead of with a mount ID.

Hit this with a real containerd OverlayFS mount where a long lowerdir=
list produced a 6131-byte mountinfo entry, causing criu dump to fail with
"Bad format in <pid> mountinfo".

Switch parse_mountinfo() and get_mountinfo_sdev_from_mntid() to use
breadline(), replacing fopen_proc() with open_proc() + bfdopenr() and
fclose() with bclose(), consistent with how smaps and status are read
elsewhere in proc_parse.c. The global buf/BUF_SIZE are left untouched
since all other fgets() call sites read short bounded lines.

Fixes: #3129
Signed-off-by: Rohan <rohanpxquantim@gmail.com>

### DIFF
--- a/criu/proc_parse.c
+++ b/criu/proc_parse.c
@@ -1649,18 +1649,29 @@ out:
 static int get_mountinfo_sdev_from_mntid(int mnt_id, unsigned int *sdev)
 {
 	int exit_code = -1;
-	FILE *f;
+	struct bfd f = { .fd = -1 };
+	char *line;
 
-	f = fopen_proc(PROC_SELF, "mountinfo");
-	if (!f)
+	f.fd = open_proc(PROC_SELF, "mountinfo");
+	if (f.fd < 0)
 		return -1;
+	if (bfdopenr(&f))
+		goto err;
 
-	while (fgets(buf, BUF_SIZE, f)) {
+	while (1) {
 		unsigned int kmaj, kmin;
 		int id;
 
-		if (sscanf(buf, "%i %*i %u:%u", &id, &kmaj, &kmin) != 3) {
-			pr_err("Failed to parse mountinfo line %s\n", buf);
+		line = breadline(&f);
+		if (!line)
+			break;
+		if (IS_ERR(line)) {
+			pr_err("Failed to read mountinfo\n");
+			goto err;
+		}
+
+		if (sscanf(line, "%i %*i %u:%u", &id, &kmaj, &kmin) != 3) {
+			pr_err("Failed to parse mountinfo line %s\n", line);
 			goto err;
 		}
 
@@ -1671,7 +1682,7 @@ static int get_mountinfo_sdev_from_mntid(int mnt_id, unsigned int *sdev)
 		}
 	}
 err:
-	fclose(f);
+	bclose(&f);
 	return exit_code;
 }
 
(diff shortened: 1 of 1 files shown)


RohanOnKeys:
@avagin can you please have look n approve ci

dfeigin-nv (commented):
You beat me to it :) 
In alignment with @avagin 's comment on the issue - maybe we should use breadline(), as it is better understood. getline may vary by libc impl.

RohanOnKeys:
> You beat me to it :) In alignment with @avagin 's comment on the issue - maybe we should use breadline(), as it is better understood. getline may vary by libc impl.

ahuh, i almost scrolled his comment away on your issue, MB ;)
i implemented your suggestions cuz i felt it might work (+it was easies to call) , 
aight then . lemme see how breadline() works first then imma replace thigns...

RohanOnKeys:
@dfeigin-nv this breadline() is a criu internal wrapper eh... will take a while to get the changes on my head

criu/proc_parse.c
```diff
@@ -1702,13 +1705,16 @@ int get_sdev_from_fd(int fd, unsigned int *sdev, bool parse_mountinfo)
...
-	f = fopen_proc(pid, "mountinfo");
-	if (!f)
+	f.fd = open_proc(pid, "mountinfo");
+	if (f.fd < 0)
+		return NULL;
+	if (bfdopenr(&f))
 		return NULL;
 
-	while (fgets(buf, BUF_SIZE, f)) {
+	while ((line = breadline(&f))) {
```
dfeigin-nv: breadline can return `ERR_PTR(-EIO)`. You can call `IS_ERR_OR_NULL()`

criu/proc_parse.c
```diff
@@ -1649,18 +1649,21 @@ int parse_timens_offsets(struct timespec *boff, struct timespec *moff)
 static int get_mountinfo_sdev_from_mntid(int mnt_id, unsigned int *sdev)
 {
 	int exit_code = -1;
```
dfeigin-nv, line 1651: The commit title and description still use getline

criu/proc_parse.c
```diff
@@ -1649,18 +1649,21 @@ int parse_timens_offsets(struct timespec *boff, struct timespec *moff)
 static int get_mountinfo_sdev_from_mntid(int mnt_id, unsigned int *sdev)
 {
```
dfeigin-nv, line 1650: What about get_mountinfo_sdev_from_mntid?

criu/proc_parse.c
avagin: we are usually avoiding assignments in check expressions. You can rewrite it like this:
```
while (1) {
  line = breadline(&f);
  if (!line)
    break;
  if (IS_ERR(line)) {
    pr_err(...);
    goto err;
  }
```

criu/proc_parse.c
avagin: all functions that parse mountinfo using fgets should be fixed.

RohanOnKeys:
Done. I think, updated `get_mountinfo_sdev_from_mntid()` to use `breadline()` as well, so all mountinfo parsers using `fgets()` are covered now. the ci (hopefully) passes

RohanOnKeys:
@dfeigin-nv yaayayayayayay!! LFG , merged!!! hoping to work more with you guys, where together we snniff out issues and patch fixes

PR #3138, opened 2026-09-09
plugins: pass plugin options as argv

avagin:
Plugins are loaded after CRIU parses command-line and RPC options. A plugin therefore cannot add its own parser without changing CRIU for every new setting.

Keep namespaced PLUGIN.NAME=VALUE strings in order and expose them as an argv array. This lets plugins use getopt_long() and ignore options they do not recognize. Append RPC and libcriu values instead of replacing matching names, and clear request values between service requests.


criu/config.c
```diff
@@ -410,8 +411,9 @@ static int pre_parse(int argc, char **argv, bool *usage_error, bool *no_default_
 	return 0;
 }
 
-void init_opts(void)
+int init_opts(void)
 {
+	cr_plugin_options_free();
```
rst0git, line 416: As you mentioned, we might need to avoid resetting the plugin options here so that service workers can retain startup defaults: https://github.com/rst0git/criu/commit/d0819bc400e9af6a68df9239f749a21f2021c644 (feel free to squash both patches into a single commit)

avagin: It is not plugin specific. Plugin options are not different from other options, so this is unrelated problem and should be solved separately. CRIU service workers have to parse global configs after init_opts.

rst0git (approved):
LGTM

PR #3140, opened 2026-09-11
cgroup: clone /sys/fs/cgroup mount when dumping cgroup v2

avagin:
When /sys/fs/cgroup is the root cgroup v2 mount (detected using statmount), use sys_open_tree() with OPEN_TREE_CLONE in open_cgroupfs() to obtain a detached mount of the cgroup v2 hierarchy.

This preserves existing superblock and mount options while stripping any child overmounts that may exist in subdirectories of /sys/fs/cgroup.


kolyshkin:
Will it fix https://github.com/checkpoint-restore/criu/issues/3029?

avagin:
> Will it fix #3029?

It will fix it in a standard environment where cgroupfs is mounted in /sys/fs/cgroup.

kolyshkin:
> It will fix it in a standard environment where cgroupfs is mounted in /sys/fs/cgroup.

@avagin for cgroup v2 this is pretty much a standard (for example, runc and (I think) crun have `/sys/fs/cgroup` path hardcoded for v2, and I have yet to see a cgroup v2 system in which cgroupfs is mounted to a different path.

avagin:
@kolyshkin we know that some users run CRIU inside a container to avoid dumping/restoring namespaces and to ensure that the restored workload won't break container isolation. We probably won't have a /sys/fs/cgroup mount in a standard K8s container...

PR #3141, opened 2026-09-11
files-reg: restore ghost device files through usernsd in user namespaces

### LINKED ISSUE #92: Ghost device files cannot be restored in user namespace
The mknod() system call is protected with global CAP_MKNOD, thus ghost device restoration fails.

RohanOnKeys:
## Summary

Ghost device files are currently restored with a direct `mknod()` call. During restores inside a user namespace, that call executes in the restoring process and can fail because creating character and block device nodes requires capabilities that are not available there.

This change routes ghost device node creation through `userns_call()`, allowing the existing `usernsd` helper to perform `mknod()` from the appropriate privilege context. The restore path for regular files, directories, FIFOs, and symlinks is unchanged.

## Changes

* Add a `userns_mknod()` callback that follows the existing `userns_call()` callback pattern.
* Add `ghost_mknod_args` to pass the path, mode, and device number to the helper.
* Replace the direct `mknod()` call for ghost character and block devices with `userns_call(userns_mknod, ...)`.
* Leave the restore path unchanged for regular files, directories, FIFOs, and symlinks.

## Why this approach

CRIU already uses `userns_call()` for operations that must execute outside a restricted user namespace, including `open_by_handle()`, `mount_root()`, `apply_sb_flags()`, and cgroup related operations. Reusing the same mechanism keeps the implementation consistent with the existing restore architecture.

## Verification

* `make -j$(nproc)` completed successfully.
* `git diff --check` reported no formatting issues.
* The change is limited to `criu/files-reg.c`.
* Full ZDTM restore tests could not be run in GitHub Codespaces because the cgroup mount setup fails with `permission denied`.

Fixes: #92

Signed-off-by: Rohan Pattanayak rohanpxquantum@gmail.com

### DIFF
--- a/criu/files-reg.c
+++ b/criu/files-reg.c
@@ -427,6 +427,22 @@ static int ghost_apply_metadata(const char *path, GhostFileEntry *gfe)
 	return 0;
 }
 
+struct ghost_mknod_args {
+	char path[PATH_MAX];
+	mode_t mode;
+	dev_t rdev;
+};
+
+static int userns_mknod(void *arg, int fd, pid_t pid)
+{
+	struct ghost_mknod_args *a = arg;
+
+	(void)fd;
+	(void)pid;
+
+	return mknod(a->path, a->mode, a->rdev);
+}
+
 static int create_ghost_dentry(char *path, GhostFileEntry *gfe, struct cr_img *img)
 {
 	int ret = -1;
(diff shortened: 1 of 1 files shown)


avagin:
How these files were created in a container?

RohanOnKeys:
> How these files were created in a container?

which files ?? the ghost files you mean,
I have not reproduced the original 2015 scenario yet. My patch was based on the current restore path and the issue description. cuz i was running this on a gh codespace and i did not have priveleged mode

i believe ghost device nodes would have to exist before checkpointing. In a container they could be created only if the container had the necessary privileges, for example --privileged or CAP_MKNOD together with the appropriate device cgroup permissions.

RohanOnKeys:
I traced the failure from ghost device restoration to create_ghost_dentry(), where device ghosts were recreated via a direct mknod(). I then compared it with other privileged operations already routed through userns_call() (open_by_handle, mount_root, userns_move) and found that ghost device creation was the only path bypassing that mechanism.

RohanOnKeys:
@avagin could you please lmk if we should proceed with this approach or give instructions to reproduce this issue. As this was quite old, I tried to guess behaviour by running close appx steps and deduced the soln

rst0git:
> > How these files were created in a container?
> which files ?? the ghost files you mean

@RohanOnKeys Usually, the OCI runtime creates these device files. For example, see `createDeviceNode` in runc: https://github.com/opencontainers/runc/blob/930f199c/libcontainer/rootfs_linux.go#L989

> I have not reproduced the original 2015 scenario yet. My patch was based on the current restore path and the issue description. cuz i was running this on a gh codespace and i did not have priveleged mode

It looks like the issue above describes an edge case that wasn't supported. I'm not sure if there exist real-world use-cases where we need to restore ghost device files inside a user namespace.

RohanOnKeys:
> > I have not reproduced the original 2015 scenario yet. My patch was based on the current restore path and the issue description. cuz i was running this on a gh codespace and i did not have priveleged mode
> 
> It looks like the issue above describes an edge case that wasn't supported. I'm not sure if there exist real-world use-cases where we need to restore ghost device files inside a user namespace.

true Facts.

PR #3145, opened 2026-09-13
stream: allow --stream with incremental images and over RPC

dlunev:
--stream is all or nothing today. Two comments in the tree say so:

/* Image streaming lacks support for incremental images */
/* Image streaming is not supported with CRIU's service feature ... */

This series removes both, and fixes two combinations that fail without being diagnosed.

Fixes (patches 1-2, independent, with Fixes: tags). --stream --auto-dedup trips BUG_ON(flags != img_streamer_mode), because --auto-dedup opens the pages image O_RDWR while the streamer serves only the mode it was started with, and dedup is skipped for streamed pages anyway. --stream --page-server on a dump silently produces an unrestorable checkpoint, because the page server writes the pages images outside the stream; it is now refused.

Incremental images (patches 3-8). A streamed dump can sit on an on-disk parent snapshot: the current snapshot streams, parents are read from disk via O_FORCE_LOCAL. "Streamed" moves from the global opts.stream onto the page read, so try_open_parent() passes PR_FORCE_LOCAL down a whole chain. This is what makes memory pre-copy usable with streaming. The zdtm test asserts PE_PARENT entries exist, not just that the restore works — without the series it still restores, it merely re-sends every page, so a correctness-only test would gate nothing.

RPC (patches 9-11). image_dir_mode() moves to image.c so the RPC path can derive the open mode instead of hardcoding -1, and criu_opts gains a stream field. A new test in test/others/rpc runs a streamed dump and restore over RPC and asserts no image lands in the directory. It skips itself where criu-image-streamer is absent.

Testing. Full zdtm suite plain, --stream, and --stream --pre 2; plus --stream --dedup, --pre 5, --snaps, --pre-dump-mode read; plus make  -C test/others/rpc/ run. Every patch builds on its own. Against an unpatched --stream run on the same machine, the series adds no failures.

Not included. The patch 2 refusal covers only the dump side; --stream with --lazy-pages is separate work coming as its own series, and adding a refusal here only to remove it there seemed worse. maps09/maps10 stay excluded from the streaming job — MAP_HUGETLB mappings are not premapped, so their contents are deferred to the restorer, which preads a pages image that streaming serves once. That predates this series.


rst0git:
Adding a reference also to the changes here: https://github.com/checkpoint-restore/criu-image-streamer/pull/73

rst0git:
@dlunev It looks like the current design assumes that "parent" snapshots are still needed for `--stream` with incremental images. This is not actually the case. When we perform several iterations of `criu pre-dump`, followed by `criu dump`, during restore we need only the final version of the memory pages. There is no need to preserve the earlier version of pages.

I think better design would be for criu-image-streamer to implement a mechanism that overwrites the pages from previous iterations when a newer version of the same page arrives. In this way, during restore criu-image-streamer already has the final version of the memory pages and CRIU doesn't need to handle "parent" snapshots, i.e., the idea is similar to how you can squash several git commits (incremental changes) into a single one.

avagin:
> @dlunev It looks like the current design assumes that "parent" snapshots are still needed for `--stream` with incremental images. This is not actually the case. When we perform several iterations of `criu pre-dump`, followed by `criu dump`, during restore we need only the final version of the memory pages. There is no need to preserve the earlier version of pages.

> I think better design would be for criu-image-streamer to implement a mechanism that overwrites the pages from previous iterations when a newer version of the same page arrives. In this way, during restore criu-image-streamer already has the final version of the memory pages and CRIU doesn't need to handle "parent" snapshots, i.e., the idea is similar to how you can squash several git commits (incremental changes) into a single one.

This isn't how CRIU works today, nor is it specific to stream mode. criu-image-streamer is intentionally agnostic to image contents, its job is just to stream image data, not parse or modify pagemap and pages files. 

When we run criu pre-dump and then criu dump, CRIU produces two separate sets of memory images (with pagemap entries in the final dump referencing the parent images for unchanged pages). With --auto-dedup enabled, outdated pages in the parent images are already punched out, so we don't keep duplicate copies. Still, on restore, CRIU must read both sets of images to reconstruct the full address space.

The reason we don't "squash" them into a single image set is simple: reading all the unchanged memory pages from the parent images and rewriting them into the final image set would be very expensive and time-consuming.

PR #3149, opened 2026-09-17
uffd: multithreaded lazy-pages background drain

### LINKED ISSUE #3148: multi-threaded VMA restore
it would be nice if  we could populate memory on restore with multiple threads, so we can exceed the ~1gbyte/sec speed limit for restoring data.

I have a set of changes for CRIU that I had Claude write for me that seems to work for this. I see a 2x speed improvement in practice (3x if the data comes from a remotely backed storage, where access latency is larger).  I was wondering what the best way is to go forward with these patches.  I do have an overall understanding of what is going on, but not sure if that is enough for taking this through code review.

I'm attaching the Claude plan for reference.

[multithreaded-lazy-pages-design.md](https://github.com/user-attachments/files/32244012/multithreaded-lazy-pages-design.md)

hanwen-flow:
test.sh has a quick test, dumping and restoring a process with a 10G heap, measuring how long the lazypages server takes to complete. On my AMD Ryzen 7 PRO 8840HS:

   * Criu 4.2:

     0.02user 3.00system 0:03.16elapsed 95%CPU (0avgtext+0avgdata 12324maxresident)k 0inputs+0outputs (0major+3202minor)pagefaults 0swaps

   * this change:

     0.04user 13.10system 0:01.43elapsed 914%CPU (0avgtext+0avgdata 72080maxresident)k 1528inputs+0outputs (0major+17065minor)pagefaults 0swaps

This is a noticeable (2x) improvement, at 10x CPU cost. Some more tuning is clearly needed.

Claude: 

The lazy-pages daemon fills a restored task's remaining lazy ranges serially after restore finishes: one epoll loop reads a range into a single staging buffer and issues one UFFDIO_COPY, topping out around a single thread's memory bandwidth.

UFFDIO_COPY is safe for concurrent callers on one userfaultfd (racing fills lose with EEXIST, already tolerated), so parallelize the background drain of a task backed by a raw local pages image. A pool of worker threads (sized like asyncd, min(cpus, 16)) fills disjoint ranges through the same uffd at once.

The main epoll thread keeps sole ownership of all page_read cursor state and all lazy_iov list surgery: it resolves each range's pages image file offset, hands workers self-contained {fd, offset, dst, nr} jobs, waits for the batch and accounts the results. Workers only pread() image content and issue the ioctl; they never touch the page_read or the iov lists. Draining one batch between epoll waits keeps demand faults and non-cooperative events serviced.

Tasks whose content is remote, compressed, streamed or backed by a parent chain, and the demand-fault path, keep using the existing serial code, as does any run where the pool fails to start.

See multithreaded-lazy-pages-design.md for the full design.

Assisted-by: Claude Code:claude-opus-4-8
Signed-off-by: Han-Wen Nienhuys <hanwen@engflow.com>
Fixes #3148


hanwen-flow:
Uploaded the last version of this for anyone who is curious. It looks like lz4 support is a better fit, though.

Issue #2503, opened 2024-10-24
Error with the combination of page-server and iterative migration


obsidian0215:
**Description**
I want to iteratively migrate a RunC container with page-server, so I tried to use pre-dump with --page-server. However, I found that criu finally dump the entire container which the pre-dumped pages transferred to destination by page-server did not effect the C/R process. I wonder how to combine --page-server with multiple pre-dump and a final dump.

**Steps to reproduce the issue:**
1. At destination, start a page server for pre-copy: `criu page-server --images-dir /runc/containers/redis-pre-d/parent0 --port 27 --auto-dedup -v4 -o /home/debian/CRIU-xiezuo/v4.0/dirty-track/mig-scripts/logs/ps.log`
2. At source, run a pre-copy iteration for the container "redis-pre-d": `runc checkpoint --pre-dump --page-server 192.168.37.154:27 --image-path parent0 redis-pre-d`
3. At source, checkpoint the container "redis-pre-d": `runc checkpoint --image-path image --shell-job --tcp-established --parent-path ../parent0 redis-pre-d`
4. At destination, restore the container "redis-pre-d": `runc restore --console-socket /runc/containers/redis-pre-d/console.sock -d --image-path /runc/containers/redis-pre-d/image --work-path /runc/containers/redis-pre-d/image --shell-job --tcp-established redis-pre-d`

**Describe the results you received:**
At source, I found that criu didn't write anything in parent dir, that dumped the entire container.
```
runc checkpoint --pre-dump --page-server 192.168.37.154:27 --image-path parent0 redis-pre-d finished after 1177.880 ms with 0
the total size of /runc/containers/redis-pre-d/parent0 with pattern pages is 0Byte
runc checkpoint --image-path image --shell-job --tcp-established --parent-path ../parent0 redis-pre-d finished after 427.326 ms with 0
DUMP size: 153M	/runc/containers/redis-pre-d/image
Transferring DUMP to 192.168.37.154

sent 24.55M bytes  received 1.18K bytes  16.36M bytes/sec
total size is 159.80M  speedup is 6.51
DUMP transfer time 1037.512 ms
the total size of /runc/containers/redis-pre-d/image is 152.4MB
```
![image](https://github.com/user-attachments/assets/2836d76f-c821-4e14-a124-97ed930cdcef)

At destination, I got pagemaps and pages images in parent dir, which means page-server worked. However, the image dir still contained pagemaps and pages images which are bigger than those in parent dir.
[...]

PR #3150, opened 2026-09-18
page-xfer: support page-server pre-dumps as parents of local final dumps

obsidian0215:
This implements the page-server + iterative migration workflow reported in #2503.

With `--pre-dump --page-server`, the pre-dump pagemap and page payload are written on the page-server side. The source-side parent directory therefore has no normal parent pagemap for a later local final dump. When that final dump is run with `--parent-path`, CRIU cannot tell which ranges already exist in the remote parent and sends the memory payload again.

This series keeps compact source-side coverage metadata for page-server pre-dumps. It records both newly transferred ranges and ranges inherited from an earlier remote round, so the coverage describes the effective remote parent chain. If a later local dump cannot open a normal parent pagemap, `page_xfer` can use this metadata to validate `PE_PARENT` ranges without fetching page data back to the source.

Restore is unchanged: it still consumes the ordinary destination-side parent image chain. The coverage files are source-side coordination metadata and are not part of the restore image ABI.

Coverage files are written under temporary names and published only after pre-dump finalization succeeds. Failed rounds remove metadata created by that round. The page server closes and flushes its output images before returning the final success status. Shared-memory pages use parent holes only when explicit shared-memory dirty tracking is enabled; otherwise their contents are copied.

The first two patches fix two read-mode pre-dump problems exercised by this workflow. The first keeps read and splice results signed and handles the first recovered iovec without accessing a preceding entry. The second prevents a `SPLICE_F_GIFT` buffer from being refilled while the pipe or socket path may still reference its pages.

Testing includes:

- `make unittest`
- `make -C test/others/page-server-remote-parent check`
- one and two page-server pre-dump rounds followed by a local final dump and restore
- read and splice pre-dump modes
- restored page-generation checks
- source and destination image-write failures
- page-server disconnects during transfer and before the final status reply
- malformed/conflicting coverage metadata and rollback paths
- shared-memory restore and aliasing with the default conservative fallback

I also reproduced the runc workflow from #2503 with a 256 MiB anonymous-memory workload, dirtying 1 MiB after the remote pre-dump. Upstream and patched CRIU were run three times each in alternating order, and every trial restored the container and verified the complete memory contents.

| Metric | Upstream base | This series |
| --- | ---: | ---: |
| Final page payload | 268,500,992 B | 1,069,056 B |
| Final dump wall time | 144 ms | 79 ms |
| Restore wall time | 107 ms | 113 ms |

The payload reduction is the expected effect of using the remote pre-dump as the final dump's parent. The timing numbers are from this focused local reproduction rather than a general performance benchmark.

Related: #2503


test/others/page-server-remote-parent/shared_regression.py
```diff
@@ -0,0 +1,293 @@
...
+            for iteration in (1, 2):
+                source = base / f'source-{iteration}'
+                target = base / f'target-{iteration}'
+                source.mkdir()
+                target.mkdir()
+                observed_dirty = {pid: dirty_private_pages(pid, private, private_size, page_size)
+                                  for pid in pids}
+                # Keep the listening socket bound; CRIU accepts the actual data
+                # connection via --ps-socket. No free-port/probe race is needed.
+                with socket.socket() as listener:
```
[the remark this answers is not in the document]
obsidian0215: Fixed in the current revision by importing the socket constructor explicitly as `Socket` and using `Socket(AF_INET, SOCK_STREAM)`. This also avoids the ambiguity reported by CodeQL. Thanks.

criu/remote-parent.c
```diff
@@ -0,0 +1,513 @@
...
+		return -1;
+	}
+
+	writer->have_pending = false;
+	return 0;
+}
+
+int remote_parent_writer_open(int fd_type, unsigned long img_id, struct remote_parent_writer **out)
+{
+	struct remote_parent_disk_header header = {
```
avagin: do we really need to introduce another image format?

obsidian0215: I added it because the source still needs to know which ranges were transferred to the page server, while the actual images are on the destination.

But I agree that a separate format is probably unnecessary. CRIU can reuse the existing pagemap format on the source and let the final dump use it for parent checks without requiring the page data locally.

I'll rework it that way.

avagin: could you please research if we can do a final dump without a parent image

PR #3153, opened 2026-09-21
Multithreaded PIE restore

hanwen-flow:
Proof of concept. For context, see https://github.com/checkpoint-restore/criu/issues/3148

This was generated Claude, Opus 4.8. The plan is in the first commit.


hanwen-flow:
A quick benchmark (10G heap on AMD Ryzen 7 PRO 8840HS, CPU @ 2Ghz) shows:

        this change
        0.00user 0.00system 0:01.49elapsed 0%CPU (0avgtext+0avgdata 7060maxresident)k
    88inputs+8outputs (1major+565minor)pagefaults 0swaps

        CRIU 4.2
        0.00user 0.03system 0:09.70elapsed 0%CPU (0avgtext+0avgdata 8436maxresident)k
    8inputs+8outputs (0major+2160minor)pagefaults 0swaps

    ie. a ~6x speedup.

avagin:
I would try to avoid any extra complexity in the PIE code. Could you try benchmarking a case where all private VMAs are restored from non-PIE code? I think you will need to do something like:
```
index 745235f58..55709a2e9 100644
--- a/criu/mem.c
+++ b/criu/mem.c
@@ -1202,7 +1202,7 @@ static int premap_priv_vmas(struct pstree_item *t, struct vm_area_list *vmas, vo
                        continue;
                }
 
-               if (vma->pvma == NULL && pr->pieok && !needs_premap && !vma_force_premap(vma, &vmas->h)) {
+               if (vma->pvma == NULL && !needs_premap && !vma_force_premap(vma, &vmas->h)) {
                        /*
                         * VMA in question is not shared with anyone. We'll
                         * restore it with its contents in restorer.
```

hanwen-flow:
See here https://github.com/checkpoint-restore/criu/pull/3161

The numbers are similar to the PIE version. I'll close this.

PR #3158, opened 2026-09-22
ci: do not assume the checkout directory is named criu

snowhiteohno:
## Problem

`scripts/ci/loongarch64-qemu-test.sh` archives the source tree with:

```sh
tar -cf criu.tar ../../../criu
```

The script is started from `scripts/ci` (`scripts/ci/Makefile`, target `loongarch64-qemu-test`), so `../../../criu` is *the parent of the source tree* plus a hard coded `criu` component. It resolves correctly only when the checkout directory happens to be called `criu`.

That holds on GitHub Actions, where `actions/checkout` names the workspace after the repository, so CI is not currently broken. It does not hold for a local run, or for a checkout under any other name.

Reproduced with a checkout named `criu-dev-checkout`:

```
$ cd criu-dev-checkout/scripts/ci
$ tar -cf criu.tar ../../../criu
tar: Removing leading `../../../' from member names
tar: ../../../criu: Cannot stat: No such file or directory
tar: Exiting with failure status due to previous errors
```

The script sets `errexit`, so the job stops there.

## Changes

Derive the top of the source tree from the location of the script:

```sh
CRIU_ROOT="$(cd "$(dirname "$0")/../.." && pwd)"
tar --exclude=.git -cf criu.tar -C "$CRIU_ROOT" .
```

and unpack into `/root/criu` explicitly, since the archive now holds the contents of the tree rather than a `criu/` directory. Everything after that, including the build and the test list, is unchanged.

`.git` is also excluded, since only the sources are needed to build and test inside the VM.

## What is deliberately not changed

The third party VM image, `sshpass` with the hard coded password, and the set of zdtm tests are all left alone to keep this to one problem.

## Testing

I cannot run the job end to end here: it needs Docker in Docker plus the loongarch64 VM image, and emulation is not available in this environment. So this is verified at the level of the code that changed.

With a checkout named `some-other-name`:

* `CRIU_ROOT` resolves to the source tree.
* `tar --exclude=.git -cf criu.tar -C "$CRIU_ROOT" .` succeeds, produces 5181
  entries and zero `.git` entries.
* Unpacking it the way the script now does leaves `Makefile`,
  `criu/crtools.c`, `test/zdtm.py` and `test/zdtm/Makefile` at `/root/criu`,
  which is what the following `make` and `zdtm.py` steps expect.

`shellcheck scripts/ci/loongarch64-qemu-test.sh` is clean.

Someone with the CI able to run the loongarch64 job should confirm it still passes before this is merged.


avagin:
lgtm

Issue #3106, opened 2026-07-25
can criu implement tar compression internally ？


ningmingxiao:
(00.065856) mnt: Dumping mountpoints
(00.065862) mnt:        159: 21:/ @ ./dev/shm
(00.065892) mnt:        158: 20:/755dee6e503064753207b4905038873cf16b26056bf5bef59dab6e8d9568126a-be5e0220e778981a-resolv.conf @ ./etc/resolv.conf
(00.065897) mnt:        157: 20:/755dee6e503064753207b4905038873cf16b26056bf5bef59dab6e8d9568126a-c3747c7d0fe784e3-hosts @ ./etc/hosts
(00.065901) mnt:        156: 20:/755dee6e503064753207b4905038873cf16b26056bf5bef59dab6e8d9568126a-289602c847374178-hostname @ ./etc/hostname
(00.065904) mnt:        155: 1a:/ @ ./sys/fs/cgroup
(00.065907) mnt:        154: 16:/ @ ./sys
(00.065910) mnt:        153: 1f:/ @ ./dev/mqueue
(00.066032) mnt:        152: 29:/ @ ./dev/pts
(00.066038) mnt:        151: 28:/ @ ./dev
(00.066040) mnt: Mount is not fully visible ./dev(151)
(00.066088) mnt:        mount has children ./dev(151)
tar: unrecognized option '--no-unquote'
BusyBox v1.36.0 (2026-05-05 08:00:00 CST) multi-call binary.

because --no-unquote is not supported on busybox
Could CRIU implement compression via C libraries instead of spawning the `tar` command? This would yield better compatibility for CRIU.

Utkal059:
The tar dependency is confined to two functions in `criu/filesystems.c` — `tmpfs_dump()` and `tmpfs_restore()`. It's only used for tmpfs/devtmpfs content, which lands in `tmpfs-dev-*.img`.

Worth noting `--no-unquote` isn't the only blocker. Dump runs tar with `--create --gzip --no-unquote --no-wildcards --one-file-system --check-links --preserve-permissions --sparse --numeric-owner --directory`, and busybox tar only implements `--create/--gzip/--numeric-owner/--directory` (plus `--same-permissions`). So dropping that one flag would just move the failure to the next one — and it isn't safe to drop anyway: it was added in 8360d9aa precisely so that arbitrary characters in path names survive the round trip.

[...]

PR #3159, opened 2026-09-22
docs: describe how tmpfs content is dumped and restored

snowhiteohno:
## Problem

The content of a tmpfs mount is the one part of a CRIU image set that CRIU does not produce itself: `criu/filesystems.c` runs `tar` and stores the resulting archive as `tmpfs-dev-<s_dev>.tar.gz.img`.

None of that is written down, so the image descriptors involved, the reason for each `tar` flag, the user namespace handling and the practical consequences of shelling out all have to be rediscovered from the code.

## Changes

Add `Documentation/under-the-hood/tmpfs-content.md` and link it from the index under "Files, Mounts & I/O". It covers:

* the two image descriptors, `CR_FD_TMPFS_DEV` and `CR_FD_TMPFS_IMG`, their file names and the fallback on restore that keeps older image sets working;
* the exact `tar` invocations and what each flag is there for;
* why dump goes through `cr_system_userns()` while restore uses `cr_system()`;
* consequences of the approach: the runtime dependency on GNU `tar`, tmpfs content bypassing the CRIU pipeline and `crit`, `--xattrs` not being requested, and the coarse error reporting;
* what the existing `tempfs*` tests do and do not cover.

Everything is a description of current behaviour taken from the code. The page proposes no changes.

## Testing

`codespell` on the new file is clean. The change is documentation only, with no effect on the build or the tests.

## Related

This came out of looking at #3106, where the question of replacing `tar` with a native image format is being discussed. 
This PR is deliberately independent of that question, and is useful either way.

I did not include a row for the `tempfs_content` test proposed in #3157, so that this page stands on its own. If both land I am happy to send a one line follow up adding it.


Documentation/under-the-hood/tmpfs-content.md
```diff
@@ -0,0 +1,108 @@
...
+| descriptor | file name | key |
+| --- | --- | --- |
+| `CR_FD_TMPFS_DEV` | `tmpfs-dev-%u.tar.gz` | superblock device (`s_dev`) |
+| `CR_FD_TMPFS_IMG` | `tmpfs-%u.tar.gz` | mount id (`mnt_id`) |
+
+Both are opened with `O_NOBUF`, since the payload is written by an external
+process rather than through the usual image writing helpers.
+
+Dump always writes `CR_FD_TMPFS_DEV`. On restore, `tmpfs_restore()` tries
+`CR_FD_TMPFS_DEV` first and falls back to `CR_FD_TMPFS_IMG` when that image is
```
avagin: we need to explain that CR_FD_TMPFS_IMG has been deprecated and why it was deprecated.

Issue #3004, opened 2026-04-21
cr_veth test fails on 7.0.0 kernel in CI


adrianreber:
```
2026-04-20T20:18:39.8728790Z ======================== Run zdtm/static/cr_veth in ns =========================
2026-04-20T20:18:39.8729584Z Skipping rtc at root
2026-04-20T20:18:39.8729981Z Start test
2026-04-20T20:18:39.8730269Z Test is SUID
2026-04-20T20:18:39.8730727Z ./cr_veth --pidfile=cr_veth.pid --outfile=cr_veth.out
2026-04-20T20:18:39.8731334Z     PID TTY      STAT   TIME COMMAND
2026-04-20T20:18:39.8731796Z       1 ?        S+     0:00 python3 zdtm.py
2026-04-20T20:18:39.8732292Z       3 ?        S+     0:00  \_ python3 zdtm.py
2026-04-20T20:18:39.8732965Z      69 ?        S+     0:00      \_ make --no-print-directory -C zdtm/static cr_veth.pid
2026-04-20T20:18:39.8733830Z     104 ?        S+     0:00      |   \_ ./cr_veth --pidfile=cr_veth.pid --outfile=cr_veth.out
2026-04-20T20:18:39.8734976Z     105 ?        Ss     0:00      |       \_ ./cr_veth --pidfile=cr_veth.pid --outfile=cr_veth.out
2026-04-20T20:18:39.8735871Z     107 ?        S      0:00      |           \_ ./cr_veth --pidfile=cr_veth.pid --outfile=cr_veth.out
2026-04-20T20:18:39.8736787Z     108 ?        Ss     0:00      |               \_ ./cr_veth --pidfile=cr_veth.pid --outfile=cr_veth.out
2026-04-20T20:18:39.8737841Z      70 ?        S+     0:00      \_ /bin/bash zdtm/static/cr_veth.hook --post-start
2026-04-20T20:18:39.8738522Z     111 ?        R+     0:00          \_ ps xaf
2026-04-20T20:18:39.8738964Z -> 105
2026-04-20T20:18:39.8739240Z + ip l l
2026-04-20T20:18:39.8739953Z 1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1000
2026-04-20T20:18:39.8740928Z     link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
[...]

PR #3160, opened 2026-09-22
tmpfs: preserve extended attributes of tmpfs content

snowhiteohno:
## Problem

Extended attributes on files inside a tmpfs mount are silently lost across checkpoint/restore.

The content of a tmpfs mount is dumped and restored with `tar` (`criu/filesystems.c`, `tmpfs_dump()` and `tmpfs_restore()`). Neither invocation passes `--xattrs`, and GNU tar does not store extended attributes by default.

I raised this as a possibility in #3106 but flagged it as unverified. It is now confirmed, at two levels.

**1. The tar round trip.** Using exactly the flags CRIU passes today:

```
$ setfattr -n user.myattr -v myvalue /mnt/src/file
$ getfattr -d -m - /mnt/src/file
user.myattr="myvalue"

$ tar --create --gzip --no-unquote --no-wildcards --one-file-system \
      --check-links --preserve-permissions --sparse --numeric-owner \
      --directory /mnt/src . > t.tgz
$ tar --extract --gzip --no-unquote --no-wildcards --directory /mnt/dst < t.tgz

$ getfattr -d -m - /mnt/dst/file
(nothing)
```

Adding `--xattrs` to the same two commands preserves the attribute.

**2. A real checkpoint/restore.** The new `zdtm/static/tempfs_xattr` test fails on current `criu-dev`:

```
FAIL: tempfs_xattr.c:63: Can't read user.zdtm back:
      (errno = 61 (No data available))
```

## Changes

Pass `--xattrs --xattrs-include=user.*` on both the dump and the restore side.

The include pattern deliberately limits this to the `user.*` namespace. `security.*` is left alone so this does not interact with the LSM handling in `criu/lsm.c`, or with a tmpfs mounted using a `context=` option, and `trusted.*` needs privileges that are not obviously appropriate to replay here. If you would rather see a wider or narrower scope, that is an easy change.

The second commit adds the regression test.

## Compatibility

I checked the three cases that seemed most likely to cause trouble. All tested with GNU tar 1.35.

| case | result |
| --- | --- |
| existing image (no xattr records) restored by a CRIU with this change | extract exits 0, unchanged behaviour |
| image written with this change restored by an older CRIU | extract exits 0, the extra records are ignored |
| restore onto a filesystem that cannot store the xattr | tar warns on stderr and **still exits 0**, so the restore succeeds and file content is intact |

The third case is the one that matters for cross kernel migration, since tmpfs only gained `user.*` xattr support in Linux 6.6. I verified it by extracting an archive whose only failing operation was the xattr write:

```
tar: setxattrat: Cannot set 'security.zdtmtest' extended attribute
     for file './f': Operation not permitted
exit=0
```

So a dump taken on a newer kernel still restores on an older one. It logs a tar warning and drops the attribute, which is the current behaviour anyway.

This adds no new dependency. CRIU already requires GNU tar, since `--no-unquote`, `--check-links`, `--sparse` and `--numeric-owner` are all GNU extensions, and `contrib/dependencies/apk-packages.sh` and `pacman-packages.sh` already install it explicitly.

## Testing

Run in a privileged container on x86_64, Ubuntu 24.04, kernel 7.0.0-31-generic, GNU tar 1.35, using the container options from `scripts/ci`.

```
zdtm/static/tempfs_xattr          -f ns    FAIL before, PASS after
zdtm/static/tempfs_xattr          -f uns   PASS after
zdtm/static/tempfs                -f ns    PASS
zdtm/static/tempfs_ro             -f ns    PASS
zdtm/static/tempfs_ro02           -f ns    PASS
zdtm/static/tempfs_subns          -f ns    PASS
zdtm/static/tempfs_overmounted    -f ns    PASS
zdtm/static/tempfs_overmounted01  -f ns    PASS
zdtm/static/mountpoints           -f ns    PASS
```

`make unittest` is OK and `make indent OPTS=--diff BASE=origin/criu-dev` is clean.

The test skips itself if the kernel reports that extended attributes are not supported, so it should not fail on pre 6.6 kernels.

## Related

Follows up the fourth point in my comment on #3106. Independent of #3157 and #3159, which touch only tests and documentation.


snowhiteohno:
Pushed a fix for the code linter failure and amended it into the test commit.

**Linter.** My mistake. `test/zdtm/static/tempfs_xattr.c` used

```c
fail("Can't read %s back: %m", XATTR_NAME);
```

which trips the `Do not use %m with pr_* or fail` rule in the `lint` target.
The `fail()` macro already appends `(errno = %d (%s))`, so the `%m` was both
redundant and forbidden. Dropped it.

I had run the individual linters locally but not the full `make lint` target,
so I missed the git-grep rules at the end of it. I have now run the complete
target in a Fedora container the way the workflow does, and it exits 0. I also
checked the other rules in that group (`\n` appended to `fail`, `errno` passed
to `fail`) across all the files I have touched, and there are no other hits.

The test still passes in both flavors after the change:

```
zdtm/static/tempfs_xattr -f ns    PASS
zdtm/static/tempfs_xattr -f uns   PASS
```

**CentOS Stream 10.** This one looks unrelated to the change, but I would
rather flag it than quietly ignore it. The failing test is
`zdtm/static/cr_veth02`:

```
Error (criu/util.c:1722): ip6tables-legacy-restore -V failed
FAIL: cr_veth02.c:49: Net config differs after restore
```

The restored veth is missing its `192.168.124.124/32` address.

`cr_veth02` is a veth and external netdev test. Its source contains no
reference to tmpfs, mounts or extended attributes, and this PR only changes
the `tar` flags used for tmpfs content plus adds a new zdtm test, so I cannot
see a mechanism by which it would be affected. In the same job,
`zdtm/static/tempfs_xattr` itself passed in both `ns` and `uns`.

There is also a previously closed report of `cr_veth` failing on 7.0.0
kernels in CI, #3004, which is the kernel this runner is on.

I have not been able to reproduce it locally, since I do not have the CentOS
Stream 10 Lima environment here. If it reproduces on the re-run triggered by
this push I will dig further, but if it is environmental please let me know
and I will leave it alone.

avagin:
lgtm

PR #3161, opened 2026-09-22
Multithreaded host restore

hanwen-flow:
proof of concept for #3148 . Claude plan included.


hanwen-flow:
@avagin - how is this? If you agree with the general direction, I can polish this up for general consumption.

maqiao-mq:
hi, @hanwen-flow @avagin 

Would it be better to offload the mem fill to a dedicated thread pool, instead of doing it on the host side?
This is somewhat like combining PR [#3119](https://github.com/checkpoint-restore/criu/pull/3119) and PR [#3149](https://github.com/checkpoint-restore/criu/pull/3149).

First, start a dedicated mem fill offload process. It is essentially a UDS server plus a thread pool.

Then, each process being restored just sends the full iovec array to the offload process over the UDS. It can send iovec array at any time, host or PIE.
Inside the offload process, the array is split and the data is fetched in parallel to fill the target process. There are several ways to do the fill:

1. `process_vm_writev` — involves one extra copy, but is general enough.
2. `uffd UFFDIO_MOVE` — saves one copy, but does not work for some VMAs, and requires Linux v6.8+.
3. Shared memfd — only works for shmem.

Benefits:
1. CRIU can abstract a unified interface that serves both extmem and concurrent restore acceleration.
2. Decoupling: no need to add more code to the existing mem restore path.
3. The thread pool size is controllable, and we don't need to worry about the pool's PIDs occupying the PIDs needed by the restored process tree (the pool lives outside the pid ns, not inside).

hanwen-flow:
i'm closing this. The lz4 support works for me.

avagin:
> i'm closing this. The lz4 support works for me.

You are right we support parallel memory restore when compression is enabled, so the next steps are to extend it to uncompressed memory and improve the worker scheduling. Currently, it works in batches-assigning tasks to workers, kicking them, and waiting for all workers to complete. Ideally, we should switch to a task queue where workers continuously pull and execute tasks without synchronization points that stop all workers.

avagin:
> hi, @hanwen-flow @avagin
> 
> Would it be better to offload the mem fill to a dedicated thread pool, instead of doing it on the host side? This is somewhat like combining PR [#3119](https://github.com/checkpoint-restore/criu/pull/3119) and PR [#3149](https://github.com/checkpoint-restore/criu/pull/3149).
> 
> First, start a dedicated mem fill offload process. It is essentially a UDS server plus a thread pool.
> 
> Then, each process being restored just sends the full iovec array to the offload process over the UDS. It can send iovec array at any time, host or PIE. Inside the offload process, the array is split and the data is fetched in parallel to fill the target process. There are several ways to do the fill:
> 
> 1. `process_vm_writev` — involves one extra copy, but is general enough.
> 2. `uffd UFFDIO_MOVE` — saves one copy, but does not work for some VMAs, and requires Linux v6.8+.
> 3. Shared memfd — only works for shmem.
> 
> Benefits:
> 
> 1. CRIU can abstract a unified interface that serves both extmem and concurrent restore acceleration.
> 2. Decoupling: no need to add more code to the existing mem restore path.
> 3. The thread pool size is controllable, and we don't need to worry about the pool's PIDs occupying the PIDs needed by the restored process tree (the pool lives outside the pid ns, not inside).

Overall this approach could make sense, but in practice things are more complicated for a few reasons:
* CRIU still needs to support restoring processes directly in the current PID namespace (not just in a new PID namespace).
* userfaultfd is not available in all environments (for example, during unprivileged restore).
* We cannot always defer or offload memory population arbitrarily. In some cases, we must populate the contents of private VMAs at a specific stage before fork() or creating subsequent VMAs so that Copy-on-Write (CoW) pages are properly shared across processes.
* Using process_vm_writev from an external process can actually be slower than restoring VMAs directly within the target process.

